Issue #137, opened 2010-02-25
Bundler.autoload method


indirect:
`Bundler.autoload` would work the same way as `Bundler.require`, but would autorequire the relevant library files instead of requiring them.

PR #1169, opened 2011-05-14
Bundler.autoload(*groups)

nevir:
Hopefully knocks out #137

Allows you to swap out a `Bundler.require` call with `Bundler.autoload` and things should work (with maybe some Gemfile tweaking)

Added a gem option of `:autoload => 'Symbol' || [:Symbols]` to specify which symbols should cause that gem to autoload.  `:autoload => false` forces the gem to be _required_ and `:require => false` keeps the gem from being autoloaded or required.

The one caveat is that if someone else overrides `Object.const_missing` and forgets to proxy the call back, this facility will break.  (Doesn't seem like a huge concern - most libraries seem to be good about aliasing those calls)

I also have it merged into the [1.0 branch over on my fork](https://github.com/nevir/bundler)


gerrywastaken:
Wow, looking at an 18-19 second bundle.require, this makes me so sad. I love the thorough discussion you both went through though. Thank you. :)

PR #1820, opened 2012-04-04
Use smart http on :github dsl option

etehtsea:


c9s:
please allow ssh+git protocol to be configured for GitHub. With https you have to enter the username password for the private repository.
and with ssh+git clone, it is actually faster than https

PR #2569, opened 2013-07-27
Use https:// instead of git:// on GitHub URL's

fuadsaud:
See #1820


c9s:
we should allow ssh+git protocol to be configured for GitHub. With https you have to enter the username password for the private repository.
and with ssh+git clone, it is actually faster than https

PR #2802, opened 2014-01-05
Add GPL license support for new gems using the --license cli parameter.

andrein:
I've added a --license switch to the gem command since I prefer to use the GPL instead of the MIT license.

When running `bundle gem GEM --license GPL` it creates a COPYING.txt file with the GPL license and sets the license in the gemspec accordingly.

I can add more licenses, if this gets merged (LGPL, AGPL, BSD?)

Thanks,
Andrei


mvz:
Since `bundle gem` has `--mit`, why not add `--gpl`?!

mvz:
Or, the other way 'round, drop `--mit` as well for full minimalism :slightly_smiling_face:.

fwolfst:
Agreed. Also, I do not fully understand the proposed path

> Our gem generator is deliberately minimal—we think that features and options in a gem generator are best done in a library outside of Bundler.

How would that look like?
And I can understand the reasoning, but adding a license is often the first thing I (have to) do after using the otherwise perfectish `bundle gem` generator.

PR #2812, opened 2014-01-09
Don't remember install options

a-chernykh:
Work in progress for https://github.com/bundler/bundler-features/issues/26 here. But with some failing specs. Would appreciate a quick review.


asymmetric:
I would be happy to take a stab at this, provided three's still interest in the feature.

I tried rebasing @andreychernih's branch, but it's too far away in the past, so I'd have to start from scratch (using his work as a guide)

Issue #2556, opened 2013-07-22
Bundler falls back to legacy mode when you have more than 100 gems in your Gemfile.


judofyr:
Example Gemfile: https://gist.github.com/whitequark/cfe5e5f70ef22a28cab5 (donated by @whitequark)

https://github.com/bundler/bundler/blob/6423da7150b880a68d8378c200de11f8c0a4b070/lib/bundler/source/rubygems.rb#L229

``` ruby
if idx.size < API_REQUEST_LIMIT && dependency_names.size < API_REQUEST_LIMIT
  # use legacy mode
else
  # use the new API
end
```

I don't quite understand this code. The comment talks about "non-api sites", but in this example there are no "non-api sites"; there's just good ol' rubygems.org.

Issue #2448, opened 2013-04-23
Regression in 1.3.0 always fetches source index


bkeepers:
As of 4e7a5e91f98d24ce33da962d25eef69f10a8bf88 (/cc @indirect), a project with > 100 dependencies will always fetch the full source index, even when only one dependency changes.  From the commit, it appears that this is by design.  Could someone give a little more context for this decision? It seems like it should be possible to resolve the new dependency using the cached gemspecs and a few requests to the dependency api.

Related: I still need to debug it, but fetching the full source index now takes about 20 minutes on my machine, all while taking up 100^ % cpu and ballooning in memory usage. In the mean time, I've rolled back to 1.2.5.

PR #2863, opened 2014-02-07
Increase API_REQUEST_LIMIT to 750.

burke:
The current behaviour of switching to the modern index when asked for >100 gems is quite wasteful.
1. `specs.4.8.gz` is fetched from S3 at 1.6MB (this part is okay).
2. Each of the 70,000 gems' metadata is [immediately instantiated as an object](https://github.com/bundler/bundler/blob/master/lib/bundler/fetcher.rb#L187-L191).
3. An [expensive traversal](https://github.com/bundler/bundler/blob/master/lib/bundler/index.rb#L104-L114) is done of these 70,000 specs.

In browsing history, it looks to me like the current limit of 100 was chosen somewhat arbitrarily, and based on a very old index size (ie. O(n^2)ish on 10000 is a lot cheaper than O(n^2)ish on 70000). Steps 2 and 3 each take ~7 seconds, running on a 2.3GHz i7. Increasing this limit brings my total resolve time down to ~6 seconds.

We (Shopify) likely have one of the higher gem counts going, at ~275 in total including all indirect dependencies. I propose changing this value to at least 750. I haven't benchmarked this high, but even assuming completely linear scaling, we're still looking at 18 seconds with this change vs. 30 seconds without at 750 gems.

Related reading: #2448 and #2556.

It's entirely possible I'm misunderstanding what's going on here, or the motivation behind this `API_REQUEST_LIMIT`, but this _definitely_ speeds up bundler runs substantially on our project. Discussion welcome.

### DIFF
--- a/lib/bundler/source/rubygems.rb
+++ b/lib/bundler/source/rubygems.rb
@@ -5,7 +5,7 @@ require 'rubygems/spec_fetcher'
 module Bundler
   class Source
     class Rubygems < Source
-      API_REQUEST_LIMIT = 100 # threshold for switching back to the modern index instead of fetching every spec
+      API_REQUEST_LIMIT = 750 # threshold for switching back to the modern index instead of fetching every spec
 
       attr_reader :remotes, :caches
       attr_accessor :dependency_names


wlipa:
In lieu of a config setting which doesn't seem to have happened, I suggest retuning this a little.  100 is not a lot of gem dependencies these days.  I measured a 4x slowdown for the gem resolution step of 'bundle install' on my Rails app (125 gem dependency_names).  This is definitely quite noticeable as it hits on every capistrano deploy with a gem change (24 seconds vs. 6).

My suspicion is that more users are in a low latency environment than a high latency one so overall the setting of 100 is suboptimal.

PR #2902, opened 2014-02-24
changing plain-git to https

mrphs:
It's much wiser and a lot more secure to use HTTPS protocol instead of plain-git.


steakknife:
git:// is a completely different protocol from ssh:// or remotes that contain an `@`.

[See the list here](https://gist.github.com/steakknife/5289e3aaf452937978f4)

PR #2942, opened 2014-03-26
Adds a --no-install flag to bundle package

d-reinhold:
This PR is a rough draft of the bundle package --no-install flag suggested in this issue: https://github.com/bundler/bundler-features/issues/43

We'd like some feedback on how to better integrate this feature. This patch works and vendors the gems without installing them, but we don't have the familiarity with bundler to know if there are workflows we haven't considered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Bugfixes:
 Features:
 
   - warn informatively when `bundle install` is run as root (#2936, @1337807)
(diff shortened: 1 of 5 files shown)


TimMoore:
@d-reinhold I know this feature has been sitting on master unreleased for quite a long time, but a few bugs have been reported in it already.

See #3003 and the comment thread in #3327. Are you able to help fix these? Thanks!

PR #3115, opened 2014-07-27
nicer default gemspec

grosser:
- simpler require / version should always be loadable without loadin any other dependencies
 - use normal quoting
 - no need for encoding unless you add weird characters / no need for it on 2.0+
 - require_path is the default, no need to add it


segiddins:
@indirect can I merge this into 2-0-dev?

indirect:
@segiddins Keep the %{} strings, because it allows both kinds of quotes inside the docstrings. The other changes are :+1:

segiddins:
Applied manually to 2-0-dev in 24789a1

PR #3123, opened 2014-08-02
prevent naming mistake

tonytonyjan:
These method is borrowed from `rails new`
ref: https://github.com/rails/rails/blob/master/railties/lib/rails/generators/rails/app/app_generator.rb#L297

### DIFF
--- a/lib/bundler/cli/gem.rb
+++ b/lib/bundler/cli/gem.rb
@@ -36,6 +36,7 @@ module Bundler
         :test             => options[:test],
         :ext              => options[:ext]
       }
+      valid_name?(opts)
 
       templates = {
         "Gemfile.tt" => "Gemfile",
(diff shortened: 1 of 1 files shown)


rohit:
Firstly, the build is passing the one incomplete run is because of a timeout issues as I understand it.

Regarding the PR, the call to `valid_name?(opts)` looks a bit awkward to me. I mean the method name it self. Maybe something along the lines of `ensure_safe_gem_name(opts[:name], opts[:constant_array])`. I'm against passing the whole `opts` object too so I think passing just the `name` and `constant_array` should do?

Don't make any changes unless one of the maintainers confirm/deny my suggestion which is based purely on taste. :grin:

BM5k:
bump?

indirect:
@bm5k want to address the issues mentioned?

BM5k:
@indirect do you have an opinion on @rohit’s suggestion?

indirect:
yeah, it sounds good to me.

TimMoore:
These changes were merged from #3424

PR #3237, opened 2014-11-06
Fix rubygems 2.2 compatibility

voxik:
This is follow up of #3234

### DIFF
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -130,6 +130,19 @@ module Bundler
       yield
     end
 
+    def loaded_gem_paths
+      # RubyGems 2.2+ can put binary extension into dedicated folders,
+      # therefore use RubyGems facilities to obtain their load paths.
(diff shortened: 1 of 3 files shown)


indirect:
@voxik I'm happy to merge this as soon as Rubygems fixes rubygems/rubygems#784, but we can only allow these changes to happen in versions of Rubygems that have fixed that bug.

This patch needs to either: 1) work around the rubygems issue by not activating unless the rubygems issue is already fixed or 2) fix the rubygems issue via monkeypatching for the exact versions of 2.2 that are broken. Is either one of those feasible for you?

steakknife:
Submitted https://github.com/rubygems/rubygems/pull/1137 to unblock this.  Credit to @voxik

indirect:
ping @steakknife @voxik can either of you rebase this and check the rubygems version for a compatible one before applying the fix?

steakknife:
@indirect @voxik Can you guys confirm that a patch is even needed?  Latest RubyGems@master reverted conflicting native/Ruby loading behavior.

PR #3277, opened 2014-11-26
cleanup tempdir

obfusk:
Hi,

I noticed bundler sometimes left (empty) temporary directories behind. Not quite the end of the world of course, but still. I'm not very familiar with the codebase and am therefore not quite sure this is the best way to fix it, so please tell me what you think.

\- Felix


scalp42:
Still running into the same issue.

PR #3291, opened 2014-12-09
Optimized bundle-exec

kirs:
(from https://trello.com/c/bf90QcsN/66-exec-as-soon-as-possible, originally it's @indirect's idea)

Right now the exec command loads the entire Gemfile, the entire lockfile, resolves the gemfile against the lockfile, and then execs. Instead, let's check $PATH for the exec command and exec instantly if it's there.


indirect:
Merged as fd0acaa7. Thanks!

PR #3313, opened 2014-12-17
Extended gem generation and generation settings

kirs:
As discussed in https://github.com/bundler/bundler/pull/3305#issuecomment-67024669, the gem generator should not generate default files that change the rights or responsibilities of gem authors without their explicit consent. Let's change the gem generator to ask gem authors what they want, and allow them to change it via config or --flags.

* Remove LICENSE from default gem
* Remove CODE_OF_CONDUCT from default gem
* Add --coc flag to generate CODE_OF_CONDUCT
* Add --mit flag to generate LICENSE
* On gem generation, if not set to true/false, ask user if they are willing to license their code permissively under the MIT license
* On gem generation, if not set to true/false, ask user if they are willing to add a Code of Conduct to their gem
* On gem generation, if not set to false/rspec/minitest, ask if the user would like to generate tests along with their gems. Save the answer as bundle config gem.tests.
* Ensure that --coc, --mit, --test flags overrule any config settings that may be set.

(from https://trello.com/c/1349U0gF/67-require-consent-when-generating-files-in-gems)

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -342,6 +342,9 @@ module Bundler
     method_option :test, :type => :string, :lazy_default => 'rspec', :aliases => '-t', :banner =>
       "Generate a test directory for your library: 'rspec' is the default, but 'minitest' is also supported."
 
+    method_option :mit, :type => :boolean, :banner => "Use MIT license"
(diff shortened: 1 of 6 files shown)


indirect:
merged as 25f55d96e79f34575fb45da27379091666d48f61. Thanks a lot!

PR #3314, opened 2014-12-17
add operators to Ruby version comparison

aderyabin:
Currently I can point Ruby version toughly:
```ruby
 ruby '2.1.0'
```

But sometimes we need to set flexible version. For example that Ruby is not so old:
```ruby
 ruby '>= 2.0.0'
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 Features:
 
+  - add operators to Ruby version comparison (@aderyabin)
   - add support for SVN sources (@msnexploder)
   - add metadata allowed_push_host to new gem template (#3002, @juanitofatas)
   - adds a `--no-install` flag to `bundle package`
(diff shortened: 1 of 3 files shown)


indirect:
Made a Trello card for a version of this feature that we can actually merge: https://trello.com/c/gAWOh4Jk

sheerun:
Is locking ruby version really necessary? If someone wants to lock ruby version, she can explicitly state ruby version in lockfile.

For example I wound't want to lock ruby version. I'm happy if anyone uses any ruby matching `~> 2.1.0`.

indirect:
@sheerun feel free to read the trello card and implement the feature!

sheerun:
@indirect My question is actually about that trello card.

indirect:
@sheerun yes, it is required to lock to a single ruby version. That's like saying "I'm happy if my application uses `~> rails 4.1.0`. It's great that you're okay with that, but you have only developed and run your tests against a single Rails version, and that is the version that needs to be used on other developer machines and on production machines.

PR #3316, opened 2014-12-18
Move dev deps to gemfile

e2:
- add note about not using dev deps in gemspec
- move build and test gems to Gemfile within suggested groups
- add "--without development" on Travis


indirect:
This pull request doesn't seem to have an explanation of what it changes, or why those changes are better than how things are now. It's also not possible to merge anymore.

If you'd like to propose a change, please open a feature request ticket in the bundler-features repo, or update this pull request to include an explanation of the point of this change, why it is a good idea, and how it improves things.

Thanks!

PR #3320, opened 2014-12-23
`bundle config --local gemfile /foo/bar/MyGemfile` now works

dholdren:
- `bundle config` can be used to set a custom gemfile name for a
  project (specifying an absolute path)
- using other bundle commands will now search up parent directories
  for a .bundle dir and use the gemfile specified in .bundle/config if
  it is set
- does not require a file named 'Gemfile' to exist
- Revert "Correct bundle-config manpage"
  This reverts commit 0a814573363d48f8cd860d672f1c935fe1d89fca.
- An added bonus: you can now both set and *delete* a global bundler
  setting from *anywhere*: `bundle config --delete foo`

### Manual Test:
In an empty directory `/foo/bar/`:
`echo "source 'https://rubygems.org'" > MyGemfile`
`echo "gem 'rake'" >> MyGemfile`
`bundle config --local gemfile /foo/bar/FooGemfile`
`bundle install`
`bundle show` # Should show the rake gem
`mkdir spec && cd spec && bundle show` # Should also show the rake gem

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -193,7 +193,13 @@ module Bundler
     end
 
     def root
-      @root ||= default_gemfile.dirname.expand_path
+      @root ||= begin
+                  default_gemfile.dirname.expand_path
+                rescue GemfileNotFound
+                  bundle_dir = default_bundle_dir
+                  raise GemfileNotFound, "Could not locate Gemfile or .bundle/ directory" unless bundle_dir
+                  Pathname.new(File.expand_path("..", bundle_dir))
+                end
     end
 
     def app_config_path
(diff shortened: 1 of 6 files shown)


indirect:
Yeah, this seems like a legitimate forwards-compatible change. Thanks!

dholdren:
@indrect awesome. thanks for merging.
FYI, I forgot to modify the changelog, in case you want to add this to the list of new features

indirect:
The changelog will get written at release time, so no worries.

ineiti:
This is really great - is it possible to add the usage of the new Gemfile also in ```require 'bundler/setup'``` ? Or am I missing something?

PR #3322, opened 2014-12-23
Change the initial version of new gems to 0.1.0

petedmarsh:
I noticed that the FAQ section of the Semantic Versioning 2.0.0 spec suggests that you should start versioning from `0.1.0` but bundler currently sets the version of new gems to `0.0.1` - this is a very small change that makes new gems start with `0.1.0` as their version.


indirect:
Works for me.

PR #3331, opened 2015-01-01
Small code improvement.

gfvcastro:


BanzaiMan:
You are changing the logic. Look up De Morgan's Laws, please. Besides, it is not clear how `unless` is superior to `if !` in these contexts.

SleeplessByte:
[This article](http://www.railstips.org/blog/archives/2008/12/01/unless-the-abused-ruby-conditional/) has a nice write-up on `unless` and it's (mis)use. Basically, avoid `if not`, _unless_ there is more than one conditional.

Apart from the apparent change in logic,

````Ruby
if !Bundler.settings[:path] && !options[:force]
````

is clearer than

````Ruby
unless Bundler.settings[:path] || options[:force]
````

But

````Ruby
unless File.directory?(path)
````

is clearer than

````Ruby
if !File.directory?(path)
````

gfvcastro:
Thanks all.

indirect:
In general, I'm okay trailing `unless` only when there is a single condition, so I'm going to pass on this PR. Thanks, though!

PR #3334, opened 2015-01-02
Added an option to create a pry console in the gem with a custom prompt

onetwopunch:
Though it is similar to bundle console with the pry option set, this feature allows for a customizable script and pry prompt. Having a custom prompt and shell in a gem is handy way to show collaborators that they are in the proper context. There could be many other uses for it as well. I found this method in the rugged gem and I thought it was more intuitive to have a shell script ship with the gem. Also you could speed up development by renaming `console` to `c` so you can just call `./bin/c` to hit the pry console.

This feature allows users to call:

    bundle gem coolgem --console
    ./bin/console

Now they are in a pry console with the following prompt:
    
    [coolgem] main > 

I explain the process in my blog: http://lowbrowcoder.me/2015/01/01/how-to-add-a-pry-console-in-your-ruby-gem-for-testing/

Please let me know if there are things you'd like me to add or change to make the merge.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -341,7 +341,7 @@ module Bundler
     method_option :ext, :type => :boolean, :default => false, :banner => "Generate the boilerplate for C extension code"
     method_option :test, :type => :string, :lazy_default => 'rspec', :aliases => '-t', :banner =>
       "Generate a test directory for your library: 'rspec' is the default, but 'minitest' is also supported."
-    method_option :console, type: :boolean, default: false, aliases: '-c', banner: "Generate a binary pry console that can be run by executing ./bin/console"
+    method_option :console, :type => :boolean, :default => false, :aliases => '-c', :banner => "Generate a binary pry console that can be run by executing ./bin/console"
 
     def gem(name)
       require 'bundler/cli/gem'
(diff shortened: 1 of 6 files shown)


rohit:
IMHO I don't think Bundler should favor a particular console. You can apparently already set the console called when running `bundle console` https://github.com/bundler/bundler/pull/2620. Does this help somehow?

onetwopunch:
Thanks for the response! My team is a big fan of using distributed scripts like this as opposed to using stock bundle config and more importantly `.consolerc`. The reason for this is that in gems that use extensions or need other kinds initialization or compilation, code can be in a visible shell script as opposed to a hidden rc file. This gives everyone who uses/tests the gem a similar experience. Though I agree that the method in place is very handy, I think there could be a use for this simple option that creates a shell script that encapsulates the console as well. 

That being said, if this is something that you'd like to include in Bundler, I would be more than happy to add options for irb and Ripl just like [#2620](https://github.com/bundler/bundler/pull/2620). Please let me know what you think.

indirect:
I guess we've independently come up with the idea of adding a `bin/console`. :) We're we're definitely going to do that, just not as part of this pull request (so that we can avoid a hard pry dependency). Thanks for this pull request proving the concept!

onetwopunch:
@indirect Glad to know! Let me know if you would like any help developing that feature :+1:

indirect:
It's already inplemented! You can try it out with `gem install bundler --pre`

PR #3335, opened 2015-01-03
use `Minitest::Test` instead of deprectaed `MiniTest::Unit::TestCase`

y-yagi:
`MiniTest::Unit::TestCase` was renamed to `Minitest::Test` in minitest 5.0.


simi:
You'll need to lock `minitest` in [gemspec](https://github.com/bundler/bundler/blob/master/lib/bundler/templates/newgem/newgem.gemspec.tt) to 5 also.

zzak:
@y-yagi ありがとうございます！

y-yagi:
@zzak Thanks!

Issue #3261, opened 2014-11-19
Regex parsing of YAML in config leads to 2^n growth of config file size


rpdillon:
Bundler writes the settings `-j4` to a config file.  If that config file does not have the setting already written, Bundler [reads the file using a regex](https://github.com/bundler/bundler/blob/1fe23219d42365693326b716eb3d9a50973a4137/lib/bundler/settings.rb#L154), but [writes it using the YAML module](https://github.com/bundler/bundler/blob/1fe23219d42365693326b716eb3d9a50973a4137/lib/bundler/settings.rb#L140).

For long, multi-line environment variables, such as those [written by the Heroku RGeo buildpack](https://github.com/cyberdelia/heroku-geo-buildpack/blob/master/bin/compile#L113) this leads to the the number of single quotes delimiting the variable to double on each write.  This circumstance arises in cases such as a Heroku deploy.

After 27 such deploys, the size of the resulting `.bundle/config` file is 128MB, which consumes ~2G of RAM, wedging a 512MB instance and preventing startup of the app.

[This gist](https://gist.github.com/rpdillon/8e7c40213e4dabd453b3) demonstrates the issue in isolation, using code pulled directly from Bundler.

Presumably, YAML is parsed using a regex to improve startup time in the common read-only case.  The change was made in [this commit](https://github.com/bundler/bundler/commit/156d6885fbb83146087dbc0e109b5c7434a7f0c5).

The issue is not present when the YAML in the config file is parsed using the YAML module.

PR #3338, opened 2015-01-06
Don't add extra quotes to quoted wrapped config values

aroben:
Long config values get wrapped to a second line. If they contain certain special characters they will also get surrounded by quotes. When this happened, we would add extra quotes to the value each time the config file was saved. 94fd250935314aee86759de3e9b0e98a709823ed fixed this issue for non-wrapped lines. Now it is fixed for wrapped lines as well.

/cc @indirect #3261

### DIFF
--- a/lib/bundler/settings.rb
+++ b/lib/bundler/settings.rb
@@ -155,9 +155,10 @@ module Bundler
     def load_config(config_file)
       valid_file = config_file && config_file.exist? && !config_file.size.zero?
       if !ignore_config? && valid_file
-        config_regex = /^(BUNDLE_.+): (?:['"](.*)['"]|(.+(?:\n(?!BUNDLE).+))|(.+))$/
+        config_regex = /^(BUNDLE_.+): (['"]?)(.*(?:\n(?!BUNDLE).+)?)\2$/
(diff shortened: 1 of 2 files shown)


rohit:
Looks like the build failures are due to timeouts. Can somebody restart the build?

aroben:
I think there are some non-timeout failures too, e.g., https://travis-ci.org/bundler/bundler/jobs/46099347:

```
  1) .bundle/config quoting saves quotes at beginning of value
     Failure/Error: expect(out).to eq("'something")
       
       expected: "'something"
            got: "''something"
       
       (compared using ==)
     # ./spec/commands/config_spec.rb:209:in `block (3 levels) in <top (required)>'
  2) .bundle/config quoting saves quotes surrounding value
     Failure/Error: expect(out).to eq("'something'")
       
       expected: "'something'"
            got: "''something''"
       
       (compared using ==)
     # ./spec/commands/config_spec.rb:221:in `block (3 levels) in <top (required)>'
```

These are new tests I added, so I need to figure out they would have passed without this PR or if there is an existing bug being demonstrated here. I hope to get back to this next week.

PR #3341, opened 2015-01-07
Implement --no-version flag

eapache:
Removes the SHA/revision from directories in vendor/cache.

See https://github.com/bundler/bundler-features/issues/76 for motivation and
explanation.

I suspect I probably need to update documentation (and possibly tests) etc. as well, but I figured I'd ask for feedback on the idea before putting in that work.


indirect:
The entire point of including the SHA is so that we know what version of the gem is being cached. If we don't know what version of the gem is being cached, Bundler can't tell if the code in `vendor/cache` is the right code or the wrong code. I don't think this is a workable change, sorry.

Issue #3296, opened 2014-12-10
bundler 1.7.7 changed source dependency API support check behavior


dpiddy:
I have a `Gemfile` with a `source` hosted at S3. It does not support the dependency API and only makes available the files necessary to support index-based usage. All other requests return a 403.

Starting with bundler 1.7.7, my source no longer works. I get an error like this:

```
Authentication is required for https://my-s3-source.local
Please supply credentials for this source. You can do this by running:
 bundle config https://my-s3-source.local username:password
```

I believe it is caused by [this rescue change](https://github.com/bundler/bundler/compare/v1.7.6...v1.7.7#diff-23f4d7633f08ecf9c4a66f7189e34f0cR280).

dpiddy:
Any update on this?

TimMoore:
I haven't had a chance to look into this yet, sorry.

hotgazpacho:
We're running into this as well, with sources pointing to internal geminabox servers.

schneems:
This is a blocker to upgrading Heroku to 1.7.x any ideas what caused the regression?

TimMoore:
@schneems the regression was caused by #3260, which fixed a problem using Gemfury as a private gem server. I can revert that change, but it will break things for other users. I would expect that to also block upgrading Heroku.

Fundamentally, the issue is that different gem servers have inconsistent behaviours with the meaning of a 403 status code. Some servers return it when the request is not authenticated, and re-sending the request with authentication fixes the problem, others return it to mean that they don't support the dependency API at all.

If anyone else would like to try a fix, that would be helpful... I haven't had much time for Bundler work recently.

dpiddy:
A start might be to explicitly document bundler's expectations around sources, considering things like:

* when the dependency API will be used
* how dependency API support is checked
* when falling back to a full index will happen

That would let gem repo providers know what to do to be compatible.

Also how about the ability to force behavior per-source? For example, if I know a source should only use the full index, perhaps I could specify that as an option to `source`.

TimMoore:
> Also how about the ability to force behavior per-source? For example, if I know a source should only use the full index, perhaps I could specify that as an option to source.

Yeah, I think that would be a good idea... even if everything were working correctly, it would save an extra request.

TimMoore:
I have a proposed fix for this over at #3342. If anyone who has run into this problem can test that branch and report back, that would be very helpful. The more the better, because I've seen a variety of different behaviors from different servers.

[...]

Issue #3180, opened 2014-09-17
bundle install fails on Ruby 1.9.3 if HTTP auth stored in bundle.config


plutino:
STR:
1. save private gem server credentials in ~/.bundle/config or with `bundle config` and remove any credentials in project Gemfile.
2. switch to Ruby 1.9.3 (tested on both rvm and chruby).
3. in Rails project directory do `bundle install`.

O:
Get a 401 error.  `bundle -V` shows that credentials are not added to to the HTTP request

D:
gems successfully installed as in other Ruby versions (tested on Ruby 2.1.2)

This happens on Bundler version 1.6.4-1.7.2

TimMoore:
Finally have a proposed fix for all known issues with source credentials at #3342.

TimMoore:
Fixed in 1.7.12.

PR #3342, opened 2015-01-08
Always use source credentials

### LINKED ISSUE #3297: When credentials are configured for a source, use them unconditionally
Right now (as of 1.7.9), it tries to make an unauthenticated request first, then retries with authentication if the first request fails.

This causes issues on some gem servers. For example, Gemfury supports the dependency API and allows a mix of public and private gems in the same repository. It does not forbid unauthenticated requests, but it returns partial information.

See discussion on #3180, #3260 and #3296.

TimMoore:
Previously, it tried to make an unauthenticated request first, then retried with authentication if the first request failed.

This causes issues on some gem servers. For example, Gemfury supports the dependency API and allows a mix of public and private gems in the same repository. It does not forbid unauthenticated requests, but it returns partial information.

Now configured credentials are always used.

Fixes #3297, #3296, #3180.

### DIFF
--- a/lib/bundler/fetcher.rb
+++ b/lib/bundler/fetcher.rb
@@ -95,8 +95,7 @@ module Bundler
       @api_timeout    = 10 # How long to wait for each API call
       @max_retries    = 3  # How many retries for the API call
 
-      @remote_uri = Bundler::Source.mirror_for(remote_uri)
-      @anonymizable_uri = AnonymizableURI.new(@remote_uri.dup) unless @remote_uri.nil?
+      @anonymizable_uri = resolve_remote_uri(remote_uri)
 
       Socket.do_not_reverse_lookup = true
       connection # create persistent connection
@@ -104,14 +103,14 @@ module Bundler
 
     def connection
       @connection ||= begin
-        needs_ssl = @remote_uri.scheme == "https" ||
+        needs_ssl = remote_uri.scheme == "https" ||
           Bundler.settings[:ssl_verify_mode] ||
           Bundler.settings[:ssl_client_cert]
         raise SSLError if needs_ssl && !defined?(OpenSSL::SSL)
 
         con = Net::HTTP::Persistent.new 'bundler', :ENV
 
-        if @remote_uri.scheme == "https"
+        if remote_uri.scheme == "https"
           con.verify_mode = (Bundler.settings[:ssl_verify_mode] ||
             OpenSSL::SSL::VERIFY_PEER)
           con.cert_store = bundler_cert_store
@@ -130,7 +129,7 @@ module Bundler
     end
 
     def uri
-      @anonymizable_uri.without_credentials unless @anonymizable_uri.nil?
+      @anonymizable_uri.without_credentials
     end
 
     # fetch a gem specification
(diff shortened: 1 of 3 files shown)


TimMoore:
If you came here looking to test this fix, here's what you need to do:

```
git clone https://github.com/TimMoore/bundler.git`
cd bundler
git checkout issue-3297-always-use-source-credentials
rake install
```

That will install it to your system gems. Alternatively, you can create a `dbundle` alias as described in [the `DEVELOPMENT.md` guide](https://github.com/bundler/bundler/blob/master/DEVELOPMENT.md#development-setup) to run it from source.

Thanks!

arronmabrey:
I will try and set aside some time later today and test this out.

lib/bundler/fetcher.rb
```diff
@@ -374,35 +371,35 @@ def bundler_cert_store
...
+    def resolve_remote_uri(uri)
+      add_configured_credentials(Bundler::Source.mirror_for(uri))
     end
 
-  private
+    def add_configured_credentials(uri)
+      auth = Bundler.settings[uri.to_s]
+      if auth
+        uri = uri.dup
+        uri.user, uri.password = *auth.split(":", 2)
```
TimMoore: Hmm... I just realised that this will now make credentials in config override ones in the `Gemfile`.

We never really clearly defined what would happen when you have both, but in the past I suppose it would have tried the ones in the `Gemfile`, then fail immediately if the original ones didn't work (it got a 401 or 403 response).

I guess, intuitively, I would expect that the ones in the `Gemfile` would override the ones in config, but what do other people think?

indirect: I would expect ENV, then Gemfile, then config file, myself, I think? But I’m not positive.

TimMoore: Hmm, yeah well right now ENV and config file are normalised to a single settings object.

will:
It works for me

```
➤ bundler version
Bundler version 1.7.3
➤ bundle update 
Fetching source index from https://redacted.s3.amazonaws.com/redacted/
Fetching additional metadata from https://rubygems.org/........
Resolving dependencies...
Using …
Your bundle is updated!


➤ gem install bundler
➤ bundle version
Bundler version 1.7.11
➤ bundle update 
Authentication is required for https://redacted.s3.amazonaws.com/redacted/.
Please supply credentials for this source. You can do this by running:
 bundle config https://redacted.s3.amazonaws.com/redacted/ username:password


➤ dbundle version
Bundler version 1.7.11-dev
# I edited version.rb to add -dev to be sure
➤ dbundle update
Fetching source index from https://redacted.s3.amazonaws.com/redacted/
Fetching additional metadata from https://rubygems.org/........
Resolving dependencies...
Using …
Your bundle is updated!
```

lib/bundler/fetcher.rb
indirect: Since ENV and config file are combined, let’s make them lower priority than the in-Gemfile credentials.

indirect:
Hooray! Thanks for all your work on this, @TimMoore.

arronmabrey:
:tada: @TimMoore :tada:

lib/bundler/fetcher.rb
TimMoore: OK with the code as merged I think it's the other way around. I raised a new issue for that: #3343

dpiddy:
Thanks! Looking forward to 1.7.12 being released with this fix.

indirect:
1.7.12 is released!

PR #3345, opened 2015-01-12
Fix s3 source url parsing / auth

pangloss:
The URI class will not actually parse a url in the format s3://AKHAJIHAFOW6WWJUV5RA:7UVCy7cYjEzwfwLwwQR/DlOdJ7V+c7GFWKZDn8yx9@bucket/ because s3 secrets are generated with characters like / in them that are invalid as a password in http simple auth.

I've made the corresponding change in a pull request in Rubygems: https://github.com/rubygems/rubygems/pull/1134

PS. the AWS keys above are based on (but not precisely) an actual key pair.

### DIFF
--- a/lib/bundler/s3_fetcher.rb
+++ b/lib/bundler/s3_fetcher.rb
@@ -11,26 +11,32 @@ module Bundler
     # Instead of taking a dependency on aws-sdk, use a method modeled on
     # the signing method in https://github.com/rubygems/rubygems/pull/856
     def sign(uri, expiration = default_expiration)
-      uri = uri.dup
-      unless uri.user && uri.password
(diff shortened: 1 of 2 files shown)


indirect:
Thanks for the patch! Unfortunately, Bundler can't use the Gem constant directly, due to the way different versions of Rubygems have highly varying APIs. In addition, Bundler doesn't read or respect .gemrc. Please update this pull request to get S3 credentials from the Gemfile or from the Bundler configuration system, and then it will be good to go.

pangloss:
Thanks for the feedback @indirect. We've decided to take a different approach with our gem hosting for the time being, so I'm going to step back from this. I hope someone else makes the minor tweaks it needs and it gets in. If not, perhaps I'll revisit in the future.

PR #3349, opened 2015-01-15
Fix the path to the Gemfile during evaluation.

cstrahan:
`Bundler.default_gemfile` was being used instead of the explicitly passed Gemfile path.

This prevents errors like the following when `gemfile` argument isn't `"$PWD/Gemfile"`:

    /nix/store/na27qjv9g32rg2nh98m51jd7ihrg7w7n-ruby-2.1.3-p0-bundler-1.7.9/lib/ruby/gems/2.1.3/gems/bundler-1.7.9/lib/bundler/shared_helpers.rb:26:in `default_gemfile': Could not locate Gemfile (Bundler::GemfileNotFound)
            from /nix/store/na27qjv9g32rg2nh98m51jd7ihrg7w7n-ruby-2.1.3-p0-bundler-1.7.9/lib/ruby/gems/2.1.3/gems/bundler-1.7.9/lib/bundler.rb:262:in `default_gemfile'
            from /nix/store/na27qjv9g32rg2nh98m51jd7ihrg7w7n-ruby-2.1.3-p0-bundler-1.7.9/lib/ruby/gems/2.1.3/gems/bundler-1.7.9/lib/bundler/dsl.rb:27:in `initialize'
            from /nix/store/na27qjv9g32rg2nh98m51jd7ihrg7w7n-ruby-2.1.3-p0-bundler-1.7.9/lib/ruby/gems/2.1.3/gems/bundler-1.7.9/lib/bundler/dsl.rb:9:in `new'
            from /nix/store/na27qjv9g32rg2nh98m51jd7ihrg7w7n-ruby-2.1.3-p0-bundler-1.7.9/lib/ruby/gems/2.1.3/gems/bundler-1.7.9/lib/bundler/dsl.rb:9:in `evaluate'
            from /nix/store/na27qjv9g32rg2nh98m51jd7ihrg7w7n-ruby-2.1.3-p0-bundler-1.7.9/lib/ruby/gems/2.1.3/gems/bundler-1.7.9/lib/bundler/definition.rb:25:in `build'
            from /nix/store/yzpjmgpc5fdhll3rbacq5r5ay9kjxxym-ruby-2.1.3-p0-bundix-0.1.0/lib/ruby/gems/2.1.3/gems/bundix-0.1.0/lib/bundix/cli.rb:55:in `expr'
            from /nix/store/hkqpgmrq1x2c8icjhl433d2yhkzy33rn-ruby-2.1.3-p0-thor-0.19.1/lib/ruby/gems/2.1.3/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
            from /nix/store/hkqpgmrq1x2c8icjhl433d2yhkzy33rn-ruby-2.1.3-p0-thor-0.19.1/lib/ruby/gems/2.1.3/gems/thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
            from /nix/store/hkqpgmrq1x2c8icjhl433d2yhkzy33rn-ruby-2.1.3-p0-thor-0.19.1/lib/ruby/gems/2.1.3/gems/thor-0.19.1/lib/thor.rb:359:in `dispatch'
            from /nix/store/hkqpgmrq1x2c8icjhl433d2yhkzy33rn-ruby-2.1.3-p0-thor-0.19.1/lib/ruby/gems/2.1.3/gems/thor-0.19.1/lib/thor/base.rb:440:in `start'
            from /nix/store/yzpjmgpc5fdhll3rbacq5r5ay9kjxxym-ruby-2.1.3-p0-bundix-0.1.0/lib/ruby/gems/2.1.3/gems/bundix-0.1.0/bin/bundix:4:in `<main>'

### DIFF
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -24,10 +24,12 @@ module Bundler
       @platforms       = []
       @env             = nil
       @ruby_version    = nil
+      @gemfile         = nil
       add_git_sources
     end
 
     def eval_gemfile(gemfile, contents = nil)
+      @gemfile = Pathname.new(gemfile)
       contents ||= Bundler.read_file(gemfile.to_s)
       instance_eval(contents, gemfile.to_s, 1)
     rescue SyntaxError => e
@@ -44,7 +46,8 @@ module Bundler
       path              = opts && opts[:path] || '.'
       name              = opts && opts[:name] || '{,*}'
       development_group = opts && opts[:development_group] || :development
-      expanded_path     = File.expand_path(path, Bundler.default_gemfile.dirname)
+      gemfile           = @gemfile || Bundler.default_gemfile
+      expanded_path     = File.expand_path(path, gemfile.dirname)
 
       gemspecs = Dir[File.join(expanded_path, "#{name}.gemspec")]
 
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -46,8 +46,7 @@ module Bundler
       path              = opts && opts[:path] || '.'
       name              = opts && opts[:name] || '{,*}'
       development_group = opts && opts[:development_group] || :development
-      gemfile           = @gemfile || Bundler.default_gemfile
-      expanded_path     = File.expand_path(path, gemfile.dirname)
+      expanded_path     = path_relative_to_gemfile(path)
 
       gemspecs = Dir[File.join(expanded_path, "#{name}.gemspec")]
 
@@ -133,7 +132,7 @@ module Bundler
     end
 
     def path(path, options = {}, &blk)
-      with_source(@sources.add_path_source(normalize_hash(options).merge("path" => Pathname.new(path))), &blk)
+      with_source(@sources.add_path_source(normalize_hash(options).merge("path" => path_relative_to_gemfile(path))), &blk)
     end
 
     def git(uri, options = {}, &blk)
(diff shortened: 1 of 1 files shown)


cstrahan:
One thing ea3ded94 doesn't address is when you don something like:

``` ruby
gem "pg", :path => "../pg"
```

`gemspec` looks for gemspecs relative to the Gemfile, so I believe the above path should also be relative to the Gemfile.

cstrahan:
The last commit (b233205) implements what I suggested earlier: all paths during Gemfile evaluation will be expanded relative to the Gemfile.

cstrahan:
One problem with b233205 - the path sources will be full paths, which leaks into the lock files and breaks the tests.

I'm not sure what the best way to fix this is. If we want to support gemfiles outside of `$PWD` (which, for a couple upcoming PRs, I believe we do), we need to handle the fact that paths relative to the Gemfile won't be the same as paths relative to `$PWD`. Using full paths fixes this, but then we need a way to drop back to paths relative to the Gemfile when we create the Gemfile.lock.

Could someone lend me a hand with this? I'm on IRC (as cstrahan) if someone would like to discuss this.

indirect:
Thanks for the patch! The change seems like an improvement in general, but it definitely needs to handle paths carefully. We've had a lot of issues with relative paths in the past, and it is absolutely required that relative paths in the Gemfile stay relative paths in the lock. They should be expanded for use while Bundler is running, but they should always be stored in their original form so that they stay compatible across machines. We can't merge a pull request that expands path sources unless it maintains the relative paths inside the lock file.

indirect:
To follow up here, if you can fix the tests, this is mergable, but if you can't it's not.

PR #3351, opened 2015-01-19
Allow vendor/cache override 

jnraine:
I found myself needing to temporarily maintain two Gemfile for a project and needed to maintain two separate vendor/cache directories. I was able to get around this by symlinking `vendor/cache` to the appropriate directory before running `bundle install` or by monkey-patching `Bundler.app_cache`. However, neither are ideal so I've added a flag to the install command:

```bash
$ bundle install --path Gemfile-rails4 --vendor-cache vendor/cache-rails4
```

It can also be set via environment variable:

```bash
$ BUNDLE_VENDOR_CACHE=vendor/cache-rails4 bundle install --path Gemfile-rails4
```

This PR doesn't have any tests – my specs didn't end up caching the gems for some reason. Perhaps I'm misusing the test helpers?

```ruby
it "caches gems at given path" do
  gemfile bundled_app("Gemfile"), <<-G
    gem 'rack'
  G

  bundle "install --vendor-cache vendor/cache-foo"
  expect(bundled_app("vendor/cache-foo/rack-1.0.0.gem")).to exist
end
```

Before I continue work on the tests, I wanted to gauge interesting in this feature. I'd be happy to add tests if there is interest in merging.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Features:
   - generates a `.consolerc` file with new gems and tries to load it on `bundle console` (@andremedeiros)
   - tries to find `gems.rb` and it's new counterpart, `gems.locked` (@andremedeiros)
   - Change the initial version of new gems from `0.0.1` to `0.1.0` (@petedmarsh)
+  - add support for setting cache path via --cache-path or BUNDLE_CACHE_PATH (@jnraine)
 
 Documentation:
   - add missing Gemfile global `path` explanation (@agenteo)
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -210,7 +210,7 @@ module Bundler
 
     def app_cache(custom_path = nil)
       path = custom_path || root
-      path.join("vendor/cache")
+      path.join(Bundler.settings.cache_path)
     end
 
     def tmp(name = Process.pid.to_s)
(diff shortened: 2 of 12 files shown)


indirect:
I think this is probably a good thing to make configurable. Can you change the name of the option to `cache-path`? In the docs, it's also worth noting that it can be set using `bundle config cache_path somepath/cache` as well as the CLI flag and the ENV variable. It also definitely needs working tests to get merged. :) Thanks!

jnraine:
OK, I've fixed most of the failed tests and added tests to cover `--cache-path`. The only failing tests on >=1.9.3 are also failing on master so I've left them alone. 

There are a lot of failing tests under 1.8.7, but it doesn't seem related to these changes. Do you know why that might be?

I left the command line option description as is because adding the `bundle config cache_path somepath/cache` made it longer than most other options. Is that where you wanted it to be added or is there somewhere more appropriate?

I've updated the changelog as well.

jnraine:
Woohoo! Thanks for adding this feature and cleaning up the implementation. I did not realize `Bundler::Settings` had all the env variable handling built in. Very cool. :smile::ok_hand:

jnraine:
I've tested it locally and it works great!

I had a question about `bundle install`. On my machine running `bundle install` causes the gems to be installed and then packaged into the cache path. For my use case, specifying the cache path using an environment variable is preferred, but does it make sense to allow it to be set via --cache-path option as well? Right now, that produces an error:

```shell
$ bundle install --cache-path vendor/cache-foo
Unknown switches '--cache-path'
```

indirect:
Since the `install` command is primarily about installing, and only updates the cache as a side thing, I thought it made more sense to ask people to run `package --cache path` or `config cache_path` before running `install`.

jnraine:
Sounds good to me. Fewer options is probably a good thing.

Issue #3353, opened 2015-01-19
Errors printed to STDOUT, not STDERR.


jnraine:
All errors rescued by [friendly_error.rb](https://github.com/bundler/bundler/blob/master/lib/bundler/friendly_errors.rb) are printed to STDOUT.

```shell
$ bundle install --foo-bar
Unknown switches '--foo-bar'
$ bundle install --foo-bar > /dev/null
$ bundle install --foo-bar 2> /dev/null
Unknown switches '--foo-bar'
```

Is this intentional? It tripped me up while writing specs for https://github.com/bundler/bundler/pull/3351 – the error messages were not being printed after the bundle command failed because the standard error buffer was empty: https://github.com/bundler/bundler/blob/master/spec/support/helpers.rb#L136

It's a pretty subtle issue that isn't likely to affect many people, but if it's unintentional behaviour it may be worth updating.

indirect:
Yeah, we should fix this.

jnraine:
I spend some time looking into this tonight and it's not as simple as I'd have hoped. 

`#with_friendly_errors` passes a string to `Bundler.ui.error` which calls `Bundler::UI::Shell#tell_me` with the string and `:red`. Eventually, the string is passed to `Thor::Shell::Basic#say` where it is formatted and printed to stdout. 

The tricky bit is that `Thor::Shell::Basic` doesn't provide a way to print a message to stderr except for `Thor::Shell::Basic#error`, which is intended for internal use by Thor only and doesn't provide colour or width formatting.

So, either we write methods that do the formatting and printing to STDERR outside of Thor, write a method that calls into the guts of Thor for formatting and then print it ourselves, or add a method / modify a method in Thor directly (e.g., `Thor::Shell::Basic#say_error` or `Thor::Shell::Basic#say("foo", error: true)`.

None of these options are great, but for the sake of doing the least work and not hacking dependencies, I think the second option is my favourite. Maybe something like:

```ruby
# lib/bundler/ui/shell.rb
def say_error(message = '', color = nil, force_new_line = (message.to_s !~ /( |\t)\Z/))
  buffer = @shell.prepare_message(message, *color)
  buffer << "\n" if force_new_line && !message.end_with?("\n")

  @shell.stderr.print(buffer)
  @shell.stderr.flush
end
```

With tests, we'll notice if Thor changes it's guts when `vendor/` is upgraded and can tweak the method at that time.

I'm all ears for a better solution, as this is a bit gross.

[...]

PR #3354, opened 2015-01-20
Issue #3353 - Print error messages to STDERR

jnraine:
This solution uses a less-than-ideal method to add Thor-formatted printing to STDERR. Because Thor doesn't offer this out-of-the-box, I've used Thor's internals to format a message and pass it to Thor's stderr.

Since this uses protected methods, changes under the hood in Thor may break this. If this happens, tests will fail.

I also added basic tests to ensure other print methods from `Bundler::UI::Shell` continue printing to STDOUT.

#3353

### DIFF
--- a/lib/bundler/ui/shell.rb
+++ b/lib/bundler/ui/shell.rb
@@ -14,23 +14,23 @@ module Bundler
       end
 
       def info(msg, newline = nil)
-        tell_me(msg, nil, newline) if level("info")
+        tell_stdout(msg, nil, newline) if level("info")
       end
 
       def confirm(msg, newline = nil)
-        tell_me(msg, :green, newline) if level("confirm")
+        tell_stdout(msg, :green, newline) if level("confirm")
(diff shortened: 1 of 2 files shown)


TimMoore:
I'm pretty uneasy about breaking the encapsulation of Thor like this. Can we make a change upstream to Thor to enable us to do this more cleanly?

indirect:
I am okay with this approach because Thor is effectively part of our local codebase, but I would also prefer that Thor offered `Shell#say_error` or something like that. @jnraine can you fix the tests for this change and/or submit a PR to Thor that adds a method to print to stderr?

segiddins:
Thanks @jnraine, this was cherry-picked into 2-0-dev as https://github.com/bundler/bundler/commit/300fcbe2a06b9b44d1935e01a1548be22949ffaa.

jnraine:
Awesome @segiddins.

PR #3363, opened 2015-01-26
Changelog entry for #3291

kirs:
@indirect I forgot to add the changelog entry for my feature.


indirect:
oh, I always do the changelog when we're ready to release. :) but thanks!

PR #3365, opened 2015-01-27
Api source index fix

keiths-osc:
Consider the following scenario:

* You have two API gem sources named A and B.  (It is important that both sources are API sources).
* Source A and only source A contains version 1.0.0 of a gem named foo.
* The foo gem explicitly depends on version 1.2.0 of a gem named bar.
* Version 1.2.0 of bar exists only in source B.  However, another version of bar exists in source A.  (The version in source A does not satisfy the dependency of the foo gem).

In this case, a bundle install of the following Gemfile will fail with an error message saying it can't find version 1.2.0 of bar in any sources:

    source "http://sourcea.example.com"
    source "http://sourceb.example.com"
    
    gem 'foo', '1.0.0'
 
Removing the bar gem from source A causes the error to stop happening.
 
We ran into this situation in real life with an in-house geminabox gem server as the first source and rubygems.org as the second source.  We have an in-house gem (that I'll call foo) located in our in-house gem server.  The foo gem depends on thor ~>19.0.  Our in-house gem server happens to have a forked version of thor with a version number less than 19 that doesn't exist on rubygems.org.  When we do a bundle install, we get an error message stating that thor ~> 19.0 cannot be found in any of the sources.

This pull request adds a test that exposes the problem, and then in a subsequent commit adds code that makes the test start working.

### DIFF
--- a/lib/bundler/index.rb
+++ b/lib/bundler/index.rb
@@ -101,11 +101,15 @@ module Bundler
 
     # returns a list of the dependencies
     def unmet_dependency_names
-      names = []
-      each{|s| names.push(*s.dependencies.map{|d| d.name }) }
-      names.uniq!
+      names = dependency_names
       names.delete_if{|n| n == "bundler" }
       names.select{|n| search(n).empty? }
+    end
+
+    def dependency_names
+      names = []
+      each{|s| names.push(*s.dependencies.map{|d| d.name }) }
+      names.uniq
     end
 
     def use(other, override_dupes = false)
--- a/lib/bundler/source/rubygems.rb
+++ b/lib/bundler/source/rubygems.rb
@@ -309,6 +309,18 @@ module Bundler
               Bundler.ui.info "" if !Bundler.ui.debug? # new line now that the dots are over
             end
 
+            # Suppose the gem Foo depends on the gem Bar.  Foo exists in Source A.  Bar has some versions that exist in both
+            # sources A and B.  At this point, the API request will have found all the versions of Bar in source A,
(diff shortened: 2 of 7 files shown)


keiths-osc:
At the moment I don't believe the test failures were related to my changes.  Of all the various ruby/rubygems combinations that it tests against, the ones that failed were ruby 2.1 with rubygems 2.2.2, and ruby 2.1 with rubygems master.  I ran both of these locally and the tests passed.  The 2.1/master run appears to have failed because of a timeout when trying to talk to rubygems.org.  The 2.1/2.2.2 run complained about not being able to require sinatra over and over again and I'm wondering if that also could have been related to a hiccup talking to rubygems.org. Is there a way to just try the test run in Travis again?

indirect:
Thanks for this patch! We knew this was a theoretical problem, but weren't sure how to reproduce it in a test.

I've restarted both travis failures, which look to be caused by network issues, and I'll merge as soon as they pass.

keiths-osc:
Thanks for pulling this in!  Do I need to also create pull requests for the 2-0-dev and master branches, or will this be merged in at some point?

indirect:
Nope, it will  be merged into master on release (which should be in the next day or two).

TimMoore:
Thanks, @keiths-osc!

@indirect should we backport this to `1-7-stable`?

indirect:
@tmoore if you could that would be awesome, we can release it in 1.7.13.

TimMoore:
Done 0514a3e9ce60fc560220ff11fc972943cdc1d642

PR #3383, opened 2015-02-09
Remove ref from git cache path

bkeepers:
When using `bundle cache --all` with git dependencies, bundler includes the ref as part of the path in `vendor/cache`. When `bundle update` is run, the old cache is removed and the new cache is added, making diffs very noisy. This is a proposal to remove the ref from the path and store it in `.bundlercache`.

This is potentially a breaking change, so it might be best to put it behind a config option.

### DIFF
--- a/lib/bundler/source/git.rb
+++ b/lib/bundler/source/git.rb
@@ -183,6 +183,7 @@ module Bundler
         git_proxy.checkout if requires_checkout?
         git_proxy.copy_to(app_cache_path, @submodules)
         serialize_gemspecs_in(app_cache_path)
+        File.write(app_cache_path.join(".bundlecache"), cached_revision || revision)
       end
 
       def load_spec_files
@@ -209,7 +210,7 @@ module Bundler
       end
 
       def app_cache_dirname
-        "#{base_name}-#{shortref_for_path(cached_revision || revision)}"
+        base_name
       end
 
       def revision
--- a/spec/cache/git_spec.rb
+++ b/spec/cache/git_spec.rb
@@ -16,24 +16,23 @@ end
   describe "bundle #{cmd} with git" do
     it "copies repository to vendor cache and uses it" do
       git = build_git "foo"
-      ref = git.ref_for("master", 11)
+      ref = git.ref_for("master")
 
       install_gemfile <<-G
         gem "foo", :git => '#{lib_path("foo-1.0")}'
       G
 
       bundle "#{cmd} --all"
-      expect(bundled_app("vendor/cache/foo-1.0-#{ref}")).to exist
-      expect(bundled_app("vendor/cache/foo-1.0-#{ref}/.git")).not_to exist
-      expect(bundled_app("vendor/cache/foo-1.0-#{ref}/.bundlecache")).to be_file
+      expect(bundled_app("vendor/cache/foo-1.0")).to exist
+      expect(bundled_app("vendor/cache/foo-1.0/.git")).not_to exist
+      expect(bundled_app("vendor/cache/foo-1.0/.bundlecache").read).to eql(ref)
 
       FileUtils.rm_rf lib_path("foo-1.0")
       should_be_installed "foo 1.0"
     end
 
     it "copies repository to vendor cache and uses it even when installed with bundle --path" do
-      git = build_git "foo"
-      ref = git.ref_for("master", 11)
+      build_git "foo"
 
       install_gemfile <<-G
         gem "foo", :git => '#{lib_path("foo-1.0")}'
@@ -42,8 +41,8 @@ end
       bundle "install --path vendor/bundle"
       bundle "#{cmd} --all"
 
-      expect(bundled_app("vendor/cache/foo-1.0-#{ref}")).to exist
-      expect(bundled_app("vendor/cache/foo-1.0-#{ref}/.git")).not_to exist
+      expect(bundled_app("vendor/cache/foo-1.0")).to exist
+      expect(bundled_app("vendor/cache/foo-1.0/.git")).not_to exist
 
       FileUtils.rm_rf lib_path("foo-1.0")
       should_be_installed "foo 1.0"
@@ -66,7 +65,7 @@ end
 
     it "tracks updates" do
       git = build_git "foo"
-      old_ref = git.ref_for("master", 11)
+      old_ref = git.ref_for("master")
 
       install_gemfile <<-G
         gem "foo", :git => '#{lib_path("foo-1.0")}'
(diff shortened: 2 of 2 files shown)


indirect:
This seems like a lot of overhead, including a magic file that could easily be deleted without deleting the files themselves, for a very edge case. Can you tell me why you feel like this is a good idea?

bkeepers:
> Can you tell me why you feel like this is a good idea?

We quit using `cache --all` at GitHub because every time we updated a dependency the diff would be unreadable. With this change, the diff clearly shows what changed in the dependency, which makes it extremely convenient to do reviews.

indirect:
So, two things: first, why are you caching git repos? Do your dependencies have really high churn that means you're not able to release them as gems with versions? Second, if you're going to cache your dependencies in your repos, we _strongly_ recommend using a submodule, so that the primary repo isn't bloated by the .gem files or the dependency files themselves. As an example, you can check out [how RubyGems.org uses a submodule for that](http://blog.rubygems.org/2015/02/01/rewriting-history.html).

bkeepers:
> first, why are you caching git repos?

Because we need to be able to deploy GitHub if GitHub is down, so we have all our external dependencies vendored. Other people should really be doing this as well. Having deploy dependencies on external services is not a good idea for any service.

> Do your dependencies have really high churn that means you're not able to release them as gems with versions?

Yes, two frequent use cases:

1. a new feature that is in heavy development
2. using a fork of a repository that we don't manage releases for, which we [do often](https://github.com/github?utf8=%E2%9C%93&query=+only%3Aforks).

> Second, if you're going to cache your dependencies in your repos, we strongly recommend using a submodule…

Same reason as the first question. We need to be able to deploy without any external dependencies.

> …so that the primary repo isn't bloated by the .gem files or the dependency files themselves.

`cache --all` with git repos doesn't include the `.gem` files, which is one of the reasons I think it would be really nice if we could just use `:git` dependencies in bundler.

Maybe it would make sense to deprecate support for caching Git repos with `cache --all` if use of this feature is discouraged?

indirect:
> we have all our external dependencies vendored. Other people should really be doing this as well

Your assertion that GitHub's decisions about how to handle tradeoffs are the Only Possible One True Way™ makes me really unhappy. What makes sense for GitHub may or may not make sense for other people who have different needs, and you don't know what their needs are. Please don't do that, it's patronizing and can be very misleading.

> Maybe it would make sense to deprecate support for caching Git repos with cache --all if use of this feature is discouraged?

Since I haven't said anything to discourage use of this feature, in this ticket or anywhere else, this comes across as both passive-aggressive and smarmy. No, we're not going to deprecate a perfectly good feature that people already use and are happy with because it doesn't meet your precise needs.

To the actual point:

I don't think it makes sense to sit here and argue back and forth about the relative merits of all the possible options, so I'm just going to summarize the current results of the core team's discussions about this.

As for this particular pull request, using a dotfile to effectively re-implement git submodules is not a sufficiently robust solution for the core team to be comfortable agreeing to merge it and support it for the next five years or more.

In general, we'd like to come up with a good solution for deploying without external service dependencies that is both robust and flexible enough to satisfy the needs of people using both regular gems and git gems. The best results that I've heard so far about have used regular gems over git gems whenever possible, and involved either a repo or a build artifact that corresponded to each commit in the main repo rather than full-on vendoring everything directly.

bkeepers:
I have a lot of respect for bundler and specifically your work on it, so sorry that I have offended you.

> Your assertion that GitHub's decisions about how to handle tradeoffs are the Only Possible One True Way™ makes me really unhappy.

Sorry, you are right. I re-read what I wrote and it did come off as rather prescriptive. In my experience, it's really painful when you need to deploy and an external service is unavailable.

> As for this particular pull request, using a dotfile to effectively re-implement git submodules…

I'm curious how this is different than the current approach of using the directory name. This proposal just moves it from the directory name to the existing `.bundlecache `. It doesn't change the functionality of the feature.

> No, we're not going to deprecate a perfectly good feature that people already use and are happy with because it doesn't meet your precise needs.

There are [1.7 million Gemfiles](https://github.com/search?utf8=%E2%9C%93&q=filename%3AGemfile&type=Code&ref=searchresults) in public repositories on GitHub, and almost [1/3 of them have a git or github dependency](https://github.com/search?utf8=%E2%9C%93&q=filename%3AGemfile+git+OR+github&type=Code&ref=searchresults). There are [191,000 repos that have `vendor/cache/rack-1.6.0.gem`](https://github.com/search?utf8=%E2%9C%93&q=filename%3Arack-1.6.0.gem+path%3Avendor%2Fcache&type=Code&ref=searchresults) (the latest release of the most popular gem), but only [183 `.bundlecache` files in public repositories on GitHub](https://github.com/search?o=desc&q=filename%3A.bundlecache+is%3Apublic&ref=searchresults&s=indexed&type=Code&utf8=%E2%9C%93), which is present when a git dependency is cached.  

This suggests to me that people care about vendoring their gem dependencies, but don't vendor git dependencies for one reason or another. This could just be because they don't know about the `--all` flag. In my experience, it's because updating the git dependency makes the diff on updates unnecessarily noisy.

I think bundler's caching is a great feature and want to make it so GitHub is not a single point of failure for everyone deploying ruby applications.

indirect:
> This proposal just moves it from the directory name to the existing .bundlecache. It doesn't change the functionality of the feature.

This is a really good point. I went back and looked at your pull request again, and it's much simpler than I thought it was while writing my last response—I think I was confusing this pull request with another PR to do the same thing with a much more complicated implementation. Regardless of how it happened, I was judging your PR much more harshly than it actually deserves, and I'm sorry about that.

> This suggests to me that people care about vendoring their gem dependencies, but don't vendor git dependencies for one reason or another.

Yeah, that seems true. The reason that `--all` isn't the default is because we were trying to maintain backwards compatibility with 1.x releases. Bundler 2.0 will remove the `--all` option and simply do that all the time.

I think I'm now convinced that this PR is an overall improvement. Assuming that I go back to the core team and they agree with me, there are some new concerns about merging this. :) The first issue is that this change completely breaks compatibility—previously cached git gems will no longer work, and git gems cached in the future will not be readable by older versions of Bundler. That makes this change a candidate for Bundler 2.0, but not possible to merge immediately.

Are you willing to implement code that reads existing caches and migrates old git gems to the new cache format? If so, we can possibly merge and release this in Bundler 1.10.

bkeepers:
> Regardless of how it happened, I was judging your PR much more harshly than it actually deserves, and I'm sorry about that.

:beers: No problem, I wasn't communicating very effectively either.

> Are you willing to implement code that reads existing caches and migrates old git gems to the new cache format? 

:+1:, but this could be tricky. If I'm running `1.10.x` and migrate to the new cache format, but my co-workers are running `1.9.2`, then they won't get the cache.

How do you typically handle changes like this? Is it ok to assume that everyone on the project is using a compatible bundler version, or would it make more sense to put this behind a config that is disabled by default for `1.x`?

indirect:
We've just merged a PR that tracks the current Bundler version in the lockfile, so this will get easier in the future (since it will print a warning that you're out of date when you bundle).

With cache migration this change will only break forwards compatibility, so maybe we can just add a warning to Bundler 1.9.3 that explains that the cache format requires 1.10 or newer?

arthurnn:
:heart: just dropping this here, because I love open source and bundler :smile_cat: (good job communicating back-and-forth both of you)

PR #3389, opened 2015-02-11
Remove white space at the end of the line

claudiob:
If you create a gem without `config[:bin]`, then your README file will have an extra white character at the end of the line.

This commit removes that extra character.


TimMoore:
The test failures are unrelated.

indirect:
Cherry-picked as 3adb756a8dee0d9757146048112467ee98f11d5c.

PR #3394, opened 2015-02-12
No longer use gemcutter. Update to rubygems.org

dsisnero:


segiddins:
You also need to update the tests: https://travis-ci.org/bundler/bundler/jobs/50557995#L2108

indirect:
also, the `:gemcutter` symbol maps to `rubygems.org` already.

TimMoore:
Unfortunately, `:gemcutter` maps to `http://rubygems.org` and not `https` (and gives a deprecation warning).

I think we should make this change and fix the tests.

indirect:
:+1:

TimMoore:
er... it isn't obvious from how GitHub formats this, but I fixed the test in the merge commit

PR #3400, opened 2015-02-14
Skip second build of definition

k0kubun:
This is just a trivial optimization to save some milliseconds.

If you run `bundle install`, Gemfile is evaluated [here](https://github.com/bundler/bundler/blob/13d5e08de1e1b957bbf4c5fecd8d3e532a09c121/lib/bundler/cli/install.rb#L78) for the first time.
In [`Bundler.definition`](https://github.com/bundler/bundler/blob/13d5e08de1e1b957bbf4c5fecd8d3e532a09c121/lib/bundler.rb#L155), it creates `Definition.build(default_gemfile, default_lockfile, nil)`.
Even though `Bundler::Installer` has it in `@definition`, it evaluates the same thing again [here](https://github.com/bundler/bundler/blob/13d5e08de1e1b957bbf4c5fecd8d3e532a09c121/lib/bundler/installer.rb#L69). I think this can be skipped.

### DIFF
--- a/lib/bundler/installer.rb
+++ b/lib/bundler/installer.rb
@@ -66,8 +66,7 @@ module Bundler
       if Bundler.default_lockfile.exist? && !options["update"]
         local = Bundler.ui.silence do
           begin
-            tmpdef = Definition.build(Bundler.default_gemfile, Bundler.default_lockfile, nil)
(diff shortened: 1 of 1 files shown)


k0kubun:
Tests are broken, maybe there are some cases I didn't care about. Closing for now. Sorry.

PR #3403, opened 2015-02-16
Check for bool value as well

EduardoBautista:
When being asked for the first time whether you want to have a COC or license file added to your gems or not, the value is a boolean and will always return true even if you chose not to generate the file. Because the config file is read on future gem creations, the value is a string and will work as intended, so it the issue only occurs the first time you create a gem.

### DIFF
--- a/lib/bundler/settings.rb
+++ b/lib/bundler/settings.rb
@@ -136,7 +136,7 @@ module Bundler
     end
 
     def to_bool(value)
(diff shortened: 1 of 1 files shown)


indirect:
good catch, thanks!

indirect:
Cherry-picked as 5b30fed7673ea5a7709e418484e78aafa4f5a9b6.

Issue #3411, opened 2015-02-17
Does not respect the versioning ~> directive


joneslee85:
Since last week, I've had few errors related to versioning conflicts while working on PR for rollbar-gem and newrelic_rpm. What I found that the `~>` does not work, for my case, ActiveSupport 3.2 depends on ~> 0.6 yet bundler (1.8.2) install version 0.7 instead of 0.6.9.

Here is the Travis CI failed build for your investigation: https://travis-ci.org/newrelic/rpm/builds/51044928

TimMoore:
I think you've misunderstood how the `~>` specifier works.

`~> 0.6` expands to `>= 0.6, < 1.0`. In your case, you want to use `~> 0.6.0` (note the extra digit), which expands to `>= 0.6.0, < 0.7`.

Full documentation is at http://guides.rubygems.org/patterns/#pessimistic-version-constraint

PR #3412, opened 2015-02-17
Add a failed test case for ~> directive

joneslee85:
Related to https://github.com/bundler/bundler/issues/3411


indirect:
Thanks for the test case!

TimMoore:
Thanks, but this is working as intended. See my comment on #3411.

Issue #2849, opened 2014-01-30
Can't find compatible versions for gem "bundler" for gemspec generated with pre bundler.


simi:
Hello. I'm just testing `1.6.0.pre.1` (awesome work). I found one issue, but I'm not sure if that's typical usecase.

I created gem with `bundle gem testgem` and in generated gemspec was this line:

``` ruby
spec.add_development_dependency "bundler", "~> 1.6"
```

Since bundler 1.6 wasn't released yet, bundle can't find this version and it ends with this message (as expected):

```
Bundler could not find compatible versions for gem "bundler":
  In Gemfile:
    bundler (~> 1.6) ruby

  Current Bundler version:
    bundler (1.6.0.pre.1)

This Gemfile requires a different version of Bundler.
Perhaps you need to update Bundler by running `gem install bundler`?
```

I think that's a acceptable tax for using pre bundler, but maybe we can use some kind of more descriptive message for this usecase. Any ideas?

PR #3415, opened 2015-02-17
Allow dev versions when creating a new gem

TimMoore:
The changed behaviour is to see if the last segment in the Bundler version contains any letters, and if so, tack it on.

So with Bundler 1.9.0.dev, you get:

```ruby
  spec.add_development_dependency "bundler", "~> 1.9.dev"
```

and with Bundler 1.8.2, you get the same as before:

```ruby
  spec.add_development_dependency "bundler", "~> 1.8"
```

### DIFF
--- a/lib/bundler/cli/gem.rb
+++ b/lib/bundler/cli/gem.rb
@@ -164,10 +164,10 @@ module Bundler
     end
 
     def bundler_dependency_version
-      major, minor, *etc = Bundler::VERSION.split(".")
-      version = [major, minor]
-      version << etc.last if etc.last =~ /[a-z]/i
-      version.join(".")
+      v = Gem::Version.new(Bundler::VERSION)
+      req = v.segments[0..1]
+      req << v.segments.last if v.prerelease?
+      req.join(".")
     end
 
   end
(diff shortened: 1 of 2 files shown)


TimMoore:
This should fix the build on master.

lib/bundler/cli/gem.rb
```diff
@@ -162,5 +163,12 @@ def ask_and_set_test_framework
       test_framework
     end
 
+    def bundler_dependency_version
+      major, minor, *etc = Bundler::VERSION.split(".")
+      version = [major, minor]
+      version << etc.last if etc.last =~ /[a-z]/i
+      version.join(".")
```
indirect: I think we can just use `Gem::Version` to do the work for us:

```ruby
v = Gem::Version.new(Bundler::VERSION)
req = v.approximate_recommendation
req << v.segments.last if v.prerelease?
req
```

TimMoore: Even better. I'll update.

TimMoore: Hmm... `approximate_recommendation` is only available with RG 2.0+. Previously, it had a different name: https://github.com/rubygems/rubygems/commit/f5035c61750b8a5af376dc30a1ed74e87b0152f3

I could monkey patch in a method alias if it's not already defined, but it probably isn't worth it just for this. I'll use a hybrid of my original code and your change.

indirect: Heh, probably not. Sounds good!

simi:
probably related to https://github.com/bundler/bundler/issues/2849

TimMoore:
Yep, thanks for the pointer @simi. Not sure why that one was closed.

Issue #3417, opened 2015-02-18
“Could not find gem …” error, apparent conflict between `gem … :path` and `gem … :bundle`


kch:
Minimal repro app here: https://github.com/kch/bundler-source-bug

Original SO question with some details: http://stackoverflow.com/questions/28557998/could-not-find-gem-error-with-bundler-using-custom-gem-source

The SO question doesn't get into what seems to be the issue, which is:

If you require a gem with `gem foo, bundle:…` things work fine. If you require a gem with `gem bar, path:…`, but as soon as you require both at in the same Gemfile, things get borked.

More specifically, bundler is then unable to find any gems with a custom source.

This occurs when the source is passed as a parameter to `gem` or when `source` is called with a block, but not with the traditional `source url` at the top.

/cc @hone @schneems yo.

indirect:
Have you tried this with 1.8.2, which fixed at least one similar issue?

kch:
Yeah versions are listed in the SO question, all latest.

```
$ ruby -v    # => ruby 2.2.0p0 (2014-12-25 revision 49005) [x86_64-darwin14]
$ gem -v     # => 2.4.5
$ bundle -v  # => Bundler version 1.8.2
```

kch:
Try the repro app. It has zero actual code in it. Unless you count the Gemfile. Just follow the Readme.

eagletmt:
It seems same with #3414 .

kch:
Yup, seems to be the same.

TimMoore:
Closing this in favor of #3414.

PR #3418, opened 2015-02-18
LocalRubygems is required for source options

### LINKED ISSUE #3414: `path` conflicts with secondary `source` in Gemfile
See https://gist.github.com/harto/0dfcee489e248900feb8 for reproduction.

Suppose a Gemfile references one gem from a local `path` and another from a secondary `source`, e.g.:
```
source "https://rubygems.org"
gem "local_gem", path: "local_gem"
gem "other_gem", source: "https://gems.example.com"
```

If I `bundle install`, the dependencies are apparently resolved and installed correctly. However, the bundle fails at runtime:
```
$ bundle exec ruby -e 'puts "hello"'
Could not find gem 'other_gem (>= 0) ruby' in rubygems repository https://gems.example.com/.
Source does not contain any versions of 'other_gem (>= 0) ruby'
```

Removing `local_gem` and running `bundle install` again appears to fix the problem:
```
$ bundle exe ruby -e 'puts "hello"'
hello
```

The `local_gem` doesn't have any dependencies. Again, see https://gist.github.com/harto/0dfcee489e248900feb8 for a minimal complete example.

eagletmt:
Since the `:path` option forces to re-resolve dependencies each time,
the index for source options has to see local gems index even in
`bundle check` or `bundle exec` mode.

It should fix #3414 and #3417 .

### DIFF
--- a/lib/bundler/source_list.rb
+++ b/lib/bundler/source_list.rb
@@ -19,7 +19,7 @@ module Bundler
     end
 
     def add_rubygems_source(options = {})
-      add_source_to_list Source::Rubygems.new(options), @rubygems_sources
+      add_source_to_list Source::LocalRubygems.new(options), @rubygems_sources
     end
 
     def add_rubygems_remote(uri)
--- a/spec/bundler/source_list_spec.rb
+++ b/spec/bundler/source_list_spec.rb
@@ -69,7 +69,7 @@ describe Bundler::SourceList do
       end
 
       it "returns the new rubygems source" do
-        expect(@new_source).to be_instance_of(Bundler::Source::Rubygems)
+        expect(@new_source).to be_instance_of(Bundler::Source::LocalRubygems)
       end
 
       it "passes the provided options to the new source" do


indirect:
Thanks for the patch! My intention with LocalRubygems was to avoid reading the installed specs once per source, and instead only read them once.

@tmoore Can we solve this problem by providing @rubygems_aggregate as a source during `check` and `exec` as well? It seems like right now they aren't able to find gems from that source.

indirect:
(Also: we should definitely add a regression test for this particular bug!)

TimMoore:
> @tmoore Can we solve this problem by providing @rubygems_aggregate as a source during check and exec as well?

I wasn't aware that it isn't. After quick look back at the code, looks like it's always created. There could be an order-of-operations dependency in there somewhere. I'll have to debug.

TimMoore:
Problem could be that `SourceList#combine_rubygems_sources` should be returning a `LocalRubygems` instance. I can give it a closer look tonight after work, but if anyone wants to run with it, go right ahead.

TimMoore:
Closing in favor of #3427

PR #3421, opened 2015-02-20
allow a gem with a source to override a gem without a source

ccutrer:
In a complex dependency structure, a project could have plugins (that are gems), but also allow inclusion of Gemfile fragments from that plugin, so that it can specify the source of its (unreleased) further dependencies. However, if the plugin depends on a development version of the same gem that the main project depends on, bundler won't allow the same gem to be listed with multiple sources, even though the project itself doesn't care where the gem comes from. This PR allows that.

Note that this also fixes the Gemfile 
```ruby
gem 'bob', github: 'bob'
gem 'bob'
```
producing the incomprehensible error
```
You specified that bob (>= 0) should come from https://github.com/bob/bob.git (at master) and 
```

### DIFF
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -96,10 +96,14 @@ module Bundler
             @dependencies.delete current
           elsif dep.type == :development
             return
+          elsif !dep.source
+            return
+          elsif !current.source
+            @dependencies.delete current
           else
             raise GemfileError, "You cannot specify the same gem twice coming from different sources.\n" \
                             "You specified that #{dep.name} (#{dep.requirement}) should come from " \
-                            "#{current.source || 'an unspecified source'} and #{dep.source}\n"
+                            "#{current.source} and #{dep.source}\n"
           end
         end
       end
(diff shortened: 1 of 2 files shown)


indirect:
Inclusion of Gemfile fragments is not a supported feature. If you'd like, you can mention this request on the feature request for including Gemfiles in other Gemfiles. The reason that we have avoided implementing Gemfile inclusion so far is that it adds complexity and confusing edge cases, like this one.

ccutrer:
And here I was thinking it was supported cause I just found the eval_gemfile method the other day, which simplified our implementation of it. Thanks for the pointer, I'll go check it out.

indirect:
The supported Bundler Ruby API includes _only_ the Gemfile DSL, `Bundler.setup`, and `Bundler.require`. All other methods are for use at your own risk. :)

ccutrer:
`eval_gemfile` _is_ a public method in `Bundler::Dsl`, and no comments saying it's for internal use. Then again, so is `to_definition`.

indirect:
The public Gemfile DSL is documented in [the Gemfile help](http://bundler.io/man/gemfile.5.html) (which can also be read by running `bundle help gemfile`, and not than by the public Ruby methods. I'd like to fix that at some point, but other things have been more important so far.

PR #3422, opened 2015-02-20
Add Rubinius to Build Matrix

bjfish:
This currently has an allowed failure listing. Trying to add rbx to builds of popular Gems to test for compatibility.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,7 @@ rvm:
   - 2.0.0
   - 1.9.3
   - 1.8.7
+  - rbx-2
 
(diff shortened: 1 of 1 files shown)


indirect:
Does it pass yet?!

bjfish:
@indirect The build passes. Rubinius is marked as an allowed failure.

I think it is okay to for Rubinius to fail here for now. We'll have visibility into this failure on this dashboard here http://bjfish.github.io/rubinius-gem-build-status/ and can investigate why this fails.

indirect:
Great! And sorry for the long delay merging this.

PR #3427, opened 2015-02-22
Fix multi source unlocking

TimMoore:
Sometimes, it's disappointing to compare the amount of time it took to track down a problem with the amount of code it took to fix it :smile:

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -406,11 +406,18 @@ module Bundler
 
       if locked
         unlocking = @locked_specs.any? do |locked_spec|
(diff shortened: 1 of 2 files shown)


indirect:
The build passes and this appears to resolve the issue, so I'm going to merge it.

tamird:
@indirect could you cut a release containing this fix?

indirect:
@tamird does this fix your issue? You can check by cloning this repo, running `git checkout 1-8-stable`, and then `rake install`, which will build and install Bundler from source.

indirect:
Released 1.8.3 with this fix.

Issue #3286, opened 2014-12-03
Gem path points to "gemfiles" directory


avit:
I'm trying to use my own fork of this gem: https://github.com/premailer/premailer/

Note, the original gem source has a "gemfiles" directory at the root.

The original gem works if loaded from rubygems. When I point to the git repo from my Gemfile, it seems to recognize the "gemfiles" directory instead of the correct gem root:

```ruby
# Gemfile
gem "premailer", github: "premailer/premailer", branch: "master"
```

```
$ bundle update premailer
...
$ bundle show premailer
/opt/rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/bundler/gems/premailer-28e3cb96befb/gemfiles
```

The result is that none of the gem files can be found; `require "premailer"` fails. Manually deleting the "gemfiles" seems to resolve to the correct directory:

```
$ rm -r /opt/rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/bundler/gems/premailer-28e3cb96befb/gemfiles

$ bundle show premailer
/opt/rbenv/versions/2.1.3/lib/ruby/gems/2.1.0/bundler/gems/premailer-28e3cb96befb
```

indirect:
Ticketed in [this Trello ticket](https://trello.com/c/DQx0OKJl).

PR #3428, opened 2015-02-24
[Source::Path] Sort gemspecs by relative depth

segiddins:
Attempts to close https://github.com/bundler/bundler/issues/3286, https://trello.com/c/DQx0OKJl/82-search-for-gemspecs-near-the-root-of-a-git-repo-first

\c @TimMoore

### DIFF
--- a/lib/bundler/source/path.rb
+++ b/lib/bundler/source/path.rb
@@ -130,7 +130,8 @@ module Bundler
         index = Index.new
 
         if File.directory?(expanded_path)
-          Dir["#{expanded_path}/#{@glob}"].sort_by { |p| p.split(File::SEPARATOR).size }.each do |file|
+          # We sort depth-first since `<<` will override the earlier-found specs
+          Dir["#{expanded_path}/#{@glob}"].sort_by { |p| -p.split(File::SEPARATOR).size }.each do |file|
             spec = Bundler.load_gemspec(file)
             if spec
               spec.loaded_from = file.to_s
--- a/lib/bundler/source/path.rb
+++ b/lib/bundler/source/path.rb
@@ -130,7 +130,7 @@ module Bundler
         index = Index.new
 
         if File.directory?(expanded_path)
-          Dir["#{expanded_path}/#{@glob}"].each do |file|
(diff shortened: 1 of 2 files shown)


indirect:
hah, clever. I was going to have two globs, the first only matching the top level. I think this is a better general solution

segiddins:
Let me know if you want specs for this change

-Samuel E. Giddins

On Feb 23, 2015, at 6:25 PM, André Arko <notifications@github.com> wrote:

hah, clever. I was going to have two globs, the first only matching the top level. I think this is a better general solution

indirect:
I think the existing specs probably cover this sufficiently? Of course, I may regret saying that in a few weeks or months, so if you could add a spec that would be great. :+1:

segiddins:
Hm, I can't use the premailer example to make a realworld spec, since only their development dependencies differ.

TimMoore:
:+1: thanks @segiddins!

A spec would be good, if only to make sure nobody breaks it again in the future. It might work to use the `build_lib` helper to build a simple gem with an invalid gemspec (one with just `raise "FAIL"` in it for example) in a subdirectory that sorts before the real one alphabetically. Then the spec can verify that the path gem still works in a `Gemfile`.

segiddins:
@TimMoore would be happy to -- any pointers on how to do that with `build_lib`? It always just seems to be used like `build_lib 'foo'`

TimMoore:
@segiddins it can take a block, and yields a builder object that can be used to write files into the path. See https://github.com/bundler/bundler/blob/master/spec/install/path_spec.rb#L56 for example

```
      build_lib "rack", "1.0.0", :to_system => true do |s|
        s.write "lib/rack.rb", "raise 'FAIL'"
      end
```

segiddins:
@TimMoore updated with a spec + comments :)

Turned out the issue was `Index#<<` overriding the earlier specs -- in other words, the glob pattern was not the actual problem.

indirect:
Yay specs! :)

TimMoore:
LGTM :+1:

PR #3429, opened 2015-02-24
Source blocks don't work in the presence of `path`

tamird:
Failing test case.

@indirect cc @tarcieri


indirect:
Thanks!

tamird:
Looks like this might be a dupe of https://github.com/bundler/bundler/issues/3414

TimMoore:
This should be fixed in Bundler 1.8.3 (released today).

Issue #3433, opened 2015-02-26
bundle install - include commit hash for git source


revett:
When using a `:git`source for a gem, the output when running `bundle install` does not include the commit hash, simply the branch:

> **Output**

```
Fetching git@github.com:revett/composition.git
Using composition 0.0.1 from git@github.com:revett/composition.git (at master)
```

This makes it look as though it's using the latest version of the gem from **master**. However the `Gemfile.lock` locks the gem to **any** previous commit hash (whenever the gem was last updated). 

Would be great to have this information in the output - very helpful in the console output of Jenkins CI jobs for example.

> **Gemfile**

```
source "https://rubygems.org/"

group :production do
  gem "composition", :git => "git@github.com:revett/composition.git"
  gem "faraday"
  gem "mustache", "~> 0.99.5"
end
```

revett:
Similiar to https://github.com/bundler/bundler/issues/178.

PR #3434, opened 2015-02-26
Show revision on bundle install from git

darnould:
A possible solution for issue #3433 - which requests that the git revision (provided that git is the source) is displayed when `bundle install`ing in addition to the branch name.  I agree that such information might be of assistance.

### DIFF
--- a/lib/bundler/source/git.rb
+++ b/lib/bundler/source/git.rb
@@ -70,7 +70,7 @@ module Bundler
         else
           ref
         end
-        "#{uri} (at #{at} - #{revision})"
+        "#{uri} (at #{at})"
       end
 
       def name
@@ -171,7 +171,7 @@ module Bundler
         if requires_checkout? && spec.post_install_message
           Installer.post_install_messages[spec.name] = spec.post_install_message
         end
-        ["Using #{version_message(spec)} from #{to_s}", nil, debug]
+        ["Using #{version_message(spec)} from #{to_s} - #{revision}", nil, debug]
       end
 
       def cache(spec, custom_path = nil)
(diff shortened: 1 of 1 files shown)


darnould:
Note: the build was failing before I went near it ;-)

segiddins:
@darnould I believe some of the failures are due to changed CLI output, so you may need to update a few of the specs :)

darnould:
@segiddins Thanks for catching that, looks like I had some env issues at the time - whilst simultaneously misreading Travis! :)

https://github.com/darnould/bundler/commit/afc00709f14a68fe3388ad72c7bafbf6dbd60015 has the specs passing, although display of the SHA is outside of the parentheses (possibly undesirable).

https://github.com/darnould/bundler/commit/6cb300ef89476f816b79c4bc68c141a856508418 works with my manual use-cases, although failing on others I'm less familiar with in the specs.  Modifying `to_s` to include the SHA might be a fundamentally problematic approach given the range of states it can be in.  Have to dash, but may come back to this later: not sure if the overall feature is desirable as a whole yet!

TimMoore:
@darnould this feature sounds pretty useful to me. It might be better to show an abbreviated SHA1 instead of the whole thing (maybe the first 7 characters, like github and `git log --oneline` use). Any thoughts, @indirect?

After a quick glance at the Travis output, I think getting the tests to pass would just be a matter of changing the expectations to include the hash.

indirect:
I think the first 7 characters would be best, yeah.

indirect:
@darnould this seems good to me. If you update it to just the first 7 characters of the git sha, rebase, and squash, I'll merge. :ship:

indirect:
@darnould I'm going to close this due to inactivity, but if you get the chance to update it I would be happy to merge!

Issue #2663, opened 2013-10-05
The `g+s` is not preserved when installing gems from git.


cyphactor:
It seems that when bundle installing. The gems that are installed via Git don't preserve the `g+s` on the gem directories.

The Git installed gem directories should have the setgid bit preserved on their directories as the normal gem installs, seen below.

```
appuser@precise64:/rl/product/capture/apps/capture_api/current$ ls -ld /usr/local/rvm/gems/ruby-1.9.3-p327\@capture_api/gems/* | head
drwxrwsr-x  3 vagrant rvm 4096 Oct  4 17:08 /usr/local/rvm/gems/ruby-1.9.3-p327@capture_api/gems/actionmailer-3.2.13
```

However, the Git installed gem directories are NOT preserving the setgid bit as they should. This can be seen in the following:

```
appuser@precise64:/rl/product/capture/apps/capture_api/current$ ls -ld /usr/local/rvm/gems/ruby-1.9.3-p327\@capture_api/bundler/gems/*
drwxrwxr-x 5 vagrant rvm 4096 Oct  4 17:15 /usr/local/rvm/gems/ruby-1.9.3-p327@capture_api/bundler/gems/capture_rliapi_json-9aba32c88f89
drwxrwxr-x 5 vagrant rvm 4096 Oct  4 17:18 /usr/local/rvm/gems/ruby-1.9.3-p327@capture_api/bundler/gems/cequel-aa25f8bb57a8
drwxrwxr-x 5 vagrant rvm 4096 Oct  4 17:18 /usr/local/rvm/gems/ruby-1.9.3-p327@capture_api/bundler/gems/poltergeist-95ee94ef6da4
drwxrwxr-x 8 vagrant rvm 4096 Oct  4 17:18 /usr/local/rvm/gems/ruby-1.9.3-p327@capture_api/bundler/gems/replywise_reachlocal_event_consumer-a7edd41b0641
drwxrwxr-x 5 vagrant rvm 4096 Oct  4 17:15 /usr/local/rvm/gems/ruby-1.9.3-p327@capture_api/bundler/gems/rl_event_rest_client-f481cddf3e5b
drwxrwxr-x 5 vagrant rvm 4096 Oct  4 17:15 /usr/local/rvm/gems/ruby-1.9.3-p327@capture_api/bundler/gems/rliapi-7ecd4d45658c
```

[...]

PR #3438, opened 2015-02-27
[GitProxy] Try and preserve the umask on git gems

segiddins:
Attempts to close https://github.com/bundler/bundler/issues/2663.

I have no idea if this is right, I just implemented what @cyphactor suggested.

### DIFF
--- a/lib/bundler/source/git/git_proxy.rb
+++ b/lib/bundler/source/git/git_proxy.rb
@@ -83,7 +83,7 @@ module Bundler
             FileUtils.mkdir_p(destination.dirname)
             FileUtils.rm_rf(destination)
             git_retry %|clone --no-checkout --quiet "#{path}" "#{destination}"|
-            File.chmod((0777 & ~File.umask), destination)
+            File.chmod(((File.stat(destination).mode | 0777) & ~File.umask), destination)
           end
 
           SharedHelpers.chdir(destination) do


indirect:
This seems reasonable to me, and the build passes. So let's try it. :P

segiddins:
Pretty sure this actually won't work, since any file mode or-ed with `0777` will be `0777`

indirect:
Bummer. :( Any ideas on how to preserve the actual umask?

cyphactor:
@segiddins to my understanding your statement of "any file mode or-ed with `0777` will be `0777`" is incorrect.

The reason I say your statement is incorrect is because that digit in the mode represents special permissions including the SetGID bit. See [https://www.novell.com/coolsolutions/trench/16940.html](https://www.novell.com/coolsolutions/trench/16940.html) for details.

## How it is losing SetGID bit

The original code of

```ruby
File.chmod((0777 & ~File.umask), destination)
```

is losing that first digit in terms of file mode because it is grabbing the processes default umask (which won't have the SetGID bit set), getting the complement of it to get the File mode, and then bitwise anding it with `0777`. Lets take a look at a quick example of this.

Lets say that `~File.umask` comes back as `7777` magically. **Note:** This wouldn't actually happen because `File.umask` is the processes default umask. But, still go with it for the sake of the example.

The first digit, `7`, would be ignored when bitwise anded with `0777`, ultimately resulting in `0777`. Shorthand,  `0777 & 7777 = 0777`. Therefore, it would be losing any special permission bits originally existing in the processes default umask.

## Does the new code work?

Looking at the new code I believe it should work.

```ruby
File.chmod(((File.stat(destination).mode | 0777) & ~File.umask), destination)
```

The snippet `(File.stat(destination).mode | 0777)` should properly result in the file mode with special permissions. Lets say the file mode obtained is `7000` the snippet would result in `7777`. Which is what we would want.

When the previous snippet is bitwise anded with the complement of the umask it should include the special permission bits. For example if  `File.umask => 0027`, the complement `~0027 => 7740`, and that bitwise anded with result from the previous snippet `7740 & 1777 => 1740`. In turn maintaining the special permission bit.

How did you test it to identify it wasn't working? What wasn't working about it? Am I missing something?

indirect:
If this code does in fact fix this issue, let's write a test to prove it! I'm not comfortable with code that I don't fully understand that doesn't have a test.

andremedeiros:
I'm +1 for that test. Please.

urbanautomaton:
We've run into this issue with our multi-user capistrano deploy setup, and I'd love to contribute a test that would see this fix (or a similar one) merged.

However, I'm not entirely clear on what the purpose of this line is in the first place. It was [first introduced in this commit](https://github.com/bundler/bundler/commit/ac1fc73a23e980e470823fd7b44919ce58d351c3):

```diff
tree 23538a26d11405d9a29d9bc1803262488b5ee660
parent a5d988498fa5c88cec94091ed7272bd3b3dd7ad4
author Tony Issakov <tony@thefrontiergroup.com.au> Thu May 5 09:10:24 2011 +0800
committer Tony Issakov <tony@thefrontiergroup.com.au> Thu May 5 09:10:24 2011 +0800

fix file permissions after repo checkout. keep in sync with umask


diff --git a/lib/bundler/source.rb b/lib/bundler/source.rb
index 32baf6a..b638b18 100644
--- a/lib/bundler/source.rb
+++ b/lib/bundler/source.rb
@@ -633,6 +633,7 @@ module Bundler
           FileUtils.mkdir_p(path.dirname)
           FileUtils.rm_rf(path)
           git %|clone --no-checkout "#{cache_path}" "#{path}"|
+          File.chmod((0777 & ~File.umask), path)
         end
         Dir.chdir(path) do
           git %|fetch --force --quiet --tags "#{cache_path}"|
```

I'm unsure in what circumstances the umask wouldn't already be respected to necessitate this line (unless `git clone` does something funky, but my own experiments haven't shown it doing so).

@tissak - sorry to drag you into a discussion about a 4-year-old commit of yours, but are you able to shed any light here?

indirect:
I also have no idea why we need to try to apply the umask again, since in my experience git has also respected it. That said, since we don't have a test at the moment, I'm okay with applying this change and just hoping this obscure issue never crops up again. :stuck_out_tongue_closed_eyes:

urbanautomaton:
Thanks very much for applying the fix. :+1: 

A bit of light googling suggests a possible [original reason for the change](http://stackoverflow.com/questions/10637416/git-clone-respects-umask-except-for-top-level-project-directory), namely that `git clone` did indeed once ignore the umask. I think this was fixed in git [here](https://github.com/git/git/commit/45d4fdc), in 2012.

Given that, I would say it's safe to assume some people will still have bugged git versions installed. I don't know how large the intersection of

* people who use old git versions
* people who use very recent bundler versions
* people who know how to use umask

is though. :smile:

Issue #3437, opened 2015-02-27
Support for resolving gems and printing the lockfile to stdout


cstrahan:
This is related to #3327

-----

Right now, there's only one way to get a lockfile for a given Gemfile without first installing the gems: `bundler package --no-install`. This poses a couple problems for OS packaging:

1. It assumes that bundler can persist that `--no-install` flag in `.bundle/config`, which makes it impossible to shell out to bundler from a third-party tool without polluting the user's current directory and current bundler config.
2. It spends a great deal of time downloading the gems, which isn't necessary if all you want to do is resolve the gem versions and dependency graph, which is all we need for maintaining OS packages (and it also pollutes the current directory).
3. Printing the lockfile to stdout would be more composable, compared to putting the lockfile in the current directory.

Having a way to *just* resolve the dependency graph and print the resulting lockfile would be a huge help for people packaging Ruby applications.

("Packaging" here is a bit overloaded. I'm referring to packaging for package managers, like apt/yum/nix, etc)

cstrahan:
Another idea: instead of adding a new option to the CLI, perhaps an effort could be made to create a stable, public API for such things. A great example would [Cabal](https://www.haskell.org/cabal/), which exposes everything it knows about packing [in its API](http://hackage.haskell.org/package/Cabal).

indirect:
Is that the sound of volunteering that I hear? :)

cstrahan:
@indirect At some point, possibly... But I'm pretty burnt out at the moment. I'd sooner pay someone else to do it (can't do that at the moment, though, as I'm living off of savings).

segiddins:
```ruby
require 'bundler'
puts Bundler.definition(true).to_lock
```

indirect:
Ha. You're much more concise than I am. :) I'm convinced that we need a lock command, though, and we might as well have `lock --print` or `lock --stdout` while we're at it. If anyone wants to implement that, go for it.

segiddins:
What would said `lock` command do without options?

indirect:
Resolve the `Gemfile` and write the lockfile into `Gemfile.lock`, but not install any gems.

cstrahan:
@segiddins That snippet is not quite sufficient; if the required gems aren't installed locally, you get an error along the lines of:

[...]

PR #3439, opened 2015-02-27
[Lock] Add lock command

segiddins:
Closes https://github.com/bundler/bundler/issues/3437.

This is obviously missing specs, and I doubt I defined all the stuff correctly in `cli.rb`, but it's a start!

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -379,6 +379,20 @@ module Bundler
       Inject.new(options, name, version, gems).run
     end
 
+    desc "lock", "Creates a lockfile without installing"
+    method_option "update", :type => :boolean, :default => false, :banner =>
+      "ignore the existing lockfile"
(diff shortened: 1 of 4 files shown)


segiddins:
Added some basic specs. Is there any other documentation that I need to update when adding a new command?

indirect:
If you want to, you could write a man page. :grin:

indirect:
oh, also the --local flag should talk about "downloading gemspecs" or "downloading gem metadata", rather than "downloading gems".

cstrahan:
@indirect: Do you know if there's a rough ETA for when this feature will be released?

indirect:
You can use it today by running "rake install" from a checkout of master, but I'd also like to push a prerelease by the end of the week.

cstrahan:
@indirect: Awesome - thanks!

PR #3440, opened 2015-02-27
[Inline] Add bundler/inline to allow using Bundler without a Gemfile

segiddins:
Closes https://github.com/bundler/bundler-features/issues/47.

Enables scripts such as the following to work on their own, as long as bundler is installed:

```ruby
#!/usr/bin/env ruby

require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'json'
  gem 'nap', require: 'rest'
  gem 'cocoapods', '> 0.pre'
end

puts Pod::VERSION
```

### DIFF
--- a/lib/bundler/inline.rb
+++ b/lib/bundler/inline.rb
@@ -16,5 +16,6 @@ def gemfile(install = false, &blk)
   runtime.setup_environment
   runtime.setup.require
 
-  Bundler.send(:define_singleton_method, :root, old_root)
+  bundler_module = class << Bundler; self; end
+  bundler_module.send(:define_method, :root, old_root)
 end
--- a/lib/bundler/inline.rb
+++ b/lib/bundler/inline.rb
@@ -1,4 +1,33 @@
-def gemfile(install = false, &blk)
+# Allows for declaring a Gemfile inline in a ruby script, optionally installing
+# any gems that aren't already installed on the user's system.
+#
+# @note Every gem that is specified in this 'Gemfile' will be `require`d, as if
+#       the user had manually called `Bundler.require`. To avoid a requested gem
+#       being automatically required, add the `:require => false` option to the
+#       `gem` dependency declaration.
+#
+# @param install [Boolean] whether gems that aren't already installed on the
+#                          user's system should be installed.
+#                          Defaults to `false`.
+#
+# @param gemfile [Proc]    a block that is evaluated as a `Gemfile`.
+#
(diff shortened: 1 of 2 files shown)


segiddins:
Those 2 spec failures look like API timeouts to me.

TimMoore:
I restarted the failing jobs.

indirect:
Er, just to double-check, in the example script, you would still have to `require 'cocoapods'`, right?

segiddins:
@indirect right now, using `bundler/inline`'s `gemfile`  method will also call `require` on the environment, just as if you'd somehow done a `require 'bundler/require'` afterwards.

indirect:
There is no `bundler/require`, though. :) There's a `bundler/setup`, but that means you still need to explicitly call `require`.

segiddins:
Ah my bad. It'd be like calling `require 'bundler/setup'; Bundler.require`

indirect:
In order to encourage reasonable growth of a script into a multi-file app, can we leave out the `Bundler.require` part? Requiring the world up front is an anti-pattern that I wish Rails didn't encourage. :'(

segiddins:
André and I just agreed that not auto-requiring in this circumstance would make users do a redundant `gem 'foo'` && `require 'foo'` for every gem that they include in the inline gemfile, which is probably worse than (slightly) encouraging en-mass requiring.

indirect:
Can you document this somewhere, preferably with examples? :D (Maybe even on the Bundler website?!)

andremedeiros:
This is really nice! :+1:

cstrahan:
@segiddins Excellent! This new feature should greatly simplify the way we do Ruby packaging in NixOS. Thanks!

schneems:
> Requiring the world up front is an anti-pattern

Do you have more thoughts on this written down somewhere? I would think we want to preload to take advantage of CoW with forking webservers.

cstrahan:
@schneems I believe the anti-pattern that @indirect is referring to is the fact that - assuming you don't require what you use - your tests (as just one example) can't be run without requiring everything up front. So while the act of requiring everything up front is great (with respect to CoW), you should still require what you need where you need it.

segiddins:
This is not really a relevant discussion for this particular feature, which is targeted at single-file scripts.

schneems:
Thank you @cstrahan, makes sense. Thanks for the feature @segiddins i cede this thread back to you.

PR #3441, opened 2015-02-27
[DSL] Infer ruby version when .ruby-version exists

segiddins:
Closes https://github.com/bundler/bundler-features/issues/79

### DIFF
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -30,6 +30,7 @@ module Bundler
     def eval_gemfile(gemfile, contents = nil)
       contents ||= Bundler.read_file(gemfile.to_s)
       instance_eval(contents, gemfile.to_s, 1)
+      infer_ruby_version(gemfile)
     rescue SyntaxError => e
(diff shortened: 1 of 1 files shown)


indirect:
I think this needs to be optional, because Gemfile ruby versions cause hard failures, while .ruby-version versions don't cause failures of any kind, they just change the Ruby version when you `cd`. This would cause me infinity problems while trying to test against multiple Ruby versions, and I think means this change should be part of 2.0 if we do it.

segiddins:
So I guess in the meantime the workaround is to File.read in the Gemfile?

-Samuel E. Giddins

On Mar 31, 2015, at 6:28 PM, André Arko <notifications@github.com> wrote:

I think this needs to be optional, because Gemfile ruby versions cause hard failures, while .ruby-version versions don't cause failures of any kind, they just change the Ruby version when you cd. This would cause me infinity problems while trying to test against multiple Ruby versions, and I think means this change should be part of 2.0 if we do it.

indirect:
Yeah, the opt-in is pretty straightforward, and doesn’t come with breaking changes.

PR #3447, opened 2015-03-01
Slightly more secure github git sources

steakknife:
Pulling code into production over an insecure protocol usually isn't a good idea, so this PR moves from git -> https.  As a smaller side-benefit, this change also allows Gemfiles including github sources to pass bundler-audit.

(Granted there is not yet any checking of commits' or tags' GPG signatures in bundler, and more non-git sourced gems authors signing their gems would also be a good idea.)

### DIFF
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -140,7 +140,7 @@ module Bundler
               "either use the :git option on a gem, or specify the gems that \n" \
               "bundler should find in the git source by passing a block to \n" \
               "the git method, like: \n\n" \
-              "  git 'git://github.com/rails/rails.git' do\n" \
+              "  git 'https://github.com/rails/rails.git' do\n" \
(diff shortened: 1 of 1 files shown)


TimMoore:
Thanks, but unfortunately we can't take this into the 1.x series. Changing the URL format breaks backwards compatibility in some scenarios.

We already have this change in the 2-0-dev branch. In the meantime, we recommend either using a full, explicit `https` URL or overriding the `github` shortcut in your `Gemfile`.

steakknife:
That'll work. 



—

steakknife:
For posterity and anyone with retentive clipboard auditors, drop this in near the top of `Gemfile` to get the same effect:

```ruby
git_source(:github) do |repo_name|
  "https://github.com/#{repo_name}.git"
end if Bundler::VERSION < '2'
```

PR #3451, opened 2015-03-02
Update man-pages with new options

moeffju:
Some of the commands’ supported options were not reflected in the respective manpages.
I believe I got them all, but automating this would probably we worthwile (keep bundler.io, manpages and optparse in sync).


indirect:
We would definitely like to generate the website pages from the man pages. Would you be interested in implementing it? :)

indirect:
Closing this pull request since the changes are made in the current 1.9 release. Thanks for the patch!

PR #3452, opened 2015-03-02
add debug log for api request limit

neerfri:


lib/bundler/source/rubygems.rb
```diff
@@ -302,6 +302,7 @@ def remote_specs
           # the gemspecs of those gems, if the non-api sites contain more than
           # about 100 gems, we just treat all sites as non-api for speed.
           allow_api = idx.size < API_REQUEST_LIMIT && dependency_names.size < API_REQUEST_LIMIT
+          Bundler.ui.debug "Need to query more then #{API_REQUEST_LIMIT} gems" unless allow_api
```
andremedeiros: Can we add a bit more information? 

```
Need to query more than 200 gems. Downloading full index instead...
```

neerfri: Sure.
My first version included such info, but I've removed it since the current output will be:

```
$ DEBUG=1 bundle
HTTP GET https://bundler.rubygems.org/api/v1/dependencies
HTTP 200 OK
Need to query more then 100 gems
Fetching source index from https://rubygems.org/
```

Do you think the following is better?:
```
$ DEBUG=1 bundle
HTTP GET https://bundler.rubygems.org/api/v1/dependencies
HTTP 200 OK
Need to query more then 100 gems. Downloading full index instead...
Fetching source index from https://rubygems.org/
```

andremedeiros: I like the second one better. Cause *and* effect.

Also, could you change `then` to `than`? Thank you.

TimMoore:
:+1:

TimMoore:
Thanks!

neerfri:
My pleasure

Issue #3448, opened 2015-03-01
Warn on git protocol URLs


TimMoore:
We've had a number of pull requests submitted to change the `github` DSL shortcut to use `https` instead of the `git` protocol (most recently #3447).

This breaks backwards compatibility because the URL is expanded before it goes in the lock file, so updating Bundler would create a change in the lock file even without changing the `Gemfile` at all. This can break deployment if the server has a newer version of Bundler than what was used at development time.

The 2.0 branch has already made the change, but in 1.9 we should warn on any `git` protocol URLs (including ones expanded from `github`) and suggest replacing them with explicit URLs or overriding the shortcut in the `Gemfile`.

(Maybe we should even get rid of the shortcut entirely in 2.0 and encourage people to define their own.)

indirect:
:+1:

rgb-one:
I think this issue can be closed following #3453

PR #3453, opened 2015-03-02
[SourceList] Warn when adding a git source using the git procotol

segiddins:
Closes https://github.com/bundler/bundler/issues/3448.

### DIFF
--- a/lib/bundler/source_list.rb
+++ b/lib/bundler/source_list.rb
@@ -15,7 +15,12 @@ module Bundler
     end
 
     def add_git_source(options = {})
-      add_source_to_list Source::Git.new(options), git_sources
+      source = add_source_to_list Source::Git.new(options), git_sources
+      if source.uri =~ %r{^git:}
+        Bundler.ui.warn "The git source `#{source.uri}` uses the `git` protocol, " \
+                        "please consider changing it to `https`, which is more secure."
+      end
+      source
     end
 
     def add_rubygems_source(options = {})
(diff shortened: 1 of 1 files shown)


TimMoore:
Thanks! :blue_heart: 

Couple of things :smile: 

1. The tests are failing now
2. I was thinking we should say something different for people using the `github` shortcut, since they might not know that it's using a git protocol URL, or that they can redefine the shortcut

So I think the best place to issue the warnings is from `dsl.rb`, but it will take a bit of refactoring to make it possible to issue different warnings for github vs git.

segiddins:
I don't have the time to do that sort of refactor at the moment, but I think I fixed the test failures.

indirect:
Maybe in the meantime we should add a `github_https` option? Basically I would like to deprecate github immediately, but we can't. :(

TimMoore:
I'd like to deprecate all built-in shortcuts. I think the translation logic needs to be explicit in the `Gemfile`, because we can never change it in Bundler without breaking things.

indirect:
I feel like I'm probably okay with `https` shortcuts because they seem very unlikely to change in the future, and work for both public and private repos... but famous last words, maybe. :P 

Github is _such_ a common case that it seems completely valid to have that one built in, but I think we should drop everything else in 2.0, and document the shortcut DSL better.

indirect:
Ready to merge this as soon as the tests pass and there is a conflig flag to disable it (and mention the config flag in the warning).

segiddins:
Done

pducks32:
The flags have to be set as true in settings.rb in order to be returned as true. It's fine for true because everything resolves to true but for false it will return `"false"` which is true.

segiddins:
@indirect it looks like this has been cherry-picked into 1-99-dev already?

indirect:
Yup, thanks!

PR #3461, opened 2015-03-07
[Travis] Ruby 1.8.7 and 1.9.3 have been retired

maclover7:
https://www.ruby-lang.org/en/news/2013/06/30/we-retire-1-8-7/

https://www.ruby-lang.org/en/news/2015/02/23/support-for-ruby-1-9-3-has-ended/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,6 @@ rvm:
   - 2.2
   - 2.1
   - 2.0.0
-  - 1.9.3
   - 1.8.7
 
 # Rubygems versions MUST be available as rake tasks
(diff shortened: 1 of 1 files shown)


BanzaiMan:
"They are no longer supported" and "They are no longer used so Bundler should not support them" are completely different things.

Allow them to fail, maybe, but do not remove them.

indirect:
Thanks for the pull request! While 1.8.7 and 1.9.3 will no longer get patches or security updates, Bundler follows [SemVer](semver.org), which means that every 1.x version of Bundler needs to do everything that 1.0 did. That includes working on 1.8.7 and 1.9.3. We plan to drop support for 1.8.7 and 1.9.3 in Bundler 2.0, since that is the first time we will be able to make changes that are not backwards compatible.

Issue #3436, opened 2015-02-27
CODE_OF_CONDUCT always created


shortdudey123:
It appears the the `CODE_OF_CONDUCT.md` file is created even when `n` is selected on the first run

```
vagrant@vagrant-ubuntu-trusty-64:~/tmp$ bundle gem test
Creating gem 'test'...
Do you want to include a code of conduct in gems you generate?
Codes of conduct can increase contributions to your project by contributors who prefer collaborative, safe spaces. You can read more about the code of conduct at contributor-covenant.org. Having a code of conduct means agreeing to the responsibility of enforcing it, so be sure that you are prepared to do that. For suggestions about how to enforce codes of conduct, see bit.ly/coc-enforcement. y/(n): n
Do you want to license your code permissively under the MIT license?
This means that any other developer or company will be legally allowed to use your code for free as long as they admit you created it. You can read more about the MIT license at choosealicense.com/licenses/mit. y/(n): y
Do you want to generate tests with your gem?
Type 'rspec' or 'minitest' to generate those test files now and in the future. rspec/minitest/(none): rspec
      create  test/Gemfile
      create  test/.gitignore
      create  test/lib/test.rb
      create  test/lib/test/version.rb
      create  test/test.gemspec
      create  test/Rakefile
      create  test/README.md
      create  test/bin/console
      create  test/bin/setup
      create  test/CODE_OF_CONDUCT.md
      create  test/LICENSE.txt
      create  test/.travis.yml
      create  test/.rspec
      create  test/spec/spec_helper.rb
      create  test/spec/test_spec.rb
Initializing git repo in /home/vagrant/tmp/test
vagrant@vagrant-ubuntu-trusty-64:~/tmp$ 
```
```
vagrant@vagrant-ubuntu-trusty-64:~/tmp$ cat ~/.bundle/config 
---
BUNDLE_GEM__COC: false
BUNDLE_GEM__MIT: true
BUNDLE_GEM__TEST: rspec
vagrant@vagrant-ubuntu-trusty-64:~/tmp$ 
```

Subsequent `bundle gem XXX` calls don't create it
```
vagrant@vagrant-ubuntu-trusty-64:~/tmp$ bundle gem test2
Creating gem 'test2'...
      create  test2/Gemfile
      create  test2/.gitignore
      create  test2/lib/test2.rb
      create  test2/lib/test2/version.rb
      create  test2/test2.gemspec
      create  test2/Rakefile
      create  test2/README.md
[...]

PR #3465, opened 2015-03-09
Fixed ’no’ option in gem generator

kirs:
`result || Bundler.settings["gem.#{key}"]` didn't gave expected result when `result` was `false`

Fix for #3436

### DIFF
--- a/lib/bundler/cli/gem.rb
+++ b/lib/bundler/cli/gem.rb
@@ -120,17 +120,21 @@ module Bundler
     end
 
     def ask_and_set(key, header, message)
-      result = options[key]
+      choice = options[key]
       if !Bundler.settings.all.include?("gem.#{key}")
-        if result.nil?
+        if choice.nil?
           Bundler.ui.confirm header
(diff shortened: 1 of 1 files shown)


TimMoore:
Thanks!

If you have some more time, could I ask for a couple more things?

1. A spec to make sure there aren't future regressions
2. The pull request to be against the 1-8-stable branch instead of master

kirs:
Do you have any idea how to test `Bundler.ui.ask` from specs? Tests are missing because I couldn't find any solution yet.

indirect:
IIRC you can use the bundle test helper with a block to get access to the stdin for that process. Check the helper definition to be sure, though.

kirs:
Not really, because [`Silent`](https://github.com/bundler/bundler/blob/master/lib/bundler/ui/silent.rb) is used in tests.

indirect:
Cherry-picked to the 1.8 branch as edb4b50e74dcd5dec877027218cfc5970a60bcc7, tests added in 22590270f3af681838e7cbef699fa25f734126f4.

kirs:
:+1:

PR #3466, opened 2015-03-10
Remove unused dups

kohgpat:
Hello!

I've started looking into bundler sources recently and found that dups call in cli/install.rb can be removed, options already dups in cli.rb before passing into cli/install.rb initializer.

Tests successfully passes locally and I hope I've no broke anything.

Thank you for your hard work!

Sincerelly, Nikolay.


indirect:
Thanks!

Issue #3457, opened 2015-03-04
RuntimeError: can't modify frozen String


easyrider-new:
Error details

    RuntimeError: can't modify frozen String
    /usr/lib/ruby/1.9.1/rubygems/version.rb:191:in `strip!'
    /usr/lib/ruby/1.9.1/rubygems/version.rb:191:in `initialize'
    /home/kk/projects/ultravie/Gemfile:47:in `new'
    /home/kk/projects/ultravie/Gemfile:47:in `eval_gemfile'
    /home/kk/.rvm/gems/ruby-1.8.7-p358@ultravie/gems/bundler-1.8.3/lib/bundler/dsl.rb:32:in `instance_eval'
    /home/kk/.rvm/gems/ruby-1.8.7-p358@ultravie/gems/bundler-1.8.3/lib/bundler/dsl.rb:32:in `eval_gemfile'
    /home/kk/.rvm/gems/ruby-1.8.7-p358@ultravie/gems/bundler-1.8.3/lib/bundler/dsl.rb:10:in `evaluate'
    /home/kk/.rvm/gems/ruby-1.8.7-p358@ultravie/gems/bundler-1.8.3/lib/bundler/definition.rb:25:in `build'
    /home/kk/.rvm/gems/ruby-1.8.7-p358@ultravie/gems/bundler-1.8.3/lib/bundler.rb:155:in `definition'
    /home/kk/.rvm/gems/ruby-1.8.7-p358@ultravie/gems/bundler-1.8.3/lib/bundler/cli/install.rb:78:in `run'
    /home/kk/.rvm/gems/ruby-1.8.7-p358@ultravie/gems/bundler-1.8.3/lib/bundler/cli.rb:157:in `install'
    /home/kk/.rvm/gems/ruby-1.8.7-p358@ultravie/gems/bundler-1.8.3/lib/bundler/vendor/thor/command.rb:27:in `run'
    /home/kk/.rvm/gems/ruby-1.8.7-p358@ultravie/gems/bundler-1.8.3/lib/bundler/vendor/thor/invocation.rb:121:in `invoke_command'
    /home/kk/.rvm/gems/ruby-1.8.7-p358@ultravie/gems/bundler-1.8.3/lib/bundler/vendor/thor.rb:363:in `dispatch'
    /home/kk/.rvm/gems/ruby-1.8.7-p358@ultravie/gems/bundler-1.8.3/lib/bundler/vendor/thor/base.rb:440:in `start'
[...]

Issue #3187, opened 2014-09-25
`strip!': can't modify frozen String (RuntimeError)


dipolesource:
Hi Bundler team!

Got the following failure error on MRI1.9.3 but not when I switched to MRI2.1.2. I don't need resolution. Just reporting because it asked me to. Cheers!

[master]lotus: git remote -v
origin	https://github.com/lotus/lotus.git (fetch)
origin	https://github.com/lotus/lotus.git (push)

[master]lotus: git rev-parse head
e2f2b8112018cddf9fc9020986732a7e9ad8fd0e

````
[master]lotus: bundle
Unfortunately, a fatal error has occurred. Please report this error to the Bundler issue tracker at https://github.com/carlhuda/bundler/issues so that we can fix it. Please include the full output of the command, your Gemfile and Gemfile.lock. Thanks!
/Users/dipolesource/.rbenv/versions/1.9.3-p125/lib/ruby/1.9.1/rubygems/version.rb:191:in `strip!': can't modify frozen String (RuntimeError)
	from /Users/dipolesource/.rbenv/versions/1.9.3-p125/lib/ruby/1.9.1/rubygems/version.rb:191:in `initialize'
	from /Users/dipolesource/.rbenv/versions/1.9.3-p125/lib/ruby/1.9.1/rubygems/version.rb:178:in `new'
	from /Users/dipolesource/.rbenv/versions/1.9.3-p125/lib/ruby/1.9.1/rubygems/version.rb:178:in `create'
	from /Users/dipolesource/.rbenv/versions/1.9.3-p125/lib/ruby/1.9.1/rubygems/specification.rb:2092:in `version='
	from /Users/dipolesource/Sites/lotus/lotusrb.gemspec:8:in `block in <main>'
	from /Users/dipolesource/.rbenv/versions/1.9.3-p125/lib/ruby/1.9.1/rubygems/specification.rb:1346:in `initialize'
	from /Users/dipolesource/Sites/lotus/lotusrb.gemspec:6:in `new'
	from /Users/dipolesource/Sites/lotus/lotusrb.gemspec:6:in `<main>'
	from /Users/dipolesource/.rbenv/versions/1.9.3-p125/lib/ruby/gems/1.9.1/gems/bundler-1.1.2/lib/bundler.rb:280:in `eval'
[...]

PR #3469, opened 2015-03-11
Catch and report RuntimeError on Gemfile eval

arthurnn:
[Fixes https://github.com/bundler/bundler/issues/3457 and https://github.com/bundler/bundler/issues/3187]

review @segiddins @indirect @andremedeiros 

For more context on the error see my spec, and both issues.

(not sure where the Changelog entry should go, if to 1.8.x or 1.9.x, let me know)

### DIFF
--- a/spec/bundler/dsl_spec.rb
+++ b/spec/bundler/dsl_spec.rb
@@ -180,4 +180,12 @@ describe Bundler::Dsl do
         to raise_error(Bundler::GemfileError, /Gemfile syntax error/)
     end
   end
+
+  describe "Runtime errors", :unless => Bundler.current_ruby.on_18? do
+    it "will raise a Bundler::GemfileError" do
+      gemfile "s = 'foo'.freeze; s.strip!"
+      expect { Bundler::Dsl.evaluate(bundled_app("Gemfile"), nil, true) }.
+        to raise_error(Bundler::GemfileError, /There was an error in your Gemfile/)
+    end
+  end
 end
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -33,7 +33,7 @@ module Bundler
     rescue SyntaxError => e
       syntax_msg = e.message.gsub("#{gemfile}:", 'on line ')
       raise GemfileError, "Gemfile syntax error #{syntax_msg}"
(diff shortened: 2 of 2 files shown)


andremedeiros:
I like it. Let the specs run and we'll merge it.

As for Changelog, that gets done whenever we release, so don't worry!

segiddins:
Why not just rescue all exceptions that come from the eval?

andremedeiros:
I'd rather add them as we go, so that we can maybe define nice error messages for each, than have a blanket `rescue` there.

arthurnn:
@segiddins That was kinda my first thought too, Like catch StandardErrors.. But I didnt want to change behavior that much... I can do that tho, if you think it is right.

arthurnn:
rebased

arthurnn:
the `string.freeze` wont fail on ruby 1.8, so the regression test should run for 1.9+

spec/bundler/dsl_spec.rb
```diff
@@ -180,4 +180,14 @@
         to raise_error(Bundler::GemfileError, /Gemfile syntax error/)
     end
   end
+
+  if RUBY_VERSION > "1.9"
+    describe "Runtime errors" do
+      it "will raise a Bundler::GemfileError" do
+        gemfile "s = 'foo'.freeze; s.strip!"
```
segiddins: `raise RuntimeError, "Error message"`?

andremedeiros:
@arthurnn can you use this instead?

```ruby
unless Bundler.current_ruby.on_18?
  describe "Runtime errors" do
    # Spec goes here
  end
end
```

spec/bundler/dsl_spec.rb
arthurnn: I didnt want to force raise a RuntimeError, in case that if ruby changes what `freeze` raises we would catch that on this regression test.

arthurnn:
@andremedeiros done.

andremedeiros:
Actually, sorry, I'm a huge idiot. This is correct:

```ruby
describe "Runtime errors", unless: Bundler.current_ruby.on_18? do
  # Spec goes here
end
```

arthurnn:
@andremedeiros done.

andremedeiros:
:+1:

PR #3471, opened 2015-03-12
[Gemfile] Add a post_install hook

segiddins:
Closes https://github.com/bundler/bundler-features/issues/83.

This time, I wrote specs!

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -5,7 +5,7 @@ module Bundler
   class Definition
     include GemHelpers
 
-    attr_reader :dependencies, :platforms, :ruby_version, :locked_deps
+    attr_reader :dependencies, :platforms, :ruby_version, :post_install_hook, :locked_deps
 
     # Given a gemfile and lockfile creates a Bundler definition
     #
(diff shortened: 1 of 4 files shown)


TimMoore:
amaze :sparkle: 

(just waiting for specs to pass before merging)

segiddins:
Let's wait to merge this until we're sure we want it.

rzane:
Hi @segiddins, this looks really good. You're a speed deamon. I was working on this myself, pushed, and realized you also took a stab. I like that you passed some useful info to the block.

My version is a little different, so maybe we can work together and take the best aspects from both. Here's what I did differently:
1. It adds pre and post install hooks (before_install / after_install)
2. You can add as many hooks as you want.
3. When passed a string, it'll execute it as a shell command, which is what NPM does.

See https://github.com/rzane/bundler/tree/hooks.

maclover7:
@segiddins Nice PR! For the other hooks we are going to be creating, do you suggest we follow a similar process?

indirect:
This PR spurred some discussion, and we decided to implement hooks in a slightly different way. We're working on it!

RochesterinNYC:
@indirect where is the discussion regarding this PR or the context on how hooks were implemented in a slightly different way? Thanks.

PR #3475, opened 2015-03-14
chore(template): remove repository specific files from gem files

sanemat:
Remove repository specific files from gem files

e.g. `.travis.yml`, `circle.yml`, `.drone.yml`, `.tachikoma.yml`

We don't use these files from gem, and sometimes these files contain encrypted keys.


TimMoore:
This is going to be very project-specific. Maybe it should exclude all dot files by default? I'm not sure.

sanemat:
I see. This is too project-specific. Thanks

PR #3476, opened 2015-03-14
Rubygems remote refactor

TimMoore:
This is part one of a larger refactoring to pull more code into smaller unit-tested classes. I'll try to submit small, focused pull requests so that it's easier to review (and so that I can get feedback before going too deep).

This one just renames `AnonymizableURI` to `Source::Rubygems::Remote`. I plan to move more stuff into `Remote` from `Fetcher` after this.


lib/bundler/source/rubygems.rb
```diff
@@ -84,10 +86,10 @@ def install(spec)
 
         # Download the gem to get the spec, because some specs that are returned
         # by rubygems.org are broken and wrong.
-        if spec.source_uri
+        if spec.remote
           # Check for this spec from other sources
-          uris = [spec.source_uri.without_credentials]
-          uris += source_uris_for_spec(spec)
+          uris = [spec.remote.anonymized_uri]
+          uris += remotes_for_spec(spec).map { |remote| remote.anonymized_uri }
```
TimMoore: I think I'll turn this into an array of remotes instead of an array of URIs in a future change. Need to implement `==` for `Remote` first.

indirect:
yessssssssssss 👏👹

PR #3478, opened 2015-03-15
Move remote URI resolution logic into Remote

TimMoore:


indirect:
At this rate, we're going to start having nice code soon. :D

Issue #82, opened 2010-02-15
Add a way to effect load order in lieu of explicit dependency between gems


midas:
There are some fringe cases where a gem is made as a standalone that can work with some other gem if available.  In this scenario, the gem author does not want to explicitly depend on the other gem but we need a way to effect load order to get convenience functionality to mix in.

Example:

Gem1 implements functionality that can be used in any Ruby project, but also has a convenience module that works with ActiveRecord.  If ActiveRecord is defined when Gem1 is included, the functionality will be mixed-in, otherwise it will not.  However, Gem1's author does not wish to force users of Gem1 to install the ActiveRecord gem if they are not using it.

When you try to use bundler in this scenario, your hands are tied and your desired functionality may not get mixed in, depending on how the lock environment file is ordered.  Some kind of option for the gem method that would allow a user defined dependency to be set up could alleviate this issue:

gem 'Gem1', '>= 1.0.0', :ensure_loaded => 'activerecord'
  or
gem 'Gem1', '>= 1.0.0', :ensure_loaded => %w(activerecord mysql)

Then bundler will have the information it needs to order the locked environment file correctly.

PR #3479, opened 2015-03-17
Implemented use-oldest feature. (#82)

Sxl1092:
@ajgrover, @rrperez, @bgott, and I are 4 Stanford Students who in issue #82 requested a feature where we can use the flag --use-oldest to run out code using the oldest versions of the gems our gemfile claims. This will be used to test to ensure that our code works with the oldest versions it claims it does. We added the code to make this small addition and have a test to go along with it. The only issue right now is that running bundle with this flag will not properly work if the Gemfile.lock file is present, because it will simply install the most recent versions again. The fix currently is to delete the file before running bundle --use-oldest. Thanks in advance for any feedback!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Features:
 
   - prefer gemspecs closest to the directory root (#3428, @segiddins)
   - debug log for API request limits (#3452, @neerfri)
+  - implemented use-oldest feature (#82, @sxl1092)
 
 "Features":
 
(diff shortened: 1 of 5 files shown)


indirect:
It seems like this change broke some tests, and some gems are being installed as the oldest rather than the newest. Can you fix the failing specs?

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ Features:
 
   - prefer gemspecs closest to the directory root (#3428, @segiddins)
   - debug log for API request limits (#3452, @neerfri)
+  - implemented use-oldest feature (#82, @sxl1092)
```
segiddins: this doesn't belong here

segiddins:
this definitely needs some more specs for more complicated Gemfiles

indirect:
@Sxl1092, @ajgrover, @rrperez, @bgott: thanks for the pull request! the specs are failing. can you fix them? if you do, we're still interested in this patch.

PR #3480, opened 2015-03-17
[DSL] Add prettier DSL error

segiddins:
Goes hand in hand with https://github.com/CocoaPods/Core/pull/226.

### DIFF
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -179,7 +179,6 @@ module Bundler
     end
 
     def method_missing(name, *args)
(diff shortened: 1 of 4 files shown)


indirect:
This is great! Instead of saying "invalid Gemfile", can we keep the error message closer to the original? Something like "There was an error while parsing the Gemfile:". It's a small change, but keeping error messages as human-friendly as possible is important.

segiddins:
This should finally be green! It turns out, different rubies have really different ideas on how to report that `}` is an invalid ruby file :)

PR #3483, opened 2015-03-19
Improve generating module name from gem name in gem generator

janlelis:
I encountered the bug that the gem generator would fail on gem names containing two dashes (like `a--a`). I updated the module name generator to use a more robust regex based approach.

### DIFF
--- a/lib/bundler/cli/gem.rb
+++ b/lib/bundler/cli/gem.rb
@@ -20,8 +20,8 @@ module Bundler
 
       underscored_name = name.tr('-', '_')
       namespaced_path = name.tr('-', '/')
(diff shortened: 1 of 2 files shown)


indirect:
Thank you for the patch! Unfortunately we can't accept it because your change breaks the tests. Could you update this pull request so the tests pass, and add a test for the gem with two dashes in its name? Thanks.

janlelis:
I rewrote the regex to be 1.8 compatible and added a simple test.

janlelis:
The travis fail seems to be a timeout.

indirect:
Thanks!

PR #3485, opened 2015-03-20
Track Bundler version in lockfile

smlance:
Let me know if anything should be changed!


lib/bundler/definition.rb
```diff
@@ -256,9 +256,39 @@ def lock(file)
...
     end
 
+    # Returns the version of Bundler that is creating or has created
+    # Gemfile.lock. Used in #to_lock.
+    def lock_version
+      lockfile = @lockfile_contents
+      lock_ver = ""
+      curr_ver = ""
+
+      if !lockfile.empty? and lockfile.match(/^  \[(.*)\]$/)
```
indirect: Since the idiomatic `!thing.empty?` in ruby is `thing.any?`, I think it would probably make sense to switch to that.

lib/bundler/definition.rb
```diff
@@ -256,9 +256,39 @@ def lock(file)
         "#{File.expand_path(file)}"
     end
 
+    # Returns the version of Bundler that is creating or has created
+    # Gemfile.lock. Used in #to_lock.
+    def lock_version
+      lockfile = @lockfile_contents
+      lock_ver = ""
+      curr_ver = ""
```
indirect: Unless I'm missing something, these variables can just be `nil`, and then your `.empty?` checks can simply be `.nil?` checks

lib/bundler/definition.rb
segiddins: Except I believe `thing.any?` is linear on size?

lib/bundler/definition.rb
```diff
@@ -256,9 +256,39 @@ def lock(file)
...
+          ver = curr_ver
+        else
+          ver = lock_ver
+        end
+      else
+        ver = Bundler::VERSION
+      end
+
+      ver
+    end
```
indirect: So I kept mulling this over, and I think I see some ways to cut down on the amount of code that exists without making it (much?) more obscure:

```ruby
lock_ver = @lockfile_contents[/^  \[(.*)\]$/, 1] if lockfile

if lock_ver && Gem::Version.new(lock_ver) < Gem::Version.new(Bundler::VERSION)
  new_ver = Bundler::VERSION
end

new_ver || lock_ver || Bundler::VERSION
```

smlance:
Yes, that part is a bit verbose. Thanks!

lib/bundler/definition.rb
smlance: Yes, that part is a bit verbose. Thanks!

lib/bundler/definition.rb
```diff
@@ -256,9 +256,30 @@ def lock(file)
         "#{File.expand_path(file)}"
     end
 
+    # Returns the version of Bundler that is creating or has created
+    # Gemfile.lock. Used in #to_lock.
+    def lock_version
+      lockfile = @lockfile_contents
+      lock_ver = nil
+      curr_ver = nil
```
indirect: Oh, with the new succinct code, all three of these lines can be removed. :)

lib/bundler/definition.rb
```diff
@@ -256,9 +256,30 @@ def lock(file)
...
+
+      lock_ver = @lockfile_contents[/^  \[(.*)\]$/, 1] if lockfile
```
indirect: and I lied, this needs to be @lockfile_contents before those three lines can be removed.

indirect:
Seems like this build is still failing, any chance for a fix?

lib/bundler/definition.rb
```diff
@@ -256,9 +256,26 @@ def lock(file)
...
+      end
+
+      new_ver || lock_ver || Bundler::VERSION
+    end
+
     def to_lock
       out = ""
 
+      # Record the version of Bundler that was used to create the lockfile
+      out << "LOCKED WITH\n"
```
segiddins: `BUNDLED WITH`?

lib/bundler/lockfile_parser.rb
```diff
@@ -31,6 +32,22 @@ def initialize(lockfile)
 
       @rubygems_aggregate = Source::Rubygems.new
 
+
+      # If the lockfile's recorded version is older than the current version
+      # of Bundler, we warn the user to update Bundler.
+      if lockfile.match(/^  \[(.*)\]$/)
```
segiddins: any reason this doesn't use the state-based parsing?

lib/bundler/lockfile_parser.rb
```diff
@@ -31,6 +32,22 @@ def initialize(lockfile)
...
+        lock_ver = $1
+        curr_ver = Bundler::VERSION
+        prerelease_text = lock_ver.match(/pre/) ? " --pre" : ""
```
segiddins: this ought to parse the `Gem::Version` to see if it's a pre-release version

indirect:
This looks great! Can you squash this pull request down so it's a single logical change? Then it's good to merge. :ship:

smlance:
Closing since #3546 was merged.

PR #3486, opened 2015-03-20
Missing regex escape in bin/bundler

niku4i:


segiddins:
I think this is more _technically_ correct, but I don't think it fixes any bugs?

indirect:
I'm happy to be as precise as possible with that, because we've had issues with insufficiently-specific regexes in the past :P

niku4i:
Thanks for merging!  Why I sent this patch is that I was just exploring the difference between bin/bundle and bin/bundler. So then I noticed there was just this tiny diff.

PR #3491, opened 2015-03-21
Add bundle_ruby deprecation warning

smlance:


bin/bundle_ruby
```diff
@@ -41,6 +41,8 @@ module Bundler
   end
 end
 
+puts "Warning: bundle_ruby will be deprecated in Bundler 2.0.0."
```
segiddins: pretty sure this needs to go to STDERR, otherwise piping the output of `bundle_ruby` will be broken

PR #3496, opened 2015-03-22
[GemTasks] Add a basic spec that the relevant tasks are installed

segiddins:
\c @indirect

### DIFF
--- a/bundler.gemspec
+++ b/bundler.gemspec
@@ -16,10 +16,11 @@ Gem::Specification.new do |s|
   s.required_ruby_version     = '>= 1.8.7'
   s.required_rubygems_version = '>= 1.3.6'
 
-  s.add_development_dependency 'mustache', '0.99.6'
+  s.add_development_dependency 'mustache',  '0.99.6'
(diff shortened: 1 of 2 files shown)


indirect:
looks like it's failing on 1.8 and 1.9, could you take a look?

segiddins:
It looks like it's failing because the rake command isn't available or something?

indirect:
Hmmmm. Looks like `sys_exec "rake"` is running a system-level rake executable without `GEM_HOME` and things set correctly for rake to load.

indirect:
So it turned out to be that GEM_HOME was set to the testing GEM_HOME, which didn't have rake in it. I fixed the tests by pointing to the test support gems, which include rake, instead.

(This wasn't an error on Ruby 2+ because IIRC those versions of Ruby include Rake in the stdlib, so it was impossible to not have rake installed.)

indirect:
51db7efb0db6a077a5e3b9222ff74d1aabf9f169 is the commit that fixed the tests for me locally on 1.8 and 1.9.

Issue #3009, opened 2014-05-03
Unexpected behavior when `Bundler.setup` is called multiple times


alexeymuranov:
With the following setup:
```ruby
# Gemfile
source 'https://rubygems.org'
ruby '2.1.1'
gem 'rack-test', :group => :test
```
```ruby
# app.rb
require 'bundler'
Bundler.setup(:default, :test)
# Bundler.setup(:test)
Bundler.setup(:default)
require 'rack/test'
```
running
```shell
ruby app.rb
```
shows no errors.

However, if the line with `Bundler.setup(:test)` is *uncommented*, i get the error:
```
app.rb:5:in `require': cannot load such file -- rack/test (LoadError)
	from app.rb:6:in `<main>'
```
This is very unexpected to me.

PR #3497, opened 2015-03-22
[Setup] Fix multiple non-additive invocations of Bundler.setup with diff...

segiddins:
...erent groups

Closes https://github.com/bundler/bundler/issues/3009.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -120,12 +120,7 @@ module Bundler
         # Load all groups, but only once
         @setup = load.setup
       else
-        @completed_groups ||= []
-        # Figure out which groups haven't been loaded yet
-        unloaded = groups - @completed_groups
-        # Record groups that are now loaded
-        @completed_groups = groups
-        unloaded.any? ? load.setup(*groups) : load
+        load.setup(*groups)
       end
     end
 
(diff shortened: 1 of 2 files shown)


segiddins:
@indirect even though I closed this PR, maybe you'd like to take a look?

indirect:
Oh. This looks correct, yeah. Any reason we shouldn't merge this?

indirect:
Oh, I see, this conflicts with the ability to replace the load path with a subsequent call to `setup`. Hmm. :/ We need to preserve the current semantics for backwards compatibility. What's the argument for changing the behaviour to this instead for 2.0?

segiddins:
Could you explain what exactly the current semantics are?

indirect:
To provide some total pseudocode, the current semantics are that `Bundler.setup(*groups)` is equivalent to `$LOAD_PATH = groups.map{|g| g.load_path }.flatten`. Your change makes it so that `Bundler.setup(*groups)` is additive instead of replaces. The reason for the replacing semantics is so that some people (I don't remember who anymore) could use `Bundler.setup` to remove certain groups from the load path.

segiddins:
Are there any special semantics for the default group, perchance?

indirect:
Not that I recall? But I haven't checked the code for them.

segiddins:
Think I might've actually 'fixed' things by making them more consistent (accomplished entirely by deleting code)

indirect:
Huh, amazing. :)

PR #3499, opened 2015-03-22
Apply test framework when install with asks

### LINKED ISSUE #3410: rspec BUNDLE_GEM__TEST setting doesn't add rspec to Gemfile
Causes this error: 
```
bundler-1.8.2/lib/bundler/rubygems_integration.rb:261:in `block in replace_gem': rspec-core is not part of the bundle. Add it to Gemfile. (Gem::LoadError)
```

Wouldn't it make sense to add rspec to the Gemfile when `BUNDLE_GEM__TEST` is set to `rspec` and the new gem is generated with spec directory?

marocchino:
- resolve #3410

### DIFF
--- a/lib/bundler/cli/gem.rb
+++ b/lib/bundler/cli/gem.rb
@@ -74,6 +74,7 @@ module Bundler
       end
 
       if test_framework = ask_and_set_test_framework
+        config[:test] = test_framework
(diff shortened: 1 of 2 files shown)


indirect:
Thanks!

PR #3501, opened 2015-03-23
add before_install to .travis.yml template for new gems

kodnin:
As seen in [this](https://travis-ci.org/kodnin/defekt/jobs/55475451) Travis CI build and the screenshot below, Bundler 1.9 is not installed by default on Travis CI. This leaves us a problem, which can be resolved by installing Bundler 1.9 [before](http://docs.travis-ci.com/user/languages/ruby/#Dependency-Management) running Bundler.

screen shot 2015-03-23 at 13 49 15


indirect:
Seems to make sense. Thanks!

PR #3505, opened 2015-03-24
disable_shared_gems independent from current path setting

bobishh:
Hi there, in this PR i tried to fix some strange behaviour. - Bundler doesn't respect global config for disable_shared_gems setting. Feel free to point me things you might want to change! <3 <3 <3

### DIFF
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -70,7 +70,7 @@ module Bundler
       Bundler.settings[:clean]    = options["clean"] if options["clean"]
       Bundler.settings.without    = options[:without]
       Bundler::Fetcher.disable_endpoint = options["full-index"]
-      Bundler.settings[:disable_shared_gems] = Bundler.settings[:path] ? '1' : nil
+      Bundler.settings[:disable_shared_gems] ||= Bundler.settings[:path] ? '1' : nil
 
       # rubygems plugins sometimes hook into the gem install process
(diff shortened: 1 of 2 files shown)


indirect:
Thanks for the patch! Can you explain how you are using Bundler that led you to notice this, and why the way it works now is confusing?

bobishh:
Sure, the reason for PR is such behaviour - http://showterm.io/40f712b4af501b91ca2e2, if global config is intended to be default why can't I combine 2 global config values? It seems like "disable_shared_gems" variable is dependent on "path" variable which seems unintuitive. @brainopia mentioned this at first

indirect:
Well, `disable_shared_gems` is an internal config, and Bundler officially does not support enabling shared gems with path. So you're trying to do something we don't support, and don't have tests for. :)

bobishh:
What about adding support for that then? I wrote few specs that emulate behaviour shown in term session above, and they pass, nothing seem to be broken with it. And it simply omits setting disable_shared_path based on path if it is already present in global config. It might be the rare case but someone might want it, at least few of my friends do.

indirect:
We used to support this mode (in Bundler 0.9) and found that it created hard to understand and hard to debug situations. We decided against allowing it on purpose to avoid those problems, and so we do not support or recommend the configuration that you are describing here.




Bundler 2, coming sometime in the future, will share .gem files across all bundles on the system, even with --path.

bobishh:
Ok. Closing the PR then.

PR #3513, opened 2015-03-26
Improve Minitest gem template

### LINKED ISSUE #3512: Minitest template is wrong
Bundler creates `test/test_#{my_gem}.rb` file with class `TestMyGem`, but common practice is to create `test/#{my_gem}_test.rb` file with class `MyGemTest`.

Also - why test helper called `minitest_helper`?

arthurnn:
Default minitest test/ folder structure should be:
- test/test_helper.rb
- test/GEMNAME_test.rb

[fixes #3512]

review @indirect @andremedeiros 
cc @yuriihabrusiev


simi:
:+1: I was thinking about similar PR few days ago.

spec/commands/newgem_spec.rb
```diff
@@ -499,20 +499,20 @@ def create_temporary_dir(dir)
       end
 
       it "builds spec skeleton" do
-        expect(bundled_app("test-gem/test/test_test/gem.rb")).to exist
```
arthurnn: this test illustrate this PR!
`bundle gem test-gem` would generete:
before: `test-gem/test/test_test/gem.rb`
after: `test-gem/test/test/gem_test.rb`

indirect:
Yep, seems legit. Thanks!

yuriihabrusiev:
:+1:

PR #3515, opened 2015-03-26
Add test files to default rake test task

arthurnn:
After this PR, we can run `bundle gem foo; cd foo; bundle exec rake` and it should run all the tests.

(this needs to go after https://github.com/bundler/bundler/pull/3513)

review @indirect @andremedeiros

### DIFF
--- a/lib/bundler/templates/newgem/Rakefile.tt
+++ b/lib/bundler/templates/newgem/Rakefile.tt
@@ -4,6 +4,8 @@ require "rake/testtask"
 
 Rake::TestTask.new(:test) do |t|
   t.libs << "test"
+  t.libs << "lib"
+  t.test_files = FileList['test/**/*_test.rb']
 end
 
 task :default => :test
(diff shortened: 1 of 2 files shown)


lib/bundler/templates/newgem/Rakefile.tt
```diff
@@ -4,6 +4,8 @@ require "rake/testtask"
 
 Rake::TestTask.new(:test) do |t|
   t.libs << "test"
```
indirect: Shouldn't we remove "test" from libs now? Or am I misunderstanding TestTask?

arthurnn: `libs` is like `ruby -I`.
So we need to include the lib folder, and the test.
`test` is necessary because of `require test_helper`

indirect: Got it. Thanks!

PR #3518, opened 2015-03-26
[Fetcher] Refactor all the things

segiddins:
This will make adding new index support _really_ easy.

\c @indirect


indirect:
❤️💚💜💙💛

segiddins:
@indirect this is green now!

indirect:
:dancers: :dancers: :dancers:

lib/bundler/fetcher.rb
```diff
@@ -53,6 +56,10 @@ def initialize(remote_uri)
     class << self
       attr_accessor :disable_endpoint, :api_timeout, :redirect_limit, :max_retries
 
+      Fetcher.redirect_limit = 5  # How many redirects to allow in one request
```
indirect: Pretty sure this can just be `self.` instead of `Fetcher.`. Also, lolololololo that these were class-lever attr_accessors that we didn't use.

lib/bundler/fetcher.rb
```diff
@@ -154,37 +130,26 @@ def fetch_spec(spec)
       elsif cached_spec_path = gemspec_cached_path(spec_file_name)
         Bundler.load_gemspec(cached_spec_path)
       else
-        Bundler.load_marshal Gem.inflate(fetch(uri))
+        Bundler.load_marshal Gem.inflate(fetchers.first.fetch uri)
```
indirect: Seems like maybe `fetch` should be a class method if it literally doesn't matter which fetcher instance it uses?

lib/bundler/fetcher.rb
```diff
@@ -200,183 +165,76 @@ def specs(gem_names, source)
       end
 
       index
-    rescue CertificateFailureError => e
+    rescue CertificateFailureError
       Bundler.ui.info "" if gem_names && use_api # newline after dots
-      raise e
+      raise
```
indirect: :+1:

lib/bundler/source/rubygems.rb
```diff
@@ -293,7 +293,7 @@ def cached_specs
       end
 
       def api_fetchers
-        fetchers.select{|f| f.use_api }
+        fetchers.select(&:use_api)
```
indirect: O_O I thought this broke on 1.8.7? You are magic :raised_hands:

lib/bundler/fetcher/dependency_fetcher.rb
```diff
@@ -0,0 +1,75 @@
+require 'bundler/fetcher/fetcher_impl'
+
+module Bundler
+  class Fetcher
+    class DependencyFetcher < FetcherImpl
```
indirect: This is the level of true nits, but I think the constant names would be slightly improved by `Fetcher::FetcherImpl` → `Fetcher::Base` or `Fetcher::Shared` and `DependencyFetcher` → `Fetcher::Dependency`.

indirect:
I left the nittiest picks I had. It's good and you should feel good. :ship:

PR #3519, opened 2015-03-27
[Installer] Add option to force installation

segiddins:
Closes https://trello.com/c/baZN0BmP/78-add-bundle-install-force-option.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -134,6 +134,8 @@ module Bundler
       "Do not attempt to fetch gems remotely and use the gem cache instead"
     method_option "no-cache", :type => :boolean, :banner =>
       "Don't update the existing gem cache."
+    method_option "force", :type => :boolean, :banner =>
+      "Force downloading every gem."
     method_option "no-prune", :type => :boolean, :banner =>
       "Don't remove stale gems from the cache."
     method_option "path", :type => :string, :banner =>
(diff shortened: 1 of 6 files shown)


indirect:
😮😬 you are a machine!

segiddins:
Maybe this would be cleaner if implemented as a setting?

indirect:
Nah, I'm pretty sure this is effectively "bundle pristine"... actually wait hmm. Can we make this "bundle pristine --download" instead, and have pristine without download just install without downloading? That would close like 3 tickets at once.

indirect:
@segiddins, any comment on the `pristine` challenge? Should I just create a Trello card for that and link to this pull as the starting point?

segiddins:
Just haven't gotten around to it yet. First I need to figure out exactly what the scope is!

-Samuel E. Giddins

On Mar 31, 2015, at 6:08 PM, André Arko <notifications@github.com> wrote:

@segiddins, any comment on the pristine challenge? Should I just create a Trello card for that and link to this pull as the starting point?

segiddins:
@indirect if you wrote up exactly what `bundle pristine` should be, I might be able to make it happen :)

indirect:
The `bundle pristine` command would basically be `bundle exec gem pristine`. That re-installs the gem from the tarball that's kept in $GEM_PATH/cache. Bundler's pristine would also need to do a `git reset --hard HEAD && git clean -fd` for each git gem. I guess the download flag would force gems to be downloaded even if they're already in the local cache.

I am really excited about adding checksum support so we won't need that. :P

indirect:
I give up, this is not the same as pristine. Let's just merge.

segiddins:
@indirect sorry about this not being `pristine` :stuck_out_tongue_closed_eyes:

PR #3521, opened 2015-03-27
Make `outdated` output more "grepable"

smlance:
Addresses https://github.com/bundler/bundler-features/issues/63. I know the spec is a bit sparse, so let me know if you have suggestions for more specs or parts of specs.


segiddins:
Let's try and use some more semantic variable names?

smlance:
`fmt_groups` -> `groups`? I'm unsure what to do with `pl`.

smlance:
The failing spec is at spec/realworld/edgecases_spec.rb:57... I don't think this has to do with this PR.

Issue #3522, opened 2015-03-27
Bundler sometimes spams one gem in Gemfile.lock


stalker37:
Bug #3216  return.

Version 1.9.1.
bug is alive
rvm use 2.1.5 do bundle install --path ../../shared/gems

Fetching gem metadata from http://rubygems.org/.........
Fetching version metadata from http://rubygems.org/..
Resolving dependencies....
Installing rake 10.4.2
Installing multi_json 1.11.0
..
Installing ansi 1.5.0
Using bundler 1.9.1

in Gemfile.lock i see

treetop (1.4.15)
polyglot
polyglot
polyglot
..
polyglot (>= 0.3.1)
polyglot (>= 0.3.1)
polyglot (>= 0.3.1)

segiddins:
Can confirm. I'll look into fixing it this weekend, thanks for the bug report @stalker37

arthurnn:
@segiddins the `treetop` gem has bad data in our dependencies db see  http://api.rubygems.org/api/v1/dependencies.json?gems=treetop

indirect:
Going to release 1.9.2 with the fix for this shortly.

PR #3524, opened 2015-03-27
[LazySpecification] Uniq dependencies in lockfile representation

segiddins:
Closes https://github.com/bundler/bundler/issues/3522, but I'm not convinced this is the right thing to do.

### DIFF
--- a/lib/bundler/lazy_specification.rb
+++ b/lib/bundler/lazy_specification.rb
@@ -41,7 +41,7 @@ module Bundler
         out = "    #{name} (#{version}-#{platform})\n"
       end
 
-      dependencies.sort_by {|d| d.to_s }.each do |dep|
+      dependencies.sort_by {|d| d.to_s }.uniq.each do |dep|
         next if dep.type == :development
         out << "    #{dep.to_lock}\n"
       end


indirect:
I mean, it's illegal to have multiple dependencies on the same gem now, but we'll still have to handle this one version of treetop for a long time. Not sure if there is a "right thing" beyond deleting the dupes :P

indirect:
Also it's failing on Ruby 1.8, could you take a look?

TimMoore:
It's a problem with the gemspec in an older version of the gem, so there's not really a way to fix it on the server.

arthurnn:
https://rubygems.org/gems/treetop/versions/1.4.10 . We could ask the author to yank that version..

segiddins:
@arthurnn we shouldn't make them yank the version just because the tooling on our end is faulty, especially when it's in our power to correct the  'issue'

indirect:
Yanking that version doesn’t fix the problem. The gem file, and the gemspec inside it, are on thousands of servers and machines all over the world. That version is also locked in thousands of Gemfile.locks. Yanking would just break `bundle install` for all those people, and the problem would still happen every time they run `bundle install` on a machine that already has it installed.

segiddins:
These spec failures on 1.8 are all totally unrelated

indirect:
I fixed the specs! (Also, @segiddins fixed the specs.) Merging.

PR #3525, opened 2015-03-27
[Molinillo] Make warning-clean

chastell:
This makes `Bundler::Molinillo::Resolver::Resolution` warning-clean, in turns making Bundler warning-lcean.

I [opened this in Molinillo](https://github.com/CocoaPods/Molinillo/pull/21) as well, but it seems Bundler’s version is vendored and namespaced, so I wasn’t sure whether this should or should not be opened here as well.

### DIFF
--- a/lib/bundler/vendor/molinillo/lib/molinillo/resolution.rb
+++ b/lib/bundler/vendor/molinillo/lib/molinillo/resolution.rb
@@ -73,6 +73,19 @@ module Bundler::Molinillo
         end_resolution
       end
 
+      # @return [Integer] the number of resolver iterations in between calls to
+      #   {#resolver_ui}'s {UI#indicate_progress} method
+      attr_accessor :iteration_rate
+      private :iteration_rate
+
+      # @return [Time] the time at which resolution began
+      attr_accessor :started_at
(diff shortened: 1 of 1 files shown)


indirect:
I suspect that Molinillo will not make this change, but we do want Bundler to be warning-clean. Thanks!

segiddins:
@indirect please don't make downstream changes to Molinillo

indirect:
Bundler has a warning-clean policy, because it's infrastructure. The warnings are silly, but people need to be able to find warnings in their own code, without seeing any from using Bundler.

segiddins:
I understand that, but these changes belong upstream, to be released, and then pulled down as a released version in bundler. 

-Samuel E. Giddins

On Mar 27, 2015, at 5:42 PM, André Arko <notifications@github.com> wrote:

Bundler has a warning-clean policy, because it's infrastructure. The warnings are silly, but people need to be able to find warnings in their own code, without seeing any from using Bundler.

indirect:
Oh, alright. Sounds good. 👍

segiddins:
These changes made it properly into 1-9-stable via Molinillo 0.2.3

PR #3526, opened 2015-03-28
Clean up parallel installer

pducks32:
So I was going through the Installer code and thought it was confusing and hard to read (and I guess CodeClimate agreed) and I thought that the parallel code was especially complex. So to get the class more closely aligned with SOLID and Sandi Metz's principles I ripped out that part into a new section.

Parallel code is very hard to test and I'm not always sure about whether things work on older versions of Ruby so i'm just as interested as anyone on how the tests turn out. I'm going to keep researching the best way to test parallelization but for right now I just left it.

### DIFF
--- a/lib/bundler/installer.rb
+++ b/lib/bundler/installer.rb
@@ -269,7 +269,7 @@ module Bundler
     end
 
     def install_in_parallel(size, standalone)
-      ParallelInstaller.call(size, standalone)
+      ParallelInstaller.call(self, specs, size, standalone)
     end
 
 
--- a/lib/bundler/installer.rb
+++ b/lib/bundler/installer.rb
@@ -1,6 +1,7 @@
 require 'erb'
 require 'rubygems/dependency_installer'
 require 'bundler/worker'
+require 'bundler/installer/parallel_installer'
 
 module Bundler
   class Installer < Environment
--- a/lib/bundler/installer.rb
+++ b/lib/bundler/installer.rb
@@ -1,6 +1,5 @@
 require 'erb'
 require 'rubygems/dependency_installer'
-require 'bundler/worker'
 require 'bundler/installer/parallel_installer'
 
 module Bundler
@@ -86,6 +85,7 @@ module Bundler
       # installation is just SO MUCH FASTER. so we let people opt in.
       jobs = [Bundler.settings[:jobs].to_i-1, 1].max
       if jobs > 1 && can_install_in_parallel?
+        require 'bundler/installer/parallel_installer'
         install_in_parallel jobs, options[:standalone]
       else
         install_sequentially options[:standalone]
--- a/lib/bundler/installer.rb
+++ b/lib/bundler/installer.rb
@@ -269,48 +269,7 @@ module Bundler
     end
 
     def install_in_parallel(size, standalone)
-      name2spec = {}
-      remains = {}
-      enqueued = {}
-      specs.each do |spec|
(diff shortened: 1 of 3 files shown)


pducks32:
What does:
```
"/home/travis/build/bundler/bundler/lib/bundler/installer/parallel_installer.rb:4: warning: loading in progress, circular require considered harmful
```

and how do I fix it?

indirect:
It means you're requiring a file (say installer.rb) that requires a file (say parallel_installer.rb) that requires the original file (in this case installer.rb). It usually means that you need to pull out the actual base into a different file that's required by both files.

lib/bundler/installer.rb
```diff
@@ -1,6 +1,6 @@
 require 'erb'
 require 'rubygems/dependency_installer'
-require 'bundler/worker'
+require 'bundler/installer/parallel_installer'
```
indirect: You can drop this line by adding an `autoload` below.

lib/bundler/installer.rb
```diff
@@ -1,6 +1,6 @@
...
 
 module Bundler
   class Installer < Environment
```
indirect: Put the autoload at the top, here: `autoload :ParallelInstaller, 'bundler/installer/parallel_installer'`

lib/bundler/installer.rb
```diff
@@ -85,6 +85,7 @@ def run(options)
       # installation is just SO MUCH FASTER. so we let people opt in.
       jobs = [Bundler.settings[:jobs].to_i-1, 1].max
       if jobs > 1 && can_install_in_parallel?
+        require 'bundler/installer/parallel_installer'
```
indirect: With the autoload, you can remove this, too.

lib/bundler/installer/parallel_installer.rb
```diff
@@ -0,0 +1,94 @@
...
+      end
+    end
+  end
+
+  def self.call(*options)
+    new(*options).call
+  end
+
+  def self.max_threads
+    @@max_threads ||= [Bundler.settings[:jobs].to_i-1, 1].max
```
indirect: Class variables aren't thread-safe, so this makes me nervous. Will this ever be read from inside a thread? Can it be be an ivar set during `initialize` instead?

spec/realworld/parallel_spec.rb
```diff
@@ -6,7 +6,7 @@
       source "https://rubygems.org"
       gem 'activesupport', '~> 3.2.13'
       gem 'faker', '~> 1.1.2'
-      gem 'i18n', '~> 0.6.0' # Because 1.7+ requires Ruby 1.9.3+
+      gem 'i18n', '~> 0.7.0' # Because 1.7+ requires Ruby 1.9.3+
```
indirect: If you have to change the tests for your refactor to pass the tests, I think that means the refactor doesn't work. :) Please make sure these tests pass without changing them!

indirect:
I really like this! :+1: If you can address the things I commented inline, and then squash and rebase this, I'd be really happy to merge it. :ship:

pducks32:
Awesome will do ASAP.

indirect:
@pducks32 any progress on this?

pducks32:
Yea I was working on it a little while ago. Getting there.

-Patrick Metcalfe

indirect:
👍

indirect:
Closing this since there's nothing we can merge yet, but feel free to reopen once it's ready!

PR #3527, opened 2015-03-28
[Gem] Fix making executables executable on 1.8

segiddins:
\c @indirect


indirect:
lol I wrote this code on the plane at the same time

indirect:
Since my commit uses the `Pathname` instance rather than `to_s`ing it, I'll keep mine. Thanks for fixing this though!

PR #3530, opened 2015-03-29
Add Gemfile deprecation warnings [WIP]

smlance:
We still need to ensure deprecation warnings are printed only once per bundler execution.

### DIFF
--- a/spec/bundler/bundler_spec.rb
+++ b/spec/bundler/bundler_spec.rb
@@ -3,6 +3,41 @@ require 'spec_helper'
 require 'bundler'
 
 describe Bundler do
+  describe "version 1.99" do
+    context "when bundle is run" do
+      it "should print a single deprecation warning" do
+        # install_gemfile calls `bundle :install, opts`
+        install_gemfile <<-G
+          source "file://#{gem_repo1}"
+          gem "rack"
+        G
+
+        expect(err).to lack_errors
+      end
+    end
+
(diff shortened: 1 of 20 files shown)


spec/bundler/bundler_spec.rb
```diff
@@ -3,6 +3,41 @@
...
+      it "should print a single deprecation warning" do
+        install_gemfile <<-G
+          source "file://#{gem_repo1}"
+          gem "rack", :group => :test
+        G
+
+        ruby <<-RUBY
+          require 'rubygems'
+          require 'bundler'
+          Bundler.setup
```
indirect: To verify the behavior we're looking for, we should have two lines of `Bundler.setup` here.

indirect:
Looking good! I'm going to see if we can merge the STDOUT pull request into master within the next day or two, and then we can add a `Bundler.ui.deprecate` method that delegates to `Bundler.ui.warn`.

smlance:
Closing in favor of https://github.com/bundler/bundler/pull/3535.

smlance:
Closing in favor of https://github.com/bundler/bundler/pull/3535.

indirect:
@smlance hey, can you rebase this against the current 1.99 branch and squash it down to one commit? We do still need to warn that `Gemfile` is deprecated and will be going away in 2.0.

smlance:
@indirect: Here you go: https://github.com/bundler/bundler/pull/3569

PR #3531, opened 2015-03-30
 Implement optional groups

jhass:
The lack of discussion in https://github.com/bundler/bundler-features/issues/68 and https://github.com/bundler/bundler-features/issues/59 is depressing, so here's some meat to get things going, hopefully.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -151,6 +151,8 @@ module Bundler
         Bundler.rubygems.security_policy_keys.join('|')
     method_option "without", :type => :array, :banner =>
       "Exclude gems that are part of the specified named group."
+    method_option "with", :type => :array, :banner =>
+      "Include gems that are part of the specified named group."
 
     def install
       require 'bundler/cli/install'
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -10,9 +10,34 @@ module Bundler
 
       warn_if_root
 
-      if options[:without]
-        options[:without] = options[:without].map{|g| g.tr(' ', ':') }
+      [:with, :without].each do |option|
+        if options[option]
+          options[option] = options[option].join(":").tr(" ", ":").split(":")
+        end
       end
+
+      if options[:without] && options[:with]
+        conflicting_groups = options[:without] & options[:with]
+        unless conflicting_groups.empty?
+          Bundler.ui.error "You can't list a group in both, --with and --without." \
+          "The offending groups are: #{conflicting_groups.join(", ")}."
+          exit 1
+        end
+      end
+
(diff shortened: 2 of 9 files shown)


lib/bundler/cli.rb
```diff
@@ -151,6 +151,8 @@ def check
         Bundler.rubygems.security_policy_keys.join('|')
     method_option "without", :type => :array, :banner =>
       "Exclude gems that are part of the specified named group."
+    method_option "with", :type => :array, :banner =>
+      "Include gems that are part of the specified named group."
```
indirect: We should probably say "specified optional group" here, to make it clear that `with` only applies to optional groups.

indirect:
This looks like the most reasonable implementation of `with` that we've seen so far. :) I'm not sure how I feel about `group :foo, :optional => true` vs `optional_group :foo`.

How do you feel about this change, @bundler/core?

lib/bundler/cli.rb
jhass: I thought about that but actually it does not only apply to optional groups. Applying it to a normal group has no effect but also doesn't throw an error, however applying it to a normal group that was previously excluded with `without`, removes it from the excluded groups.

indirect: Yeah, that makes sense. I read the rest of the diff after I made that comment. :)

jhass: I did most reasoning in https://github.com/bundler/bundler-features/issues/68, which is why I spared it in the PRs description :)

TimMoore:
:+1: thanks @jhass!

jhass:
Alright, anything else you need from my side?

indirect:
Don't think so! Was just waiting for consensus from the core team and haven't had time to circle back around to this pull yet. Thanks!

jhass:
<3 Thanks!

I know, annoying question, but any rough ETA on when I can see this in a release?

indirect:
I'd like to publish a prerelease within the next week, and then the final will be a week or two after all the known bugs are worked out.

jhass:
Awesome news, thanks!

rymohr:
The advantage of the `--only` approach is that it guarantees the listed groups are the only groups that will be installed.  Ever.  If you want to include new groups in your build you have to do so explicitly within your build process.

The `:optional => true` approach doesn't have the same purity.  When you're adding new gems to a project you're usually not thinking about the obscure conditions of your build process. You're thinking about development, staging, and production.  You're not thinking about the enterprise build you only run a few times a year that runs on a completely different platform.  What happens if you add a group and forget to make it optional?  You break the build.

The shift in control is subtle but I think the `bundle install` caller should have the ultimate say on what gets installed, instead of relying on flags within the Gemfile.

salimane:
@rymohr +1

PR #3533, opened 2015-03-30
more secure suggestion for spec.metadata['allowed_push_host']

seamusabshere:
if you use the option but fail to require rubygems >=2.2, then accidental pushes are **still possible** with old versions that don't respect it

### DIFF
--- a/lib/bundler/templates/newgem/newgem.gemspec.tt
+++ b/lib/bundler/templates/newgem/newgem.gemspec.tt
@@ -24,9 +24,11 @@ Gem::Specification.new do |spec|
   spec.extensions    = ["ext/<%=config[:underscored_name]%>/extconf.rb"]
 <%- end -%>
 
-  if spec.respond_to?(:metadata)
-    spec.metadata['allowed_push_host'] = "TODO: Set to 'http://mygemserver.com' to prevent pushes to rubygems.org, or delete to allow pushes to any server."
-  end
+  # If this is a private gem, you can prevent accidental pushes to rubygems.org with this section (otherwise delete it)
+  # Note: "required_rubygems_version" is important because "allowed_push_host" won't be respected in earlier versions.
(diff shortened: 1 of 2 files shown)


seamusabshere:
@indirect i removed the tests because i turned the suggestion into a comment only

indirect:
Commenting this out means that by default, even Rubygems >=2.2 will not prevent pushes to Rubygems.org. That seems worse than what we have now. :/

seamusabshere:
@indirect actually 80% of the time you don't want this feature - most people make public gems

and the 20% of the time you want a private gem, it's quite insecure to suggest setting `metadata['allowed_push_host']` inside of a `if spec.respond_to?(:metadata)`... as i say in the comments, running `gem push my_private_gem` will work 100% if you have less than rubygems < 2.2

indirect:
I have no idea how you found those 80%/20% numbers, but I'd be very interested to see where you got them.

When a gem _is_ supposed to be private, the current gemspec stops it from being pushed to RubyGems.org 100% of the time for users running RubyGems 2.2 or newer. Your pull request blocks pushes to RubyGems.org 0% of the time, for users on any version of RubyGems. I'd rather keep partial protection that's on by default than get perfect protection that's off by default.

seamusabshere:
@indirect manual intervention is required 100% of the time to make this effective security

1. currently: you have to (a) add your private push server and (b) remove the `if if spec.respond_to?(:metadata)`
2. my version: you have to (a) uncomment and (b) add your private push server

the reason for this pull request is people may do (1)(a) and think that they are protected, but they are not.

if (2)(b) is somehow unnecessary or you think people will do it 100% of the time, please let me know.

seamusabshere:
@indirect would it make more sense if i just removed the `if spec.respond_to?(:metadata)` part? that's really what i don't like

seamusabshere:
@indirect perfect, thanks!

PR #3535, opened 2015-03-31
Bundler.ui.deprecate [WIP]

smlance:


smlance:
Too many commits. I suppose I should have waited until https://github.com/bundler/bundler/pull/3530 is merged.

segiddins:
Hm, you can't really say that something _will_ be deprecated.

lib/bundler/ui/shell.rb
```diff
@@ -11,6 +11,7 @@ def initialize(options = {})
         end
         @shell = Thor::Base.shell.new
         @level = ENV['DEBUG'] ? "debug" : "info"
+        @deprecation_messages = Hash.new
```
segiddins: This looks like it should be a `Set`?

indirect: Yeah, I think it probably should be. `Set` is [slower than `Hash`](http://blog.x-aeon.com/2012/02/07/using-data-sets-in-ruby-performance-of-hash-vs-set/), but that really shouldn't matter here.

indirect:
Can you rebase this against the current 1-99-dev and squash it down to one commit? I'm ready to merge if the specs pass.

indirect:
Cherry-picked as ff1d598ce3e2fae5bce450be9470f429e154c9ed. Thanks!

PR #3536, opened 2015-03-31
Bindir is bin

matiasow:
When 'bundle gem' is invoked, /bin dir is created for gem executable scripts. Then spec.bindir should be set to "bin" instead of "exe".


indirect:
The bindir is not actually `bin`, it is `exe`. There's more explanation in [this blog post](http://bundler.io/blog/2015/03/20/moving-bins-to-exe.html).

PR #3542, opened 2015-04-02
Allow pushing to custom gem servers

mlarraz:
Although there's already several ways to do this without Bundler, it would be convenient to be able to do so without having to deal directly with RubyGems configuration


indirect:
Why not read this directly out of the gemspec's allowed_push_host?

indirect:
Closing due to lack of activity...

PR #3551, opened 2015-04-08
Deprecate and remove capistrano.rb

smlance:


indirect:
We should remove `lib/bundler/deployment.rb` and `lib/bundler/vlad.rb`, too!

smlance:
Done!

Issue #3066, opened 2014-06-14
Universal gems do not match x64-mingw32 platform


joshcooper:
I have a Gemfile:

<pre>
source "https://rubygems.org"
gem 'windows-api'
</pre>

Currently, `windows-api` is at version [`0.4.2`](https://rubygems.org/gems/windows-api/versions/0.4.2). It has a dependency on [`win32-api`](https://github.com/djberg96/windows-api/blob/windows-api-0.4.2/windows-api.gemspec#L17).

In ruby 1.9.3p374, bundler successfully resolves the most recent version of the gem, which happens to be [`win32-api-1.5.1-universal-mingw32`](https://rubygems.org/gems/win32-api/versions/1.5.1-universal-mingw32), containing precompiled binaries for both x86 and x64. 

<pre>
C:\work\foo>bundle install --path .bundle
...
Activated:
  windows-api (0.4.2)
Requirements:
  win32-api (>= 1.4.5) x86-mingw32
Attempting:
  win32-api (>= 1.4.5) x86-mingw32
  Activating: win32-api (1.5.1)
    * windows-api (>= 0)
    * win32-api (>= 1.4.5)
    Dependencies

Installing win32-api 1.5.1
Installing windows-api 0.4.2
Using bundler 1.6.2
</pre>

However, on ruby 2.0.0p481 x64, the universal gem does not match, and bundler falls back to [`win32-api-1.4.8`](https://rubygems.org/gems/win32-api/versions/1.4.8), which is a `ruby` version that requires the DevKit to be installed to compile the native extension.

<pre>
C:\work\foo>ruby --version
ruby 2.0.0p481 (2014-05-08) [x64-mingw32]
C:\work\foo>gem --version
2.0.14
C:\work\foo>bundle --version
Bundler version 1.6.2
C:\work\foo>bundle install --path .bundle
...
Activated:
  windows-api (0.4.2)
Requirements:
  win32-api (>= 1.4.5) x64-mingw32
Attempting:
  win32-api (>= 1.4.5) x64-mingw32
  Activating: win32-api (1.4.8)
    * windows-api (>= 0)
    * win32-api (>= 1.4.5)
    Dependencies

Gem::InstallError: The 'win32-api' native gem requires installed build tools.
</pre>

However, `gem install windows-api` on ruby 2.0.0 x64 correctly pulls in the `win32-api-1.5.1-universal-mingw32` gem as expected:

<pre>
C:\work\foo>ruby --version
[...]

PR #3553, opened 2015-04-09
Fix universal gem installation issues

jdmundrawala:
Trying to install `win32-api` on windows 64-bit ruby since it is published as a `universal-mingw32` gem instead of `x64-mingw32`.
This patch should fix https://github.com/bundler/bundler/issues/3066

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ before_script: travis_retry rake spec:travis:deps
 
 branches:
   only:
+    - universal
     - master
     - 1-9-stable
     - 1-8-stable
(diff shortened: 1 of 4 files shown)


indirect:
Thanks for the patch! Can you write a test, or at least describe how to test this, so we can make sure that it's actually fixed?

jdmundrawala:
Happy to write the tests, probably goes in https://github.com/bundler/bundler/blob/master/spec/resolver/platform_spec.rb? 
The case this covers is where the gem has a universal cpu

Any ideas why travis is not running? I wasn't actually able to run any of the specs on windows and was hoping I could see if I broke anything.

indirect:
Travis isn't running because you create a branch whose name isn't in the travis whitelist, I think?

jdmundrawala:
I'm not sure if that's the reason, but let's give it a shot.
Travis should build PRs against branches that are whitelisted, and master is whitelisted.

jdmundrawala:
Looks like it worked

indirect:
awesome, thanks!

indirect:
Cherry picked as bd078f22c3a9720f7d050c73039f53cca8ff90dd and b7d5d902715e266758a32fda849c3590217fcbb0.

PR #3559, opened 2015-04-11
Closes #3558

### LINKED ISSUE #3558: NoMethodError: undefined method `spec' for nil:NilClass
That exception occurs when running `bundle install` on the current `ruby-head` ...
Have found this twice so far, and suspect I'll keep finding it if I keep looking:

* abstractive/rspec-log_split#2
* celluloid/timers#47

digitalextremist:
`ruby-head` seems to be more strict about how instance variables are referenced, at the moment.

### DIFF
--- a/lib/bundler/source/path/installer.rb
+++ b/lib/bundler/source/path/installer.rb
@@ -16,7 +16,7 @@ module Bundler
         end
 
         def generate_bin
-          return if spec.executables.nil? || spec.executables.empty?
+          return if @spec.executables.nil? || @spec.executables.empty?
 
           if Bundler.requires_sudo?
             FileUtils.mkdir_p(@tmp_bin_dir) unless File.exist?(@tmp_bin_dir)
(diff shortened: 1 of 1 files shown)


segiddins:
I believe this is actually caused by https://github.com/rubygems/rubygems/commit/620910aa75c7446e8f8d5af1df5f981e4da86197.

segiddins:
@digitalextremist this PR doesn't seem to fix things, please take a look at the test failures?

digitalextremist:
@segiddins seeing the test failures, but at first glance it seems like it's failing on tests *not* anticipating @tenderlove's changes to rubygems. Looking further.

digitalextremist:
@segiddins the give-away that this is more about how an instance variable is referenced versus whether caching is being done, for me, is that this is `ruby-head` isolated so far.

segiddins:
well, does adding an `attr_accessor :spec` fix things?

digitalextremist:
I pulled my comment from @tenderlove's commit since it was out of the context of that commit, and I'm looking through rubygems right now to attempt putting `@spec` within scope of its own instances which reference it as `spec` .. purely for science, because you asked. As far as I can see, what you're talking about is *another* similar issue.

indirect:
I have basically no idea what is even going on here. What is broken, and under what circumstances? What is this patch supposed to fix, and why does the change make the tests fail?

indirect:
This is fixed by f4481a7e3dbaf7f563e32afbe2cbbb84e3091f16.

Issue #3554, opened 2015-04-09
10.10.3 with XCode 6.3 Command Line Tools?


barts:
Hi, I upgraded to OS 10.10.3 today and XCode 6.3 with the XCode 6.3 Command Line Tools from here:
https://developer.apple.com/downloads/index.action#

When I try to do a `bundle install` I get this error, which is so long GitHub won't let me paste it into this form (and the plain text file for it is 1.1MB large...):
https://www.dropbox.com/s/qotscp9lmkd8r0z/bundler.txt?dl=0

indirect:
It looks like that error doesn't actually include the full text. Perhaps your terminal is running out of backscroll? Can you try running `bundle install > output.txt` and then upload the output.txt file to dropbox or a gist?

TimMoore:
This gist is decoded from the URL parameter in the file linked above: https://gist.github.com/TimMoore/0ed2626e9dba2ae30639

It looks like a problem building the `debugger` gem. @barts does `gem install debugger` work for you?

TimMoore:
Also @barts the `debugger` gem is not compatible with Ruby 2.0+. You might want to try `byebug` instead. See https://github.com/cldwalker/debugger#known-issues

barts:
Installing `byebug` (and `pry-byebug` just in case) seems to have fixed it. Thank you!

TimMoore:
I can't seem to reproduce whatever caused the excessively long error message. I get much more useful output when I try to build a `Gemfile` with the `debugger` gem on a newer Ruby:

```
Fetching gem metadata from https://rubygems.org/....
Fetching version metadata from https://rubygems.org/..
Resolving dependencies...
Installing columnize 0.9.0
Installing debugger-linecache 1.2.0
Installing debugger-ruby_core_source 1.3.8

Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

    /Users/tmoore/.rubies/ruby-2.2.0/bin/ruby -r ./siteconf20150412-15238-1tt0hsp.rb extconf.rb
*** extconf.rb failed ***
Could not create Makefile due to some reason, probably lack of necessary
libraries and/or headers.  Check the mkmf.log file for more details.  You may
need configuration options.

Provided configuration options:
	--with-opt-dir
	--with-opt-include
	--without-opt-include=${opt-dir}/include
	--with-opt-lib
	--without-opt-lib=${opt-dir}/lib
	--with-make-prog
	--without-make-prog
	--srcdir=.
[...]

PR #3562, opened 2015-04-12
Only use first line of error messages in the issue search URL

TimMoore:
This is to prevent super-long error messages like in #3554.


indirect:
👍🚢

PR #3563, opened 2015-04-12
Fixes a build failure introduced in a72df72

TimMoore:
a72df72c7719bcb7677e5cb140cf7cdb11be652d removes the exception that prevents public pushes on RubyGems < 2.0, but one test was restoring the gemspec back to the original contents (https://github.com/bundler/bundler/commit/a72df72c7719bcb7677e5cb140cf7cdb11be652d#diff-7e091e8293b0ca921df62d6995b5b909R137).

Now it restores the TODOs, but not the removed exception.


indirect:
Heh, I was wondering I had had missed one. Thanks!

TimMoore:
Somehow it seems to have broken everything else!

TimMoore:
I amended the commit to fix the problem on newer RG versions.

indirect:
Hmm, looks like I need to backport the fix for rubygems master.

PR #3565, opened 2015-04-13
platform can be of type string so it must be cast into a Gem::Platform

jdmundrawala:
#3553 didn't actually fix anything and made things worse (universal gems no longer work on 32 bit ruby on windows)

It seems that the platform can be of type string, so wrapping in in Gem::Platform is necessary.

sorry for the regression

### DIFF
--- a/lib/bundler/match_platform.rb
+++ b/lib/bundler/match_platform.rb
@@ -7,8 +7,7 @@ module Bundler
     def match_platform(p)
       Gem::Platform::RUBY == platform or
       platform.nil? or p == platform or
-      generic(Gem::Platform.new(platform)) == p or
-      Gem::Platform.new(platform) === p
+      generic(Gem::Platform.new(platform)) === p
     end
   end
 end
--- a/lib/bundler/match_platform.rb
+++ b/lib/bundler/match_platform.rb
@@ -8,7 +8,7 @@ module Bundler
       Gem::Platform::RUBY == platform or
       platform.nil? or p == platform or
       generic(Gem::Platform.new(platform)) == p or
-      platform === p
+      Gem::Platform.new(platform) === p
     end
   end
 end
--- a/spec/other/ext_spec.rb
+++ b/spec/other/ext_spec.rb
@@ -5,6 +5,14 @@ describe "Gem::Specification#match_platform" do
     darwin = gem "lol", "1.0", "platform_specific-1.0-x86-darwin-10"
     expect(darwin.match_platform(pl('java'))).to eq(false)
   end
+
+  context "when platform is a string" do
+    it "matches when platform is a string" do
+      lazy_spec = Bundler::LazySpecification.new("lol", "1.0", "universal-mingw32")
+      expect(lazy_spec.match_platform(pl('x86-mingw32'))).to eq(true)
+      expect(lazy_spec.match_platform(pl('x64-mingw32'))).to eq(true)
+    end
+  end
 end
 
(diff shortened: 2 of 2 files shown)


lib/bundler/match_platform.rb
```diff
@@ -8,7 +8,7 @@ def match_platform(p)
       Gem::Platform::RUBY == platform or
       platform.nil? or p == platform or
       generic(Gem::Platform.new(platform)) == p or
-      platform === p
+      Gem::Platform.new(platform) === p
```
segiddins: Since `Gem::Platform.new(platform)` is done twice, maybe it should be wrapped in a variable?

segiddins:
@jdmundrawala maybe it would just be better to do this in the initializer for `LazySpecification`?

jdmundrawala:
@segiddins I'm a little worried about changing `LazySpecification` and `EnpointSecification` to do that.
A similar thing was done in rubygems, and produced a bug like: https://github.com/rubygems/rubygems/pull/1184

segiddins:
Ah actually you’re right, I made that very mistake this weeked

jdmundrawala:
So reading this again, maybe the `===` comparison is all that is needed and the `==` can be dropped

lib/bundler/match_platform.rb
```diff
@@ -5,10 +5,11 @@ module MatchPlatform
     include GemHelpers
 
     def match_platform(p)
+      generic_platform = generic(Gem::Platform.new(platform))
       Gem::Platform::RUBY == platform or
       platform.nil? or p == platform or
```
indirect: The vast, vast majority of these comparisons are going to exit on the first three conditions, though, so I think it's a performance loss to create a `Platform` before we do the compare... can we switch to just the one `===` and wait to create the `Platform` until then?

jdmundrawala:
Not really certain why the tests keep failing....
```ruby
  it "installs the latest version of gxapi_rails", :ruby => "1.9"  do
    install_gemfile <<-G
    source "https://rubygems.org"

    gem "sass-rails"
    gem "rails", "~> 3"
    gem "gxapi_rails"
    G
    expect(out).to include("gxapi_rails 0.0.6")
  end
```
seems to never return. Tried running locally(on a mac) with rubygems 2.2.3 and ruby 1.9.3 with the following Gemfile
```ruby
source "https://rubygems.org"

gem "sass-rails"
gem "rails", "~> 3"
gem "gxapi_rails"
```
never returns. Tried with bundler 1.9.2 and bundler 1.9.3

never in this case means
```
Resolving dependencies..................................................................................................................................................................................................................................................................................................................................................^C
```

segiddins:
@jdmundrawala that would imply that the resolver is either looping infinitely, or it's missing the match that it needs to terminate in a sane amount of time.

jdmundrawala:
bundle install  171.70s user 1.46s system 97% cpu 2:58.17 total
It just took a while

indirect:
@segiddins that seems like a pretty pathological resolver case, any ideas?

indirect:
(For the record, 1-8-stable takes 19 seconds to finish resolving that Gemfile, while 1-9-stable takes... much longer.)

indirect:
@jdmundrawala Thanks for the patch!

indirect:
Cherry-picked to 1-9-stable as 99a71299d1991e5b31a651e48b31a651d7065510 and de368e1313dcf4411527c9e8654b1b6cfa7833c5.

PR #3566, opened 2015-04-13
fix GemNotFound message in Resolver

### LINKED ISSUE #3564: Beginner-friendly error message for misspelled or non-existent gem names
Let's say I'm new to Ruby, Rails, Bundler, RubyGems, etc.  Clear error messages are a huge help.

Given a Gemfile:
```ruby
source 'https://rubygems.org'
gem 'rails'
gem 'misspelled-gem-name'
```

Bundling will report that the 'misspelled-gem-name' is not available on "this machine":

```console
$ bundle
Could not find gem 'misspelled-gem-name (>= 0) ruby' in the gems available on this machine.
```

I think this message could be improved by adding a line that says the gem could not be found in RubyGems, when the source for a Gemfile includes rubygems.org.  For example:

```console
Could not find gem 'misspelled-gem-name (>= 0) ruby' at https://rubygems.org/.
```

Does anyone else think this is would be a welcome improvement?

For further illustration, this spec fails:
```ruby
  describe "failed bundle install" do
    before :each do
      gemfile <<-G
        gem "rails"
        gem "misspelled-gem-name", :group => :development
      G
    end

    it "should report a helpufl error message" do
      bundle :install
      expect(out).to include("Fetching gem metadata from https://rubygems.org/")
      expect(out).to include("Could not find gem 'misspelled-gem-name (>= 0) ruby' at https://rubygems.org/.")
    end
  end
```

```console
Failure/Error: expect(out).to include("Could not find gem 'misspelled-gem-name (>= 0) ruby' at https://rubygems.org/.")
       expected "Fetching gem metadata from https://rubygems.org/...........\nFetching version metadata from https://rubygems.org/..\nCould not find gem 'misspelled-gem-name (>= 0) ruby' in the gems available on this machine." to include "Could not find gem 'misspelled-gem-name (>= 0) ruby' at https://rubygems.org/."
```

sealocal:
Fixes #3564 (Beginner-friendly error message for misspelled or non-existent gem names).

Feedback from anyone is welcome.

### DIFF
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -323,12 +323,7 @@ module Bundler
               message << "Source does not contain any versions of '#{requirement}'"
             end
           else
-            message = "Could not find gem '#{requirement}' "
-            if Bundler.definition.has_rubygems_remotes?
-              message << "in any of the gem sources listed in your Gemfile or installed on this machine."
-            else
-              message << "in the gems available on this machine."
-            end
+            message = "Could not find gem '#{requirement}' in any of the gem sources listed in your Gemfile or installed on this machine."
           end
           raise GemNotFound, message
         end
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -324,8 +324,8 @@ module Bundler
             end
           else
             message = "Could not find gem '#{requirement}' "
-            if @index.source_types.include?(Bundler::Source::Rubygems)
-              message << "in any of the gem sources listed in your Gemfile."
+            if Bundler.definition.has_rubygems_remotes?
+              message << "in any of the gem sources listed in your Gemfile or installed on this machine."
             else
               message << "in the gems available on this machine."
             end
--- a/spec/install/post_bundle_message_spec.rb
+++ b/spec/install/post_bundle_message_spec.rb
@@ -81,6 +81,22 @@ describe "post bundle message" do
         expect(out).to include(bundle_complete_message)
       end
     end
+
+    describe "with misspelled or non-existent gem name" do
+      before :each do
+        gemfile <<-G
+          source 'https://rubygems.org/'
+          gem "rails"
+          gem "misspelled-gem-name", :group => :development
+        G
+      end
+
+      it "should report a helpufl error message" do
+        bundle :install
+        expect(out).to include("Fetching gem metadata from https://rubygems.org/")
+        expect(out).to include("Could not find gem 'misspelled-gem-name (>= 0) ruby' in any of the gem sources listed in your Gemfile or installed on this machine.")
+      end
+    end
   end
 
   describe "for second bundle install run" do
--- a/lib/bundler/index.rb
+++ b/lib/bundler/index.rb
@@ -81,10 +81,6 @@ module Bundler
       end
     end
 
-    def source_types
-      sources.map{|s| s.class }.uniq
-    end
-
     alias [] search
 
     def <<(spec)


lib/bundler/resolver.rb
```diff
@@ -324,8 +324,8 @@ def verify_gemfile_dependencies_are_found!(requirements)
             end
           else
             message = "Could not find gem '#{requirement}' "
-            if @index.source_types.include?(Bundler::Source::Rubygems)
-              message << "in any of the gem sources listed in your Gemfile."
+            if Bundler.definition.has_rubygems_remotes?
```
segiddins: I'd prefer we didn't reference the `Bundler.definition` when not necessary, as it makes this method reliant on global state

spec/install/post_bundle_message_spec.rb
```diff
@@ -81,6 +81,22 @@
         expect(out).to include(bundle_complete_message)
       end
     end
+
+    describe "with misspelled or non-existent gem name" do
```
segiddins: I'd think these tests should probably go in the resolver spec?

lib/bundler/resolver.rb
```diff
@@ -324,8 +324,8 @@ def verify_gemfile_dependencies_are_found!(requirements)
...
+              message << "in any of the gem sources listed in your Gemfile or installed on this machine."
             else
               message << "in the gems available on this machine."
             end
```
sealocal: `@index.source_types` always returns `[Bundler::Index]`, because a source must be an instance `Index` as seen in `Index#add_source`.  So, an instance of `Bundler::Source::Rubygems)` will never be found.

indirect: that explains that bug! oops :grimacing:

lib/bundler/resolver.rb
sealocal: OK. So, perhaps `Index#source_types` should be re-implemented?

I don't understand how to find the Gemfile's sources from `@index`.  I was thinking `@index.sources.each {|source| source.sources}` might be helpful, but after experimentation any source's source seems to always be empty.

Another simpler solution would be to remove this conditional completely so that the message here is always `in any of the gem sources listed in your Gemfile or installed on this machine.`.

indirect:
It is _extremely_ uncommon to have zero sources, so I am fine with changing the hardcoded message.

spec/install/post_bundle_message_spec.rb
sealocal: The reason I put it in `install/post_bundle_message_spec` is because I really just wanted to spec a helpful error message, not that the Resolver is able to do its job.

I might be able to move it to the resolver spec, but I will have to study `spec/support/indexes` for the example Gemfile that I provided.

indirect: I am :+1: on having one full test for this. It's been broken for many releases, I would like to make sure that doesn't happen again.

sealocal:
@indirect When you say to change the hardcoded message, do you mean that removing the conditional on line 327 of Resolver would be reasonable?

indirect:
yes, I think removing the conditional makes sense.

spec/install/post_bundle_message_spec.rb
sealocal: Bear with me, but I don't know what you mean by "one full test."  Does that mean you'd prefer the test style in the Resolver spec of the "post bundle message" spec?

indirect: Oh, sorry... I am happy to have tests in the resolver, but I think we should have at least this one test that is the full integration style. Please don't remove this test!

sealocal: Great, thanks! I'll see if I can compose a resolver test as well.

indirect:
Cherry-picked into 1-9-stable as dc442f4ff848ed1a550a4437293c14fb71dae9ef and 70262b2c8bf5999bd8a207ebdfecb6d7e58e9f0c. Thanks!

PR #3575, opened 2015-04-17
Allow fuzzy ruby version requirements.

panthomakos:
This is my first PR to this repo. I tried to follow guidelines as best as possible, but I am sure there are plenty of adjustments that need to be made.

* Use `Gem::Requirement` and `Gem::Version` to parse and check
  ruby version, engine_version and patchlevel requirements.
* Added new class `RubyVersionRequirement` to handle this functionality.
* Added tests that cover the `RubyVersionRequirement` class.

I wasn't sure about the `CHANGELOG.md` formatting since I am referencing this issue: https://github.com/bundler/bundler-features/issues/70

### DIFF
--- a/lib/bundler/ruby_version.rb
+++ b/lib/bundler/ruby_version.rb
@@ -38,26 +38,6 @@ module Bundler
         patchlevel     == other.patchlevel
     end
 
-    # Returns a tuple of these things:
-    #   [diff, this, other]
-    #   The priority of attributes are
-    #   1. engine
-    #   2. ruby_version
-    #   3. engine_version
-    def diff(other)
-      if engine != other.engine && @input_engine
-        [ :engine, engine, other.engine ]
-      elsif version != other.version
-        [ :version, version, other.version ]
-      elsif engine_version != other.engine_version && @input_engine
-        [ :engine_version, engine_version, other.engine_version ]
-      elsif patchlevel != other.patchlevel && @patchlevel
-        [ :patchlevel, patchlevel, other.patchlevel ]
-      else
-        nil
-      end
-    end
-
     def host
       @host ||= [
         RbConfig::CONFIG["host_cpu"],
@@ -112,6 +92,34 @@ module Bundler
 
     def patchlevel
       RUBY_PATCHLEVEL.to_s
+    end
+  end
+
+  class RubyVersionRequirement < RubyVersion
+    # Returns a tuple of these things:
+    #   [diff, this, other]
+    #   The priority of attributes are
+    #   1. engine
+    #   2. ruby_version
+    #   3. engine_version
+    def diff(other)
+      if engine != other.engine && @input_engine
+        [ :engine, engine, other.engine ]
+      elsif !version || !matches?(version, other.version)
+        [ :version, version, other.version ]
+      elsif @input_engine && !matches?(engine_version, other.engine_version)
+        [ :engine_version, engine_version, other.engine_version ]
+      elsif @patchlevel && !matches?(patchlevel, other.patchlevel)
+        [ :patchlevel, patchlevel, other.patchlevel ]
+      else
+        nil
+      end
+    end
+
+    private
+
+    def matches?(requirement, version)
+      Gem::Requirement.create(requirement).satisfied_by?(Gem::Version.new(version))
     end
   end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## Next Release
+
+Features:
+
+  - Allow fuzzy ruby version, patchlevel and engine version requirements (#70, @panthomakos)
+
 ## 1.9.4 (2015-04-13)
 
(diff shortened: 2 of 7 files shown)


indirect:
Thanks for the pull request! As you may have noticed, while the Gemfile can contain fuzzy version numbers, the Gemfile.lock must contain specific locked version numbers. In order to make this change to Bundler, we need to start saving the ruby version in the Gemfile.lock, and allow running `bundle update ruby` to update the locked version of Ruby to the current version (as long as the current version meets the fuzzy specifier in the Gemfile). Are you willing to add that to this pull request?

panthomakos:
@indirect 

I don't know if I am ready to take on that level of code change yet, but it also doesn't make sense to me. Perhaps you can provide some more guidance or some additional explanation.

It seems that `Gemfile.lock` is responsible for listing the exact gem versions that Bundler has installed. Since Bundler does not install Ruby, and a Gemfile can support multiple versions/platforms of Ruby, does it not make sense to leave the implementation as is - after all the ruby version check does happen before all bundle commands (exec, install, etc.) so we are guaranteed to be compliant.

A practical use-case that I am using this for is the following. In our Rails application we want to be able to upgrade Ruby versions without tying the Ruby package upgrade to a particular project source control version, but we also want to enforce a specific set of Ruby versions for developers and for production instances. For instance, we might want to run 90% of traffic in Ruby `2.0.0` and 10% in Ruby `2.1.0` for testing purposes. A directive of `ruby '> 2.0.0'` would accomplish this without tying us to a particular Ruby version - since that's not something Bundler can control anyways. Does that make sense?

lib/bundler/ruby_version.rb
```diff
@@ -114,4 +94,32 @@ def patchlevel
       RUBY_PATCHLEVEL.to_s
     end
   end
+
+  class RubyVersionRequirement < RubyVersion
```
andremedeiros: I'm starting to feel that this file should be broken down into several. It's already defining 3 classes :disappointed:

andremedeiros:
@indirect actually I looked it up and there _is_ a gem called `ruby`. Would your example still be valid?

indirect:
@andremedeiros the gem named ruby was reserved as an empty gem specifically so that things like this would never conflict.

@panthomakos you can accomplish that simply by putting `raise("Wrong ruby version!") unless RUBY_VERSION > "2.0.0"` in your Gemfile. I'm not sure why that would need a patch to Bundler. :) The fundamental reason that Bundler exists is to make it possible to run the exact same versions on every machine—development, production, and everywhere else. That's why putting the Ruby version in the lockfile is needed once the Gemfile allows more than one version. In order to guarantee that the same code is running in development and deployment, Bundler has to lock down to a single Ruby version. If a single Ruby version isn't what you want, that's fine, but I don't think that that makes sense to make part of Bundler.

simi:
@andremedeiros I had an similar idea. But `ruby` gem pushed to rubygems.org can be deleted (since it is empty gem) and ruby can be reserved rubygems word.

But there is problem with custom gemservers.

Maybe `bundle update --ruby` can work for this?

lib/bundler/ruby_version.rb
panthomakos: Happy to do that as part of this PR.

panthomakos:
@indirect - okay, that makes sense. I am not sure how to get started on adding something to Gemfile.lock or adding tests to ensure it parses/works properly. If there is incremental progress I can make towards that goal, I would be happy to contribute. Otherwise, I'll close this PR since it's not the right approach. Thanks!

jtarchie:
@indirect and @panthomakos:

This is a feature that we are looking to implement, but did our do diligence to make sure no one hadn't already.

Our use case is specific deployments of applications to a cloud platform -- Cloud Foundry. We make a security guarantee that an application cannot use insecure version of Ruby (and other runtimes). If a user had been developing locally with Ruby 2.2.1 and pushed to a Cloud Foundry, they would get an error that 2.2.1 is unsupported.

We are looking to do a track of work of ensuring that it Just Works™. We want to allow the user to specify in an idiomatic way that they trust Ruby 2.2.x will support their 2.2.1 tested app.

Is the only blocker from this PR being accepted that the version information of Ruby is not in the `Gemfile.lock`?

If so, a proposal for a possible specification. Assuming Ruby 2.2.2 is `ruby --version`.

Gemfile

```ruby
source 'https://rubygems.org'
ruby '~> 2.2.1'
gem 'sinatra'
```

Gemfile.lock

```
GEM
  remote: https://rubygems.org/
  specs:
    rack (1.5.2)
    rack-protection (1.5.2)
      rack
    sinatra (1.4.4)
      rack (~> 1.4)
      rack-protection (~> 1.4)
      tilt (~> 1.3, >= 1.3.4)
    tilt (1.4.1)

PLATFORMS
  ruby (2.2.2)

DEPENDENCIES
  sinatra
```

What do you think?

PR #3578, opened 2015-04-20
Fix boolean `bundle gem` settings

### LINKED ISSUE #3443: [bundle gem] Replace interactive prompts with command options
Interactive prompts are annoying and disrupt workflow. They should be replaced with command-line options with appropriate default values. Consider adding a `--[no-]-code-of-conduct` option.

pducks32:
Fixes #3443 (nmarley's problem at the bottom)
The issue was that `alpha =~ beta` returns nil if `alpha` doesn't contain `beta` and the position—e.g. 6—instead of `false` and `true` respectively like was assumed by whomever wrote the original logic. It coincidentely worked for yes because everything—like 6 using my above example—except for `false` and `nil` resolve to `true`. However, when they selected no for coc or mit they were always reasked because it resolved to `nil` which for boolean settings doesn't get transformed into `false`.

### DIFF
--- a/lib/bundler/cli/gem.rb
+++ b/lib/bundler/cli/gem.rb
@@ -139,7 +139,7 @@ module Bundler
 
       if choice.nil?
         Bundler.ui.confirm header
-        choice = (Bundler.ui.ask("#{message} y/(n):") =~ /y|yes/)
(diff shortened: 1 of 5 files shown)


pducks32:
Okay messed up this PR will make a new one...one of these days I'll finally understand git perfectly!!

PR #3580, opened 2015-04-20
Use specification stubs with RubyGems 2.1+ when constructing the local i...

segiddins:
...ndex

\c @indirect @tenderlove


segiddins:
(This could probably do with some profiling on a RubyGems install that has lots of gems to see if it's really an improvement)

PR #3582, opened 2015-04-20
Fix rubygems 2.2+ compatibility

voxik:
This is follow up of #3237 including the fix to the last test failure on Ruby 1.8.7 and RubyGems 2.2.2.

### DIFF
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -130,6 +130,19 @@ module Bundler
       yield
     end
 
+    def loaded_gem_paths
+      # RubyGems 2.2+ can put binary extension into dedicated folders,
+      # therefore use RubyGems facilities to obtain their load paths.
+      if Gem::Specification.method_defined? :full_require_paths
+        loaded_gem_paths = Gem.loaded_specs.map {|n, s| s.full_require_paths}
+        loaded_gem_paths.flatten!
+      else
+        $LOAD_PATH.select do |p|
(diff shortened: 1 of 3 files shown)


indirect:
Maybe I'm slow, but I'm still not sure what this is fixing. Can you explain why this is needed, now that RubyGems correctly loads .rb and .so files again?

voxik:
The original explanation is in #3234:

> RubyGems allows since version 2.2 to reconfigure place, where binary
> extensions are stored. This might allow easier sharing of gems between
> various Ruby interpreters. It can be enabled by redefining
> Gem.default_ext_dir_for method.

And we are using this feature in Fedora [1]. The attached test case simulates this by the ```s.extension_dir = '#{ext_dir}'``` line.

To be precise, it is not about loading, the .rb and .so files are correctly loaded, but the $LOAD_PATH might not be cleaned correctly. This patch set narrows it by using RubyGems facilities.

http://pkgs.fedoraproject.org/cgit/ruby.git/tree/operating_system.rb#n103

voxik:
Just to give you example, you probably assume that this is the directory structure of RubyGems:

~~~
/usr/local/ruby/2.2.1/extensions
/usr/local/ruby/2.2.1/gems
~~~

while it can be

~~~
/usr/lib64/gems/ruby/
/usr/ruby/gems/gems/
~~~

i.e. they don't have common prefix as the current code expects.

indirect:
Got it. I understand now, thank you for explaining. And thank you for the patch!

voxik:
Thanks!

PR #3586, opened 2015-04-23
Configurable timeout/retries/redirects

### LINKED ISSUE #3583: make bundler connection options configurable
In environments where the remote server serving gems is slow to respond and or the work performed by the remote server to generate the response is significant ( long list of gems ) [the hardcoded connection timeout of 10 seconds](https://github.com/bundler/bundler/blob/master/lib/bundler/fetcher.rb#L61:L63) may not be enough to complete the request. Bundler will reset it's connection after 10 seconds and try again 3 times.

It would be helpful to provide a configuration option to adjust the connection timeout option so that bundler users can work around slowly performing hosts.

Since the logic and code touched would be near identical, consider making retries and redirect limit configurable as well, although less important for the purposes of this report.

These settings are typically more important in corporate environments where requests are routed through all manner of proxies and firewalls.

pducks32:
Fixes #3583
Works just like you’d expect it to like all settings and is easily config’d with `bundle config`

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -20,7 +20,7 @@ module Bundler
       current_cmd = args.last[:current_command].name
       custom_gemfile = options[:gemfile] || Bundler.settings[:gemfile]
       ENV['BUNDLE_GEMFILE']   = File.expand_path(custom_gemfile) if custom_gemfile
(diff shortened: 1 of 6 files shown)


indirect:
Looks good, thanks!

PR #3590, opened 2015-04-24
Edit bundle-config.ronn and bundle-install.ronn re: previously remembered flags

smlance:


indirect:
Oh wow, I completely forgot—we want to remove the `--binstubs` option entirely in 2.0. It conflicts with any application that has its own binstubs, like Rails apps, and even the current `gem` command, which creates `bin/setup` and `bin/console`. Instead, we suggest that people use the `bundle binstubs` command to generate the binstubs that they specifically want to use, and then check those in to source control. I'll add a Trello card about removing it.

indirect:
This is exactly what I had in mind, thanks!

PR #3592, opened 2015-04-25
Replace locked gem sources with Gemfile equivalents.

### LINKED ISSUE #3585: Gem found in multiple sources
I had a similar problem as reported in #3378, and it appears to come up when a gem from the primary source depends on a gem from a custom source. For example, `stripe-ruby-mock` depends on `stripe`, and even when the `source` is specified for the `stripe` gem, Bundler is giving the "found in multiple sources" error.

Here's a quick demo to reproduce this. Note: if you have the `stripe` gem installed, you'll have to uninstall it in order to get the error message.

```bash
# if necessary, uninstall stripe gem
gem uninstall stripe

mkdir bundler-test && cd bundler-test
echo "source 'https://rubygems.org'"                   >> Gemfile
echo "gem 'json'"                                      >> Gemfile
echo "gem 'stripe', source: 'https://code.stripe.com'" >> Gemfile
echo "gem 'stripe-ruby-mock'"                          >> Gemfile
bundle install
```

This produces the following output:

```
Fetching source index from https://code.stripe.com/
Fetching version metadata from https://rubygems.org/.....
Fetching source index from https://code.stripe.com/
Using blankslate 3.1.3
Using dante 0.2.0
Using unf_ext 0.0.7.1
Using unf 0.1.4
Using domain_name 0.5.24
Using http-cookie 1.0.2
Using multi_json 1.11.0
Using rack 1.6.0
Using mime-types 2.4.3
Using netrc 0.10.3
Using rest-client 1.8.0
Using jimson-temp 0.9.5
Using json 1.8.2
Installing stripe 1.20.1
Using stripe-ruby-mock 2.1.1
Using bundler 1.9.4
Bundle complete! 3 Gemfile dependencies, 16 gems now installed.
Warning: the gem 'stripe' was found in multiple sources.
Also found in:
  * https://code.stripe.com/
For example:
    gem 'stripe', :source => 'https://rubygems.org/'
Then uninstall the gem 'stripe' (or delete all bundled gems) and then install again.
```

This is on OS X 10.9 with Ruby 2.2.2 (p95) and Bundler 1.9.4.

Notes:
* The `json` gem is present as it seemed to be necessary to trigger the error message, possibly indicating a problem with mixed `source`s in gems in a dependency hierarchy (`stripe-ruby-mock` depends on `stripe` depends on `json`).
* Running `bundle install` again does not report an error. (Not a problem necessarily, just an observation for trying to reproduce this.)
* The error seemingly only appears on a fresh install (when `bundler` is the only installed gem prior to running `bundle install`) or when all of the gems in my example are installed except for `stripe`.

TimMoore:
Previously, an up-to-date lock file would retain the aggregate gem
source on all dependencies, so if the gems were not installed locally,
you'd get an ambiguous gem warning and possibly the wrong source
selected.

Fixes #3585

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -184,11 +184,10 @@ module Bundler
     # @return [SpecSet] resolved dependencies
     def resolve
       @resolve ||= begin
+        last_resolve = converge_locked_specs
         if Bundler.settings[:frozen] || (!@unlocking && nothing_changed?)
-          @locked_specs
+          last_resolve
         else
-          last_resolve = converge_locked_specs
-
           # Run a resolve against the locally available gems
           last_resolve.merge Resolver.resolve(expanded_dependencies, index, source_requirements, last_resolve)
         end
@@ -518,7 +517,9 @@ module Bundler
 
       converged = []
       @locked_specs.each do |s|
-        s.source = sources.get(s.source)
+        # Replace the locked dependency's source with the equivalent source from the Gemfile
+        dep = @dependencies.find { |dep| s.satisfies?(dep) }
+        s.source = (dep && dep.source) || sources.get(s.source)
 
         # Don't add a spec to the list if its source is expired. For example,
         # if you change a Git gem to Rubygems.
@@ -566,7 +567,7 @@ module Bundler
     end
 
     def satisfies_locked_spec?(dep)
-      @locked_specs.any? { |s| s.satisfies?(dep) && (!dep.source || s.source == dep.source) }
+      @locked_specs.any? { |s| s.satisfies?(dep) && (!dep.source || s.source.include?(dep.source)) }
     end
 
     def expanded_dependencies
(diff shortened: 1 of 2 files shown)


lib/bundler/definition.rb
```diff
@@ -184,11 +184,10 @@ def specs_for(groups)
     # @return [SpecSet] resolved dependencies
     def resolve
       @resolve ||= begin
+        last_resolve = converge_locked_specs
```
TimMoore: Moving this out is a bigger behavioural change than I would have liked, but it was the smallest code change I could work out. All of the tests pass (locally)... is there anything else this could break?

indirect: I think you're right. Now that we have sources (and don't save those sources in the lock), we have to resolve the sources using the Gemfile even when it's frozen/locked/etc.

PR #3601, opened 2015-04-28
add ignore messages documentation

### LINKED ISSUE #3600: Fix (& document) ignore_message config
Right now the ignore_message config isn't documented in the `config` manpage, and it also causes empty post-install messages to be printed out for every gem that isn't actively ignored. We should fix that for 1.10 :)

/cc @pducks32

pducks32:
Fixes #3600 part 1


indirect:
:+1:

PR #3602, opened 2015-04-28
Fix ignore messages

### LINKED ISSUE #3600: Fix (& document) ignore_message config

pducks32:
Fixes #3600 part 2

### DIFF
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -117,8 +117,10 @@ module Bundler
         Bundler.ui.confirm "Use `bundle show [gemname]` to see where a bundled gem is installed."
       end
 
-      Installer.post_install_messages.to_a.each do |name, msg|
-        print_post_install_message(name, msg) unless Bundler.settings["ignore_messages.#{name}"]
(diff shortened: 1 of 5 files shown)


segiddins:
This needs specs!

indirect:
I like this change! Being able to disable all messages is useful. However, I'm not sure this actually fixes the problem with ignoring a single gem. When I ignored just one gem, it caused empty install messages to be printed for every other gem in the bundle. That's what I was talking about in #3600.

pducks32:
It does because before `has_post_install_message` evaluated to true when it was `empty?` which is obviously the opposite which meant that the ones that did print were empty.

pducks32:
Is there any way to rerun the tests without having to commit a trivial commit as its rubygems' error.

indirect:
Unfortunately, RubyGems master is actually broken right now. :(

PR #3603, opened 2015-04-29
Bundler >= 1.9.0 fails to create gem plugins

amatsuda:
Since 17a4fc47bfad02de553e5a53b00ad38b4c905e18 (#3123), any gem name starting with an
existing const name has been regarded as "Invalid gem name" by `bundle gem` command.
However, "gem-foo" or "rails-bar" should be totally valid and rather preferable
naming for plugins for RubyGems/Rails.

This patch changes the rule to raise error only when the new gem name fully
matches an existing const name.

invalid: `gem`, `gem-version` (because `Gem::Version` already exists)
valid : `gem-foobar`, `gem-version-foobar` (when `FooBar`module does not exist)

### DIFF
--- a/lib/bundler/cli/gem.rb
+++ b/lib/bundler/cli/gem.rb
@@ -21,8 +21,8 @@ module Bundler
       underscored_name = name.tr('-', '_')
       namespaced_path = name.tr('-', '/')
       constant_name = name.gsub(/-[_-]*(?![_-]|$)/){ '::' }.gsub(/([_-]+|(::)|^)(.|$)/){ $2.to_s + $3.upcase }
-
       constant_array = constant_name.split('::')
+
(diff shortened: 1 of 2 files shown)


indirect:
Oh no! Thank you for fixing this. I will merge it to the 1.9 branch.

indirect:
@amatsuda it seems like the tests are failing on older Rubies. Could you check the test failures on Travis?

amatsuda:
@indirect Oh, sorry! Writing code that works on Ruby 1.8.7..2.2 is so hard! You're really doing a great job here!
Fixed and `push -f`ed the patch. Hope it works now..

indirect:
Thanks!

amatsuda:
:cake:

PR #3604, opened 2015-04-29
[Documentation] Add suggestion to remove pre-installed bundler

cdwort:
Hi all, 

I ran into all sorts of errors getting my development environment set up due to the fact that I chose a ruby at random and forgot that I had bundler installed in that ruby. 

`rake spec:deps` did call the code in the local repo, but that code found the classes defined by that pre-installed version. Given that it was quite a few versions back, things broke pretty badly.

I'm not sure if uninstalling bundler in the version of ruby you're using is the best solution, but it was the best one I could think of.

I'm new to bundler, but I'm not new to ruby. I recognize that this is ultimately a completely obvious mistake and good for a chuckle, but I do think it's important to provide some sort of warning.

Let me know if you'd rather handle this as an issue first and then a PR, or if you have suggestions for how to improve what I've written.

Thanks, 
Amy

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -20,15 +20,23 @@ bugs](https://github.com/bundler/bundler/issues?labels=small&state=open) and [sm
 
 Bundler doesn't use a Gemfile to list development dependencies, because when we tried it we couldn't tell if we were awake or it was just another level of dreams. To work on Bundler, you'll probably want to do a couple of things.
 
-  1. Install Bundler's development dependencies
+  1. Make sure Bundler is not already installed in your gemset
+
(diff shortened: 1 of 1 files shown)


segiddins:
Can you elaborate the issues you're running into? I've never had any problem running `rake spec:deps` on an MRI install that already had bundler.

indirect:
Hi Amy, thanks for reporting this! What you're describing sounds like a bug in the way Bundler uses rake or runs its tests. It definitely ought to be possible to have Bundler installed while working on Bundler. I'd like to make sure the problem you ran into wont happen to anyone else, but I'm not sure how to reproduce your problem myself. Could you answer the questions in [ISSUES](https://github.com/bundler/bundler/blob/master/ISSUES.md) (some won't apply to you, just skip those), and tell us exactly what commands you ran, as well as the exact error messages that you saw? Thanks!

cdwort:
Sounds good. Thanks @indirect and @segiddins. I'll close this PR and open an issue with steps to repo.

PR #3606, opened 2015-04-29
loop through the spec list to find gems with a certain name

tenderlove:
Since rubygems/rubygems@4fa03bb7aac9f25f44394e818433fdda9962ae8d
rubygems lazily loads specs from the filesystem for a particular name.
So if you request the "foo" gem, then rubygems will go to the FS and
find the gemspecs with the "foo" name.  **Before** the change, rubygems
would search through the loaded spec list for a gem with that name.
Bundler assumed that rubygems would always search through that spec
list, so it sets the specs and relies on that internal behavior.  Since
the internal behavior changed, we need to take that in to account in the
bundler internals.

This patch changes bundler to search through the spec list for a gem
with a particular name.  Gem::Specification.stubs should be supported in
the future (though not recommended because loading every spec isn't
super performant).

### DIFF
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -480,7 +480,9 @@ module Bundler
       end
 
       def find_name(name)
-        Gem::Specification.find_all_by_name name
+        Gem::Specification.stubs.find_all { |spec|
+          spec.name == name
+        }.map(&:to_spec)
       end
     end
 
@@ -521,7 +523,9 @@ module Bundler
       end
 
       def find_name(name)
-        Gem::Specification.find_all_by_name name
+        Gem::Specification.stubs.find_all { |spec|
+          spec.name == name
+        }.map(&:to_spec)
       end
 
       def fetch_specs(source, name)


lib/bundler/rubygems_integration.rb
```diff
@@ -480,7 +480,9 @@ def all_specs
       end
 
       def find_name(name)
-        Gem::Specification.find_all_by_name name
```
segiddins: I don't believe `.stubs` existed pre `2.1` or something?

segiddins:
See https://github.com/bundler/bundler/pull/3580 for another PR that adds stubs use.

lib/bundler/rubygems_integration.rb
```diff
@@ -521,7 +523,9 @@ def all_specs
...
+        Gem::Specification.stubs.find_all { |spec|
```
segiddins: is there any reason `Gem::Specification.stubs_for` wouldn't be appropriate here?

indirect:
@tenderlove thanks for the patch! I merged #3580, so we're using stubs now too—hopefully this will get everything fully stubbed and back to working.

PR #3608, opened 2015-05-01
[RubyGemsIntegration] Use .stubs_for when available

segiddins:
Depends on https://github.com/rubygems/rubygems/pull/1239.

### DIFF
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -607,10 +607,16 @@ module Bundler
         Gem::Ext::Builder::CHDIR_MONITOR
       end
 
-      def find_name(name)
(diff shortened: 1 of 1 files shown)


segiddins:
This won't work because `stubs_for` will bypass the array of specs we pass in to `all=` because `@@stubs_by_name[name]` will return `nil` and then RubyGems will use it's own search paths, instead of restricting itself to bundler's approved gems.

indirect:
Restarting the build for the latest RubyGems master...

segiddins:
I don't know why travis is showing this as having error?

segiddins:
This is now green. Let's :shipit: ?

indirect:
Yay! 🚢

PR #3612, opened 2015-05-02
Warn when upgrading lockfile major version & error when major version is...

segiddins:
... too low

Closes https://trello.com/c/1jnFZSbb/101-ensure-that-bundler-1-x-will-abort-and-request-an-upgrade-if-bundled-with-is-2-0-or-greater.

\c @smlance

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -249,12 +249,14 @@ module Bundler
         return
       end
 
-      locked_major = @locked_bundler_version.segments.first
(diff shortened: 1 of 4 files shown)


indirect:
testssssssssssssss

indirect:
Actually, can one of the tests include a lockfile with separate source sections as in #3593? I want to make sure that Bundler 1 hitting a Bundler 2 lockfile can still ask for an upgrade.

segiddins:
Updated

indirect:
\o/

PR #3616, opened 2015-05-05
set `@wrappers` and `@env_shebang` to true

tenderlove:
these conditionals would always evaluate to `true`, so just directly set
them to true without checking the option value

### DIFF
--- a/lib/bundler/source/path/installer.rb
+++ b/lib/bundler/source/path/installer.rb
@@ -11,8 +11,8 @@ module Bundler
           @gem_bin_dir       = "#{Bundler.rubygems.gem_dir}/bin"
           @bin_dir           = Bundler.requires_sudo? ? @tmp_bin_dir : @gem_bin_dir
           @gem_dir           = Bundler.rubygems.path(spec.full_gem_path)
-          @wrappers          = options[:wrappers] || true
-          @env_shebang       = options[:env_shebang] || true
+          @wrappers          = true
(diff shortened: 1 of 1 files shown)


lib/bundler/source/path/installer.rb
```diff
@@ -11,8 +11,8 @@ def initialize(spec, options = {})
           @gem_bin_dir       = "#{Bundler.rubygems.gem_dir}/bin"
           @bin_dir           = Bundler.requires_sudo? ? @tmp_bin_dir : @gem_bin_dir
           @gem_dir           = Bundler.rubygems.path(spec.full_gem_path)
-          @wrappers          = options[:wrappers] || true
-          @env_shebang       = options[:env_shebang] || true
+          @wrappers          = true
```
segiddins: maybe this should be `options.fetch(:wrappers, true)`?

indirect: I'd rather not start allowing it to be set if we haven't allowed it this entire time so far....

segiddins: It seems we were trying to set it to `false` in `path.rb`

indirect: Given that it changed 5 years ago in 93033523 and no one noticed, I think we should probably keep it the way it is now.

segiddins:
The single failure looks like a Travis issue, so this should be good

PR #3618, opened 2015-05-05
add test for #3277

hone:


indirect:
I don't think this is the underlying cause of #3277

indirect:
Er, I mean, I don't think it is only sudo installs that cause Bundler to leave empty directories behind in /tmp. Especially since this test passes, and Bundler (apparently) still leaves empty directories behind in /tmp. :)

segiddins:
@indirect we're pretty sure that it only did that pre-1.8

indirect:
oh. hmmmm....

indirect:
Nope, that test still passes on 1-7-stable. :/

hone:
@indirect hmm. removing the cleanup stuff made the test fail on my computer locally.

spec/install/gems/sudo_spec.rb
```diff
@@ -106,6 +109,20 @@
       expect(default_bundle_path("gems/rack-1.0.0")).to exist
       should_be_installed "rack 1.0"
     end
+
+    it "cleans up the tmpdirs generated", focused: true do
```
segiddins: 1.9 hash syntax is no good

hone: focused tests are also bad. let me fix that.

PR #3619, opened 2015-05-05
[LockfileParser] Leniently handle unknown states

segiddins:


segiddins:
Basically, Bundlers pre-1.10 choke on the `BUNDLED WITH` section

indirect:
Gah. Thanks for this. I guess I'm inclined to now change the format of the `BUNDLED_WITH` section, because I _really_ want 1.9 to be able to install lockfiles created by 1.10 :/

segiddins:
If we have any more releases of them, this should also go into 1.7/1.8

Issue #3233, opened 2014-11-05
Distinguish CI command runs from the rest


andremedeiros:
Right now, the commands are being reported and counted without differentiation as to their source.

@indirect suggests we should split them up by source when they run on CI systems.

A brief research wielded these environment variables we can use to pinpoint what CI system is running:

- Travis CI: `TRAVIS` (or, even better, `HAS_JOSH_K_SEAL_OF_APPROVAL`); /cc @joshk
- CircleCI: `CIRCLECI` (or any of the other env variables it sets when starting a container);
- Semaphore: `SEMAPHORE`;
- Jenkins: `JENKINS_URL`;
- Buildbox: `BUILDBOX`;
- Codeship: `CI_NAME` set to `codeship`;
- Go: `GO_SERVER_URL`;
- SnapCI: `SNAP_CI`;

Let me know if I've missed anything before I do the work for this.

simi:
What about to distinguish command runs from non-interactive shells (CI, deploy, ...) also? But I'm not sure if that's easy to do universal check.

fotanus:
If this turns out to be interesting, I can look for a way on how to do. Maybe open a new issue for that one?

segiddins:
closed by https://github.com/bundler/bundler/pull/3646.

PR #3622, opened 2015-05-06
Distinguish CI command runs

fotanus:
This patch add CI information on the user_agent in order for us to track if a given command was run by a CI, as discussed in #3233.

### DIFF
--- a/lib/bundler/fetcher.rb
+++ b/lib/bundler/fetcher.rb
@@ -183,18 +183,16 @@ module Bundler
     def cis
       env_cis = {
         "TRAVIS" => "travis",
-        "HAS_JOSH_K_SEAL_OF_APPROVAL" => "josh_k_seal",
         "CIRCLECI" => "circle",
         "SEMAPHORE" => "semaphore",
         "JENKINS_URL" => "jenkins",
         "BUILDBOX" => "buildbox",
         "GO_SERVER_URL" => "go",
-        "SNAP_CI" => "snap"
+        "SNAP_CI" => "snap",
+        "CI_NAME" => ENV["CI_NAME"],
+        "CI" => "ci"
       }
-      cis = env_cis.find_all{ |env, ci| ENV[env]}.map{ |env, ci| ci }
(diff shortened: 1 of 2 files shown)


fotanus:
This needs to accepted along with [this PR on bundler-api](https://github.com/bundler/bundler-api/pull/75)

lib/bundler/fetcher.rb
```diff
@@ -178,6 +180,23 @@ def inspect
...
+        "HAS_JOSH_K_SEAL_OF_APPROVAL" => "josh_k_seal",
+        "CIRCLECI" => "circle",
+        "SEMAPHORE" => "semaphore",
+        "JENKINS_URL" => "jenkins",
+        "BUILDBOX" => "buildbox",
+        "GO_SERVER_URL" => "go",
+        "SNAP_CI" => "snap"
+      }
+      cis = env_cis.find_all{ |env, ci| ENV[env]}.map{ |env, ci| ci }
+      cis << ENV["CI_NAME"] if ENV["CI_NAME"]
```
segiddins: why aren't these in the hash?

```ruby
{
  "CI_NAME" => ENV["CI_NAME"],
  "CI" => "ci",
```

fotanus: makes sense, will change it :+1:

fotanus:
Oh, well, it seems that it is not passing on travis, because this is all about CIs. Let me rewrite this test to be more generic...

indirect:
Happy to include this in the 1.10 release, as long as the specs are passing before we are ready to release. :)

andremedeiros:
Actually, this is failing because you're adding `josh_k_seal`. Take that out and I'm sure it will pass ;-)

andremedeiros:
To give you context, using that environment variable to detect Travis was more of a joke than anything, and we should probably not use it :-P

fotanus:
@andremedeiros I wonder how I managed to break all that in my last commit? o.O

andremedeiros:
Leave it to me. I'll have a look!

fotanus:
closing this in favor of #3646

PR #3626, opened 2015-05-07
Speed up the build and release tasks

pjump:
by not skipping man:clean.

Doing man:clean on every build defeats the purpose of having the
txt2roff/ronn2roff tasks declared as file tasks, rendering the file-task
associated timestamp checking useless and it slows down `rake build` by
about 3.5s on my laptop.

This removes the man:clean dependencies.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -3,6 +3,7 @@ $:.unshift File.expand_path("../lib", __FILE__)
 require 'shellwords'
 require 'benchmark'
 
+
 RUBYGEMS_REPO = File.expand_path("tmp/rubygems")
 BUNDLER_SPEC = Gem::Specification.load("bundler.gemspec")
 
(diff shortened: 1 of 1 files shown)


indirect:
The point of defining the tasks as file tasks is so that you can run `rake man:build` repeatedly. It has nothing to do with build/release. The reason that it is _required_ to run `man:clean` before any release is that manpages can be added and removed between releases, and the only way to get a clean build of the manpages for a particular release is to run `clean` and then `build`.

pjump:
Well, you still don't have to clobber all of your generated manpages and start from scratch, you just have to mirror deletes in the source directory, which takes abot 3--4 LOCs and it massively speeds up the whole thing:
https://github.com/bundler/bundler/compare/master...pjump:rakefile?expand=1

indirect:
Can you explain why you want to optimize `rake build` so much? I'm willing to accept a pull request that speeds it up, I just haven't ever even felt like it was slow. It really only needs to be run at release-time...

Issue #3621, opened 2015-05-06
Unable to `bundle viz` if a dependency is satisfied by a prerelease


aprescott:
This came out of #3217, which describes hitting a `NoMethodError`:

```
`block in _populate_relations': undefined method `runtime_dependencies' for nil:NilClass (NoMethodError)
```

I spent some time looking into this situation and have a potential fix (given below), but I didn't know whether I should PR it, since I'm unclear on whether bundler should change or whether rubygems should change, given the solution.

I figured I would give an explanation of what I think is happening, along with a fix, and then some discussion could happen.

I'm using bundler 1.9.5 here. Entire reproduction steps are given below.

## Reproducing

The setup here is a gem, `eco`, which requires `eco-source`. If the `eco-source` dependency happens to be satisfied by a prerelease version, then you can't `bundle viz` on Rubygems 2.4.0+. This forces `eco-source` to be a prerelease, just to illustrate the point.

```bash
# 2.3.0: WORKS
# 2.4.0: FAILS
gem_version_for_repro=2.4.0

cd /tmp
rm -rf viz-dependency-error
mkdir viz-dependency-error
cd viz-dependency-error

echo "2.2.2" > .ruby-version

# get any .ruby-version reloading
cd ..
cd -
ruby -v # should be 2.2.2., just for isolating the version fully

cat <<EOF > Gemfile
gem "eco", "= 1.0.0"
gem "eco-source", "= 1.1.0.rc.1"
EOF

gem update --system $gem_version_for_repro
gem install bundler -v '= 1.9.5' --no-document
bundle install

bundle viz
```

The `Gemfile.lock` here is:

```
GEM
  specs:
    coffee-script (2.4.1)
      coffee-script-source
      execjs
    coffee-script-source (1.9.1.1)
    eco (1.0.0)
      coffee-script
      eco-source
      execjs
    eco-source (1.1.0.rc.1)
    execjs (2.5.2)

PLATFORMS
  ruby

DEPENDENCIES
  eco (= 1.0.0)
  eco-source (= 1.1.0.rc.1)
```

Note that `eco` requires any `eco-source`, but here it happens to be satisfied by a prerelease gem.

## Outline

Let's say you have an application which relies on `haml-rails`. Suppose it resolves to `haml-rails` 0.7.0, which has this set of dependencies:

```ruby
haml-rails (0.7.0)
  actionpack (>= 4.0.1)
  activesupport (>= 4.0.1)
  haml (>= 3.1, < 5.0)
[...]

PR #3629, opened 2015-05-07
Allow bundle viz to work with dependencies satisfied by prereleases

aprescott:
Suppose a Gemfile specifies a dependency, `foo`:

```
foo
```

Let's say `foo` requires a certain `bar`:

```
foo
  bar (>= 0.5)
```

If this dependency happens to be satisfied by a prerelease `bar`, e.g., 0.6.pre, then `bundle viz` will error.

The cause of this error is that `Bundler::Graph#_populate_relations` executes

```ruby
child_dependencies = dependency.to_spec.runtime_dependencies.to_set
```

and if `dependency` (here, `bar`), is a prerelease, then `dependency.to_spec` will be `nil` on Rubygems 2.4.0 and higher because of changes to `#to_spec`.

By forcing `dependency.prerelease = true`, then `to_spec` won't ignore the prerelease `bar` dependency, which prevents `dependency.to_spec` to from being `nil`.

This should address #3621 and #3217.

### DIFF
--- a/spec/commands/viz_spec.rb
+++ b/spec/commands/viz_spec.rb
@@ -1,0 +1,46 @@
+require "spec_helper"
+
+describe "bundle viz" do
+  before :each do
+    `gem install ruby-graphviz`
+  end
+
+  context "with a standard Gemfile" do
+    before :each do
+      `gem install httpi -v "= 2.2.7"` # requires any rack
+      `gem install rack -v "= 1.1.1"`
+      # TODO: this whole approach is bad
+
+      install_gemfile <<-G
+        source "file://#{system_gem_path}"
+        gem "httpi", "= 2.2.7"
+        gem "rack", "= 1.1.1"
+      G
+    end
+
+    it "correctly generates a graph" do
+      bundle "viz"
+
+      expect(out).to include("gem_graph.png")
+    end
+  end
+
+  context "with a Gemfile that has a prerelease as a satisfying dependency" do
+    before :each do
+      `gem install httpi -v "= 2.2.7"` # requires any rack
+      `gem install rack -v "= 1.1.1.pre"`
+
+      install_gemfile <<-G
+        source "file://#{system_gem_path}"
+        gem "httpi", "= 2.2.7"
+        gem "rack", "= 1.1.1.pre"
+      G
+    end
+
+    it "correctly generates a graph" do
+      bundle "viz"
+
+      expect(out).to include("gem_graph.png")
+    end
+  end
+end
--- a/lib/bundler/graph.rb
+++ b/lib/bundler/graph.rb
@@ -36,6 +36,9 @@ module Bundler
         else
           tmp = Set.new
           parent_dependencies.each do |dependency|
+            # if the dependency is a prerelease, allow to_spec to be non-nil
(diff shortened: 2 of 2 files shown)


spec/commands/viz_spec.rb
```diff
@@ -0,0 +1,46 @@
+require "spec_helper"
+
+describe "bundle viz" do
+  before :each do
+    `gem install ruby-graphviz`
```
segiddins: umm this should probably be done in `rake spec:deps` if absolutely necessary?

spec/commands/viz_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  context "with a standard Gemfile" do
+    before :each do
+      `gem install httpi -v "= 2.2.7"` # requires any rack
```
segiddins: you should be able to create an index that has everything you need

spec/commands/viz_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      `gem install rack -v "= 1.1.1"`
+      # TODO: this whole approach is bad
```
aprescott: I struggled with getting this test set up correctly.

There are a bunch of problems here, beyond what this comment is about:

1. The test output can contain a _ton_ of warnings from `ruby-graphviz`'s output, like: "warning: duplicated key at line 986 ignored: "red".
2. I'm using backticks to install gems into the `system_gem_path`, so that they're available to `install_gemfile`. This is because I don't understand the `spec/` structure enough to get the necessary test conditions in place. That means there's an actual gem install happening, I believe. I spent a ton of time trying to find this pair of `httpi` and `rack` — they install quickly enough, with no native extensions, etc., but this can probably be improved.
3. I needed to `gem install ruby-graphviz` in the outer `before`, so that I could get `bundle viz` working. Again, same problem as (2), really.

spec/commands/viz_spec.rb
aprescott: I'm unfortunately not familiar enough with the Bundler test suite to know how to do that. :( I mentioned that in https://github.com/bundler/bundler/pull/3629#discussion-diff-29895133.

indirect:
Thanks! I'm going to squash this down into a single commit and use fake gems in the test rather than real ones. Merged as 072fd0f.

aprescott:
@indirect that commit doesn't seem to exist in Bundler. Maybe something hasn't been pushed? (I thought I'd only had 1 commit too...)

indirect:
Oops! My bad. I merged it on a plane, and couldn't get the wifi working well enough to push it back. It'll be there later today. Thanks again!

indirect:
also I got the commit wrong, it's actually e3e9f4add3ab25776c43816b916543995db8fb4e. but now it's pushed!

indirect:
by the way _thank you_ for writing specs for the viz command! it's great to finally have some :)

aprescott:
:heart: very happy to help where I can! It can be pretty difficult to persuade yourself to backfill, so I'm glad I stopped myself from being lazy.

Thanks for spending the time cleaning up the tests. And of course, for all your continuing work on Bundler!

Issue #3624, opened 2015-05-07
Bundler version warning printed 3 times


indirect:
Maybe we need to use the same set-based strategy here that we do for deprecations?

```

$ cat Gemfile.lock
GEM
  specs:
    rake (10.4.2)

PLATFORMS
  ruby

DEPENDENCIES
  rake

BUNDLED WITH
  1.99.0.pre

$ dbundle install
Warning: the running version of Bundler is older than the version that created the lockfile. We suggest you upgrade to the latest version of Bundler by running `gem install bundler --pre`.
Warning: the running version of Bundler is older than the version that created the lockfile. We suggest you upgrade to the latest version of Bundler by running `gem install bundler --pre`.
Using rake 10.4.2
Warning: the running version of Bundler is older than the version that created the lockfile. We suggest you upgrade to the latest version of Bundler by running `gem install bundler --pre`.
Using bundler 1.10.0.pre.1
Bundle complete! 1 Gemfile dependency, 2 gems now installed.
Use `bundle show [gemname]` to see where a bundled gem is installed.
```

/cc @segiddins

presidentJFK:
It looks like this can be closed with https://github.com/bundler/bundler/pull/3716 being merged in.

PR #3631, opened 2015-05-07
Avoid showing warning messages twice

fotanus:
This is my tentative for fixing #3624.

It don't involve sets :(

### DIFF
--- a/lib/bundler/ui/shell.rb
+++ b/lib/bundler/ui/shell.rb
@@ -11,6 +11,7 @@ module Bundler
         end
         @shell = Thor::Base.shell.new
         @level = ENV['DEBUG'] ? "debug" : "info"
+        @told_messages = []
       end
 
       def info(msg, newline = nil)
@@ -80,6 +81,9 @@ module Bundler
 
       # valimism
       def tell_me(msg, color = nil, newline = nil)
+        return if @told_messages.include? msg
+        @told_messages << msg
+
         msg = word_wrap(msg) if newline.is_a?(Hash) && newline[:wrap]
         if newline.nil?
           @shell.say(msg, color)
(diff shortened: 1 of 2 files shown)


lib/bundler/ui/shell.rb
```diff
@@ -80,6 +81,9 @@ def silence
 
       # valimism
       def tell_me(msg, color = nil, newline = nil)
+        return if @told_messages.include? msg
```
segiddins: this might catch more than warning messages

fotanus: Yeah, the question is whether is desirable to show a message more than once to the user. When intended, the message can be add twice in a row, in the same string. Right?

segiddins: well in theory the same message can be intended twice in different context. Ideally, we'd limit de-duplication just to warning messages

fotanus: I can make that happen if you guys want to merge this way. 

Those are the levels currently:
```
LEVELS = %w(silent error warn confirm info debug)
```
Confirming, should I filter only by warn level?

fotanus:
Please note this avoid showing **any** warning message twice.

lib/bundler/ui/shell.rb
```diff
@@ -11,6 +11,7 @@ def initialize(options = {})
         end
         @shell = Thor::Base.shell.new
         @level = ENV['DEBUG'] ? "debug" : "info"
+        @told_messages = []
```
indirect: @fotanus change the name of this ivar to `warning_history` or `past_warnings`, and this looks good to go.

fotanus:
Closing this in favor of #3716 with the requested fix. This is so because I deleted this branch or something :stuck_out_tongue:

PR #3633, opened 2015-05-08
fast clean manpage build

pjump:
@indirect I don't care for it. As a matter of fact, I think it's an awesome idea that each `rake build`, regardless of whether it has actual work to do or not, should give people an obligatory 3 seconds to ponder about how awesome bundler is. :)

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -269,6 +269,7 @@ begin
   namespace :man do
     directory "lib/bundler/man"
 
+    generated_files = []
     Dir["man/*.ronn"].each do |ronn|
       basename = File.basename(ronn, ".ronn")
       roff = "lib/bundler/man/#{basename}"
@@ -281,6 +282,7 @@ begin
         sh "groff -Wall -mtty-char -mandoc -Tascii #{roff} | col -b > #{roff}.txt"
       end
 
+      generated_files << roff << "#{roff}.txt"
       task :build_all_pages => "#{roff}.txt"
     end
 
@@ -289,7 +291,8 @@ begin
 
     desc "Clean up from the built man pages"
     task :clean do
-      rm_rf "lib/bundler/man"
+      dirt = Dir["lib/bundler/man/*"] - generated_files
+      rm dirt
     end
 
     task(:require) { }


indirect:
You completely ignored my question, though: why are you building Bundler? If you want a released version, you can just download it from RubyGems.org, and if you want a non-released version, you can just use it directly without building it.

indirect:
Unfortunately, this patch also results in releasing a gem with incorrect manpages. For example:

```
$ git checkout 1-9-stable
$ rake release
$ git checkout master
$ rake release
```

If this patch were applied, the manpages in the release from master would be out of date and apply to the 1.9 bugfix release.

pjump:
How? A git checkout is just like any other update. The changed files in the tree will have newer timestamps (even if you check out older commits---git doesn't store file timestamps), causing their respective targets to get rebuilt. The new clean task will mirror deletes invoked by that checkout.

pjump:
If you want the current behavior, you can get it with less code by deleting the clean task and any references thereto and converting all the file task involved in building manpages to regular tasks. Declaring a file task only to have it always paired with a clean task seems weird. But then again, you should be able to make it more efficient with file tasks and this patch. And it should work with git checkouts too.

indirect:
Huh, totally didn't realize that. Cool. I still feel weird about using global state inside file tasks, so I think I'll reimplement this to just diff the two arrays of source and destination files. Thanks for the patch!

pjump:
You could've just hit merge and add the clobber task if that's what you wanted. Your solution didn't eliminate the `generated_files` block-local variable shared between inner blocks (which, I assume, is what you referred to by global state, which I think there's little point to eliminating in Rakefiles as those are inherently about global state (but it doesn't matter, because rakefiles aren't libraries to be reshared)), only replaced it with the `source` block-local variable that's shared among inner blocks (same thing, different name),  made it more "indirect", about twice as long, less efficient (the each + include? should be much less efficient that `Array#-`, but that shouldn't matter here with the fileset size as small as it is) and introduced two bugs (leftover *.txt files don't get cleaned now and the /gemfile\.5(\.txt)?/ is overly broad and should be /^gemfile\.5(\.txt)?$/  or better yet, if you do the mappings right, you shouldn't need to give this file special treatment).

indirect:
Petr, every single thing that you’ve written in this issue tracker has been hostile, condescending, passive-aggressive, or all three at once. I've been trying to ask questions, understand why you care about making these changes, and come up with a result that works for everyone.

You've repeatedly ignored my questions, demanded changes without any explanation of why those changes matter. To top everything off, you've been wrong repeatedly while also being insufferably rude. At this point, however, it doesn’t actually matter whether what you’ve been saying is technically correct or not—your hostile behavior violates the [Bundler project code of conduct](https://github.com/bundler/bundler/blob/master/CODE_OF_CONDUCT.md), and you need to stop.

Please keep the code of conduct in mind if you open any issues or pull requests in the future.

PR #3635, opened 2015-05-08
[Resolver] Speed up cases where a dependency with more possible versi…

### LINKED ISSUE #3634: Simple Gemfile causes "bundle update" to never finish
When running ```bundle update``` with the following Gemfile, the command never completes.  It runs with "Resolving dependencies" forever.  This happens on new installations using RVM for MRI and JRuby.

```ruby

gem 'rails', '<4.2.0'
gem 'actionpack-page_caching'
gem 'simple_workflow'
```

segiddins:
…ons is more constrained

Closes #3634.

\c @Who828

### DIFF
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -297,6 +297,7 @@ module Bundler
         [
           activated.vertex_named(name).payload ? 0 : 1,
           @prereleases_cache[dependency.requirement] ? 0 : 1,
+          dependency.requirements_list.size,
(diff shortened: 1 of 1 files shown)


indirect:
This is green enough for me to at least be happy trying it. Thanks!

@Who828, how do you feel about creating a project that runs benchmarks against known Gemfiles? Do you still have that collection of Gemfiles you used to test resolver speed? We might even be able to add the output to RubyBench.org in the future...

PR #3636, opened 2015-05-08
[LockfileParser] Allow reverting BUNDLED WITH padding to 2 spaces

segiddins:


indirect:
I'm guessing this is so that 2.0 and up can use two spaces and everything can go back to being happy? Sounds good.

segiddins:
Correct!

PR #3638, opened 2015-05-10
Add option to use docker to run specs

stevenjack:
### Problem

I was trying to run the specs but the version of `rspec` I had on my machine wasn't correct and I couldn't use `bundle exec` to select the correct version of the gems as I was testing `bundler` itself and `bundler` doesn't use a `Gemfile`.

### Solution

Have a containerised, predictable environment to run the specs in. The `Dockerfile` adds the spec dependencies then adds `rake` as the entry point and `spec` as the command. To aid running the specs I've also added `docker-compose.yml` so you don't have to construct the whole `docker` command.

There could be another option instead of having `docker-compose` installed as it's  couple of dependencies, an image could be created in the hub on each push to `master` so you could run the following:

`docker run -v /path/to/my/clone:app bundler`

Which mounts your version of bundler, then uses the publisher bundler image that would be updated on each push to master to run the specs.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -20,6 +20,8 @@ bugs](https://github.com/bundler/bundler/issues?labels=small&state=open) and [sm
 
 Bundler doesn't use a Gemfile to list development dependencies, because when we tried it we couldn't tell if we were awake or it was just another level of dreams. To work on Bundler, you'll probably want to do a couple of things.
 
+## Standard method
+
   1. Install Bundler's development dependencies
 
         $ rake spec:deps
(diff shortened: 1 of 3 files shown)


stevenjack:
@indirect what I was talking about in IRC earlier this evening..

segiddins:
@stevenjack given that the `Rakefile` will select the appropriate version of `rspec` to load, I don't really see why this is necessary?

indirect:
I'm kind of ambivalent about this, myself... I definitely want to make it as easy as possible to get started working on Bundler. On the other hand, installing docker and docker-compose is hard enough that this is both the first request we've gotten for docker support and the first time we've heard a report of the problem that this exists to solve. Let's wait to see if there's anyone else who would be helped by a Dockerfile before merging this.

indirect:
Since no one else has been interested in this yet, I think we're going to pass on merging this. I'm happy to link to a repo or gist that explains how to get Bundler up and running using Docker, though.

PR #3643, opened 2015-05-14
Autoconfig gemfile

ineiti:
This finishes off the work with different Gemfiles. Now if you do a

```
bundle install --gemfile Gemfile.dev
```

the new ```Gemfile.dev``` will be saved to the configuration and be used automatically for all further references.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -145,6 +145,7 @@ module Bundler
       @definition = nil if unlock
       @definition ||= begin
         configure
+        ENV['BUNDLE_GEMFILE'] ||= Bundler.settings[:gemfile]
         upgrade_lockfile
         Definition.build(default_gemfile, default_lockfile, unlock)
       end
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -16,8 +16,10 @@ module Bundler
     def initialize(*args)
       super
       current_cmd = args.last[:current_command].name
-      custom_gemfile = options[:gemfile] || Bundler.settings[:gemfile]
-      ENV['BUNDLE_GEMFILE']   = File.expand_path(custom_gemfile) if custom_gemfile
+      if custom_gemfile = options[:gemfile]
+        Bundler.settings[:gemfile] = custom_gemfile
+        ENV['BUNDLE_GEMFILE'] = File.expand_path(custom_gemfile)
+      end 
       Bundler.settings[:retry] = options[:retry] if options[:retry]
       Bundler.rubygems.ui = UI::RGProxy.new(Bundler.ui)
       auto_install if AUTO_INSTALL_CMDS.include?(current_cmd)


simi:
I think it is not good idea to set ENV variable.

ineiti:
It only sets it if it is not already set, so the priorities are:

1. use Gemfile from BUNDLE_GEMFILE
2. use Gemfile from command-line (--gemfile)
3. use Gemfile from .bundle/config
4. use default Gemfile

In fact, there are quite some places in the code where ENV['BUNDLE_GEMFILE'] is set... Which is a bad excuse to go on, I know.

segiddins:
We're actually moving away from auto-setting config variables in 2.0, so I'm not really sure that we want to add another right now?

indirect:
Since we are removing remembered options, I don't think it makes sense to add one here. Please use `bundle config gemfile Gemfile.dev` in order to change the Gemfile location.

ineiti:
Le 16/05/2015 22:38, André Arko a écrit :
> Since we are removing remembered options, I don't think it makes sense
> to add one here. Please use |bundle config gemfile Gemfile.dev| in order
> to change the Gemfile location.

OK - your call.

But I still have the problem that after |bundle config gemfile 
Gemfile.dev| the |require 'bundler/setup'| doesn't pick up the new 
config, it still uses |Gemfile|. Is that on purpose? Part of my patch 
'fixes' that, too.

segiddins:
@ineiti the proper fix for that is to probably alter `SharedHelpers.default_gemfile` to also check the config at each step up

ineiti:
Le 18/05/2015 05:29, Samuel E. Giddins a écrit :
> @ineiti <https://github.com/ineiti> the proper fix for that is to
> probably alter |SharedHelpers.default_gemfile| to also check the config
> at each step up

@samuel yes, but SharedHelpers.default_gemfile gets called the first 
time BEFORE the config-file is available, in Bundler.root

This is why I put the code in Bundler.definition, which seemed the first 
place where the config-file is available.

segiddins:
I think it would make sense to modify shared helpers to fix the issue there, as the config is found relative to the Gemfile path. 

-Samuel E. Giddins

On May 18, 2015, at 2:12 AM, Linus Gasser <notifications@github.com> wrote:

Le 18/05/2015 05:29, Samuel E. Giddins a écrit :
> @ineiti <https://github.com/ineiti> the proper fix for that is to
> probably alter |SharedHelpers.default_gemfile| to also check the config
> at each step up

@samuel yes, but SharedHelpers.default_gemfile gets called the first 
time BEFORE the config-file is available, in Bundler.root

This is why I put the code in Bundler.definition, which seemed the first 
place where the config-file is available.

ineiti:
Le 18/05/2015 14:08, Samuel E. Giddins a écrit :
> I think it would make sense to modify shared helpers to fix the issue
> there, as the config is found relative to the Gemfile path.

OK, so the .bundle/config depends on the Gemfile. But the name of the 
Gemfile is found in .bundle/config :(

Would it be acceptable to modify _SharedHelpers.find_gemfile_ to 
_search_up_ for 'Gemfile' and '.config' and then act accordingly?

```
if .config found
   look for new name
elsif Gemfile found
   ok
end
```

And why is the 'gems.rb' in the search_up-path?

segiddins:
Yes, a modification like that would make sense in my opinion. gems.rb is an alternate name for the Gemfile.

ineiti:
OK, before I try a new pull-request, how about that:

1. search for global __$HOME/.bundle/config__ and initialize ```Bundler.settings```
2. ```search_up``` for ```[ .bundle, Bundler.settings[:gemfile], 'Gemfile' ]```
3. if ```.bundle``` is found, initialize ```Bundler.settings``` and search for ```[ Bundler.settings[:gemfile], 'Gemfile' ]``` in that directory

One problem is ```SharedHelpers.in_bundle?``` which is called from ```bundler/setup``` before any configuration is read. So we'd have to initialize the settings in ```SharedHelpers``` instead of ```bundler.rb```, as this is the only thing required in ```bundler/setup```.

PR #3646, opened 2015-05-14
Distinguish CI command runs

fotanus:
This patch add CI information on the user_agent in order for us to track
if a given command was run by a CI, as discussed in #3233.

### DIFF
--- a/spec/bundler/fetcher_spec.rb
+++ b/spec/bundler/fetcher_spec.rb
@@ -16,5 +16,21 @@ describe Bundler::Fetcher do
       expect(fetcher.user_agent).to match(/ruby\/(\d.)/)
       expect(fetcher.user_agent).to match(/options\/foo,bar/)
     end
+
+    describe "include CI information" do
+      it "from one CI" do
+        ENV["JENKINS_URL"] = "foo"
+        ci_part = fetcher.user_agent.split(' ').find{|x| x.match(/\Aci\//)}
+        expect(ci_part).to match("jenkins")
+      end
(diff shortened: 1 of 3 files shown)


spec/bundler/fetcher_spec.rb
```diff
@@ -16,5 +16,21 @@
       expect(fetcher.user_agent).to match(/ruby\/(\d.)/)
       expect(fetcher.user_agent).to match(/options\/foo,bar/)
     end
+
+    describe "include CI information" do
+      it "from one CI" do
+        ENV["JENKINS_URL"] = "foo"
```
simi: this ENV should be cleaned

fotanus: Done! Thanks!

simi: It will be better to keep and reassign old value (instead of forcing nil).

fotanus: Of course! Silly me!

fotanus:
Can I have this job on travis run again? I just tried locally and it worked with ruby 2.2.2. Looking at the travis log, it seems like a transient error. It have messages like this:

```
Could not fetch specs from file:/home/travis/build/bundler/bundler/tmp/gems/remote2/
```

indirect:
restarted the build, I think it was failing because RubyGems master was broken

fotanus:
@indirect Thank you!

indirect:
Thanks!

PR #3648, opened 2015-05-14
[Path] Allow arbitrary nesting of gemspecs inside a path source

segiddins:
Closes https://github.com/bundler/bundler-features/issues/87

### DIFF
--- a/lib/bundler/source/path.rb
+++ b/lib/bundler/source/path.rb
@@ -8,7 +8,7 @@ module Bundler
       attr_writer   :name
       attr_accessor :version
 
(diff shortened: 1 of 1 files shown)


indirect:
I believe (although I'm not positive) that we once changed this away from that because of the performance impact. Maybe it would be easier to let people opt in with specific paths rather than recursively stat every directory in the entire repo?

segiddins:
@indirect it's been this way since 1.0, so I can't find any commit where the behavior that I'm introducing was reverted.

segiddins:
Given that merging this would break using Rails as a git gem, let's not merge it.

PR #3651, opened 2015-05-15
closes  #3649 - Stop printing replacement message when config is not changed

### LINKED ISSUE #3649: Stop printing "replacement" message when config is not changed
When setting a config value, even if it does not change, we print `You are replacing the current global value of <KEY>, which is currently "<VALUE>"`. We should not print this if the value didn't change.

fotanus:
closes #3649

### DIFF
--- a/lib/bundler/cli/config.rb
+++ b/lib/bundler/cli/config.rb
@@ -59,7 +59,7 @@ module Bundler
               "This will take precedence over the global value you are setting"
           end
 
-          if global = locations[:global]
+          if global = locations[:global] and global != locations[:global]
             Bundler.ui.info "You are replacing the current global value of #{name}, which is currently #{global.inspect}"
           end
         end
--- a/lib/bundler/cli/config.rb
+++ b/lib/bundler/cli/config.rb
@@ -59,7 +59,7 @@ module Bundler
               "This will take precedence over the global value you are setting"
           end
 
-          if global = locations[:global] and global != locations[:global]
+          if (global = locations[:global]) && (global != locations[:global])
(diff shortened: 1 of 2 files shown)


lib/bundler/cli/config.rb
```diff
@@ -59,7 +59,7 @@ def run
               "This will take precedence over the global value you are setting"
           end
 
-          if global = locations[:global]
+          if global = locations[:global] and global != locations[:global]
```
segiddins: can we use `&&` instead?

fotanus: Done! Is there a code style guide for me to follow?

Strech: I think it's [Ruby style guide](https://github.com/bbatsov/ruby-style-guide) from bbathos

segiddins: I don't think we have a formal bundler style guide, but I personally follow CocoaPods', which is a modified version of bbatsov's

indirect:
Cherry-picked as 6506ac5419248a9426d9becd4d337375a9cd24f3, and fixed so all the tests pass in 9eef958bf531e723aa84f2334708aa3dc3feabb0. Thanks!

fotanus:
It seems that one test still fails (as it is breaking our current master build):

```
  1) .bundle/config global expands the path at time of setting
     Failure/Error: expect(out).to eq(File.expand_path(Dir.pwd + "/.."))
       
       expected: "/home/fotanus/code/bundler/tmp"
            got: ".."
       
       (compared using ==)
     # ./spec/commands/config_spec.rb:111:in `block (3 levels) in <top (required)>'

```

PR #3654, opened 2015-05-16
[Resolver] Modify sorting heuristic to check how constraining the req…

segiddins:
…uirements are

This should speed up cases where a simple-looking requirement actually eliminates a lot of the possibilities

\c @indirect @Who828

### DIFF
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -190,7 +190,6 @@ module Bundler
       @base = base
       @resolver = Molinillo::Resolver.new(self, self)
       @search_for = {}
-      @prereleases_cache = Hash.new { |h,k| h[k] = k.prerelease? }
       @base_dg = Molinillo::DependencyGraph.new
       @base.each { |ls| @base_dg.add_root_vertex ls.name, Dependency.new(ls.name, ls.version) }
     end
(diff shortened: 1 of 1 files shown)


lib/bundler/resolver.rb
```diff
@@ -296,14 +295,23 @@ def sort_dependencies(dependencies, activated, conflicts)
...
           activated.vertex_named(name).payload ? 0 : search_for(dependency).count,
         ]
       end
     end
 
+    def amount_constrained(dependency)
+      @amount_constrained ||= {}
+      @amount_constrained[dependency.name] ||= begin
+        base_dep = Dependency.new dependency.name, '>= 0.a'
+        all = search_for(DepProxy.new base_dep, dependency.__platform)
```
indirect: Doesn't calling __platform cause a request to download the gemspec? I worry that this is really, really slow when we don't already have the platform of the dep...

indirect:
So if I'm understanding this correctly, it's sorting dependencies whose constraints remove the largest percentage of the total search space to the top, right? That sounds like a good deal to me. Do we have  Gemfiles we can performance test this against?

lib/bundler/resolver.rb
segiddins: pretty sure `dependency` here is a `DepProxy` that is initialized with a `platform`

segiddins:
Try this (using `bundle lock --print`) against the `gxapi_rails` realworld spec

indirect:
This doesn't seem to slow down the common cases, and this speeds up some really bad cases by a huge amount. :+1:

segiddins:
Fwiw this sped up https://github.com/bundler/bundler/issues/3634 to the point where that Gemfile takes under a second and a half to `bundle lock --print`

indirect:
Awesome.

PR #3655, opened 2015-05-17
[RubyGemsIntegration] Mark loaded specs as activated

segiddins:


indirect:
What effects does this change have? Does it fix a bug?

segiddins:
Not that I can tell, but it does mean that the resolved specs are marked as `activated`, which `Specification#activate` does, and it's possible someone is relying on that instead of using the `loaded_specs` hash.

lib/bundler/rubygems_integration.rb
```diff
@@ -39,7 +39,12 @@ def loaded_specs(name)
     end
 
     def mark_loaded(spec)
-      Gem.loaded_specs[spec.name] = spec
+      name = spec.name
+      if existing = Gem.loaded_specs[name]
+        existing.activated = false
```
segiddins: probably needs to make sure that `existing.respond_to? :activated` if we want to merge this :/

indirect: Heh, yeah. Pretty sure that `activated` was invented later than some of the RubyGems versions we support.

segiddins:
This seems to have been merged manually.

PR #3657, opened 2015-05-17
[Path] Validate the gemspecs we find when globbing

segiddins:


indirect:
If we validate all path gemspecs, then we don't need to validate in the gemspec DSL method. It just calls the path DSL method next.

indirect:
Cherry-picked as 858afa5cfec914bd3c9d6242d1ae49cd6c227fa1.

PR #3658, opened 2015-05-17
Fix spec to expand directories

fotanus:
Fix failing spec on master. Was introduced on #3651


segiddins:
This was already fixed on 1-10-stable.

PR #3661, opened 2015-05-17
rails command is already aware of Bundler

spastorino:
Do not do `bundle exec rails ...`, rails command is already aware of this.

Not sure if I'd apply this change or you should just change the example and use a gem where running it with bundle exec makes sense

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ gem install bundler
 bundle init
 echo "gem 'rails'" >> Gemfile
 bundle install
-bundle exec rails new myapp
+rails new myapp
 ```
 
 See [bundler.io](http://bundler.io) for the full documentation.


indirect:
This won't actually work. For example, what if you have two Gemfiles with different versions of Rails? You have to `bundle exec` to get the correct Rails version. What if you install the bundle into a path? You have to `bundle exec` to get the `rails` command at all.

The only time `bundle exec` is not required is when you explicitly run `bin/rails` from inside a Rails app that is already generated.

spastorino:
@indirect yep I understand, that's why I've said that I'd not apply. I should've just opened an issue :). Anyway, I think this example is not good because not running this using bundle exec I think it's the most common case.

Anyway we can disagree what's the common case but we can agree that there are examples where doing bundle exec something is always the right thing to do.

I'm concerned about this http://wyeworks.com/blog/2011/12/27/bundle-exec-rails-executes-bundler-setup-3-times

indirect:
This is why everyone should use binstubs and run `bin/rails` etc. :)

spastorino:
If you think the example if right is ok to me, I was just trying to help pointing out that what's being done is not common and really weird.

`gem 'rails'` does not install into a path and unless you do something that is not shown in the example the last Rails version will be installed and system wide.

Also after that `rails new myapp` another directory will be created with a Rails app on it and also containing another Gemfile.

I just think that all that is confusing to only show how to install and use Bundler.

:heart: :heart: :heart:

indirect:
That's a good idea. We should explain that after the rails app is generated, you should run `bin/rails` instead of `bundle exec rails`. Thanks!

Issue #3672, opened 2015-05-22
Bundler reports issues in `Gemfile.lock` regardless of the file being used as gemfile


AgentAntelope:
For example, in the event of a merge conflict in the file GemfileFast, bundler provides the following error:

![screen shot 2015-05-22 at 10 58 58](https://cloud.githubusercontent.com/assets/1962412/7768298/98adfa66-0071-11e5-9ec3-fad80c0e9299.png)

rgb-one:
Can you provide more information following the guidelines in [ISSUE](https://github.com/bundler/bundler/blob/master/ISSUES.md).

segiddins:
@rgb-one no additional info needed. 

Yes, there are places where we hardcode `Gemfile.lock` where we shouldn't.

AgentAntelope:
Thanks, @segiddins.

I've had a bit of a look at this in an attempt to fix it. It seems that the `LockfileParser`, which generates this error, has no knowledge of the filename of the file it's trying to parse. I attempted to give it that information, but the assumptions seems to be baked in in quite a few places; Not just the lockfile parser, but the `lockfile` helper method, and other places too.

Sadly I think this is too large a piece of work for me without knowing more about the direction the code is headed in.

PR #3677, opened 2015-05-26
Use the actual lockfile name in messages instead of assuming its alwa…

segiddins:
…ys Gemfile.lock

Closes https://github.com/bundler/bundler/issues/3672.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -330,7 +330,7 @@ module Bundler
     def ensure_equivalent_gemfile_and_lockfile(explicit_flag = false)
       msg = "You are trying to install in deployment mode after changing\n" \
             "your Gemfile. Run `bundle install` elsewhere and add the\n" \
-            "updated #{Bundler.default_lockfile} to version control."
+            "updated #{Bundler.default_lockfile.relative_path_from(SharedHelpers.pwd)} to version control."
 
       unless explicit_flag
         msg += "\n\nIf this is a development machine, remove the #{Bundler.default_gemfile} " \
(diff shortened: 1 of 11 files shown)


lib/bundler/definition.rb
```diff
@@ -267,7 +267,7 @@ def lock(file)
       File.open(file, 'wb'){|f| f.puts(contents) }
     rescue Errno::EACCES
       raise Bundler::InstallError,
-        "There was an error while trying to write to Gemfile.lock. It is likely that \n" \
+        "There was an error while trying to write to #{File.basename(file)}. It is likely that \n" \
```
adarsh: Why is this change different from the others? Shouldn't this be `#{Bundler.default_lockfile}`?

adarsh:
This seems like a positive change and good contribution, however Travis is failing.

lib/bundler/definition.rb
segiddins: Why rely on global state when we have the proper file name handy?

adarsh: Oh I see, this is the only place where we have `File` available so why not use it? I get that.

I lean slightly more on the side of consistency - if I were to see this break in the pattern, I'd wonder why, but I'm like a 2 out of 10 on this perspective.

@indirect thoughts? For me, this is ready to ship aside from this discussion.

segiddins:
Travis should pass now, will merge when green

lib/bundler/definition.rb
indirect: I'm fine with using the local when it's available—avoiding globals when possible is a slightly stronger preference for me than perfect consistency.

indirect:
:+1: :ship:

PR #3679, opened 2015-05-28
Run ``` bundle check ``` with ``` disable_shared_gems ``` set to '1' 

akihiro17:
Run ``` bundle check ``` with ``` disable_shared_gems ``` set to '1' if --path option is specified

### DIFF
--- a/lib/bundler/cli/check.rb
+++ b/lib/bundler/cli/check.rb
@@ -6,7 +6,10 @@ module Bundler
     end
 
     def run
-      Bundler.settings[:path] = File.expand_path(options[:path]) if options[:path]
+      if options[:path]
+        Bundler.settings[:path] = File.expand_path(options[:path])
+        Bundler.settings[:disable_shared_gems] = '1'
+      end
       begin
         definition = Bundler.definition
         definition.validate_ruby!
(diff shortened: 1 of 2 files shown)


akihiro17:
``` bundle check --path #{directory} ``` shows ``` The Gemfile's dependencies are satisfied ``` if gems are already installed in the default directory even though they are not installed in the --path directory.

Reproduction script for this issue.

```ruby
# -*- coding: utf-8 -*-
require 'tmpdir'
 
Dir.mktmpdir do |dir|
  Dir.chdir(dir) do
    File.write('Gemfile', <<-GEMFILE)
      source 'https://rubygems.org'
      gem 'activesupport'
    GEMFILE
 
    system 'bundle'
    system 'bundle check --path vendor/bundle'
    system 'ls -al vendor/bundle/ruby/*'
  end
end
```

The reason for this is probably that ``` disable_shared_gems ``` is not configured to '1' when executing ``` bundle check ``` with the --path option.
https://github.com/bundler/bundler/blob/master/lib/bundler/cli/check.rb#L9

So, this commit sets ``` disable_shared_gems```  to '1' and runs ``` bundle check ``` if --path option is specified.

indirect:
Yes, I believe you're right. Thanks!

PR #3689, opened 2015-05-29
Clarify BUNDLE_PATH documentation.

jrafanie:
My team recently ran into this and wanted to help clarify this.

I'm not intimately familiar with all the use cases for BUNDLE_PATH but I think I have the gist of what I needed clarified.

* BUNDLE_PATH is used for installing AND finding gems
* It's easy to confuse GEM_PATH and BUNDLE_PATH
 * GEM_PATH is an array of paths in addition to GEM_HOME that rubygems
   uses to find gems.
 * BUNDLE_PATH is a single path used for gem installs and retrieval.
 * GEM_PATH is ignored when you have BUNDLE_PATH set by enabling
   BUNDLE_DISABLE_SHARED_GEMS

### DIFF
--- a/man/bundle-config.ronn
+++ b/man/bundle-config.ronn
@@ -84,8 +84,10 @@ The following is a list of all configuration keys and their purpose. You can
 learn more about their operation in [bundle install(1)][bundle-install].
 
 * `path` (`BUNDLE_PATH`):
-  The location on disk to install gems. Defaults to `$GEM_HOME` in development
-  and `vendor/bundle` when `--deployment` is used
+  The location on disk where all gems in your bundle will be located regardless
+  of `$GEM_HOME` or `$GEM_PATH` values. Bundle gems not found in this location
+  will be installed by bundle install. Defaults to `Gem.dir`. When --deployment is
+  used, defaults to vendor/bundle.
 * `frozen` (`BUNDLE_FROZEN`):
   Disallow changes to the `Gemfile`. Defaults to `true` when `--deployment`
   is used.


jrafanie:
@indirect can you please review this when you get a chance,  thank you!

indirect:
I guess I'm a little bit confused about why GEM_PATH needs an explicit disclaimer—nothing in the documentation references it, and Bundler never looks at GEM_PATH under any circumstances.

It's probably most accurate to say "When the `path` option is set, Bundler will ignore GEM_HOME entirely."

jrafanie:
@indirect Thanks for reviewing this!  :smile:   

I appreciate your input since you know more about this area than I do and I'd like to get it right. 

I have pushed some edits with your comment in mind but I do believe GEM_PATH should be mentioned for the following reasons:

* bundler honors GEM_PATH when GEM_HOME is unset for normal gems (see my gist below)

* BUNDLE_PATH has a similar name to GEM_PATH but is a single path and solves a completely different use case.

* BUNDLE_PATH is more than just a [installation location](https://github.com/bundler/bundler/blob/0f93e4aa194892ceaedd9e1b2ff6976ac5add276/lib/bundler/source/git.rb#L96), the lookup path is an alias to the install path.

* I had no idea that BUNDLE_PATH disabled other lookup paths (GEM_HOME/GEM_PATH) via DISABLE_SHARED_GEMS see [1](https://github.com/bundler/bundler/blob/e8c962ef2a3215cdc6fd411b6724f091a16793d6/lib/bundler/cli/install.rb#L99) and [2](https://github.com/bundler/bundler/blob/e8c962ef2a3215cdc6fd411b6724f091a16793d6/lib/bundler.rb#L421-L422), nor did I realize that git based gems were treated differently than regular gems (see gist below).

https://gist.github.com/jrafanie/5424bc23a5c5fcbe5047

I can't really claim there's a bug here but I'm more than willing to help triage/fix what I'm seeing in that gist if you think so.  It would be nice if git based gems honored GEM_PATH like regular gems do when GEM_HOME is unset though but I didn't see any easy way to fix that.

Thanks again for your work on bundler... I remember the times before bundler and nitpicks like this are so minor compared to the problems before...  :bow:

indirect:
Interesting. I didn't actually realize that Bundler would honor GEM_PATH if GEM_HOME was not set. I guess that's because RubyGems honors it and sets `Gem.dir` based on it. I will definitely grant that BUNDLE_PATH and GEM_PATH are named similarly.

Unfortunately, it's completely impossible for git gems to honor GEM_PATH, because RubyGems itself does not support git gems. Bundler only honors GEM_PATH by accident, and we have never said that we support managing gems in your bundle via the `gem` command.

I'll add some notes to the word choice, but I think we're closing in on something that works for everyone.

man/bundle-config.ronn
```diff
@@ -84,8 +84,10 @@ The following is a list of all configuration keys and their purpose. You can
 learn more about their operation in [bundle install(1)][bundle-install].
 
 * `path` (`BUNDLE_PATH`):
-  The location on disk to install gems. Defaults to `$GEM_HOME` in development
-  and `vendor/bundle` when `--deployment` is used
+  The location on disk to install and locate gems. Defaults to `$GEM_HOME` in
+  development and `vendor/bundle` when `--deployment` is used. When the `path`
+  option is set, bundler will ignore `$GEM_HOME` and `$GEM_PATH` entirely and
+  require all gems be found in this location.
```
indirect: How about "The location on disk where all gems in your bundle will be located. Bundle gems not found in this location will be installed by `bundle install`. Defaults to `Gem.dir`. When `--deployment` is used, defaults to `vendor/bundle`."

jrafanie:
> Interesting. I didn't actually realize that Bundler would honor GEM_PATH if GEM_HOME was not set. I guess that's because RubyGems honors it and sets Gem.dir based on it. 

@indirect Yes, that was very confusing but it makes sense... clearly not something that was expected.

> I will definitely grant that BUNDLE_PATH and GEM_PATH are named similarly.

Yes, naming is hard.  In retrospect, I'd love to rename the current `BUNDLE_PATH` to something like `BUNDLE_VENDOR_PATH` since it's more for vendoring your bundle to one location.  

> Unfortunately, it's completely impossible for git gems to honor GEM_PATH, because RubyGems itself does not support git gems. Bundler only honors GEM_PATH by accident, and we have never said that we support managing gems in your bundle via the gem command.

I was looking for something in bundler that is like `GEM_PATH` for bundler (an additional location to search or both git and regular gems).  This is why I confused `BUNDLE_PATH` and `GEM_PATH` to begin with.  I don't see an easy way of doing this since so much code expects the git based installation path and lookup path to be the same thing (via the alias), so I'm content with just clarifying the documentation for now.

Thank you for helping to clarify this.   I thought your wording was closer but still didn't clarify the issue I ran into originally: `$GEM_HOME` and `$GEM_PATH` values are completely ignored when `path` is set.  I hope my small addition gets us closer.

Thanks again for your review!

indirect:
Looks good, thanks!

jrafanie:
Great, thank you! :shipit:

PR #3694, opened 2015-05-30
[EndpointSpecification] Return local load paths

### LINKED ISSUE #3693: Bundler inline cannot require if installing gem at the same time
When trying to use the example provided by the [inline feature](https://github.com/bundler/bundler/pull/3440), changing the `gemfile` line to indicate that the gems should be installed:

```
#!/usr/bin/env ruby

require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'
  gem 'json', require: false
  gem 'nap', require: 'rest'
  gem 'cocoapods', '~> 0.34.1'
end

puts Pod::VERSION # => "0.34.4"
```

I get the following at the first time I run it:

```
bundler-1.10.2/lib/bundler/runtime.rb:76:in `require': cannot load such file -- rest (LoadError)
  from bundler-1.10.2/lib/bundler/runtime.rb:76:in `block (2 levels) in require'
  from bundler-1.10.2/lib/bundler/runtime.rb:72:in `each'
  from bundler-1.10.2/lib/bundler/runtime.rb:72:in `block in require'
  from bundler-1.10.2/lib/bundler/runtime.rb:61:in `each'
  from bundler-1.10.2/lib/bundler/runtime.rb:61:in `require'
  from bundler-1.10.2/lib/bundler/inline.rb:55:in `gemfile'
  from example.rb:5:in `<main>'
```

After the first time, when the gems are already installed, everything goes fine and I get the expected output `0.34.4`.

I have:
* Bundler version 1.10.2
* Rubygems 2.4.7
* ruby 2.2.2p95 (2015-04-13 revision 50295) [x86_64-darwin14] [MRI]
* rvm 1.26.11

Let me know if you need more information. If it is a real issue, I'd love to try working on it.

segiddins:
Needed for when inline has installed from this spec

Closes #3693.

### DIFF
--- a/lib/bundler/endpoint_specification.rb
+++ b/lib/bundler/endpoint_specification.rb
@@ -29,17 +29,6 @@ module Bundler
       end
     end
 
-    # needed for inline
-    def load_paths
-      if @remote_specification
-        @remote_specification.load_paths
-      elsif _local_specification
-        _local_specification.load_paths
-      else
-        super
-      end
-    end
-
     # needed for binstubs
     def executables
       if @remote_specification
--- a/lib/bundler/endpoint_specification.rb
+++ b/lib/bundler/endpoint_specification.rb
@@ -29,6 +29,16 @@ module Bundler
       end
     end
 
+    # needed for inline
+    def load_paths
+      # remote specs aren't installed, and can't have load_paths
+      if _local_specification
+        _local_specification.load_paths
+      else
+        super
+      end
+    end
+
     # needed for binstubs
     def executables
       if @remote_specification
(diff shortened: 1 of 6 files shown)


indirect:
any way to add this to the inline tests?

segiddins:
Well, it requires installing the spec specifically from the v1 dependency API that isn't available locally.

indirect:
we can do that is specs, using artifice

segiddins:
I can't reproduce those spec failures locally, would appreciate some help on that

segiddins:
@indirect your fix doesn't appear to have worked

indirect:
yeah, working on it :/

segiddins:
@indirect need to update the specs for not validating on < 1.7 now

segiddins:
@indirect if you're fine with that spec change, let's :shipit:

indirect:
:ship: :ship: :ship:

Issue #3671, opened 2015-05-21
Bundler issues warning when a gem is on both the global source and a one-off source


cowboyd:
I have a Gemfile that looks like this:

```ruby
source "https://rubygems.org"
gem "rails"
gem 'capistrano-graphite', '1.0.5.spiceworks.1', require: false, source: "http://gem-server.spice.spiceworks.com"
```

The custom source `gem-server.spice.spiceworks.com` contains both the `capistrano-graphite` gem as well as `rails`.

When I run bundle install, it succeeds, but indicates several warnings about some of the gems being on both sources:

```sh
/C/S/tmp ❯❯❯ bundle install
Fetching gem metadata from http://gem-server.spice.spiceworks.com/...
Fetching gem metadata from https://rubygems.org/......
Fetching version metadata from http://gem-server.spice.spiceworks.com/...
Fetching version metadata from https://rubygems.org/..
Fetching gem metadata from http://gem-server.spice.spiceworks.com/...
Fetching version metadata from http://gem-server.spice.spiceworks.com/...
Fetching dependency metadata from http://gem-server.spice.spiceworks.com/..
Using i18n 0.7.0
Using builder 3.2.2
Using erubis 2.7.0
Using mini_portile 0.6.2
Using rack 1.6.1
Using mime-types 2.5
Using arel 6.0.0
Using colorize 0.7.7
Using net-ssh 2.9.2
Using bundler 1.9.9
Using thor 0.19.1
Using json 1.8.2
Using rake 10.4.2
Using thread_safe 0.3.5
Using minitest 5.6.1
Using nokogiri 1.6.6.2
Using rack-test 0.6.3
Using sprockets 3.1.0
Using mail 2.6.3
Using net-scp 1.2.1
Using tzinfo 1.2.2
Using loofah 2.0.2
Using sshkit 1.7.1
Using rails-html-sanitizer 1.0.2
Using capistrano 3.4.0
Installing capistrano-graphite 1.0.5.spiceworks.1
Installing activesupport 4.2.1
Using rails-deprecated_sanitizer 1.0.3
Using globalid 0.3.5
Using activemodel 4.2.1
Using rails-dom-testing 1.0.6
Using activejob 4.2.1
Using actionview 4.2.1
Installing actionpack 4.2.1
Using actionmailer 4.2.1
Using railties 4.2.1
Using sprockets-rails 2.3.1
Installing activerecord 4.2.1
Installing rails 4.2.1
[...]

PR #3696, opened 2015-05-31
Failing spec for #3671

TimMoore:


segiddins:
Did we ever make any progress on fixing this?

indirect:
Not yet...

TimMoore:
I made a failed attempt a while back... it's a bug whack-a-mole. I managed to fix this spec but broke several others :sob:

spec/install/gems/sources_spec.rb
```diff
@@ -266,6 +266,79 @@
...
+          end
+
+          update_repo gem_repo3 do
+            build_gem "rack", "1.0.0" do |s|
+              s.write "lib/rack.rb", "RACK = 'FAIL'"
+            end
+          end
+        end
+
+        it "installs the dependency from the top-level source without warning" do
```
segiddins: @TimMoore shouldn't it install from the pinned source here rather than the top-level source?

TimMoore: Not in this case, I think, because the thing that depends on it comes from the top-level source.

but then I don't know what should happen when you have multiple gems from multiple different sources that depend on a third gem

segiddins: yup, ended up getting it working :D

segiddins:
This PR will be auto-closed once 2-0-dev is merged to master

Issue #3697, opened 2015-05-31
Do not update BUNDLED WITH if no other changes to the lock happened


jhass:
Running `bundle install` (e.g. as a dev on a project) and having a newer version of bundler, writes that newer version into the lock. This makes the lock dirty, leaving you with the option to 1) make a meaningless commit to bump the bundler version referenced in it or 2) constantly run `git checkout Gemfile.lock` as you switch branches, rebase and so on.

If there were no other changes to the lock, we know that the previous version of Bundler was not only working for the lock and the corresponding Gemfile, in fact we know that there's very likely not even the slightest difference in behaviour for it.

I'm happy to look into a fix if you have some pointers, like where one could decide to not write the lock at all if it wasn't changed.

ashchan:
This is also annoying when you have v1.10, and other team members still have 1.9.

segiddins:
Bundler will only ever increase that version, not decrease it. The reason we need to do that, even when there are seemingly no other changes, is to know what version you, the developer, have verified "works" with your project.

jhass:
Can I have "please do a commit to all your projects on every bundler minor release" as an official statement clearly spelled out then?

segiddins:
Well, it's of course up to you if you want to commit those changes or not :)

-Samuel E. Giddins

On May 31, 2015, at 9:56 AM, Jonne Haß <notifications@github.com> wrote:

Can I have "please do a commit to all your projects on every bundler minor release" as an official statement clearly spelled out then?

jhass:
Well, variant 2 that I described in my initial issue is an even bigger pain in the ass.

jhass:
I'm just asking what the officially endorsed workflow is here. Also keep in mind that things like spring constantly write the lock on changes (I didn't investigate why, I just see it very clearly happening).

SuperTux88:
This is no problem, when there is only one developer ... the Gemfile.lock always contains the version from this developer.

But this is a really big issue, when there are multiple developers ... when in the Gemfile.lock on the main-branch is e.g. 1.10.0 and one developer has a newer version (e.g. 1.10.2) then he has always an outgoing change on the Gemfile.lock. But he can't commit the Gemfile.lock to his bugfix/feature-branch, because it has nothing to do with the bug/feature he is developing.

So someone needs to commit a new Gemfile.lock to the main-branch every time a new bundler version is released? Is this the intended workflow?

indirect:
[...]

PR #3722, opened 2015-06-05
Implicit lock preservation

indirect:
When either `bundle check` is run, or any application requires the `bundler/setup` file, Bundler will automatically check whether it is possible to lock the Bundle. During the lock process, Bundler updates the lock if the implicit locking changes the lock file.

Starting with the 1.10 release, Bundler includes a lockfile section named BUNDLED WITH that includes the version of Bundler that generated the lockfile. In order to minimize git churn, and guarantee that the lockfile will only be changed when the user runs an explicit Bundler command, Bundler will now only add or update the BUNDLED WITH section during commands where the user asks for changes to the lock. This includes, but is not limited to, `install`, `update`, `lock`, and `package`.

Running the `check` command or loading an application that uses Bundler will still now add or update the BUNDLED WITH section if, and only if, the lockfile has also changed for a different reason (such as a gem being updated).

Simply using an application, by running `bundle exec` commands or by running `bin/rails` and the like, will not change the lockfile. As a result, the intended workflow with the BUNDLED WITH section is now slightly different than it was before:

1. When running `bundle install`, Bundler will update the version in the lockfile if newer than the version present.

2. Then, check in the lockfile change, exactly as you would after running install to change any other gem version.

3. Older versions of Bundler will not change the number in the lock, but will warn the user that they are out of date.

refs bundler/bundler-features#80
refs #3697

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -240,14 +240,14 @@ module Bundler
       dependencies.map { |d| d.groups }.flatten.uniq
     end
 
-    def lock(file)
+    def lock(file, preserve_bundled_with = false)
       contents = to_lock
 
       # Convert to \r\n if the existing lock has them
       # i.e., Windows with `git config core.autocrlf=true`
       contents.gsub!(/\n/, "\r\n") if @lockfile_contents.match("\r\n")
 
-      return if @lockfile_contents == contents
+      return if lockfiles_equal?(@lockfile_contents, contents, preserve_bundled_with)
 
       if Bundler.settings[:frozen]
         Bundler.ui.error "Cannot write a changed lockfile while frozen."
@@ -658,5 +658,15 @@ module Bundler
     def requested_groups
       self.groups - Bundler.settings.without - @optional_groups + Bundler.settings.with
     end
+
+    def lockfiles_equal?(current, proposed, preserve_bundled_with)
+      if preserve_bundled_with
+        pattern = /\n\nBUNDLED WITH\n.*\n/
+        current.sub(pattern, "\n") == proposed.sub(pattern, "\n")
+      else
+        current == proposed
+      end
+    end
+
   end
 end
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -1,3 +1,4 @@
+require "bundler/lockfile_parser"
 require "digest/sha1"
 require "set"
 
@@ -661,7 +662,7 @@ module Bundler
 
     def lockfiles_equal?(current, proposed, preserve_bundled_with)
       if preserve_bundled_with
-        pattern = /\n\nBUNDLED WITH\n.*\n/
+        pattern = /\n\n#{LockfileParser::BUNDLED}\n\s+#{Gem::Version::VERSION_PATTERN}\n/
         current.sub(pattern, "\n") == proposed.sub(pattern, "\n")
       else
         current == proposed
--- a/spec/commands/check_spec.rb
+++ b/spec/commands/check_spec.rb
@@ -314,9 +314,11 @@ describe "bundle check" do
     end
 
     context "is newer" do
-      it "does not change the lock" do
+      it "does not change the lock but warns" do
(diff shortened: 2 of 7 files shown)


andremedeiros:
LGTM. :ship:

lib/bundler/definition.rb
```diff
@@ -658,5 +658,15 @@ def pinned_spec_names(specs)
     def requested_groups
       self.groups - Bundler.settings.without - @optional_groups + Bundler.settings.with
     end
+
+    def lockfiles_equal?(current, proposed, preserve_bundled_with)
+      if preserve_bundled_with
+        pattern = /\n\nBUNDLED WITH\n.*\n/
```
TimMoore: Might be better as a constant.

TimMoore:
Makes sense :+1:

lib/bundler/definition.rb
segiddins: use the parser constant

spec/commands/check_spec.rb
```diff
@@ -275,4 +275,59 @@
...
+        bundle :check
+        lockfile_should_be lock_with(nil)
+      end
+    end
+
+    context "is newer" do
+      it "does not change the lock" do
+        lockfile lock_with(Bundler::VERSION.succ)
+        bundle :check
+        lockfile_should_be lock_with(Bundler::VERSION.succ)
```
segiddins: please verify the warning is still printed

spec/other/cli_dispatch_spec.rb
```diff
@@ -14,7 +14,7 @@
   end
 
   it "print a friendly error when ambiguous" do
-    bundle "i"
+    bundle "in"
```
segiddins: why is this change necessary?

andremedeiros: Because "bundle i " is now mapped to install

spec/runtime/setup_spec.rb
```diff
@@ -899,4 +899,58 @@ def Bundler.require(path)
...
+        ruby "require 'bundler/setup'"
+        lockfile_should_be lock_with(nil)
+      end
+    end
+
+    context "is newer" do
+      it "does not change the lock" do
+        lockfile lock_with(Bundler::VERSION.succ)
+        ruby "require 'bundler/setup'"
+        lockfile_should_be lock_with(Bundler::VERSION.succ)
```
segiddins: check the warning is printed

spec/other/cli_dispatch_spec.rb
TimMoore: I'm not sure why it's on this branch, though. I committed it directly to `1-10-stable`. @andremedeiros looks like you cherry-picked the commit?

indirect: I'll just rebase. Please stand by...

spec/runtime/setup_spec.rb
indirect: I don't think the warning should be printed. This is the runtime; the user hasn't run a bundler command, bundler is being loaded potentially without the end user's knowledge. I think we should limit warnings to times when the user knows they are interacting with Bundler.

spec/commands/check_spec.rb
indirect: It does at least make sense to me to print the warning here, but I feel like I would rather not warn on `bundle check` than add a third special case to the BUNDLED WITH code :/

spec/runtime/setup_spec.rb
segiddins: either way, there should be specs around the warning being printed or not being printed

parndt:
Thanks, @indirect !

spec/commands/check_spec.rb
indirect: Ironic after my last commit, but the existing behavior is my preferred one: the `check` command warns, but `bundler/setup` does not.

indirect:
@segiddins there's your specs

indirect:
Specs are green, so I'm merging this. Thanks for the review, @segiddins!

PR #3723, opened 2015-06-05
Clarify the purpose of --deployment

mattbrictson:
The man page originally explained that `--deployment` is not appropriate for the `test` environment, which is misleading. In a CI environment, you *do* want to use `--deployment`, despite the fact that CI will likely be using `RAILS_ENV=test` when running tests.

This PR clarifies that `--deployment` is for production and CI, and avoids explicitly referencing Rails environment names.


indirect:
:+1: thanks!

PR #3724, opened 2015-06-05
Remove Gemfile.lock updates outside of bundle explicit commands

hartator:
Thanks a lot for awesome work to each of you!

I feel it's counterintuitive to update the `Gemfile.lock` outside of explicit `bundle` commands like `install`, `update` or `lock`. For instance, you don't really expect `bundle exec rails server` or just `require "bundler/setup"` to actually change your lock file. I feel it breaks in a way the principle of least surprise.

Even if I don't really get the use cases of having the lock updated at runtime, I feel it might be edge cases that can be moved to an opt in (like `:force_lock_generation`) whereas now everyone has their `Gemfile.lock` regenerated.

It might also improve performance a bit specially in tools like Spring or Zeus that's continuously reloading rails environment. 

Sorry, I might be missing something obvious.

References:
- https://github.com/bundler/bundler/pull/3722
- https://github.com/bundler/bundler/issues/3697
- https://github.com/rails/spring/issues/409

### DIFF
--- a/spec/commands/check_spec.rb
+++ b/spec/commands/check_spec.rb
@@ -23,7 +23,7 @@ describe "bundle check" do
     expect(out).to eq("The Gemfile's dependencies are satisfied")
   end
 
-  it "creates a Gemfile.lock by default if one does not exist" do
+  it "does not creates a Gemfile.lock by default if one does not exist" do
     install_gemfile <<-G
       source "file://#{gem_repo1}"
       gem "rails"
@@ -33,7 +33,7 @@ describe "bundle check" do
 
     bundle "check"
 
-    expect(bundled_app("Gemfile.lock")).to exist
+    expect(bundled_app("Gemfile.lock")).not_to exist
   end
 
   it "does not create a Gemfile.lock if --dry-run was passed" do
--- a/spec/install/gemfile/git_spec.rb
+++ b/spec/install/gemfile/git_spec.rb
@@ -214,7 +214,7 @@ describe "bundle install with git sources" do
       expect(out).to eq("LOCAL")
     end
 
-    it "updates specs on runtime" do
+    it "doesn't update specs on runtime" do
       system_gems "nokogiri-1.4.2"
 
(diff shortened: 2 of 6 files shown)


segiddins:
This seems like a duplicate of https://github.com/bundler/bundler/pull/3722?

spec/commands/check_spec.rb
```diff
@@ -33,7 +33,7 @@
 
     bundle "check"
 
-    expect(bundled_app("Gemfile.lock")).to exist
+    expect(bundled_app("Gemfile.lock")).not_to exist
```
segiddins: this is a backwards-incompatible change

spec/install/gemfile/git_spec.rb
```diff
@@ -235,7 +235,7 @@
       run "require 'rack'"
 
       lockfile1 = File.read(bundled_app("Gemfile.lock"))
-      expect(lockfile1).not_to eq(lockfile0)
+      expect(lockfile1).to eq(lockfile0)
```
segiddins: this is also backwards-incompatible

segiddins:
Unfortunately, this change is backwards incompatible, so we'll have to go with @indirect's PR

hartator:
#3722 is perfectly fine. 

I am more questioning why are we regenerating the `Gemfile.lock` outside of Bundle commands and the issues fixed by #3722 were un-directly related.

indirect:
@hartator we're doing it because it caused an unbelievable support burden. In previous versions of Bundler, we required users to run `bundle install` after changing the Gemfile, even when the gems were already installed. There were dozens of tickets reporting it as a bug.

Bundler users expect to be able to change their Gemfile, run their application, and have it just work as long as the needed gems are already installed.

hartator:
Thanks Indirect for the explanation. Alright, that's weird expectation, maybe they were confused by the `bundle install` command name that's suggesting installation but not lock. `bundle bundlify` ?

PR #3731, opened 2015-06-08
Added optional remote argument for the rake release task

skateman:
Useful when working with multiple remotes (origin/upstream/downstream).
`rake release[remote]` will push tags into the specified remote instead of the default one.


indirect:
This seems good, thanks!

Issue #3732, opened 2015-06-09
Bundler 1.10.3 resolves dependencies indefinitely.


adammathys:
I'm currently unable to use Bundler 1.10 on most projects due to it taking forever to resolve dependencies.

Output of `bundle env` (with some minor data scrubbing):
```shell
Environment

    Bundler   1.10.3
    Rubygems  2.4.8
    Ruby      2.1.6p336 (2015-04-13 revision 50298) [x86_64-linux]
    GEM_HOME  /home/user/.gem/ruby/2.1.6
    GEM_PATH  /home/user/.gem/ruby/2.1.6:/home/user/.rubies/ruby-2.1.6/lib/ruby/gems/2.1.0
    Git       1.9.1

Bundler settings

    path
      Set for the current user (/home/user/.bundle/config): "vendor/bundle"
    disable_shared_gems
      Set for your local app (/home/user/Work/project/.bundle/config): "1"
      Set for the current user (/home/user/.bundle/config): "1"

Gemfile

    source 'https://rubygems.org'
    
    # Provides basic authentication functionality for testing parts of your engine
    gem 'spree', github: 'spree/spree', branch: '2-2-stable'
    gem 'spree_auth_devise', github: 'spree/spree_auth_devise', branch: '2-2-stable'
    gem 'byebug'
    
    gemspec

Gemfile.lock

    <No /home/user/Work/project/Gemfile.lock found>
```

Result of running `bundle install` for ~11 minutes:
```shell
Fetching git://github.com/spree/spree.git
Fetching git://github.com/spree/spree_auth_devise.git
Fetching gem metadata from https://rubygems.org/.............
Fetching version metadata from https://rubygems.org/...
Fetching dependency metadata from https://rubygems.org/..
[...]

PR #3736, opened 2015-06-10
[Resolver] Add optimization for deps where theres a path/gemspec source

segiddins:
Closes https://github.com/bundler/bundler/issues/3732.

### DIFF
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -310,12 +310,12 @@ module Bundler
         else
           base_dep = Dependency.new dependency.name, '>= 0.a'
           all = search_for(DepProxy.new base_dep, dependency.__platform).size.to_f
-          search = search_for(dependency).size.to_f
           if all.zero?
             0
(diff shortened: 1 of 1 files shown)


lib/bundler/resolver.rb
```diff
@@ -309,11 +309,14 @@ def amount_constrained(dependency)
           dependency.requirement.satisfied_by?(base.first.version) ? 0 : 1
         else
           base_dep = Dependency.new dependency.name, '>= 0.a'
-          all = search_for(DepProxy.new base_dep, dependency.__platform)
-          if all.size == 0
+          all = search_for(DepProxy.new base_dep, dependency.__platform).size.to_f
+          search = search_for(dependency).size.to_f
+          if all.zero?
```
indirect: can't we skip the work of `search_for` in this case?

segiddins: I suppose we could!

segiddins:
merging presuming travis passes

Issue #3715, opened 2015-06-04
`install --force` should delete existing gems before installing


indirect:
> @segiddins that's a bug in force, then. we need to nuke the directory, since it's possible that a malicious gem could install files that don't get overwritten when the non-malicious gem is installed. :/

karlo57:
@indirect Can you provide more details? I'd like to help fix this bug. Thanks!

indirect:
@karlo57 sure, no problem.

Right now, when you run `bundle install --force`, we download each .gem file again (even if we already have a copy downloaded), and we install each gem again (even if that gem is already installed). The additional thing we need to do is _uninstall_ each gem before we download and install it again. Or if not uninstall, at least delete the directory containing the installed/unpacked files, to guarantee that the only files that are left afterward came from the gem that was just installed with `--force`.

karlo57:
@indirect  I'm deleting the bundle path on first run of the install process if the `--force` flag is ticked. Is this correct? And should I also delete the user's directories for instances that the user has defined a different install path?

Thanks!

https://github.com/karlo57/bundler/commit/1f1688ae8440827342f0dd54aaf1e5a7ac104470

karlo57:
I think we should close this now? I'll create a new issue for the new bug found out.

agis-:
@karlo57 Resolved by which commit?

agis-:
@indirect Should we close this now that https://github.com/bundler/bundler/commit/13c445680f65293ee50f11718e8a86bfa87abd2e is merged? It's not clear if this is an issue fixed by that commit or it's a non-issue.

indirect:
I guess it was a non-issue? Either way, seems resolved by a test now.

PR #3743, opened 2015-06-15
Delete bundle path if using bundle install --force

karlo57:
#3715

EDIT: #3715 doesn't apply anymore. But a new bug was found regarding `bundle install --force` reinstalling bundler. This pull request is for that.

### DIFF
--- a/spec/install/force_spec.rb
+++ b/spec/install/force_spec.rb
@@ -33,5 +33,14 @@ describe "bundle install" do
       expect(out).to include "Using bundler 1.10.4"
     end
 
+    it "works on first bundle install" do
+      bundle "install --force"
+
+      expect(out).to include "Installing rack 1.0.0"
+      should_be_installed "rack 1.0.0"
+      expect(exitstatus).to eq(0) if exitstatus
+    end
+
+
   end
 end
--- a/spec/install/force_spec.rb
+++ b/spec/install/force_spec.rb
@@ -7,14 +7,15 @@ describe "bundle install" do
         source "file://#{gem_repo1}"
         gem "rack"
       G
-
-      bundle "install"
     end
 
     it "re-installs installed gems" do
+      bundle "install"
       bundle "install --force"
+
       expect(out).to include "Installing rack 1.0.0"
       should_be_installed "rack 1.0.0"
+      expect(exitstatus).to eq(0) if exitstatus
     end
   end
 end
--- a/spec/install/force_spec.rb
+++ b/spec/install/force_spec.rb
@@ -17,5 +17,13 @@ describe "bundle install" do
       should_be_installed "rack 1.0.0"
       expect(exitstatus).to eq(0) if exitstatus
     end
+
+    it "doesn't reinstall bundler" do
+      bundle "install"
+      bundle "install --force"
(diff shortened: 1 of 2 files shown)


indirect:
Well, at a minimum this needs tests. :) This approach seems like it will probably work, though.

segiddins:
This seems way overly-aggressive to me? We should only be cleaning up after the gems we're re-installing

indirect:
If it's a bundle install --path --force it seems fine? But if it's --system --force, yes, this destroys everything. :P

karlo57:
I'll add tests and look into cleaning up only the gems we're installing. Thanks guys!

karlo57:
@indirect @segiddins I updated this to clean up after the gems we're only installing. I'll add tests tomorrow.

spec/install/force_spec.rb
```diff
@@ -2,19 +2,41 @@
...
+    end
+
+    # How do I test this
+    it "deletes installed gems" do
+      gemfile <<-G
+        source "file://#{gem_repo1}"
+        gem "rack"
+      G
       bundle "install"
+
```
indirect: On this line, add a file into the directory where `rack` is installed. IIRC, that's something like `default_bundle_path("gems/rack-1.0.0/garbage").open{|f| f.write("blah blah blah") }`.

spec/install/force_spec.rb
```diff
@@ -2,19 +2,41 @@
...
+      bundle "install --force"
+
```
indirect: On this line, check to make sure the file you added no longer exists. For example, `expect(default_bundle_path("gems/rack-1.0.0/garbage")).to_not exist`.

indirect:
See the inline comments for testing suggestions. :)

karlo57:
I added the lines to the test. It works. Although now I'm seeing that this is a non-issue? I commented the  lines which uninstall the gems and the spec still passes. I also have verified that the garbage file is actually written so there's nothing wrong with the tests.

indirect:
Well, I guess I was right about there being a bug, but only by accident! Heh. Thanks for writing these tests, they show the real bug!

```
Installing bundler 1.10.4

Bundler::GemNotFound: Could not find bundler-1.10.4.gem for installation
Installing rack 1.0.0
An error occurred while installing bundler (1.10.4), and Bundler cannot continue.
Make sure that `gem install bundler -v '1.10.4'` succeeds before bundling.
```

indirect:
I wound up making these changes to show that problem:

```
Installing bundler 1.10.4

Bundler::GemNotFound: Could not find bundler-1.10.4.gem for installation
Installing rack 1.0.0
An error occurred while installing bundler (1.10.4), and Bundler cannot continue.
Make sure that `gem install bundler -v '1.10.4'` succeeds before bundling.
```

indirect:
Looks like we don't need to actively uninstall, but we definitely need these tests, and we need to not force-install Bundler itself.

karlo57:
Cool, I updated the code now. Please review. Thanks!

indirect:
Merged as 13c445680f65293ee50f11718e8a86bfa87abd2e. Thanks!

PR #3744, opened 2015-06-16
Don't Autoremember Flags in 2.0

pducks32:


lib/bundler/settings.rb
```diff
@@ -256,5 +280,11 @@ def normalize_uri(uri)
       uri
     end
 
+    def print_remembered_flags_deprecation_warning
+      Bundler.ui.warn "Starting in Bundler 2.0, flags passed into commands will no longer be remember past that command." \
+        " Instead please set flags you want remembered between commands using `bundle config <setting name> <setting value>`" \
+        " To ignore this warning use `bundle config ignore_warnings.remembered_flags true`"
+      set_current "ignore_warnings.remembered_flags", true
```
segiddins: why this change?

lib/bundler/settings.rb
```diff
@@ -86,6 +105,7 @@ def gem_mirrors
     def locations(key)
       key = key_for(key)
       locations = {}
+      location[:current] = @current_config[key] if @current_config.key?(key)
```
segiddins: `locations`?

spec/commands/check_spec.rb
```diff
@@ -242,7 +242,7 @@
     it "returns success" do
       bundle "check --path vendor/bundle"
       expect(exitstatus).to eq(0) if exitstatus
-      expect(out).to eq("The Gemfile's dependencies are satisfied")
+      expect(out).to include("The Gemfile's dependencies are satisfied")
```
pducks32: Because I know someone might ask, I noticed this while testing and thought that the test is the one that should be fixed, not the code. Checking that output equals something is overly rigid, we should only be checking that it includes the text—i.e. the task was done.

lib/bundler/settings.rb
pducks32: which change? The `set_current`? so that it doesn't show up every time `settings[]=` is called.

lib/bundler/settings.rb
pducks32: Ugh thanks for picking that up. As my friend told me: "You really wouldn't believe English is your first language."

smlance:
Shouldn't we merge this into 2-0-dev?

indirect:
Yeah. If you think it's ready, you can do that yourself using the `hub` command (`brew install hub`):

```bash
$ git checkout 2-0-dev
$ hub merge https://github.com/bundler/bundler/pull/3744
$ git push
```

That will notify this issue, and then we can close the PR.

lib/bundler/settings.rb
smlance: Considering https://github.com/bundler/bundler/pull/3590/files, is this change necessary? If so, shouldn't it be printed as a deprecation warning in 1-99-dev rather than as a ui.warn in 2-0-dev?

indirect: Yeah, we should warn in 1-99 and just not remember flags in 2-0. I don't think @pducks32 knew that 1-99 existed when he wrote this PR, though...

smlance:
https://github.com/bundler/bundler/commit/4d5d6f4865213c67fc06b36b715ecb131f259ae5

smlance:
I'm closing this since I cherry-picked the commits into 2-0-dev: https://github.com/bundler/bundler/commit/3d76c8fefb4eed7a082719e693396b7cdb0d58b0. Thanks!

lib/bundler/settings.rb
```diff
@@ -26,13 +27,31 @@ def [](name)
       end
     end
 
+    # In Bundler pre 2.0 any command ever set was stored in the local config
+    # file. In Bundler 2.0 they are only saved for that current command and
+    # `bundle config <setting name> <setting value>` is used for setting
+    # remembered arguments.
+    # See https://trello.com/c/yGsPNDpg
     def []=(key, value)
+      @use_current ||= (Bundler::VERSION.split(".")[0].to_i >= 2)
```
smlance: Because `Bundler::VERSION` is less than 2 (https://github.com/pducks32/bundler/blob/8a8b1513e4c1651f39c5361e1d5d93a41411a202/lib/bundler/version.rb#L5), the behavior of bundler isn't changed by this PR.

pducks32: As smlance and I talked about. The issue is that there are tons of instances where specs or lib assume auto remembered flags. So I'm going to go through and refactor them all to use the new style.

PR #3745, opened 2015-06-16
Fix Version Matchers

pducks32:
Matchers were set up so they had to be updated everytime one of the gems (rails in this case) updated.


segiddins:
I think for these tests, we'd rather know about changes so we can verify their correctness manually.

pducks32:
It just seems weird to be specing whether or not Rubygems version resolution works, but I see your point.

segiddins:
It's bundler resolution....

pducks32:
oh right.

PR #3749, opened 2015-06-17
Improve error message for circular dependency case

### LINKED ISSUE #3738: "gems that depend on each other" message is incorrect
I'm running bundler 1.10.3
when I bundle update I get this message

> Your Gemfile requires gems that depend on each other, creating an infinite loop. Please remove either gem 'linkeddata' or gem 'rdf-vocab' and try again.

However, only the `rdf-vocab` gem is in my `Gemfile`.  linkeddata is in my `Gemfile.lock` though.

asymmetric:
Makes it more clear that the circular dependent gems are not necessarily declared in the `Gemfile` (they could be in the `Gemfile.lock`).

Closes #3738.


asymmetric:
FYI, tests fail because of a spec trying to match against 3.2.21, while 3.2.22 has been released.

indirect:
Thanks!

jcoyne:
:clap:

PR #3750, opened 2015-06-17
update test to check for rails-3.2.22

asymmetric:
Since Rails 3.2.22 has been released, [this](https://github.com/bundler/bundler/blob/master/spec/realworld/edgecases_spec.rb#L39) test has been broken, since it matched against 3.2.21.

Now we match against 3.2.22


segiddins:
This was already merged into 1-10-stable

Issue #3761, opened 2015-06-18
Bundler doesn't respect `bundle update --source ...` when gem and git repo differ in name


neoeno:
This is related to and explained by #3759.

Let's say I have a Gemfile

```ruby
source 'https://rubygems.org'
gem 'rubocop', github: 'neoeno/rubocop_renamed'
```

And I've got a previously generated Gemfile.lock. 

But the repository has just been updated, so I run:

```shell
$ bundle update --source rubocop_renamed
# Look here for how the source name is inferred: https://github.com/bundler/bundler/blob/master/lib/bundler/source/git.rb#L76-L78
```

I get the following output:

```
Updating git://github.com/neoeno/rubocop_renamed.git
Fetching gem metadata from https://rubygems.org/.........
Fetching version metadata from https://rubygems.org/..
Resolving dependencies...
Could not find rubocop-0.32.0 in any of the sources
```

Since I'm updating the source, I'd expect the source and all its gems to be unlocked and thereby updated.

There's an example in http://github.com/neoeno/bundle_test

rgb-one:
I can reproduce this issue with the `Gemfiile.lock` provided in the `bundle_test` repository, however, when I delete that `Gemfile.lock` and run `bundle update --source rubocop_renamed`, it works fine updating rubocop to version 0.32.0.2.

When did you clone the rubocop repository? It could be that because you cloned after the `revision: 2c669b384421bb163aa6a58b045d2eb256ef295b` and also renamed your clone of the repository to `rubocop_renamed`, it is causing this issue with rubocop-0.32.0 not being found because it is looking for a point where, in the `rubocop_renamed` repository that specific commit was made which doesn't not exist. This is just an assumption.

neoeno:
That's a valid possibility — worth investigating. Here's what I did to investigate:

I pulled down the repository, then the following:

1. `bundle update`
2. Pushed a new version bump commit to `rubocop_renamed` (https://github.com/neoeno/rubocop_renamed/commit/2bf754be48cdeddad6d66585bec689a3e235ea2b)
3. `bundle update --source rubocop_renamed`

Log:

```shell
$ bundle update
Updating git://github.com/neoeno/rubocop_renamed.git
Fetching gem metadata from https://rubygems.org/.........
Fetching version metadata from https://rubygems.org/..
Resolving dependencies...
Using ast 2.0.0
Using parser 2.2.2.5
Using astrolabe 1.3.0
Using powerpack 0.1.1
[...]

PR #3763, opened 2015-06-18
Fix `--source` option causing incorrect gem unlocking

### LINKED ISSUE #3759: `bundle update --source gem` updates conservatively
As noted in https://github.com/bundler/bundler/issues/2016, using `bundle update --source gem` appears to do the same as `bundle update gem` while refusing to unlock any of the gems' dependencies. It's becoming a more popular technique.

Some people consider this useful behaviour, but the docs don't seem to indicate that this is intended behaviour. The docs say:

```
––source=<name>

The name of a :git or :path source used in the Gemfile(5). For instance, with a :git source of http://github.com/rails/rails.git, you would call bundle update ––source rails
```

This would appear to be intended to allow git sources to be unlocked. I reason this way because of the [comment and code in definition.rb](https://github.com/bundler/bundler/blob/master/lib/bundler/definition.rb#L500-L510):

```ruby
gemfile_sources.each do |source|
  # If the source is unlockable and the current command allows an unlock of
  # the source (for example, you are doing a `bundle update <foo>` of a git-pinned
  # gem), unlock it. For git sources, this means to unlock the revision, which
  # will cause the `ref` used to be the most recent for the branch (or master) if
  # an explicit `ref` is not used.
  if source.respond_to?(:unlock!) && @unlock[:sources].include?(source.name)
    source.unlock!
    changes = true
  end
end
```

Which is fine. The question is — what does this code do when `--source` is passed a rubygem rather than a git gem. It does nothing — because `lib/bundler/source/rubygems.rb` doesn't have `#unlock!`.

So, by that reasoning, `bundle update --source rubygem_gem` should be a no-op. But it isn't.

So, I found [definition.rb line 529 in #converge_locked_specs](https://github.com/bundler/bundler/blob/master/lib/bundler/definition.rb#L558):

```ruby
@locked_specs.each do |s|
  s.source = sources.get(s.source)
  
  # Don't add a spec to the list if its source is expired. For example,
  # if you change a Git gem to Rubygems.
  next if s.source.nil? || @unlock[:sources].include?(s.name)
```

When we run `bundle update --source rubygem_gem` and `s.name == 'rubygem_gem'` —

1. `s.source` is not nil
2. `@unlock[:sources]` is `['rubygem_gem']`

Therefore the condition returns true and the result of all this is that `#converge_locked_specs` returns a `SpecSet` that _omits_ `rubygem_gem`. This means that the gem is effectively unlocked, allowing it to be updated on its own without having any of its dependencies unlocked.

All beyond this line is speculation:

I looked for the commit where this was introduced, and found [this commit back in 2010.](https://github.com/bundler/bundler/commit/8a4dc7a12e150dae3f0a7e2f6de5a18f899191ac#diff-ec267c71117bf68f38c928a3691ba98cL152). It's a little hard to work out what's equivalent, but I found [this removal in `#converge`](https://github.com/bundler/bundler/commit/8a4dc7a12e150dae3f0a7e2f6de5a18f899191ac#diff-ec267c71117bf68f38c928a3691ba98cL152):

```ruby
@locked_specs.each do |s|
  if source = @sources.find { |source| s.source == source }
    s.source = source
  else
    @unlock << s.name
  end
end
```

Which would seem to go through locked specs, try to find their source, and if it can't find it then unlock the spec.

Correspondingly, [this addition in `#converge_locked_specs`](https://github.com/bundler/bundler/commit/8a4dc7a12e150dae3f0a7e2f6de5a18f899191ac#diff-ec267c71117bf68f38c928a3691ba98cR162):

```ruby
@locked_specs.delete_if do |s|
  s.source.nil? || (@unlock[:sources] || []).include?(s.name)
end
```

Which would seem to go through locked specs, and delete them if the source is nil, or if `@unlock[:sources]` contains... the spec name? Seems unexpected. Shouldn't that be the source name? It _could_ be unintended.

---

So, I've no strong opinion on whether you should be _able_ to 'update conservatively', but if the functionality is there it should probably be documented as such.

Sorry for the long read! Hope it was clear.

neoeno:
Fixes #3759 and #3761.

And also another issue I found where if:

1. You have a Gemfile with a group,
2. That group contains a gem with the same name as a source
3. And you run `bundle update --group group_name`
4. Then all gems associated with the source will be updated.

The 3 cases fixed here are all pretty subtle. The specs should make it clearer, and the github issues above add more context.

Note that this PR removes the ability to use `bundle update --source gem_name` as a hacky way to update a gem 'conservatively'. This is detailed in #3759.

### DIFF
--- a/lib/bundler/cli/update.rb
+++ b/lib/bundler/cli/update.rb
@@ -30,7 +30,7 @@ module Bundler
 
         if groups.any?
           specs = Bundler.definition.specs_for groups
-          sources.concat(specs.map(&:name))
+          gems.concat(specs.map(&:name))
         end
 
         Bundler.definition(:gems => gems, :sources => sources)
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -557,6 +557,11 @@ module Bundler
         # Don't add a spec to the list if its source is expired. For example,
         # if you change a Git gem to Rubygems.
         next if s.source.nil? || @unlock[:sources].include?(s.source.name)
+
+        # XXX This is a backwards-compatibility fix to preserve the ability to
+        # unlock a single gem by passing its name via `--source`. See issue #3759
+        next if s.source.nil? || @unlock[:sources].include?(s.name)
+
         # If the spec is from a path source and it doesn't exist anymore
         # then we just unlock it.
 
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -556,7 +556,7 @@ module Bundler
 
         # Don't add a spec to the list if its source is expired. For example,
         # if you change a Git gem to Rubygems.
-        next if s.source.nil? || @unlock[:sources].include?(s.name)
+        next if s.source.nil? || @unlock[:sources].include?(s.source.name)
         # If the spec is from a path source and it doesn't exist anymore
         # then we just unlock it.
 
--- a/spec/commands/update_spec.rb
+++ b/spec/commands/update_spec.rb
@@ -132,6 +132,8 @@ describe "bundle update" do
 
   describe "with --source option" do
     it "should not update gems not included in the source that happen to have the same name" do
+      pending("Allowed to fail to preserve backwards-compatibility")
+
       install_gemfile <<-G
         source "file://#{gem_repo2}"
         gem "activesupport"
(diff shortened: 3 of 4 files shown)


indirect:
Thanks for the patch! As I commented over in #3759, we can't remove the `--source` hack until 2.0. If you remove that commit from this PR, I'd be happy to merge here. That said, I would also be super happy to merge that patch into `2-0-dev` if you can submit your patch as a PR against that branch.

I'd also love to ship a documented `--only` option that limits updates to a single gem or gems. Since you're currently the world expert on this code, would you be willing to implement that in a PR?

Thanks!

neoeno:
Ha — I just realised why the tests are failing. Too much of the wrong kind of thinking! New commit to (hopefully) fix.

Anyway, thanks for the response! Hearing you on the backwards-compatibility thing. For this PR, I could have a go at fixing #3761 and the issue detailed in the OP — while preserving the undocumented `--source` behaviour. Does that sound like it'd be useful?

Also happy to take a look at `2-0-dev` and look at applying the `--source` fix to there.

Piqued my curiosity on the potential `--only` flag so I'll take a look at that too.

indirect:
Awesome! That all sounds great. :)

neoeno:
Looks like I've still got plenty of work to do making this pass with old versions of RubyGems(?) but I think this is more or less in its final form.

It fixes #3761, and the `--group` issue detailed in this PR — along with adding tests for the behaviour. It preserves the `--source` behaviour detailed in #3759 and adds tests for it too (since it's now expected). There's also a test for the 'correct' behaviour which I've left pending with a comment to explain the situation.

I'll have a go at correcting the behaviour in the `2-0-dev` branch next.

indirect:
This is super impressive, thanks! :grinning:

neoeno:
So I'm finding the tests here a little hard to figure out — seems like travis isn't running on this PR anymore? And also I'm not sure if master is green right now? Could you advise?

As far as I can work out the tests here pass (for me), except `spec/commands/help_spec.rb:28` which fails for me on master too.

segiddins:
@neoeno I had travis re-run the PR, it looks like the only failures are due to rubocop, `bin/rake spec:deps rubocop:auto_correct`

indirect:
Rebased and corrected styles at #3948, which will merge when the tests pass.

neoeno:
Ah — sorry about that! Thought I'd caught all those.

neoeno:
(CC #3948) OK —

* Rebased in @indirect's rubocop fixes
* Amended some testing message terminology
* Moved some tests out of context blocks as re-building git repos in the same test run seemed to cause the issues in older versions of RubyGems.

One test fail remains: https://travis-ci.org/bundler/bundler/jobs/76903851 — is this expected or unexpected? Seems like an odd state it's got into.

indirect:
We've been fighting with that particular periodic test failure lately—just merged something to master that I think will help. In the meantime, I've just restarted that particular job and I expect it to pass. Thanks for sticking with this, and thanks again for all your work on this set of issues!

indirect:
@homu r+

neoeno:
Looks like another strange result. Let me know if there's anything I should be doing!

indirect:
Ugh, that must be the RubyGems connectivity issues that are happening right now. :(

PR #3764, opened 2015-06-19
Change default install path to ./.bundle/gems [WIP]

smlance:


lib/bundler/settings.rb
```diff
@@ -138,7 +138,7 @@ def path
       if path = self[:path]
         "#{path}/#{Bundler.ruby_scope}"
       else
-        Bundler.rubygems.gem_dir
+        Pathname.new(File.dirname(__FILE__)).join(".bundle/gems")
```
segiddins: Should this take into account the ruby scope?

smlance: `Bundler.ruby_scope` is a string like `ruby/1.9.1`; why would this be necessary in the path?

indirect: @smlance the Ruby scope string defines the Ruby ABI (application binary interface), which breaks compatibility for compiled gems when it changes. The path gems are installed to always needs to include it.

@segiddins the Rubygems gem path already includes the scope string, so we don't need to add it ourselves.

segiddins: Except this is no longer using the RG path?

-Samuel E. Giddins

On Jun 19, 2015, at 12:46 PM, André Arko <notifications@github.com> wrote:

In lib/bundler/settings.rb:

> @@ -138,7 +138,7 @@ def path
>        if path = self[:path]
>          "#{path}/#{Bundler.ruby_scope}"
>        else
> -        Bundler.rubygems.gem_dir
> +        Pathname.new(File.dirname(__FILE__)).join(".bundle/gems")
@smlance the Ruby scope string defines the Ruby ABI (application binary interface), which breaks compatibility for compiled gems when it changes. The path gems are installed to always needs to include it.

@segiddins the Rubygems gem path already includes the scope string, so we don't need to add it ourselves.

lib/bundler/settings.rb
```diff
@@ -138,7 +138,7 @@ def path
       if path = self[:path]
         "#{path}/#{Bundler.ruby_scope}"
       else
-        Bundler.rubygems.gem_dir
+        Pathname.new(@root)
```
indirect: Ah, this needs to be `File.join(@root, Bundler.ruby_scope)`.

spec/support/path.rb
```diff
@@ -15,7 +15,9 @@ def home(*path)
     end
 
     def default_bundle_path(*path)
-      system_gem_path(*path)
+      root = tmp.join("bundled_app")
+      FileUtils.mkdir_p(root)
+      root.join(".bundle").join(*path)
```
indirect: You can definitely implement this in terms of `bundled_app`, defined below. Something like `bundled_app(".bundle", Bundler.ruby_scope)` should do it.

indirect:
Here's an example of fixing a test by changing the assumption about where gems get installed: https://github.com/bundler/bundler/tree/default-install-path-2-0.

In the future, if you push branches to the bundler/bundler repo for opening pull requests, we can collaborate on them... might be easier. :)

indirect:
Moving this to #3784.

PR #3766, opened 2015-06-19
Integrate Code of Covenant 1.1.0 changes

strand:
See also: http://contributor-covenant.org/version/1/1/0/code_of_conduct.md

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -20,6 +20,8 @@ If you are being harassed, notice that someone else is being harassed, or have a
 
 We expect everyone to follow these rules anywhere in the Bundler project’s codebases, issue trackers, IRC channel, group chat, and mailing lists.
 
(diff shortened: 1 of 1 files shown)


indirect:
:+1: thanks

TimMoore:
As a general procedural point, I think it would be good if we could @-mention @bundler/core for changes to the CoC to make sure everyone has a chance to understand and approve of the changes, since we are committed to abide by and enforce it. I just want to make sure that everyone is aware, since it's easy to miss stuff that happens quickly in all of the activity on the project.

For the record, I'm +1 on this particular change. :smile:

indirect:
@TimMoore thanks, that's a really good point! I saw this as explicitly stating something that we had already agreed on in @bundler/core, but you're right, and I will make sure to check in the future. :)

andremedeiros:
:+1 @TimMoore.

And :+1: on the change.

segiddins:
I'm :+1: on both counts.

PR #3768, opened 2015-06-20
Update checked rails version in realworld spec to make it pass

karlo57:
I noticed that the spec fails because rails has updated to `3.2.22`.


TimMoore:
Thanks for the pull request! We had already merged in a similar change on the `1-10-stable` branch, but had not yet merged that branch back to master. I've just done that now, which should hopefully fix the build running now https://travis-ci.org/bundler/bundler/builds/67709660.

PR #3769, opened 2015-06-20
Expose setup with friendly errors and custom groups

habibalamin:
Allows user to `require 'bundler/setup_friendly` and just call Bundler.setup_friendly with custom groups, while still adding Bundler to the load path and printing friendly errors.

#3770.


habibalamin:
Issue has been solved.

Issue #3765, opened 2015-06-19
Network s unreachable


vipul2690:
--- ERROR REPORT TEMPLATE -------------------------------------------------------
- What did you do?

  I ran the command `/home/vipul/.rvm/gems/ruby-2.2.0/bin/bundle install`

- What did you expect to happen?

  I expected Bundler to...

- What happened instead?

  Instead, what actually happened was...

Error details

    Errno::ENETUNREACH: Network is unreachable - connect(2) for "rubygems.org" port 443
      /home/vipul/.rvm/rubies/ruby-2.2.0/lib/ruby/2.2.0/net/http.rb:879:in `initialize'
      /home/vipul/.rvm/rubies/ruby-2.2.0/lib/ruby/2.2.0/net/http.rb:879:in `open'
      /home/vipul/.rvm/rubies/ruby-2.2.0/lib/ruby/2.2.0/net/http.rb:879:in `block in connect'
      /home/vipul/.rvm/rubies/ruby-2.2.0/lib/ruby/2.2.0/timeout.rb:74:in `timeout'
      /home/vipul/.rvm/rubies/ruby-2.2.0/lib/ruby/2.2.0/net/http.rb:878:in `connect'
      /home/vipul/.rvm/rubies/ruby-2.2.0/lib/ruby/2.2.0/net/http.rb:1456:in `begin_transport'
      /home/vipul/.rvm/rubies/ruby-2.2.0/lib/ruby/2.2.0/net/http.rb:1409:in `transport_request'
      /home/vipul/.rvm/rubies/ruby-2.2.0/lib/ruby/2.2.0/net/http.rb:1383:in `request'
      /home/vipul/.rvm/gems/ruby-2.2.0/gems/bundler-1.10.4/lib/bundler/vendor/net/http/persistent.rb:999:in `request'
      /home/vipul/.rvm/gems/ruby-2.2.0/gems/bundler-1.10.4/lib/bundler/fetcher/downloader.rb:45:in `request'
      /home/vipul/.rvm/gems/ruby-2.2.0/gems/bundler-1.10.4/lib/bundler/fetcher/downloader.rb:15:in `fetch'
      /home/vipul/.rvm/gems/ruby-2.2.0/gems/bundler-1.10.4/lib/bundler/fetcher.rb:87:in `fetch_spec'
      /home/vipul/.rvm/gems/ruby-2.2.0/gems/bundler-1.10.4/lib/bundler/remote_specification.rb:46:in `_remote_specification'
[...]

PR #3771, opened 2015-06-20
added Errno::ENETUNREACH to list of handled HTTP errors

steverob:
Is this enough to handle #3765 (where an issue related to not handling Errno::ENETUNREACH was reported)?


indirect:
I think you're right, thanks!

PR #3773, opened 2015-06-20
Show deprecation warning when using --binstubs

smlance:


indirect:
:+1:

smlance:
Should a deprecation message also be shown in the `bundle help` output (`lib/bundler/cli.rb:120`)?

indirect:
Nah, I don't think so. It's more important to print the deprecation when the option is used.

PR #3774, opened 2015-06-20
Remove --binstubs from cli/install.rb

smlance:
- Also, remove `bundle install --binstubs` documentation from
  the config and install manpages


smlance:
man/bundle-install.ronn:104 needs to be taken care of, as well.

indirect:
I'll wait to merge this until the manpage is edited, then...

spec/runtime/executable_spec.rb
```diff
@@ -8,142 +8,9 @@
     G
   end
```
smlance: It seems like this file was primarily testing `bundle install --binstubs`, so most of its specs have been removed.

smlance:
Because of https://github.com/bundler/bundler/pull/3662/files, we needed to remove the first spec in `spec/commands/help_spec.rb`.

UPGRADING.md
```diff
@@ -16,4 +16,10 @@ This change is not expected to impact production deployment scripts, because tho
...
\ No newline at end of file
+The `bundler/capistrano` and `bundler/vlad` deployment helper files have been removed. Capistrano 3 has built-in Bundler integration in the form of the `capistrano-bundler` gem, and everyone using Capistrano 3 should be using that instead. If for some reason, you are still using Capistrano 2, feel free to copy the Capistrano tasks out of the Bundler 1 file `bundler/deployment.rb` and put them into your app.
+
+#### Removed: `bundle install --binstubs`
+
+The `--binstubs` option has been removed from `bundle install` and replaced with the `bundle binstubs` command. This means that binstubs must be created and checked into version control individually.
+
+The `bundle binstubs [GEM]` command accepts two optional arguments: `--force`, which overwrites existing binstubs if they exist, and `--path=PATH`, which specifies the directory in which the binstubs will be installed (./bin by default).
```
segiddins: place the `./bin` in backticks

spec/runtime/executable_spec.rb
segiddins: van you make sure `bundle binstubs` is properly tested, and if not, change these specs to test it?

smlance: It looks like binstubs is tested in `spec/commands/binstubs_spec.rb`.

smlance: Although we could change these specs so they use `bundle binstubs` and include them, as long as they're not (too) redundant.

segiddins: :+1:

segiddins:
@smlance try not to comment on individual commits, but rather on the PR itself?

segiddins:
> What does the \"\" do at spec/runtime/executable_spec.rb:115? I'm unsure why this spec is failing, but perhaps this has line something to do with it.

It's intentionally trying to install to a path denoted by the empty string

PR #3775, opened 2015-06-21
allow for finding git sources on $GEM_PATH

cstrahan:
__This is still a work-in-progress, and needs tests.__

------

## Feature Description

This PR makes it possible for Bundler to find caches and installation paths for git sources in `$GEM_PATH`.

## Motivation

RubyGems/Bundler already allows for finding normal gem installations across both `$GEM_HOME` and `$GEM_PATH`; this makes git sources behave the same way.

More importantly, to package Ruby applications on [NixOS](http://nixos.org/) (and similar distros, like [GuixSD](http://www.gnu.org/software/guix/)), we need a deterministic way to fetch git sources. This means that we must fetch git sources individually, given a sha256 hash of directory contents. As of right now, I maintain an unsustainable set of monkey-patches which, prior to `bundle install`ing, point Bundler at the pre-fetched git repos instead of allowing Bundler to fetch repos across the network. Ideally, our package infrastructure would direct `gem install` to install gems into the right places to then be picked up from `$GEM_PATH` (without additional hacks, we can't use `$GEM_HOME`: Nix-like distros install packages into unique prefixes, so it's not possible to point `$GEM_HOME` to a single all encompassing file path).

## Feedback?

Would this be a welcome contribution? If so, I'll write up some tests.

### DIFF
--- a/lib/bundler/source/git.rb
+++ b/lib/bundler/source/git.rb
@@ -84,6 +84,10 @@ module Bundler
         @install_path ||= begin
           git_scope = "#{base_name}-#{shortref_for_path(revision)}"
           path = Bundler.install_path.join(git_scope)
+          unless path.exist?
+            candidates = find_in_gem_path("bundler", "gems", git_scope)
+            path = candidates.first unless candidates.empty?
+          end
 
           if !path.exist? && Bundler.requires_sudo?
             Bundler.user_bundle_path.join(Bundler.ruby_scope).join(git_scope)
@@ -202,7 +206,15 @@ module Bundler
           if Bundler.requires_sudo?
             Bundler.user_bundle_path.join("cache/git", git_scope)
           else
-            Bundler.cache.join("git", git_scope)
+            default_cache_path = Bundler.cache.join("git", git_scope)
+
+            candidates = [default_cache_path] + find_in_gem_path("cache", "bundler", "git", git_scope)
+            cache_path = candidates.detect do |path|
+              proxy = GitProxy.new(path, uri, ref, cached_revision, self)
+              proxy.has_revision_cached?
+            end
+
+            cache_path || default_cache_path
           end
         end
       end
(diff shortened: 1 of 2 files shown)


indirect:
Huh, I had no idea that Bundler respected $GEM_PATH. I wonder if that's always been true, or if that happened sometime within the last few years due to changes inside RubyGems...

Anyway, more to the specific point, have you considered packaging Ruby applications on NixOS by using the `path:` option in the Gemfile instead of doing git shenanigans? Bundler does not and has never supported a path consisting of multiple directories, and it's basically an accident that it honors $GEM_PATH. We only ever intended to support $GEM_HOME or $BUNDLE_PATH, as single directories containing all of the dependencies of a single application.

cstrahan:
> Huh, I had no idea that Bundler respected $GEM_PATH

Ah, I just realized (AFAICT) Bundler doesn't respect $GEM_PATH.

> Anyway, more to the specific point, have you considered packaging Ruby applications on NixOS by using the path: option in the Gemfile instead of doing git shenanigans? 

There are a couple problems with that. IIRC, using `path:` doesn't allow for building c-extensions. Even if it did, we wouldn't want to have to rewrite and maintain every Gemfile from every respective project. Also, we want to compose applications from packages of the dependencies, rather than having one monolithic package that tries to install everything at once; using separate packages, we prevent massive rebuilds when only a few dependencies change, it gives us build parallelism for free (for example, gitlab installs 20% faster), makes our build infrastructure much, much simpler, and - should a gem fail to build/install - it makes it much easier to track down why/where the build is failing (and, because the successfully built dependencies don't need to be rebuilt, only the gem that failed needs to be rebuilt, not everything needs to built over again.)

Anyway, we're working around this by creating a symlink forest to give the appearance that everything is installed in the same prefix.

indirect:
I'm not opposed to this in principle, especially since we do find installed gems everywhere in $GEM_PATH. @cstrahan would you be willing to add tests for this?

Fair warning, Bundler 2.0 is focusing more on installing gems for a single application directly into that application, and we will be changing the `--system` default to a `--path .bundle` default instead. I'm not sure how or even if that interacts with the way NixOS handles gems.

indirect:
This seems to have been dropped, so I'm closing it, at least for now.

PR #3777, opened 2015-06-23
Custom console

pducks32:
It might just be me, but switching to Pry is the first thing I do when I make a new gem so I thought this would be helpful to others as well.
It's done in a little strange of a manner to allow any console library to work—a new console shouldn't have to come to us for their users to be able to use it, that would be unfair.


segiddins:
Closing as a duplicate of https://github.com/bundler/bundler/pull/3334.

PR #3779, opened 2015-06-24
[Definition] Dont updated bundled with when 100% unnecessary

segiddins:
\c @indirect @andremedeiros

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -248,21 +248,22 @@ module Bundler
       # i.e., Windows with `git config core.autocrlf=true`
       contents.gsub!(/\n/, "\r\n") if @lockfile_contents.match("\r\n")
 
+      if @locked_bundler_version
+        locked_major = @locked_bundler_version.segments.first
+        current_major = Gem::Version.create(Bundler::VERSION).segments.first
+
+        if updating_major = locked_major < current_major
+          Bundler.ui.warn "Warning: the lockfile is being updated to Bundler #{current_major}, " \
+                          "after which you will be unable to return to Bundler #{@locked_bundler_version.segments.first}."
+        end
+      end
+
+      preserve_bundled_with ||= !updating_major && (Bundler.settings[:frozen] || !@unlocking)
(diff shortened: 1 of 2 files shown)


lib/bundler/definition.rb
```diff
@@ -248,23 +248,24 @@ def lock(file, preserve_bundled_with = false)
...
-
       if @locked_bundler_version
         locked_major = @locked_bundler_version.segments.first
         current_major = Gem::Version.create(Bundler::VERSION).segments.first
 
-        if locked_major < current_major
-          Bundler.ui.warn "Warning: the lockfile is being updated to Bundler #{Bundler::VERSION.split('.').first}, " \
+        if updating_major = locked_major < current_major
+          Bundler.ui.warn "Warning: the lockfile is being updated to Bundler #{current_major}, " \
                           "after which you will be unable to return to Bundler #{@locked_bundler_version.segments.first}."
```
indirect: Is it actually a warning that you have upgraded Bundler? I feel like the git diff will make it clear that this happened anyway...

segiddins: this has been there since 1.10.0, and I feel like we _should_ warn when making a backwards-incompatible update

indirect: OH that's major version. I was reading it as minor. :grimacing:

indirect:
This makes sense to me. :ship:

lib/bundler/definition.rb
```diff
@@ -248,23 +248,24 @@ def lock(file, preserve_bundled_with = false)
...
 
+      preserve_bundled_with ||= !updating_major && (Bundler.settings[:frozen] || !@unlocking)
+      return if lockfiles_equal?(@lockfile_contents, contents, preserve_bundled_with)
+
+      if Bundler.settings[:frozen]
+        Bundler.ui.error "Cannot write a changed lockfile while frozen."
+        return
+      end
+
       File.open(file, 'wb'){|f| f.puts(contents) }
```
andremedeiros: Can we fix the spacing here?

andremedeiros:
LGTM :shipit:

PR #3784, opened 2015-06-25
Change default install path to ./.bundle/gems

indirect:
Opening this PR from a branch in the bundler repo, so we can all collaborate on it... /cc @smlance

### DIFF
--- a/spec/install/path_spec.rb
+++ b/spec/install/path_spec.rb
@@ -85,13 +85,12 @@ describe "bundle install" do
         set_bundle_path(type, bundled_app("vendor").to_s)
 
         bundle :install
-
-        expect(bundled_app('vendor/gems/rack-1.0.0')).to be_directory
+        expect(bundled_app("vendor/#{Bundler.ruby_scope}/gems/rack-1.0.0")).to be_directory
         should_be_installed "rack 1.0.0"
       end
 
       it "installs gems to BUNDLE_PATH relative to root when relative" do
-        # If the bundle_path is `"vendor"` instead of
+        # FIXME: If the bundle_path is `"vendor"` instead of
         # `bundled_app("vendor").to_s`, this spec fails. As is, this spec
         # may not test what happens when `path` is relative.
         set_bundle_path(type, bundled_app("vendor").to_s)
@@ -101,7 +100,7 @@ describe "bundle install" do
           bundle :install
         end
 
-        expect(bundled_app('vendor/gems/rack-1.0.0')).to be_directory
+        expect(bundled_app("vendor/#{Bundler.ruby_scope}/gems/rack-1.0.0")).to be_directory
         should_be_installed "rack 1.0.0"
       end
     end
--- a/spec/install/path_spec.rb
+++ b/spec/install/path_spec.rb
@@ -91,7 +91,10 @@ describe "bundle install" do
       end
 
       it "installs gems to BUNDLE_PATH relative to root when relative" do
-        set_bundle_path(type, "vendor")
+        # If the bundle_path is `"vendor"` instead of
+        # `bundled_app("vendor").to_s`, this spec fails. As is, this spec
+        # may not test what happens when `path` is relative.
+        set_bundle_path(type, bundled_app("vendor").to_s)
 
         FileUtils.mkdir_p bundled_app('lol')
         Dir.chdir(bundled_app('lol')) do
--- a/lib/bundler/cli/config.rb
+++ b/lib/bundler/cli/config.rb
@@ -49,6 +49,18 @@ module Bundler
         new_value = args.join(" ")
         locations = Bundler.settings.locations(name)
 
+
+        if name == "path.system" and Bundler.settings[:path] and new_value == "true"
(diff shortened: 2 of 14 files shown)


indirect:
Reviewing the failing `clean` specs makes me realize something... we don't allow saving flags on install (like `install --system`) in 2.0. So, how can you use system gems via `bundle config`? Do you run `bundle config path --system`? Just clearing the value won't work, because we're changing the default. @bundler/core, any thoughts?

indirect:
Talked this over with @smlance, and the current plan is:

- [ ] bundle config path.system true` prints a warning if `path` is set when you set it
- [ ] `bundle config path foo` prints a warning if `path.system` is set when you set it
- [ ] unset the conflicting option when we print the warning

smlance:
This is looking good, so I'm going to squash the commits, open a new PR, and merge!

hone:
I must have missed a ton of context here, but what's the reasoning for
changing this from `vendor/bundle`? Just having a single bundler directory?

wycats:
fwiw: changes like this, while they may seem innocuous, can definitely break scripts designed to work with bundler. It should be considered a semver breaking change, and I agree with @hone that it would be good to articulate the rationale for changes with breaking potential.

segiddins:
This is targeting the Bundler 2.0 branch already, @wycats ?

indirect:
@hone @wycats we've been talking about the plan for 2.0 in the Slack for a year, and had Trello cards clearly outlining the exact plan for over six months. I'm happy to discuss if you would like, but this pull request is not the right place to discuss why this particular change is being made.

wycats:
@segiddins even if it's targeting 2.0, bundler should still be careful about the breaking changes it makes. A major version bump means bundler **can** make changes, but the cost of the changes should still outweigh the benefits.

wycats:
@indirect I have mostly stayed out of bundler discussions because the project maintained the basic assumptions and assumed the basic constraints we agreed to when we started the project.

As a daily user of bundler, and someone who has evangelized bundler's ideas to other communities (and very recently implemented a similar strategy for Rust), I think it's appropriate for me to express concerns about the number of changes in 2.0, even though I stayed hands-off of the development process.

Since I'm not involved in the development process, you can feel free to ignore anything I've said; them's the break with open source. I'm just letting you know how I feel.

indirect:
@wycats thanks for letting us know how you feel. FWIW, Cargo's default strategy is much closer to this change than it is to Bundler 1.0's default strategy. Again, I'm happy to discuss, as long as it's somewhere else. This PR is for code review of @smlance's Google Summer of Code work implementing the long-standing plans for 2.0.

wycats:
@indirect No problem. Name a place :smile: 

For what it's worth, one of the basic principles was keeping things stable between releases of bundler, so that people can generally use `Gemfile.lock`s and directory structures produced by a new version of bundler with an old version of bundler. Obviously, any new features in the DSL would not work with old versions of bundler, but people tend to have new versions of bundler on their system well before they start using the new features in their projects.

We've talked about all of this before, and I think 2.0 is an ok time to make people upgrade "all at once" if absolutely necessary.

In case it's not clear, I am extremely thankful for all the hard work you and others have done to maintain bundler over the years. When I say I've kept out of development because you've done a good job maintaining the aspects of bundler I care about, I'm not just saying it; it's really true.

segiddins:
@wycats https://trello.com/c/K4zdOFl8/51-default-installs-to-path-bundle-gems is the troll card for this change in particular, and there's a whole column for breaking changes in 2.0 at https://trello.com/b/DioUaKgx/bundler.

hone:
Thanks @segiddins. I didn't mean to hijack the thread. I was just curious on the reasoning. Talked with @indirect a bit about it. I follow up on the card. Cheers and thanks for working on bundler @smlance!

smlance:
@indirect and I thought it made sense to change the `install_path` and `bin_path` in `source/rubygems.rb` as well, since these variables are used when `bundle install` is run. Are there any other default bundle path variables in the code that need to be changed?

smlance:
The `bundle install --deployment` path redundancy can be seen here: https://gist.github.com/smlance/d8214f81c2652de98e3a

indirect:
looking good :+1:

spec/install/path_spec.rb
```diff
@@ -85,20 +85,22 @@ def set_bundle_path(type, location)
...
+        expect(bundled_app("vendor/#{Bundler.ruby_scope}/gems/rack-1.0.0")).to be_directory
         should_be_installed "rack 1.0.0"
       end
 
       it "installs gems to BUNDLE_PATH relative to root when relative" do
-        set_bundle_path(type, "vendor")
+        # FIXME: If the bundle_path is `"vendor"` instead of
+        # `bundled_app("vendor").to_s`, this spec fails. As is, this spec
+        # may not test what happens when `path` is relative.
+        set_bundle_path(type, bundled_app("vendor").to_s)
```
indirect: Maybe replace this line with `bundle "config path vendor"`, and see if the test still passes? If it does, we're good. If not, we should fix it. :)

PR #3791, opened 2015-06-27
[WIP] Add integration specs for not autoremembering flags

smlance:
@pducks32 and I thought that modifying `@use_current` in `Bundler.settings#[]=` was the most sensible way to test that flags are not autoremembered between commands, since we cannot modify `Bundler::VERSION` within specs.

Since users can run `bundle config use_current true`, should we document this setting, or should we leave it as a hidden setting used solely in the added specs?

Relevant: https://github.com/bundler/bundler/pull/3744


segiddins:
@smlance yeah, you can modify the version in specs :P

lib/bundler/settings.rb
```diff
@@ -33,7 +33,8 @@ def [](name)
     # remembered arguments.
     # See https://trello.com/c/yGsPNDpg
     def []=(key, value)
-      @use_current ||= (Bundler::VERSION.split(".")[0].to_i >= 2)
+      @use_current ||= Bundler.settings[:use_current] ||
+                       (Bundler::VERSION.split(".")[0].to_i >= 2)
```
segiddins: since this is always on the `2.0` branch, this line will always return `true`

smlance:
> yeah, you can modify the version in specs :P
How?

> since this is always on the 2.0 branch, this line will always return true
The problem is that the version number in 2-0-dev is currently set to 1.10.4.

segiddins:
@smlance a PR before this that changes the version and fixes the failing specs because of that would be excellent!

You can use the `simulate_bundler_version` helper to change the bundler version inside the given block.

smlance:
Thanks; I'll look into both of those shortly!

smlance:
https://github.com/bundler/bundler/pull/3793

segiddins:
@smlance what's the status of this?

smlance:
@segiddins It's on hold until we implement not autoremembering flags.

segiddins:
This'll need a rebase

PR #3792, opened 2015-06-27
Add deprecation warning for not autoremembering flags

smlance:
It'd be better if I squashed these commits...

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -15,7 +15,6 @@ module Bundler
 
     def initialize(*args)
       super
-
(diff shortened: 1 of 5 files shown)


lib/bundler/cli.rb
```diff
@@ -408,6 +408,11 @@ def env
 
     private
 
+
+    def given_flags?
```
segiddins: `options_given?`

segiddins:
I'd prefer to call it `options` instead of `flags`

smlance:
We may also want to add a setting that users can toggle with `bundle config` to stop showing this warning.

PR #3793, opened 2015-06-28
Change version number to 2.0.0 & fix specs [WIP]

smlance:
Many of the failing specs test that flags are saved between commands. Since this is deprecated, the specs can be fixed if we expect the flags to not be saved between commands by inverting the expectations. But I can see how this could be considered redundant, since they're all testing the same functionality. Should I remove these specs, or should I continue to modify them?


lib/bundler/version.rb
```diff
@@ -2,5 +2,5 @@ module Bundler
   # We're doing this because we might write tests that deal
   # with other versions of bundler and we are unsure how to
   # handle this better.
-  VERSION = "1.10.4" unless defined?(::Bundler::VERSION)
+  VERSION = "2.0.0" unless defined?(::Bundler::VERSION)
```
segiddins: `2.0.0.dev` ?

indirect:
I think we want to tests to cover: 1) does the command honor the flag when it's passed, 2) does the flag honor the config setting when it's set.

I'm honestly okay with not having any tests to cover "this flag was not saved", as long as there are unit tests on the `Settings` class that cover that?

indirect:
should we wait on this? should we try to merge it soon?

smlance:
Sorry for the delay -- I'll look into this shortly. I'm waiting on the RuboCop merge right now.

smlance:
Closing in favor of https://github.com/bundler/bundler/pull/3863.

PR #3802, opened 2015-07-01
Change default install path to ./.bundle/gems

smlance:
See https://github.com/bundler/bundler/pull/3784.

In addition, this will:

- Handle conflicting path config settings; behavior may be undefined if the user sets both manually in their config file

- Ensure that all specs pass


smlance:
Never mind; closing in favor of https://github.com/bundler/bundler/pull/3784, because this is a duplicate PR. (Thanks, @segiddins!)

PR #3807, opened 2015-07-04
Plugin System Proof of Concept

suhastech:
Hi,

I had created a proof of concept for the Plugin system. The sample plugin can be found at https://github.com/suhastech/bundler-sample-plugin 

It is still early stage. Only subcommand module has been implemented.

Currently in the pipeline: Source and Lifecycle.


segiddins:
@suhastech the next step is probably to write a bunch of tests for this, as I don't see any at the moment?

lib/bundler/cli.rb
```diff
@@ -73,6 +74,32 @@ def self.handle_no_command_error(command, has_namespace = $thor_runner)
       Kernel.exec(command_path, *ARGV[1..-1])
     end
 
+    
+
+    Bundler.plugin("1").manager.commands.all.each do |command,proc|
```
segiddins: what's this doing?

lib/bundler/plugin/state_file.rb
```diff
@@ -0,0 +1,46 @@
+require "yaml"
+
+module Bundler
+  module Plugin
+    class StateFile
+
+      # A super basic persitance mechanism for installed plugin. 
+      # As of now, it only stores the require entry points. It could be much more detailed.
+      def initialize
+      	@path = File.expand_path("~/.bundler/plugins.yml")
```
segiddins: this is only for global plugins? since it apples to every ruby, on every project

lib/bundler/cli.rb
suhastech: It is creating a method for every registered command. That's how Thor defines sub commands. I did look for ways to do it dynamically but couldn't find one that worked. There's probably a better way to do it.

lib/bundler/plugin/state_file.rb
suhastech: Yup, only for global plugins. 
Entry points for project based plugins are in the Gemfile itself. I figured, you don't need another file.

segiddins: In that case, this file's location needs to be managed via `settings`

spec/plugin/lifecycle_hooks_spec.rb
```diff
@@ -0,0 +1,44 @@
+require "spec_helper"
+
+describe "bundler lifecycle plugin" do
+  before :each do
+    build_gem "bundler-foo", :to_system => true do |s|
+      main = <<-EOF
+class BundlerCustomLifecycle < Bundler.plugin("1")
+  name "Custom lifecycle plugin"
+  lifecycle ["before_install", "after_install"] do
+    require_relative "lifecycle"
```
segiddins: this will fail on 1.8.7

spec/plugin/source_plugin_spec.rb
```diff
@@ -0,0 +1,67 @@
+require "spec_helper"
+
+describe "bundler source plugin" do
+  before :each do
+    build_gem "bundler-foo", :to_system => true do |s|
+      main = <<-EOF
+class BundlerCustomSource < Bundler.plugin("1")
+  name "Custom source plugin"
+  source "foo" do
+    require_relative "source"
```
segiddins: not available on 1.8.7

spec/plugin/subcommand_plugin_spec.rb
```diff
@@ -0,0 +1,51 @@
+require "spec_helper"
+
+describe "bundler subcommand plugin" do
+  before :each do
+    build_gem "bundler-foo", :to_system => true do |s|
+      main = <<-EOF
+class BundlerCustomSubcommand < Bundler.plugin("1")
+  name "Custom subcommand plugin"
+  command "foo" do
+    require_relative "command"
```
segiddins: 1.8.7

spec/plugin/subcommand_plugin_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    @command_name = "Foo [OPTIONS]"
+    @command_short_description = "This will print foo + argument"
+    @command_long_description =
+    <<-D
+          This will print hi + argument
+    D
+  end
+
+  def run(options, args)
+    puts "hi "+args[0]
```
segiddins: please test all of the provided arguments here

spec/plugin/subcommand_plugin_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  it "uses the plugin" do
+    bundle "foo bar"
+    expect(out).to include("hi bar")
+  end
+
+  it "uninstalls the subcommand plugin" do
+    bundle "plugin uninstall foo"
+    expect(out).to include("Bundler plugin 'foo' has been successfully uninstalled")
```
segiddins: also verify it is no longer available

segiddins:
@suhastech this needs a rebase.
I'd also like to see significantly more test coverage, particularly unit tests.

bin/bundle
```diff
@@ -14,6 +14,14 @@ $LOAD_PATH.each do |path|
   end
 end
 
+# Require all the user level plugins
+Bundler.settings.plugins.each do |plugin|
```
segiddins: doing this in the exe is going to be potentially very slow, is there any way we could avoid loading plugins when not 100% necessary?

segiddins: (i.e. I'd love to avoid it on `bundle exec`)

lib/bundler.rb
```diff
@@ -46,6 +49,32 @@ module Bundler
...
+  PLUGIN_COMPONENTS = Registry.new.tap do |c|
+    c.register(:"1")                  { Plugin::V1::Plugin    }
+    c.register([:"1", :command])      { Plugin::V1::Command   }
+    c.register([:"1", :source])       { Plugin::V1::Source    }
+    c.register([:"1", :lifecycle])    { Plugin::V1::Lifecycle }
+
+  end
+
+
+  def self.plugin(version, component=nil)
```
segiddins: I'd prefer this to be `Bundler::Plugin.register`

lib/bundler.rb
```diff
@@ -170,6 +204,15 @@ def definition(unlock = nil)
       end
     end
 
+    def plugin_definition(unlock = nil)
```
segiddins: let's avoid adding even more methods to the top-level `Bundler` module please?

lib/bundler/cli.rb
```diff
@@ -73,6 +74,35 @@ def self.handle_no_command_error(command, has_namespace = $thor_runner)
...
+        else
+          desc "Plugin [OPTIONS]", "Unimplemented short description"
+          long_desc "Unimplemented long description"
+        end
+
+        define_method(command) do |*args|
+          begin
+            command_object.run(options, args)
+          rescue => e
+            Bundler.ui.error "Error executing plugin subcommand #{command}: #{e.message}"
```
segiddins: shouldn't treat plugin errors any differently

lib/bundler/cli.rb
```diff
@@ -102,6 +132,20 @@ def check
...
     end
 
+
+    desc "plugin --install [PLUGIN]", "Install or uninstall plugins"
+    long_desc <<-D
+      Plugins
+    D
+    method_option "install", :type => :string, :banner =>
+      "installs a plugin"
+    method_option "uninstall", :type => :string, :banner =>
```
segiddins: these should be subcommands, not options

lib/bundler/cli/install.rb
```diff
@@ -101,10 +101,27 @@ def run
       # rubygems plugins sometimes hook into the gem install process
       Gem.load_env_plugins if Gem.respond_to?(:load_env_plugins)
 
+      Bundler.plugin("1").manager.call_hook(:before_install)
```
segiddins: does this not also need to take any args?

lib/bundler/cli/install.rb
```diff
@@ -101,10 +101,27 @@ def run
...
+      plugin_definition = Bundler.plugin_definition
+
+      if plugin_definition
+        Bundler.ui.info "Setting up plugins..."
+        plugin_definition.validate_ruby!
+
+        Installer.install_plugin(plugin_definition, options)
+        plugin_definition.specs.to_a.each do |spec|
+          next if spec.name == 'bundler'
+          require spec.name
```
segiddins: shouldn't we only require a specific plugin file, so normal gems can also come with bundler plugins?

lib/bundler/cli/plugin.rb
```diff
@@ -0,0 +1,64 @@
...
+    attr_reader :options
+    attr_reader :args
+    def initialize(options, args)
+      @options = options
+      @args = args
+    end
+
+    def run
+      plugins = Bundler.settings.plugins
+      action = @args[0]
```
segiddins: please do this assignment in the initializer

lib/bundler/cli/plugin.rb
```diff
@@ -0,0 +1,64 @@
+module Bundler
+  class CLI::Plugin
+    attr_reader :options
+    attr_reader :args
+    def initialize(options, args)
+      @options = options
+      @args = args
+    end
+
+    def run
```
segiddins: this method is way too long

lib/bundler/cli/plugin.rb
```diff
@@ -0,0 +1,64 @@
...
+                Bundler.ui.info "Bundler plugin '#{name}' is already installed"
+              rescue LoadError
+                Bundler.ui.error "Bundler plugin '#{name}' is installed but can't be found in the system"
+              end
+            end
+
+
+          end
+        rescue => e
+          Bundler.ui.error e.message
```
segiddins: why are we swallowing this error?

lib/bundler/cli/plugin.rb
```diff
@@ -0,0 +1,64 @@
...
+
+          if !plugins.include?("bundler-#{name}")
+            plugins.push("bundler-#{name}")
+            plugin_definition.specs.to_a.each do |spec|
+              next if spec.name == 'bundler'
+              begin
+                require spec.name
+                Bundler.ui.confirm "Bundler plugin '#{name}' has been successfully installed"
+              rescue LoadError
+                Bundler.ui.error "Bundler plugin '#{name}' couldn't be installed"
```
segiddins: why is this error being swallowed?

lib/bundler/dsl.rb
```diff
@@ -302,6 +320,19 @@ def normalize_options(name, version, opts)
         end
       end
 
+      plugins.each do |type, proc|
+        if opts[type.to_s]
+          if version.first && version.first =~ /^\s*=?\s*(\d[^\s]*)\s*$/
```
segiddins: please comment this regex, as it stands I can't tell what it's doing

lib/bundler/plugin.rb
```diff
@@ -0,0 +1,5 @@
+module Bundler
+  module Plugin
+    autoload :V1,        "bundler/plugin/v1"
```
segiddins: I don't feel like this sort of versioning is necessary -- plugins in their gemspecs could just declare bundler version requirements?

lib/bundler/plugin/v1/command.rb
```diff
@@ -0,0 +1,22 @@
...
+        attr_reader :command_name, :command_short_description, :command_long_description
+
+        def initialize
+          @command_name = "Plugin [OPTIONS]"
+          @command_short_description = "Unimplemented short description"
+          @command_long_description = "Unimplemented long description"
+        end
+
+        def run(options, args)
+          Bundler.ui.info "The plugin hasn't implemented the run method"
```
segiddins: unnecessary

lib/bundler/plugin/v1/command.rb
```diff
@@ -0,0 +1,22 @@
+module Bundler
+  module Plugin
+    module V1
+      class Command
+        attr_reader :command_name, :command_short_description, :command_long_description
+
+        def initialize
+          @command_name = "Plugin [OPTIONS]"
+          @command_short_description = "Unimplemented short description"
+          @command_long_description = "Unimplemented long description"
```
segiddins: can these please be nil?

lib/bundler/plugin/v1/lifecycle.rb
```diff
@@ -0,0 +1,15 @@
+module Bundler
+  module Plugin
+    module V1
+      class Lifecycle
+
+        def run(hook_name)
```
segiddins: why not allow for an arbitrary hook payload to be passed in?

lib/bundler/plugin/v1/manager.rb
```diff
@@ -0,0 +1,58 @@
...
+        def lifecycle_hooks
+          Registry.new.tap do |result|
+            @registered.each do |plugin|
+              result.merge!(plugin.components.lifecycle_hooks)
+            end
+          end
+        end
+
+        def call_hook (hook)
+          if lifecycle_hooks.all[hook].is_a?(Array)
```
segiddins: prefer to avoid `is_a` checks

lib/bundler/plugin/v1/plugin.rb
```diff
@@ -0,0 +1,104 @@
+require "bundler/plugin/v1/components"
+require "bundler/plugin/v1/lifecycle"
+module Bundler
+  module Plugin
+    module V1
+      # This is the superclass for all V1 plugins.
+      class Plugin
+
+        ROOT_CLASS = self
+        UNSET_VALUE = Object.new
```
segiddins: what's wrong with `nil`?

lib/bundler/plugin/v1/plugin.rb
```diff
@@ -0,0 +1,104 @@
...
+        def self.command(name, &block)
+
+
+          # Validate the name of the command
+          if name.to_s !~ /^[-a-z0-9]+$/i
```
segiddins: constant for the regex

lib/bundler/plugin/v1/plugin.rb
```diff
@@ -0,0 +1,104 @@
...
+          if Bundler.plugin_install_mode
+            Bundler.ui.info "Setting up '#{name}' subcommand"
+          end
+
+          # Return the registry
+          components.commands.register(name.to_sym) do
+            block
+          end
+
+          nil
```
segiddins: explicit nil returns are unnecessary -- just declare the method to return void

lib/bundler/plugin/v1/plugin.rb
```diff
@@ -0,0 +1,104 @@
...
+
+          # Return the registry
+          components.sources.register(name.to_sym) do
+            block
+          end
+
+          nil
+        end
+
+        def self.lifecycle(name_array, &block)
```
segiddins: `names`

lib/bundler/plugin/v1/plugin.rb
```diff
@@ -0,0 +1,104 @@
...
+          raise ArgumentError, "must provide an array of hooks to register" unless name_array.is_a?(Array)
```
segiddins: if you must check type, let it be `Enumerable`

lib/bundler/plugin/v1/plugin.rb
```diff
@@ -0,0 +1,104 @@
...
+
+            if Bundler.plugin_install_mode
+              Bundler.ui.info "Setting up '#{hook}' lifecycle hook"
+            end
+
+            lifecycle_class = block.call
+            lifecycle_object = lifecycle_class.new
+
+
+            # Return the registry
```
segiddins: nothing here is actually returned

lib/bundler/plugin/v1/plugin.rb
```diff
@@ -0,0 +1,104 @@
...
+            components.lifecycle_hooks.register_hook(hook.to_sym, lifecycle_object)
+          end
+
+          nil
+        end
+
+
+        def self.name(name=UNSET_VALUE)
```
segiddins: `nil`

lib/bundler/plugin/v1/plugin.rb
```diff
@@ -0,0 +1,104 @@
+require "bundler/plugin/v1/components"
+require "bundler/plugin/v1/lifecycle"
+module Bundler
+  module Plugin
+    module V1
+      # This is the superclass for all V1 plugins.
+      class Plugin
```
segiddins: shouldn't be a class if it has no instance methods

lib/bundler/plugin/v1/source.rb
```diff
@@ -0,0 +1,39 @@
+module Bundler
+  module Plugin
+    module V1
+      class Source
```
segiddins: why doesn't this inherit from the existing `Source` class?

lib/bundler/plugin/v1/source.rb
```diff
@@ -0,0 +1,39 @@
...
+        attr_accessor :dependency_names
+
+        def initialize (options)
+        end
+
+        def unmet_deps
+          specs.unmet_dependency_names
```
segiddins: there is no `specs` variable

lib/bundler/plugin_dsl.rb
```diff
@@ -0,0 +1,49 @@
+require 'bundler/dsl'
+module Bundler
+  class PluginDsl
```
segiddins: `DSL` since it's an acronym

lib/bundler/plugin_dsl.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    def initialize
+      @dsl = Dsl.new
+      methods = Dsl.instance_methods(false)
+      methods.delete(:plugin)
+
+      self.class.instance_eval do
+        methods.each do |method|
+
+          define_method (method) do |*args|
```
segiddins: should probably also ignore the `&blk`

lib/bundler/plugin_dsl.rb
```diff
@@ -0,0 +1,49 @@
...
+      builder.eval_gemfile_for_plugins(gemfile)
+      builder.to_plugin_definition(lockfile, unlock)
+    end
+
+    def initialize
+      @dsl = Dsl.new
+      methods = Dsl.instance_methods(false)
+      methods.delete(:plugin)
+
+      self.class.instance_eval do
```
segiddins: why is this done for the class each `initialize`? Should be done at the class scope

segiddins: at that, why isn't this a subclass of the existing `DSL` class?

lib/bundler/plugin_dsl.rb
```diff
@@ -0,0 +1,49 @@
...
+    def eval_gemfile_for_plugins(gemfile)
+      contents ||= Bundler.read_file(gemfile.to_s)
+      instance_eval(contents, gemfile.to_s, 1)
+    rescue Exception => e
+      message = "There was an error parsing `#{File.basename gemfile.to_s}`: #{e.message}"
+      raise Dsl::DSLError.new(message, gemfile, e.backtrace, contents)
+    end
+
+    def plugin(name, *args)
+      @dsl.gem("bundler-#{name}", *args)
```
segiddins: imho should allow any gem names to be plugins, not just bundler-prefixed ones

lib/bundler/plugin_dsl.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+
+    def to_plugin_definition(lockfile, unlock)
+      if @dsl.dependencies.count > 0
+        @dsl.to_definition(lockfile, unlock)
```
segiddins: why not always return a definition?

lib/bundler/registry.rb
```diff
@@ -0,0 +1,72 @@
...
+      @items = {}
+      @results_cache = {}
+
+    end
+
+    # Register a key with a lazy-loaded value.
+    #
+    # If a key with the given name already exists, it is overwritten.
+    def register(key, &block)
+      raise ArgumentError, "block required" if !block_given?
```
segiddins: unnecessary if you also explicitly take the `&block` parameter

lib/bundler/registry.rb
```diff
@@ -0,0 +1,72 @@
...
+      @items[key] = block
+    end
+
+    # Register a key with a lazy-loaded value.
+    #
+    # If a key with the given name already exists, it is overwritten.
+    def register_hook(key, object)
+      @items[key] = [] if @items[key].nil?
```
segiddins: `||=`

lib/bundler/registry.rb
```diff
@@ -0,0 +1,72 @@
...
+      @items[key] << object if !@items[key].include?(object)
```
segiddins: why not always append, and then `uniq` ?

lib/bundler/registry.rb
```diff
@@ -0,0 +1,72 @@
...
+    end
+
+    # Get a value by the given key.
+    #
+    # This will evaluate the block given to `register` and return the
+    # resulting value.
+    def get(key)
+      return nil if !@items.key?(key)
```
segiddins: `unless`

lib/bundler/registry.rb
```diff
@@ -0,0 +1,72 @@
...
+      return @results_cache[key] if @results_cache.key?(key)
+      @results_cache[key] = @items[key].call
+    end
+    alias :[] :get
```
segiddins: `alias_method`

lib/bundler/registry.rb
```diff
@@ -0,0 +1,72 @@
...
+
+    # Returns all items in the registry
+    #
+    def all
```
segiddins: this should just be including `Enumerable`

lib/bundler/registry.rb
```diff
@@ -0,0 +1,72 @@
...
+      end
+    end
+
+    # Like #{merge} but merges into self.
+    def merge!(other)
+      @items.merge!(other.__internal_state[:items])
+      self
+    end
+
+    def __internal_state
```
segiddins: `private` ?

lib/bundler/settings.rb
```diff
@@ -129,6 +129,14 @@ def with
       get_array(:with)
     end
 
+    def plugins
+      @global_config[key_for(:plugins)] ? @global_config[key_for(:plugins)].split(":") : []
```
segiddins: what's wrong with `get_array` ?

lib/bundler/settings.rb
```diff
@@ -129,6 +129,14 @@ def with
...
+    end
+
+    def plugins=(array)
+      set_global(:plugins, (array.empty? ? nil : array.join(":")))
```
segiddins: `set_array` ?

lib/bundler/source.rb
```diff
@@ -13,7 +13,7 @@ def unmet_deps
     def version_message(spec)
       message = "#{spec.name} #{spec.version}"
 
-      if Bundler.locked_gems
+      if !Bundler.plugin_install_mode && Bundler.locked_gems
```
segiddins: why this change?

lib/bundler/source_list.rb
```diff
@@ -1,13 +1,25 @@
...
+                :git_sources,
+                :plugin_sources
 
     def initialize
       @path_sources       = []
       @git_sources        = []
       @rubygems_aggregate = Source::Rubygems.new
       @rubygems_sources   = []
+      @plugin_sources     = {}
+      @plugin_registry    = {}
```
segiddins: imho the source list shouldn't need to know about the registry

lib/bundler/cli/install.rb
suhastech: Only gems with 

`plugin "whatever"`
in the Gemfile

will be required. This is essential for Source plugins to work.

lib/bundler/cli/plugin.rb
suhastech: I'll raise it instead

lib/bundler/plugin/v1/command.rb
suhastech: I was thinking that improperly implemented plugin shouldn't be allowed to crash bundler. That's why some errors were "swallowed". 

Of course, we can let bundler crash and let the user decide to uninstall the misbehaving plugin. Now that I think about it, it does make sense.

lib/bundler/plugin/v1/lifecycle.rb
suhastech: Sure, I'll add that.

lib/bundler/plugin/v1/source.rb
suhastech: That should be implemented by the plugin

lib/bundler/plugin_dsl.rb
suhastech: The old DSL class was named Dsl. So, I figured a similar name would do.

lib/bundler/plugin_dsl.rb
suhastech: If we subclass the existing Dsl, it will run as the definition. This "pre flight" Dsl is used to only process 

`plugin "whatever"`

and ignore the rest of the definition. Hence, we need all the methods defined as the original Dsl but they shouldn't do anything. Hence the empty methods.

lib/bundler/plugin_dsl.rb
suhastech: I figured this would be much more readable.

lib/bundler/plugin_dsl.rb
suhastech: If we do that, it'll still have "bundler" spec by default when trying to install. 

When installing, it'll print out

`Using bundler 1.0.0`

This is break most of the tests.

So, it'll only try to install if plugins are actually defined.

lib/bundler/settings.rb
suhastech: Global settings are stored in ~/.bundle/config 

We wouldn't want to lose information about user level plugins.

lib/bundler/source.rb
suhastech: While installing a plugin in a directory without Gemfile, 

like so
`bundle plugin install whatever`

when I use the installer, to install the plugins, the installer checks for the locked gems to show "(locked at whatever)" and crashes.

So, I created a "plugin_install_mode" where it won't worry about Gemfile and Lockfile. It doesn't seem all that elegant, but works.

lib/bundler/cli/plugin.rb
suhastech: Done

suhastech:
@all Is this still on the agenda?

indirect:
@suhastech this is still on the agenda eventually, but we're focusing on launching 1.12 and the new index on the server for right now

PR #3808, opened 2015-07-05
Fix man page link

EduardoBautista:


EduardoBautista:
On second thought, maybe it should link to the folder that contains all the man pages in this repo?

indirect:
This looks like a bug! The site is supposed to render the current version of the man pages at /man/, similar to how Github shows the newest readme at /master/. Thanks for noticing!

indirect:
Fixed in bundler/bundler-site@137997ba1d4abbf2539181b691c8559cb3f584a0. Thanks!

PR #3809, opened 2015-07-05
Use bang methods for speed and memory efficiency

EduardoBautista:
The bang methods will edit the string in place so it won't allocate more memory and is much faster because of it.  I will try to find more places where this can be done, maybe we can make some small speed improvements across the entire codebase this way.


EduardoBautista:
Hmmm, I am getting some failures in the build, but I am not sure if they are from my code changes or not.

EduardoBautista:
Wait, I think I did break them...

EduardoBautista:
Ah...the strings are frozen...ok then...

EduardoBautista:
Lol, nvm then.

Issue #3806, opened 2015-07-03
bundle gem generates test framework dev dependency without version constraint


tf:
While I think `bundle gem -t rspec` is a nice touch, I wonder if there should be a version constraint on the generated gemspec entry. At the moment it only says `add_development_dependency 'rspec'`. Given that `Gemfile.lock` is git ignored, I think it would be best practice to depend upon a specific major version. Not sure though how to figure out the appropriate major version. Maybe simply hard coding it is enough since the generator will have to be checked for RSpec 4.0 compatibility anyway once that is released.

andremedeiros:
You're probably right, although I'm not sure how we would even go about this.

I'm open to suggestions as to how to pick versions.

@indirect @segiddins do you guys want to weigh in?

segiddins:
I agree hard coding a major version for rspec in the generator should be fine

andremedeiros:
So you mean, for RSpec, `~> 3.0` should do it?

tf:
I'd say so. I guess there would need to be some hash keeping major versions for the different test frameworks supported by the `-t` option.

indirect:
I'm fine with just tracking the current major version of RSpec and Minitest. Hard coded version strings seems fine, major releases aren't even every year.

tf:
Yes, but the name of the testing framework is [passed to the gemspec template as a variable](https://github.com/bundler/bundler/blob/master/lib/bundler/templates/newgem/newgem.gemspec.tt#L41).I'd suggest adding `config[:test_framework_version]`, which would be set 3.0 for RSpec and 5.0 for Minitest in [`Bundler::CLI::Gem`](https://github.com/bundler/bundler/blob/master/lib/bundler/cli/gem.rb#L68).

Would you be interested in a PR? Is there some sort of test coverage for the gem generator?

segiddins:
@tf yeah, a PR would be great! There's some test coverage in `spec/commands/newgem_spec.rb`

tf:
I've submitted a first draft. Maybe we can continue our discussion there.

tf:
Could somebody help me figure out why the tests of my PR #3811 fail for RubyGems < 2.0? Thanks!

PR #3811, opened 2015-07-05
Pessimistic test framework versions for bundle gem command

tf:
Here's a first draft of the change discussed in #3806.

Use a pessimistic version requirement for RSpec and Minitest in
gemspecs generated by the `bundle gem` command. It is best practice to
at least depend on major versions of development dependency gems.

* Hard code major versions of the two testing frameworks.
* Add a spec that tests that only specific version requirements are
  used in development dependencies.

Some questions:
* The test for`bundle gem` contains some duplication for cases concerning different gem names (with dash, with underscore etc). The shared example group that I added could have been used in more of the nested contexts, but I did not feel that those were testing different paths in my case.
* I mostly struggled with the naming of the shared example group. Probably someone could come up with something a little less stilted...

Looking forward to your feedback.

### DIFF
--- a/lib/bundler/cli/gem.rb
+++ b/lib/bundler/cli/gem.rb
@@ -2,6 +2,11 @@ require 'pathname'
 
 module Bundler
   class CLI::Gem
+    TEST_FRAMEWORK_VERSIONS = {
+      'rspec' => '3.0',
+      'minitest' => '5.0'
+    }
+
     attr_reader :options, :gem_name, :thor, :name, :target
 
     def initialize(options, gem_name, thor)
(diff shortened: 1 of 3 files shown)


tf:
Not sure why the specs are failing for Rubygems < 2.0. `remove_push_guard` seems to be called in the before block. Shouldn't that prevent the error?

indirect:
A couple of reasons... the guard removal code wasn't actually being called where you inserted the test, and the `specific?` method only exists on RubyGems 1.8.1 and higher. I pushed a commit with fixed tests to #3885.

tf:
Thanks! :+1:

indirect:
Merged in 26c4665788cd2601211cc54935411b8227c2f132. Thanks for the feature!

PR #3812, opened 2015-07-07
Add new binstubs for rake and rspec in bin, move executables to exe

segiddins:
\c @smlance @bundler/core


indirect:
Happy to merge this as soon as the rest of the specs are changed to match.

segiddins:
This should pass now

PR #3817, opened 2015-07-09
Removes a warning in jruby

deivid-rodriguez:
It seems that jruby's exec does not understand that option:

```
uri:classloader:/jruby/kernel/kernel.rb:28: warning: unsupported exec option: close_others

```

The warning was skipped for rubies under 2.0 (like latest released
jruby) but the new jruby 9.0.0.0 is 2.0 compatible by default so the
check needs to be refined.

### DIFF
--- a/lib/bundler/cli/exec.rb
+++ b/lib/bundler/cli/exec.rb
@@ -1,3 +1,5 @@
+require 'bundler/current_ruby'
+
 module Bundler
   class CLI::Exec
     attr_reader :options, :args, :cmd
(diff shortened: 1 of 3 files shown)


indirect:
Thanks!

indirect:
Unfortunately this change causes the specs to fail on Ruby versions < 2. Could you take a look?

deivid-rodriguez:
Ooops, my bad, I didn't run the specs on other rubies. I will take a look tomorrow!

deivid-rodriguez:
Fixed!

PR #3819, opened 2015-07-09
Fix install_if not clearing install conditions

danieltdt:
The new `install_if` command was not clearing `install_conditions` properly.
All gems defined after a `install_if -> false do ... end` would not be installed.

I think that fixes the problem.


indirect:
Thanks!

segiddins:
yikes, terrible copy-pasting on my part. Thanks for catching that, @danieltdt !

PR #3820, opened 2015-07-10
fix with and without groups

pducks32:
cc @smlance
This is still going to fail but at least with and without groups will be fine. Now they won't be auto remembered and will also override each other. As it turns out they didn't before, they would do it bluntly and not take into account the actual groups.


indirect:
I'm okay with merging this as progress towards a green build, but I'll hold off until it's ready. :)

smlance:
Update: we'll split this feature into separate PRs, because this change uproots a lot of old Bundler functionality. This PR is a good start, though -- thanks, @pducks32!

segiddins:
Should this be closed at this point?

indirect:
Yeah, I think so.

PR #3823, opened 2015-07-12
Fix option description for `bundle gem`

mattbrictson:
The `--bin` option incorrectly had the description "Set a default with `bundle config gem.mit true`." I've moved this text to the appropriate place in the `--mit` description.


segiddins:
:+1:

PR #3827, opened 2015-07-14
Remove `bundle install --path` and print an error if `bundle install --path` is used

smlance:


lib/bundler/cli/install.rb
```diff
@@ -44,9 +44,15 @@ def run
...
-                         "as well as --system. Please choose."
+      if options[:path]
+        Bundler.ui.error "You have specified an installation path with the "\
+         "path flag. Please use `bundle config path #{options[:path]}` instead."
+        exit 1
+      end
+
+      if (Bundler.settings[:path] || options[:deployment]) && options[:system]
+        Bundler.ui.error "You have configured a path to install your gems to, \n" \
+                         "and specified the --system path. Please choose."
```
segiddins: Please only use one?

lib/bundler/source/rubygems.rb
```diff
@@ -130,6 +130,9 @@ def install(spec, opts = {})
             bin_path     = Bundler.system_bindir
           end
 
+          $stderr.puts "install_path: #{install_path}"
```
segiddins: delete these

smlance:
Specs fail at the moment (local failures here https://gist.github.com/smlance/7be52e21745eae585271); it's possible this is related to https://github.com/bundler/bundler/commit/3202988684406fbd2f03a83c281fb9811d3d2c9a.

lib/bundler/cli/install.rb
smlance: Forgot about that -- thanks!

lib/bundler/source/rubygems.rb
smlance: Absolutely. This is a WIP PR.

lib/bundler/cli/install.rb
smlance: Actually, why? They're distinct errors.

segiddins: I meant changing `Please choose` to `Please only use one`

smlance:
The specs should pass now, but this needs to be rebased onto 2-0-dev.

indirect:
@smlance can you use git to squash this down to a single commit?

smlance:
Coming right up!

segiddins:
:+1: . @indirect r?

indirect:
@homu r+

PR #3831, opened 2015-07-16
[CLI] Print help as the default no-args command

segiddins:
Addresses https://trello.com/c/OpuOdTZl/112-print-help-when-bundle-is-run-without-arguments.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -34,7 +34,7 @@ module Bundler
     check_unknown_options!(:except => [:config, :exec])
     stop_on_unknown_option! :exec
 
-    default_task :install
+    default_command :help
     class_option "no-color", :type => :boolean, :desc => "Disable colorization in output"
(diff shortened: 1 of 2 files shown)


segiddins:
Spec failures seem to be existing on 2-0-dev, @smlance ?

indirect:
Yeah, failures are existing. I'm hoping to take a look shortly.

smlance:
I can look into this as well.

chrismo:
my 2 cents would be to leave this on `install`. What's the desire for this behavior? I know that it would fall in line with a lot of tools that do this if no action is specified, but ... personally it'd be a bother to me to have to alias or just get used to typing bundle install each time

segiddins:
@chrismo I think you meant to put this on https://github.com/bundler/bundler/issues/5725?

chrismo:
yeah, i guess. i clicked the PR link in slack that seemed nearest to that conv. I'll move it - thx.

PR #3833, opened 2015-07-16
Considerate .gemrc when determine http_proxy.

PG-kura:
Bundler doesn't see http_proxy written in .gemrc. So I have to set environment var corresponding to each host. I think, usually, host specific setting of rubygem is customize and adjust using .gemrc.
This patch consider http_proxy written in .gemrc when determine proxy.

### DIFF
--- a/lib/bundler/fetcher.rb
+++ b/lib/bundler/fetcher.rb
@@ -172,6 +172,14 @@ module Bundler
       @fetchers ||= FETCHERS.map { |f| f.new(downloader, remote_uri, fetch_uri, uri) }
     end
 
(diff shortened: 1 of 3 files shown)


segiddins:
Could you please add a test for this fix?

segiddins:
Also, anything that calls into the  `Gem` module should probably live in `rubygems_integration.rb`

PG-kura:
I modified code and added test.

segiddins:
Merged via https://github.com/bundler/bundler/issues/3936.

PG-kura:
Thank you!!

PR #3837, opened 2015-07-16
Ensure configured relative paths are expanded from the Gemfile directory

smlance:
Fixes this failing spec: https://github.com/bundler/bundler/pull/3784#discussion-diff-34627198L94

The path code looks a bit scattered now, and it probably deserves some explanation.

### DIFF
--- a/lib/bundler/settings.rb
+++ b/lib/bundler/settings.rb
@@ -12,6 +12,10 @@ module Bundler
       @global_config = load_config(global_config_file)
     end
 
+    def root
+      @root
+    end
+
     def [](name)
       key = key_for(name)
       value = (@local_config[key] || ENV[key] || @global_config[key] || DEFAULT_CONFIG[name])
@@ -141,7 +145,7 @@ module Bundler
 
       if path = self[:path]
         path = "#{path}/#{Bundler.ruby_scope}" if path != Bundler.rubygems.gem_dir
-        File.expand_path(path)
+        path
       else
         File.join(@root, Bundler.ruby_scope)
       end
--- a/lib/bundler/settings.rb
+++ b/lib/bundler/settings.rb
@@ -165,10 +165,8 @@ module Bundler
         install_path = set_path
       else
          # all relative paths (configured by the user)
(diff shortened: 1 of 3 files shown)


indirect:
Yeah, I think it makes sense to consolidate this logic into Bundler.settings.path.

lib/bundler/settings.rb
```diff
@@ -130,21 +134,41 @@ def with
...
       if path = self[:path]
         path = "#{path}/#{Bundler.ruby_scope}" if path != Bundler.rubygems.gem_dir
-        File.expand_path(path)
+        set_path = path
+      else
+        set_path = File.join(@root, Bundler.ruby_scope)
+      end
+
+      if set_path == File.join(Bundler.settings.root, Bundler.ruby_scope)
+        # ?
```
segiddins: what's this comment saying?

lib/bundler/settings.rb
```diff
@@ -130,21 +134,41 @@ def with
...
+      elsif Pathname.new(set_path).absolute?
+        # all other absolute paths
+        install_path = set_path
       else
-        File.join(@root, Bundler.ruby_scope)
+         # all relative paths (configured by the user)
+        install_path = File.join(Bundler::root, Bundler.settings.path)
       end
+
+      install_path
```
segiddins: no need for this statement, as all the cases of the conditional end with the return value

lib/bundler/settings.rb
smlance: When I first wrote this code, line 157 seemed useful, but I forgot why I added it. I need to sit down, refactor, and comment this block of code later, because it's not very transparent right now.

indirect:
Once the build passes, I'm happy to merge this in order to get 2-0-dev green. Let's plan to get the TODO refactoring done before we ship 2.0 though. :)

indirect:
@homu r+

indirect:
:'(

smlance:
2-0-dev is green! :tada:

indirect:
:tada:

PR #3838, opened 2015-07-16
Update README.md.tt

ethanez713:
Under the "Development" header, the README was telling users to run "rake rspec" if the RSpec gem is being used. However, in lib/bundler/templates/newgem/Rakefile.tt, the rake task that runs RSpec is named 'spec', rather than 'rspec'. 

I changed the generated README text to read 'rake spec' instead of 'rake rspec'.

Also: Should I write a test for this change?

### DIFF
--- a/lib/bundler/templates/newgem/README.md.tt
+++ b/lib/bundler/templates/newgem/README.md.tt
@@ -26,7 +26,7 @@ TODO: Write usage instructions here
 
 ## Development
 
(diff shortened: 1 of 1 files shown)


maxjacobson:
+1, when I generated a gem with `bundle gem` and chose rspec, the README recommended running `rake rspec`, but the registered task name was `rake spec`

indirect:
Good catch, thanks!

PR #3846, opened 2015-07-17
Fix Lint/UnreachableCode

esasse:


segiddins:
:+1:

esasse:
@indirect I think you merged a broken PR. Sorry if I shouldn't have left it open. But I still didn't have the time to investigate why it was breaking in an older version of ruby and rubygems.

indirect:
If it only fails on a single build matrix entry, it is almost always a network error while talking to RubyGems.org... pretty sure we're safe. :)

PR #3847, opened 2015-07-17
Fix Style/TrailingComma

esasse:


segiddins:
IIRC we might need to disable this one for 1.8.7 compatibility? let's see what CI says

maxjacobson:
If so, rubocop can require the absence of a trailing comma https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/style/trailing_comma.rb#L24

esasse:
You were right @segiddins. What do you want me to do? Just disable this one or add the check to require absence of trailing comma as @maxjacobson suggests? I vote for Max idea.

segiddins:
I'll handle it, thanks for the PR though!

PR #3854, opened 2015-07-18
Fix Style/SpaceAroundOperators

esasse:


lib/bundler/source/git.rb
```diff
@@ -50,12 +50,7 @@ def hash
       end
 
       def eql?(o)
-        o.is_a?(Git)         &&
-        uri == o.uri         &&
-        ref == o.ref         &&
-        branch == o.branch   &&
-        name == o.name       &&
-        version == o.version &&
+        o.is_a?(Git) && uri == o.uri && ref == o.ref && branch == o.branch && name == o.name && version == o.version &&
```
segiddins: don't love this

segiddins:
In general, I prefer this, but I'll let the rest of the team chime in on this one

lib/bundler/source/git.rb
esasse: Me neither, but unless you do it, rubocop will fail. I can investigate if it's possible to rubocop ignore just this part of code, but maybe it's not a good idea.

segiddins: I mean we can still have them each on their own line

esasse: Done! :+1:

indirect:
I wish there was a way to enforce this without breaking the places where there's multi-line alignment, but I'd rather have enforced spaces than the ability to align things across lines.

PR #3856, opened 2015-07-18
downgrade invalid specs to a warning for now

indirect:
since invalid specs were allowed by 1.0-1.9, this change allows them in all 1.x versions. we'll go back to throwing an exception on invalid gemspecs in Bundler 2.0.

/cc @segiddins @smlance

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -367,11 +367,15 @@ module Bundler
         else
           spec = eval_gemspec(path, contents)
         end
-        Bundler.rubygems.validate(spec) if spec && validate
+        warn_on_invalid_specs(spec) if validate
         spec
       end
+    end
+
+    def warn_on_invalid_specs(spec)
+      spec && Bundler.rubygems.validate(spec)
     rescue Gem::InvalidSpecificationException => e
-      raise InvalidOption, "The gemspec at #{file} is not valid. " \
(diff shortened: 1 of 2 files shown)


segiddins:
:cry: I don't _really_ approve of this change, since it leaves bundler open to broken gemspecs, which is exactly what we were trying to fix

indirect:
I completely agree with that sentiment. :cry: It's really important to me that we not break working Bundles in minor releases, and it turns out that this started breaking previously-working bundles. Since we don't have a practical way to only target the broken bundles, and 2.0 is imminent, let's just blanketly implement the change there.

segiddins:
I don't think that backwards compatibility means we guarantee _every_ bundle work, merely that every _supported_ behavior remain so.

indirect:
You're totally right that simple backwards compatibility does not obligate us to do this. This is about taking care of users, making sure that they always feel like they can safely update without breaking anything. That's something specific to Bundler, and I think that Bundler has seen the level of adoption and willingness to upgrade that it has today because of that that level of commitment to avoiding breakage.

indirect:
Interestingly, it turns out that the spec that checks raised errors on invalid specs still passes even if only a warning is printed instead. We'll want to keep the `expect(out).to_not include("ERROR REPORT")` when we revert this on 2.0.

PR #3857, opened 2015-07-18
Fix Rubocop offenses

esasse:


esasse:
@indirect I've squashed that commit like you asked.

indirect:
After merging all your other fixes, I think this either needs to be rebased or just --auto-correct'ed again. Thanks for all your work on this!

esasse:
I think you should be able to merge now. I haven't used rebased that much, so I hope I'm doing it right. :smile:

indirect:
:+1:

PR #3863, opened 2015-07-21
Change version number to 2.0.0.dev

smlance:


segiddins:
Only issue with this is that, technically, `2.0.0.dev` > `2.0.0.beta.1`, otherwise this should be good to go

indirect:
We use 2.0.0.pre.1, so it's all good. :star:

segiddins:
@homu r+

segiddins:
@homu retry

PR #3864, opened 2015-07-21
Switch to speedy new Travis boxes

indirect:


indirect:
@homu r+

.travis.yml
```diff
@@ -1,4 +1,5 @@
 language: ruby
+sudo: 9000
```
segiddins: what exactly does this change?

indirect: It switches our builds to the beta Travis builders running on Gogle Compute. They can finish a Bundler build in about 10-11 minutes, which is 2x the speed of the other builders.

segiddins: Cool!

segiddins:
@homu r+

indirect:
Starting to think that Travis isn't running on the auto branch, and so homu will never succeed...

segiddins:
Any reason this is into 2-0-dev instead of master?

segiddins:
@homu r+

indirect:
The beta infra doesn't have Ruby 1.8 on it. :/

indirect:
Hmm seems like the tests must still not be running?

PR #3865, opened 2015-07-21
Support :mri_23 platform

amatsuda:
Current ruby trunk is versioned as "2.3.0".

```
% ruby -v -e 'p RUBY_VERSION'
ruby 2.3.0dev (2015-07-21 trunk 51319) [x86_64-darwin14]
"2.3.0"
```

We'd like to run tests on some gems against ruby-trunk, so here's a support for MRI 2.3.

Also, added some missing platform mappings for 2.2 series.


indirect:
oh man, this is reaching ludicrous proportions. hopefully we can fix this in 2.x. :sweat_smile: 

Thanks for the patch!

indirect:
@homu r+

segiddins:
@homu r+

PR #3867, opened 2015-07-21
bin should be exe in `alias dbundle` instructions

mattbrictson:
It looks like bundler's executable was moved from `bin` to `exe` at some point but this documentation was not updated. This tripped me up when I was setting up my dev environment for the first time today. An easy fix!


segiddins:
:+1: 
Thanks @mattbrictson , forgot to make this change last week
@homu r+

PR #3871, opened 2015-07-22
[Resolver] Allow mis-matches in bundler requirements, and warn

segiddins:
\c @smlance @indirect


indirect:
Can we avoid an extra loop over all the deps by warning around line 164?

segiddins:
Not really, since we won't know if we've actually activated said spec at that point.

indirect:
Alright then, @homu r+

segiddins:
Failing spec fixed

indirect:
@homu r+

segiddins:
@homu retry

PR #3872, opened 2015-07-22
Summarize the number of using gems

hbd225:
For this issue bundler/bundler-features#33
Thanks for making great gem :)

### DIFF
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -105,6 +105,7 @@ module Bundler
       Installer.install(Bundler.root, definition, options)
       Bundler.load.cache if Bundler.app_cache.exist? && options["cache"] && !Bundler.settings[:frozen]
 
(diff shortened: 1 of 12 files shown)


lib/bundler/cli/install.rb
```diff
@@ -105,6 +105,7 @@ def run
       Installer.install(Bundler.root, definition, options)
       Bundler.load.cache if Bundler.app_cache.exist? && options["cache"] && !Bundler.settings[:frozen]
 
+      Bundler.ui.confirm "Using #{Installer.using_gems.size} already installed gems" if Installer.using_gems?
```
segiddins: just use `Installer.using_gems.size > 0`?

hbd225: Fixed.

segiddins:
:+1:

indirect:
@homu r+

PR #3878, opened 2015-07-25
Make test suite work with local git configs

cdwort:
On my work laptop, I was seeing failures in `spec/commands/newgem_spec.rb` due to the presence of  local `.git/config` files.

On my work computer, my global config is my work (private gitlab) credentials. Each github repo has a local config with my “cdwort” credentials. For example, I have one placed in `/Path/To/Bundler/.git/config`. 

Unfortunately, when a new gem is created within the test file noted above, it is created within the /Path/To/Bundler/ directory. This means it picks up my local config. All the tests for when there is no git config information fail because instead of getting nothing (which those tests expect because the test overrides my global amy.unger@corporatecompany.com setting), it gets my GitHub email.

I am using the `--remove-section` flag because if you only `--unset user.email` and `--unset user.name`, you get the following junk in your .git/config file. Yes, at the very bottom is the appropriate config.

screen shot 2015-07-24 at 11 51 41 pm

This has one drawback - it rearranges your file. I prefer my `[user]` section at the top of the file (good for troubleshooting when I've done stupid things). This places it at the bottom of the file, because the section is removed and then recreated each time for each test.

I'll take that trade-off for running my tests on my work laptop, but I can see that this is unexpected behavior and might prohibit this PR from going forward. Thoughts?

### DIFF
--- a/spec/commands/newgem_spec.rb
+++ b/spec/commands/newgem_spec.rb
@@ -28,7 +28,7 @@ describe "bundle gem" do
   end
 
   after do
-    `rm "#{ENV['GIT_CONFIG']}"` if File.exist?(ENV["GIT_CONFIG"])
+    `rm "#{ENV["GIT_CONFIG"]}"` if File.exist?(ENV["GIT_CONFIG"])
     ENV["GIT_CONFIG"] = @git_config_location
   end
 
--- a/spec/commands/newgem_spec.rb
+++ b/spec/commands/newgem_spec.rb
@@ -16,15 +16,20 @@ describe "bundle gem" do
   end
 
   before do
-    @git_name = `git config --global user.name`.chomp
-    `git config --global user.name "Bundler User"`
-    @git_email = `git config --global user.email`.chomp
-    `git config --global user.email user@example.com`
+    git_config_content = <<-EOF
+    [user]
+      name = "Bundler User"
+      email = user@example.com
+    EOF
+    @git_config_location = ENV["GIT_CONFIG"]
+    path = "#{File.expand_path("../../tmp", File.dirname(__FILE__))}/test_git_config.txt"
+    File.open(path, "w") {|f| f.write(git_config_content) }
+    ENV["GIT_CONFIG"] = path
   end
 
   after do
-    `git config --global user.name "#{@git_name}"`
-    `git config --global user.email "#{@git_email}"`
+    `rm "#{ENV['GIT_CONFIG']}"` if File.exist?(ENV["GIT_CONFIG"])
+    ENV["GIT_CONFIG"] = @git_config_location
   end
 
   shared_examples_for "git config is present" do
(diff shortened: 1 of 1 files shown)


segiddins:
Oooooh actually, maybe we could use the `GIT_CONFIG` environment variable to force it to use a scratch file?

cdwort:
Yes, if that's a thing! Not sure from looking at this: https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables

segiddins:
from `man git-config`
```
ENVIRONMENT
       GIT_CONFIG
           Take the configuration from the given file instead of .git/config. Using the "--global" option forces this to ~/.gitconfig. Using the "--system"
           option forces this to $(prefix)/etc/gitconfig.

       GIT_CONFIG_NOSYSTEM
           Whether to skip reading settings from the system-wide $(prefix)/etc/gitconfig file. See git(1) for details.
```

cdwort:
Oh, danke. I'll look into that tomorrow. That sounds like a far better option.

cdwort:
Alright, so this is now working without altering either the global or the local config.

It does not:
- Alter the user's global git config
- Alter the user's local git config
- Alter the "GIT_CONFIG" env variable

It does, however, cause the test suite to be littered with:

```
fatal: BUG: lockfile filename corrupt
fatal: BUG: lockfile filename corrupt
fatal: BUG: lockfile filename corrupt
```

I'll work on that if this looks like a strategy we want to pursue. Personally, I think this is cleaner than mucking about in the user's config files.

indirect:
If we can do this without messing with user files, that would be great. 👍 from me

cdwort:
Yeah, it's a tradeoff, but I'm happier messing with environment variables than I am with messing with files.

cdwort:
Err, I suppose my comment above needs clarification:

With this PR, this file now does not:

- Alter the user's global git config
- Alter the user's local git config
- Result in a change of the "GIT_CONFIG" env variable once the `after` block is run.

cdwort:
Test suite is happy now locally, without code alteration. 

I was altering ENV variables on the command-line and had a beautiful moment where I forgot that setting a variable to empty string is not the same as `unset`ing it. 

@segiddins @indirect If the above looks :+1: then can we get a CI build? Also, would you like the two commits on this PR squashed?

indirect:
Normally, I'm pretty indifferent to squashing, but in this case the second commit is by "Foo Bar", so a squash would be good. :)

segiddins:
Yay, looks great @cdwort ! Will :+1: when squashed.

cdwort:
Hm, seems like Rubocop is unhappy about single quotes even though by the message, I don't think it should be:

```
spec/commands/newgem_spec.rb:31:16: C: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
    `rm "#{ENV['GIT_CONFIG']}"` if File.exist?(ENV["GIT_CONFIG"])
               ^^^^^^^^^^^^
```

I'll fix, but I wonder if Rubocop is confused by the backtic?

indirect:
Oh. It's not confused by the backtick—we have rubocop set to enforce double-quotes everywhere (and double quotes inside interpolation inside double quotes don't need to be escaped).

cdwort:
Heh, yup, just realized the latter point.

cdwort:
All traces of Foo Bar gone! Should be ready for CI.

indirect:
@homu r+

segiddins:
@homu retry

PR #3879, opened 2015-07-25
Always check that subcommands succeeded

segiddins:
@indirect @smlance I need help finishing this up

### DIFF
--- a/lib/bundler/settings.rb
+++ b/lib/bundler/settings.rb
@@ -2,7 +2,7 @@ require "uri"
 
 module Bundler
   class Settings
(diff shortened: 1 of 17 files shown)


indirect:
What are we doing here? Just testing the exit codes for subcommands?

segiddins:
@indirect yeah. Generally, the idea is to make the tests more robust by forcing us to explicitly call out the subprocesses that are expected to fail, just like we would have to for normal method invocations

segiddins:
If I finish this up, you good with it @indirect ?

indirect:
yeah, I'm good with this.

PR #3881, opened 2015-07-27
fix bug that causes "run `rake false`" to appear in README.md

roseweixel:
If `BUNDLE_GEM__TEST: false` or `BUNDLE_GEM__TEST: "false"` are in `~/.bundle/config`, the README.md that is generated when creating a new gem includes instructions under the Development heading that say "Then, run \`rake false\` to run the tests."

The logic on line 29 of  README.md.tt (`if config[:test]`) fails because `config[:test]` returns the string "false" instead of the boolean.

As a patch, I changed the logic of `Bundler::Settings#[]` to convert the string "false" to a boolean.

It seems that the root of the issue is that `Bundler::Settings::BOOL_KEYS` does not include "gem.test", and it can't because it is only sometimes a boolean (when it is not 'rspec' or 'minitest'). A future fix for this could be to have two separate options, a boolean for whether or not the gem has tests and another for the type of test framework.

### DIFF
--- a/lib/bundler/settings.rb
+++ b/lib/bundler/settings.rb
@@ -17,9 +17,11 @@ module Bundler
       value = (@local_config[key] || ENV[key] || @global_config[key] || DEFAULT_CONFIG[name])
 
       case
-      when !value.nil? && is_bool(name)
+      when value.nil?
+        nil
+      when is_bool(name) || value == "false" 
         to_bool(value)
-      when !value.nil? && is_num(name)
+      when is_num(name)
         value.to_i
       else
         value
--- a/lib/bundler/settings.rb
+++ b/lib/bundler/settings.rb
@@ -19,7 +19,7 @@ module Bundler
       case
       when value.nil?
         nil
-      when is_bool(name) || value == "false" 
+      when is_bool(name) || value == "false"
         to_bool(value)
       when is_num(name)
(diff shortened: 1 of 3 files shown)


indirect:
This is a great patch, thanks! :grin: If you're interested in contributing more to Bundler in the future, drop me an email, and I'll invite you to the Bundler Slack!

indirect:
@homu r+

lib/bundler/settings.rb
```diff
@@ -17,9 +17,11 @@ def [](name)
       value = (@local_config[key] || ENV[key] || @global_config[key] || DEFAULT_CONFIG[name])
 
       case
-      when !value.nil? && is_bool(name)
+      when value.nil?
+        nil
+      when is_bool(name) || value == "false"
```
segiddins: trailing whitespace here

roseweixel:
bc94ea0 fixes the whitespace issue - should be ok now!

segiddins:
@homu r+

segiddins:
@homu retry

indirect:
Okay, I'm tired of non-deterministic failures. Merging! Thanks again for the fix.

roseweixel:
Thanks for merging, happy to contribute!

PR #3882, opened 2015-07-27
Experimental fast setup

byroot:
Just to be clear I don't expect a merge at this point, I'd just like inputs on the overall feasibility of the thing.

### Problem

To make gems accessible, Bundler populate the `$LOAD_PATH`, then when `require` is called, the VM try to locate the
required file in all the `$LOAD_PATH` entries one by one. Long story short Ruby's require is `O(n)`, `n` being the `$LOAD_PATH.size`.

### Acknowledgements 

- The list of files that can be required in single gem is unlikely to change at runtime.

- Therefore the list of files that can be required in the sum of all the gems is simply function of the `$LOAD_PATH` and can be cached.

- `require` is basically `O(1)` when provided with an absolute path

### Solution

This patch, during bundler setup, compute the list of all the files that can be required, and store it in a `Hash[relative_path => absolute_path]`

The cache is serialized in a file, with the all the gems `load_paths` as a cache key. If a single gem is added / removed / updated, the cache key change.

Then `require` is monkey patched to absolutize paths, using the cache.

Finally, bundler don't add the gems to the `$LOAD_PATH`, so that the `require` for code not loaded by bundler is not slowed down.

### Drawbacks

Some c extensions gems call the native `rb_require` or `rb_autoload` directly, therefore, the files they require must be kept in the `$LOAD_PATH`.
Gems can be blacklisted as follow `Bundler::CachedRuntime.blacklist += %w(snappy oj stackprof)`.

Unfortunately, this issue prevent this cached strategy from being the default one, that is why this patch is purely opt-in.

### Bugs

I haven't quite figured it out why yet, but `autoload` when called from the main object, doesn't behave properly.
I made it call require directly instead. That just "works" for now. But I'm convinced I can make it work.

### Benchmark

Benchmark can be found here: https://github.com/byroot/fast-bundler-demo 

For a plain new Rails app with 53 gems, the improvement is barely 4%:

```sh
$ bundle install
Bundle complete! 11 Gemfile dependencies, 53 gems now installed

$ ./script/benchmark
== Booting without cache:
RequireCache disabled
$LOAD_PATH.size => 77
        1.73 real         1.47 user         0.24 sys

== Warming cache (unless already present):
RequireCache enabled
$LOAD_PATH.size => 26
        1.67 real         1.47 user         0.18 sys

== Fast boot:
RequireCache enabled
$LOAD_PATH.size => 26
        1.67 real         1.47 user         0.19 sys
```

But for a sligthly bigger app with 136 gems, the app boot 20% faster.

```sh
$ bundle install
Bundle complete! 65 Gemfile dependencies, 136 gems now installed.

$ ./script/benchmark
== Booting without cache:
RequireCache disabled
$LOAD_PATH.size => 184
        3.29 real         2.59 user         0.69 sys

== Warming cache (unless already present):
RequireCache enabled
$LOAD_PATH.size => 39
        2.74 real         2.44 user         0.29 sys

== Fast boot:
RequireCache enabled
$LOAD_PATH.size => 39
        2.62 real         2.33 user         0.27 sys
```

I tested it with a real world app with 324 gems and lots of initializers, the saving is around 25%.

```sh
RequireCache disabled
$LOAD_PATH.size => 439
     14.14 real        10.85 user         3.19 sys

RequireCache enabled
$LOAD_PATH.size => 95
     10.79 real         9.52 user         1.18 sys
```

### DIFF
--- a/lib/bundler/cached_runtime.rb
+++ b/lib/bundler/cached_runtime.rb
@@ -14,8 +14,8 @@ module Bundler
       self
     end
 
-    def cache
-      @cache ||= RequireCache.new(@load_paths)
+    def require_cache
+      @require_cache ||= RequireCache.new(@load_paths)
     end
 
     private
@@ -23,7 +23,7 @@ module Bundler
         super if self.class.blacklist.include?(spec.name)
         @load_paths ||= []
         @load_paths += spec.load_paths
-        @cache = nil
+        @require_cache = nil
       end
   end
 end
--- a/lib/bundler/cached_runtime.rb
+++ b/lib/bundler/cached_runtime.rb
@@ -1,0 +1,29 @@
+require 'bundler/runtime'
+require 'bundler/require_cache'
+
+module Bundler
+  class CachedRuntime < Runtime
+    class << self
+      attr_accessor :blacklist
+    end
+    self.blacklist = []
+
+    def setup(*)
+      super
+      require_relative 'require_patch'
+      self
+    end
+
+    def cache
+      @cache ||= RequireCache.new(@load_paths)
+    end
+
+    private
+      def register_load_paths(spec)
+        super if self.class.blacklist.include?(spec.name)
+        @load_paths ||= []
+        @load_paths += spec.load_paths
+        @cache = nil
+      end
+  end
+end
--- a/lib/bundler/require_patch.rb
+++ b/lib/bundler/require_patch.rb
@@ -1,0 +1,33 @@
+class Module
+  def autoload_with_cache(const, path)
+    autoload_without_cache(const, Bundler.load.cache[path] || path)
+  end
+
+  alias_method :autoload_without_cache, :autoload
+  alias_method :autoload, :autoload_with_cache
+end
+
+class << Kernel
+  def require_with_cache(path)
+    require_without_cache(Bundler.load.cache[path] || path)
+  end
+
+  alias_method :require_without_cache, :require
+  alias_method :require, :require_with_cache
+end
+
+module Kernel
+  def require_with_cache(path)
+    require_without_cache(Bundler.load.cache[path] || path)
+  end
+
+  alias_method :require_without_cache, :require
+  alias_method :require, :require_with_cache
+
+  def autoload_with_cache(const, path)
+    autoload_without_cache(const, Bundler.load.cache[path] || path)
+  end
+
+  alias_method :autoload_without_cache, :autoload
+  alias_method :autoload, :autoload_with_cache
+end
--- a/lib/bundler/require_patch.rb
+++ b/lib/bundler/require_patch.rb
@@ -1,6 +1,6 @@
 class Module
   def autoload_with_cache(const, path)
-    autoload_without_cache(const, Bundler.load.cache[path] || path)
+    autoload_without_cache(const, Bundler.load.require_cache[path] || path)
   end
 
   alias_method :autoload_without_cache, :autoload
@@ -9,7 +9,7 @@ end
 
 class << Kernel
   def require_with_cache(path)
-    require_without_cache(Bundler.load.cache[path] || path)
+    require_without_cache(Bundler.load.require_cache[path] || path)
   end
 
   alias_method :require_without_cache, :require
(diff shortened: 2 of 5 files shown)


segiddins:
Why not just make this change in RubyGems?

byroot:
> Why not just make this change in RubyGems?

To be honest I'm not quite sure what rubygems does exactly these days.

But what I wanted to do was to reduce the size of the `$LOAD_PATH`, and it's bundler that is responsible to populate it, so it seemed logical to me to start here. But I could totally be missing something.

segiddins:
I don't think we can ever _not_ populate the load path, and rubygems already monkey patches `Kernel#require` to deal with gems, so ¯\_(ツ)_/¯

byroot:
> I don't think we can ever not populate the load path

Why is that? If you look at the demo repository I linked in the description, You'll see that even Rails app can boot without populating the load path. Appart from native extensions that bypass the API, it works just fine.

> rubygems already monkey patches Kernel#require to deal with gems, so ¯_(ツ)_/¯

Yes but AFAIK rubygems have little to no knowledge about which gems are part of the bundle.

segiddins:
No, the list of activated specs is known to rubygems. And I think it breaks the contract of `require` if it _doesnt_ search the `LOADPATH`

indirect:
This is pretty interesting to me! It's clearly not compatible with the RubyGems way of doing things, but neither is Bundler's standalone mode. As an incompatible opt-in performance improvement, I think it might be even better than standalone mode is.




If we had the names of native extensions (say, as gem metadata perhaps?) do you think we could require those directly without needing to put them on the load path?

segiddins:
(It's also incompatible with a user manually manipulating the load path, unshifting something that conflicts with a gem in their bundle)

byroot:
> If we had the names of native extensions (say, as gem metadata perhaps?) do you think we could require those directly without needing to put them on the load path?

Unfortunately no. Maybe I wasn't clear. The problem with some native extensions is that they call `rb_require` directly, so the LOAD_PATH is searched in C directly, without delegating to `Kernel#require`.

See https://github.com/ohler55/oj/blob/0eea2cb689f0fe972595e2302e0f26474e49fa7e/ext/oj/oj.c#L2075-L2079 for instance. 

The solution would be to call `Kernel#require` from C instead of `rb_require`.

`autoload` have the same issue in `> 2.3`: https://bugs.ruby-lang.org/issues/11140

It's just a matter of using `rb_funcall(rb_mKernel, rb_intern("require"), 1, rb_str_new2("foo"));` instead of `rb_require("foo")`.

On a longer term, I think it would make sense for MRI to expose another API than `$LOAD_PATH` for library inclusion.

byroot:
> (It's also incompatible with a user manually manipulating the load path, unshifting something that conflicts with a gem in their bundle)

How so? If you look here: https://github.com/bundler/bundler/pull/3882/files#diff-a9429187b3bcb02b801b757f0a8fef61R37

You'll see that the remaining `$LOAD_PATH` will be looked up first, and then on miss only, the gems in the bundler will be looked up. 

I didn't notice any compatibility breakage beside the native `rb_require` issue. And I tested with a fairly complex Rails app of 300+ gems. And we're running an alternative version of this patch since a couple months in production successfully.

vidarh:
An alternative that I've tested in an app with startup times >30 seconds (slow disks, lots of gems) is to wrap the calls to the underlying require with resetting $LOAD_PATH to what it was before + just the paths to the gems that the gem you're currently loading wrong requires. That handles the aforementioned C-extensions etc. just fine, though there are still potential issues e.g. with gems that expect to be in the load path later during runtime. 

I'm looking forward to testing this code - bundlers $LOAD_PATH mangling leads to some really pathological cases in larger apps (the app mentioned earlier had ca. 100k failed stat calls while trying to look up files before I put in place a hacky solution based on wrapping as above, and less than 30k afterwards)

lib/bundler/cached_runtime.rb
```diff
@@ -0,0 +1,29 @@
...
+module Bundler
+  class CachedRuntime < Runtime
+    class << self
+      attr_accessor :blacklist
+    end
+    self.blacklist = []
+
+    def setup(*)
+      super
+      require_relative 'require_patch'
```
segiddins: `require_relative` is no good on 1.8.7

lib/bundler/require_patch.rb
```diff
@@ -0,0 +1,34 @@
...
+module Kernel
+  def require_with_cache(path)
+    require_without_cache(Bundler.load.require_cache[path] || path)
+  end
+
+  alias_method :require_without_cache, :require
+  alias_method :require, :require_with_cache
+
+  def autoload_with_cache(const, path)
+    STDERR.puts "WARNING: Top level autoload is not properly supported yet, #{const} was eager loaded"
```
segiddins: Why not just have it autoload the absolute path?

byroot:
Closing in favor of https://github.com/byroot/bootscale

Ref: https://github.com/bundler/bundler/pull/3958#issuecomment-133896913

PR #3884, opened 2015-07-27
Run Rubocop before specs

TimMoore:
This fails faster on style issues.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -238,6 +238,9 @@ begin
     task :travis do
       rg = ENV["RGV"] || raise("Rubygems version is required on Travis!")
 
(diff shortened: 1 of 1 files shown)


indirect:
:+1: @homu +r

indirect:
lol oops @homu r+

segiddins:
@TimMoore I don't believe this will actually make it fail fast, since we're calling `safe_task` ?

indirect:
Yeah, I think that's true.

indirect:
How did this ever fail when rubocop failed, anyway? We weren't even checking for it on line 269.

indirect:
@homu r+

TimMoore:
Good catch, thanks @indirect & @segiddins.

PR #3889, opened 2015-07-28
Fix Style/WordArray

esasse:


indirect:
Can we use %w[word array] instead?

esasse:
Done. 
And maybe we can enforce %w[], do you want me to do it?

indirect:
@esasse yeah, that would be great.

esasse:
Looks like it's not possible to enforce it. But I'll keep looking and send another PR if needed, ok?

indirect:
Sounds good.

indirect:
@homu r+

esasse:
Looks like a GitHub error during build, I'm restarting it.

esasse:
git checkout error again, I'm not sure what to do. Any ideas?

PR #3890, opened 2015-07-29
Fix Style/AlignParameters

esasse:


indirect:
I really dislike this change :( Can we make rubocop enforce a single indent instead?

esasse:
"Yes, we can!"
And it's done! :smile:

indirect:
Seems less bad to me. :) Thanks! @homu r+

esasse:
Same git checkout issue here as in #3889. Any ideas?

PR #3892, opened 2015-07-29
Use CGI instead of URI for escaping URL params

EduardoBautista:
URI.encode prints a warning that it's obsolete.

Fixes #3892


indirect:
@homu r+

segiddins:
I'm not seeing any such warning?

PR #3893, opened 2015-07-30
Try failing fast if prereq gems fail to install

indirect:
One of the causes of builds that fail late is gems that didn't install but also didn't raise an error (until the tests that depend on them fail, much much later). This tries to fail those quickly, and maybe means that we can add some retry logic.

### DIFF
--- a/spec/support/sometimes.rb
+++ b/spec/support/sometimes.rb
@@ -1,0 +1,52 @@
+module Sometimes
+  def run_with_retries(example_to_run, retries)
+    example = RSpec.current_example
+    example.metadata[:retries] ||= retries
+    retries.times do |t|
+      example.metadata[:retried] = t + 1
+      example.instance_variable_set(:@exception, nil)
+      example_to_run.run
+      break unless example.exception
+    end
+    if e = example.exception
+      new_exception = e.exception(e.message + "[Retried #{retries} times]")
+      new_exception.set_backtrace e.backtrace
+      example.instance_variable_set(:@exception, new_exception)
+    end
+  end
+end
+
+RSpec.configure do |config|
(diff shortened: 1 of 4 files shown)


indirect:
@homu r+ (I want to see how the tests do, this still needs code review)

indirect:
@homu r+

spec/support/sometimes.rb
```diff
@@ -0,0 +1,52 @@
+module Sometimes
+  def run_with_retries(example_to_run, retries)
+    example = RSpec.current_example
```
segiddins: couldn't this just be `example_to_run`?

spec/support/sometimes.rb
```diff
@@ -0,0 +1,52 @@
...
+    }
+    retried_examples = RSpec.world.example_groups.map do |g|
+      g.descendants.map do |d|
+        d.filtered_examples.select do |e|
+          e.metadata[:sometimes] && e.metadata.fetch(:retried, 1) > 1
+        end
+      end
+    end.flatten
+    message.call(retried_examples.empty? ? :green : :yellow, "\n\nRetried examples: #{retried_examples.count}")
+    unless retried_examples.empty?
```
segiddins: no need for this check

spec/support/sometimes.rb
indirect: no, they are inexplicably different objects. :'(

spec/support/sometimes.rb
indirect: oh yeah, good catch

indirect:
@segiddins r?

segiddins:
@homu r+

PR #3896, opened 2015-07-30
Print a warning when a gem can't be cached due to a missing source

EduardoBautista:
I still need to create a test, but I am not entirely sure where to write it.

### DIFF
--- a/lib/bundler/source/rubygems.rb
+++ b/lib/bundler/source/rubygems.rb
@@ -189,7 +189,11 @@ module Bundler
         cached_path = cached_path(spec)
         if cached_path.nil?
           remote_spec = remote_specs.search(spec).first
-          cached_path = fetch_gem(remote_spec)
+          if remote_spec
+            cached_path = fetch_gem(remote_spec)
+          else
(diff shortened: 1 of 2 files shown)


segiddins:
I don't think this is correct, iirc we have a faux rubygems source to contain the local gems?

indirect:
@segiddins we paired on this, and it fixes the problem--built in gems don't have .gem files, and if there's no remote source with a copy, it causes the exception we've seen reported.

segiddins:
@indirect ah ok, cool

EduardoBautista:
So where should the test be written?

indirect:
I copied a cache-builtin spec and removed the source `e872985f4f4988437e3f47efe18842513bd62452`. @EduardoBautista can you cherry-pick that commit into this PR so we can run the tests?

EduardoBautista:
@indirect Done.

segiddins:
@homu r+

indirect:
Merged with fixed whitespace as 69475cb66c86c271494299e595b0ad3f04bbda31. Thanks!

PR #3900, opened 2015-07-31
path caching: make sure intersecting paths are copied

steved:
This is part bug-fix, part feature request.

971edd3 fixes a bug where a gem with a path that is similar to the bundled application path would not be copied on `package --all`. For example, a gem with path `/test_gem/` would not be copied into `/test/vendor/cache`.

570e60d is a feature request to preserve the buggy behavior with an option. In my case, it would be helpful because we can restart a docker container that has the gem path mapped in without having to rerun `package` and copy it into `vendor/cache`.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -239,6 +239,7 @@ module Bundler
     method_option "all",  :type => :boolean, :banner => "Include all sources (including path and git)."
     method_option "all-platforms", :type => :boolean, :banner => "Include gems for all platforms, not just the current one"
     method_option "no-prune",  :type => :boolean, :banner => "Don't remove stale gems from the cache."
+    method_option "no-copy-paths", :type => :boolean, :banner => "Do not include paths when caching gems using --all"
     def cache
       require "bundler/cli/cache"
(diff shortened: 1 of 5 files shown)


segiddins:
I don't really see why we need to add an option for this, as doing so would obligate us to support and test it for a while. The bug fix looks good, though!

indirect:
Thanks for the patch!

Honestly, it seems a little weird to me to have a flag where 50% of the reason it exists is to cache path deps, and then have yet another flag whose sole point is to cancel out that half of the first flag. If you don't want to cache path deps, you can always remove them from vendor/cache after you run `bundle pack`.

indirect:
Cherry-picked in #3907 as 7e8d399.

steved:
Fine by me, just putting it out there. Thanks.

grosser:
FYI solving this with:
 - not checking if path gems need to be packaged via [BundlePackageCheck](https://github.com/grosser/bundle_package_check) `ignore_path: true`
 - removing all path gems `rm -rf #{Dir["vendor/cache/*"].reject { |x| x.include?("-").join(" ") }`

PR #3904, opened 2015-08-02
Fix Style/ParallelAssignment

esasse:


segiddins:
:+1:

indirect:
This goes against my grumpy old-school rubyist, but it does seem easier to understand. Thanks.

PR #3905, opened 2015-08-02
remove unneeded conditional for path

asymmetric:
the `self[:path]` call already returns the configured path with the correct priority, i.e. local, env, global, default.


segiddins:
This seems to be causing some spec failures?

indirect:
Closing since the specs fail...

PR #3906, opened 2015-08-03
Use locked_requirement platform to print it out

indirect:


indirect:
@segiddins r?

lib/bundler/resolver.rb
```diff
@@ -3,12 +3,20 @@ class Resolver
     require "bundler/vendored_molinillo"
 
     class Molinillo::VersionConflict
+      def printable_dep(dep)
+        case dep
+        when DepProxy; dep.to_s.strip
+        when Bundler::Dependency; DepProxy.new(dep, dep.platforms.join(", ")).to_s.strip
+        else; raise("Don't know how to print #{dep.inspect}")
```
segiddins: Gem::Dep?

indirect: It seems we only ever have a Bundler::Dep (which is what defines #platforms) in our suite.

segiddins: I don't see a need to raise here, could fallback to `dep.to_s.strip` for everything but a `Bundler::Dependency` ?

indirect: Sure, works for me.

segiddins:
@homu r+

indirect:
@homu r=segiddins

segiddins:
@homu r+

indirect:
@homu retry

PR #3907, opened 2015-08-03
make sure intersecting gem paths are cached

indirect:
When a gem has the same path prefix as the application that is being bundled,
the gem will not be added to vendor/cache.

For example: a gem with a path /home/test/src/bundled_gem and an application with a path
/home/test/src/bundled that references it.

### DIFF
--- a/lib/bundler/source/path.rb
+++ b/lib/bundler/source/path.rb
@@ -77,7 +77,7 @@ module Bundler
       def cache(spec, custom_path = nil)
         app_cache_path = app_cache_path(custom_path)
         return unless Bundler.settings[:cache_all]
-        return if expand(@original_path).to_s.index(Bundler.root.to_s) == 0
+        return if expand(@original_path).to_s.index(Bundler.root.to_s + '/') == 0
 
         unless @original_path.exist?
           raise GemNotFound, "Can't cache gem #{version_message(spec)} because #{self} is missing!"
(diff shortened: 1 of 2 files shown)


indirect:
@homu r+

indirect:
@homu r+

lib/bundler/source/path.rb
```diff
@@ -77,7 +77,7 @@ def install(spec, force = false)
       def cache(spec, custom_path = nil)
         app_cache_path = app_cache_path(custom_path)
         return unless Bundler.settings[:cache_all]
-        return if expand(@original_path).to_s.index(Bundler.root.to_s) == 0
+        return if expand(@original_path).to_s.index(Bundler.root.to_s + '/') == 0
```
segiddins: double quotes

spec/cache/path_spec.rb
```diff
@@ -29,6 +29,24 @@
       should_be_installed "foo 1.0"
     end
 
+    it "copies when the path is outside the bundle and the paths intersect" do
+      libname = File.basename(Dir.pwd) + '_gem'
```
segiddins: double quotes

indirect:
@homu r+

indirect:
@homu r+

segiddins:
@homu r+

indirect:
@homu retry

PR #3911, opened 2015-08-04
Add command plugin support to help command

tf:
Bundler already delegates to executables of the form `bundler-<command>` when called as `bundle <command>` with an unknown command. Allow viewing help entries for such command plugins by
turning calls of the form `bundle help <command>` into `bundler-<command> --help`.


indirect:
Thanks!

PR #3913, opened 2015-08-05
Fix for warning on jruby 9k

mastfish:
Jruby 9k is ruby 2.2.2, but doesn't support processes. 

So running `bundle exec` leads to a warning `warning: unsupported exec option: close_others`.

This should fix that warning.

### DIFF
--- a/lib/bundler/cli/exec.rb
+++ b/lib/bundler/cli/exec.rb
@@ -9,7 +9,7 @@ module Bundler
       @cmd = args.shift
       @args = args
 
-      if Bundler.current_ruby.ruby_2?
+      if Bundler.current_ruby.ruby_2? && (!Bundler.current_ruby.jruby?)
         @args << { :close_others => !options.keep_file_descriptors? }
       elsif options.keep_file_descriptors?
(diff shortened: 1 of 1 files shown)


lib/bundler/cli/exec.rb
```diff
@@ -9,7 +9,7 @@ def initialize(options, args)
       @cmd = args.shift
       @args = args
 
-      if Bundler.current_ruby.ruby_2?
+      if Bundler.current_ruby.ruby_2? && (!Bundler.current_ruby.jruby?)
```
segiddins: parentheses shouldn't be necessary

mastfish: Parentheses have been removed

indirect:
Thanks!

perlun:
Any chance we could get a release including this fix? It's quite annoying to see the warning message all the time... :smile:

thomasfedb:
I'd also like to see this fix release. https://github.com/carrierwaveuploader/carrierwave/issues/1814

perlun:
FYI: The JRuby warning is gone with recent versions of `bundler`, methinks. But not your issue.

PR #3918, opened 2015-08-06
Tweaks git checkout UI output

jordanthomas:
Tweaking the "Updating ..." output of the checkout method to reduce
confusion when running `outdated`. Opting instead for "Fetching" as it
does when doing a clone operation.

### DIFF
--- a/lib/bundler/source/git/git_proxy.rb
+++ b/lib/bundler/source/git/git_proxy.rb
@@ -66,7 +66,7 @@ module Bundler
         def checkout
           if path.exist?
             return if has_revision_cached?
-            Bundler.ui.confirm "Updating #{uri}"
+            Bundler.ui.info "Fetching #{uri}"
             in_path do
               git_retry %|fetch --force --quiet --tags #{uri_escaped_with_configured_credentials} "refs/heads/*:refs/heads/*"|
             end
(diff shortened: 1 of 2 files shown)


lib/bundler/source/git/git_proxy.rb
```diff
@@ -66,7 +66,7 @@ def version
         def checkout
           if path.exist?
             return if has_revision_cached?
-            Bundler.ui.confirm "Updating #{uri}"
+            Bundler.ui.info "Fetching #{uri}"
```
segiddins: Why the change from `confirm` to `info` here?

jordanthomas: The colored output of `confirm` looked out of place next to the other "Fetching" outputs. It was also unclear to me what the coloring was trying to convey when the same method doesn't use `confirm` when cloning.

Picture from my machine for reference:
bundler

indirect: I agree that git updates should be info rather than confirm—in the `install` output, confirm means a gem is being installed, while info means a gem is already installed and is simply being used.

segiddins: :+1:

indirect:
@homu r+

PR #3920, opened 2015-08-08
[RuboCop] Update to 0.33.0

segiddins:


lib/bundler/graph.rb
```diff
@@ -135,7 +135,7 @@ def run
         @groups.each do |group|
           g.add_nodes(
             group,
-            { :style     => "filled",
+            { :style => "filled",
```
indirect: this makes no sense to me... why is the first line not allowed to be aligned, but the rest are?

segiddins: oooooh good question, didn't notice that one (autocorrect)

segiddins:
@indirect we ought to get this merged so that `master` is back passing?

andremedeiros:
Looks good to me. :shipit:

indirect:
I was just waiting for some resolution on the weird "first line can't align but the other lines can" cop. I guess it expects the `{` to be on a line by itself.

segiddins:
@indirect probably, but I didn't think it was worth keeping the build broken over

PR #3921, opened 2015-08-08
[Rakefile] Use automatiek to manage vendoring

segiddins:
This uses my new gem to manage vendoring.

\c @indirect


indirect:
Looks good to me.

PR #3922, opened 2015-08-08
Fix Lint/UnusedMethodArgument

Koronen:
* ~~Remove non-referenced explicit block arguments in favor of depending on the   implicit block provided to every method.~~
* Remove unused method `Spec::Builders#build_dep`.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -25,11 +25,6 @@ Lint/NonLocalExitFromIterator:
 Lint/RescueException:
   Enabled: false
 
-# Offense count: 38
-# Cop supports --auto-correct.
-Lint/UnusedMethodArgument:
-  Enabled: false
-
 # Offense count: 7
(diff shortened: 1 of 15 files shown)


segiddins:
I don't like this, I far prefer explicit block arguments.

segiddins:
This is failing on CI

segiddins:
@indirect iirc you dislike this?

indirect:
I prefer explicit &block to make it clear that the method takes a block, and I prefer not prefixing variables with underscores. Maybe those are cases where we can improve the code, or maybe they are cases where it's pointless to enumerate the method arguments.

Koronen:
If we don't agree with the rule, maybe we should instead disable the rule in `.rubocop.yml` and remove it from `.rubocop_todo.yml`?

segiddins:
@Koronen yup

Koronen:
Closing in favor of #3956.

PR #3923, opened 2015-08-08
Fix a bunch of style offences

Koronen:
Most of these were low-hanging fruit, fixable by using RuboCop's auto-correct feature.

### DIFF
--- a/lib/bundler/gem_helper.rb
+++ b/lib/bundler/gem_helper.rb
@@ -158,7 +158,11 @@ module Bundler
 
     def sh(cmd, &block)
       out, code = sh_with_code(cmd, &block)
-      code == 0 ? out : raise(out.empty? ? "Running `#{cmd}' failed. Run this command directly for more detailed output." : out)
+      if code == 0
+        out
+      else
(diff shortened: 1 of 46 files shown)


lib/bundler/gem_helper.rb
```diff
@@ -83,7 +83,7 @@ def build_gem
 
     def install_gem(built_gem_path = nil, local = false)
       built_gem_path ||= build_gem
-      out, _ = sh_with_code("gem install '#{built_gem_path}'#{" --local" if local}")
+      out, _code = sh_with_code("gem install '#{built_gem_path}'#{" --local" if local}")
```
segiddins: this causes a warning on 1.9.3, please keep this as `_`

segiddins:
I'm :+1: on this other than the one comment, which is causing CI to fail

lib/bundler/gem_helper.rb
Koronen: Since this was the only warning for `Style/TrailingUnderscoreVariable`, I just dropped that commit for now.

segiddins:
@homu r+

segiddins:
@homu r+

PR #3924, opened 2015-08-08
Fix Style/RegexpLiteral

Koronen:


lib/bundler/source/path.rb
```diff
@@ -165,7 +165,7 @@ def load_spec_files
       end
 
       def relative_path
-        if path.to_s.match(%r{^#{Regexp.escape Bundler.root.to_s}})
+        if path.to_s.match(/^#{Regexp.escape Bundler.root.to_s}/)
```
Koronen: Maybe use `String#start_with?` here?
```ruby
if path.to_s.start_with?(Bundler.root.to_s)
```

segiddins:
@homu r+

PR #3928, opened 2015-08-09
Specify the actual exception to test for

agis-:
Gets rid of the following RSpec warning:

    Using the `raise_error` matcher without providing a specific error or
    message risks false positives, since `raise_error` will match when Ruby
    raises a `NoMethodError`, `NameError` or `ArgumentError`, potentially
    allowing the expectation to pass without even executing the method you
    are intending to call. Instead consider providing a specific error class
    or message.


indirect:
@homu r+

PR #3929, opened 2015-08-09
Be stricter when accepting rubygems source URIs

### LINKED ISSUE #3925: bundle command throws error
- What did you do?
- What happened instead?

Error details

    ArgumentError: no HTTP request path given
    /usr/local/Cellar/ruby/2.2.2/lib/ruby/2.2.0/net/http/generic_request.rb:24:in `initialize'
    /usr/local/Cellar/ruby/2.2.2/lib/ruby/2.2.0/net/http/request.rb:14:in `initialize'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/fetcher.rb:294:in `new'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/fetcher.rb:294:in `request'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/fetcher.rb:270:in `fetch'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/fetcher.rb:242:in `use_api'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/source/rubygems.rb:301:in `block in api_fetchers'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/source/rubygems.rb:301:in `select'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/source/rubygems.rb:301:in `api_fetchers'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/source/rubygems.rb:306:in `block in remote_specs'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/index.rb:9:in `build'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/source/rubygems.rb:305:in `remote_specs'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/source/rubygems.rb:79:in `specs'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/definition.rb:204:in `block (2 levels) in index'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/definition.rb:202:in `each'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/definition.rb:202:in `block in index'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/index.rb:9:in `build'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/definition.rb:199:in `index'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/definition.rb:193:in `resolve'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/definition.rb:132:in `specs'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/definition.rb:121:in `resolve_remotely!'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/installer.rb:79:in `run'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/installer.rb:18:in `install'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/cli/install.rb:80:in `run'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/cli.rb:157:in `install'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/vendor/thor/lib/thor/invocation.rb:126:in `invoke_command'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/vendor/thor/lib/thor.rb:359:in `dispatch'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/vendor/thor/lib/thor/base.rb:440:in `start'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/cli.rb:10:in `start'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/bin/bundle:20:in `block in <top (required)>'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/lib/bundler/friendly_errors.rb:7:in `with_friendly_errors'
    /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.9.4/bin/bundle:18:in `<top (required)>'
    /usr/local/bin/bundle:23:in `load'
    /usr/local/bin/bundle:23:in `<main>'

Environment

    Bundler   1.9.4
    Rubygems  2.4.5
    Ruby      2.2.2p95 (2015-04-13 revision 50295) [x86_64-darwin14]
    GEM_HOME  /usr/local/lib/ruby/gems/2.2.0
    GEM_PATH  /usr/local/lib/ruby/gems/2.2.0:/Users/Scosentino/.gem/ruby/2.2.0:/usr/local/Cellar/ruby/2.2.2/lib/ruby/gems/2.2.0
    Git       2.3.5

You can find the repo here https://github.com/Scosentino/towboatislamorada

agis-:
Since `URI#parse` accepts URIs such as *http:rubygems.org* as per RFC 2396,
we reject such sources manually early in the process, in order to provide more
helpful error messages.

Fixes #3925.

### DIFF
--- a/lib/bundler/source/rubygems.rb
+++ b/lib/bundler/source/rubygems.rb
@@ -262,7 +262,8 @@ module Bundler
         uri = uri.to_s
         uri = "#{uri}/" unless uri =~ %r'/$'
         uri = URI(uri)
-        raise ArgumentError, "The source must be an absolute URI" unless uri.absolute?
+        raise ArgumentError, "The source must be an absolute URI. For example:\n" \
+          "source 'https://rubygems.org'" if !uri.absolute? || (uri.is_a?(URI::HTTP) && uri.host.nil?)
         uri
       end
 
--- a/spec/bundler/source/rubygems_spec.rb
+++ b/spec/bundler/source/rubygems_spec.rb
@@ -22,4 +22,12 @@ describe Bundler::Source::Rubygems do
       end
     end
   end
+
+  describe "#add_remote" do
+    context "when the source is an HTTP(s) URI with no host" do
+      it "raises error" do
+        expect { subject.add_remote("https:rubygems.org") }.to raise_error(ArgumentError)
+      end
+    end
+  end
 end


lib/bundler/fetcher/downloader.rb
```diff
@@ -36,6 +36,7 @@ def fetch(uri, counter = 0)
 
       def request(uri)
         Bundler.ui.debug "HTTP GET #{uri}"
+        raise GemfileError, "Incorrect gem source URL: #{uri}" unless uri.request_uri
```
segiddins: This shouldn't be raising a `GemfileError`, and are you 100% certain that `uri.request_uri` will return nil instead of throwing?

indirect:
Can we validate the url-ness of source urls when we add them as sources in the DSL class?

agis-:
@indirect Sure, sounds better indeed; will get back soon with the changes.

agis-:
Updated to check the validity when the source URIs are added in the `DSL` class.

segiddins:
Seems good to me! Thanks, @agis- !

indirect:
Perfect. Thanks!

indirect:
@homu r+

agis-:
It seems this change breaks compatibility with sources such as *"file:///home/travis/build/bundler/bundler/tmp/gems/remote1"*, since they are absolute but don't have a host: https://travis-ci.org/bundler/bundler/jobs/76237612.

Should we exclude the check for host if the scheme is "file" or be explicit and only perform it when the scheme is "http" or "https"?

indirect:
Hmmmm. Yeah, I think that makes sense. You could even use `uri.schema == "file"`, I believe.

agis-:
Thing is, if I only use `uri.scheme == "file"` it may truly be an invalid URI such as `file:foo/bar` and the check would give a false positive, since:

```ruby
URI('file:///foo/bar').host # => nil
URI('file://foo/bar').host # => "foo"
```

If you mean using `uri.schema` *and* testing for that 3rd slash, then yes, this would also work, but I think it'd be redundant and we'd better do this in one pass (via a regex).

What do you think?

agis-:
Apparently, the suggested change breaks [tests] that rely on Ruby's buggy behavior:

```ruby
URI('file:///foo/bar').to_s # => file:/foo/bar
```

This bug is also described [here](https://bugs.ruby-lang.org/issues/9912). So if the provided rubygem source is of the format `file:///path/to/foo`, then the URI is erroneously converted to `file:/path/to/foo` [here](https://github.com/bundler/bundler/blob/master/lib/bundler/source/rubygems.rb#L262).

I assume that such URIs should not work with bundler at all, based on the current implementation. TBH I'm not sure how bundler plays correctly with such URIs all that time. Or does it?

I'm not sure what is the correct approach here (re. URIs that ommit the host). Maybe we should reject them in bundler alltogether?

agis-:
@indirect I'm starting to believe it doesn't worth the fuss to bother with `file` URIs. I've updated the code to only reject non-host addresses *only* if they're HTTP/HTTPS.

indirect:
Makes sense to me. I'm happy with that. Thanks! @homu r+

indirect:
@homu r+

segiddins:
ummmmm homu pretty sure they passed

PR #3937, opened 2015-08-14
Remove `bundle install --cache`

smlance:
Removes the --cache option and replaces it with an error that asks the user to run `bundle cache`.

### DIFF
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -16,6 +16,12 @@ module Bundler
         end
       end
 
+      if options[:cache]
+        Bundler.ui.error "Please use `bundle cache` instead of `bundle "\
+         "install --cache`."
+        exit 1
+      end
+
       if options[:without] && options[:with]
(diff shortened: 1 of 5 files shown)


lib/bundler/cli/install.rb
```diff
@@ -16,6 +16,12 @@ def run
         end
       end
 
+      if options[:cache]
+        Bundler.ui.error "Please use `bundle cache` instead of `bundle "\
```
segiddins: make sure this is a warning on 1.99

smlance:
This currently violates some RuboCop style guidelines, but it should pass after I rebase master onto 2-0-dev and rebase 2-0-dev onto this branch.

segiddins:
👍🏻

lib/bundler/cli/install.rb
segiddins: @smlance ^

segiddins:
@homu r+

PR #3938, opened 2015-08-14
[2.0] Don't persist install options

asymmetric:
Implements bundler/bundler-features#26.

Some specs are failing, but wanted to get  some early feedback.

* What do you think about resetting the `BUNDLE_WITH/WITHOUT` ENV variables after each spec?
* If that's OK, should setting `ENV` be the preferred method in most specs, compared to `bundle config`, `config`?

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -137,8 +137,6 @@ module Bundler
       "Force downloading every gem."
     method_option "no-prune", :type => :boolean, :banner =>
       "Don't remove stale gems from the cache."
-    method_option "path", :type => :string, :banner =>
-      "Specify a different path than the system default ($BUNDLE_PATH or $GEM_HOME)."
     method_option "quiet", :type => :boolean, :banner =>
       "Only output warnings and errors."
     method_option "shebang", :type => :string, :banner =>
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -143,8 +143,6 @@ module Bundler
       "Specify a different shebang executable name than the default (usually 'ruby')"
     method_option "standalone", :type => :array, :lazy_default => [], :banner =>
       "Make a bundle that can work without the Bundler runtime"
-    method_option "system", :type => :boolean, :banner =>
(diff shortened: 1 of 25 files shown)


segiddins:
How does this differ from https://github.com/bundler/bundler/commit/3d76c8fefb4eed7a082719e693396b7cdb0d58b0 ?

asymmetric:
@segiddins AFAICS, the actual implementation (i.e. removing flags, moving logic) is not in that commit, nor in the `2-0-dev` branch.

segiddins:
OK, @smlance will need to review this then :)

asymmetric:
@homu thanks, I'll take a look as soon as I come back from the holidays (~2 weeks).

smlance:
@asymmetric: I've been using parts of your code to tackle this and have already removed the cache and path flags. I'm currently reworking the way `bundle config` parses the with and without flags. See https://github.com/bundler/bundler/pull/3827, https://github.com/bundler/bundler/pull/3937, and https://github.com/bundler/bundler/pull/3939.

asymmetric:
@smlance Glad to have been of help (although attribution would have been even nicer :cry:)

I guess this PR can be closed then.

smlance:
https://github.com/bundler/bundler/pull/3939 uses the deletions you made in `cli/install.rb` and the line you added to `cli/config.rb`. If there's some way I can give more credit (@segiddins?), let me know. Thanks for the contributions!

segiddins:
@smlance the proper way to give attribution is to use someone's git commits.

indirect:
To expand on that, there are a few ways you can give attribution:

 - you can directly use the actual git commits via `git cherry-pick <SHA1>` (usually then adding your own commits later if needed)
 - you can use the author information from their commit when you commit what is mostly their change via `git commit -C <SHA1>` (if their commit didn't quite work as is)
 - worst case, you can explicitly acknowledge inspiration/code in the commit message

In this case, I think you could likely use some combination of cherry-picking commits and writing your own commits to get the same result, but it's hard to be sure without trying it myself.

PR #3941, opened 2015-08-15
Add comment describing Bundler.settings.path

smlance:
An improvement over https://github.com/smlance/bundler/commit/426c36551823594f381b1fd136e9836a12f82d8e#diff-75219bd6bc0defc0e7f4ba322d92f950R157.

* The `install_path` variable isn't necessary, but I left it in for the sake of clarity
* The `if` branches at L155-167 can be shortened since three return values are the same, but I left it lengthened for clarity via the comments

Thoughts?

### DIFF
--- a/lib/bundler/settings.rb
+++ b/lib/bundler/settings.rb
@@ -153,7 +153,7 @@ module Bundler
       end
 
       if set_path == File.join(Bundler.settings.root, Bundler.ruby_scope)
-        # ?
+        # root bundler (gems.rb) path
         install_path = set_path
       elsif set_path == Bundler.rubygems.gem_dir
         # system gems path


indirect:
It looks to me like all of the conditionals could be replaced by the single `Pathname.new(set_path).absolute?` Is that true?

Also, (and I missed this before), prefer `Bundler.root` over `Bundler::root`. The `::` for methods is considered obsolete and is in danger of being deprecated.

segiddins:
@indirect none of that code was added in this PR -- it just added the comment

indirect:
Yeah, I know. But those were my thoughts. :stuck_out_tongue:

smlance:
@indirect I think you're right; I can simplify the conditionals, move the comments above the conditionals, and switch to `Bundler.root`. Can I do this in a new PR?

indirect:
@smlance yeah, go for it.

smlance:
Done: https://github.com/bundler/bundler/pull/3943

PR #3898, opened 2015-07-30
Private access credentials for private git repositories

frsyuki:
Pull-request #2825 added support for authorization credentials provided by bundle-config for rubygems source. This change adds the same support for HTTP(S) URL of git source.

Users can provide username and password as following:

    bundle config https://github.com/bundler/bundler.git username:password

or

    export BUNDLE_GITHUB__COM=username:password

Especially for private repositories on Github, users can use personal OAuth tokens:

    export BUNDLE_GITHUB__COM=abcd0123generatedtoken:x-oauth-basic

This enables us to easily deploy ruby code hosted on github private repositories to production servers where storing shared password in plain text at `~/.git-credentials` is security risk.

PR #3942, opened 2015-08-15
Rebase 2-0-dev onto master

smlance:
I'm unsure how to fix the spec added in https://github.com/bundler/bundler/pull/3898 that now fails after the rebase. Thoughts?


indirect:
Without looking into the code yet, maybe the changes to make the test pass were dimly lost n the rebase?

smlance:
@indirect Yes, it's possible. I didn't consistently stick with master or 2-0-dev when resolving merge conflicts; is there a good strategy for resolving them so they don't cause failures like these? https://gist.github.com/smlance/398e5995507714b2c9fe

PR #3943, opened 2015-08-16
Refactor and add comments describing Bundler.settings.path

smlance:
See https://github.com/bundler/bundler/pull/3941.


smlance:
The specs pass, but there are RuboCop violations (which will be taken care of by https://github.com/bundler/bundler/pull/3942).

PR #3939, opened 2015-08-14
Remove `--with` and `--without` flags

smlance:

PR #3955, opened 2015-08-21
Don't autoremember flags

smlance:
Based on https://github.com/bundler/bundler/pull/3744.

### DIFF
--- a/lib/bundler/cli/config.rb
+++ b/lib/bundler/cli/config.rb
@@ -155,8 +155,6 @@ module Bundler
     def resolve_group_conflicts(name, new_value, scope = "global")
       groups = new_value.split(":").map(&:to_sym)
 
-      # TODO: Look here.
-
       if (name == "with") && without_conflict?(groups, scope)
         without_scope = groups_conflict?(:without, groups, :local, scope) ? "locally" : "globally"
         conflicts = conflicting_groups(:without, groups, without_scope == "locally" ? :local : :global, scope)
--- a/lib/bundler/cli/config.rb
+++ b/lib/bundler/cli/config.rb
@@ -51,6 +51,18 @@ module Bundler
         new_value.gsub!(/\s+/, ":") if name == "with" || name == "without"
         locations = Bundler.settings.locations(name)
 
+        if name == "path.system" and Bundler.settings[:path] and new_value == "true"
+          Bundler.ui.warn "`path` is already configured, so it will be unset."
+          Bundler.settings.set_local("path", nil)
+          Bundler.settings.set_global("path", nil)
+        elsif name == "path" and Bundler.settings["path.system"]
+          Bundler.ui.warn "`path.system` is already configured, so it will be unset."
+          Bundler.settings.set_local("path.system", nil)
+          Bundler.settings.set_global("path.system", nil)
+        elsif (name == "with" || name == "without")
+          new_value.gsub!(/\s+/, ":")
+        end
+
         if scope == "global"
           if locations[:local]
             Bundler.ui.info "Your application has set #{name} to #{locations[:local].inspect}. " \
--- a/lib/bundler/cli/config.rb
+++ b/lib/bundler/cli/config.rb
@@ -39,6 +39,8 @@ module Bundler
       when "delete"
         delete_config(name)
       when "local", "global"
+        # NOTE: `"current"` is not a valid scope, because current settings are
+        # not remembered between commands.
         if args.empty?
           Bundler.ui.confirm "Settings for `#{name}` in order of priority. The top value will be used"
           thor.with_padding do
@@ -164,7 +166,7 @@ module Bundler
         if difference == []
           delete_config("without", scope)
         else
-          Bundler.settings.without = difference
+          Bundler.settings.without = difference, :local
         end
 
         :conflict
@@ -179,7 +181,7 @@ module Bundler
         if difference == []
           delete_config("with", scope)
         else
-          Bundler.settings.with = difference
+          Bundler.settings.with = difference, :local
         end
 
         :conflict
@@ -191,7 +193,9 @@ module Bundler
     # Deletes (sets to `nil`) the given configuration variable within the
     # given scope. If the scope is `"global"`, only the local value is deleted,
     # and vice versa. If the scope is `nil` or anything else, both the local
-    # and global values are unset.
+    # and global values are unset. Note that we do not delete the current config
+    # since it is not used in #run or any conflict resolution and current
+    # configurations are not persisted between commands.
     #
     # @param  [String] name
     #         the name of the option being deleted.
--- a/lib/bundler/cli/config.rb
+++ b/lib/bundler/cli/config.rb
@@ -51,18 +51,6 @@ module Bundler
         new_value.gsub!(/\s+/, ":") if name == "with" || name == "without"
         locations = Bundler.settings.locations(name)
 
-        if name == "path.system" and Bundler.settings[:path] and new_value == "true"
-          Bundler.ui.warn "`path` is already configured, so it will be unset."
-          Bundler.settings.set_local("path", nil)
-          Bundler.settings.set_global("path", nil)
-        elsif name == "path" and Bundler.settings["path.system"]
-          Bundler.ui.warn "`path.system` is already configured, so it will be unset."
-          Bundler.settings.set_local("path.system", nil)
-          Bundler.settings.set_global("path.system", nil)
-        elsif (name == "with" || name == "without")
-          new_value.gsub!(/\s+/, ":")
-        end
-
         if scope == "global"
           if locations[:local]
             Bundler.ui.info "Your application has set #{name} to #{locations[:local].inspect}. " \
--- a/lib/bundler/cli/config.rb
+++ b/lib/bundler/cli/config.rb
@@ -94,9 +94,7 @@ module Bundler
       end
     end
 
-    # Clears `path` if `system` is being set, and vice versa. Also, checks
-    # whether `path` is already set and `path.system` is being set, and
-    # vice versa.
+    # Clears `path` if `path.system` is being set, and vice versa.
     #
     # @param  [String] name
     #         the name of the option being set by the user.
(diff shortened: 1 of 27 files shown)


smlance:
(Waiting for https://github.com/bundler/bundler/pull/3939 to be merged.)

smlance:
It looks like certain matchers in `matchers.rb`, e.g., `should_be_installed`, rely on settings configured in previous invocations of `bundle`. This makes certain specs that use `bundle` commands with flags fail, because these settings are not persisted to the matcher commands. Thoughts about how to deal with this? @indirect, @segiddins?

segiddins:
@smlance allow passing flags to `should_be_installed`?

indirect:
I think it makes sense to allow `path` as an argument to `should_be_installed`... is there anything else we need to support? As an alternative to an argument, we could also potentially create a more generic block form for config settings, something like `config(:path => "/wherever") { should_be_installed("rack 1.0") }`

lib/bundler/settings.rb
```diff
@@ -176,11 +185,14 @@ def groups_array(group_type, scope)
     # Always returns an absolute path to the bundle directory
     # TODO: Refactor this method
     def path
+      # TODO: Shoudl we consider the @current_config here?
```
segiddins: why should / shouldn't we?

lib/bundler/settings.rb
```diff
@@ -260,8 +274,22 @@ def get_array(key)
...
-      self[key] = (array.empty? ? nil : array.join(":")) if array
+    def set_array(key, array, scope)
+      if scope != :current and scope != :local and scope != :global
+        Bundler.ui.error "Invalid scope #{scope} given. Please use :local or :global."
+        exit 1
+      end
+
+      if array
+        value = (array.empty? ? nil : array.join(":"))
+        if scope == :current
```
segiddins: please use a `case` statement for this, and do the scope checking in there

spec/commands/clean_spec.rb
```diff
@@ -470,12 +472,19 @@ def should_not_have_gems(*gems)
...
     G
     bundle "install --system"
-    bundle "clean --force"
 
-    expect(out).to include("Removing foo (1.0)")
-    sys_exec "gem list"
-    expect(out).not_to include("foo (1.0)")
-    expect(out).to include("rack (1.0.0)")
+    # See cli/install.rb:L76.
+    # FIXME: [user-unfriendly] `clean` only happens if path is set to
```
segiddins: this looks like a bug that needs to get fixed

spec/commands/config_spec.rb
```diff
@@ -14,9 +14,12 @@
       bundle "config path vendor/bundle"
       bundle "install"
 
-      expect(bundled_app(".bundle")).not_to exist
-      expect(tmp("foo/bar/config")).to exist
-      should_be_installed "rack 1.0.0"
+      # See cli/install.rb:L86
```
segiddins: please don't lint to lines -- use methods, which are less likely to go out of date

lib/bundler/settings.rb
smlance: Never mind -- I don't think we need to. The comment just below it explains why: `--path` is no longer used, so `bundle install --path foo` or any command that previously accepted `--path` would now be set with `bundle config`, which only set options in the `local` and `global` scopes.

lib/bundler/settings.rb
```diff
@@ -9,15 +9,15 @@ class Settings
     attr_reader :root
 
     def initialize(root = nil)
-      @root          = root
-      @local_config  = load_config(local_config_file)
-      @global_config = load_config(global_config_file)
+      @root           = root
+      @current_config ||= {}
```
segiddins: no need to `||=` in `initialize`

lib/bundler/settings.rb
```diff
@@ -260,8 +289,20 @@ def get_array(key)
       self[key] ? self[key].split(":").map(&:to_sym) : []
     end
 
-    def set_array(key, array)
-      self[key] = (array.empty? ? nil : array.join(":")) if array
+    def set_array(key, array, scope)
+      unless [:current, :local, :global].include? scope
+        Bundler.ui.error "Invalid scope #{scope} given. Please use :local or :global."
```
segiddins: raise an exception?

lib/bundler/settings.rb
```diff
@@ -160,11 +185,13 @@ def with(scope = nil)
...
+      when nil
         get_array(group_type)
-      elsif scope == :global
+      when :global
         @global_config[key] ? @global_config[key].split(" ").map(&:to_sym) : []
-      elsif scope == :local
+      when :local
         @local_config[key] ? @local_config[key].split(" ").map(&:to_sym) : []
       else
         Bundler.ui.error "Invalid scope #{scope} given. Please use :local or :global."
```
segiddins: raise an exception here?

spec/support/builders.rb
```diff
@@ -365,7 +365,6 @@ def build_with(builder, name, args, &blk)
       versions = args.last || "1.0"
       spec     = nil
 
-      # TODO: [2.0] Change this, since we won't use options[:path]
       options[:path] ||= @_build_path
```
segiddins: shouldn't we change this?

spec/support/helpers.rb
```diff
@@ -1,5 +1,13 @@
 module Spec
   module Helpers
+    def set_temp_config(config)
+      if block_given?
+        config.each {|k, v| Bundler.settings.set_local(k, v) }
+        yield
+        config.each {|k, _| Bundler.settings.set_local(k, nil) }
```
segiddins: shouldn't this restore the settings?

smlance: Oh -- good point.

spec/support/builders.rb
smlance: Hm. I'll look into this later.

smlance: No. It looks like this options hash is specific to this file and has nothing to do with the options hash passed to `bundle` commands.

segiddins: ok, so the `TODO` was incorrect?

smlance: Yes. I added it incorrectly.

lib/bundler/cli/config.rb
```diff
@@ -109,7 +112,17 @@ def run
...
+        Bundler.ui.warn "`path` is already configured, so it will be unset."
+        delete_config("path")
+        # We still want `system` to be set.
+        :no_conflict
+      elsif name == "path" and Bundler.settings[:system]
+        Bundler.ui.warn "`system` is already configured, so it will be unset."
+        delete_config("system")
+        # We still want `path` to be set.
+        :no_conflict
+      elsif name == "path.system" and Bundler.settings[:path] and new_value == "true"
```
indirect: Oops. We wound up with `path`, `system`, and `path.system`. I intended us to only have two. :) I think `path` and `path.system` is the clearest. Let's get rid of just plain `system`.

spec/support/helpers.rb
```diff
@@ -1,5 +1,18 @@
 module Spec
   module Helpers
+    def set_temp_config(config)
```
indirect: I think renaming this to `with_config` would match with Ruby conventions for methods that take a block and then unset after.

spec/realworld/parallel_spec.rb
```diff
@@ -9,7 +9,9 @@
       gem 'i18n', '~> 0.6.0' # Because 0.7+ requires Ruby 1.9.3+
     G
 
-    bundle :install, :jobs => 4, :env => { "DEBUG" => "1" }
+    bundle "config jobs 4"
+    # NOTE: The env option below cannot be configured with `bundle "config env debug"`
+    bundle :install, :env => { "DEBUG" => "1" }
```
indirect: it's actually possible to just use `bundle install --verbose` here (verbose output contains the thread numbers, which is why we use verbose output in this test)

lib/bundler/settings.rb
```diff
@@ -30,13 +31,17 @@ def [](name)
...
 
-    def []=(key, value)
+    def set_current(key, value)
+      @current_config[key_for(key)] = value
+    end
+
+    alias_method :[]=, :set_current
+
+    def set_local(key, value)
       local_config_file or raise GemfileNotFound, "Could not locate #{SharedHelpers.gemfile_name}"
```
segiddins: please don't use `or`

spec/support/helpers.rb
```diff
@@ -1,5 +1,18 @@
...
+    def with_config(config)
+      if block_given?
+        old_config = {}
+
+        config.each do |k, v|
+          old_config[k] = Bundler.settings[k]
+          Bundler.settings.set_local(k, v)
+        end
+        yield
+        config.each {|k, _| Bundler.settings.set_local(k, old_config[k]) }
```
segiddins: this needs to be in an `ensure`

segiddins:
This seems good?

spec/support/helpers.rb
smlance: Which part?

lib/bundler/settings.rb
smlance: Would `unless` be preferable?

segiddins:
@smlance you'll need to squash the `wip` commits out, they don't provide any information

smlance:
Now that I removed the flags from the list of `method_option`s, `thor` detects that the flags are invalid before `bundler` does, so the error messages I added won't ever be shown.

Should each command have its own `path` setting? They all shared the same `path` configuration before, and this seemed to only be a problem for `binstubs` (see the fixed commits in https://github.com/smlance/bundler/commit/fe9e4246804ecab4b6a5b98f563371b8358625c8). Based on the specs, this isn't a problem for the other commands, so I'm okay with allowing them to share it.

lib/bundler/settings.rb
segiddins: Yes

spec/support/helpers.rb
segiddins: Restoring the old value

indirect:
This looks good!

indirect:
@homu r+

PR #3956, opened 2015-08-22
Disable Lint/UnusedMethodArgument

Koronen:
We don't agree with this rule. See discussion in <https://github.com/bundler/bundler/pull/3922>.


segiddins:
@homu r+

PR #3958, opened 2015-08-22
fast static require

grosser:
@indirect got inspired by @tenderlove's red-dot talk to make require O(1) ... took our app boot time from 18s->12s (worked for ~3600 of ~4000 requires = 90% cache hits with 14k files in the cache total) ... are you interested in adding something like this ... obviously needs more work / tests, but wanted to get your feedback

### DIFF
--- a/lib/bundler/static_setup.rb
+++ b/lib/bundler/static_setup.rb
@@ -1,0 +1,43 @@
+require 'bundler/setup'
+
+# cache rack -> /foo/bar/vendor/bundle/rack/lib/rack.rb
+module Bundler
+  PRELOADED_REQUIRE = {}
+  $LOAD_PATH.reverse_each do |path|
+    Dir["#{path}/**/*.rb"].each do |f|
+      start = path.size + 1
+      count = f.size - start - 3 # without '.rb' = last 3
+      PRELOADED_REQUIRE[f.slice(start, count)] = f
+    end
+  end
+  PRELOADED_REQUIRE[:state] = $LOAD_PATH.join('')
+
+  # Bundler::StaticSetup.verification = :warn
+  module StaticSetup
+    class << self
+      attr_accessor :verification
+    end
+  end
+end
+
+module Kernel
+  alias require_without_bundler_preload require
+  def require(path)
+    verification = ::Bundler::StaticSetup.verification
+    if !verification.nil? && ::Bundler::PRELOADED_REQUIRE.fetch(:state) != $LOAD_PATH.join('')
+      case verification
+      when :warn then warn "$LOAD_PATH was changed"
+      when :raise then raise "$LOAD_PATH was changed"
+      when :skip then require_without_bundler_preload(path)
+      else
+        raise ::ArgumentError, "Unknown verification option #{verification}"
+      end
+    end
+
+    if f = ::Bundler::PRELOADED_REQUIRE[path]
+      require_without_bundler_preload(f)
+    else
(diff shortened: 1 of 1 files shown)


indirect:
Looks like #3882 is relevant to your interests. I'm not yet sure about the tradeoffs between performance and correctness in these options; if we can make require more performant without any loss of correctness, that would be pretty fantastic.

byroot:
Well in #3882 I took it a bit farther by wanting not to populate the $LOAD_PATH. But a tamer version that would let the LOAD_PATH would still provide a big speedup while being 100% correct.

The downside is that the correctness is tied on reimplementing the require precedence logic in ruby. I mean this: https://github.com/byroot/bundler/blob/3fc5230b7956a148558b545d7c12c50040302683/lib/bundler/require_cache.rb#L42-L61

segiddins:
I think this approach is cool, but seeing as it has absolutely no bundler-specific code... I'm just unsure it belongs in Bundler.

grosser:
In a non-bundler app this simple approach makes less sense since the load-path is added to all the time, also it has to be done after bundler/setup ... so I figured here is a good place ... maybe only to kick off a discussion / being told 'this already exists' :)

@indirect this loses correctness once something is added/removed to the LOAD_PATH ... but we could have a 'strict' mode where it blows up when the load-path got changed or warn or 'stop being fast' to be correct.

I'm not worried about native gems etc foobar since they only make up a small percentage of requires -> lets get that easy 90% :)

grosser:
added verification, so you can optionally make sure it is correct

grosser:
so if someone is worried about correctness they would set it to raise, boot their app and fix all the extra load-paths that get added in random places

```
require bundler/setup'
$LOAD_PATH << 'custom_foo'
require bundler/static_setup'
```

lib/bundler/static_setup.rb
```diff
@@ -0,0 +1,43 @@
+require 'bundler/setup'
+
+# cache rack -> /foo/bar/vendor/bundle/rack/lib/rack.rb
+module Bundler
+  PRELOADED_REQUIRE = {}
+  $LOAD_PATH.reverse_each do |path|
+    Dir["#{path}/**/*.rb"].each do |f|
+      start = path.size + 1
+      count = f.size - start - 3 # without '.rb' = last 3
```
simi: Is this (`.rb`) really universal?

grosser: with just .rb hits 4698 - misses 405
with .so,.o added it's the same ... so seems to not make any difference ...
and even if it does probably not worth the extra complexity/time (search
with pattern + extracting extname) ... on miss it falls back to default
ruby anyway

byroot:
Since it seems complicated to have a patch like this in bundler or rubygems, I published my version of this trick as a gem: https://github.com/byroot/bootscale. I think it's simple enough to integrate for people who want to do so.

@grosser if you'd like to try to integrate that gem in Zendesk, I'd love to hear your feedback. I tested it on Shopify in replacement of our old patch and it works fine.

grosser:
The static Kernel patch did not help me, but the autoload patch fixed
another 300 files :)

Seems to work nicely! Thx for gemifying :)

byroot:
> The static Kernel patch did not help me

It does pick up a relatively low number of files for me. I'll try to figure out what the difference is.

> Seems to work nicely! Thx for gemifying :)

My pleasure!

PR #3959, opened 2015-08-22
Wrap sudo access prompts in a mutex

### LINKED ISSUE #3910: Parallel jobs break system gems installation
When bundler is configured for using parallel jobs, the installation of system gems (which request sudo permissions) fails.

Specifically, on `bundle install` execution, several password prompts show up at the same time (as many as the parallel jobs configured), and even if the correct password is inserted, it will be rejected as wrong.

The workaround is to either override the parallel jobs parameter to 1, or to disable this feature from the configuration at all.

The bundler version is 1.10.6.

agis-:
Fixes #3910.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -1,6 +1,7 @@
 require "fileutils"
 require "pathname"
 require "rbconfig"
+require "thread"
 require "bundler/gem_path_manipulation"
 require "bundler/rubygems_ext"
 require "bundler/rubygems_integration"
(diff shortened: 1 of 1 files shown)


indirect:
Thanks!

segiddins:
@homu r+

agis-:
@segiddins I've updated the patch to work on Ruby 1.8 also, so I guess you could cancel https://travis-ci.org/bundler/bundler/builds/76799207.

Sorry for the inconvenience! :)

segiddins:
@homu r+

PR #3961, opened 2015-08-23
maybe we can install missing system gems?

indirect:
(this seems to be the current leading reason for CI job failures)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -73,26 +73,21 @@ Style/TrailingComma:
 
 # Metrics
 
-# Arbitrary max lengths for classes simply do not work and enabling this will
-# lead to a never ending stream of annoyance and changes.
+# We've chosen to use Rubocop only for style, and not for complexity or quality checks.
 Metrics/ClassLength:
   Enabled: false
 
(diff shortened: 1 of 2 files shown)


segiddins:
I don't see how this will help things? Also, it failed on CI.

indirect:
It passed except for Rubocop, which failed it on `Metrics/ModuleLength: Module has too many lines. [368/363]`. I thought we disabled that?

indirect:
To elaborate: the current most common reason for CI jobs to fail is the error message `you need to rm -rf tmp`. That error message is only printed when Bundler is unable to find rake inside the test system gems. I'm adding logic to delete the test system gems and try to install them again, one time, and then fail if that doesn't work. Since the code in question only needs to successfully run once at the beginning of the test suite, it seems like this will reduce or eliminate our current occasional CI failures.

segiddins:
Please do disable that rule -- then feel free to 👍🏻

indirect:
@homu r=segiddins

indirect:
(this is why r= is useful!)

PR #3965, opened 2015-08-26
Show friendly errors when ref is not found

agis-:
Before:

    $ bundle install
    Fetching git://github.com/agis-/some_gem.git
    fatal: ambiguous argument 'FOO': unknown revision or path not in the working tree.
    Use '--' to separate paths from revisions, like this:
    'git  [...] -- [...]'
    Git error: command `git rev-parse WRONG` in directory /cache/path
    has failed.
    Ref 'foo' was not found. Perhaps you mispelled it?
    # ...

After:

    $ bundle install
    Fetching git://github.com/agis-/some_gem.git
    fatal: Needed a single revision
    Git error: command `git rev-parse --verify WRONG` in directory /cache/path
    has failed.
    Ref 'foo' was not found. Perhaps you mispelled it?
    # ...

Fixes #3879.

### DIFF
--- a/lib/bundler/source/git/git_proxy.rb
+++ b/lib/bundler/source/git/git_proxy.rb
@@ -19,8 +19,9 @@ module Bundler
       end
 
       class GitCommandError < GitError
-        def initialize(command, path = nil)
+        def initialize(command, path = nil, extra_info = nil)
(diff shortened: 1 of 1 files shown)


indirect:
Looks good to me. @homu r+

PR #3966, opened 2015-08-27
Failing spec for BUNDLE_NO_INSTALL breaking `bundle install`

chulkilee:
Problem:

1. Run `bundle package --all --no-installl` on build machine during build
2. Run `bundle install --deployment` on target machine
3. It does not install any gems because `BUNDLE_NO_INSTALL` is set at `.bundle/config`.

A possible workaround is to remove the line from the config, but I believe `bundle install` should ignore `BUNDLE_NO_INSTALL`.

Another way to solve it is adding a new option to `bundle install` to force install `--force-install`. If we really need to keep the current behavior, we should do that. Otherwise, this PR would work.

### DIFF
--- a/spec/commands/package_spec.rb
+++ b/spec/commands/package_spec.rb
@@ -41,6 +41,18 @@ describe "bundle package" do
       should_not_be_installed "rack 1.0.0"
       expect(bundled_app("vendor/cache/rack-1.0.0.gem")).to exist
     end
+
+    it "does not prevent installing gems with bundle install" do
+      gemfile <<-D
+        source "file://#{gem_repo1}"
+        gem 'rack'
+      D
+
+      bundle "package --no-install"
+      bundle "install"
+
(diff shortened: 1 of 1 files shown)


segiddins:
This breaks the tests. https://travis-ci.org/bundler/bundler/jobs/77554699#L2217

chulkilee:
Oh, it's more complicated.. It looks like

- CLI::Package uses CLI::Install to fetch gems before package
- CLI::Package uses Bundle.settings to pass `no-install` option down to `Bundler::Source::Rubygems`, which is pretty bad like a global variable :(

For now, I just added a failing test to show` bundle install` wouldn't install after `bundle package --no-install`.

segiddins:
I think I have a fix for this, just let me finish running the specs and I'll make a PR. Thanks so much for the failing test case, @chulkilee !

PR #3968, opened 2015-08-28
make it able to specify gem host with GEM_HOST env

pmenglund:
when you in your `.gemspec` specify

```
  if spec.respond_to?(:metadata)
    spec.metadata['allowed_push_host'] = 'https://gems.example.com'
  end
```

you can't use the `bundler` rake tasks, as it can't specify the alternate host. This change picks up the `allowed_push_host` and uses it when pushing.

### DIFF
--- a/lib/bundler/gem_helper.rb
+++ b/lib/bundler/gem_helper.rb
@@ -93,9 +93,10 @@ module Bundler
     def rubygem_push(path)
       if Pathname.new("~/.gem/credentials").expand_path.exist?
         gem_command = "gem push '#{path}'"
-        gem_command << " --host #{ENV['GEM_HOST']}" if ENV['GEM_HOST']
+        allowed_push_host = @gemspec.metadata["allowed_push_host"]
+        gem_command << " --host #{allowed_push_host}" if allowed_push_host
         sh(gem_command)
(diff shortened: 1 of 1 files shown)


indirect:
Why not just read the gem host out of the gemspec's `allowed_push_host` if it's present? Then you don't need to do anything with env vars at all...

pmenglund:
good point! I'll modify it to do that instead...

lib/bundler/gem_helper.rb
```diff
@@ -92,8 +92,11 @@ def install_gem(built_gem_path = nil, local = false)
 
     def rubygem_push(path)
       if Pathname.new("~/.gem/credentials").expand_path.exist?
-        sh("gem push '#{path}'")
-        Bundler.ui.confirm "Pushed #{name} #{version} to rubygems.org."
+        gem_command = "gem push '#{path}'"
+        allowed_push_host = @gemspec.metadata["allowed_push_host"]
```
segiddins: isn't it possible that `metadata` isn't implemented, on a sufficiently old RubyGems?

segiddins:
:+1:

indirect:
@homu r+

PR #3969, opened 2015-08-30
Output gemspec validation errors on exec commands

### LINKED ISSUE #3895: Bundler 1.10.6 can't find gems with invalid gemspec files
I have a custom gem that is installed from a git source in my Gemfile. After upgrading from Bundler 1.9 to Bundler 1.10.6, I started receiving errors that the gem could not be found when trying to load anything that interacted with the gem (in this example, it's spinning up a rails server):

```
$ rails s
Could not find mygem-0.1.0 in any of the sources
```

After downgrading to 1.10.5, things would still fail to start, but I at least got a more helpful error that led me to the actual problem:

```
$ rails s
The gemspec at /Users/nmuerdte/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/bundler/gems/mygem-51004a1371dd/mygem.gemspec is not valid. The validation error was '"FIXME" or "TODO" is not a description'
```

So while this is certainly a solvable problem with my gem (since the gemspec is indeed invalid), the current behavior in 1.10.6 is somewhat confusing, since `bundle install` reports installing the gem without any warnings, but then when trying to load the gem it simply reports it as missing.

I think the change in behavior is related to this change: https://github.com/bundler/bundler/pull/3856 I think the attempt with that pull request was to actually allow these invalid gemspecs to continue working on 1.10, but it seems like it's perhaps just deferred the error to later when the loading fails (but in a less obvious way). Did something else maybe change earlier on the 1.10 branch that's completely excluding these invalid gemspecs from the gems Bundler will load?

I'm seeing this under Ruby 2.2.1 and RubyGems 2.4.5.

Thanks!

agis-:
Fixes #3895.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -398,7 +398,7 @@ module Bundler
         spec
       end
     rescue Gem::InvalidSpecificationException => e
-      Bundler.ui.warn "The gemspec at #{file} is not valid. " \
+      UI::Shell.new.warn "The gemspec at #{file} is not valid. " \
         "The validation error was '#{e.message}'"
       nil
     end
--- a/spec/commands/exec_spec.rb
+++ b/spec/commands/exec_spec.rb
@@ -318,4 +318,31 @@ describe "bundle exec" do
     bundle "exec rackup"
     expect(out).to include("Installing foo 1.0")
   end
+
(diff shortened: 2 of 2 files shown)


segiddins:
@homu r+

segiddins:
@homu retry

PR #3971, opened 2015-08-30
Distinguish Gemfile syntax and evaluation errors

### LINKED ISSUE #3783: Gemfile friendly error should allow ruby requirement exceptions
The new friendly Gemfile error handler mistakenly treats exceptions raised from the Ruby version checker as if they are syntax errors or something similar:

```
[!] There was an error parsing `Gemfile`: ruby_version must match the :engine_version for MRI. Bundler cannot continue.
 #  from /home/travis/build/bundler/bundler/tmp/bundled_app/Gemfile:2
 #  source "file:///home/travis/build/bundler/bundler/tmp/gems/remote1"
 >  ruby "1.8.7", :engine => 'ruby', :engine_version => '1.2.4'
 #  
```

This happens during the spec named "raises an error if engine version doesn't match ruby version for MRI", among others.

Instead, we should simply treat the exception as a normal friendly error, and print the message out in red and exit non-zero.

agis-:
Fixes #3783.

### DIFF
--- a/exe/bundle_ruby
+++ b/exe/bundle_ruby
@@ -8,6 +8,8 @@ require "bundler/shared_helpers"
 
 module Bundler
   class GemfileError < RuntimeError; end
+  class GemfileEvalError < GemfileError; end
+
   class Dsl
     include RubyDsl
 
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -96,6 +96,7 @@ module Bundler
     status_code(24)
   end
 
+  class GemfileEvalError < GemfileError; end
   class MarshalError < StandardError; end
 
   class PermissionError < BundlerError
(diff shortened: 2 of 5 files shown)


indirect:
This would probably be easier as `GemfileParseError` and `GemfileEvalError`, that way you wouldn't have to check on strings all over the place. :)

agis-:
Sounds better indeed (I just wasn't sure we want to introduce more
exception classes). I will update this tomorrow :)

agis-:
@indirect You mean something like:

```ruby
class GemfileEvalError < GemfileError; end
class GemfileParseError < GemfileError; end
```

and then in `RubyDsl#ruby`:

```ruby
raise GemfileEvalError, "There was an error evaluating etc. etc."
```

And then update the other places in `RubyDsl#ruby` accordingly?

indirect:
You could probably even only add GemfileEvalError < GemfileError, and then use it when needed.

agis-:
@indirect updated :)

segiddins:
:+1:

indirect:
thanks! @homu r+

PR #3975, opened 2015-09-02
Cache gems into ~/.bundle/cache [WIP]

smlance:
We probably also want to test that this works when separate local bundles are installed on the same machine, i.e., that they don't overwrite the .gems in the cache, although I'm unsure how to do this just yet.

### DIFF
--- a/lib/bundler/runtime.rb
+++ b/lib/bundler/runtime.rb
@@ -125,12 +125,10 @@ module Bundler
       cache_path = Bundler.global_cache
       FileUtils.mkdir_p(cache_path) unless File.exist?(cache_path)
 
-      Bundler.ui.info "Updating files in #{Bundler.settings.global_cache_path}"
+      Bundler.ui.info "Caching gems in #{Bundler.settings.global_cache_path}"
 
       specs.each do |spec|
         next if spec.name == "bundler"
-        spec.source.send(:fetch_gem, spec) if Bundler.settings[:cache_all_platforms] && spec.source.respond_to?(:fetch_gem, true)
-
         spec.source.global_cache(spec, custom_path) if spec.source.respond_to?(:global_cache)
       end
       Dir[cache_path.join("*/.git")].each do |git_dir|
--- a/lib/bundler/runtime.rb
+++ b/lib/bundler/runtime.rb
@@ -121,7 +121,7 @@ module Bundler
       prune_cache(cache_path) unless Bundler.settings[:no_prune]
     end
 
-    def global_cache(custom_path = nil)
+    def global_cache
       cache_path = Bundler.global_cache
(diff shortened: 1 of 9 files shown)


lib/bundler/runtime.rb
```diff
@@ -121,6 +121,22 @@ def cache(custom_path = nil)
       prune_cache(cache_path) unless Bundler.settings[:no_prune]
     end
 
+    def global_cache
```
segiddins: `cache_globally`?

lib/bundler/runtime.rb
```diff
@@ -121,6 +121,22 @@ def cache(custom_path = nil)
...
+      cache_path = Bundler.global_cache
+      FileUtils.mkdir_p(cache_path) unless File.exist?(cache_path)
```
segiddins: Leave off the `unless`

segiddins:
Only gems with no extensions can be safely cached globally, I think.

indirect:
@segiddins this cache is for the .gem files, not for the installed gems. so all gems can be cached here

lib/bundler/settings.rb
```diff
@@ -222,6 +222,10 @@ def app_cache_path
       end
     end
 
+    def global_cache_path
+      "~/.bundle/cache"
```
indirect: `~` doesn't work on Windows, so well need to use whatever Bundler currently uses to find the global settings file.

indirect:
This implementation looks good to me, but as I read it I realized that I think we can't do global caching this way. Because `load_global_cache` reads every cached `.gem` file, ungzipping and untarring it to extract the .gemspec file and then evaluating the gemspec, it's going to get slower and slower as the global cache grows.

Instead of loading gemspecs from the global cache, I think we will have to use it as a pure download cache instead—when the RubyGems source is asked to download a .gem file, it checks the global cache and provides the cached .gem if present, and downloads the gem into the global cache if it is not. Make sense?

lib/bundler/source/rubygems.rb
```diff
@@ -20,7 +20,7 @@ def initialize(options = {})
         @dependency_names = []
         @allow_remote = false
         @allow_cached = false
-        @caches = [Bundler.app_cache, *Bundler.rubygems.gem_cache]
+        @caches = [Bundler.app_cache, *Bundler.rubygems.gem_cache, Bundler.global_cache]
```
indirect: we need to remove this change--the global cache should be for downloads only. It will be too slow to use the global cache for gem metadata, because getting the metadata out of .gem files means decompressing them and evaluating lots of ruby code.

smlance:
See https://github.com/bundler/bundler/pull/3983/

PR #3977, opened 2015-09-05
Improve hint on conflict error message

seanlinsley:
Given this previous configuration:
```ruby
gem 'rails',      '~> 4.1.11'
gem 'sass-rails', '~> 4.0.3'
```

When changing both gem versions:
```ruby
gem 'rails',      '~> 4.2.4'
gem 'sass-rails', '~> 5.0.0'
```

And running `bundle update rails`, you see this error message:
```
You have requested:
  sass-rails ~> 5.0.0

The bundle currently has sass-rails locked at 4.0.3.
Try running `bundle update sass-rails`
```

This commit adds a hint to run `bundle update rails sass-rails`:
```
You have requested:
  sass-rails ~> 5.0.0

The bundle currently has sass-rails locked at 4.0.3.
Try running `bundle update sass-rails`

If you are updating multiple gems in your Gemfile at once,
try passing them all to `bundle update`
```

### DIFF
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -326,7 +326,9 @@ module Bundler
             message = "You have requested:\n" \
               "  #{requirement.name} #{requirement.requirement}\n\n" \
               "The bundle currently has #{requirement.name} locked at #{version}.\n" \
-              "Try running `bundle update #{requirement.name}`"
+              "Try running `bundle update #{requirement.name}`\n\n" \
+              "If you are updating multiple gems in your Gemfile at once,\n" \
+              "try passing them all to `bundle update`"
           elsif requirement.source
             name = requirement.name
             versions = @source_requirements[name][name].map(&:version)


seanlinsley:
https://twitter.com/seanlinsley/status/636303643151470592
https://rubytogether-members.slack.com/archives/computering/p1440628494000034

seanlinsley:
I would have provided the direct example to run the command with any/all gems currently being updated, but I wasn't sure where I could find that information

indirect:
Uh... huh. So this is, I guess, very counter-intuitive, but what you actually wanted from the very beginning was to just run `bundle install`. That will install the versions you ask for in your Gemfile, period.

The `bundle update` command is for when you _don't_ change the Gemfile. You might have put e.g. `gem "rails", "~> 4.2.0"`. When Rails 4.2.1 comes out, you can run `bundle update rails` to update your lock file from 4.2.0 to 4.2.1.

Does that make sense?

seanlinsley:
`bundle update` is needed, because it involves changes to the lockfile. Simply changing the Gemfile as shown above and running `bundle install` results in this error:
```
You have requested:
  rails ~> 4.2.4

The bundle currently has rails locked at 4.1.11.
Try running `bundle update rails`
```

Hasn't that always been how Bundler works? I've always thought that `bundle update` was required if the lockfile needs to be changed.

I didn't specify before, but I'm using Bundler 1.10.6

indirect:
😳 You're completely right. Once a gem is locked, you do need to use `update` to unlock it and install a different version. Sorry about that, what I said was completely wrong.

Now I'm really curious... knowing that gems will stay locked unless you pass their name to `bundle update` (or run it without arguments), what were you expecting to happen when you changed two gems in the Gemfile, even though they were both already locked in the lock file, and then only unlocked one at a time using `bundle update`? Were you expecting them to both unlock simultaneously because you changed both of their version numbers?

segiddins:
👍🏻

seanlinsley:
> Were you expecting them to both unlock simultaneously because you changed both of their version numbers?

I don't know what I was expecting, but I didn't yet know that I could pass multiple gems to `bundle update`, so I didn't know how to get around the error while updating only those gems.

Thanks for merging this :smile_cat: my first bundler PR!

PR #3978, opened 2015-09-06
Dsl#with_source should restore the previous source

### LINKED ISSUE #3974: Git reference inside a source block changes .lock file unexpectedly
If you reference a git repo inside of a source block, any gems after said git reference will be changed in `Gemfile.lock`.  

##### Starting project, with the `Gemfile` and `Gemfile.lock` as expected:
https://github.com/craig-day/bearded-nemesis

I have the base gems for a simple rails app in my `Gemfile`, then I also want to use https://rails-assets.org for my asset management, so I added a source block at the bottom to do so.  All is fine and behaving expectedly. 

##### Example problem branch: 
https://github.com/craig-day/bearded-nemesis/tree/broken_gemfile

This example seems like a silly thing to do, but it illustrates the point well. I added a gem with a git reference to the source block.  If you look at the diff for `Gemfile.lock` ([here](https://github.com/craig-day/bearded-nemesis/compare/broken_gemfile)) you will notice that it removed the `!` from all of the gems referenced after the git reference. 

To put this into context, I can explain a real situation.  At work we use a private gem server to host published gems, and Github to manage our git repos. In the normal situation, gems will get published to the gem server and all is well. If I want to test a gem feature branch for example, the simplest thing to do would be to tell that gem to reference the branch on Github.  This works fine, but it makes unnecessary changes to the lock file. I would expect referencing git to only change the single gem I changed.

agis-:
Fixes #3974.

### DIFF
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -229,13 +229,14 @@ module Bundler
     end
 
     def with_source(source)
+      old_source = @source
       if block_given?
         @source = source
         yield
       end
       source
     ensure
-      @source = nil
+      @source = old_source
     end
 
     def normalize_hash(opts)
--- a/spec/bundler/dsl_spec.rb
+++ b/spec/bundler/dsl_spec.rb
@@ -195,4 +195,20 @@ describe Bundler::Dsl do
         to raise_error(Bundler::GemfileError, /There was an error parsing `Gemfile`: can't modify frozen String. Bundler cannot continue./i)
     end
   end
+
+  describe "#with_source" do
+    context "if there was a rubygem source already defined" do
+      it "restores it after it's done" do
+        other_source = double("other-source")
+        allow(Bundler::Source::Rubygems).to receive(:new).and_return(other_source)
+
+        subject.source("https://other-source.org") do
+          subject.gem("dobry-pies", :path => "foo")
+          subject.gem("foo")
+        end
+
+        expect(subject.dependencies.last.source).to eq(other_source)
+      end
+    end
+  end
 end


spec/realworld/edgecases_spec.rb
```diff
@@ -210,4 +210,21 @@
...
+        source 'https://rails-assets.org' do
+          gem 'rails-assets-angular', git: "git@github.com:sharoo/rails-assets-angular.git"
+          gem "i18n"
+        end
+      G
+
+      expect(out).to include("Could not find gem 'i18n' in rubygems repository https://rails-assets.org")
+      expect(exitstatus).to eq(7) if exitstatus
+    end
+  end
```
agis-: This could also be tested inside `Dsl` but a realworld test is more suitable for such a bug imho. That said, I'm not happy with the fact that we depend on `rails-assets.org` being up *and* not containing `i18n`. Ideas?

cc @indirect @segiddins

lib/bundler/dsl.rb
```diff
@@ -230,12 +230,13 @@ def add_git_sources
 
     def with_source(source)
       if block_given?
+        old_source = @source
```
segiddins: Needs to go outside the conditional

lib/bundler/dsl.rb
```diff
@@ -230,12 +230,13 @@ def add_git_sources
...
         @source = source
         yield
       end
       source
     ensure
-      @source = nil
+      @source = old_source || nil
```
segiddins: The `||` is unnecessary

segiddins:
Awesome catch! Id actually prefer to see this covered by just a normal test as well, there are some helpers for setting up sources.

agis-:
@segiddins Yep that was my thought in https://github.com/bundler/bundler/pull/3978#discussion_r38821336. Cool, I'll do it.

agis-:
@segiddins Updated!

lib/bundler/dsl.rb
```diff
@@ -229,13 +229,14 @@ def add_git_sources
...
     def with_source(source)
+      old_source = @source
       if block_given?
         @source = source
         yield
       end
       source
     ensure
-      @source = nil
+      @source = old_source
```
grosser: man what an obvious bug :(

grosser:
:+1:

lib/bundler/dsl.rb
agis-: yep :)

indirect:
awesome, thanks!

@homu r+

TimMoore:
Good catch!

PR #3979, opened 2015-09-07
Dsl#with_source should restore the previous source

### LINKED ISSUE #3974: Git reference inside a source block changes .lock file unexpectedly

indirect:
Before this change, a gem inside a rubygems source block that also
defined a custom source, would cause subsequent gems inside that same
block to be fetched from rubygems.org.

For example:

    source "https://rubygems.org"

    source "https://foo.bar" do
      gem "rails", git: "git@github.com/rails/rails.git"
      gem "i18n"
    end

In that case, i18n would be fetched from rubygems.org instead of
foo.bar.

Fixes #3974.

Backport of #3978

### DIFF
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -226,13 +226,14 @@ module Bundler
     end
 
     def with_source(source)
+      old_source = @source
       if block_given?
         @source = source
         yield
       end
       source
     ensure
-      @source = nil
+      @source = old_source
     end
 
     def normalize_hash(opts)
--- a/spec/bundler/dsl_spec.rb
+++ b/spec/bundler/dsl_spec.rb
@@ -188,4 +188,20 @@ describe Bundler::Dsl do
     end
   end
 
+  describe "#with_source" do
+    context "if there was a rubygem source already defined" do
+      it "restores it after it's done" do
+        other_source = double("other-source")
+        allow(Bundler::Source::Rubygems).to receive(:new).and_return(other_source)
+
+        subject.source("https://other-source.org") do
+          subject.gem("dobry-pies", :path => "foo")
+          subject.gem("foo")
+        end
+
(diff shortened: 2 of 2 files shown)


indirect:
@homu r+

agis-:
@homy retry? :)

agis-:
@indirect this probably needs a @homu retry?

indirect:
ugh, why did this never finish...

@homu retry

agis-:
@indirect I think @homu doesn't like this patch for 1.10.. :smiley:

indirect:
Ugh, well, I will force-merge it then. :grimacing:

PR #3983, opened 2015-09-10
Add cross-application user cache [3975]

smlance:
@indirect wrote this in response to https://github.com/bundler/bundler/pull/3975

### DIFF
--- a/lib/bundler/source/rubygems.rb
+++ b/lib/bundler/source/rubygems.rb
@@ -450,9 +450,14 @@ module Bundler
       def download_cache_path(*paths)
         raise "Caching is only possible for sources with one URL" if remotes.size > 1
         uri = remotes.first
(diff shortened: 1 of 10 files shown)


lib/bundler/source/rubygems.rb
```diff
@@ -423,6 +424,39 @@ def builtin_gem?(spec)
...
+        cache_file = download_cache_path("#{spec.full_name}.gem")
+        if cache_file && !cache_file.exist?
+          cache_path = File.dirname(cache_file)
+          FileUtils.mkdir_p(cache_path)
+          FileUtils.cp(local_cache_path, cache_path)
+        end
+      end
+
+      def download_cache_path(*paths)
+        raise "Caching is only possible for sources with one URL" if remotes.size > 1
```
segiddins: Raise a proper exception subclass?

PR #3986, opened 2015-09-12
Add ignore dependencies to avoid access network in install:local

Eric-Guo:
Otherwise it will access network if no dependencies match, even there are higher version gems available in local

### DIFF
--- a/lib/bundler/gem_helper.rb
+++ b/lib/bundler/gem_helper.rb
@@ -83,7 +83,7 @@ module Bundler
 
     def install_gem(built_gem_path = nil, local = false)
       built_gem_path ||= build_gem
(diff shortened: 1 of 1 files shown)


indirect:
If no dependencies match, will the gem even work at all?

Eric-Guo:
@indirect Bundler will not consider pre-release gems, so will try to install gems if those non-pre release gems can not found in local (which is right behavior), But then Bundler need to access network to install them even you say  `--local` (seems currently only resolve at local if given local option).

Anyway, I close this PR.

indirect:
Bundler will consider prerelease gems if the gemspec you are installing allows them! I suggest you just do that.

PR #3989, opened 2015-09-15
expect positive case, instead of negative

asymmetric:
The previous expectation didn't make it clear which version of the gem was supposed to be still there after the update.

**NOTE:** I'm not sure why, but it's version 1.0, whereas IMO it should be version 1.0.1.

This expectation also makes it easier to spot false positives, if any.

---
*SIDENOTE:* Why is either gem being cleaned? And where is the code that does this? The specs could be improved to make this clearer.

### DIFF
--- a/spec/commands/clean_spec.rb
+++ b/spec/commands/clean_spec.rb
@@ -454,7 +454,7 @@ describe "bundle clean" do
     bundle :update
 
     sys_exec "gem list"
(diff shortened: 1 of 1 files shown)


indirect:
I don't know why either gem is being cleaned. I would actually expect both rack 1.0 and 1.0.1 to be present at the end of this test. :/

indirect:
Found the problem! The test was testing the wrong thing. :) Thanks for pointing this out.

PR #3991, opened 2015-09-15
Ignore ENV["BUNDLE_GEMFILE"] if it's empty

### LINKED ISSUE #3678: brittle handling of ENV['BUNDLE_GEMFILE']
Setting `BUNDLE_GEMFILE=` (as opposed to unsetting the env variable) causes Bundler to go crazy.

The problem is with these 2 lines here: https://github.com/bundler/bundler/blob/e8c962ef2a3215cdc6fd411b6724f091a16793d6/lib/bundler/cli.rb#L19-L20

When `BUNDLE_GEMFILE` happens to be `""`:
1. `custom_gemfile` becomes "", which evaluates to `true`
2. `File.expand_path(custom_gemfile)` expands to the current directory and ...
3. `ENV['BUNDLE_GEMFILE']` now contains a *directory* instead of a file - which causes Bundler to blow up.

Use case: I'm running `bundle install` as root on one `Gemfile`, and then running another, unrelated `bundle install` on another `Gemfile` - the variable is set and propagated and deleting the env variable is tricky on my setup (my workaround is: `/usr/bin/env -u BUNDLE_GEMFILE foo`, which isn't portable).

agis-:
Fixes #3678.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -17,7 +17,7 @@ module Bundler
       super
 
       custom_gemfile = options[:gemfile] || Bundler.settings[:gemfile]
-      ENV["BUNDLE_GEMFILE"] = File.expand_path(custom_gemfile) if custom_gemfile
+      ENV["BUNDLE_GEMFILE"] = File.expand_path(custom_gemfile) if custom_gemfile && !custom_gemfile.empty?
 
       Bundler.settings[:retry] = options[:retry] if options[:retry]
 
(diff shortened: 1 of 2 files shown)


lib/bundler/cli.rb
```diff
@@ -17,7 +17,7 @@ def initialize(*args)
       super
 
       custom_gemfile = options[:gemfile] || Bundler.settings[:gemfile]
-      ENV["BUNDLE_GEMFILE"] = File.expand_path(custom_gemfile) if custom_gemfile
+      ENV["BUNDLE_GEMFILE"] = File.expand_path(custom_gemfile) if custom_gemfile && !custom_gemfile.empty?
```
simi: What about this :question: 

```ruby
ENV["BUNDLE_GEMFILE"] = File.expand_path(custom_gemfile) if not custom_gemfile.to_s.empty?
```

indirect:
Thanks!

PR #3992, opened 2015-09-16
add option in dsl "ca_cert" to specify ssl ca_certificate to verify server connection with

esse:
It's common to use self-signed ca certificates inside local work environment.

This PR add option to bundler to allow easier use of custom CA certificates - by allowing to specify `ca_cert` option inside Gemfile itself - which makes it simpler for users.

Example usage inside Gemfile: `ca_cert 'test.pem'`

### DIFF
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -124,6 +124,10 @@ module Bundler
         @sources.add_rubygems_remote(source)
       end
     end
+    
+    def ca_cert(source)
+      ENV['SSL_CERT_FILE'] = source
+    end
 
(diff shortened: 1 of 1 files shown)


segiddins:
This doesn't have tests, and I'm not sold on the benefit of us adding this to the DSL when you could just set the env variable in the Gemfile already.

indirect:
I think it makes sense for this to be an option on the source, but not a global setting. it would be great to get a PR that allows setting the CA cert for sources that need it.

esse:
@indirect this is great idea, I will try to look into it.

PR #3993, opened 2015-09-16
Persist path to configuration when there are conflicts

asymmetric:
Previously, if the user tried to set `path` when `system` was set (and viceversa), the conflicting option would be removed, but the passed option would not be saved.


indirect:
this looks good to me. @smlance, what do you think?

smlance:
Looks good!

PR #3994, opened 2015-09-16
Remove --system flag from install

asymmetric:
As [planned](https://github.com/bundler/bundler-features/issues/26#issuecomment-129487171), this PR removes the `--system` flag from `install`.

Users are expected to use `bundle config path.system true` instead.

Also fixes a bug introduced in ddafe06ab60c42943cc7c8d64973d4745d6a26ca where passing the `--system` option would set `Bundler.settings[:path]`, which in turn would be used to `disable_shared_gems` when doing a system install - which seems wrong.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -142,8 +142,6 @@ module Bundler
       "Specify a different shebang executable name than the default (usually 'ruby')"
     method_option "standalone", :type => :array, :lazy_default => [], :banner =>
       "Make a bundle that can work without the Bundler runtime"
-    method_option "system", :type => :boolean, :banner =>
-      "Install to the system location ($BUNDLE_PATH or $GEM_HOME) even if the bundle was previously installed somewhere else for this application"
     method_option "trust-policy", :alias => "P", :type => :string, :banner =>
       "Gem trust policy (like gem install -P). Must be one of " +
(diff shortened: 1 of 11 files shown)


segiddins:
Seems good

asymmetric:
Just to be clear: should `disable_shared_gems` be **on** or **off** when `path.system == true`?

I'd say **off**, as I consider "shared" to mean "system" in this context. The previous implementation however did the opposite thing, and there's a [card](https://trello.com/c/yGsPNDpg/48-stop-auto-remembering-any-command-flags) on Bundler's Trello where it's explicitly stated that it should be **on**.

So, which is it?

indirect:
Sorry for the confusion--disable_shared_gems should be set to true anytime that `path` is set, but false anytime `path.system` is set. In fact, the setting doesn't even need to be written out into the config file, and we can just use the path/path.system setting instead.

In a very old version of Bundler, about six years ago, it was possible to use `path` and still use shared/system gems at the same time. It's turned out to be very confusing for users, and so we stopped allowing it before Bundler 1.0. Then we never got around to removing the setting. :P

segiddins:
Does this need anything else, or is it ready for review?

asymmetric:
There's no more work to be done here AFAIK, so I'd say it's ready.

indirect:
:+1: Looks right to me.

@homu r+

PR #3995, opened 2015-09-17
Update deployment mode explanation

asymmetric:
The deployment mode section in `install`s man page still read as if installation to the system path were the default.


segiddins:
@homu r+

PR #3996, opened 2015-09-17
Update config man for 2.0

asymmetric:
Fixes some outdated information in `bundle config`'s man page:

* Avoids using invalid flags
* Tells users to pass options to `bundle config` instead of `bundle install`


segiddins:
👍🏻

segiddins:
@homu r+

PR #3999, opened 2015-09-18
Remove `disable_shared_gems` flag

asymmetric:
As [mentioned](http://git.io/vnqjO), this flag is a leftover from a previous implementation.


spec/commands/config_spec.rb
```diff
@@ -11,15 +11,12 @@
   describe "BUNDLE_APP_CONFIG" do
     it "can be moved with an environment variable" do
       ENV["BUNDLE_APP_CONFIG"] = tmp("foo/bar").to_s
-      bundle "config path vendor/bundle"
-      bundle "install"
+      bundle "config --local foo bar"
```
asymmetric: NOTE: I'm not sure why `--local` is needed here, but it is.

indirect:
Thanks!

PR #4002, opened 2015-09-19
Clean RUBYLIB as well in `with_clean_env`

### LINKED ISSUE #3982: with_clean_env does not replace original RUBYLIB environmental variable
[with_clean_env](https://github.com/bundler/bundler/blob/9cb77343377856af5920691d99ba9075a8bccec2/lib/bundler.rb#L274-L284) does not reset the RUBYLIB environmental variable. This results in any calls to bundler in that environment will use the current bundler version. For example, I would expect
```
bundle _1.10.0_ exec ruby -e "require 'bundler'; Bundler.clean_system('bundle _1.9.1_ exec ruby -e \"require \\'bundler\\'; puts Bundler::VERSION\"')"
```
to return 1.9.1, but it returns 1.10.0.

agis-:
Fixes #3982.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -275,10 +275,17 @@ module Bundler
       with_original_env do
         ENV["MANPATH"] = ENV["BUNDLE_ORIG_MANPATH"]
         ENV.delete_if {|k, _| k[0, 7] == "BUNDLE_" }
-        if ENV.key? "RUBYOPT"
+
+        if ENV.key?("RUBYOPT")
           ENV["RUBYOPT"] = ENV["RUBYOPT"].sub "-rbundler/setup", ""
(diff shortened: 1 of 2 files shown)


agis-:
@indirect The original branch [`3982-clean-rubylib`](https://github.com/bundler/bundler/commits/3982-clean-rubylib) was based on `1-10-stable`. However, since the issue is also present on `master` also, this PR targets `master` (this required fixing some rubocop offenses too).

I didn't yet squash for clarity (I'll squash after the review).

segiddins:
Looks good! 👍🏻

agis-:
Squashed.

indirect:
@homu r+

ojab:
Please take a look @ bundler/issues#4165, looks like this commit has broken [rails's spring](https://github.com/rails/spring).

PR #4005, opened 2015-09-20
Revert to using sub-processes

### LINKED ISSUE #4003: Change `path` setting test back to subprocess
Inadvertently changed to testing the running test process instead of a new process, should be changed back:  https://github.com/bundler/bundler/commit/fab9d5529949ee4b714425450742b503d4f8899e#commitcomment-13317478

asymmetric:
the current running process cannot reliably report the state of the
bundler config file when it's changed on disk (cfr. git.io/vnclk)

Closes #4003.


segiddins:
👍🏻

Issue #3692, opened 2015-05-30
Fatal: No live threads left. Deadlock?


rogierslag:
Might be related to https://github.com/bundler/bundler/issues/2813

We have this problem with our testing infrastructure, where I'm unable to build new Docker containers. Since we test in Docker, in order to keep the testing speed up we have pre-prepared containers, which house all the gems already. In case of a new gem, only a little bit needs to be installed instead of the whole thing (reducing lead time). However, while preparing a new container, Bundle seems to encounter a deadlock.

The following script is used to build this container

```bash
#!/bin/bash
#
# Map the application in Docker to `/test`, so we have `/test/Gemfile` and `/test/Gemfile.lock`

# Update everything
sudo apt-get update
sudo apt-get install curl -y
sudo apt-get update && sudo apt-get upgrade -y
curl -sL https://deb.nodesource.com/setup | sudo bash -
sudo apt-get install nodejs build-essential libmysqlclient-dev redis-server postgresql libpq-dev git -y
npm install -g bower

# Install rvm
gpg --keyserver hkp://keys.gnupg.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3
\curl -sSL https://get.rvm.io | bash -s stable --ruby
source /usr/local/rvm/scripts/rvm

# Check the ruby versions
VERSIONS=('2.2.2')

# Install all dependencies
cd /test
for v in $VERSIONS; do rvm install $v && rvm use $v && gem install bundler && bundle install --system && echo "Installed ruby-${v} and gems correctly"; done

# Clear the container a bit
sudo apt-get autoremove && sudo apt-get clean
rm -rf /var/lib/apt/lists/*
```

Setting the thread number differently with `--jobs=X` does not make any difference.

We use the following Gemfile, the most recent addition is the `pg` gem in order to prepare a switch from MariaDB to Postgresql. However, when using incremental builds, this works on our Vagrant machines.

The exact failure as reported by Bundler is

```
Error details

    fatal: No live threads left. Deadlock?
      /usr/local/rvm/gems/ruby-2.2.2/gems/bundler-1.10.1/lib/bundler/worker.rb:35:in `pop'
      /usr/local/rvm/gems/ruby-2.2.2/gems/bundler-1.10.1/lib/bundler/worker.rb:35:in `deq'
      /usr/local/rvm/gems/ruby-2.2.2/gems/bundler-1.10.1/lib/bundler/installer/parallel_installer.rb:94:in `process_specs'
[...]

PR #3800, opened 2015-07-01
switch reject to select

pducks32:
Fixes #3799 
Reject is clearly not the right method to use here, now I’m trying to figure out what state of mind I was in when I wrote something that is so obviously wrong.

PR #4012, opened 2015-09-24
Now won't wait for dependencies that will never install. Partial fix for #3692

lukaso:
Before this fix, if for any reason a dependency is not on the install list
it will cause a deadlock. This now means the bundle will no longer fail
in that instance, which matches the behavior of the sequential installer.
It isn't clear why the gem is not making it onto the list however, which
appears to be a separate bug.

This fixes my instance of #3692 which was still occurring despite the issue being closed.

### DIFF
--- a/lib/bundler/installer/parallel_installer.rb
+++ b/lib/bundler/installer/parallel_installer.rb
@@ -41,8 +41,8 @@ class ParallelInstaller
     # Represents only the non-development dependencies, the ones that are
     # itself and are in the total list.
     def dependencies(all_spec_names)
-      @dependencies ||= all_dependencies.reject {|dep| ignorable_dependency? dep }
-                                        .select {|dep| all_spec_names.include? dep.name }
+      @dependencies ||= all_dependencies.reject {|dep| ignorable_dependency? dep }.
+                                         select {|dep| all_spec_names.include? dep.name }
     end
 
     # Represents all dependencies
--- a/lib/bundler/installer/parallel_installer.rb
+++ b/lib/bundler/installer/parallel_installer.rb
@@ -35,13 +35,14 @@ class ParallelInstaller
     # sure needed dependencies have been installed.
     def dependencies_installed?(all_specs)
       installed_specs = all_specs.select(&:installed?).map(&:name)
-      dependencies.all? {|d| installed_specs.include? d.name }
+      dependencies(all_specs.map(&:name)).all? {|d| installed_specs.include? d.name }
     end
 
-    # Represents only the non-development dependencies and the ones that
-    # are itself.
-    def dependencies
+    # Represents only the non-development dependencies, the ones that are
+    # itself and are in the total list.
+    def dependencies(all_spec_names)
       @dependencies ||= all_dependencies.reject {|dep| ignorable_dependency? dep }
+                                        .select {|dep| all_spec_names.include? dep.name }
     end
 
     # Represents all dependencies
--- a/spec/install/parallel/spec_installation_spec.rb
+++ b/spec/install/parallel/spec_installation_spec.rb
@@ -57,5 +57,18 @@ describe ParallelInstaller::SpecInstallation do
         expect(spec.dependencies_installed?(all_specs)).to be_falsey
       end
     end
+
+    context "when dependencies that are not on the overall installation list are the only ones not installed" do
+      it "returns true" do
+        dependencies = []
+        dependencies << instance_double("SpecInstallation", :spec => "alpha", :name => "alpha", :installed? => true, :all_dependencies => [], :type => :production)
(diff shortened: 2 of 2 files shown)


lukaso:
It also builds on this pr #3800. The original problem was that a dependency 'therubyracer' wasn't in the `DEPENDENCIES` section of the `Gemfile.lock`, which led to the deadlock condition. Now the parallel version will fail silently, just as the sequential version does.

Once I added `therubyracer` into the `DEPENDENCIES` section, my `Gemfile.lock` changed in the following way:

```
--- a/.bundle/config
+++ b/.bundle/config
@@ -1,6 +1,6 @@
 ---
 BUNDLE_PATH: vendor/bundle
 BUNDLE_DISABLE_SHARED_GEMS: '1'
-BUNDLE_JOBS: '4'
-BUNDLE_CACHE_ALL: true
-BUNDLE_CACHE_ALL_PLATFORMS: true
+BUNDLE_JOBS: 4
+BUNDLE_CACHE_ALL: 'true'
+BUNDLE_CACHE_ALL_PLATFORMS: 'true'
diff --git a/Gemfile.lock b/Gemfile.lock
index 8fef9e1..84c25f4 100644
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -548,6 +548,7 @@ GEM
       addressable (~> 2.3)
     letter_opener (1.1.2)
       launchy (~> 2.2)
+    libv8 (3.16.14.11)
     listen (3.0.3)
       rb-fsevent (>= 0.9.3)
       rb-inotify (>= 0.9)
@@ -685,6 +686,7 @@ GEM
     redis (3.2.1)
     redis-namespace (1.5.2)
       redis (~> 3.0, >= 3.0.4)
+    ref (2.0.0)
     request_store (1.0.5)
     require_all (1.3.2)
     responders (1.1.2)
@@ -782,6 +784,9 @@ GEM
     teaspoon-jasmine (2.2.0)
       teaspoon (>= 1.0.0)
     terminal-table (1.5.2)
+    therubyracer (0.12.2)
+      libv8 (~> 3.16.14.0)
+      ref
     thor (0.19.1)
     thread_safe (0.3.5)
     tilt (1.4.1)
```

I'll see if there is an issue for this or open one.

segiddins:
I'm just interested in how we even get into that invalid state to begin with... I'm tempted to add some informative assertions about the invarients around the installer to try and catch this

indirect:
My best guess is git merges that create mutant lock files?

segiddins:
Ah, so we're trusting the lock file there to be correct, but in reality its missing some lines? Seems like a bug elsewhere in the installer then, and this is just a symptom

indirect:
Yes and yes. I’m happy to accept this and at least back to the -j 1 brokenness rather than throw a deadlock, though, and investigate ways to detect bad lockfiles as a separate thing.

segiddins:
@indirect sounds good

lukaso:
In my case it was definitely a bad merge. Detecting and throwing a better
error is relatively straightforward in the parallel installer (throw an
exception in the 'dependencies_installed?' method), but harder in the
sequential. One option might be to get rid of the sequential installer and
use parallel with one worker by default.

indirect:
I like that idea, assuming that we’re able to make it so that the parallel installer with one worker actually works all the time. :)

segiddins:
CI failure: https://travis-ci.org/bundler/bundler/jobs/82041909#L251

lib/bundler/installer/parallel_installer.rb
```diff
@@ -35,13 +35,14 @@ def ignorable_dependency?(dep)
     # sure needed dependencies have been installed.
     def dependencies_installed?(all_specs)
       installed_specs = all_specs.select(&:installed?).map(&:name)
-      dependencies.all? {|d| installed_specs.include? d.name }
+      dependencies(all_specs.map(&:name)).all? {|d| installed_specs.include? d.name }
```
segiddins: pretty sure this change is breaking things?

lukaso: I fixed that particular error but it looked like Travis went wonky after my last commit.

lukaso:
Should I resubmit this PR against `1-10-stable`? I realised that I didn't fully follow the bug fix procedure.

Also, there's no indication in the procedure (which I should have followed) about `rubocop`. Even on master where there's now a rake task.

lib/bundler/installer/parallel_installer.rb
lukaso: @segiddins I don't see any failing tests for this line in Travis. Or am I missing something?

segiddins:
@lukaso well manually cherry pick back to stable once this is merged.

lukaso:
This is superceded by #4020. Closing.

PR #4017, opened 2015-09-26
Show ruby engine and version

A5308Y:
instead of a static string.

### DIFF
--- a/lib/bundler/installer.rb
+++ b/lib/bundler/installer.rb
@@ -2,6 +2,7 @@ require "erb"
 require "rubygems/dependency_installer"
 require "bundler/worker"
 require "bundler/installer/parallel_installer"
+require "bundler/installer/standalone"
 
 module Bundler
   class Installer < Environment
@@ -68,7 +69,7 @@ module Bundler
       install(options)
 
       lock unless Bundler.settings[:frozen]
-      generate_standalone(options[:standalone]) if options[:standalone]
+      Standalone.new(options[:standalone], @definition).generate if options[:standalone]
     end
 
     def install_gem_from_spec(spec, standalone = false, worker = 0, force = false)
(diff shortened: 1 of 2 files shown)


segiddins:
@A5308Y good catch! think you could add a test case for this?

A5308Y:
Wow! That was fast! Actually I just stumbled upon this without any real understanding of the code. I'll try to add a test tomorrow (I don't have time today). If I don't understand enough: Where do I get help? Is the bundler slack a good place?

segiddins:
Awesome! Feel free to just ask questions on the PR and we'll do our best to answer them :)

A5308Y:
Ok. I was wrong. The not interpolated interpolation variables are intentional. Because this confused me and wasn't immediately obvious to @segiddins as well I refactored a bit to ease understanding. I'm not sure about some code quality preferences. This is my first attempt at contributing to bundler. Please tell me, if you think I might understanding harder somewhere.
I have one failing spec: spec/commands/help_spec.rb:32, but that failed before and doesn't seem to be related.

lib/bundler/installer.rb
```diff
@@ -67,7 +68,7 @@ def run(options)
       install(options)
 
       lock unless Bundler.settings[:frozen]
-      generate_standalone(options[:standalone]) if options[:standalone]
+      Standalone.new(options[:standalone], @definition).generate if options[:standalone]
```
asymmetric: I wonder if it would make more sense for this to be `Standalone.generate(options[:standalone], @definition)`. `Standalone` could then be turned into a `Module`.

Don't see much sense in keeping state in an object, when we throw the object away.

indirect: Making Standalone its own class seems good to me. :+1:

A5308Y:
Rebased and converted Standalone into a module.
Thanks for the feedback everyone!

@asymmetric: I usually avoid Singletons in Ruby because I find them harder to test. In this case Standalone is still tested through the installer, so that might not be a concern. 
To make sure I don't miss anything:
1. The benefit of using a Singleton here is less memory usage. Is there another one?
2. Why use a module instead of a Singleton *class*? I browsed the web a bit and the reason I found was so that initialize is undefined. Is that it, or is there another reason?

indirect:
Ahh, shoot. I meant that it's good to turn it into a class, and it doesn't need to be a module. I really don't think that single object allocation is going to make any practical difference. I think it's actually better to use a regular class than a singleton class here—a singleton makes it impossible to use Standalone twice with different parameters, and also means that instances can never be garbage collected. With Standalone as a class, it can be instantiated, perform its work, and then be GCed.

Sorry for the confusion!

A5308Y:
@indirect: My bad. I read your comment wrong. I'll just push 428e08a again.

segiddins:
https://travis-ci.org/bundler/bundler/jobs/82685824#L251 rubocop failure.

A5308Y:
Wow. This takes a lot of attempts. Is that common?

segiddins:
If you run the specs locally before pushing, you won't have to wait until Travis fails to get feedback

A5308Y:
I just reread my comment. I didn't want to sound critical. I did. I'm sorry... It's a shame there's no tone of voice on the internet. I should've added some emoticons maybe. I just wanted some hope that it will get better, because this is my first real attempt to contribute to open source.
I really appreciate all your feedback here and your work for the ruby community in general! Thank you for both.

I did run the specs locally, I just always get one failing spec (spec/commands/help_spec.rb:32). And this one fails on bundler/master as well (on my machine) everything else is green. 
Running Rubocop on my machine shows 3 offenses:

lib/bundler/dsl.rb:446:34: C: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
          indicator  = indent.tr('#', ">")
                                 ^^^
lib/bundler/installer/standalone.rb:29:49: C: Style/StringLiterals: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.
          gem_path(path, spec).sub(version_dir, '#{ruby_engine}/#{ruby_version}')
                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
spec/support/indexes.rb:65:13: C: Style/OneLineConditional: Favor the ternary operator (?:) over if/then/else/end constructs.

So I didn't think you would have them checked for PRs.
It seems to me that this latest failing check only happens on 1.8.7, so I guess I'll have to install that and figure out what syntax wasn't supported then. I'll do that.
I thought it would take too long time running the specs for all supported ruby versions locally. Maybe there's a way to speed it up I don't know about?

But all this lead me not to rely on my local results for specs and rubocop that much.

Ok. So. Sorry again. Thanks again and I'll dig into ruby changes from 1.8.7 to 1.9.3 as soon as I can.

segiddins:
@homu r+

nicely done, @A5308Y !

A5308Y:
Thanks @segiddins!

PR #4018, opened 2015-09-26
Removes duplication from bundle clean command

philnash:
I was not feeling inspired, but I do like to delete code! Here's a small simplification as reported [in Code Climate](https://codeclimate.com/github/bundler/bundler/Bundler::Runtime).


segiddins:
Awesome, thanks @philnash !
@homu r+

segiddins:
@homu retry

PR #4016, opened 2015-09-26
Friendly filesystem permission errors

agis-:
This introduces `SharedHelpers#filesystem_access` which can wrap any piece of code that makes a modification to the file system (eg. creating a directory). Doing so, results in user-friendly errors instead of the original ones (ie `Errno:EACCES`).

In the future we could (and probably should) handle more error types within `#filesystem_access`.

Closes #4009.
Closes #3988.

https://trello.com/c/OicgrmDd/102-improve-permissions-error-messages.

PR #4019, opened 2015-09-26
Improve permissions error messages

robinboening:
Related to this: https://trello.com/c/OicgrmDd/102-improve-permissions-error-messages
Greetings from ROSSConf with @isidorapetreska, @gregkare


segiddins:
I think we're going to go with https://github.com/bundler/bundler/pull/4016 since it's a more comprehensive solution, but I hope @agis- takes that test case :)
Thanks for the PR, @robinboening  !

PR #4020, opened 2015-09-29
Error on missing dependency - partial fix for #3692 (alternative)

lukaso:
This pull request implements the strategy suggested in the comments for https://github.com/bundler/bundler/pull/4012, namely to raise an error when dependencies are not present (due to a corrupted `Gemfile.lock`).

This also removes the `install_sequentially` option.

I wasn't totally sure about the error message formatting.

### DIFF
--- a/lib/bundler/installer.rb
+++ b/lib/bundler/installer.rb
@@ -215,7 +215,9 @@ module Bundler
     def generate_standalone(groups)
       standalone_path = Bundler.settings[:path]
       bundler_path = File.join(standalone_path, "bundler")
-      FileUtils.mkdir_p(bundler_path)
+      SharedHelpers.filesystem_access(bundler_path) do |p|
+        FileUtils.mkdir_p(p)
+      end
 
       paths = []
 
@@ -262,7 +264,9 @@ module Bundler
     end
 
     def create_bundle_path
-      Bundler.mkdir_p(Bundler.bundle_path.to_s) unless Bundler.bundle_path.exist?
+      SharedHelpers.filesystem_access(Bundler.bundle_path.to_s) do |p|
+        Bundler.mkdir_p(p)
+      end unless Bundler.bundle_path.exist?
     rescue Errno::EEXIST
       raise PathError, "Could not install to path `#{Bundler.settings[:path]}` " \
         "because of an invalid symlink. Remove the symlink so the directory can be created."
--- a/lib/bundler/installer.rb
+++ b/lib/bundler/installer.rb
@@ -1,6 +1,7 @@
 require "erb"
 require "rubygems/dependency_installer"
 require "bundler/worker"
+require "bundler/installer/parallel_installer"
 
 module Bundler
   class Installer < Environment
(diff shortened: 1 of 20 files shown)


lukaso:
referenceing #3692 as well.

lib/bundler/installer/parallel_installer.rb
```diff
@@ -35,13 +35,21 @@ def ignorable_dependency?(dep)
...
-    def dependencies
-      @dependencies ||= all_dependencies.reject {|dep| ignorable_dependency? dep }
+    # Represents only the non-development dependencies, the ones that are
+    # itself and are in the total list.
+    def dependencies(all_spec_names)
+      @dependencies ||= begin
+        deps = all_dependencies.reject {|dep| ignorable_dependency? dep }
+        missing = deps.reject {|dep| all_spec_names.include? dep.name }
+        if missing.size > 0
+          raise Bundler::GemfileError, "Your Gemfile.lock is corrupt. The following #{missing.size > 1 ? "gems are" : "gem is"} missing\n" \
```
segiddins: Don't manually try and line break the output

segiddins:
This looks pretty good to me

lib/bundler/installer/parallel_installer.rb
lukaso: There were a few other cases where that was happening for errors. For example here: https://github.com/bundler/bundler/blob/90c6d3ad2afcd9085db2d8ef8b5c55f1e8e37284/lib/bundler/dsl.rb#L92-L93

Happy to change it.

segiddins: I don't like it, I think bundler should take care of wrapping things it prints via `Bundler.ui` instead of doing it manually, otherwise things start to look silly on a 300 character terminal width

lukaso: If I take the newline out, is that enough, or is there an example I can copy from?

segiddins: you should be good to just take it out :)

lukaso: Done in ced3616

lib/bundler/installer/parallel_installer.rb
```diff
@@ -35,13 +35,21 @@ def ignorable_dependency?(dep)
...
-    def dependencies
-      @dependencies ||= all_dependencies.reject {|dep| ignorable_dependency? dep }
+    # Represents only the non-development dependencies, the ones that are
+    # itself and are in the total list.
+    def dependencies(all_spec_names)
+      @dependencies ||= begin
+        deps = all_dependencies.reject {|dep| ignorable_dependency? dep }
+        missing = deps.reject {|dep| all_spec_names.include? dep.name }
+        if missing.size > 0
+          raise Bundler::GemfileError, "Your Gemfile.lock is corrupt. The following #{missing.size > 1 ? "gems are" : "gem is"} missing " \
```
segiddins: this should be a `LockfileError`

segiddins:
I'm :+1:  on this, but I want a confirmation from @indirect before I merge

lib/bundler/installer.rb
```diff
@@ -1,6 +1,7 @@
 require "erb"
 require "rubygems/dependency_installer"
 require "bundler/worker"
+require "bundler/installer/parallel_installer"
```
asymmetric: How come this wasn't needed before?

segiddins: because it wasn't loaded when jobs was == 1

segiddins:
❤️

@homu r+

PR #4021, opened 2015-09-29
Use exe instead of bin for BUNDLE_BIN_PATH fallback

jaym:
Looks like these got moved.
I'm having a hard time getting this to actually fail in the tests. Here's a gist of something that fails:

Gemfile
----------
```ruby
source 'https://rubygems.org'
gem 'rake'
```

Rakefile
-----------
```ruby
desc "foo"
task :foo do
  sh 'bundle -v'
end
```

> bundle install
> bundle install --deployment
> bundle exec rake foo

```
bundle -v
/Users/jmundrawala/.rbenv/versions/2.1.2/bin/bundle:23:in `load': cannot load such file -- /Users/jmundrawala/.rbenv/versions/2.1.2/lib/ruby/gems/2.1.0/gems/bundler-1.10.6/bin/bundle (LoadError)
	from /Users/jmundrawala/.rbenv/versions/2.1.2/bin/bundle:23:in `<main>'
rake aborted!
Command failed with status (1): [bundle -v...]
/Users/jmundrawala/workspace/bundler/foo/Rakefile:4:in `block in <top (required)>'
Tasks: TOP => foo
```

### DIFF
--- a/lib/bundler/runtime.rb
+++ b/lib/bundler/runtime.rb
@@ -192,7 +192,7 @@ module Bundler
       begin
         ENV["BUNDLE_BIN_PATH"] = Bundler.rubygems.bin_path("bundler", "bundle", VERSION)
       rescue Gem::GemNotFoundException
-        ENV["BUNDLE_BIN_PATH"] = File.expand_path("../../../bin/bundle", __FILE__)
+        ENV["BUNDLE_BIN_PATH"] = File.expand_path("../../../exe/bundle", __FILE__)
       end
 
       # Set BUNDLE_GEMFILE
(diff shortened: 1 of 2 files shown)


segiddins:
We'll need a test for this, otherwise good catch!

jaym:
Added a spec, however I couldn't actually get it to fail with the old code. It might be something weird about my environment. Thoughts?

segiddins:
Ah so I can't think of when the real `bin_path` won't exist, but this change is definitely correct. :+1: 

@indirect ideas on how to test?

indirect:
Inside the spec, inside `gemfile`, monkeypatch `Bundler.rubygems.bin_path` to raise `Gem::GemNotFound`. :sweat_smile:

segiddins:
@jaym any chance you could try out Andre's suggestions?

jaym:
@segiddins 
hah, i thought he was joking.
Anyway, tried it, still does not fail when i run against master. Pushed the changes up anyway, maybe I'm doing something stupid

segiddins:
@homu r+

PR #4027, opened 2015-10-03
Fix Style/TrailingUnderscoreVariable

esasse:


asymmetric:
Honestly not sure why the second form is considered better - I'd argue it's even less clear.

esasse:
I agree, just sent the PR so that we can discuss and decide if we'll use rubocop default or the current style with underscore.

segiddins:
I'm :+1: with this

agis-:
IMHO clearest of both is:

```ruby
out, _foo = sh_with_code("gem install '#{built_gem_path}'#{" --local" if local}")
```

Ruby won't warn for unused variables beginning with `_`.

indirect:
I agree that leaving it out is less understandable than just making it `_`. Let's set the rubocop styles to require just a single `_`, like we do elsewhere when we have unused variables.

esasse:
Ok, so what I did was just move the config to the main rubocop file since this is the setting we want for the project. And I could not find a way to enforce this style.

segiddins:
@homu r+

PR #4033, opened 2015-10-04
Also print gemspecs in `bundle env`

agis-:
This change has the nice side-effect that `bundle env` will throw an error for Gemfiles with invalid DSLs.

Closes https://trello.com/c/QOX9jyYN.


agis-:
Sample output: https://gist.github.com/agis-/2e753ebd4d89fc81d644.

segiddins:
👍🏻 if we can add a simple test for this

agis-:
@segiddins Added test.

P.S. In the future I will add some more tests about `Env`.

segiddins:
Awesome!
@homu r+

PR #4038, opened 2015-10-06
clean: add test for not removing git extensions artifacts

Sirupsen:
Unfortunately https://github.com/bundler/bundler/pull/4030 was merged before I had a chance to write a test. I confirmed that the test did not pass before my patch.

Any ETA of 1.10.7 which includes #4030? :heart: 

@indirect @segiddins @andremedeiros @skottler

### DIFF
--- a/spec/commands/clean_spec.rb
+++ b/spec/commands/clean_spec.rb
@@ -613,4 +613,26 @@ describe "bundle clean" do
     should_have_gems "thin-1.0", "rack-1.0.0", "weakling-0.0.3"
     should_not_have_gems "foo-1.0"
   end
+
+  it "doesn't remove extensions artifacts from bundled git gems after clean", :rubygems => "2.2" do
+    build_git "very_simple_git_binary" do |s|
+      s.add_c_extension
+    end
+
+    revision = revision_for(lib_path("very_simple_git_binary-1.0"))
+
+    gemfile <<-G
+      source "file://#{gem_repo1}"
+
+      gem "very_simple_git_binary", :git => "#{lib_path("very_simple_git_binary-1.0")}", :ref => "#{revision}"
+    G
+
+    bundle "install --path vendor/bundle"
+    bundle :clean
+
+    expect(out).to eq("")
+
+    expect(vendored_gems("bundler/gems/extensions")).to exist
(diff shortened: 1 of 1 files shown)


indirect:
@Sirupsen awesome, thanks for following up!

@homu r+

Sirupsen:
@homu retry

Sirupsen:
Looks like a bug with older Rubygems versions. Will take a look.

Sirupsen:
It looks like the only affected Rubygems version of my specific symptom is `2.2.x`, as compilation on `#activate` was removed [in `2.3.x`](https://github.com/rubygems/rubygems/blob/master/History.txt#L302). However, it can still compile unnecessarily in other scenarios on later versions.

Sirupsen:
No idea why that test is failing :(

indirect:
@Sirupsen you're right! it wasn't you. rack-cache just released a new version that cannot be installed on ruby 1.9, breaking all the 1.9 specs. I just fixed master, so Homu should be able to build successfully now!

indirect:
@homu retry

indirect:
@homu r+

indirect:
boo, I screwed it up, going to have to have to retry again

indirect:
@homu retry

Sirupsen:
@indirect :sparkles: Sorry to ask again, but any news on when a new version of bundler will be out that includes this fix? Would really help us in production to speed up our builds in certain cases and avoid compiling Nokogiri when booting a Rails console (and avoid running a fork!). :)

segiddins:
We don't follow a fixed release schedule, and given that this is the only fix to be backported to the 1.10 series, I'm not entirely sure that I can provide a window for a release. In the meantime, you're welcome to use bundler off of master.

Sirupsen:
@segiddins that's fair, thanks for the response! We'll build off `master`.

rafaelfranca:
@segiddins will this with the related patch backported to the 1-10-stable branch or you only backport when releasing? I'm not seeing this commit there https://github.com/bundler/bundler/compare/v1.10.6...1-10-stable

segiddins:
Well only backport if we do a patch release.

PR #4040, opened 2015-10-06
Send X-Gemfile-Source header to source mirrors

agis-:
When there is a mirror configured for a rubygems source, we now set the
'X-Gemfile-Source' header, which contains the original source URI and is
sent along with the requests to the rubygems server.

This is discussed in https://github.com/bundler/gemstash/issues/2.

Closes https://trello.com/c/iysZYwlD.


pcarranza:
:+1:

smellsblue:
Ok, so dependencies work perfect with Gemstash, but when it comes to downloading the gem, this fails, and I think I found why.

When downloading the gem, the code goes through https://github.com/bundler/bundler/blob/master/lib/bundler/rubygems_integration.rb#L586 which as you can see invokes `mirror_for` separate from where you capture that the mirror is happening. Even in this file there are other calls to `mirror_for` that might be worth looking at. However, outputting the `uri` at this point reveals that it isn't the problem... it is already set as the mirror location at this point.

Instead, if you go up the stack, you get to this: https://github.com/bundler/bundler/blob/master/lib/bundler/source/rubygems.rb#L400 which **is** the problem. If you dump `spec.remote` you will see that it is indeed an instance of the `Remote` that you've modified to capture the original uri, and it has both the URI and the original URI. Unfortunately, you can see here that it is just pulling out the URI before the header can even be considered to be added. Nevermind that the `download_gem` method is going to then check for a mirror... that will further complicate a possible fix here :disappointed:

indirect:
Ahh, dang. :( We'll probably have to implement downloading gems ourselves rather than relying on the rubygems method for this to work. Sorry. :/

pcarranza:
@indirect Should we open an issue for the case of probing that the gemstash server is there? just opening a TCP connection to the upstream, and if that fails in half a second instead of using the mirror use the actual upstream server?

agis-:
@indirect So we *do* want to do this, right?

indirect:
@agis- yes, I think the gain in functionality is worth the effort. Maybe we can just copy the implementation from rubygems for now?

agis-:
@indirect Copying the implementation from Rubygems is not very easy, since the path is quite long:

1) `Gem::RemoteFetcher#download`
2) `Gem::RemoteFetcher#cache_update_path`
3) `Gem::RemoteFetcher#fetch_path`
4) `Gem::RemoteFetcher#request`
5) and finally `Gem::Request#fetch`

If we could hijack the `Gem::Request` object things would've been easier since we could just do `request.add_header('X-Gemfile-Source', foo)`.

Or were you talking about copying `Gem::RemoteFetcher` and friends all together?

indirect:
We already monkey-patch a lot of other RubyGems things, it wouldn't be the worst thing ever if we were to extend Gem::Request. But if we can do it another way, that would be better. :)

agis-:
@smellsblue This still lacks some tests but can you give it a try now?

segiddins:
LGTM :+1:

agis-:
@smellsblue Added tests for when downloading the gem. Can you try now? /cc @indirect

smellsblue:
@agis- my tests pass :+1: thanks for all the work!

What version would this be in? I have a placeholder in the Gemstash docs to indicate which version of Bundler supports mirroring with this header.

segiddins:
This will probably be in 1.11

indirect:
Awesome, thanks! This should land in 1.11, which I expect to be out (at least as a prerelease) within 4 weeks.

PR #4041, opened 2015-10-06
Fix bitbucket source deprecation warning

Zorbash:
The deprecation warning tries to access `user_name` which is not yet
assigned and falls back to `method_missing` which raises an error.

### DIFF
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -230,8 +230,8 @@ module Bundler
 
       # TODO 2.0 remove this deprecated git source
       git_source(:bitbucket) do |repo_name|
-        warn_deprecated_git_source(:bitbucket, "https://#{user_name}@bitbucket.org/#{user_name}/#{repo_name}.git")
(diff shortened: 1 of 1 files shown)


segiddins:
Looks like we need a test that hits the bitbucket code path...

Zorbash:
@segiddins There are currently 2 failing examples hitting the bitbucket code path:

* https://github.com/bundler/bundler/blob/1-99-dev/spec/bundler/dsl_spec.rb#L59
* https://github.com/bundler/bundler/blob/1-99-dev/spec/bundler/dsl_spec.rb#L53

segiddins:
@Zorbash but why did those specs not catch this failure?

Zorbash:
@segiddins those specs seem to catch this failure. See [1-99-dev failed build](https://travis-ci.org/bundler/bundler/bundler/jobs/71274288#L4480).

agis-:
:+1: Let's merge this.

PR #4052, opened 2015-10-12
Remove excessive periods from DSL error messages

agis-:
Apparently, a period is already added when interpolating those messages,
some lines below.


segiddins:
@homu r+

agis-:
I guess this needs a retry.

PR #4063, opened 2015-10-17
Extracting a new "GemInstaller" from installer.rb

A5308Y:
My goal was to reveal the main part of install_gem_from_spec. From my
perspective the main part is:
1. Install
2. Generate stubs
3. Print message to the user
(4. handle exceptions)

While working on it I found that Bundler::Installer's run method and
instance variables weren't used when installing gems with
install_gem_from_spec. The installer was (and still is) only used to
generate executable stubs.
So I inserted the new GemInstaller in ParallelInstaller; passing the
instantiated Bundler::Installer to the GemInstaller only to generate
the executable stubs.

Based on this I would continue by extracting two BinstubGenerator
classes and removing GemInstaller's dependency on Bundler::Installer,
possibly allowing for a more concise way to generate binstubs in other
parts of bundler.

I was a bit intimidated by the amount of installers and weary to add
another one to this contested namespace. So I checked in with
@indirect who approved of the general direction of the refactoring and
suggested to rename the old GemInstaller to RubyGemsGemInstaller.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -27,7 +27,7 @@ module Bundler
   autoload :Fetcher,               "bundler/fetcher"
   autoload :GemHelper,             "bundler/gem_helper"
   autoload :GemHelpers,            "bundler/gem_helpers"
-  autoload :GemInstaller,          "bundler/gem_installer"
+  autoload :RubyGemsGemInstaller,  "bundler/rubygems_gem_installer"
   autoload :Graph,                 "bundler/graph"
   autoload :Index,                 "bundler/index"
   autoload :Installer,             "bundler/installer"
--- a/lib/bundler/gem_remote_fetcher.rb
+++ b/lib/bundler/gem_remote_fetcher.rb
@@ -4,7 +4,7 @@ module Bundler
   # Adds support for setting custom HTTP headers when fetching gems from the
   # server.
   #
-  # TODO Get rid of this when and if gemstash only supports RubyGems versions
+  # TODO: Get rid of this when and if gemstash only supports RubyGems versions
   # that contain https://github.com/rubygems/rubygems/commit/3db265cc20b2f813.
   class GemRemoteFetcher < Gem::RemoteFetcher
     attr_accessor :headers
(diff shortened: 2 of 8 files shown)


segiddins:
👍🏻 I like

segiddins:
@indirect r+?

indirect:
@homu r+

indirect:
@A5308Y can you fix the failing tests? At least one is RuboCop, I'm not sure about the rest. Thanks!

A5308Y:
Sorry for the delay. I had a bike accident and hurt my dominant hand. I didn't even look at github-things for quite a while now. It's a lot better now. I still have to type carefully though. I see what I can do today in the (European) evening.

Regarding the failure:
Rubocop says not to rescue Exception but StandardError. Exception was rescued before. So I guess my change only allowed to see it. 
Would rescuing StandardError a functionality change in bundler? Would that be covered by the specs?

segiddins:
Just doing `rescue => e` should be sufficient

indirect:
I think it was a big that we previously rescued Exception, since that would stop ctrl-c from working. Let's change to StandardError (which is the default, and so you can just `rescue => e` like @segiddins said).

A5308Y:
@indirect @segiddins I fixed this on Tuesday and just realized that there might not have been a notification. So, this is it.

segiddins:
:ship:

A5308Y:
Hmm. Long time no comment. Can I do something to help the process?

segiddins:
@homu r+

barosl:
Strangely, it seems that the problem persists. The status for #4077 (9 days ago) was correctly reported by Travis as you can see in [here](https://api.github.com/repos/bundler/bundler/statuses/9e6f9d8), but starting with #4085 (6 days ago) the API shows [nothing](https://api.github.com/repos/bundler/bundler/statuses/8957464). So do the recent two PRs, including [the current PR](https://api.github.com/repos/bundler/bundler/statuses/b4ff4c4).

Did you happen to change the Travis configuration, maybe around 7 days ago?

indirect:
@barosl that's really, really strange. The Travis build log shows the merge for #4085 being built, and even references the correct commit: https://travis-ci.org/bundler/bundler/builds/88536039. Maybe Travis has a bug and is not reporting back to Github correctly? I'll check with them.

segiddins:
@homu retry

segiddins:
@homu retry

PR #4069, opened 2015-10-21
Make SSL Verify Mode a number setting

goughy000:
Needed to turn off SSL verify mode, most the code seemed to be there but the setting was being cast to a string which caused it to fail. Added it to NUMBER_KEYS and it's ok now


indirect:
Thanks! Could you add a test for this?

agis-:
LGTM

goughy000:
Thanks :)

indirect:
Thank you!

PR #4070, opened 2015-10-21
Add --shebang option to binstubs command

Zorbash:
bundle install `--binstubs` option is to be removed from Bundler 2.0 and currently 
supports setting the shebang using the `--shebang` option.

See #1467 introducing the `--shebang` option, which with this commit
is ported to the binstubs command.


segiddins:
:+1:

Zorbash:
Amended the commit with info in [UPGRADING.md](https://github.com/bundler/bundler/blob/2-0-dev/UPGRADING.md#removed-bundle-install---binstubs).

PR #4071, opened 2015-10-22
Include revision hash in Source::Git#to_s

### LINKED ISSUE #3433: bundle install - include commit hash for git source

agis-:
This effectively changes the output of commands that use Source::Git#to_s
(ie. `bundle install`). For example:

    Using rack 1.0.0 from git@github.com:rack/rack.git (at master@574b147)

Closes #3433.

### DIFF
--- a/lib/bundler/source/git.rb
+++ b/lib/bundler/source/git.rb
@@ -65,7 +65,14 @@ module Bundler
         else
           ref
         end
-        "#{uri} (at #{at})"
+
+        rev = begin
+                "@#{shortref_for_display(revision)}"
+              rescue GitError
+                nil
+              end
+
+        "#{uri} (at #{at}#{rev})"
       end
 
       def name
--- a/spec/install/deploy_spec.rb
+++ b/spec/install/deploy_spec.rb
@@ -207,7 +207,7 @@ describe "install with --deployment or --frozen" do
 
       bundle "install --deployment"
       expect(out).to include("deployment mode")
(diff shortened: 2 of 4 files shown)


agis-:
@indirect Should this be based in `2-0-dev` since it's a breaking change?

segiddins:
It's only adding info, so it should be fine?

agis-:
Cool!

P.S. I assumed changing the output of `bundle install` was considered a breaking change.

segiddins:
I'll leave that up to @indirect

indirect:
This looks good to me! The output is a superset of the current output, so I think this OK as a feature addition in the next 1.x release.

revett:
Thanks for this @agis-

agis-:
@revett You're welcome :)

PR #4075, opened 2015-10-26
Fix Gem::SystemExitException initialization

agis-:
Previously, we would attempt to initialize `Gem::SystemExitException`
without an argument, which would result in an `ArgumentError` since the
argument is required (from Rubygems 1.3 all the way to 2.4).

I don't know how we missed this all this time. Maybe that path is not
even hit anymore.

Anyway, we don't even have to initialize a new exception, just re-raising the
one we've rescued is sufficient.

### DIFF
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -66,7 +66,7 @@ module Bundler
     rescue Gem::SystemExitException => e
       Bundler.ui.error "#{e.class}: #{e.message}"
       Bundler.ui.trace e
-      raise Gem::SystemExitException
+      raise
     end
 
     def ruby_engine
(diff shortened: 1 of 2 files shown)


lib/bundler/rubygems_integration.rb
```diff
@@ -66,7 +66,7 @@ def configuration
     rescue Gem::SystemExitException => e
       Bundler.ui.error "#{e.class}: #{e.message}"
       Bundler.ui.trace e
-      raise Gem::SystemExitException
+      raise e
```
segiddins: Can just do raise without an argument here

agis-: Done!

segiddins:
@homu r+

agis-:
hmm @homu retry?

segiddins:
@homu retry

PR #4077, opened 2015-10-27
Extract exception classes to a separate file

agis-:


lib/bundler.rb
```diff
@@ -9,6 +9,7 @@
 require "bundler/version"
 require "bundler/constants"
 require "bundler/current_ruby"
+require "bundler/exceptions"
```
segiddins: `bundler/errors` ?

agis-: Done.

indirect:
@homu r+

Issue #3528, opened 2015-03-28
Default Source


billywatson:
I think intelligent defaults are not only acceptable but a generally good practice.

I feel that if there are no sources in a Gemfile, that instead of raising an error, bundler should puts a warning message to STDOUT and then use https://rubygems.org instead of telling the human to put that in the gemfile and failing out.

If this is okay, I'd be happy to issue a pull request.

indirect:
There is another use that I guess you haven't thought of, which is Gemfiles that only use local gems. We don't provide a default source so that isn't always clear where gems will come from, even if that means te gems only come from the local machine. If you don't like typing "source 'https://rubygems.org'", please run "bundle init" instead.

TimMoore:
The other problem with an intelligent default is that you can never change it without breaking backward compatibility. For something like a `Gemfile`, explicit is the way to go.

Issue #118, opened 2010-02-22
If no sources defined, should default to rubygems.org


trevorturk:
...since that's the default now...?

PR #4078, opened 2015-10-27
Add "source :default" support.

voxik:
I'd love to configure my sources once for RubyGems and later share this setting by all my projects. As far as I understand, there is no way how to achieve this at the moment, it is always necessary to specify the "source" in my Gemfile. Hence I am proposing to use "source :default" to inherit my RubyGems settings.

Please note that that although this is in principle request similar to #118, #2051 of #3528, the main difference is that this has to be explicitly enabled and hence any possible issues are user responsibility.

### DIFF
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -121,11 +121,17 @@ module Bundler
 
     def source(source, &blk)
       source = normalize_source(source)
+
+      if source == :default
+        Bundler.rubygems.configuration
+        source = Bundler.rubygems.sources
+      end
+
       if block_given?
         with_source(@sources.add_rubygems_source("remotes" => source), &blk)
       else
         check_primary_source_safety(@sources)
-        @sources.add_rubygems_remote(source)
+        Array(source || []).reverse_each {|r| @sources.add_rubygems_remote(r) }
       end
     end
 
@@ -348,7 +354,7 @@ module Bundler
           "requests are insecure.\nPlease change your source to 'https://" \
           "rubygems.org' if possible, or 'http://rubygems.org' if not."
         "http://rubygems.org"
-      when String
+      when String, :default
         source
       else
         raise GemfileError, "Unknown source '#{source}'"
--- a/spec/bundler/dsl_spec.rb
+++ b/spec/bundler/dsl_spec.rb
@@ -211,4 +211,16 @@ describe Bundler::Dsl do
       end
     end
   end
+
+  describe "#source" do
+    it ":default is using sources specified by RubyGems" do
+      other_source = double("default-source")
+      allow(Bundler::Source::Rubygems).to receive(:new).and_return(other_source)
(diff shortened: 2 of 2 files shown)


segiddins:
@voxik this will cause different behavior on each user's machine, kinda defeating the point of using bundler?

indirect:
I'm confused about what this enables... do you change the default source often? If you have control over all of the Gemfiles yourself, what is your objection to putting the server URL in the Gemfile?

strzibny:
> kinda defeating the point of using bundler?

I disagree. Bundler is used by people to lock the versions of gems they use and put them on path. Enabling them to be more flexible about sources still make Bundler to do its job.

> If you have control over all of the Gemfiles yourself

That might not be the case, you might want to provide a different source for all projects within your organization and still allow people to get the gems from RubyGems.org. E.g. let them develop the project publicly upstream, but use different source for internal builds for production.

TimMoore:
This sounds like the use case for the Gem Source Mirrors feature that was added in 1.5 http://bundler.io/v1.5/bundle_config.html#gem-source-mirrors

strzibny:
> This sounds like the use case for the Gem Source Mirrors feature that was added in 1.5 http://bundler.io/v1.5/bundle_config.html#gem-source-mirrors

I know about this feature, but then you have no option to opt-out from it.

voxik:
I want to have the same gems installed either I'll use ```gem install``` or ```bundle install```. That is what is important for my use case.

indirect:
@voxik Thanks for the pull request (and the feedback). Can you tell us about why you need `gem install` and `bundle install` to install from the same remote server?

@strzibny The mirrors feature can be used either per-project and per-user. You could opt out (or back in) with a single `bundle config` command.

The use case for this feature is extremely narrow: people who care that `gem` and `bundle` always use the same source. Since very few people need this use case, and because it is already possibly to achieve this outcome without changing any code in Bundler, I'm going to decline this PR (at least for now).

To accomplish this result without changes to Bundler, instead of `source :default` in your Gemfile, do something like `source YAML.load_file("~/.gemrc")[:sources].first`. Thanks!

voxik:
> Can you tell us about why you need gem install and bundle install to install from the same remote server?

If we have RubyGems mirror in company, I want everybody to use just this mirror, always and for both, RubyGems as well as for Bundler. Of course there are several reasons for internal RubyGems mirror, such as bandwidth, security and availability.

> The mirrors feature can be used either per-project and per-user.

But it can't be configured system wide. Even if it was possible, then you still have to configure RubyGems separately.

@indirect Would you mind to elaborate what is the real reason to reject this PR?

indirect:
@voxik the "real reason" is that the mirror setting already works for all bundles, and you are the first person to care in five years whether the `gem` command uses the same source.

You can already accomplish your goal without adding code to Bundler, and I don't want to agree to maintain perfect backwards compatibility for five years or more for code only used by a few people.

PR #4082, opened 2015-10-28
Don't rebuild git extensions

csfrancis:
@andremedeiros /cc @Sirupsen

This is somewhat related to https://github.com/bundler/bundler/pull/4030 (and requires it to work). Bundler is not checking the `gem.build_complete` for git extensions, and as a result is recompiling extensions when performing a `bundle install`, for example.

This PR simply checks if `gem.build_complete` exists and bypasses the build step. This is a similar approach to what is done in [Rubygems](https://github.com/ruby/ruby/blob/v2_1_7/lib/rubygems/specification.rb#L1420).

In my Shopify development environment this brings `bundle install` time (when there are no Gemfile changes) from nearly 20 seconds to 3 seconds.

### DIFF
--- a/lib/bundler/source/git.rb
+++ b/lib/bundler/source/git.rb
@@ -170,7 +170,11 @@ module Bundler
           @copied = true
         end
 
-        return nil if gem_build_complete?(spec.extensions_dir)
+        if spec.respond_to?(:missing_extensions?)
+          return nil unless spec.missing_extensions?
+        else
+          return nil if gem_build_complete?(spec.extensions_dir)
+        end
 
         generate_bin(spec)
 
--- a/lib/bundler/source/git.rb
+++ b/lib/bundler/source/git.rb
@@ -169,6 +169,9 @@ module Bundler
           serialize_gemspecs_in(install_path)
           @copied = true
         end
+
+        return nil if gem_build_complete?(spec.extensions_dir)
+
         generate_bin(spec)
 
         requires_checkout? ? spec.post_install_message : nil
@@ -221,6 +224,14 @@ module Bundler
       end
 
     private
+
+      def gem_complete_path(extensions_dir)
+        File.join(extensions_dir, extension_dir_name, "gem.build_complete")
+      end
+
+      def gem_build_complete?(extensions_dir)
+        File.exists? gem_complete_path(extensions_dir)
+      end
 
       def serialize_gemspecs_in(destination)
         expanded_path = destination.expand_path(Bundler.root)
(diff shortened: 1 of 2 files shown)


lib/bundler/source/git.rb
```diff
@@ -223,6 +226,14 @@ def allow_git_ops?
 
     private
 
+      def gem_complete_path(extensions_dir)
+        File.join(extensions_dir, extension_dir_name, "gem.build_complete")
+      end
+
+      def gem_build_complete?(extensions_dir)
+        File.exists? gem_complete_path(extensions_dir)
```
byroot: ` warning: File.exists? is a deprecated name, use File.exist? instead`

csfrancis:
Looks like some of the CI failures are legit - I'll investigate.

indirect:
@csfrancis how's it going fixing the specs? This definitely seems like an improvement over the current state. :+1:

csfrancis:
Sorry, I haven't had time to look at this yet. I'll try to get to it in a couple days.

csfrancis:
This is a really hard issue to debug. I've probably spent at least 6-7 hours try to figure out where things are going wrong.

At a high level, it looks like with the latest Rubygems master, the caching of `extensions_dir` is causing problems. In the test I added, the test gem is installed once and then installed again to ensure that we don't recompile the extension. The problem is that on the second installation, `extensions_dir` is not the same as it was on the first invocation, which causes the check for `gem.build_complete` to fail. On the second invocation, it looks like `extensions_dir` is indirectly cached at the beginning of the call to `bundle install`, which seems to validate all existing gem specs (this doesn't happen on the first invocation because the gem hasn't yet been installed). 

Anyway, if I have time in the next few days I'll spent a bit more time trying to figure this out.

lib/bundler/source/git.rb
```diff
@@ -170,6 +170,9 @@ def install(spec, force = false)
           serialize_gemspecs_in(install_path)
           @copied = true
         end
+
+        return nil if gem_build_complete?(spec.extensions_dir)
```
segiddins: any reason we can't just check `spec.missing_extensions?`

segiddins: arg, this seems to have broken more things :(
@indirect do we need to run `git clean -xdf` before switching branches?

indirect: Sounds like we do? :/

segiddins: nope, we copy the git dir elsewhere as the gem dir, so I'm flummoxed

csfrancis:
Ping did you guys end up making any progress on this?

indirect:
not yet, build still failing. want to take a stab at it?

segiddins:
Closing in favor of https://github.com/bundler/bundler/pull/4272.

PR #4091, opened 2015-11-03
Send X-Gemfile-Source when bundling with --full-index

### LINKED ISSUE #4090: Bundling with `--full-index` doesn't work against Gemstash with headers
Bundling against [Gemstash](https://github.com/bundler/gemstash) with the new `X-Gemfile-Source` header works (yay!), unfortunately it doesn't work when bundling with the `--full-index` option.

I believe the header isn't being sent when Bundler requests `/specs.4.8.gz`, and it's possible it might also not send the header when it requests `/quick/Marshal.4.8/some-gem-0.1.0.gemspec.rz` after pulling the specs.

You can run (roughly) the following to replicate the issue in a branch of Gemstash I have set up:
```
$ git clone git@github.com:bundler/gemstash.git
$ cd gemstash
$ git checkout tmp-bundler-headers-full-index
$ bundle
$ (cd ~/path_to_repo_for/bundler; gem uninstall bundler --executables; rake clobber; rake install)
$ rspec spec/bundler_headers_spec.rb
```

This will set up and run the specs that demonstrate the problem. One spec will pass and the other will fail. You may want to tweak what happens in the subshell, but it is important to have bundler installed from master (so you get the header behavior, otherwise neither spec will pass).

It appears the problem is starting here: https://github.com/bundler/bundler/blob/master/lib/bundler/rubygems_integration.rb#L563-L570 where Bundler requests `/specs.4.8.gz`.

smellsblue:
This fixes #4090

I did a little refactoring that made sense to me along the way, but if I went too far or might be breaking something I'm not aware of, please let me know.

**Existing issue:** This doesn't fix bundling with full index when using Rubygems < 2.0, as that code path seemed a bit trickier to fix. I added a *TODO* comment at the relevant method. I thought of a couple possibilities:
* Subclass `Gem::SpecFetcher` like `Bundler::GemRemoteFetcher`, it seems like it might be as simple as calling the super constructor and then replace `@fetcher` with a `Bundler::GemRemoteFetcher` with the headers added ahead of time (see https://github.com/rubygems/rubygems/blob/v1.8.30/lib/rubygems/spec_fetcher.rb#L66). This may involve different solutions the further back in Rubygems you go.
* Use the newer implementation of `fetch_all_remote_specs` for older Rubygems as well (I don't know if that would work though, but it might be worth a try).

I haven't yet added additional specs for this, but I wanted to get this out there first to get some feedback if possible.

### DIFF
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -184,13 +184,21 @@ module Bundler
       [] # if we can't download them, there aren't any
     end
 
-    def fetch_all_remote_specs
+    # TODO: This is for older versions of Rubygems... should we support the
+    # X-Gemfile-Source header on these old versions?
+    # Maybe the newer implementation will work on older Rubygems?
+    # It seems difficult to keep this implementation and still send the header.
+    def fetch_all_remote_specs(remote)
+      old_sources = Bundler.rubygems.sources
+      Bundler.rubygems.sources = [remote.uri.to_s]
       # Fetch all specs, minus prerelease specs
       spec_list = fetch_specs(true, false)
       # Then fetch the prerelease specs
       fetch_prerelease_specs.each {|k, v| spec_list[k] += v }
 
       spec_list
+    ensure
+      Bundler.rubygems.sources = old_sources
     end
 
     def with_build_args(args)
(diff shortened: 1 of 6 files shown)


segiddins:
Can we add a spec for setting the header?

lib/bundler/rubygems_integration.rb
```diff
@@ -560,41 +568,46 @@ def find_name(name)
...
-        proxy = configuration[:http_proxy]
-        dns = Resolv::DNS.new
-        fetcher = Bundler::GemRemoteFetcher.new(proxy, dns)
+        fetcher = gem_remote_fetcher
         fetcher.headers = { "X-Gemfile-Source" => spec.remote.original_uri.to_s } if spec.remote.original_uri
         fetcher.download(spec, uri, path)
       end
 
+      def gem_remote_fetcher
+        require "resolv"
```
segiddins: is this in the stdlib?

segiddins:
@homu r+

segiddins:
@homu retry

PR #4095, opened 2015-11-06
Remove outdated advice

deivid-rodriguez:
I'm pretty sure you're no longer supporting Rails 2.3 users


segiddins:
@indirect ?

indirect:
We are in fact still supporting Rails 2.3 users. 😕 This seems like a good change for 2.0, though.

deivid-rodriguez:
:open_mouth:

Should I open a PR to `2.0-dev`?

indirect:
Sounds good!

PR #4097, opened 2015-11-06
Improve DEVELOPMENT.md readability

deivid-rodriguez:
I was reading `DEVELOPMENT.md` in `nvim` and found it a bit unreadable. So I typed `gqG` and thought I may add a PR in case you are interested. :wink:

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -1,23 +1,48 @@
-Great to have you here! Here are a few ways you can help out with [Bundler](http://github.com/bundler/bundler).
+Great to have you here! Here are a few ways you can help out with
+[Bundler](http://github.com/bundler/bundler).
 
 # Where should I start?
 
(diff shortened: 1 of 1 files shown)


segiddins:
👍🏻

deivid-rodriguez:
Mi no entender.

segiddins:
@indirect r+?

indirect:
I don't like hard wrapping text, because I don't think forcing your window width on everyone else is a good solution. My vim is configured to soft wrap markdown, maybe you can configure your own vim to do the same?

deivid-rodriguez:
Yep, that's a good idea for projects that don't follow any line length conventions.

Perhaps you could move [this](https://github.com/bundler/bundler/blob/master/.rubocop_todo.yml#L51-L55) to the `.rubocop.yml` file and set it to `Enabled: false`.

PR #4103, opened 2015-11-11
Retry `Fetcher#specs` with `Bundler::Retry`

jingweno:
As mentioned in https://github.com/heroku/heroku-buildpack-ruby/pull/435, bundler failed at the step of fetching version metadata:

```
Fetching version metadata from https://rubygems.org/Net::HTTPInternalServerError: <?xml version=\"1.0\" encoding=\"UTF-8\"?>
       <Error><Code>InternalError</Code><Message>We encountered an internal error.
       Please try
       again.</Message><RequestId>E7402EA19C5D6803</RequestId><HostId>hzuvaA1JAZX6ST+OL4ARYeAqZ/tgkM2yOjZgBq6Panu10YzWtfNNozOg8N5qR3gxFE/sUfYGP48=</HostId></Error>
```

It turns out the retry logic doesn’t cover fetching source index, gem metadata, version metadata and dependency metadata. A new method `Fetcher#specs_with_retry` is added to wrap `Fetcher#specs` with `Bundler::Retry` which reties when bundler fails in the mentioned cases.

/cc @schneems

### DIFF
--- a/lib/bundler/fetcher.rb
+++ b/lib/bundler/fetcher.rb
@@ -92,6 +92,13 @@ module Bundler
         "Your network or your gem server is probably having issues right now."
     end
 
+    # return the specs in the bundler format as an index with retries
+    def specs_with_retry(gem_names, source)
+      Bundler::Retry.new("fetcher").attempts do
+        specs(gem_names, source)
+      end
+    end
+
     # return the specs in the bundler format as an index
     def specs(gem_names, source)
       old = Bundler.rubygems.sources
--- a/lib/bundler/source/rubygems.rb
+++ b/lib/bundler/source/rubygems.rb
@@ -337,7 +337,7 @@ module Bundler
           # gather lists from non-api sites
           index_fetchers.each do |f|
             Bundler.ui.info "Fetching source index from #{f.uri}"
-            idx.use f.specs(nil, self)
+            idx.use f.specs_with_retry(nil, self)
           end
 
           # because ensuring we have all the gems we need involves downloading
@@ -350,7 +350,7 @@ module Bundler
           if allow_api
             api_fetchers.each do |f|
               Bundler.ui.info "Fetching gem metadata from #{f.uri}", Bundler.ui.debug?
-              idx.use f.specs(dependency_names, self)
+              idx.use f.specs_with_retry(dependency_names, self)
               Bundler.ui.info "" unless Bundler.ui.debug? # new line now that the dots are over
             end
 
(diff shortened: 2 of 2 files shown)


segiddins:
Why not add the retry in the specs method?

jingweno:
@segiddins I could do that as well. But wrapping it will prob. break less tests :).

jingweno:
And of course separate the concerns.

jingweno:
Some of the tests are failing and I'm not sure whether it's expected: https://travis-ci.org/bundler/bundler/builds/90627361.

segiddins:
@jingweno yup, expected for now

jingweno:
@segiddins Looking forward to ship this. Let me know if there's anything I missed or I need to adjust.

jingweno:
Any update on this?

indirect:
I would prefer this it handle the case of no internet connection and stop requesting, rather than trying three times to connect to a server over a network that doesn't exist...

jingweno:
@indirect I believe we could catch the right exception [here](https://github.com/bundler/bundler/blob/master/lib/bundler/retry.rb#L39) to achieve it. Bundler already retries when downloading dependencies and git checkout.

jingweno:
BTW, I just met the same issue in deployment and the the deployment was skewed...we should really fix it...

schneems:
@indirect is there a good way to distinguish between the two cases? I think you run into the https://en.wikipedia.org/wiki/Two_Generals%27_Problem since you don't know why your packets aren't getting through. The easiest is to retry all the things. If you've got some implementation ideas @jingweno is pretty motivated. I can help work through a solution. Let us know when you get back home and get settled.

segiddins:
@schneems iirc the Retry class can be passed an array of Exceptions to retry on

schneems:
It looks like it is for filtering and re-raising exceptions. It doesn't stop the retry behavior https://github.com/bundler/bundler/blob/7be4f29c7caaccead7a6d9014839a8b3a63da315/lib/bundler/retry.rb#L35-L41.

segiddins:
The re-raise will stop retrying

schneems:
Ahh, got it. Thanks, looks like it was introduced in https://github.com/bundler/bundler/commit/39f64b962361316f6028d2bb60b06355054f0936. Is there already a specific exception to capture the network being down or name not resolving as Andre was asking for?

segiddins:
pretty sure it's just an ERRNO but I don't know for sure

indirect:
@homu r+

jingweno:
Nice! It's merged! Thanks all!

schneems:
It is also deployed in a pre-release https://github.com/bundler/bundler/blob/43addfdad722639dc26c58ec85369e0563793ec5/CHANGELOG.md#1110pre1-2015-11-29

When it goes stable, we can update the Ruby buildpack.

jingweno:
@schneems :+1:

PR #4105, opened 2015-11-12
[Lock] Add support for specify gems to `bundle lock` command

JuanitoFatas:
This Pull Request adds the ability to specify gems in bundle lock command via command-line option,
so that this is allowed:

```
bundle lock --gems devise
```

or

```
bundle lock --gems devise ominiauth
```

In this way, we can use lock command for specific gems without downloading any gem.

Thanks!

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -399,8 +399,8 @@ module Bundler
     end
 
     desc "lock", "Creates a lockfile without installing"
-    method_option "update", :type => :boolean, :default => false, :banner =>
-      "ignore the existing lockfile"
+    method_option "update", :type => :array, :lazy_default => [], :banner =>
+      "ignore the existing lockfile, update all gems by default, or update list of given gems"
     method_option "local", :type => :boolean, :default => false, :banner =>
       "do not attempt to fetch remote gemspecs and use the local gem cache only"
     method_option "print", :type => :boolean, :default => false, :banner =>
--- a/lib/bundler/cli/lock.rb
+++ b/lib/bundler/cli/lock.rb
@@ -16,8 +16,14 @@ module Bundler
       ui = Bundler.ui
       Bundler.ui = UI::Silent.new if print
 
-      unlock = options[:update]
-      definition = Bundler.definition(unlock)
+      gems = options[:update]
+
+      if gems && !gems.empty?
+        definition = Bundler.definition(:gems => gems)
+      else
+        definition = Bundler.definition(true)
+      end
+
       definition.resolve_remotely! unless options[:local]
 
       if print
--- a/spec/commands/lock_spec.rb
+++ b/spec/commands/lock_spec.rb
@@ -94,4 +94,12 @@ describe "bundle lock" do
     expect(read_lockfile "lock").to eq(@lockfile)
     expect { read_lockfile }.to raise_error(Errno::ENOENT)
   end
+
+  it "update specific gems using --update" do
+    lockfile @lockfile.gsub("2.3.2", "2.3.1").gsub("10.0.2", "10.0.1")
+
+    bundle "lock --update rails rake"
+
+    expect(read_lockfile).to eq(@lockfile)
+  end
 end


lib/bundler/cli/lock.rb
```diff
@@ -12,12 +12,30 @@ def run
         exit 1
       end
 
+      unless Bundler.default_lockfile.exist?
```
segiddins: This defeats the entire purpose of having the lock command?

segiddins:
I can't tell what the point of this PR is, could you try and explain in the PR message a summary of what this change is doing?

lib/bundler/cli/lock.rb
JuanitoFatas: Could you elaborate the purpose of lock command, thank you.

JuanitoFatas:
@segiddins Thanks for review, I have updated the Pull Request body.

segiddins:
This should be done via flags, not as raw arguments, and some of the changes will need to be reverted since they break the intended behavior of the existing command

JuanitoFatas:
I pass in gems via flags and revert most of changes. :bow:

segiddins:
:+1:

indirect:
If you lock only some gems, doesn't that mean that the lock might not work for all the gems? I am worried that this allows users to produce locks that conflict with the full Gemfile.

JuanitoFatas:
> If you lock only some gems, doesn't that mean that the lock might not work for all the gems? I am worried that this allows users to produce locks that conflict with the full Gemfile.

Hi @indirect, thanks for your concern, I have tried two commands against a project with Gemfile and Gemfile.lock: 

- `bundle lock --gems gem1 gem2`
- `bundle update gem1 gem2`

The [resulting `Gemfile.lock` from these two commands is the same](https://gist.github.com/JuanitoFatas/608fe8cd13601fdcea03).

Is this correct? Thank you.

indirect:
@JuanitoFatas I see. Is the goal to be able to _update_ gems without downloading them? If that is the case, I do not think `bundle lock --gem` is a good name.

JuanitoFatas:
> Is the goal to be able to update gems without downloading them?

Yes! To get a updated lockfile without downloading them.

>  If that is the case, I do not think bundle lock --gem is a good name for the command.

Do you think this `bundle lock` command should rename to `bundle resolve`? Resolve and update lockfile. This name is not confused with the [`bundle lock` command from Bundler version 0.9](http://bundler.io/v0.9/bundle_lock.html).

Thanks!

segiddins:
No, we can't rename the lock command until 2.0 anyways. I think Andre is just asking for a better option name.

JuanitoFatas:
> No, we can't rename the lock command until 2.0 anyways. I think Andre is just asking for a better option name.

Thank you for the heads up.

I list some alternatives:

0. `bundle lock --resolve`
1. `bundle lock --resolves`
2. `bundle lock --resolves-for`
3. `bundle lock --resolve-gems`
4. `bundle lock --update-gems`
5. `bundle lock --updates-for`

I vote for number 3, `bundle lock --resolve-gems` (can have a shorter alias).

What do you think?

spec/commands/lock_spec.rb
```diff
@@ -94,4 +94,13 @@ def read_lockfile(file = "Gemfile.lock")
     expect(read_lockfile "lock").to eq(@lockfile)
     expect { read_lockfile }.to raise_error(Errno::ENOENT)
   end
+
+  it "update specific gems using --gems" do
+    lockfile @lockfile.gsub("2.3.2", "2.3.1")
+    lockfile @lockfile.gsub("10.0.2", "10.0.1")
```
segiddins: the two `gsub`s need to be chained, since `@lockfile` isn't being updated here

lib/bundler/cli.rb
```diff
@@ -407,6 +407,8 @@ def inject(name, version, *gems)
       "print the lockfile to STDOUT instead of writing to the file system"
     method_option "lockfile", :type => :string, :default => nil, :banner =>
       "the path the lockfile should be written to"
+    method_option "gems", :type => :array, :default => [], :banner =>
```
segiddins: what if we just turned the `--update` option into taking an array, and pass in `true` if that array is empty?

indirect:
I think `--update` makes sense, and it can work like the update command, taking a list of gems or updating all gems if no list is given.

spec/commands/lock_spec.rb
JuanitoFatas: Thanks for the catch! :+1:

lib/bundler/cli.rb
JuanitoFatas: Nice! I've made this change.

segiddins:
👍🏻

JuanitoFatas:
@indirect Since #4108 merged, should I update the manual here or as a separate Pull Request? Thanks!

indirect:
@homu r+

Issue #3609, opened 2015-05-01
Bundler should NOT include private credentials in Gemfile.lock from source urls in Gemfile


joelpresence:
Hi,

I use GemFury for private gems.  In my `Gemfile` I have code like:
```
# In Gemfile
source "https://#{ENV['FURY_AUTH']}@gem.fury.io/devpresence/" do
  gem 'presence_common_web', '~> 0.1.4'
end
```
The environment variable `FURY_AUTH` has value `_MY_SECRET_FURY_TOKEN_`.  When I run `bundle install` or `bundle update`, bundler updates `Gemfile.lock` to include:
```
# In Gemfile.lock
remote: https://_MY_SECRET_FURY_TOKEN_@gem.fury.io/devpresence/
```
This is a risk because now `_MY_SECRET_FURY_TOKEN_` risks getting committed to source control.  I took great care to *not* list my secret token in plain text in `Gemfile` (e.g. by specifying it as an env var) so bundler/gems should not now just write the secret token to the `Gemfile.lock`.

When I deploy this, say to Heroku, I further specify an env var as in:
```
bash> heroku config:set FURY_AUTH=_MY_SECRET_FURY_TOKEN_
```
so even Heroku deploys don't contain my secret token (except in private runtime config).

Further, when I run tests, bundler seems to periodically update `Gemfile.lock`.  So even if I manually edit `Gemfile.lock` to remove `_MY_SECRET_FURY_TOKEN_`, bundler seems to periodically re-add it, without me even running bundler manually (maybe rspec is doing this?).

I would propose that if a source url in the Gemfile contains valid HTTP credentials, and the bundler-created `Gemfile.lock` remote url contains credentials as in:
```
remote: https://token@host/path
remote: https://user:password@host/path
```
then bundler should automatically strip these when it's creating the `Gemfile.lock` from `Gemfile`.

So
```
# CASE 1: CURRENT BEHAVIOR
# In Gemfile
source "https://#{ENV['FURY_AUTH']}@gem.fury.io/devpresence/" do
  gem 'presence_common_web', '~> 0.1.4'
end

# Becomes in Gemfile.lock
remote: https://_MY_SECRET_FURY_TOKEN_@gem.fury.io/devpresence/

# CASE 2: DESIRED BEHAVIOR
# In Gemfile
source "https://#{ENV['FURY_AUTH']}@gem.fury.io/devpresence/" do
  gem 'presence_common_web', '~> 0.1.4'
end

[...]

PR #4107, opened 2015-11-13
When passing https:// credentials into git, don't write them to the lockfile

denmat:
Context:
Currently bundler writes sensitive credentials in the lockfile when passed in as environment variables. 

`Gemfile`
```
gem foo,  git: "https://#{ENV[GITHUB_TOKEN]}@github.com/foo/foo.git"
```

`Gemfile.lock`
```
 GIT
-  remote: git@github.com:foo/foo.git
+  remote: https://<sensitive_token>:x-oauth-basic@github.com/foo/foo.git
   revision: b668ddd4661836eb31ec780160048d2d41f4393b
   branch: master
   specs:
```

This should never happen.

Rather than using `bundle config` which stores clear text sensitive information, I would rather be able to pass in gpg encrypted credentials file as ENV values.

Change:
This patch sanitises those credentials from the lockfile at the time the lockfile is written.

Relates to: https://github.com/bundler/bundler/issues/3609

### DIFF
--- a/spec/bundler/source/git_spec.rb
+++ b/spec/bundler/source/git_spec.rb
@@ -1,0 +1,26 @@
+require 'spec_helper'
+
+describe Bundler::Source::Git do
+
+  describe "#to_lock" do
+    let(:git_proxy) { double(:git_proxy, revision: "ABC123") }
+
+    it "removes credentials from https uri" do
+      expect(Bundler::Source::Git::GitProxy).to receive(:new).exactly(2).times.and_return(git_proxy)
+      expect(Bundler).to receive(:requires_sudo?).exactly(2).times.and_return(false)
+      expect(Bundler).to receive(:cache).exactly(2).times.and_return(Pathname.new("Idontcare"))
+      {
+        "https://u:p@github.com/foo/foo.git" => "https://github.com/foo/foo.git",
+        "http://u:p@github.com/foo/foo.git" => "http://github.com/foo/foo.git"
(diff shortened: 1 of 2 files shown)


spec/bundler/source/git_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+        "http://u:p@github.com/foo/foo.git" => "http://github.com/foo/foo.git"
+      }.each do |uri_in, uri_out|
+        g = described_class.new({"revision" => "ABC123", "uri" => uri_in})
+        expect(g.to_lock).to eq(<<-STR)
+GIT
+  remote: #{uri_out}
+  revision: ABC123
+  specs:
+STR
+      end
```
nemski: Should this be two different tests instead of http and https in one test?

I think it would like a little cleaner this way.

nemski:
:+1:

indirect:
Bundle config and environment variables are 100% interchangeable. Set an env var with the same name as the key in `.bundle/config` and delete the config file.

indirect:
The `master` branch already contains support for setting git usernames and passwords via both ENV and `bundle config`: https://github.com/bundler/bundler/pull/3898. This will be released very shortly as version 1.11. Please use these supported mechanisms that do not save credentials in the lockfile instead of interpolating your username and password into the source URL.

denmat:
Sorry for the late, reply. Looking forward to the next release then. Thanks.

Issue #2859, opened 2014-02-06
make the :github shorthand produce https urls


grosser:
atm github shorthand produces `git://github.com/owner/repo.git`,
but would be ssl/nice if it produces `https://github.com/owner/repo.git`

PR #4109, opened 2015-11-16
Use HTTPS protocol for :github shortcut

sikachu:
This improves security of cloning a Git repository using `:github` directive, as the clone process would go through a secured HTTPS protocol instead of insecure `git://` protocol.

Note that GitHub already starting to retire usage of `git://` protocol as you can notice by the clone URL on every project page. Also, given both `:gist` and `:bitbucket` already uses HTTPS protocol, I don't think
this will follow the good convention of choosing the best protocol for the users.

### DIFF
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -152,7 +152,7 @@ module Bundler
               "either use the :git option on a gem, or specify the gems that \n" \
               "bundler should find in the git source by passing a block to \n" \
               "the git method, like: \n\n" \
-              "  git 'git://github.com/rails/rails.git' do\n" \
+              "  git 'https://github.com/rails/rails.git' do\n" \
               "    gem 'rails'\n" \
               "  end"
         raise DeprecatedError, msg
@@ -220,7 +220,7 @@ module Bundler
     def add_git_sources
       git_source(:github) do |repo_name|
         repo_name = "#{repo_name}/#{repo_name}" unless repo_name.include?("/")
-        "git://github.com/#{repo_name}.git"
+        "https://github.com/#{repo_name}.git"
       end
 
       git_source(:gist) {|repo_name| "https://gist.github.com/#{repo_name}.git" }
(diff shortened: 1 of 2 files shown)


seanpdoyle:
@sikachu does this change have any effect on how bundler would consume private repositories?

sikachu:
@seanpdoyle It should not. I don't think you were able to use `:github` with private repos anyway, so this actually will make you _able_ to use `:github` with it. ;)

seanpdoyle:
> so this actually will make you able to use :github with it. :wink:

:clap:

segiddins:
This is already done on the 2.0 branch -- we can't ship it before then because it's a backwards incompatible change.

dblandin:
@segiddins I'm unclear on why this would be a backwards incompatible change. Would you mind explaining that a bit further?

Are there scenarios in which Bundler 1.x would fail to fetch a gem's source via secure `git@github.com:tenderlove/nokogiri.git` or `https://github.com/tenderlove/nokogiri.git ` URL formats while a gem's source is fetched successfully via the insecure `git://github.com:tenderlove/nokogiri.git` URL format?

Thanks!

indirect:
@dblandin it is backwards incompatible because we actually shipped this change in a previous version of Bundler and many people reported that their Gemfiles no longer worked.

You can change how the Github shortcut works in your own Gemfiles using the `git_source` method, so I would suggest you do that until Bundler 2.0.

sikachu:
Excuse me for self-promotion-ish, but I wrote [a blog post](https://sikac.hu/secure-your-github-sources-in-your-gemfile-f99784015200) on how to set that `git_source` to use secure `https` URL for `:github` source. I hope that's helpful for everyone before Bundler 2.0 comes out.

dblandin:
>  it is backwards incompatible because we actually shipped this change in a previous version of Bundler and many people reported that their Gemfiles no longer worked.

@indirect Gotcha. Thanks for the context! Found the initial [PR](https://github.com/bundler/bundler/pull/2569) and some [related](https://github.com/bundler/bundler/issues/2859) [issues](https://github.com/bundler/bundler/pull/2902).

Thanks for your work on Bundler and fielding the same question over and over again! :persevere:

PR #4110, opened 2015-11-16
Fix bundle console IRB fallback

felixbuenemann:
The fallback to IRB for `bundle console` is currently broken, due to a missing `require 'irb'`:
```bash
bundle config console pry
bundle console
Couldn't load console pry
Could not find constant IRB
```

This fixes the fallback and also alters the error message to indicate the fallback to irb.

While investigating the fix I noticed that the `bundle console` specs were broken and only worked because the expected strings were both contained in the error message.

In order to test an alternate console and the fallback behavior, I added a fake pry gem that implements the minimum required api. I also replaced `__callee__` with `__method__` for ruby 1.8.7 compatibility.

### DIFF
--- a/lib/bundler/cli/console.rb
+++ b/lib/bundler/cli/console.rb
@@ -18,7 +18,8 @@ module Bundler
       require name
       get_constant(name)
     rescue LoadError
-      Bundler.ui.error "Couldn't load console #{name}"
+      Bundler.ui.error "Couldn't load console #{name}, falling back to irb"
+      require "irb"
       get_constant("irb")
     end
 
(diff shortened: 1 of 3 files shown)


segiddins:
:+1:

segiddins:
@homu r+

felixbuenemann:
If there will be another 1.10.x release it would be great if this could be backported to 1-10-stable.

indirect:
@felixbuenemann 1.11 pre is out now. :)

felixbuenemann:
@indirect Thanks, I'll try it out.

PR #4111, opened 2015-11-17
Generate HTTPS URIs for github-targetted gems

dwb:
The [bundler-audit gem rightly raises an error](https://github.com/rubysec/bundler-audit/blob/96cc110974e2c26702fa99e22bc430a425696f4e/lib/bundler/audit/scanner.rb#L77-L116) when a Gemfile uses a `git://` or `http://` source, these being insecure protocols.

This patch sets an `https://` URI in the lockfile when you do something like
```
gem 'environmentor', github: 'madebymany/environmentor'
```


indirect:
Duplicate of #4109.

dwb:
@indirect ah sorry. I thought I searched well enough; clearly not.

PR #4113, opened 2015-11-17
Check silence_root_warning option to skip warning on root install cmds

blackxored:
Adds a new setting `silence_root_warning` (environment variable: `BUNDLE_SILENCE_ROOT_WARNING`) that skips the default warning when `install` is run as root.

The warning is useful but I think it should be able to be disabled if the user knows what he's doing, it's handy when run inside Docker containers (which by default run as root), or other environments where root is used.

Also adds a new `preserve_env` option to `bundle` helper in tests which allows running commands under sudo with the `-E` flag (preserve environment) as opposed to the default env_reset. Without this, the command would be run without our flag which makes it impossible to test. There might be cleaner ways to achieve this, though.

### DIFF
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -8,7 +8,7 @@ module Bundler
     def run
       Bundler.ui.level = "error" if options[:quiet]
 
-      warn_if_root unless ENV['SILENCE_ROOT_WARNING']
+      warn_if_root
 
       [:with, :without].each do |option|
         if options[option]
@@ -157,7 +157,7 @@ module Bundler
   private
 
     def warn_if_root
-      return if Bundler::WINDOWS || !Process.uid.zero?
+      return if Bundler.settings[:silence_root_warning] || Bundler::WINDOWS || !Process.uid.zero?
       Bundler.ui.warn "Don't run Bundler as root. Bundler can ask for sudo " \
         "if it is needed, and installing your bundle as root will break this " \
         "application for all non-root users on this machine.", :wrap => true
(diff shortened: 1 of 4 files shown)


segiddins:
https://travis-ci.org/bundler/bundler/jobs/91675727#L257-L282 failures

lib/bundler/cli/install.rb
```diff
@@ -8,7 +8,7 @@ def initialize(options)
     def run
       Bundler.ui.level = "error" if options[:quiet]
 
-      warn_if_root
+      warn_if_root unless ENV['SILENCE_ROOT_WARNING']
```
segiddins: this should be done via the bundler config

blackxored:
@segiddins I'll fix the style issues later this afternoon. As for the bundler config, I'm not really sure it belongs there as I don't want to avoid the warnings everywhere (i.e. make the setting project-wide) but only on specific environments. Let me hear your thoughts about it.

segiddins:
@blackxored it does belong in settings, since that's how bundler handles checking for settings.

blackxored:
I've fixed the style issues, updated to use `Bundler.settings` and the PR title and description. Let me know.

segiddins:
Looks good to me! r+, @indirect ?

segiddins:
@homu r+

envygeeks:
This is awesome, thanks so much!!!!!!

blackxored:
Thanks! I'm now trying to make it the default in the Docker Ruby images, not sure how that'd go.

PR #4116, opened 2015-11-21
[CompactIndex] Handle yanked versions

segiddins:
@indirect @fotanus look good?

### DIFF
--- a/lib/bundler/vendor/compact_index_client/lib/compact_index_client/cache.rb
+++ b/lib/bundler/vendor/compact_index_client/lib/compact_index_client/cache.rb
@@ -22,10 +22,14 @@ class Bundler::CompactIndexClient
       lines(versions_path).each do |line|
         name, versions_string, info_checksum = line.split(" ", 3)
         info_checksums_by_name[name] = info_checksum || ""
-        versions = versions_string.split(",").map! do |version|
-          version.split("-", 2).unshift(name)
+        versions_string.split(",").each do |version|
+          if version.start_with?("-")
+            versions_by_name[name].delete(version[1..-1])
+          else
+            versions = version.split("-", 2).unshift(name)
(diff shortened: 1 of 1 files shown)


indirect:
This moves concat inside the loop, increasing calls from 1 to N... since N is huge, maybe this is a significant performance difference? But I haven't benchmarked yet.

segiddins:
Ah it's also wrong... Needs to be append instead of concat

lib/bundler/vendor/compact_index_client/lib/compact_index_client/cache.rb
```diff
@@ -22,10 +22,14 @@ def versions
...
         name, versions_string, info_checksum = line.split(" ", 3)
         info_checksums_by_name[name] = info_checksum || ""
-        versions = versions_string.split(",").map! do |version|
-          version.split("-", 2).unshift(name)
+        versions_string.split(",").each do |version|
+          if version.start_with?("-")
+            versions_by_name[name].delete(version[1..-1])
+          else
+            versions = version.split("-", 2).unshift(name)
+            versions_by_name[name].concat(versions)
```
fotanus: Not 100% sure, but It seems that `versions` is something like `["rails", "1.0.0", "ruby"]`, and this will be appended to `versions_by_name`. The thing is that I expect `versions_by_name` take a name and return versions, which is not the case. Do we really need the `unshift` if we already have the nameas the hash key?

Also, not sure if makes sense on this context, but you might want to have some structure for this stuff if gets too complex.

segiddins:
Fixed

segiddins:
@fotanus please review?

fotanus:
puuuuuurfect :smile_cat:

segiddins:
:shipit:

PR #4124, opened 2015-11-26
More rubocop_todo cleanup

segiddins:


simi:
I see no benefit and git mess only. :(

indirect:
@simi Bundler was written for years and years without any consistent code style, and now it is very hard for contributors to know if their pull requests will be accepted or rejected based on style. By making RuboCop enforce style, contributors can easily know in advance if their style matches the project style. It's a big improvement.

indirect:
@homu r+

simi:
Sure I understand. I just prefer "increment style" of those changes = change style while you're changing the code to keep history. It will be worth it to notice the bundler ruby style in https://github.com/bundler/bundler/blob/master/DEVELOPMENT.md at least.

segiddins:
@simi we already do, since running `rake spec` will also run rubocop.

PR #4130, opened 2015-11-30
[GemHelper] Fix crash when pushing gem

segiddins:
\c @indirect


segiddins:
Failures look unrelated

segiddins:
@homu r+

segiddins:
We really need to write better specs for the gem helper, though.

PR #4139, opened 2015-12-02
Use travis container-based infrastructure

benlovell:
Jobs running on container-based infrastructure start up faster and permit the
use of caches for public repositories. As a default it makes sense to enable
this setting for generated gems.

### DIFF
--- a/lib/bundler/templates/newgem/.travis.yml.tt
+++ b/lib/bundler/templates/newgem/.travis.yml.tt
@@ -1,3 +1,4 @@
+sudo: false
 language: ruby
 rvm:
   - <%= RUBY_VERSION %>


segiddins:
@indirect thoughts on this?

indirect:
This actually seems like the right default, but a magic and unrelated-looking comment makes me uneasy.

segiddins:
¯\_(ツ)_/¯ that's how travis does it

BanzaiMan:
This is worth exploring, but:

* Note that `sudo: false` by itself does not enable caching.
* Given a large build matrix of Bundler, the savings in job startup time is probably negligible, and it increases the probability that cache may become corrupt, and yields false positive (this is a shortcoming on the Travis CI side).

indirect:
@BanzaiMan this PR is for the travis config that ships with new gems generated by `bundle gem`. The Bundler build sadly cannot run on the new infrastructure, because it needs sudo. Also, Bundler doesn't really have anything to cache, so we don't use the cache either.

BanzaiMan:
@indirect Ah! Thanks for the explanation. Carry on!

segiddins:
@indirect please merge with homu

indirect:
@segiddins dammit sorry

PR #4141, opened 2015-12-04
give a better error message when a remote ref no longer exists

### LINKED ISSUE #4140: ambiguous error message when refs are missing from a git dependency
This is related to, but not quite the same as, #3897.

When you run `bundle install` in a project that has a dependency on a git ref, and that ref no longer exists, you get an error message like
```
Git error: command `git reset --hard 0d4f9b7dcc50fe5235abd5035a49c7075ae514d6` in directory /home/doy/tmp/bundler-test/vendor/bundle/ruby/2.1.0/bundler/gems/bundler-test-0d4f9b7dcc50 has
failed.
If this error persists you could try removing the cache directory '/home/doy/tmp/bundler-test/vendor/bundle/ruby/2.1.0/cache/bundler/git/bundler-test-a59c53842b1d2c30f68c5498ce0cca55e39b6e20'
```

This is pretty unhelpful, since removing the cache directory isn't actually going to fix anything in this case. A much more helpful response here would be for bundler to automatically remove the cache directory and try again, and if it fails the second time, give an error message indicating that the remote ref actually doesn't exist.

doy:
if we can't find the given ref after fetching, try clearing the cache
ourselves and trying again. if we still can't find it, we know that it
actually just doesn't exist, so tell the user that.

Fixes #4140.

### DIFF
--- a/lib/bundler/source/git/git_proxy.rb
+++ b/lib/bundler/source/git/git_proxy.rb
@@ -27,6 +27,13 @@ module Bundler
         end
       end
 
+      class GitRevisionError < GitError
+        def initialize(ref, repo)
+          msg = "Revision #{ref} does not exist in the repository #{repo}. Maybe you misspelled it?"
+          super msg
+        end
+      end
+
       # The GitProxy is responsible to interact with git repositories.
       # All actions required by the Git source is encapsulated in this
       # object.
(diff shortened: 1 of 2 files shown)


lib/bundler/source/git/git_proxy.rb
```diff
@@ -27,6 +27,13 @@ def initialize(command, path = nil, extra_info = nil)
         end
       end
 
+      class GitRevisionError < GitError
```
segiddins: `MissingGitRevisionError`

segiddins:
Sweet! Think you could throw in a few tests for this?

doy:
How does this look? I wanted to add some tests for an actual corrupted git repo cache, but I couldn't figure out how the caching stuff works in tests. This covers the major functionality though.

segiddins:
Looks good! I'd also like to see a test for removing the cache and retrying when the first attempt to find the revision fails

doy:
I did some more digging, and I can't actually find a situation where the first attempt would fail but the second attempt would succeed. Everything I tried led to either the cache being corrupted enough that it failed at a much earlier stage, or it worked fine the first time. Would it be reasonable to remove the "remove the cache and retry" logic? I'm not entirely convinced that I didn't miss anything (I'm not familiar with this codebase at all), but leaving it in "just in case" without being able to test it also doesn't sound ideal.

indirect:
I'm okay removing the remove cache option, and we can see how it goes.

doy:
Done.

segiddins:
@homu r+

indirect:
@homu retry

doy:
Anything else I need to do here? It doesn't look like this failure is related to my branch.

segiddins:
@homu r+

segiddins:
@homu retry

PR #4142, opened 2015-12-07
Modified newgem generator to exclude Gemfile from the gemspec spec.files.

jperville:
This PR modifies the `bundle gem` generator to exclude Gemfile from the gemspec spec.files (the list of files that will be copied in the generated gem).

The reason is that:
 - the `Gemfile` file is not used by rubygems (as far as I know)
 - the `Gemfile` file may contain private sources (possibly with credentials) that we do not want to leak into a packaged gem.

### DIFF
--- a/lib/bundler/templates/newgem/newgem.gemspec.tt
+++ b/lib/bundler/templates/newgem/newgem.gemspec.tt
@@ -24,7 +24,7 @@ Gem::Specification.new do |spec|
     raise "RubyGems 2.0 or newer is required to protect against public gem pushes."
   end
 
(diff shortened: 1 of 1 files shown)


indirect:
Please use `bundle config` to store credentials instead of the Gemfile.

That said, if your Gemfile contains private sources, why are you pushing this gem in public?

jperville:
Hi @indirect, we are not publishing these gems publicly but we are considering giving them in *.gem form to a few selected integrators (without giving them access to our gem server). That's why we would like to avoid passing credentials by accident if the Gemfile is bundled into the *.gem file.

For now, we will explore passing the credentials with "bundle config" instead, it is good practice.

PR #4143, opened 2015-12-09
[WIP] Better support of Ruby 2.3 with enabled frozen string literal

deepj:
This commit prevents several bugs with `RuntimeError: can't modify frozen String` with enabled string frozen literal in Ruby 2.3

### DIFF
--- a/lib/bundler/lazy_specification.rb
+++ b/lib/bundler/lazy_specification.rb
@@ -35,10 +35,12 @@ module Bundler
     end
 
     def to_lock
+      out = String.new
+
       if platform == Gem::Platform::RUBY || platform.nil?
-        out = "    #{name} (#{version})\n"
+        out << "    #{name} (#{version})\n"
       else
-        out = "    #{name} (#{version}-#{platform})\n"
+        out << "    #{name} (#{version}-#{platform})\n"
       end
 
       dependencies.sort_by(&:to_s).uniq.each do |dep|
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -63,6 +63,10 @@ Style/TrailingComma:
 Style/TrailingUnderscoreVariable:
   Enabled: false
 
+# `String.new` is preferred style with enabled frozen string literal
+Style/EmptyLiteral:
+  Enabled: false
+
 # 1.8.7 support
 
 Style/HashSyntax:
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,6 +91,11 @@ matrix:
     # Ruby-head (we want to know how we're doing, but not fail the build)
     - rvm: ruby-head
       env: RGV=master
+    # Ruby-head with enabled frozen string literal
+    - rvm: ruby-head
+      env:
+        - RGV=master
+        - RUBYOPT="--enable-frozen-string-literal --debug=frozen-string-literal"
   allow_failures:
     - rvm: 1.8.7
       env: RGV=v2.1.11
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -53,7 +53,7 @@ module Bundler
       @specs           = nil
       @ruby_version    = ruby_version
 
-      @lockfile_contents      = ""
+      @lockfile_contents      = String.new
       @locked_bundler_version = nil
 
       if lockfile && File.exist?(lockfile)
@@ -280,7 +280,7 @@ module Bundler
     end
 
     def to_lock
-      out = ""
+      out = String.new
 
       sources.lock_sources.each do |source|
         # Add the source header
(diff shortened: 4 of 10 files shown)


segiddins:
This shouldn't be necessary as none of the bundler files have the frozen string literal magic comment.

deepj:
Of course, but for example if I try run RSpec dev tests with enabled string frozen literal (I want to use RSpec with a project with Ruby 2.3+ and string frozen literal). Without these changes in Bundler I can't move further. These changes don't bring any backward incompatibilities. So why not?

segiddins:
In that case, we need to test that bundler supports being run with frozen string literals enabled

deepj:
OK. I take a look at Travis settings. And is it okay to disable `Style/EmptyLiteral` in Rubocop (https://travis-ci.org/bundler/bundler/jobs/95840769#L256)? Alternative is to use `''.dup`. But it seems `String.new` is preferred style. See: https://bugs.ruby-lang.org/issues/11759#note-12

segiddins:
Yes, I definitely prefer `String.new`

segiddins:
Hm, I can't even get the RubyGems specs to _run_ with frozen string literals enabled (.rubies/2.3.0-preview1/lib/ruby/2.3.0/x86_64-darwin15/rbconfig.rb:247:in `replace': can't modify frozen String (RuntimeError)). Given Ruby core itself isn't working... I don't know how much it's worth getting bundler to work with it enabled globally yet?

deepj:
I forgot to mention, `ruby-head` needs some additional patches due to this. I have some fixes on my side. See: https://github.com/ruby/ruby/pulls (https://github.com/ruby/ruby/pull/1116)

Let's make this PR like WIP

segiddins:
I think it also needs https://github.com/ruby/ruby/pull/1136

segiddins:
`can't modify frozen String, created at /home/travis/.rvm/rubies/ruby-head/lib/ruby/site_ruby/2.3.0/rubygems/dependency.rb:309` :cry:

lib/bundler/lazy_specification.rb
```diff
@@ -41,6 +41,8 @@ def to_lock
         out = "    #{name} (#{version}-#{platform})\n"
       end
 
+      out = out.dup
```
bquorning: Could we set an `out = String.new` variable at the beginning of the method instead of calling `dup` here?

bquorning:
Please, also fix these:

 - https://github.com/bundler/bundler/blob/master/lib/bundler/env.rb#L14
 - https://github.com/bundler/bundler/blob/master/lib/bundler/env.rb#L65

mwpastore:
It's worth mentioning that Ruby 3 will have frozen string literals enabled by default—or at least that's the plan—so any effort put into making Bundler compatible with this pragma now will hopefully not go to waste. I for one am hoping to run my full stack under Ruby 2.3 with `RUBYOPT="--enable=frozen-string-literal"` once we're all past the initial shock of incompatibility.

deepj:
@mwpastore Unfortunately, Ruby core/stdlib isn't prepared for that. Also it seems core developers aren't interesting in that to fix it :( /cc @hsbt

lib/bundler/lazy_specification.rb
deepj: Fixed

deepj:
@bquorning This is fixed as well

mwpastore:
@deepj Agreed, the state of the preview is quite dismal.

deepj:
@mwpastore It seems they solved it by using adding `frozen_string_literal: false ` to everywhere. More here: https://github.com/ruby/ruby/commit/3e92b635fb5422207b7bbdc924e292e51e21f040

deepj:
@segiddins: Shouldn't I add `frozen_string_literal: true` to everywhere in bundler and not force frozen string literals globally like now? This way is more viable than affecting other dependencies. You made same thing in rubygems (https://github.com/rubygems/rubygems/pull/1408)

segiddins:
@deepj I really dislike encoding comments so id rather not, if possible

gregd:
I've installed Ruby 2.3.0, set --enable-frozen-string-literal cmd line switch and run my Rails 4.2 app. It stops on `rubygems.rb:65:in block in to_lock: can't modify frozen String` It would be good have the pull request merged in next Bundler release.

indirect:
@gregd if it's important to you, you're welcome to fix the tests on Ruby 2.3, and then update this pull request.

technion:
Hi, Just wondering if there were any remaining issues for the current version of this PR? I'm working on running with frozen strings and, at first glance, every issue outstanding is already covered by this patch.

gregd:
@technion Try to execute tests. This PR fixes just few places where frozen strings are modified.

indirect:
@homu r+

Issue #4094, opened 2015-11-05
--no-coc and --no-mit options does not work


bolshakov:
I'm running the following command:

```
> bundle --version
Bundler version 1.10.6
> bundle gem my_gem --no-coc --no-mit
```

But it ignores my options and generate MIT license CODE_OF_CONDUCT file

segiddins:
I seem to recall us fixing that at some point? Can you check if the issue still exists on master?

rgb-one:
I just tested it on master. When I run it the options `--no-coc`, `--no-mit` seem to be ignored as I am still prompted as to whether or not I would like to add a Code of Conduct and an MIT licence file. I would think that those two prompts would be bypassed and only the testing prompt would show but that's not the case; I get all three prompts.

Here is the input/output:
~~~
:~$ dbundle -v
Bundler version 1.11.0.pre.2
:~$ bundle gem my_gem --no-coc --no-mit
Creating gem 'my_gem'...
Do you want to include a code of conduct in gems you generate?
Codes of conduct can increase contributions to your project by contributors who prefer collaborative, safe spaces. You can read more about the code of conduct at contributor-covenant.org. Having a code of conduct means agreeing to the responsibility of enforcing it, so be sure that you are prepared to do that. For suggestions about how to enforce codes of conduct, see bit.ly/coc-enforcement. y/(n): n
Do you want to license your code permissively under the MIT license?
This means that any other developer or company will be legally allowed to use your code for free as long as they admit you created it. You can read more about the MIT license at choosealicense.com/licenses/mit. y/(n): n
Do you want to generate tests with your gem?
Type 'rspec' or 'minitest' to generate those test files now and in the future. rspec/minitest/(none): none
      create  my_gem/Gemfile
      create  my_gem/.gitignore
      create  my_gem/lib/my_gem.rb
      create  my_gem/lib/my_gem/version.rb
      create  my_gem/my_gem.gemspec
      create  my_gem/Rakefile
      create  my_gem/README.md
      create  my_gem/bin/console
      create  my_gem/bin/setup
Initializing git repo in /home/vagrant/my_gem
:~$ ls my_gem/
bin  Gemfile  lib  my_gem.gemspec  Rakefile  README.md
~~~

rgb-one:
Working on master.

RochesterinNYC:
@rgb-one Is this reopened because it's not working on master/the latest release?

rgb-one:
[...]

PR #4148, opened 2015-12-13
Fix --no-coc and --no-mit flags

RochesterinNYC:
- Behavior for the `gem` command should be that utilizing these flags (or
  the `--coc` and `--mit` flags) overrides the defaults you have set in your
  bundler config settings for that one-off execution of the gem command
- Added test coverage for testing `--coc`, `--mit`, `--no-coc`, `--no-mit` flags
  against bundle config settings
- Refactored and DRYed up `newgem_spec`

Addresses #4094

### DIFF
--- a/lib/bundler/cli/gem.rb
+++ b/lib/bundler/cli/gem.rb
@@ -148,7 +148,8 @@ module Bundler
     end
 
     def ask_and_set(key, header, message)
-      choice = options[key] || Bundler.settings["gem.#{key}"]
+      choice = options[key]
+      choice = Bundler.settings["gem.#{key}"] if choice.nil?
 
       if choice.nil?
         Bundler.ui.confirm header
--- a/spec/commands/newgem_spec.rb
+++ b/spec/commands/newgem_spec.rb
@@ -15,6 +15,21 @@ describe "bundle gem" do
     end
   end
 
+  def execute_bundle_gem(gem_name, flag = "", to_remove_push_guard = true)
+    bundle "gem #{gem_name} #{flag}"
+    remove_push_guard(gem_name) if to_remove_push_guard
+    # reset gemspec cache for each test because of commit 3d4163a
+    Bundler.clear_gemspec_cache
+  end
+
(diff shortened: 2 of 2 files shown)


indirect:
Ahh, great, thanks!

RochesterinNYC:
@indirect Hm, so regarding the broken Travis build, I saw this kind of stalled/frozen behavior locally while experimenting with different ways to call and use RSpec shared_examples. I'm guessing some of those build jobs (the ones using Ruby `1.8.7`) are using older versions of RSpec and might not have support for the shared_example behavior I'm using in the spec. How would you recommend I proceed on this?

I could drop the use of RSpec shared examples but the coverage specs I added would be relatively lengthy and repetitive without them. I could also look more into the legacy RSpec those Ruby `1.8.7` builds are using? How do I find out which version of RSpec those are using? Thanks for your help!

indirect:
@RochesterinNYC we're using the same version of RSpec in all builds, as far as I know, installed with `gem install rspec -v "~> 3.0"`. I'm perfectly happy with your tests only being called on only one of the gem names.

RochesterinNYC:
@indirect Fixed the integration on the Ruby `1.8.7` builds without having to remove any test coverage. It seems the error was with Ruby `1.8.7` and setting only partial settings of the global config instead of all three main ones?

Setting all three like so in the before blocks:
```
global_config "BUNDLE_GEM__MIT" => "true", "BUNDLE_GEM__TEST" => "false", "BUNDLE_GEM__COC" => "false"
```
Rather than:
```
global_config "BUNDLE_GEM__MIT" => "true""
```

fixed it in the Ruby 1.8.7 builds. Can't say I know exactly why this was happening though.

indirect:
Awesome, thanks!

indirect:
Cherry-picked to the 1-11-stable branch as 6c18c55406e386c6b9d591b589360face9211096. This will be released in 1.11.1.

RochesterinNYC:
Great, thanks!

Issue #4147, opened 2015-12-13
LazySpecification has not been materialized yet (calling :required_ruby_version [])


lukemelia:
- What did you do?

  I updated bundler to 1.11.0 and got an error running `bundle install`: `Your Gemfile.lock is corrupt. The following gem is missing from the DEPENDENCIES section: 'ruby_gntp'`. I then ran the command `/Users/lmelia/.rvm/gems/ruby-2.1.6@yapp/bin/bundle update terminal-notifier-guard`

- What did you expect to happen?

  I hoped this was related to the error and updating the dependency would fix it.

- What happened instead?

  Instead, what actually happened was that I received the error below.

Error details

    RuntimeError: LazySpecification has not been materialized yet (calling :required_ruby_version [])
      /Users/lmelia/.rvm/gems/ruby-2.1.6@yapp/gems/bundler-1.11.0/lib/bundler/lazy_specification.rb:75:in `method_missing'
      /Users/lmelia/.rvm/gems/ruby-2.1.6@yapp/gems/bundler-1.11.0/lib/bundler/resolver.rb:132:in `for?'
      /Users/lmelia/.rvm/gems/ruby-2.1.6@yapp/gems/bundler-1.11.0/lib/bundler/resolver.rb:275:in `block in search_for'
      /Users/lmelia/.rvm/gems/ruby-2.1.6@yapp/gems/bundler-1.11.0/lib/bundler/resolver.rb:275:in `select'
      /Users/lmelia/.rvm/gems/ruby-2.1.6@yapp/gems/bundler-1.11.0/lib/bundler/resolver.rb:275:in `search_for'
      /Users/lmelia/.rvm/gems/ruby-2.1.6@yapp/gems/bundler-1.11.0/lib/bundler/resolver.rb:332:in `block in verify_gemfile_dependencies_are_found!'
      /Users/lmelia/.rvm/gems/ruby-2.1.6@yapp/gems/bundler-1.11.0/lib/bundler/resolver.rb:330:in `each'
      /Users/lmelia/.rvm/gems/ruby-2.1.6@yapp/gems/bundler-1.11.0/lib/bundler/resolver.rb:330:in `verify_gemfile_dependencies_are_found!'
      /Users/lmelia/.rvm/gems/ruby-2.1.6@yapp/gems/bundler-1.11.0/lib/bundler/resolver.rb:199:in `start'
[...]

PR #4152, opened 2015-12-13
Handle LazySpecifications and Ruby versions

indirect:
This is a fix for #4147, as well as a change in which Ruby version we use to resolve gems against.

### DIFF
--- a/lib/bundler/lazy_specification.rb
+++ b/lib/bundler/lazy_specification.rb
@@ -54,7 +54,7 @@ module Bundler
     end
 
     def respond_to?(*args)
-      super || @specification.respond_to?(*args)
+      super || @specification ? @specification.respond_to?(*args) : nil
     end
 
     def to_s


indirect:
I give up, I can't for the life of me figure out how to create a spec that will run `bundle install` in a way that uses LazySpecifications. ¯\_(ツ)_/¯

indirect:
@segiddins r?

segiddins:
@indirect if you make it pull from an artifice server

indirect:
@segiddins no, sadly... we only create LazySpecs from the lockfile parser. Hmm...

segiddins:
I don't think the change in ruby version is correct either? What's it trying to address?

indirect:
The Ruby version that we resolve against gem's required_ruby_versions should be the running RUBY_VERSION, not the `ruby` DSL requirement. For example, if you Bundle on 1.9, Bundler should not try to install gems that require 2.0. The `ruby` DSL method is a completely separate requirement, and raises an exception if Bundler is run on a Ruby that doesn't match that requirement. So, for example, if you have `ruby "2.0"` in your Gemfile, you'll get an exception on 1.9 before the resolver starts. I hope that makes sense, I feel really tired right now.

segiddins:
It makes sense, but I intentionally wrote the feature to resolve against the gemfile ruby version

indirect:
Yeah, I think that was my bad when we were talking about the feature design. Sorry. :/

segiddins:
No, I implemented it the way I did because I think it's the correct way for the feature to work. If you want to change that now, we need to discuss, because I wrote it this way for a reason.

lib/bundler/resolver.rb
```diff
@@ -127,14 +127,15 @@ def source
...
 
-      def for?(platform, required_ruby_version)
-        if spec = @specs[platform]
-          if required_ruby_version && spec_required_ruby_version = spec.required_ruby_version
-            spec_required_ruby_version.satisfied_by?(required_ruby_version)
-          else
-            true
-          end
+      def for?(platform, required_ruby)
+        return true if required_ruby.nil?
```
segiddins: This definitely looks wrong

indirect:
@homu r+

galori:
Is this fix in bundler 1.11.1 ?

I believe we're still having the same issue with bundler 1.11.1 which is why I'm asking.

segiddins:
@galori try `1.11.2`, @indirect just shipped a patch :)

PR #4156, opened 2015-12-13
Suppress warnings about "already initialized constant"s in bundle update spec

RochesterinNYC:


RochesterinNYC:
For the `bundle update --ruby` spec in `commands/update_spec.rb`, the double execution of `install_gemfile` causes the following warnings for each context: `warning: already initialized constant RUBY_VERSION` and `warning: already initialized constant RUBY_PATCHLEVEL`. This PR suppresses these extraneous warnings and cleans up the RSpec test suite output.

spec/support/helpers.rb
```diff
@@ -195,7 +196,7 @@ def install_gemfile(*args)
       gemfile(*args)
       opts = args.last.is_a?(Hash) ? args.last : {}
       opts[:retry] ||= 0
-      bundle :install, opts
+      bundle :install, opts.merge(:suppress_stderr => true)
```
segiddins: This definitely should not be done

segiddins:
What's wrong with `expect_err`?

spec/support/helpers.rb
RochesterinNYC: Ah, this would suppress every call of `install_gemfile` in every spec. Removing that for sure.

RochesterinNYC:
@segiddins Ah, oversight on my part. Basically reinvented the `expect_err` wheel there. All the existing spec support/helper infrastructure for suppressing these warnings is already there so switched the PR to simply utilize the `expect_err` option.

segiddins:
On further inspection, I think we can get rid of those warnings just by using the spec helper for simulating a ruby version?

RochesterinNYC:
Which spec helper are you referring to? The only one I found that can simulate a ruby version seems to be `simulate_ruby_engine(engine, version = "1.6.0")`. This could simulate changing the ruby version through the ENV and the `BUNDLER_SPEC_RUBY_ENGINE_VERSION` value in particular, but I don't see how this affects these specific tests as they're primarily working with Gemfile creation with particular information and Gemfile.lock creation/updating (via `bundle update`) with particular information?

RochesterinNYC:
Also, a quick note on the design of these specs, it seems that they're asserting/setting up the scenarios incorrectly? 

```
  it "removes the Ruby from the Gemfile.lock" do
      install_gemfile <<-G
          ::RUBY_VERSION = '2.1.3'
          ::RUBY_PATCHLEVEL = 100
          ruby '~> 2.1.0'
      G

      install_gemfile <<-G
          ::RUBY_VERSION = '2.1.4'
          ::RUBY_PATCHLEVEL = 222
      G

      bundle "update --ruby"

      lockfile_should_be <<-L
       GEM
         specs:

       PLATFORMS
         ruby

       DEPENDENCIES

       BUNDLED WITH
          #{Bundler::VERSION}
      L
    end
```

In the following code, the Gemfile is getting created with the first configuration, then overwritten with the second configuration, then the `bundle update --ruby` command is being executed which actually creates the `Gemfile.lock` file I believe. Then, the assertion is made that the `Gemfile.lock` file should be a certain configuration but there was no `Gemfile.lock` file that was created with the first Gemfile configuration. The `Gemfile.lock` that exists (if one does) in the `tmp/bundled_app` directory before `bundle update --ruby` is a polluted one from some other spec that did not clean up the file. 

This seems incorrect as it seems that the assertion is supposed to test that the `Gemfile.lock` internals that are created from the first `Gemfile` configuration are different in some manner (missing `ruby 2.1.3p100`) from the `Gemfile.lock` internals set by calling `bundle update --ruby` on the second configured Gemfile. 

I think we should be running `bundle update --ruby` after the first `install_gemfile` call as well to generate the `Gemfile.lock` file with that first `Gemfile` configuration. Additionally, we should cleaning up `tmp/bundled_app/Gemfile.lock` and `tmp/bundled_app/Gemfile` before (or after?) each of these specs.

RochesterinNYC:
Something like:
```
before do 
  clear_gemfile
  clear_gemfile_lock
end
context "when the Gemfile removes the ruby" do
  it "removes the Ruby from the Gemfile.lock" do
    #No Gemfile or Gemfile.lock exists in tmp/bundled_app
    install_gemfile <<-G, :expect_err => true
        ::RUBY_VERSION = '2.1.3'
        ::RUBY_PATCHLEVEL = 100
        ruby '~> 2.1.0'
    G
    #Gemfile with first config created, no Gemfile.lock

    bundle "update --ruby"

    #Gemfile with first config created, Gemfile.lock created off first config

=begin
    Gemfile.lock should be at this state right now
    lockfile_should_be <<-L
     GEM
       specs:

     PLATFORMS
       ruby

     DEPENDENCIES

     RUBY VERSION
        ruby 2.1.3p100

     BUNDLED WITH
        #{Bundler::VERSION}
    L
=end

    install_gemfile <<-G, :expect_err => true
        ::RUBY_VERSION = '2.1.4'
        ::RUBY_PATCHLEVEL = 222
    G

    #Gemfile with second config created, Gemfile.lock off first config still there

    bundle "update --ruby"

    #Gemfile with second config still there, Gemfile.lock off created off second config
    #If bundle update --ruby wasn't working right, Gemfile.lock should not have changed 

    #Below assertion checks that Gemfile.lock has indeed changed 

    lockfile_should_be <<-L
     GEM
       specs:

     PLATFORMS
       ruby

     DEPENDENCIES

     BUNDLED WITH
        #{Bundler::VERSION}
    L
  end
end
```

Apologize if it's a bit hard to read with the unhelpful way the Ruby comments are displayed/rendered on Github.

RochesterinNYC:
Updated the PR with these corrections in test setup.

spec/support/helpers.rb
```diff
@@ -177,6 +177,16 @@ def gemfile(*args)
       create_file("Gemfile", *args)
     end
 
+    def clear_gemfile
+      gemfile_path = bundled_app("Gemfile")
+      File.delete(gemfile_path) if File.exist?(gemfile_path)
+    end
+
+    def clear_gemfile_lock
```
segiddins: `delete_lockfile`?

RochesterinNYC:
Renamed `clear_gemfile` and `clear_gemfile_lock` to `delete_gemfile` and `delete_gemfile_lock`. Not sure if we want to use these more aggressively to prevent test pollution or in a global before/after way.

segiddins:
@RochesterinNYC the bundled app directory is blown away between tests already.

RochesterinNYC:
@segiddins Oh okay, gotcha. Does that also apply for running a single test as well? I was encountering issues with inconsistent results with running the same test a few times in a row and it seemed as if the Gemfile and Gemfile.lock were persisting. If it does apply, then I'll remove `delete_gemfile` and `delete_gemfile_lock`.

segiddins:
That would be a separate issue, bundled app should be deleted before every test.

RochesterinNYC:
@segiddins Okay, removed any mention of `delete_gemfile*`.

Okay, dug around a bit and noticed that it's `reset!` that deletes the `tmp/bundled_app` directory along with some other stuff and `reset!` gets called before each spec so my flakiness must have been a personal system issue or something. The reason I was confused by that was because `reset!` isn't called after specs so my `tmp` files including `bundled_app` were still there after any specs were run, but now I see that it won't interfere with the validity of any actual specs. All clear now :+1:

segiddins:
@RochesterinNYC lingering `tmp` files are actually a plus when debugging single specs :)

RochesterinNYC:
Gotcha, valid point :+1:

RochesterinNYC:
Anything else needed on this PR?

segiddins:
Looks good to me :)
Just want @indirect to approve it

indirect:
@homu r+

PR #4160, opened 2015-12-14
Resolve gems against the running Ruby version

indirect:
The initial implementation of respecting a gem's required_ruby_version used the Gemfile `ruby` requirement. I believe that what we actually want to do (and what users will expect) is to resolve the required_ruby_version against the currently running version of Ruby, even if the Gemfile does not contain a `ruby` call.

/cc @segiddins

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -194,8 +194,7 @@ module Bundler
           last_resolve
         else
           # Run a resolve against the locally available gems
-          requested_ruby_version = ruby_version.version if ruby_version
-          last_resolve.merge Resolver.resolve(expanded_dependencies, index, source_requirements, last_resolve, requested_ruby_version)
+          last_resolve.merge Resolver.resolve(expanded_dependencies, index, source_requirements, last_resolve, Bundler.ruby_version)
         end
       end
     end
--- a/lib/bundler/lazy_specification.rb
+++ b/lib/bundler/lazy_specification.rb
@@ -54,7 +54,7 @@ module Bundler
     end
 
     def respond_to?(*args)
-      super || @specification.respond_to?(*args)
+      super || @specification ? @specification.respond_to?(*args) : nil
     end
 
     def to_s
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -127,14 +127,15 @@ module Bundler
         @source ||= first.source
       end
 
-      def for?(platform, required_ruby_version)
-        if spec = @specs[platform]
-          if required_ruby_version && spec_required_ruby_version = spec.required_ruby_version
-            spec_required_ruby_version.satisfied_by?(required_ruby_version)
-          else
-            true
-          end
+      def for?(platform, required_ruby)
+        return true if required_ruby.nil?
+        return false if @specs[platform].nil?
+
+        if @specs[platform].respond_to?(:required_ruby_version)
+          spec_ruby = @specs[platform].required_ruby_version
         end
+
+        spec_ruby.nil? ? true : spec_ruby.satisfied_by?(required_ruby)
       end
 
       def to_s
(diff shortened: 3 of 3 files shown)


segiddins:
I have disagreements with this that I will explain in due course.

indirect:
@segiddins ping for disagreements

segiddins:
So, I'm very strongly against this because it means you'll get different resolutions based solely upon what version of Ruby you're running -- and because you haven't specified a Ruby version in your Gemfile, you've implied that you'll need to support _any ruby version imaginable_.

lamont-granquist:
If I'm reading this right currently without a ruby version in a Gemfile, bundler will ignore the required_ruby_version in the gemspecs entirely?  That seems a little surprising to me, and could lead to lockfiles being created where there was no ruby version that would satisfy all the required_ruby_version constraints (although that might be a little bit contrived based on how the required_ruby_version in the gemspec usually gets used).  I'd think that either the running ruby version is used when there's no ruby in the Gemfile, or else the presence of a required_ruby_version in any gempsec causes an exception to be thrown if there isn't a ruby line in the Gemfile.

indirect:
@segiddins I might need to make the specs clearer, but I think it is _very_ surprising to ignore the current ruby version when resolving gems. In Gemfiles that don't have a specified Ruby version, that means that every resolve has an implied "and the gems in the lockfile need to allow this version of Ruby". And I would expect Bundler to raise a conflict exception if the running Ruby conflicts with the locked gems.

segiddins:
I totally don't. If you at all care about what bundler thinks about ruby versions, add a ruby version to your gemfile, especially since that now supports arbitrary requirements.

indirect:
Here's the specific situation that I am trying to improve for users:

0. Avoid using Gemfile `ruby` (because it's library, or because it's in `.ruby-version`).
1. Add a gem whose newest release has a `required_ruby_version` greater than your ruby.
2. Run `bundle install`, see Bundler say that everything is good, get an exception and a backtrace.
3. Exclaim "wtf Bundler knew my ruby version, why did it explicitly choose a version it knew couldn't be installed".

I think there are two ways to resolve this scenario:

Option 1: If the gem is not yet locked, Bundler silently chooses a version that can be installed as part of running the resolver, and than version is then locked. If the gem is already locked, and it is not compatible with the currently running version of Ruby, Bundler explains that your locked gems can't be installed on this Ruby version.

Option 2: Bundler refuses to resolve, because the resolution will be impossible to install. I would expect at a minimum a message explaining which version of the gem _will_ install on this Ruby version, so that it can be added explicitly to the Gemfile.

Is there another option I'm not thinking of? I'm not okay with "they should have known better", because that means we screwed up somewhere, not them.

segiddins:
1) Either use `ruby` or use 2.0, which will pull from `.ruby-version` (and which bundler's own repos already do)

segiddins:
IMHO we should resolve and then explain that the user should either specify a ruby version in their lockfile or specify a stricter dependency before attempting to resolve.

indirect:
Saying "use `ruby`" is not a solution to the use-case of "I ran `bundle install` and Bundler resolved some versions that throw exceptions rather than install". :/

yob:
I found this PR today while trying to resolve an issue, so thought I'd throw in my 2c.

I maintain the pdf-reader gem, and my preference is for the 1.x series to maintain Ruby  1.8.7 support. That means I'd like to run builds on travis CI for all rubies from 1.8.7 to 2.3, so my Gemfile doesn't specify a Ruby version.

The latest version of some of my dependencies require 1.9.3 (like parallel and ttfunk) and my Gemfile.lock isn't in git, so currently running `bundle install` during the build attempts to resolve a dependency graph.

On 1.8.7, the resolving fails because of the dependencies that require >=1.9.3.

Adding Ruby to the Gemfile doesn't feel like an ideal solution. It'd be nice if bundler was able to discover older versions of the dependencies that are 1.8.7 compatible and "just work". I'm not familiar with the constraints and behaviour of the bundler resolver though, so it's likely I'm missing some nuance.

Issue #4158, opened 2015-12-14
bundle package --all is broken in bundler 1.11.0


janost:
- What did you do?
  I ran the command `/usr/local/bin/bundle package --all`

- What did you expect to happen?
  I expected `bundle package --all` to work.

- What happened instead?
  Instead, the command seems to create vendor/cache/PROJECTNAME/vendor/cache/PROJECTNAME/... in and endless loop until the file system permits.

```
~ $ bundle -v
Bundler version 1.11.0
~ $ bundle gem something
Creating gem 'something'...
      create  something/Gemfile
      create  something/.gitignore
      create  something/lib/something.rb
      create  something/lib/something/version.rb
      create  something/something.gemspec
      create  something/Rakefile
      create  something/README.md
      create  something/bin/console
      create  something/bin/setup
Initializing git repo in /Users/user/something
~ $ cd something
~/something (master*) $ bundle package --all
Fetching gem metadata from https://rubygems.org/..
Fetching version metadata from https://rubygems.org/.
Resolving dependencies...
Using rake 10.4.2
Using bundler 1.11.0
Using something 0.1.0 from source at `.`
Bundle complete! 3 Gemfile dependencies, 3 gems now installed.
Use `bundle show [gemname]` to see where a bundled gem is installed.
Updating files in vendor/cache
  * rake-10.4.2.gem
--- ERROR REPORT TEMPLATE -------------------------------------------------------
- What did you do?

  I ran the command `/usr/local/bin/bundle package --all`

- What did you expect to happen?

  I expected Bundler to...

- What happened instead?

  Instead, what actually happened was...


Error details

    Errno::ENAMETOOLONG: File name too long - /Users/user/something/vendor/cache/something/vendor/cache/something/vendor/cache/something/vendor/cache/something/vendor/cache/something/vendor/cache/something/vendor/cache/something/vendor/cache/something/vendor/cache/som
ething/vendor/cache/something/vendor/cache/something/vendor/cache/something/vendor/cache/something/vendor/cache/something/vendor/cache/something/vendor/cache/something/vendor/cache/something/vendor/cache/something/vendor/cache/something/vendor/cache/something/vendor/cac
he/something/vendor/cache/something/vendor/cache/something/vendor/cache/something/vendor/cache/something/vendor/cache/something/vendor/cache/something/vendor/cache/something/vendor/cache/something/vendor/cache/something/vendor/cache/something/vendor/cache/something/vend
[...]

PR #4161, opened 2015-12-15
Fix broken `bundle package --all` command and prevent endless recursive copy

RochesterinNYC:
- Only cache gem specs that are not bundler and not the gem specified by
  the gemspec in the main directory the command is being executed from

### DIFF
--- a/spec/commands/package_spec.rb
+++ b/spec/commands/package_spec.rb
@@ -15,6 +15,50 @@ describe "bundle package" do
     end
   end
 
+  context "with --all" do
+    context "without a gemspec" do
+      it "caches all dependencies except bundler itself" do
+        gemfile <<-D
+          source "file://#{gem_repo1}"
+          gem 'rack'
+          gem 'bundler'
+        D
+
+        bundle "package --all"
+
+        expect(bundled_app("vendor/cache/rack-1.0.0.gem")).to exist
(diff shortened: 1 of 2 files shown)


RochesterinNYC:
Addresses #4158.

Verified that it works in a directory that has a `.gemspec` and in a directory that doesn't have one (ex. Rails app).

segiddins:
Could you please add a spec for this fix? Thanks!

RochesterinNYC:
@segiddins Sure! Added a spec covering the `bundle package --all` vs. with a `.gemspec` and without a `.gemspec` cases.

indirect:
@homu r+

RochesterinNYC:
Thanks!

spec/commands/package_spec.rb
```diff
@@ -15,6 +15,50 @@
...
+      end
+      it "caches all dependencies except bundler and the gemspec specified gem" do
+        gemfile <<-D
+          source "file://#{gem_repo1}"
+          gem 'rack'
+          gemspec
+        D
+
+        bundle "package --all"
+        sleep 20
```
segiddins: :sob: (Found this in another PR, so no need to fix separately)

RochesterinNYC: Oh geez, sorry about this. Think this was left over from when I was using the `sleep` to inspect the local files/what `bundle package --all` was doing during the execution of the spec.

Issue #4170, opened 2015-12-17
Permission denied @ unlink_internal


oka-haist:
- What did you do?

  I ran the command `/usr/local/bin/bundle clean --force`

- What did you expect to happen?

  I expected Bundler to remove all the gems from my system

- What happened instead?

  Instead, it removed a few and stopped. Can't go on with bundle clean

Error details

    Errno::EACCES: Permission denied @ unlink_internal - /usr/local/lib/ruby/gems/2.2.0/cache/sdoc-0.4.1.gem
      /usr/local/lib/ruby/2.2.0/fileutils.rb:1450:in `unlink'
      /usr/local/lib/ruby/2.2.0/fileutils.rb:1450:in `block in remove_file'
      /usr/local/lib/ruby/2.2.0/fileutils.rb:1455:in `platform_support'
      /usr/local/lib/ruby/2.2.0/fileutils.rb:1449:in `remove_file'
      /usr/local/lib/ruby/2.2.0/fileutils.rb:793:in `remove_file'
      /usr/local/lib/ruby/2.2.0/fileutils.rb:571:in `block in rm'
      /usr/local/lib/ruby/2.2.0/fileutils.rb:570:in `each'
      /usr/local/lib/ruby/2.2.0/fileutils.rb:570:in `rm'
      /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.10.6/lib/bundler/runtime.rb:204:in `block in clean'
      /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.10.6/lib/bundler/runtime.rb:204:in `each'
      /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.10.6/lib/bundler/runtime.rb:204:in `clean'
      /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.10.6/lib/bundler/cli/clean.rb:11:in `run'
      /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.10.6/lib/bundler/cli.rb:373:in `clean'
      /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.10.6/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
      /usr/local/lib/ruby/gems/2.2.0/gems/bundler-1.10.6/lib/bundler/vendor/thor/lib/thor/invocation.rb:126:in `invoke_command'
[...]

PR #4173, opened 2015-12-18
Use SharedHelpers#filesystem_access in bundle clean for a friendlier error message for denied permissions

RochesterinNYC:


RochesterinNYC:
Addresses #4170

lib/bundler/runtime.rb
```diff
@@ -189,8 +189,16 @@ def clean(dry_run = false)
 
       unless dry_run
         stale_files = stale_gem_bins + stale_gem_files + stale_gemspec_files
-        stale_files.each {|file| FileUtils.rm(file) if File.exist?(file) }
-        stale_git_cache_dirs.each {|dir| FileUtils.rm_rf(dir) if File.exist?(dir) }
+        stale_files.each do |file|
+          SharedHelpers.filesystem_access(File.dirname(file)) do |_p|
```
segiddins: Why the `dirname `?

segiddins:
👍🏻 perfect other than the one comment

lib/bundler/runtime.rb
RochesterinNYC: Hm, so I was reading up on what the Read-Write-Exec permissions needed for removing a file are and it seems that oddly enough, Write permissions on the directory that the file resides in are needed. Write or execute permissions on that file itself aren't needed since file removal is essentially modifying the directory and its info. (http://unix.stackexchange.com/questions/48579/why-can-rm-remove-read-only-files)

Hence, for the error we potentially throw from the `filesystem_access` method, we'd want to tell the user they need write access on the directory for the file targeted for deletion.

segiddins:
@homu r+

PR #4178, opened 2015-12-21
Redirects any usage of --help flag for a command to that command's man page

RochesterinNYC:
In response to: bundler/bundler-features#97


RochesterinNYC:
In response to: bundler/bundler-features#97

indirect:
Awesome, thanks so much!

@homu r+

PR #4184, opened 2015-12-25
[Travis] Test against Ruby 2.3.0

segiddins:
\c @indirect


indirect:
Thanks! It seems good to get this working soon.

RochesterinNYC:
So does Travis CI have the official 2.3.0 stable rvm ruby up yet? Or does it only have the release candidates and etc.? 

Ran the test suite with RVM - Ruby 2.3.0 locally and there are a few test failures and a fair amount of deprecation warnings that would be great to work on.

segiddins:
Looks like a bunch of those warnings are from rspec-mocks

ddrmanxbxfr:
@segiddins If you want to trigger a build in travis before 2.3 version works.

Try specifying 2.3.0. I've updated a few projects specifying this version and the test run great on travis :).

ddrmanxbxfr:
@RochesterinNYC Yes 2.3 is up but with alias 2.3.0

segiddins:
@homu r+

PR #4185, opened 2015-12-25
Ensure `bundle gem` default task matches framework

dthjones:
Previously running `bundle gem` with a global config containing

    BUNDLE_GEM__TEST: 'minitest'

would incorrectly set the default task in the generated Rakefile to

    task :default => :spec

whereas `bundle gem --test=minitest` worked as expected.  This happened because the test task was determined before checking the global config or asking for user input, making the test task and default task inconsistent.

This fix moves the decision to after the test framework has been properly determined.


segiddins:
👍🏻

indirect:
Awesome, thanks!

@homu r+

Issue #4128, opened 2015-11-27
Probe the configured mirror and drop to the source uri in case it fails


pcarranza:
To support using *gemstash* in the office and keep the *zero conf* style when the user is not in the same network anymore we could just probe the configured mirror for a given upstream server to check if it is still there. If it is not, we just drop down to the actual upstream server.

The behaviour would be something like this:

open tcp connection to the upstream mirrored with a connection timeout of 0.5 seconds, on timeout, keep the url as is to go for the actual upstream server.

A quick look at the code points me at the `remote.rb` file. Would it be acceptable to probe the server there? should I look any other place in the source? any particular guideline?

cc @indirect

TimMoore:
I'm not sure everyone would want this behavior. Some people want to enforce use of a mirror for security reasons. So I think it would need to be configurable and default to off.

indirect:
Yes, I think that's how I would like it to work for me, too. I would even be happy with a much lower timeout, since the gemstash server should be local if it is going to be available at all. Maybe we can provide this via a `bundle config mirror.URL.fallback true` setting?

indirect:
(I think @TimMoore is right, and we want this to be opt-in for mirror fallback, but for developers using gemstash in their office, this could be really nice.)

pcarranza:
I agree, it makes sense to make it an optional behaviour.

Then the way to enable this fallback is by providing this `.fallback=true` trailing to a mirror url.

Timeout of 0.1 secs? configurable with an option? => `bundle config mirror.URL.fallback_timeout` for ex?

I'll give it a try.

pcarranza:
Just wondering if it makes sense to just enable the `fallback_timeout` and not use the `fallback=true` at all.

TimMoore:
I was thinking the same. If there were two configs, it might be confusing what the behavior should be when `fallback_timeout` is set but `fallback` isn't. Maybe just have `fallback_timeout` which can be set to either a number of milliseconds or `true` to use the default timeout.

pcarranza:
Yep, it makes sense and keeps things neat and simple. I like it.

I'll build it as soon as I can. Thanks for the feedback.

pcarranza:
A feature suggested by @smellsblue that I think makes sense: add support for setting gemstash as the mirror for all the upstreams, something like this: `bundle config mirror.all http://localhost:9292`

That would make gemstash the default server for anything, and we should just send the header to let gemstash know what are we looking for.

[...]

PR #4186, opened 2015-12-25
Mirror fallback timeout with TCP probing

pcarranza:
cc/ @indirect 

This PR implements #4128

Feature: support probing a mirror with a fallback timeout, if the probe fails, bundler drops down to the source url instead of using the mirror.

There are basically 2 ways of declaring a mirror:
* `bundler config mirror.all uri` -> this will set a mirror for any uri provided, matching the new bundler behaviour of sending the upstream as a header to the mirror.
* `bundler config mirror.escaped_uri uri` -> old style, setting up a mirror for a specific uri.

On top of this, there is a new option which allows declaring a `fallback_timeout` for any mirror:
* `bundler config mirror.all.fallback_timeout number|boolean`
* `bundler config mirror.escaped_uri.fallback_timeout number|boolean`

If a mirror has a fallback timeout set, bundler will probe a tcp connection to it, if the connection succeeds, bundler will use the mirror, otherwise bundler will call the original url skipping the mirror completely.

The fallback timeout supports both a number (float) and a boolean, if `true` is specified it will use the default timeout of 0.1 seconds, in case false is specified it will disable probing completely.

This maintains the old behaviour in the case probing is not setup and the mirror is not available, it will just fail as it was before.

There are a few caveats: this does not support partial configuration
For example: setting up many mirrors for specific urls and then trying to setup a fallback timeout for `all` will not apply the `all` timeout for the different specific urls. 
All is treated as a specific uri that has priority over any other specific configuration. So these configurations may be ignored.

When a given mirror returns more than one address (ipv4+ipv6 for ex) all of them will be probed, this means that if a given mirror corresponds to 6 addresses, and all of them fail, there may be 6 probe timeouts, which means that the time it will take will be 0.6 seconds in the case that it is setup with the default timeout.

This was tested briefly (as it is basically a RFC for now) like this:

``` bash
~$ bundler config mirror.all http://localhost:9292
~$ cat .bundle/config 
---
BUNDLE_MIRROR__ALL: http://localhost:9292

...
make-me-an-artist(master)$ bundle
Fetching source index from http://localhost:9292/
Retrying fetcher due to error (2/4): Bundler::HTTPError Could not fetch specs from http://localhost:9292/
Retrying fetcher due to error (3/4): Bundler::HTTPError Could not fetch specs from http://localhost:9292/
Retrying fetcher due to error (4/4): Bundler::HTTPError Could not fetch specs from http://localhost:9292/
Could not fetch specs from http://localhost:9292/
...
~$ bundler config mirror.all.fallback_timeout true
~$ cat .bundle/config 
---
BUNDLE_MIRROR__ALL: http://localhost:9292
BUNDLE_MIRROR__ALL__FALLBACK_TIMEOUT: 'true'
...
make-me-an-artist(master)$ bundle
Installing arrayfields 4.9.2
...
```


lib/bundler/mirror.rb
```diff
@@ -0,0 +1,183 @@
+module Bundler
+  class Settings
+    class Mirrors
+      def initialize(prober = nil)
+        @all = Mirror.new
```
segiddins: these should all be `attr_reader`

lib/bundler/mirror.rb
```diff
@@ -0,0 +1,183 @@
+module Bundler
+  class Settings
+    class Mirrors
```
segiddins: this class name is confusing

lib/bundler/mirror.rb
```diff
@@ -0,0 +1,183 @@
...
+      def validate!(probe = nil)
+        @valid = false if uri.nil?
+        if @valid.nil?
+          @valid = fallback_timeout == 0 || (probe || TCPSocketProbe.new).replies?(self)
+        end
+        self
+      end
+    end
+
+    private
```
segiddins: I don't believe this `private` affects any methods?

lib/bundler/settings.rb
```diff
@@ -244,14 +239,16 @@ def convert_to_backward_compatible_key(key)
 
     # TODO: duplicates Rubygems#normalize_uri
     # TODO: is this the correct place to validate mirror URIs?
-    def normalize_uri(uri)
-      uri = uri.to_s
-      uri = "#{uri}/" unless uri =~ %r{/\Z}
-      uri = URI(uri)
-      unless uri.absolute?
-        raise ArgumentError, "Gem sources must be absolute. You provided '#{uri}'."
+    class AbsoluteURI
```
segiddins: should be a module with a `module_function`?

segiddins:
I'd love to see some YARD documentation for the new classes ;)

segiddins:
IMHO this also requires a integration-style spec that exercises the fallback path during a `bundle install`

lib/bundler/settings.rb
pcarranza: Sure, I'll take it back as a module function.

lib/bundler/mirror.rb
pcarranza: Nop, it's gone.

lib/bundler/mirror.rb
pcarranza: I agree that it is not specific enough, but I honestly cannot think of any better name. Any suggestion?

lib/bundler/mirror.rb
pcarranza: Not sure about this, all these are used for deciding which mirror should be returned depending on the validity and the configuration.

By making these public it would be expanding the public API of this object leaking internal implementation details.

pcarranza:
I've added documentation for the new classes, hopefully improved rubocop feelings.

I only have a question regarding the integration test, where can I find a sample of how are they implemented in bundler?

segiddins:
Any of the specs that do `install_gemfile` and use an `artifice` option -- those are the ones that will let you  mock the HTTP server.

pcarranza:
Right, I see how.

Will figure something out for the integration test, I have a few ideas already.

indirect:
We might need to do a realworld test that boots a rack server -- unfortunately, artifice intercepts network requests in a way that doesn't allow testing some headers, so I think it won't allow testing sockets either. There's an example of booting a rack server in the realworld test for the X-Gemserver header.

pcarranza:
Yeah, exactly what I was thinking.

Thanks

pcarranza:
I've added a few real world end to end specs that setup a server using rack and a mirror (which is not there) just to assert that when there is no fallback timeout defined, bundler will fail, and when it is, bundler will reach the original url.

Both using the configuration specific for one mirror, or using the configuration for **all**

indirect:
Great, sounds like exactly what we were looking for. I'm ready to merge this once the build is green. :)

pcarranza:
Yay!

There is something odd with the TCPServer, apparently it can return a port that is still in use by other socket.

I restored the linear way of looking for an unused port.

indirect:
@homu r+

Issue #4054, opened 2015-10-13
bundle install: undefined method `use_ssl=' for #<Net::HTTP::Persistent::SSLReuse


OneMain:
dongchao:xmo iclick$ bundle  install
--- ERROR REPORT TEMPLATE -------------------------------------------------------
- What did you do?

  I ran the command `/Users/iclick/.rvm/gems/ruby-1.9.1-p431/bin/bundle install`

- What did you expect to happen?

  I expected Bundler to...

- What happened instead?

  Instead, what actually happened was...

Error details

    NoMethodError: undefined method `use_ssl=' for #<Net::HTTP::Persistent::SSLReuse ruby.taobao.org:443 open=false>
      /Users/iclick/.rvm/gems/ruby-1.9.1-p431/gems/bundler-1.10.5/lib/bundler/vendor/net/http/persistent.rb:1136:in `ssl'
      /Users/iclick/.rvm/gems/ruby-1.9.1-p431/gems/bundler-1.10.5/lib/bundler/vendor/net/http/persistent.rb:626:in `connection_for'
      /Users/iclick/.rvm/gems/ruby-1.9.1-p431/gems/bundler-1.10.5/lib/bundler/vendor/net/http/persistent.rb:994:in `request'
      /Users/iclick/.rvm/gems/ruby-1.9.1-p431/gems/bundler-1.10.5/lib/bundler/fetcher/downloader.rb:45:in `request'
      /Users/iclick/.rvm/gems/ruby-1.9.1-p431/gems/bundler-1.10.5/lib/bundler/fetcher/downloader.rb:15:in `fetch'
      /Users/iclick/.rvm/gems/ruby-1.9.1-p431/gems/bundler-1.10.5/lib/bundler/fetcher/dependency.rb:7:in `api_available?'
      /Users/iclick/.rvm/gems/ruby-1.9.1-p431/gems/bundler-1.10.5/lib/bundler/fetcher.rb:135:in `block in use_api'
      /Users/iclick/.rvm/gems/ruby-1.9.1-p431/gems/bundler-1.10.5/lib/bundler/fetcher.rb:135:in `reject!'
      /Users/iclick/.rvm/gems/ruby-1.9.1-p431/gems/bundler-1.10.5/lib/bundler/fetcher.rb:135:in `use_api'
      /Users/iclick/.rvm/gems/ruby-1.9.1-p431/gems/bundler-1.10.5/lib/bundler/source/rubygems.rb:320:in `select'
[...]

PR #4188, opened 2015-12-27
Catch NoMethodError for `use_ssl=` and raise appropriate FriendlyErrors warning

RochesterinNYC:
Addresses #4054. 

- `NoMethodError` for `use_ssl=` in the downloader should now throw a `LoadError` with `cannot load such file -- openssl` as the message which will be caught and logged by the `FriendlyErrors` module as an OpenSSL Ruby error.


indirect:
thanks! @homu r+

Issue #4183, opened 2015-12-23
Errno::EAGAIN: Resource temporarily unavailable @ rb_sysopen


p-vital:
Getting this error when running bundle install and a new gem needs to be downloaded/installed:
<pre>
--- ERROR REPORT TEMPLATE -------------------------------------------------------
- What did you do?

  I ran the command `/home/user/.rbenv/versions/2.2.4/bin/bundle install`

- What did you expect to happen?

  I expected Bundler to...

- What happened instead?

  Instead, what actually happened was...

Error details

    Errno::EAGAIN: Resource temporarily unavailable @ rb_sysopen - /home/user/app/Gemfile.lock
      /home/user/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/bundler-1.11.2/lib/bundler/definition.rb:268:in `initialize'
      /home/user/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/bundler-1.11.2/lib/bundler/definition.rb:268:in `open'
      /home/user/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/bundler-1.11.2/lib/bundler/definition.rb:268:in `block in lock'
      /home/user/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/bundler-1.11.2/lib/bundler/shared_helpers.rb:114:in `filesystem_access'
      /home/user/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/bundler-1.11.2/lib/bundler/definition.rb:267:in `lock'
      /home/user/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/bundler-1.11.2/lib/bundler/environment.rb:34:in `lock'
      /home/user/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/bundler-1.11.2/lib/bundler/installer.rb:72:in `run'
      /home/user/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/bundler-1.11.2/lib/bundler/installer.rb:21:in `install'
      /home/user/.rbenv/rbenv.d/exec/gem-rehash/rubygems_plugin.rb:27:in `install'
      /home/user/.rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/bundler-1.11.2/lib/bundler/cli/install.rb:105:in `run'
[...]

PR #4189, opened 2015-12-28
Create `TemporaryResourceError` that `SharedHelpers#filesystem_access` raises for `Errno::EAGAIN`

RochesterinNYC:
Addresses #4183

Creates a `TemporaryResourceError` that is raised when an `Errno::EAGAIN` occurs when trying to perform some file op. `Errno::EAGAIN`s can occur for a variety of scenarios when certain system resources (ex. memory, in which case a fork cannot occur) aren't available at a certain moment, so the error message suggests that the user rerun the operation, which worked in the case of #4183.

### DIFF
--- a/lib/bundler/errors.rb
+++ b/lib/bundler/errors.rb
@@ -58,19 +58,32 @@ module Bundler
       @permission_type = permission_type
     end
 
-    def message
-      action = case @permission_type
-               when :read then "read from"
-               when :write then "write to"
-               when :executable, :exec then "execute"
-               else @permission_type.to_s
+    def action
+      case @permission_type
+      when :read then "read from"
+      when :write then "write to"
+      when :executable, :exec then "execute"
+      else @permission_type.to_s
       end
+    end
+
+    def message
       "There was an error while trying to #{action} `#{@path}`. " \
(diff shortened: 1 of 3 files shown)


lib/bundler/errors.rb
```diff
@@ -73,6 +73,27 @@ def message
     status_code(23)
   end
 
+  class TemporaryResourceError < BundlerError
```
segiddins: inherit from `PermissionError` and extract the `action` into a method so it doesn't need to be duplicated?

RochesterinNYC: Ah thanks. I was having some trouble trying to DRY up that part with a private module method.

Is it an issue that this TemporaryResourceError would be inheriting from PermissionsError when it's not necessarily always something that's thrown from a lack of file permissions? Things like a temporary inability to fork or memory resource issues can cause this error as well.

segiddins: as long as it requires a `permission_type`, it's a permission error to me?

RochesterinNYC:
Updated PR to have `TemporaryResourceError` inherit from `PermissionError`. 

`action` in `PermissionError` is now an instance method as well that its descendants can utilize.

segiddins:
@homu r+

PR #4191, opened 2015-12-29
Reorder Bundler error classes by correlated status code

RochesterinNYC:
Optional green to green refactor of the error classes. While implementing #4189, I would have found it much easier and logical for the error classes to be arranged by status code rather than whatever order or lack of order it seems to currently employ.

### DIFF
--- a/lib/bundler/errors.rb
+++ b/lib/bundler/errors.rb
@@ -5,27 +5,10 @@ module Bundler
     end
   end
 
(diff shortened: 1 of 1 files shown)


segiddins:
awesome, thanks @RochesterinNYC !
@homu r+

RochesterinNYC:
Additionally, I noticed there was no error with status code 9. Is there a reason it was skipped over?

PR #4195, opened 2016-01-01
Add `Net#HTTP--` errors safely (only if in version of ruby stdlib)

RochesterinNYC:
Continuation of #4154. Some errors in the ruby `Net` module that we might not want to retry on were added in later Ruby versions of the stdlib. This PR allows us to preserve backwards compatibility while adding new  errors.

### DIFF
--- a/lib/bundler/fetcher.rb
+++ b/lib/bundler/fetcher.rb
@@ -52,10 +52,16 @@ module Bundler
 
     # Exceptions classes that should bypass retry attempts. If your password didn't work the
     # first time, it's not going to the third time.
-    AUTH_ERRORS = [AuthenticationRequiredError, BadAuthenticationError, Net::HTTPBadGateway,
-                   Net::HTTPBadRequest, Net::HTTPForbidden, Net::HTTPMethodNotAllowed,
-                   Net::HTTPMovedPermanently, Net::HTTPNotImplemented, Net::HTTPNotFound,
-                   Net::HTTPRequestEntityTooLarge, Net::HTTPNoContent]
+    AUTH_ERRORS = [AuthenticationRequiredError, BadAuthenticationError]
+    NET_ERRORS_TO_ADD_SAFELY = [:HTTPBadGateway, :HTTPBadRequest, :HTTPFailedDependency,
+                                :HTTPForbidden, :HTTPInsufficientStorage, :HTTPMethodNotAllowed,
+                                :HTTPMovedPermanently, :HTTPNoContent, :HTTPNotFound,
+                                :HTTPNotImplemented, :HTTPPreconditionFailed, :HTTPRequestEntityTooLarge,
+                                :HTTPRequestURITooLong, :HTTPUnauthorized, :HTTPUnprocessableEntity,
+                                :HTTPUnsupportedMediaType, :HTTPVersionNotSupported]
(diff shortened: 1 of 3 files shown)


segiddins:
If you can add a test for this, that'd be great. If not, I will when I get back home.

RochesterinNYC:
Hrm okay, not sure exactly where to add the test for this. Looking through the code, this is a level above all the `Bundler::Retry` stuff and the fetcher specs currently don't test this. Additionally, since it's just inline code for the Fetcher class, it's difficult to test that section. One way I could potentially see doing it is  putting the logic for safely finding/loading into memory an Exception class into a helper (shared_helper maybe?) and then unit testing that.

The helper method would look maybe something like this:
```ruby
def safely_retrieve_constant(constant_name, namespace)
  return nil unless namespace.constants.include? constant_name
  namespace.const_get(constant_name) 
end
```

The AUTH_ERRORS code would look like:
```ruby
    AUTH_ERRORS = [AuthenticationRequiredError, BadAuthenticationError]
    NET_ERRORS_TO_ADD_SAFELY = [:HTTPBadGateway, :HTTPBadRequest, :HTTPFailedDependency,
                                :HTTPForbidden, :HTTPInsufficientStorage, :HTTPMethodNotAllowed,
                                :HTTPMovedPermanently, :HTTPNoContent, :HTTPNotFound,
                                :HTTPNotImplemented, :HTTPPreconditionFailed, :HTTPRequestEntityTooLarge,
                                :HTTPRequestURITooLong, :HTTPUnauthorized, :HTTPUnprocessableEntity,
                                :HTTPUnsupportedMediaType, :HTTPVersionNotSupported]
    NET_ERRORS_TO_ADD_SAFELY.each do |net_error|
      
AUTH_ERRORS << Net.const_get(net_error) if Net.constants.include? net_error
    end

```

RochesterinNYC:
Sorry, accidentally closed.

Thoughts?

indirect:
Something like this seems reasonable to me, although I would probably do something like `AUTH_ERRORS.push(*NET_ERRORS.map{|e| SharedHelpers.get_error(e) }.compact)`

segiddins:
@RochesterinNYC oooh yeah I wasn't even thinking of something unit-testable like that! That's perfect :)

RochesterinNYC:
Updated the PR with the extracted `SharedHelpers#const_get_safely` and a unit test for it.

Additionally, I noticed that most or all of the methods in the `SharedHelpers` module don't seem to have unit tests for them and rely on their usage in other modules/methods and the tests for those other modules/methods for test coverage. Would it be worth making a separate PR to add unit tests for those `SharedHelpers` methods?

segiddins:
@RochesterinNYC unit tests 😍

segiddins:
@homu r+

indirect:
Unit tests would be amazing! We want to transition to primarily having unit tests over time, because right now we have only integration tests and it is kinda awful.

RochesterinNYC:
Hm, failure on `Ruby 1.8.7` was due to `Module#constants` returning an array of strings rather than symbols (behavior in all later versions of ruby). Refactored `SharedHelpers#const_get_safely` to check for string or symbol form of the constant.

segiddins:
@homu r+

PR #4196, opened 2016-01-02
Make `bundle env` output consistent

agis-:
We don't use ":" anywhere else on `bundle env` output.


indirect:
there's a focused: true in there :)

agis-:
Oh, it's been a while :smile: 

Fixed!

indirect:
@homu r+

PR #4201, opened 2016-01-07
Fix returning case-sensitive mirror URIs in Ruby 2.3

segiddins:
I have no idea why this fixes things locally, but it does. ¯\\_(ツ)_/¯


indirect:
@homu r+

PR #4206, opened 2016-01-08
Remove `Bundler::SharedHelpers#in_bundle?` and make `Bundler::SharedHelpers#find_gemfile` public

RochesterinNYC:
- `Bundler::SharedHelpers#in_bundle?` is currently just a wrapper around solely `Bundler::SharedHelpers#find_gemfile`
  so the two are functionally equivalent
- Modify `Bundler::SharedHelpers#default_gemfile` unit tests to use
  mocking for `find_gemfile`

### DIFF
--- a/lib/bundler/setup.rb
+++ b/lib/bundler/setup.rb
@@ -1,6 +1,6 @@
 require "bundler/shared_helpers"
 
-if Bundler::SharedHelpers.in_bundle?
+if Bundler::SharedHelpers.find_gemfile
   require "bundler"
 
   if STDOUT.tty? || ENV["BUNDLER_FORCE_TTY"]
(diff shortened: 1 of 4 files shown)


RochesterinNYC:
Seems like `Bundler::SharedHelpers#in_bundle?` used to have more functionality (like exception catching) but after some modification/refactoring, it's just a straight call through to `find_gemfile` now.

RochesterinNYC:
After some searching on Github, looks like there's repos/stuff that depends on this public method (`in_bundle?`) so I guess we have to keep it to prevent breakage. 

I guess for the unit test for this method, I can just mock and check that it calls `find_gemfile`.

segiddins:
I'm 👎🏻 on this change -- it means we can't add in any other meaning to 'being in a bundle'

RochesterinNYC:
Gotcha, also a valid point. Closing.

PR #4207, opened 2016-01-09
Separate concerns of `Bundler::SharedHelpers#set_bundle_environment`

RochesterinNYC:
- Splits the implementation details/sections of `Bundler::SharedHelpers#set_bundle_environment` into three more modular, singularly responsible methods: `set_path`, `set_rubyopt`, `set_rubylib`
- Adds unit tests for `set_bundle_environment` and the new methods

### DIFF
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -74,22 +74,9 @@ module Bundler
     end
 
     def set_bundle_environment
-      # Set PATH
-      paths = (ENV["PATH"] || "").split(File::PATH_SEPARATOR)
-      paths.unshift "#{Bundler.bundle_path}/bin"
-      ENV["PATH"] = paths.uniq.join(File::PATH_SEPARATOR)
-
-      # Set RUBYOPT
-      rubyopt = [ENV["RUBYOPT"]].compact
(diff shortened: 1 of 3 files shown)


RochesterinNYC:
Not 100% sure whether `set_path`, `set_rubyopt`, `set_rubylib` should be private or public. Made them public as of right now since `set_bundle_environment` was essentially a wrapper around the three.

segiddins:
Private imho

RochesterinNYC:
Hm, okay, so then no unit tests for those methods? 

And then we would keep the `set_bundle_environment` mocked call unit spec or create/make a unit test that aggregates all the functional unit testing for the set methods?

segiddins:
@RochesterinNYC yup!

segiddins:
Some rubocop offenses in the specs: https://travis-ci.org/bundler/bundler/jobs/101209168#L264-L277

RochesterinNYC:
Updated the PR with making the `set_` methods private. Also added `ENV['RUBYLIB']` and `ENV['RUBYOPT']` to the after spec reset list to prevent the new `set_bundle_environment` unit tests from having side effects.

segiddins:
@homu r+

PR #4208, opened 2016-01-10
Finish adding unit tests for `Bundler::SharedHelpers` module

RochesterinNYC:
If this PR and #4207 are accepted/merged, then the `Bundler::SharedHelpers` module will have all its public methods unit tested. The only thing that's missing coverage in the spec file for `Bundler::SharedHelpers` is `clean_load_path`, which is used in the `Bundler::Runtime` class. This'll be hopefully addressed/covered as I look more into increasing reasonable unit test coverage to near 100%.

PR #4202, opened 2016-01-07
Start adding unit tests for SharedHelpers

RochesterinNYC:
Backfilling some unit tests for the `Bundler::SharedHelpers` module. This PR doesn't add all the unit tests yet as I'd like some feedback on whether I'm on the right track regarding the level of mocking out and the interaction of some of the testee methods with private methods.

Related to discussion at #4195 regarding unit tests.

PR #4209, opened 2016-01-11
Refactor and add unit tests for `Bundler::RubyVersion`

RochesterinNYC:
- Add extensive unit test coverage for `lib/bundler/ruby_version.rb`
- Refactors `Bundler::RubyVersionRequirement` as a subclass of `Bundler::RubyVersion` (eliminates the odd, misdirecting code involving input_engine) and redundant methods.


RochesterinNYC:
Feedback on this would be appreciated as the formatting, level of mocking/stubbing, use of shared examples, general unit test design on this is how I'd aim to proceed with backfilling further unit tests.

Moreso than PR #4202 and #4208, as I realize now that I eschewed line break formatting for easier readability, which is present in this PR.

spec/bundler/ruby_version_spec.rb
```diff
@@ -1,80 +1,390 @@
...
+      let(:other_patchlevel)     { patchlevel }
+      let(:other_engine)         { engine }
+      let(:other_engine_version) { engine_version }
+      let(:other_ruby_version)   { Bundler::RubyVersion.new(other_version, other_patchlevel, other_engine, other_engine_version) }
 
-    expect(req.diff(ver)).to eq([:version, "2.0.0", "2.0.1"])
-  end
+      shared_examples_for "two ruby versions are not equal" do
+        it "should return false" do
+          expect(subject == other_ruby_version).to be false
```
segiddins: expect(subject).not_to eq(other_ruby_version)`

RochesterinNYC: Ah, RSpec `eq` matcher uses `==` under the hood? Learned something new :+1:

spec/bundler/ruby_version_spec.rb
```diff
@@ -1,80 +1,390 @@
...
-  it "successfully matches gem requirements" do
-    req = requirement(">= 2.0.0", "< 643", "ruby", "~> 2.0.1")
-    ver = version("2.0.0", "642", "ruby", "2.0.5")
+  describe Bundler::SystemRubyVersion do
+    subject { Bundler::SystemRubyVersion.new(version, patchlevel, engine, engine_version) }
+
+    let(:system_ruby_version) { subject }
+
+    describe "#initialize" do
+      # Suppress "warning: instance variable @variable not initialized" warnings
```
segiddins: why not fix those warnings?

spec/bundler/ruby_version_spec.rb
```diff
@@ -1,80 +1,390 @@
...
+        expect(system_ruby_version.instance_variable_get(:@patchlevel)).to be_nil
+        expect(system_ruby_version.instance_variable_get(:@engine)).to be_nil
+        expect(system_ruby_version.instance_variable_get(:@engine_version)).to be_nil
+      end
+    end
+
+    describe "#version" do
+      it "should return a copy of the value of RUBY_VERSION" do
+        expect(subject.version).to eq(RUBY_VERSION)
+        expect(subject.version.object_id).to_not eq(RUBY_VERSION.object_id)
```
segiddins: `expect(subject.version).to_not be(RUBY_VERSION)`

spec/bundler/ruby_version_spec.rb
RochesterinNYC: Well, since the `initialize` method is empty and does not call the super constructor, those instance variables are inherently empty, which is what we're trying to assert. And calling `version`, 'engine`, etc. don't hit the instance variables, calling and trying to get them directly seems to be the most straightforward.

I did try to look for an RSpec is_defined matcher or something similar but couldn't find anything. This is further complicated by the fact that `defined?` seems to be a semantic operator and not a method. Any suggestions? I could see a possible fix as asserting instead that the super constructor is not being called but that seems a bit like testing implementation rather than the functional unit? Open to this though.

segiddins: ew. TBH, looking at that code, I'd kinda like to refactor it to just call through to the super initializer? It could even just be a factory method and we could probably get rid of the whole subclass

RochesterinNYC: I'd definitely be down for that. I don't feel that the whole concept of a `RubyVersionRequirement` class is necessary, especially when it's essentially just a `RubyVersion` instance that can be compared to other ones. I'd be down for moving the `diff` method and `matches?` private method up into `RubyVersion`, making the appropriate changes in the codebase for phasing out `RubyVersionRequirement`. RubyVersions can be compared to each other so I don't exactly see the point in the `RubyVersionRequirement` class and having it solely contain the `diff` method?

segiddins: We can also get rid of `SystemRubyVersion` which would make things less complicated imho

spec/bundler/ruby_version_spec.rb
RochesterinNYC: Hm, `SystemRubyVersion` seems to have a lot of use/reference I'm external repos? `RubyVersionRequirement` seems only to be used by Bundler itself.

segiddins: It's an implementation detail

RochesterinNYC: Sorry, posted this in the wrong section.

spec/bundler/ruby_version_spec.rb
RochesterinNYC: Hm, SystemRubyVersion seems to have a lot of use/reference in external repos?
https://github.com/search?p=5&q=SystemRubyVersion&ref=searchresults&type=Code&utf8=%E2%9C%93

RubyVersionRequirement seems only to be used by Bundler itself so removing it won't break anything using Bundler.

segiddins:
@RochesterinNYC none of those files point to people actually using SystemRubyVersion

RochesterinNYC:
@segiddins Ah okay, I understand what I'm seeing now. It's cached/in-repo copies of old Bundler versions that have the Bundler code with SystemRubyVersion. Gotcha.

RochesterinNYC:
Refactored with the removal of `SystemRubyVersion` and `RubyVersionRequirement`.

lib/bundler.rb
```diff
@@ -374,6 +372,28 @@ def reset!
 
   private
 
+    def ruby_engine
```
segiddins: I'd prefer these to not be in their own methods...

spec/bundler/bundler_spec.rb
```diff
@@ -3,6 +3,111 @@
...
+    it "should return an instance of Bundler::RubyVersion" do
+      expect(subject).to be_kind_of(Bundler::RubyVersion)
+    end
+
+    it "memoizes the instance of Bundler::RubyVersion" do
+      expect(Bundler::RubyVersion).to receive(:new).once.and_call_original
+      2.times { Bundler.ruby_version }
+    end
+
+    it "should return an instance of Bundler::RubyVersion" do
```
segiddins: this is a duplicate

segiddins:
:+1: 
@indirect r?

lib/bundler.rb
```diff
@@ -365,7 +363,23 @@ def git_present?
...
+                            when "ruby"
+                              RUBY_VERSION.dup
+                            when "rbx"
+                              Rubinius::VERSION.dup
+                            when "jruby"
+                              JRUBY_VERSION.dup
+                            else
+                              raise BundlerError, "RUBY_ENGINE value #{RUBY_ENGINE} is not recognized"
+      end
+      @ruby_version ||= RubyVersion.new(RUBY_VERSION.dup, RUBY_PATCHLEVEL.to_s, ruby_engine, ruby_engine_version)
```
indirect: I would dearly love to start ripping everything out of the `Bundler` god-module, so can we move `Bundler.ruby_version` over to `Bundler::RubyVersion.current` or maybe `.system`?

RochesterinNYC: I'm appreciative of that direction.

So a class method that creates and returns a new or memoized instance of `Bundler::RubyVersion` that's built with system specific params.

indirect: Yup! With the bonus of there being less Bundler.whatever methods. :)

RochesterinNYC:
Updated the PR with the moving of the system functionality to `Bundler::RubyVersion`.

segiddins:
Absolutely excellent work, @RochesterinNYC !
@homu r+

RochesterinNYC:
Thanks for all the review/critique! Always happy to contribute to Bundler.

PR #4210, opened 2016-01-11
add spec verifying that #4147 is fixed

indirect:
See #4147 
this test fails when run against the tag v1.11.0, and passes when run against v1.11.2.

### DIFF
--- a/spec/install/gemfile/ruby_spec.rb
+++ b/spec/install/gemfile/ruby_spec.rb
@@ -1,0 +1,24 @@
+require "spec_helper"
+
+describe "ruby requirement" do
+  # As discovered by https://github.com/bundler/bundler/issues/4147, there is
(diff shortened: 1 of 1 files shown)


indirect:
@segiddins r?

spec/install/gemfile/ruby_spec.rb
```diff
@@ -0,0 +1,19 @@
+require "spec_helper"
+
+describe "ruby requirement" do
```
segiddins: I'm not seeing any ruby requirement in here?

segiddins:
@indirect please make it more clear in the spec what this is actually testing

indirect:
@segiddins done. also the test actually covers the bug in question now that I have added ruby requirements to the test. :stuck_out_tongue_closed_eyes:

segiddins:
@homu r+

PR #4211, opened 2016-01-11
Fix self dependency of gems while installing parallely.

morsheed:
Thanks to @schneems for reporting and fixing the problem.We now detect
if gem has a dependency with same name and proceed with installation if
that is the case and that is the only dependency left.

### DIFF
--- a/lib/bundler/installer.rb
+++ b/lib/bundler/installer.rb
@@ -278,7 +278,6 @@ module Bundler
         name2spec[spec.name] = spec
         remains[spec.name] = true
       end
-
       worker_pool = ParallelWorkers.worker_pool size, lambda { |name, worker|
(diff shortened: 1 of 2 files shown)


segiddins:
Given this commit is almost two years old, and it doesn't merge cleanly... Did you randomly just find this patch?

indirect:
We already previously rejected patches to allow installing gems with circular dependencies. If a gem has a dependency, it might be impossible to install that gem until all its dependencies are installed. If there is a circular dependency, it makes it impossible to know which gem needs to be installed first, so it's impossible to install the gems in the correct order.

This isn't a fix, this is deliberately introducing a bug in order to allow installing certain gems with bad dependencies. :(

PR #4217, opened 2016-01-17
Extract kernel require, so we can extend it

arthurnn:
I find myself wanting to instrument this require calls, however to patch this code in bundler
is not super easy as Bundler.require is a big method. We can extract the Kernel.require to a method
and make it easier to extend it.

review @indirect 
Thoughts? Maybe you might have a better idea to accomplish what I need. let me know

thanks

### DIFF
--- a/lib/bundler/runtime.rb
+++ b/lib/bundler/runtime.rb
@@ -74,7 +74,7 @@ module Bundler
             file = dep.name if file == true
             required_file = file
             begin
-              Kernel.require file
+              kernel_require(file)
             rescue => e
               raise e if e.is_a?(LoadError) # we handle this a little later
               raise Bundler::GemRequireError.new e,
@@ -96,6 +96,10 @@ module Bundler
           end
         end
       end
+    end
+
+    def kernel_require(file)
+      Kernel.require(file)
     end
 
     def dependencies_for(*groups)


segiddins:
Why not just instrument `Kernel.require` directly?

indirect:
Instrumenting just this usage of Kernel.require gives you an instant gem-level overview of how long requiring takes, so I can understand wanting it. I remember a long ago PR to add a `bundle profile` that printed timing info for every gem in your bundle. @arthurnn, are you gong to release your work as `bundle-profile` so everyone can use it? :D

arthurnn:
bundle-profile,that sounds like a good idea.. i am definitely interested . i am working in some optimizations on rubygems.org, so i patched this locally, but i will see what else i could push to a profile command. that would be awesome.

indirect:
@arthurnn I'll accept this if it extracts all of [lines 66-98](https://github.com/arthurnn/bundler/blob/arthurnn/extract_require/lib/bundler/runtime.rb#L66-L98) into a method named `require_dependency` or something like that... then it's both a genuine improvement to the code and lets you instrument each gem getting required. :)

segiddins:
@arthurnn can we close this until you've addressed @indirect's comments?

PR #4218, opened 2016-01-19
Fixed quotes in newgem.gemspec.tt

MOZGIII:
I always end up fixing quotes manually. Should it be like that by default, or is it not convenient?

### DIFF
--- a/lib/bundler/templates/newgem/newgem.gemspec.tt
+++ b/lib/bundler/templates/newgem/newgem.gemspec.tt
@@ -13,31 +13,31 @@ Gem::Specification.new do |spec|
   spec.description   = %q{TODO: Write a longer description or delete this line.}
   spec.homepage      = "TODO: Put your gem's website or public repo URL here."
 <%- if config[:mit] -%>
-  spec.license       = "MIT"
+  spec.license       = 'MIT'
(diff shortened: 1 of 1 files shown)


segiddins:
What is this fixing, exactly?

MOZGIII:
It fixes unnecessary use of double quotes by replacing them with single quotes in `.gemspec` template.

RochesterinNYC:
The current rubocop configuration enforces double quotes.

indirect:
https://viget.com/extend/just-use-double-quoted-ruby-strings

MOZGIII:
Provide me a way to choose then. Don't decide what I need to use for me.

indirect:
@MOZGIII the Bundler gem generator is designed to be extremely minimal—the only options are for testing (rspec or minitest) and for license and code of conduct. If you want a gem generator with more options, I hear that Ore is pretty good: https://github.com/ruby-ore/ore

MOZGIII:
@indirect I see, probably that would be the way to go for me.
However, `bundler` seems to be a "default" tool to create gems nowadays. I guess we should discuss things at least and try to fix whatever we can, so #4220.

PR #4221, opened 2016-01-21
Booleanize the disable_shared_gems option

vassilevsky:
Hi 👻

I thought that the last value in the config looks a bit strange:

```
$ cat .bundle/config
---
BUNDLE_PATH: vendor/bundle
BUNDLE_JOBS: 4
BUNDLE_CLEAN: true
BUNDLE_DISABLE_SHARED_GEMS: '1'
```

So I have booleanized it.

### DIFF
--- a/lib/bundler/cli/check.rb
+++ b/lib/bundler/cli/check.rb
@@ -1,6 +1,7 @@
 module Bundler
   class CLI::Check
     attr_reader :options
+
     def initialize(options)
       @options = options
     end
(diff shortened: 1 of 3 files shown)


RochesterinNYC:
Are there any specs around this?

vassilevsky:
I was hoping to look at Travis results when it's done and find out 😂

RochesterinNYC:
Hahah gotcha. Not ideal to use Travis as a dev test suite runner but I do the same thing sometimes. Side effect of incomplete unit test coverage and a load of extensive integration/feature tests really.

segiddins:
Can you please add a test and make sure `disable_shared_gems` is on the list of Boolean keys in `settings.rb`?

vassilevsky:
Wow. I had no idea there is such a list. I have added the option there.

I am lost as to where the tests should be added. What to test?

segiddins:
just grepped the existing specs, so I actually think you're good

@homu r+

PR #4225, opened 2016-01-23
Use 'bundle gem --exe' as primary option name, in addition to '--bin'.

christhekeele:
This furthers the v1.8 transition of conventionally using the `exe/` folder for bins, leaving the `bin/` folder for binstubs, by allowing the flag that auto-generates a starter file in the `exe/` folder to logically be referred to as `--exe`.

This is a follow-up on the discussion at bundler/bundler-features#105.


segiddins:
:+1: 
@indirect r?

indirect:
@homu r+

PR #4229, opened 2016-01-24
Configure Code Climate exclusions

RochesterinNYC:
This PR configures file exclusions for Code Climate test coverage analysis (which files we don't care about "coverage" on) and code analysis.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,24 @@
+---
+engines:
+  duplication:
+    enabled: true
+    config:
+      languages:
+      - ruby
+  fixme:
+    enabled: true
+  rubocop:
+    enabled: false
+ratings:
+  paths:
+  - "**.rb"
+  - "bin/*"
+  - "exe/*"
+exclude_paths:
+- bundler.gemspec
+- "*.md"
+- lib/bundler/ssl_certs/*.pem
+- lib/bundler/vendor/**/*
+- lib/bundler/templates/**/*.tt
+- man/*
+- spec/**/*
--- a/spec/support/code_climate.rb
+++ b/spec/support/code_climate.rb
@@ -3,9 +3,22 @@ module Spec
     def self.setup
       require "codeclimate-test-reporter"
       ::CodeClimate::TestReporter.start
+      configure_exclusions
     rescue LoadError
       # it's fine if CodeClimate isn't set up
       nil
+    end
+
+    def self.configure_exclusions
+      SimpleCov.start do
+        add_filter "/bin/"
+        add_filter "/lib/bundler/man/"
+        add_filter "/lib/bundler/vendor/"
+        add_filter "/man/"
+        add_filter "/pkg/"
+        add_filter "/spec/"
+        add_filter "/tmp/"
+      end
     end
   end
 end


RochesterinNYC:
Do we want to disable their rubocop analyzer in the `.codeclimate.yml`? I'm open to this.

I'm assuming they don't use the `.rubocop.yml` we have in the repo as otherwise there would be no style warnings presumably.

.codeclimate.yml
```diff
@@ -0,0 +1,24 @@
...
+  - "exe/*"
+exclude_paths:
+- bundler.gemspec
+- "*.md"
+- lib/bundler/ssl_certs/*.pem
+- lib/bundler/vendor/**/*
+- lib/bundler/templates/**/*.tt
+- man/*
+- spec/**/*
\ No newline at end of file
```
segiddins: Trailing newline

segiddins:
That's strange, since the rubocop errors aren't showing up for master?

RochesterinNYC:
@segiddins It looks like the lack of rubocop errors on master (either locally or via Travis) are because it seems that Code Climate doesn't take into consideration the `.rubocop_todo.yml` file correctly? When we run rubocop locally or on Travis, the violations that Code Climate detects (https://codeclimate.com/github/bundler/bundler/pull/4229) are not there/detected because they are listed in the `.rubocop_todo.yml`. 

For example, the `Lint/UselessAssignment` violation in `lib/bundler/graph.rb` is suppressed on local or Travis rubocop runs by the following configuration in the `.rubocop_todo.yml` file:
```
# Offense count: 7
Lint/UselessAssignment:
  Exclude:
    - 'lib/bundler/graph.rb'
    - 'lib/bundler/index.rb'
    - 'lib/bundler/installer.rb'
```

RochesterinNYC:
Looking at the rubocop exclusions, a lot of them are unavoidable at the moment or require more extensive code rewrites. Ex. global variable stuff, no generic Exception rescuing, etc.

RochesterinNYC:
@segiddins so the issue is caused by CodeClimate not currently supporting the `inherit_from` command as per: https://buildpacks.ci.cf-app.com/pipelines/nodejs-buildpack/jobs/buildpack-CF-LTS/builds/45 So the exclusion `.rubocop.todo.yml` that we call `inherit_from` on in our `.rubocop.yml` is not considered in Code Climate's rubocop analysis.

One quick work around this might be to use the ability to specify a custom/different rubocop.yml file (one that's not named `.rubocop.yml` for the CodeClimate rubocop engine. We would create a `.code_climate_rubocop.yml` or something of that sort that includes the content from both our `.rubocop.todo.yml` and `.rubocop.yml`. Theoretically, since this `.code_climate_rubocop.yml` wouldn't need to use `inherit_from`, Code Climate's rubocop engine should process it (and our rubocop exclusions) properly. Thoughts?

segiddins:
Ugh. In that case, lets nix rubocop from code climate? I really don't want to maintain duplicate state. Do they document that limitation anywhere?

RochesterinNYC:
Oops I completely posted the wrong link up there, hahah. Should be: https://github.com/codeclimate/codeclimate-rubocop/issues/39#issuecomment-169090184

Don't believe it's documented in their official docs. Had to look through for that comment on that issue and then clarify with support.

Okay, I'll edit the PR to nix Rubocop later.

RochesterinNYC:
Updated the PR to turn off the Rubocop engine.

indirect:
Alternatively, we could merge our todo file into our main file and indicate which entries are planned to be removed via a comment?

segiddins:
@indirect but then we'd lose the ability to auto-gen the config, and also the ability to use that config file across different repos

indirect:
Alright :( I was hoping for RuboCop as a separate commit status, but I guess that's too hard for computers to accomplish :P

segiddins:
Ideally, Travis would let us segment the different jobs, but ¯\_(ツ)_/¯

RochesterinNYC:
Yeah I think we can just disable the Rubocop engine in the `.codeclimate.yml` file for now. It's seemingly in Code Climate's plans to eventually support the `inherit_from` directive so we can enable it when that occurs.

segiddins:
@homu r+

PR #4236, opened 2016-01-25
[Hax] Support simulating jruby on MRI 2.3

### LINKED ISSUE #4235: Get specs passing on 2.3
Specifically, when we stub Ruby, we'll need to add a hack in to load YAML first

segiddins:
Closes #4235.


segiddins:
@homu r+

PR #4240, opened 2016-01-26
Require 'irb/completion' in bin/console template

Zorbash:
Effortlessly providing completion in the generated `bin/console` seems like a sane default.

Rails has followed this way, at least [since 2009](https://github.com/rails/rails/commit/f0dd77c#diff-cd8bfebc5a193c9a13a282e9be10e180R14) and 'irb/completion' is present in the standard library for all Ruby versions we support.

What do you think?


segiddins:
=> @indirect

indirect:
Yeah, this seems fine... I personally change my consoles to run Pry, but if you have to use IRB, it's better with completion.

PR #4245, opened 2016-01-27
Allow passing UI object into bundler/inline

lamont-granquist:
Added an argument to the gemfile method to allow overriding the
UI object.  Useful to override the default output format and
send those log events to a logfile or decorate them (timestamps,
severity, etc) before emitting them.

### DIFF
--- a/lib/bundler/inline.rb
+++ b/lib/bundler/inline.rb
@@ -27,7 +27,9 @@
 #
 #          puts Pod::VERSION # => "0.34.4"
 #
-def gemfile(install = false, ui = nil, &gemfile)
+def gemfile(install = false, options = {}, &gemfile)
+  ui = options[:ui]
+
   require "bundler"
   old_root = Bundler.method(:root)
   def Bundler.root
--- a/lib/bundler/inline.rb
+++ b/lib/bundler/inline.rb
@@ -28,9 +28,12 @@
 #          puts Pod::VERSION # => "0.34.4"
 #
 def gemfile(install = false, options = {}, &gemfile)
-  ui = options[:ui]
-
   require "bundler"
+
+  opts = options.dup
+  ui = opts.delete(:ui) { Bundler::UI::Shell.new }
+  raise ArgumentError, "Unknown options: #{opts.keys.join(", ")}" unless opts.empty?
+
   old_root = Bundler.method(:root)
   def Bundler.root
     Bundler::SharedHelpers.pwd.expand_path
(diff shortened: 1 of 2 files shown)


segiddins:
I really don't think this is necessarily the right approach -- maybe a options hash that allows changing `silent` from `true` to `false`?

lamont-granquist:
I don't want silent, though, I want my own logger.

I can certainly do an options hash instead.

segiddins:
I'm not really sold on the benefit of you're own logger though...?

lamont-granquist:
So I can use this in chef-client and not have messy formatting on the output for one thing.  Also, when chef-client runs as a daemon I want to capture the output and send it to the configured logfile (which is not necessarily STDOUT, particularly on Windows--runit/djb's logging philosophy is good, but its far from universal).  I don't want to suppress the output because the installer may be taking actions and not surfacing those actions to the user would be terrible experience.   I also want to be able to bump the verbosity up higher when the chef-client logger is in debug mode.

The alternative is copypasta'ing the bundler/inline code directly into chef and maintaining it there in parallel, but I'd prefer to use a public API.

segiddins:
@indirect thoughts on taking an options hash that allows specifying a UI object?

indirect:
seems... ok, I guess?

lamont-granquist:
@indirect anything else you want?

lib/bundler/inline.rb
```diff
@@ -27,7 +27,9 @@
 #
 #          puts Pod::VERSION # => "0.34.4"
 #
-def gemfile(install = false, &gemfile)
+def gemfile(install = false, options = {}, &gemfile)
+  ui = options[:ui]
```
segiddins: should raise if there are any unknown options passed

segiddins: `options.fetch(:ui) { Bundler::UI::Shell.new }`

indirect:
Address @segiddins' comments, and I think this is good.

lamont-granquist:
@segiddins @indirect here you go.

there's a bonus fix to wrap the lazy require with concurrency protection.  now that i type it up and think about it, though, i'm not positive it makes any difference since i'm not sure it makes much sense to ever have multiple threads calling into bundler?  i can back that out...

lib/bundler/inline.rb
```diff
@@ -27,8 +27,13 @@
 #
 #          puts Pod::VERSION # => "0.34.4"
 #
-def gemfile(install = false, &gemfile)
-  require "bundler"
+def gemfile(install = false, options = {}, &gemfile)
+  Thread.exclusive { require "bundler" }
```
segiddins: Requiring is already tread-safe, and exclusive is deprecated.

spec/runtime/inline_spec.rb
```diff
@@ -93,4 +93,50 @@ def script(code, options = {})
...
+      end
+    RUBY
+
+    expect(out).to eq("CONFIRMED!")
+    expect(exitstatus).to be_zero if exitstatus
+  end
+
+  it "raises an exception if passed unknown arguments" do
+    script <<-RUBY, :expect_err => true
+      gemfile(true, arglebargle: true) do
```
segiddins: this fails on 1.8.7, please use hashrockets

lamont-granquist:
@indirect travis is green

segiddins:
@homu r+

Issue #3805, opened 2015-07-02
bundle outdated/update within patch, minor, major version 


ComLock:
I would like to be able to only show or update outdated gems where a newer version is within patch, or minor version. As in do the easy upgrades first.

```sh
bundle update --patch
```

A no-op flag to show what would be done, without doing it would also be nice.

```sh
bundle -n update --minor
```

Then you would not really need the outdated command :)

indirect:
This sounds lovely. Any interest in sending a pull request? :grin:

ComLock:
Sure, I'm going on vacation soon though, so it might be a while. In the mean time, anyone please feel free to make it... :grin:

agis-:
@indirect I'm interested in doing this. Any insights on how this should be approached?

indirect:
This is something that we haven't ever done before! I think what you will probably end up having to do is constructing Gem::Requirement objects with ~> so you can limit the minor or tiny versions that the resolver will look for. I imagine that --minor would end up iterating over all the gems and adding ~> X.Y constraints to each gem in the bundle before trying to resolve. I'm also not sure that would work, but it's where I would start. :)

trusche:
I've been wanting this as well for a long time. @agis let me know if I can help please.

agis-:
I won't have the time to do this so please feel free to take a shot!

bronzdoc:
Closed  by https://github.com/bundler/bundler/pull/4249

PR #4249, opened 2016-01-28
Bundle outdated with major and minor options

cirdes:
This feature adds the ability to specify a outdated command to show only the gems with a major or minor updates.

I was looking for this feature and found the issue https://github.com/bundler/bundler/issues/3805

@indirect, how can I improve this PR?

### DIFF
--- a/lib/bundler/cli/outdated.rb
+++ b/lib/bundler/cli/outdated.rb
@@ -46,6 +46,18 @@ module Bundler
             active_spec = active_spec.delete_if {|b| b.respond_to?(:version) && b.version.prerelease? }
           end
           active_spec = active_spec.last
+
+          if options[:major]
+            current_major = current_spec.version.segments.first
+            active_major = active_spec.version.segments.first
+            active_spec = nil unless active_major > current_major
+          end
+
+          if options[:minor]
+            current_minor = current_spec.version.segments[0, 2].compact.join(".")
+            active_minor = active_spec.version.segments[0, 2].compact.join(".")
+            active_spec = nil unless active_minor > current_minor
+          end
         end
         next if active_spec.nil?
 
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -242,6 +242,8 @@ module Bundler
     method_option "source", :type => :array, :banner => "Check against a specific source"
     method_option "strict", :type => :boolean, :banner =>
       "Only list newer versions allowed by your Gemfile requirements"
+    method_option "major", :type => :boolean, :banner => "Only list major newer versions"
+    method_option "minor", :type => :boolean, :banner => "Only list at least minor newer versions"
     def outdated(*gems)
       require "bundler/cli/outdated"
       Outdated.new(options, gems).run
--- a/spec/commands/outdated_spec.rb
+++ b/spec/commands/outdated_spec.rb
@@ -180,4 +180,59 @@ describe "bundle outdated" do
     bundle :outdated
     expect(out).to include("Installing foo 1.0")
   end
+
+  describe "with --major option" do
+    it "only reports gems that have a newer major version" do
+      update_repo2 do
+        build_gem "weakling", "0.2.0"
+        build_gem "activesupport", "3.0"
+      end
+
+      bundle "outdated --major"
+      expect(out).to_not include("weakling (newest")
+      expect(out).to include("activesupport (newest")
+    end
+
+    it "ignore gems not in semantic version" do
+      update_repo2 do
+        build_gem "weakling", "1"
+      end
+
+      bundle "outdated --major"
+      expect(out).to include("weakling (newest")
+    end
+  end
+
+  describe "with --minor option" do
+    it "only reports gems that have at least a newer minor version" do
+      update_repo2 do
+        build_gem "activesupport", "3.0.0"
+        build_gem "weakling", "0.2.0"
+      end
+
+      bundle "outdated --minor"
+      expect(out).to include("weakling (newest")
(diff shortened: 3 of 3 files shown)


lib/bundler/cli/outdated.rb
```diff
@@ -46,6 +46,22 @@ def run
             active_spec = active_spec.delete_if {|b| b.respond_to?(:version) && b.version.prerelease? }
           end
           active_spec = active_spec.last
+
+          if options[:major]
+            current_major = Gem::Version.new(current_spec.version.to_s.split(".").first)
```
segiddins: `current_spec.version.segments.first` ?

cirdes: Thanks!

lib/bundler/cli/outdated.rb
```diff
@@ -82,5 +94,12 @@ def run
         exit 1
       end
     end
+
+  private
+
+    def minor_trim(spec)
+      segments = spec.version.segments
```
segiddins: `spec.version.segments[0, 2].compact.join(".")` ?

lib/bundler/cli/outdated.rb
```diff
@@ -82,5 +94,12 @@ def run
         exit 1
       end
     end
+
+  private
+
+    def minor_trim(spec)
```
segiddins: please make this method name more informative

lib/bundler/cli/outdated.rb
```diff
@@ -46,6 +46,18 @@ def run
             active_spec = active_spec.delete_if {|b| b.respond_to?(:version) && b.version.prerelease? }
           end
           active_spec = active_spec.last
+
+          if options[:major]
+            current_major = Gem::Version.new(current_spec.version.segments.first)
+            active_major = Gem::Version.new(active_spec.version.segments.first)
+            active_spec = nil unless active_major > current_major
```
segiddins: no need to wrap in a `Gem::Version` object for this comparison

lib/bundler/cli/outdated.rb
cirdes: spec.version.segments[0, 2].join(".") ?

segiddins: you'd need to `compact`

cirdes: I'd tested with one segment and after segments method it returned an array with only one element(no nil element), but I'll add the compact just to be sure.

segiddins:
@indirect please review this

segiddins:
Actually, @cirdes, it would be awesome if you could add the version numbers to the specs, to decrease the likelihood of us breaking something accidentally in the future

indirect:
This seems genuinely useful. Thanks!

@homu r+

contentfree:
It would be great to add a `--patch` (or `--tiny`) flag, too. Patch level updates should be the safest to make, yet with this change we can only see the list of *not the safest* updates (minor and major upgrades).

indirect:
@contentfree when you run `outdated` with no arguments, you should be seeing the patch level updates, because you use pessimistic version constraints, right? :grin:

contentfree:
Hmm... Gemfile fail on my part then ;)

indirect:
That said, you raise a good point... it looks like #3805 actually expected the options `--patch` and `--minor`, with "no arguments" providing all possible updates. This pull request seems to implement the opposite, `--major` showing only major updates, and `--minor` showing only minor updates.

This PR provides "only major", "only minor", and "major, minor, and patch" (on no flags), but I think it would actually be more useful to provide "only patch" and "only patch and minor" as options, because no arguments already provides "patch, minor, and major". @RochesterinNYC what do you think about this?

contentfree:
@indirect: Exactly. I do use the pessimistic operator (most of the time…) but `"~> 1.1"` would show minor outdated with no flags.

contentfree:
I think having all three options would be ideal. Even if you were looking for major upgrades, sifting through the list of *all* upgrades available is worse than just seeing the exact list of major upgrades.

RochesterinNYC:
I thought we already added a `--patch` flag that will only show updates in the third, patch version section here? #4312

cirdes:
@indirect, this PR(#4249) provides "only major"(or at least major) for --major and "major and minor"(at least minor) for --minor. 

The PR #4312 change the description from "at least" to "Only list".

The reason I didn't add the "--patch" is because it would be "major, minor and patch"(at least patch) that is equivalent to outdated without arguments

indirect:
@cirdes I guess it's not clear to me whether whether `--minor` means "minor and major" or "minor and patch" or "minor only". I didn't really think about the options before accepting this PR, so I'm not sure if what we have now is the best choice. (Maybe it is! But I didn't think about it until now.)

@contentfree as @RochesterinNYC pointed out, we added a `--patch` flag! Go to town. :)

cirdes:
@indirect, you are right. It was not clear at all. I did that way because when I`m updating my gems using bundle outdated I just have to manually change the version of minor and major updates. So using bundle outdated --minor (at least) give me just what I want. I not sure if my approach is the best for everyone else but was based on my user case.

RochesterinNYC:
If you're looking for updates that only affect the minor and patch version sections, I think that `bundle outdated --minor --patch` might accomplish this. Please correct me if I'm wrong here.

cirdes:
@RochesterinNYC, I think you are right. It seems to be correct and clear.

indirect:
I agree, that seems to make sense. Each flag includes only that type of update, they can be combined, and no flags includes all types.

Issue #4248, opened 2016-01-28
Confusing error message when local gems have invalid gemspec


dbslone:
First I want to say :+1: for enforcing gemspecs to be valid before bundler will install the gems.

I just upgraded to ruby 2.3.0 and installed the latest version of bundler and when I ran `bundle install` I received the following message:

```
$ bundle install
The gemspec at /lib/rails_admin_devise_invitable/rails_admin_devise_invitable.gemspec is not valid. The validation error was '"FIXME" or "TODO" is not an author'
The gemspec at /lib/rails_admin_devise_invitable/rails_admin_devise_invitable.gemspec is not valid. The validation error was '"FIXME" or "TODO" is not an author'
Fetching gem metadata from https://rubygems.org/.......
Fetching version metadata from https://rubygems.org/...
Fetching dependency metadata from https://rubygems.org/..
Could not find gem 'rails_admin_devise_invitable' in source at `lib/rails_admin_devise_invitable`.
Source does not contain any versions of 'rails_admin_devise_invitable'
```

The error message is confusing because I thought it was unable to find the gem once I corrected the errors in the gemspec everything worked as expected.

My suggestion would be that the error message should inform the user to correct the specified gemspec file(s) then run bundle install again.

TimMoore:
Thanks for the report, @dbslone. Are you able to share the failing project with us to help reproduce the issue? Is it open source?

dbslone:
@TimMoore Unfortunately the app isn't open source but I should have some time over the weekend to create new rails app that reproduces the error that I can share

RochesterinNYC:
@dbslone what would you have expected error message wise ? Something along the lines of:

```
$ bundle install

Gem::InvalidSpecificationException: The gemspec at /lib/rails_admin_devise_invitable/rails_admin_devise_invitable.gemspec is not valid. Please fix the gemspec.
```

dbslone:
@RochesterinNYC Yeah I think that error message is clearer and would inform users that once the gemspec is corrected operations should proceed as expected.

segiddins:
Closed by https://github.com/bundler/bundler/pull/4250.

PR #4250, opened 2016-01-29
Change behavior and improve clarity of error messages when local gems have invalid gemspecs

RochesterinNYC:
Now, when a local gem's gemspec contains errors, an error is raised immediately and helpful message is provided. This message instructs the user to fix the gemspec before an attempt is made to fetch the local gem that has an incorrect gemspec.

- Addresses #4248

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -352,9 +352,9 @@ module Bundler
         spec
       end
     rescue Gem::InvalidSpecificationException => e
-      UI::Shell.new.warn "The gemspec at #{file} is not valid. " \
-        "The validation error was '#{e.message}'"
-      nil
+      error_message = "The gemspec at #{file} is not valid. Please fix this gemspec.\n" \
+        "The validation error was '#{e.message}'\n"
+      raise Gem::InvalidSpecificationException.new(error_message)
     end
 
     def clear_gemspec_cache
--- a/spec/install/gemfile/path_spec.rb
+++ b/spec/install/gemfile/path_spec.rb
@@ -157,8 +157,11 @@ describe "bundle install with explicit source paths" do
       gem "foo", :path => "#{lib_path("foo-1.0")}"
     G
 
-    expect(out).to match(/missing value for attribute version/)
     expect(out).to_not include("ERROR REPORT")
+    expect(out).to_not include("Your Gemfile has no gem server sources.")
+    expect(out).to match(/is not valid. Please fix this gemspec./)
+    expect(out).to match(/The validation error was 'missing value for attribute version'/)
+    expect(out).to match(/You have one or more invalid gemspecs that need to be fixed/)
   end
 
   it "supports gemspec syntax" do
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -150,6 +150,9 @@ module Bundler
         WARN
       end
       raise e
+    rescue Gem::InvalidSpecificationException => e
+      Bundler.ui.warn "You have one or more invalid gemspecs that need to be fixed."
+      raise e
     end
 
   private
--- a/lib/bundler/friendly_errors.rb
+++ b/lib/bundler/friendly_errors.rb
@@ -33,6 +33,8 @@ module Bundler
       when Interrupt
         Bundler.ui.error "\nQuitting..."
         Bundler.ui.trace error
+      when Gem::InvalidSpecificationException
+        Bundler.ui.error error.message, :wrap => true
       when SystemExit
       else request_issue_report_for(error)
       end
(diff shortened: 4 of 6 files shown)


spec/install/gemfile/path_spec.rb
```diff
@@ -157,8 +157,7 @@
       gem "foo", :path => "#{lib_path("foo-1.0")}"
     G
 
-    expect(out).to match(/missing value for attribute version/)
-    expect(out).to_not include("ERROR REPORT")
```
segiddins: This should stay?

RochesterinNYC:
A note on the changes in behavior.
Where is the current error message that users see (as per #4248):
```
$ bundle install
The gemspec at /lib/rails_admin_devise_invitable/rails_admin_devise_invitable.gemspec is not valid. The validation error was '"FIXME" or "TODO" is not an author'
The gemspec at /lib/rails_admin_devise_invitable/rails_admin_devise_invitable.gemspec is not valid. The validation error was '"FIXME" or "TODO" is not an author'
Fetching gem metadata from https://rubygems.org/.......
Fetching version metadata from https://rubygems.org/...
Fetching dependency metadata from https://rubygems.org/..
Could not find gem 'rails_admin_devise_invitable' in source at `lib/rails_admin_devise_invitable`.
Source does not contain any versions of 'rails_admin_devise_invitable'
```

With this PR, the behavior/output changes to:
```
$ bundle install
The gemspec at /lib/rails_admin_devise_invitable/rails_admin_devise_invitable.gemspec is not valid. Please fix this gemspec.
The validation error was '"FIXME" or "TODO" is not an author'
The gemspec at /lib/rails_admin_devise_invitable/rails_admin_devise_invitable.gemspec is not valid. Please fix this gemspec.
The validation error was '"FIXME" or "TODO" is not an author'
You have one or more invalid gemspecs that need to be fixed.
Gem::InvalidSpecificationException
```

No fetching attempts or etc. would occur until the gemspecs are correct.

lib/bundler.rb
```diff
@@ -352,9 +352,9 @@ def load_gemspec_uncached(file, validate = false)
         spec
       end
     rescue Gem::InvalidSpecificationException => e
-      UI::Shell.new.warn "The gemspec at #{file} is not valid. " \
-        "The validation error was '#{e.message}'"
-      nil
+      UI::Shell.new.warn "The gemspec at #{file} is not valid. Please fix this gemspec.\n" \
+        "The validation error was '#{e.message}'\n"
+      raise Gem::InvalidSpecificationException
```
segiddins: this should be `raise` with no args?

segiddins:
:+1: other than that last comment

lib/bundler.rb
RochesterinNYC: So the reason that I essentially re-raise the error here is to wipe away the error message.

Otherwise, with raising the original error, the output is as follows:
```
       The gemspec at /Users/pivotal/workspace/bundler-rochester/tmp/libs/foo-1.0/foo.gemspec is not valid. Please fix this gemspec.
       The validation error was 'missing value for attribute version'
       You have one or more invalid gemspecs that need to be fixed.
       missing value for attribute version
```

instead of (with this new raised error):

```
       The gemspec at /Users/pivotal/workspace/bundler-rochester/tmp/libs/foo-1.0/foo.gemspec is not valid. Please fix this gemspec.
       The validation error was 'missing value for attribute version'
       You have one or more invalid gemspecs that need to be fixed.
       Gem::InvalidSpecificationException
```

The output might look even worse for multiple validation errors. Thoughts?

segiddins: hmmmm. Why do we warn here instead of raising with the full message in the first place?

RochesterinNYC: Well, the original implementation would return `nil` and would hence iterate through all local gemspecs and presumably report all the validation errors with all those local gemspecs presumably. By changing to raise an error here instead of `nil`, we lose that ability to report on all invalid local gemspecs at once (how big of a user scenario is this?) but prevent fetching or further post-gemspec loading actions from occurring. 

Though, as an alternative, raising a new `Gem::InvalidSpecificationException` in this location with the full error message (what is currently in warn), is functionally equivalent in terms of what the end user sees I believe.

RochesterinNYC:
```
    rescue Gem::InvalidSpecificationException => e
      error_message = "The gemspec at #{file} is not valid. Please fix this gemspec.\n" \
        "The validation error was '#{e.message}'\n"
      raise Gem::InvalidSpecificationException.new(error_message)
    end
```

segiddins: yeah

RochesterinNYC:
Updated the PR with the removal of the `UI::Shell.new.warn` call.

segiddins:
👍🏻
@indirect ?

indirect:
thanks!

@homu r+

Issue #4246, opened 2016-01-27
Can't begin Bundler development


nyarly:
Referring to DEVELOPMENT.md, I ran `rake spec:deps`, `rake spec` which resulted in 400+ spec fails. 

On advice from @segiddins I added the local bin directory to my `PATH`. 

I've added `bundler/lib` and `bundler/exe/bundle` to `RUBYLIB.`

Still no dice.

I'm on Linux, running MRI 2.1.7 at the moment. There's a warning about rspec-micks ~> 3.4.0 `Unresolved specs during Gem::Specification.reset:`, and the first fail is indeed `undefined method `allow'`

Lots of the fails report `/usr/lib/ruby/site_ruby/1.9.1/rubygems/dependency.rb:315:in `to_specs': Could not find 'git_tracker' (>= 0) among 0 total gem(s) (Gem::LoadError)` which is weird since, as noted, this is under MRI 2.1.7

nyarly:
No difference - although I wouldn't have expected one with ./bin in the PATH. Anyhow, here's the output: https://gist.github.com/nyarly/f0c7a25a113b820b5c70

segiddins:
ah. I think I know what this is. try `gem cleanup rspec-mocks` ?

nyarly:
That get's me a lot farther.

I still get the errors about `Could not find 'git_tracker' (>= 0) among 2 total gem` and I get stalls at "bundle install with gem sources / the simple case / creates a Gemfile.lock on a blank Gemfile"

segiddins:
I'm not sure what to tell you -- it works locally and on CI, and we do our best to create an isolated environment for the tests.

nyarly:
I'll see if I can figure out that detail - perhaps it's worth documenting.

segiddins:
Does a patch like this help for you?

```diff
diff --git a/bin/rspec b/bin/rspec
index 94e0d89..9614dde 100755
--- a/bin/rspec
+++ b/bin/rspec
@@ -7,4 +7,8 @@ bundler_spec.dependencies.each do |dep|
   gem dep.name, dep.requirement.to_s
 end
 
+Gem::Specification.unresolved_deps.each do |name, dep|
+  gem dep.name, dep.requirement.to_s
+end
+
 load Gem.bin_path("rspec-core", "rspec")
```

nyarly:
[...]

PR #4254, opened 2016-01-30
[Worker] Add a worker name to ease debugging

segiddins:
Also helps with https://github.com/bundler/bundler/issues/4246.

@indirect r?


indirect:
@homu r+

PR #4256, opened 2016-01-31
Preserve PATH in with_clean_env

### LINKED ISSUE #4251: bundle exec's "with_original_env" doesn't preserve PATH
I'm looking for a way to restore the original environment PATH inside a ruby script executed with `bundle exec <script>`. I understand `Bundler.with_original_env` ( and `Bundler.with_clean_env`)  should provide this functionality.

It seems to me though that the **PATH is not correctly preserved when running `bundle exec`**, because `lib/bundler.rb` is loaded *twice* during such an execution (once when running the `bundle` binary, and once when executing the script with `Kernel.exec('<script>')` in `cli/exec.rb`).
Therefore `Bundler.ORIGINAL_ENV` is overwritten twice, and can't be restored to the original value.

To illustrate I've added a spec about this here (which fails): https://github.com/njam/bundler/commit/8871e4166375ee9a679b5f74f099ee88a4c3d1b8

I think the only way to pass along the original environment to the sub process is by setting yet another environment variable. Maybe bundler should store the environment serialized in `BUNDLE_ORIGINAL_ENV`, and then load it from there if present?

Is this interpretation correct, or am I using the code wrong?

segiddins:
Closes #4251? Not sure.


indirect:
@segiddins looks like it needs a rebase?

indirect:
Rebased.

indirect:
@homu r+

segiddins:
This needs a test. @RochesterinNYC any interest in addin. Unit test coverage for the new module to this pr?

RochesterinNYC:
Sure!

lib/bundler/path_preserver.rb
```diff
@@ -0,0 +1,11 @@
+module Bundler
+  module PathPreserver
+    def self.preserve_path_in_environment(env_var, env = ENV)
```
RochesterinNYC: Is the `env` parameter used at all?

indirect:
@segiddins this change somehow breaks the test that sets PATH :/

segiddins:
Yeah, I remember seeing that. I'll just need to update that test

segiddins:
@homu r+

PR #4257, opened 2016-01-31
Ruby version fixes

segiddins:
I believe these bugs should be blockers for 1.12, but I don't have it in me to write specs yet.

\c @indirect

### DIFF
--- a/lib/bundler/ruby_version.rb
+++ b/lib/bundler/ruby_version.rb
@@ -47,7 +47,7 @@ module Bundler
     end
 
     def gem_version
-      Gem::Version.new(version)
+      Gem::Requirement.create(version).requirements.first.last
     end
 
     # Returns a tuple of these things:
--- a/lib/bundler/ruby_version.rb
+++ b/lib/bundler/ruby_version.rb
@@ -1,9 +1,14 @@
 # frozen_string_literal: true
 module Bundler
   class RubyVersion
-    attr_reader :version, :patchlevel, :engine, :engine_version, :gem_version
+    attr_reader :versions,
+                :patchlevel,
+                :engine,
+                :engine_versions,
+                :gem_version,
+                :engine_gem_version
 
-    def initialize(version, patchlevel, engine, engine_version)
+    def initialize(versions, patchlevel, engine, engine_version)
       # The parameters to this method must satisfy the
       # following constraints, which are verified in
       # the DSL:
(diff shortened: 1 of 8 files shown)


indirect:
@segiddins looks like this needs a rebase, too

indirect:
rebased

segiddins:
@indirect please review this

segiddins:
Unless... do we want to support specifying multiple ruby versions?

indirect:
@segiddins I have no idea, tbh. In theory, if we allow `ruby "~> 2.3"`, we should probably also allow `ruby "~> 2.2", ">= 2.2.3"`?

segiddins:
Sure, I'll implement that later.

jtarchie:
Happy to help out. What are the bugs you are referring to?

segiddins:
@jtarchie this PR should address them.

@indirect @RochesterinNYC, please review?

spec/bundler/ruby_dsl_spec.rb
```diff
@@ -0,0 +1,93 @@
+require "spec_helper"
```
RochesterinNYC: Think this might be missing the `# frozen_string_literal: true` line at the top? Travis seems to be throwing errors/failing out whenever a spec or file is missing this.

lib/bundler/ruby_version.rb
```diff
@@ -16,25 +21,31 @@ def initialize(version, patchlevel, engine, engine_version)
...
+    def to_s(versions = self.versions)
+      output = String.new("ruby #{versions_string(versions)}")
       output << "p#{patchlevel}" if patchlevel
-      output << " (#{engine} #{engine_version})" unless engine == "ruby"
+      output << " (#{engine} #{versions_string(engine_versions)})" unless engine == "ruby"
 
       output
     end
 
+    def single_version_string
```
RochesterinNYC: Do we need a unit test for this new method?

lib/bundler/ruby_version.rb
```diff
@@ -57,17 +64,22 @@ def gem_version
...
+      elsif versions.empty? || !matches?(versions, other.gem_version)
+        [:version, versions_string(versions), versions_string(other.versions)]
+      elsif @input_engine && !matches?(engine_versions, other.engine_gem_version)
+        [:engine_version, versions_string(engine_versions), versions_string(other.engine_versions)]
       elsif patchlevel && (!patchlevel.is_a?(String) || !other.patchlevel.is_a?(String) || !matches?(patchlevel, other.patchlevel))
         [:patchlevel, patchlevel, other.patchlevel]
       end
     end
 
+    def versions_string(versions)
```
RochesterinNYC: Do we need a unit test for this new method?

RochesterinNYC:
Actual implementation and specs LGTM. Was the bug that the new version operator logic/code did not work for `Gemfile.lock` specified ruby versioning?

lib/bundler/ruby_version.rb
segiddins: its tested indirectly

lib/bundler/ruby_version.rb
segiddins: nope, it's tested indirectly via the platform specs

segiddins:
One of the bugs had to do with `bundle platform --ruby` returning a requirement rather than the locked ruby version. The other had to do with checking if a gem that had a `required_ruby_version` was compatible with the Gemfile's ruby version.

segiddins:
_finally_ passing

indirect:
@homu r+

Issue #4259, opened 2016-01-31
Confusing error message when gemspec has a different platform than what's in Gemfile.lock


tarcieri:
I maintain a gem (nio4r) which ships CRuby and JRuby backends, but shares a single repository, common tests, etc.

When I bundle under CRuby and then try to bundle again under JRuby, I get this error:

```
Could not find gem 'nio4r (>= 0) ruby' in source at ..
Source contains 'nio4r' at: 1.2.1
```

It probably has to do with this hax in my gemspec:

```ruby
  if defined? JRUBY_VERSION
    gem.files << "lib/nio4r_ext.jar"
    gem.platform = "java"
  else
    gem.extensions = ["ext/nio4r/extconf.rb"]
  end
```

The error message should probably include the platform, at least in cases where the only difference is the platform, e.g.

```
Could not find gem 'nio4r (>= 0) ruby' in source at ..
Source contains 'nio4r' at: 1.2.1 java
```

PR #4262, opened 2016-01-31
Add platform info to error message for different platform in gemspec than Gemfile.lock

RochesterinNYC:
- Addresses #4259

### DIFF
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -364,7 +364,7 @@ module Bundler
         version_platform_str = String.new(version.to_s)
         version_platform_str << " #{platform}" unless platform.nil?
       end
-      version_platform_strs.join(",")
+      version_platform_strs.join(", ")
     end
   end
 end
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -341,10 +341,11 @@ module Bundler
             "try passing them all to `bundle update`"
         elsif requirement.source
           name = requirement.name
-          versions = @source_requirements[name][name].map(&:version)
-          message  = String.new("Could not find gem '#{requirement}' in #{requirement.source}.\n")
-          message << if versions.any?
-                       "Source contains '#{name}' at: #{versions.join(", ")}"
+          specs = @source_requirements[name][name]
+          versions_with_platforms = specs.map {|s| [s.version, s.platform] }
+          message = String.new("Could not find gem '#{requirement}' in #{requirement.source}.\n")
+          message << if versions_with_platforms.any?
+                       "Source contains '#{name}' at: #{formatted_versions_with_platforms(versions_with_platforms)}"
                      else
                        "Source does not contain any versions of '#{requirement}'"
                      end
@@ -354,6 +355,16 @@ module Bundler
         end
         raise GemNotFound, message
       end
+    end
+
+    def formatted_versions_with_platforms(versions_with_platforms)
+      version_platform_strs = versions_with_platforms.map do |vwp|
+        version = vwp.first
+        platform = vwp.last
+        version_platform_str = String.new(version.to_s)
+        version_platform_str << " #{platform}" unless platform.nil?
+      end
+      version_platform_strs.join(",")
     end
   end
 end
--- a/spec/install/gemfile/git_spec.rb
+++ b/spec/install/gemfile/git_spec.rb
@@ -82,7 +82,23 @@ describe "bundle install with git sources" do
         gem "foo", "1.1", :git => "#{lib_path("foo-1.0")}"
       G
 
-      expect(out).to include("Source contains 'foo' at: 1.0")
+      expect(out).to include("Source contains 'foo' at: 1.0 ruby")
+    end
+
+    it "complains with version and platform if pinned specs don't exist in the git repo" do
+      simulate_platform "java"
+
+      build_git "only_java" do |s|
+        s.platform = "java"
+      end
+
+      install_gemfile <<-G
+        platforms :jruby do
+          gem "only_java", "1.2", :git => "#{lib_path("only_java-1.0-java")}"
+        end
+      G
+
+      expect(out).to include("Source contains 'only_java' at: 1.0 java")
     end
 
     it "still works after moving the application directory" do
--- a/spec/install/gemfile/git_spec.rb
+++ b/spec/install/gemfile/git_spec.rb
@@ -101,6 +101,27 @@ describe "bundle install with git sources" do
       expect(out).to include("Source contains 'only_java' at: 1.0 java")
     end
 
+    it "complains with multiple versions and platforms if pinned specs don't exist in the git repo" do
+      simulate_platform "java"
+
+      build_git "only_java", "1.0" do |s|
+        s.platform = "java"
+      end
+
+      build_git "only_java", "1.1" do |s|
+        s.platform = "java"
+        s.write "only_java1-0.gemspec", File.read("#{lib_path("only_java-1.0-java")}/only_java.gemspec")
+      end
+
+      install_gemfile <<-G
+        platforms :jruby do
+          gem "only_java", "1.2", :git => "#{lib_path("only_java-1.1-java")}"
+        end
+      G
+
+      expect(out).to include("Source contains 'only_java' at: 1.0 java, 1.1 java")
+    end
+
     it "still works after moving the application directory" do
       bundle "install --path vendor/bundle"
       FileUtils.mv bundled_app, tmp("bundled_app.bck")
--- a/spec/support/builders.rb
+++ b/spec/support/builders.rb
@@ -103,9 +103,14 @@ module Spec
           s.write "lib/platform_specific.rb", "PLATFORM_SPECIFIC = '1.0.0 x86-darwin-100'"
         end
 
-        build_gem "only_java" do |s|
+        build_gem "only_java", "1.0" do |s|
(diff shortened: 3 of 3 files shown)


lib/bundler/resolver.rb
```diff
@@ -344,7 +344,7 @@ def verify_gemfile_dependencies_are_found!(requirements)
           versions = @source_requirements[name][name].map(&:version)
           message  = "Could not find gem '#{requirement}' in #{requirement.source}.\n"
           if versions.any?
-            message << "Source contains '#{name}' at: #{versions.join(", ")}"
+            message << "Source contains '#{name}' at: #{versions.join(", ")} #{requirement.__platform}"
```
segiddins: Shouldn't print the platform if it's empty or ruby?

RochesterinNYC: For the ruby case, we don't want to have it print `Could not find gem 'foojava (= 1.1) ruby'` or something? Isn't it possible for the target gemspec to indicate java or some other platform, but the Gemfile.lock to indicate ruby?

Ie. we'd want
```
Could not find gem 'foojava (>= 0) java' in source at ..
Source contains 'foojava' at: 1.2.1 ruby
```
instead of:
```
Could not find gem 'foojava (>= 0) java' in source at ..
Source contains 'foojava' at: 1.2.1
```

Thoughts?

segiddins: we typically don't print the platform when it's ruby, since that's implied

RochesterinNYC: Gotcha. In that case, any context on how to specify gems in the test suite that have a non-ruby platform? I'm looking at the `Spec::Builders` module and the gems being built using `build_repo1`. The gems that specify ulterior platforms like so:

```
        build_gem "only_java" do |s|
          s.platform = "java"
          s.write "lib/platform_specific.rb", "PLATFORM_SPECIFIC = '1.0.0 JAVA'"
        end
```

seem to still have ruby (rather than java in this case) as the platform when I'm playing around with the following spec in `spec/install/gemfile/git_spec.rb`:

```
    it "complains if pinned specs don't exist in the git repo" do
      build_git "only_java"

      install_gemfile <<-G
        gem "only_java", "1.1", :git => "#{lib_path("only_java-1.0")}"
      G

      expect(out).to include("Source contains 'foo' at: 1.0")
    end
```

segiddins: try 
```ruby
build_git "only_java" do |s|
          s.platform = "java"
          s.write "lib/platform_specific.rb", "PLATFORM_SPECIFIC = '1.0.0 JAVA'"
        end
```

RochesterinNYC: Hm, okay, have tried the following so far:
```ruby
        build_gem "only_java" do |s|
          s.platform = "java"
          s.write "lib/platform_specific.rb", "PLATFORM_SPECIFIC = '1.0.0 JAVA'"
        end
```
 and

```ruby
        build_gem "only_java" do |s|
          s.platform = "java"
          s.write "lib/only_java.rb", "ONLY_JAVA = '1.0.0 JAVA'"
        end
```

and have not been able to successfully set the platform as java. I might need to dig down the rabbit hole of the test suite gem builder stuff.

RochesterinNYC:
Finally got the integration spec setup working as expected.

lib/bundler/resolver.rb
```diff
@@ -342,15 +342,17 @@ def verify_gemfile_dependencies_are_found!(requirements)
...
+          platform = requirement.__platform
           message  = String.new("Could not find gem '#{requirement}' in #{requirement.source}.\n")
-          message << if versions.any?
-                       "Source contains '#{name}' at: #{versions.join(", ")}"
-                     else
-                       "Source does not contain any versions of '#{requirement}'"
-                     end
+          if versions.any?
+            message << "Source contains '#{name}' at: #{versions.join(", ")}"
+            message << " #{platform}" unless platform.nil? || platform == "ruby"
```
segiddins: Won't you have to print the platform per spec?

RochesterinNYC: By `spec`, do you mean `requirement` in the context of the `verify_gemfile_dependencies_are_found!` method? Looking at this method, it seems to compose an error message and then raise a `GemNotFound` error with that error message when a requirement is not fulfilled. 

So, if you mean print the platforms for each `requirement`, I don't think that'd be necessary since only the first non-fulfilled `requirement` would be displayed/cause an error. Please let me know if I'm mixing up terminology/concepts here.

segiddins:
I meant the platform for each version.

RochesterinNYC:
Ah gotcha. Updated the PR with platform printing associated with each spec/version. This will print things like `Source contains 'foo' at: 1.0 ruby`. I honestly think this is a reasonable thing to maintain in terms of developer UX though. It provides more clarity and lessens confusion/uncertainty as to the platform associated with any found specs.

lib/bundler/resolver.rb
```diff
@@ -341,19 +341,30 @@ def verify_gemfile_dependencies_are_found!(requirements)
...
+          message = String.new("Could not find gem '#{requirement}' in #{requirement.source}.\n")
+          message << if versions_with_platforms.any?
+                       "Source contains '#{name}' at: #{formatted_versions_with_platforms(versions_with_platforms)}"
                      else
                        "Source does not contain any versions of '#{requirement}'"
                      end
         else
-          message = "Could not find gem '#{requirement}' in any of the gem sources " \
-            "listed in your Gemfile or available on this machine."
+          message = String.new("Could not find gem '#{requirement}' in any of the gem sources " \
```
segiddins: no need for this to be `String.new` since it isn't mutated

spec/bundler/path_preserver_spec.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+require "spec_helper"
+
+describe Bundler::PathPreserver do
```
segiddins: this looks... unrelated?

spec/install/gemfile/git_spec.rb
```diff
@@ -82,7 +82,23 @@
         gem "foo", "1.1", :git => "#{lib_path("foo-1.0")}"
       G
 
-      expect(out).to include("Source contains 'foo' at: 1.0")
+      expect(out).to include("Source contains 'foo' at: 1.0 ruby")
+    end
+
+    it "complains with platform (if not ruby) if pinned specs don't exist in the git repo" do
```
segiddins: can you test this with multiple versions?

spec/bundler/path_preserver_spec.rb
RochesterinNYC: Ah, sorry. Rebasing this PR repeatedly is causing unrelated git issues. Will fix.

spec/install/gemfile/git_spec.rb
RochesterinNYC: Hm okay, so what is the functional process for having multiple versions in the source when the git value refers to a specific directory with just a gemspec and basic lib file? I.e. `lib_path("only_java-1.0-java")` is where it looks for sources, so what does having multiple versions in that directory entail?

RochesterinNYC: Basically, not sure how to make the `Gemfile` something like this:
```
        platforms :jruby do
          gem "only_java", "1.2", :git => "#{lib_path("only_java-1.0-java")}"
        end
```
and have it detect both 1.0 and 1.1 versions of `only_java` as it seems to be going by the `only_java.gemspec` specification in the `tmp/libs/only_java-1.0-java` directory. In this gemspec, only one version can be specified.

RochesterinNYC:
@segiddins I've updated this PR. You had originally requested adding a spec for multiple versions. However, after looking into this further, I don't believe this is possible via the `:git` option for a gem in the `Gemfile`.

This is the following spec that I had originally tried to add for multiple versions (along with appropriate `build_gem` calls in `spec/support/builders.rb`:

```ruby
    it "complains with multiple versions and platforms if pinned specs don't exist in the git repo" do
      simulate_platform "java"

      build_git "only_java", "1.0" do |s|
        s.platform = "java"
      end

      build_git "only_java", "1.1" do |s|
        s.platform = "java"
      end

      install_gemfile <<-G
        platforms :jruby do
          gem "only_java", "1.2", :git => "#{lib_path("only_java-1.1-java")}"
        end
      G

      expect(out).to include("Source contains 'only_java' at: 1.0 java, 1.1 java")
    end
```

However, it only ever has the following output:

```
       Fetching /path/to/bundler/tmp/libs/only_java-1.1-java
       Fetching source index from file:/path/to/bundler/tmp/gems/remote1/
       Could not find gem 'only_java (= 1.2) java' in /path/to/bundler/tmp/libs/only_java-1.1-java (at master@6e46c59).
       Source contains 'only_java' at: 1.1 java
```

We can't specify multiple `:git` source options for `only_java` in the `Gemfile` and I don't think there's a way to put multiple gem versions in the representation for a single unpacked gem in a way that Bundler will appropriately understand.

segiddins:
@RochesterinNYC multiple gemspecs in the same repo

RochesterinNYC:
Hm okay, I tried that previously but maybe I was going about it wrong. I'll give it another go.

RochesterinNYC:
Updated the PR. Got the multi-version spec working correctly. Thanks for the help @segiddins

segiddins:
:+1:

segiddins:
@homu r+

Issue #2952, opened 2014-03-29
Remove ruby version check from "bundle exec"?


nirvdrum:
This is a continuation to #2951.  Is there any reason to do a Ruby version check when running "bundle exec" at all?  As far as I can tell there's two situations:

* Running a command that does load the Gemfile, such as "bundle exec rails c"
* Running a command that doesn't load the Gemfile at all, such as "bundle exec ruby -v"

I realize I'm stating the obvious there.  But in the former case, the Ruby version check will be performed by the subcommand anyway.  As of today, it's effectively being run twice.  In the latter case, where the Gemfile isn't being loaded, you wouldn't see the error you see today, but does that really matter?  Bundler isn't being used for anything other than a command dispatcher in that case.

segiddins:
@indirect are those spec failures OK?

PR #4265, opened 2016-02-01
[Exec] Avoid loading the definition before exec-ing

### LINKED ISSUE #2951: Don't parse lockfile for "bundle exec"
I've been looking at ways to reduce the overhead for starting up `bundle exec`.  One thing that caught me by surprise is how much evaluating and parsing goes on.  The first thing the exec command does is validate the ruby version:

https://github.com/bundler/bundler/blob/v1.6.0/lib/bundler/cli/exec.rb#L11

This ends up both eval'ing the Gemfile and parsing the lockfile.  As far as I can tell, the results of parsing the lockfile are summarily discarded, only to be re-parsed when the subcommand actually starts.  The parsing code itself is a fair bit of regex matches, so it's not particularly lightweight.  If it could be bypassed altogether here, that would be a big help.

segiddins:
Closes #2951, #2952.

@indirect it feels like I'm missing something?

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -293,12 +293,12 @@ module Bundler
     def which(executable)
       if File.file?(executable) && File.executable?(executable)
         executable
-      elsif ENV["PATH"]
-        path = ENV["PATH"].split(File::PATH_SEPARATOR).find do |p|
+      elsif path = ENV["PATH"]
+        executable_path = path.split(File::PATH_SEPARATOR).find do |p|
           abs_path = File.join(p, executable)
           File.file?(abs_path) && File.executable?(abs_path)
         end
-        path && File.expand_path(executable, path)
+        executable_path && File.expand_path(executable, executable_path)
       end
     end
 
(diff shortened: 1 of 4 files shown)


segiddins:
@nirvdrum if this is something that still interests you, would you mind checking this PR out?

indirect:
@segiddins well, this will bundle exec to a non-Ruby command without throwing an exception if the ruby version that ran `bundle exec` is wrong? but I have no idea if that's actually a problem...

segiddins:
It technically will, but I don't think that's really going to be a big problem -- if anything ever spawns a ruby process, bundler.setup will be called anyways.

segiddins:
Given we have no tests that break with this change, I think I'm OK with it?

indirect:
Yeah, my thoughts exactly.

@homu r+

nirvdrum:
Sorry, I was offline for a bit. Thanks for taking over and getting it fixed!

PR #4268, opened 2016-02-03
Place bundler loaded gems after -I and RUBYLIB

Elffers:
Previously, gems were being placed at the front of the LOAD_PATH. This
meant you couldn't override a gem by setting -I or RUBYLIB.

This patch places -I and RUBYLIB in front of loaded gems and matches the
behavior in RubyGems.

### DIFF
--- a/lib/bundler/runtime.rb
+++ b/lib/bundler/runtime.rb
@@ -39,7 +39,7 @@ module Bundler
         load_paths = spec.load_paths.reject {|path| $LOAD_PATH.include?(path) }
 
         # See Gem::Specification#add_self_to_load_path (since RubyGems 1.8)
-        insert_index = Gem.load_path_insert_index
+        insert_index = Bundler.rubygems.load_path_insert_index
 
         if insert_index
           # Gem directories must come after -I and ENV['RUBYLIB']
--- a/lib/bundler/runtime.rb
+++ b/lib/bundler/runtime.rb
@@ -37,7 +37,17 @@ module Bundler
 
         Bundler.rubygems.mark_loaded(spec)
         load_paths = spec.load_paths.reject {|path| $LOAD_PATH.include?(path) }
-        $LOAD_PATH.unshift(*load_paths)
+
+        # See Gem::Specification#add_self_to_load_path (since RubyGems 1.8)
+        insert_index = Gem.load_path_insert_index
+
+        if insert_index
+          # Gem directories must come after -I and ENV['RUBYLIB']
+          $LOAD_PATH.insert(insert_index, *load_paths)
+        else
+          # We are probably testing in core, -I and RUBYLIB don't apply
+          $LOAD_PATH.unshift(*load_paths)
+        end
       end
 
       setup_manpath
--- a/spec/runtime/setup_spec.rb
+++ b/spec/runtime/setup_spec.rb
@@ -108,29 +108,36 @@ describe "Bundler.setup" do
     end
   end
 
-  it "puts loaded gems after -I and RUBYLIB" do
-    install_gemfile <<-G
-      source "file://#{gem_repo1}"
-      gem "rack"
-    G
-
-    ruby <<-RUBY
-      require 'rubygems'
-      require 'bundler'
-      $LOAD_PATH.unshift('fake -I')
-      Bundler.setup
-
-      puts $LOAD_PATH
-    RUBY
-
-    load_path = out.split("\n")
-    rack_load_order = load_path.each_with_index do |dir, i|
-      break i if dir =~ /rack/
+  context "load order" do
+    after do
+      ENV.delete "RUBYLIB"
+      ENV.delete "RUBYOPT"
     end
 
-    expect(err).to eq("")
-    expect(load_path.first).to eq("fake -I")
-    expect(rack_load_order).to be > 0
+    it "puts loaded gems after -I and RUBYLIB" do
+      install_gemfile <<-G
+        source "file://#{gem_repo1}"
+        gem "rack"
+      G
+
+      ENV["RUBYOPT"] = "-I dash_i_dir"
+      ENV["RUBYLIB"] = "rubylib_dir"
+
+      ruby <<-RUBY
+        require 'rubygems'
+        require 'bundler'
+        Bundler.setup
+        puts $LOAD_PATH
+      RUBY
+
+      load_path = out.split("\n")
+      rack_load_order = load_path.index {|path| path.include?("rack") }
+
+      expect(err).to eq("")
+      expect(load_path[1]).to include "dash_i_dir"
+      expect(load_path[2]).to include "rubylib_dir"
+      expect(rack_load_order).to be > 0
+    end
   end
 
   it "raises if the Gemfile was not yet installed" do
--- a/spec/runtime/setup_spec.rb
+++ b/spec/runtime/setup_spec.rb
@@ -109,11 +109,6 @@ describe "Bundler.setup" do
   end
 
   context "load order" do
-    after do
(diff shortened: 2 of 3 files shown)


indirect:
r+

@segiddins r?

lib/bundler/runtime.rb
```diff
@@ -37,7 +37,17 @@ def setup(*groups)
 
         Bundler.rubygems.mark_loaded(spec)
         load_paths = spec.load_paths.reject {|path| $LOAD_PATH.include?(path) }
-        $LOAD_PATH.unshift(*load_paths)
+
+        # See Gem::Specification#add_self_to_load_path (since RubyGems 1.8)
+        insert_index = Gem.load_path_insert_index
```
segiddins: This should go into the rubygems integration (probably as the whole 'add to load path' bit)

spec/runtime/setup_spec.rb
```diff
@@ -108,6 +108,31 @@
     end
   end
 
+  it "puts loaded gems after -I and RUBYLIB" do
```
segiddins: This doesn't actually seem to test either '-I' or using RUBYLIB

spec/runtime/setup_spec.rb
```diff
@@ -108,6 +108,31 @@
...
+      require 'rubygems'
+      require 'bundler'
+      $LOAD_PATH.unshift('fake -I')
+      Bundler.setup
+
+      puts $LOAD_PATH
+    RUBY
+
+    load_path = out.split("\n")
+    rack_load_order = load_path.each_with_index do |dir, i|
```
segiddins: `= load_path.index {|path| path.include?("rack") }`

lib/bundler/runtime.rb
Elffers: Could you point me to where that is? I'm not familiar with all the rubygems integration work.

segiddins: `rubygems_integration` I believe.

drbrain: Found it [here](https://github.com/bundler/bundler/blob/master/lib/bundler/rubygems_integration.rb)

Comparing [this chunk of the PR](https://github.com/Elffers/bundler/blob/hhh_fix_ruby-I/lib/bundler/runtime.rb#L39-L50) to [RubyGems' implementation](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/specification.rb#L1539-L1554) how do you propose we handle the reject on L39 from bundler? It was introduced by @69acfd88 by @indirect with no tests so I'm unsure what fault it addresses.

indirect: We have a rubygems_integration.rb file where we wrap all interactions with `Gem`... Just so we have a centralized place to handle changes between RubyGems versions. If you create a method there to wrap this, you can call it from `Bundler.rubygems.some_method`.

indirect: @drbrain I don't see a reject in that commit... is that the right sha?

spec/runtime/setup_spec.rb
```diff
@@ -108,6 +108,38 @@
     end
   end
 
+  context "load order" do
+    after do
+      ENV.delete 'RUBYLIB'
```
segiddins: double quotes, and this shouldn't be necessary because of https://github.com/bundler/bundler/blob/0647f9645add95a9ec522b0f8a880b1745d67112/spec/spec_helper.rb#L117

lib/bundler/runtime.rb
indirect: @drbrain I don't see a reject in that commit... is that the right sha?

drbrain: The `unless` [here](https://github.com/bundler/bundler/commit/69acfd88#diff-36d67678b29b89b0dcfe3f3f09a5ed9dR32) was refactored to the `reject` later on.  This commit appears to be the origin commit for the intention.

Elffers: @indirect Is the refactor you are suggesting necessary for the bugfix this PR hopes to address? Since line 39 in this file is different from the actual behavior in Rubygems that @drbrain mentioned, I was hesitant to encapsulate this as a `Bundler.rubygems` method.

segiddins: Given it uses a RubyGems API that could potentially change, please do move this logic into rubygems_integration

drbrain: RubyGems' Gem::Specification#add_self_to_load_path adds all load paths for a spec unconditionally.

Bundler's equivalent adds only unique load paths for a spec.

I don't think bundler should have an equivalent of `add_self_to_load_path` until the reason why L39 exists is addressed.  If the bundler version does something different at some point in the future someone will say "oh, these two methods do the same thing, let's have Bundler call the RubyGems one directly" without looking closely and determining they do not do the same thing.

Then something might break.

Keeping these two similar things that behave differently separate prevents such an occurrence.

If L39 can be deleted I support creating the equivalent of `add_self_to_load_path` in Bundler, but it seems orthogonal to this PR's purpose.

indirect: @Elffers this PR is totally acceptable without refactoring the method.

@drbrain I don't think Bundler should have an add_self_to_load_path either, at least not right now.

This PR only needs one change to be mergeable, and it is moving all references to `Gem` into the `rubygems_integration.rb` file. Something like this, I imagine:

```ruby
insert_index = Bundler.rubygems.load_path_insert_index

# meanwhile, in rubygems_integration.rb
def load_path_insert_index
  Gem.load_path_insert_index
end
```

Based on past experience, it's really important to keep references to RubyGems constants or methods isolated in the `rubygems_integration.rb` file so that we can change them in one central place when needed.

indirect:
Almost there! One more change and this is good to merge. Thanks for submitting it!

lib/bundler/runtime.rb
Elffers: Ah, that makes much more sense. Thanks so much for the clarification, @indirect!

spec/runtime/setup_spec.rb
```diff
@@ -108,6 +108,38 @@
     end
   end
 
+  context "load order" do
+    after do
```
segiddins: This `after` block is unnecessary

spec/runtime/setup_spec.rb
```diff
@@ -108,6 +108,38 @@
...
+        require 'bundler'
+        Bundler.setup
+        puts $LOAD_PATH
+      RUBY
+
+      load_path = out.split("\n")
+      rack_load_order = load_path.index {|path| path.include?("rack") }
+
+      expect(err).to eq("")
+      expect(load_path[1]).to include "dash_i_dir"
```
segiddins: This is failing on 1.8.7

spec/runtime/setup_spec.rb
```diff
@@ -108,6 +108,38 @@
     end
   end
 
+  context "load order" do
+    after do
+      ENV.delete "RUBYLIB"
+      ENV.delete "RUBYOPT"
+    end
+
+    it "puts loaded gems after -I and RUBYLIB" do
```
segiddins: please split the `-I` and `RUBYLIB` tests, and add a `:ruby => '> 1.8.7'` filter to the the `-I` test

Elffers: Fixed in @1ae0067

spec/runtime/setup_spec.rb
Elffers: Addressed in https://github.com/Elffers/bundler/commit/10ccab6a0dae411d722925fb9030475fd315db01

indirect:
Thanks!

@homu r+

PR #4271, opened 2016-02-04
[Exec] Load instead of exec-ing

segiddins:
Proof-of-concept to avoid `exec` where possible, instead just using `load` when we'd be launching the same interpreter anyways. About a .2 second gain in my limited testing. Over a second gain in jruby in my testing, somehow.

Not near ready for merging, but pretty cool for an hour of hacking.

- [x] Tests
- [ ] Centralize error handling
- [ ] Raise exceptions instead of exiting

### DIFF
--- a/lib/bundler/cli/exec.rb
+++ b/lib/bundler/cli/exec.rb
@@ -75,7 +75,7 @@ module Bundler
         "#!/usr/bin/env ruby\n",
         "#!#{Gem.ruby}\n",
       ]
-      first_line = File.read(file, possibilities.map(&:size).max, :mode => "rb")
+      first_line = File.open(file, "rb") {|f| f.read(possibilities.map(&:size).max) }
       possibilities.any? {|shebang| first_line.start_with?(shebang) }
     end
   end
--- a/lib/bundler/cli/exec.rb
+++ b/lib/bundler/cli/exec.rb
@@ -56,6 +56,7 @@ module Bundler
     def kernel_load(file, *args)
       args.pop if args.last.is_a?(Hash)
       ARGV.replace(args)
+      $0 = file
       ui = Bundler.ui
       Bundler.ui = nil
       require "bundler/setup"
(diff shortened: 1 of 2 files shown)


lib/bundler/cli/exec.rb
```diff
@@ -37,10 +51,27 @@ def run
...
+
+    def kernel_load(file, *args)
+      args.pop if args.last.is_a?(Hash)
+      ARGV.replace(args)
+      ui = Bundler.ui
+      Bundler.ui = nil
+      require "bundler/setup"
+      Kernel.load(file)
+    rescue SystemExit
+      true
```
segiddins: should just re-raise this

segiddins:
@indirect please eventually talk me out of this...

indirect:
I keep trying to figure out why this would be a bad idea, and I haven't come up with any yet. :P

lib/bundler/cli/exec.rb
```diff
@@ -37,10 +51,26 @@ def run
...
+    rescue Exception => e # rubocop:disable Lint/RescueException
       Bundler.ui = ui
-      Bundler.ui.error "bundler: exec needs a command to run"
-      exit 128
+      Bundler.ui.error "bundler: failed to load command: #{cmd} (#{file})"
+      abort "#{e.class}: #{e.message}\n#{e.backtrace.join("\n  ")}"
+    end
+
+    def ruby_shebang?(file)
+      first_line = File.open(file, "rb", &:readline)
```
segiddins: probably best to only read a fixed number of bytes? an executable might not hit a `line` for a while?

indirect: yeah, but shebangs could have a very long path... maybe 4096 bytes?

segiddins: we're comparing against two known possibilities, can take the max of them

segiddins:
Specs added!

indirect:
@segiddins I'm okay landing this on master now that it has tests... do you think the other changes are required to merge this, or are good changes to make?

segiddins:
Nah, not really. I'd be happy to land this, I guess?

indirect:
:+1: 

@homu r+

PR #4273, opened 2016-02-04
bundle init: Generate a Gemfile that is Rubocop compliant

perlun:
Rationale: Rubocop complains when you use double quoted strings but not really _need_ it, since you don't perform any string interpolation.

This can be tweaked in the settings for Rubocop, but I think it would be good for Bundler to generate a Gemfile which is compliant by default (i.e. compliant with Rubocop's defalt settings).

(The same also goes for the content generated by `bundle gem`, but let's take this change as a starting point for discussion. If we can reach an agreement on this, then other files can be changed in a similar manner.)

### DIFF
--- a/lib/bundler/templates/Gemfile
+++ b/lib/bundler/templates/Gemfile
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 # A sample Gemfile
-source "https://rubygems.org"
+source 'https://rubygems.org'
 
-# gem "rails"
+# gem 'rails'


indirect:
Bundler uses RuboCop, and requires double-quoted strings. The Gemfile and generated gems all pass Bundler's RuboCop configuration.

Related: https://viget.com/extend/just-use-double-quoted-ruby-strings

perlun:
Thanks André. However, I think there's a difference here: I am not saying that we should change all the Bundler code to use single quoted strings. Instead, I am pointing out that _the current templates are generating code that isn't approved by the default Rubocop configuration_. That is specifically the problem. Any project using `bundle init` using Rubocop has to either:

- tweak their Rubocop settings to favor double-quoted strings instead of single quotes (which of some of us don't want to do, since we prefer single quoted strings), **or**
- manually tweak the Gemfile. Or the .gemspec, if you're creating a gem.

It's not as simple as "double quoted strings are better, period". Just read the blog post you refer to and see that there are comments below the post pointing out that _it's not just about the performance_.

I can see that some people want to have it like it is right now re. the templates. Fine. You will simple have to live with some of us coming in and complaining every once in a while then. A better solution would be to make `bundler` be smarter, maybe having a setting like `STRING_QUOTE_STYLE=single` in the `ENV`, to make `bundler` know and be aware of the different kinds of string quotes.

The Ruby community obviously has two common forms in this area: https://github.com/bbatsov/ruby-style-guide#strings. I think Bundler should be open to that and not too dogmatic.

TimMoore:
> Instead, I am pointing out that the current templates are generating code that isn't approved by the default Rubocop configuration.

They also don't include Rubocop by default, you have to change your project to add it.

> Any project using bundle init using Rubocop has to [...] manually tweak the Gemfile. Or the .gemspec, if you're creating a gem.

Or run `rubocop --auto-correct` if manually editing literally one line of code is too much.

The Bundler generators are intended to be a very simple starting point for customization by end-users, not a be-all and end-all template for new projects. There are other projects out there with more flexible templating that you are welcome to use.

Adding customizable quote style to the templates would add considerable complexity to these simple templates, which then has its own drawbacks such as making it harder to maintain those templates, easier to introduce bugs, and creates a barrier for new contributors. All this to save people having to edit a couple of quote characters in a new project or run `rubocop --auto-correct`?

> I can see that some people want to have it like it is right now re. the templates. Fine. You will simple have to live with some of us coming in and complaining every once in a while then.

This sounds like a threat? That's a pretty rude way to treat volunteers. I hope you can understand if the Bundler maintainers would rather spend their limited time on features and fixes to core functionality within Bundler.

I don't see this thread going anywhere constructive so I'm going to lock it.

PR #4277, opened 2016-02-04
Clarify comment in `bundle gem` gemspec

jbodah:
The comment feels a bit misleading. it sounds like this allows Rubygems pushes by default which it does not


jbodah:
Woops grammar; reopening this with a more informative branch name too

PR #4278, opened 2016-02-04
Clarify comment in `bundle gem` gemspec

jbodah:
The comment feels a bit misleading. It sounds like this allows Rubygems pushes by default which it does not


segiddins:
@homu r+

segiddins:
@homu retry

PR #4280, opened 2016-02-05
Add --parseable (with --porcelain alias) to `bundle outdated` for minimal output

RochesterinNYC:
- This flag changes the output of bundle outdated from:
```
  * activesupport (newest 3.0, installed 2.3.5, requested = 2.3.5) in groups "development, test"
```
to
```
activesupport (newest 3.0, installed 2.3.5, requested = 2.3.5)"
```
and removes the extraneous output relating to fetching gem metadata,
version metadata, git updates, and resolving dependencies.

- Addresses bundler/bundler-features#85

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -245,6 +245,8 @@ module Bundler
       "Only list newer versions allowed by your Gemfile requirements"
     method_option "major", :type => :boolean, :banner => "Only list major newer versions"
     method_option "minor", :type => :boolean, :banner => "Only list at least minor newer versions"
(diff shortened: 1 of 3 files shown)


RochesterinNYC:
I'd like some UX feedback on what the `bundle outdated --parseable` output when there are no potential updates should be. Right now, it'd be completely empty output, which could potentially be confusing to users since they might think nothing actually happened when they ran the command.

indirect:
When there are no updates, I think that no output is the right move. We're specifically trying for output that will be used by an automated system, so no lines seems fine to me.

indirect:
@homu r+

PR #4291, opened 2016-02-10
Add alternative error message for when `--deployment` and `--system` flags are both used on `bundle install`

RochesterinNYC:
Calling `bundle install --deployment --system` currently results in the following error message:

```
$ bundle install --deployment --system
You have specified both a path to install your gems to,
as well as --system. Please choose.
```
 which is a little confusing.

This PR changes the behavior to:

```
$ bundle install --deployment --system
You have specified both --deployment as well as --system. Please choose. #<-- in red
```

### DIFF
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -45,11 +45,7 @@ module Bundler
       # Just disable color in deployment mode
       Bundler.ui.shell = Thor::Shell::Basic.new if options[:deployment]
 
-      if (options[:path] || options[:deployment]) && options[:system]
-        Bundler.ui.error "You have specified both a path to install your gems to, \n" \
-                         "as well as --system. Please choose."
-        exit 1
-      end
+      check_for_options_conflicts
 
       if options["trust-policy"]
         unless Bundler.rubygems.security_policies.keys.include?(options["trust-policy"])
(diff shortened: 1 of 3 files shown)


lib/bundler/cli/install.rb
```diff
@@ -185,5 +181,13 @@ def print_post_install_message(name, msg)
       Bundler.ui.confirm "Post-install message from #{name}:"
       Bundler.ui.info msg
     end
+
+    def check_for_options_conflicts
```
segiddins: this should raise an exception

RochesterinNYC: `raise` something like an `InstallError` instead of exiting? in the rest of the `run` method, `exit 1` is used, though. Thoughts?

RochesterinNYC: This would also result in the output being:
```
$ bundle install  --system --deployment
You have specified both --deployment as well as --system. Please choose.
Bundler::InstallError #<-- in red
```

segiddins: boo. if we do it elsewhere, fine, but I'd prefer to see us raising some sort of `invalid command usage` error instead of all the `SystemExit`s

RochesterinNYC: Yeah I agree. There is an `InvalidOption` error that we could raise, and we could add a handler for that to `log_error` in `Bundler::FriendlyErrors` that would always log the message and also output `invalid command usage` ?

segiddins: seems reasonable (no need to explicitly say `invalid command usage` btw)

RochesterinNYC: Actually, raising an `InvalidOption` error results in nice output:

```
$ bundle install  --system --deployment
You have specified both --deployment as well as --system. Please choose. #<-- in red
```

I'll update the PR to raise that exception ?

segiddins:
Yes please @RochesterinNYC

RochesterinNYC:
Thanks for the feedback. Updated the PR.

indirect:
Awesome.

indirect:
@homu r+

Issue #4163, opened 2015-12-15
Errno::EPROTO: Protocol error @ dir_s_mkdir


andresabraz:
--- ERROR REPORT TEMPLATE -------------------------------------------------------
- What did you do?

  I ran the command `/home/vagrant/.rbenv/versions/2.2.3/bin/bundle install --path`

- What did you expect to happen?

  I expected Bundler to...

- What happened instead?

  Instead, what actually happened was...

Error details

    Errno::EPROTO: Protocol error @ dir_s_mkdir - /vagrant/projects/.bundle
      /home/vagrant/.rbenv/versions/2.2.3/lib/ruby/2.2.0/fileutils.rb:252:in `mkdir'
      /home/vagrant/.rbenv/versions/2.2.3/lib/ruby/2.2.0/fileutils.rb:252:in `fu_mkdir'
      /home/vagrant/.rbenv/versions/2.2.3/lib/ruby/2.2.0/fileutils.rb:226:in `block (2 levels) in mkdir_p'
      /home/vagrant/.rbenv/versions/2.2.3/lib/ruby/2.2.0/fileutils.rb:224:in `reverse_each'
      /home/vagrant/.rbenv/versions/2.2.3/lib/ruby/2.2.0/fileutils.rb:224:in `block in mkdir_p'
      /home/vagrant/.rbenv/versions/2.2.3/lib/ruby/2.2.0/fileutils.rb:210:in `each'
      /home/vagrant/.rbenv/versions/2.2.3/lib/ruby/2.2.0/fileutils.rb:210:in `mkdir_p'
      /home/vagrant/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/bundler-1.11.0/lib/bundler/settings.rb:208:in `block in set_key'
      /home/vagrant/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/bundler-1.11.0/lib/bundler/shared_helpers.rb:114:in `filesystem_access'
      /home/vagrant/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/bundler-1.11.0/lib/bundler/settings.rb:207:in `set_key'
      /home/vagrant/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/bundler-1.11.0/lib/bundler/settings.rb:33:in `[]='
[...]

Issue #3932, opened 2015-08-11
Errno::EPROTO: Protocol error - /vagrant/vendor /ruby


bibarCB:
- What did you do?

  I ran the command `/home/vagrant/.rvm/gems/ruby-2.0.0-p247/bin/bundle `

- What did you expect to happen?

  I expected Bundler to bundle on vagrant!

- What happened instead?

  Instead, what actually happened was this error below

Error details

    Errno::EPROTO: Protocol error - /vagrant/vendor /ruby
      /home/vagrant/.rvm/rubies/ruby-2.0.0-p247/lib/ruby/2.0.0/fileutils.rb:245:in `mkdir'
      /home/vagrant/.rvm/rubies/ruby-2.0.0-p247/lib/ruby/2.0.0/fileutils.rb:245:in `fu_mkdir'
      /home/vagrant/.rvm/rubies/ruby-2.0.0-p247/lib/ruby/2.0.0/fileutils.rb:219:in `block (2 levels) in mkdir_p'
      /home/vagrant/.rvm/rubies/ruby-2.0.0-p247/lib/ruby/2.0.0/fileutils.rb:217:in `reverse_each'
      /home/vagrant/.rvm/rubies/ruby-2.0.0-p247/lib/ruby/2.0.0/fileutils.rb:217:in `block in mkdir_p'
      /home/vagrant/.rvm/rubies/ruby-2.0.0-p247/lib/ruby/2.0.0/fileutils.rb:203:in `each'
      /home/vagrant/.rvm/rubies/ruby-2.0.0-p247/lib/ruby/2.0.0/fileutils.rb:203:in `mkdir_p'
      /home/vagrant/.rvm/gems/ruby-2.0.0-p247/gems/bundler-1.10.6/lib/bundler.rb:308:in `mkdir_p'
      /home/vagrant/.rvm/gems/ruby-2.0.0-p247/gems/bundler-1.10.6/lib/bundler/installer.rb:278:in `create_bundle_path'
      /home/vagrant/.rvm/gems/ruby-2.0.0-p247/gems/bundler-1.10.6/lib/bundler/installer.rb:54:in `run'
      /home/vagrant/.rvm/gems/ruby-2.0.0-p247/gems/bundler-1.10.6/lib/bundler/installer.rb:18:in `install'
      /home/vagrant/.rvm/gems/ruby-2.0.0-p247/gems/bundler-1.10.6/lib/bundler/cli/install.rb:107:in `run'
[...]

Issue #3581, opened 2015-04-20
Bundle package has space in directory name causing exception


ultiweb:
- What did you do? bundle package
- What did you expect to happen? dependencies downloaded to correct place
- What happened instead? crash

Error details

    Errno::EPROTO: Protocol error @ dir_s_mkdir - /opt/cli/vendor/bundle /ruby
    /home/vagrant/.rvm/rubies/ruby-2.2.1/lib/ruby/2.2.0/fileutils.rb:252:in `mkdir'
    /home/vagrant/.rvm/rubies/ruby-2.2.1/lib/ruby/2.2.0/fileutils.rb:252:in `fu_mkdir'
    /home/vagrant/.rvm/rubies/ruby-2.2.1/lib/ruby/2.2.0/fileutils.rb:226:in `block (2 levels) in mkdir_p'
    /home/vagrant/.rvm/rubies/ruby-2.2.1/lib/ruby/2.2.0/fileutils.rb:224:in `reverse_each'
    /home/vagrant/.rvm/rubies/ruby-2.2.1/lib/ruby/2.2.0/fileutils.rb:224:in `block in mkdir_p'
    /home/vagrant/.rvm/rubies/ruby-2.2.1/lib/ruby/2.2.0/fileutils.rb:210:in `each'
    /home/vagrant/.rvm/rubies/ruby-2.2.1/lib/ruby/2.2.0/fileutils.rb:210:in `mkdir_p'
    /home/vagrant/.rvm/gems/ruby-2.2.1/gems/bundler-1.9.4/lib/bundler.rb:310:in `mkdir_p'
    /home/vagrant/.rvm/gems/ruby-2.2.1/gems/bundler-1.9.4/lib/bundler/installer.rb:332:in `create_bundle_path'
    /home/vagrant/.rvm/gems/ruby-2.2.1/gems/bundler-1.9.4/lib/bundler/installer.rb:54:in `run'
    /home/vagrant/.rvm/gems/ruby-2.2.1/gems/bundler-1.9.4/lib/bundler/installer.rb:18:in `install'
    /home/vagrant/.rvm/gems/ruby-2.2.1/gems/bundler-1.9.4/lib/bundler/cli/install.rb:80:in `run'
    /home/vagrant/.rvm/gems/ruby-2.2.1/gems/bundler-1.9.4/lib/bundler/cli/package.rb:32:in `install'
[...]

PR #4292, opened 2016-02-10
Raise new `Bundler::VirtualProtocolError` in response to `Errno::EPROTO`

RochesterinNYC:
- Related to #4163, #3932, and #3581

Ideally, a custom error and/or error message would be returned in these cases rather than just error out because of the uncaught `Errno::EPROTO`.

I'd like some feedback on what the appropriate error message should be however. The caveat is that it seems like a wide range of errors (possibly any error?) that occur in the presence of virtualization will raise this `Errno::EPROTO` error.

### DIFF
--- a/lib/bundler/errors.rb
+++ b/lib/bundler/errors.rb
@@ -98,4 +98,13 @@ module Bundler
 
     status_code(26)
   end
+
+  class VirtualProtocolError < BundlerError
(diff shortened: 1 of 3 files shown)


indirect:
This looks good to me. :+1: 

@homu r+

PR #4293, opened 2016-02-11
Add `--count` flag to `bundle show` that outputs required and bundled gem counts

RochesterinNYC:
Output will look like:
```
$ bundle show --count
Specified gems: 3
Bundled gems: 54
```

- closes bundler/bundler-features#110


indirect:
Thanks for this! I think this is the kind of gradual-feature-creep that we'd like to avoid in core, so it's better suited for a gem that provides a `bundle-count` command, or a shell script that counts the output lines from `bundle show`.

PR #4297, opened 2016-02-13
Add :gitlab option to DSL

mib32:
Since GitLab is getting more attention in the OSS development community, i propose to include :gitlab option to the DSL, so it would be easy to specify dependencies to the gems hosted on GitLab.

`gem 'some_gem', gitlab: 'some gem' `

Inspired by https://news.ycombinator.com/item?id=11091577

### DIFF
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -248,6 +248,11 @@ module Bundler
         repo_name ||= user_name
         "https://#{user_name}@bitbucket.org/#{user_name}/#{repo_name}.git"
       end
+
+      git_source(:gitlab) do |repo_name|
+        repo_name = "#{repo_name}/#{repo_name}" unless repo_name.include?("/")
+        "https://gitlab.com/#{repo_name}.git"
+      end
     end
 
     def with_source(source)
--- a/spec/bundler/dsl_spec.rb
+++ b/spec/bundler/dsl_spec.rb
@@ -62,6 +62,18 @@ describe Bundler::Dsl do
         bitbucket_uri = "https://mcorp@bitbucket.org/mcorp/mcorp.git"
         expect(subject.dependencies.first.source.uri).to eq(bitbucket_uri)
       end
+
+      it "converts :gitlab to :git" do
+        subject.gem("sparks", :gitlab => "indirect/sparks")
+        gitlab_uri = "https://gitlab.com/indirect/sparks.git"
+        expect(subject.dependencies.first.source.uri).to eq(gitlab_uri)
+      end
+
+      it "converts 'mcorp' to 'mcorp/mcorp' in :gitlab" do
+        subject.gem("sparks", :gitlab => "mcorp")
+        gitlab_uri = "https://gitlab.com/mcorp/mcorp.git"
+        expect(subject.dependencies.first.source.uri).to eq(gitlab_uri)
+      end
     end
   end
 


fotanus:
:+1: I could use this feature

segiddins:
We've said in the past this is not something we want to have to support for the next five years.

seanlinsley:
@segiddins then bring on more people to support Bundler. I'd certainly volunteer.

The fact that Github, Gists, and Bitbucket are supported but Gitlab isn't doesn't make sense.

It's four lines of code. How much maintenance burden is that actually going to add?

seanlinsley:
When I saw this PR get opened I tried looking for the original ticket where it was decided not to support Gitlab, but I couldn't find it. Anyone know where it is? These searches didn't help:

- https://github.com/bundler/bundler/search?q=gitlab&type=Issues
- https://github.com/bundler/bundler-features/search?q=gitlab&type=Issues

seanlinsley:
Looks like all custom sources other than Github have been deprecated on the `2-0-dev` branch: 3079f1c267e91ea670767f2477874c837e11b514. There's also a good deal of conversation on the commit. I remember reading that at the time (while I still can't find the source ticket), and still fundamentally disagree.

@indirect mentioned that supporting these custom sources has costed dozens of maintainer hours that would have costed thousands of dollars had they been paid for their time. While I can certainly understand that (I maintain open source projects without pay, and support RubyTogether), it seems that more time has been consumed debating why not to add the four lines than the four lines themselves would ever cost to actually maintain.

Pinging @DouweM. It looks like GitLab is now a member of RubyTogether :smile:

segiddins:
@seanlinsley take for example the github shortcut. We've been forced to keep using the `git` protocol instead of `https` for _years_ because of backwards compatibility issues. That's a _security issue_ that exists only because that "shortcut" has to be officially supported.

segiddins:
Closing.
See https://github.com/bundler/bundler/pull/4109 for an example of the maintenance headache these shortcuts cause, and https://github.com/bundler/bundler/commit/3079f1c267e91ea670767f2477874c837e11b514 for yet another explanation of why they were removed in 2.0.

indirect:
@mib32 @seanlinsley @fotanus I would be happy to work with any of you on an API that would allow a plugin gem to provide a gitlab source... if you have time and inclination, let me know.

mib32:
@indirect My first thought that i'm not sure if it has any sense, i am personally more involved into the idea of global config. 

What you are talking about, gonna be looking like this?
```ruby
gem 'bundler-gitlab'
gem 'rails', gitlab: 'rails'
```

indirect:
I think it would need to be

```
plugin "bundler-gitlab"
gem "foo", gitlab: "foo
```

indirect:
but yes, that's what I was thinking, pretty much :)

DouweM:
@indirect Good idea! GitLab would be very happy to maintain a `bundler-gitlab` plugin and pick up any potential headache. :)

fotanus:
@indirect I totally can help! Not with much free time before day march 19, but I believe it is an easy implementation, right? I think I can make time. I also didn't knew that bundler accepted plugins.

mib32:
Awesome that everyone's happy! 
@indirect i'm also up to work anytime soon.

zedtux:
Hey guys, I'm also looking for this `gitlab:` source in the `Gemfile`. Did you start anything yet?

seanlinsley:
It's pretty easy to define yourself:
```ruby
git_source :gitlab do |repo|
  repo = "#{repo}/#{repo}" unless repo.include? '/'
  "https://gitlab.com/#{repo}.git"
end
```

PR #4298, opened 2016-02-13
Refactor path/environment preserver, make it always preserve the PATH

### LINKED ISSUE #4251: bundle exec's "with_original_env" doesn't preserve PATH

njam:
Closes #4251

@segiddins @indirect @RochesterinNYC wdyt?


njam:
I changed the backup environment variable prefix from `_ORIGINAL_` to `BUNDLE_ORIG_`, as I thought it's good to have a common prefix for all those variables. If you think there's any downside of such a change I will change it back!

segiddins:
Can you look into the failures?

njam:
It's quite difficult to understand how the environment is modified throughout the tests :)
But I think it's good now.

Wdyt? And is it okay to change the backup variable to `BUNDLE_ORIG_*`?

segiddins:
:+1: looks good to me!

indirect:
@homu r+

PR #4304, opened 2016-02-16
Bundle Viz : Options without can have multiple values

mobilutz:
I wanted to only have one 'environment' when calling `bundle viz`. There is the options `--without`, but passing two arguments or some other form of array was not possible.

Maybe this is not the best solution. But it works for me.

The format for the option is `--without=development,test`.


indirect:
Thor autosplits arguments on the colon character `:`, but I am okay with splitting single arguments on comma as well. The tests will need to pass, though.

segiddins:
Install only seems to support `:`, I think that should be sufficient for `viz` as well?

mobilutz:
@segiddins thanks for the hint. I changed the handling inside `viz` to be the same as `install`.
And furthermore I changed the help description a little bit.

What do you think?

segiddins:
👍🏻 just needs a test

mobilutz:
@segiddins I added a very rudimentary test. I don't know how else to test it because the `options` hash is being changed and then passed to a `GraphViz` initializer.

segiddins:
@indirect r?

indirect:
@homu r+

PR #4309, opened 2016-02-18
Fix `bundle outdated --minor` behavior

RochesterinNYC:
If a `Gemfile` is as follows: 
```
source 'https://rubygems.org'
gem "activesupport", "2.3.5"
```

and the latest version of `activesupport` is `3.3.5`, running `bundle outdated --minor` should not report any potential updates, which it currently does/would. Users would be using the `--minor` flag potentially because they want to see what latest updates to their gems are available that aren't huge, possibly breaking version changes. With this PR, only if the latest version of `activesupport` is something like `2.5.5` would updates be reported for `activesupport` via `bundle outdated --minor`.

- Related to discussion at bundler/bundler-features#85

### DIFF
--- a/lib/bundler/cli/outdated.rb
+++ b/lib/bundler/cli/outdated.rb
@@ -56,18 +56,12 @@ module Bundler
           end
           active_spec = active_spec.last
 
-          if options[:major]
-            current_major = current_spec.version.segments.first
-            active_major = active_spec.version.segments.first
-            active_spec = nil unless active_major > current_major
-          end
-
-          if options[:minor]
-            current_minor = current_spec.version.segments[0, 2].compact.join(".")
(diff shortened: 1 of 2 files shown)


indirect:
Awesome :grinning:

indirect:
@homu r+

lib/bundler/cli/outdated.rb
```diff
@@ -123,5 +117,22 @@ def check_for_deployment_mode
         raise ProductionError, error_message
       end
     end
+
+    def update_present_via_semver_portions(current_spec, active_spec, options)
+      current_major = current_spec.version.segments.first
+      active_major = active_spec.version.segments.first
+
+      update_present = false
```
segiddins: technically unnecessary, but nbd

PR #4312, opened 2016-02-19
Add `--patch` flag for `bundle outdated`

RochesterinNYC:
- `bundle outdated --patch` will only report updates in the patch
  version (`v#.#.patch`)
- Related to discussion at bundler/bundler-features#85


segiddins:
@homu r+

Issue #4232, opened 2016-01-24
with_clean_env / with_original_env is not threadsafe?


jjb:
I have a script that fires off many thread which each run a shell commands. Here is the code which invokes the shell commands:

```ruby
Bundler.with_clean_env { `#{command}` }
```

On one of my runs I got this error:

```
RuntimeError: can't add a new key into hash during iteration
                        []= at org/jruby/RubyHash.java:1007
                    replace at org/jruby/RubyHash.java:1744
          with_original_env at /Users/john/.rbenv/versions/jruby-9.0.4.0/lib/ruby/gems/shared/gems/bundler-1.11.2/lib/bundler.rb:198
             with_clean_env at /Users/john/.rbenv/versions/jruby-9.0.4.0/lib/ruby/gems/shared/gems/bundler-1.11.2/lib/bundler.rb:205
```

The suspicious bundler code is here. Line 198 refers to `ENV.replace(ORIGINAL_ENV)`:

```ruby
    def with_original_env
      bundled_env = ENV.to_hash
      ENV.replace(ORIGINAL_ENV)
      yield
    ensure
      ENV.replace(bundled_env.to_hash)
    end

    def with_clean_env
      with_original_env do
        ENV["MANPATH"] = ENV["BUNDLE_ORIG_MANPATH"]
        ENV.delete_if {|k, _| k[0, 7] == "BUNDLE_" }

        if ENV.key?("RUBYOPT")
          ENV["RUBYOPT"] = ENV["RUBYOPT"].sub "-rbundler/setup", ""
        end

        if ENV.key?("RUBYLIB")
          rubylib = ENV["RUBYLIB"].split(File::PATH_SEPARATOR)
          rubylib.delete(File.expand_path("..", __FILE__))
          ENV["RUBYLIB"] = rubylib.join(File::PATH_SEPARATOR)
        end

        yield
      end
    end
```

Subsequent runs of my script never caused the problem again.

I tried writing a script to stress test this race condition and reproduce the problem but I wasn't able to.

I'm not entirely sure where in that code (or elsewhere) the problem is originating. Is it simply that any two threads in a runtime can't iterate over and add a new key to a hash at the same time? Or does the violating key addition have to happen within the actual iteration block? If the former then I think it's simply a matter of a low-probability race condition. If the latter, then I have no idea where this could be occurring.

indirect:
[...]

PR #4303, opened 2016-02-21
Introduce "Bundler.clean_env"

njam:
Via https://github.com/bundler/bundler/issues/4232
Which should return the environment as it was *before* bundler modified it.

Should this use "original_env" or "clean_env" or both?
> At one time, the difference was that one went up one layer of env, while the other removed all Bundler values. I'm not sure if that's still true today.

You can assign this ticket to me if you want.


njam:
@indirect @segiddins I added `original_env` and `clean_env` (the latter with a `@deprecated` flag). I've also rewritten the "with_clean_env_spec" test, I think it's simpler now and covers more.

What do you think?

lib/bundler.rb
```diff
@@ -430,5 +440,14 @@ def upgrade_lockfile
...
       lockfile.rmtree
     end
+
+    # @param env [Hash]
+    def with_env(env)
+      backup = ENV.to_hash
+      ENV.replace(env)
+      yield
+    ensure
+      ENV.replace(backup.to_hash)
```
segiddins: The `to_hash` here is unnecessary

segiddins:
@RochesterinNYC want to review the spec changes?

spec/runtime/with_clean_env_spec.rb
```diff
@@ -2,107 +2,104 @@
...
 
 describe "Bundler.with_env helpers" do
-  shared_examples_for "Bundler.with_*_env" do
-    it "should reset and restore the environment" do
-      gem_path = ENV["GEM_PATH"]
-      path = ENV["PATH"]
+  describe "Bundler.original_env" do
+    it "should return the PATH present before bundle was activated" do
+      gemfile ""
+      bundle "install --path vendor/bundle"
```
RochesterinNYC: These two lines seem to be present in each of the `it` blocks for `describe "Bundler.original_env"` and `describe "Bundler.clean_env"`:
```ruby
      gemfile ""
      bundle "install --path vendor/bundle"
```

They could be extracted into a single `before` block.

spec/runtime/with_clean_env_spec.rb
```diff
@@ -2,107 +2,104 @@
...
+      code = "print Bundler.original_env['PATH']"
 
-      Bundler.with_clean_env do
-        expect(`echo $GEM_PATH`.strip).not_to eq(gem_path)
-        expect(`echo $PATH`.strip).not_to eq(path)
+      path = `getconf PATH`.strip + ":/foo"
```
RochesterinNYC: extract these (`code =` and `path =`) into `let` statements?

RochesterinNYC:
Other than a bit of possible DRYing up and more utilization of `rspec` features, the specs LGTM.

njam:
Thanks for the feedback!

I moved the "bundle install" code to a `before` block.
I'm not sure about `let`. To replace all assignments with `let` I would need to wrap each test in a `context` block where I would then define those values (code, path, result). Is it the way to go?

RochesterinNYC:
@njam Nope, you're right. It's possible to re-arrange these around to use `let` blocks but it's fine as is/isn't necessary. :+1:

segiddins:
@homu r+

PR #4315, opened 2016-02-22
Add `--frozen` flag support for `bundle package`

### LINKED ISSUE #3356: "bundle package" support for --frozen
Trying to use "bundle package --frozen", but it's not a supported switch as it is with "bundle install".

```
$ bundle package --frozen
Unknown switches '--frozen'
```

Internally "bundle package" calls "bundle install", which it passes its parsed options to, so it should be trivial to support this and the entire "bundle install" options should probably be supported by "bundle package".

RochesterinNYC:
- Closes #3356

### DIFF
--- a/spec/commands/package_spec.rb
+++ b/spec/commands/package_spec.rb
@@ -99,6 +99,33 @@ describe "bundle package" do
       expect(bundled_app("vendor/cache/rack-1.0.0.gem")).to exist
     end
   end
+
+  context "with --frozen" do
+    before do
+      gemfile <<-G
+        source "file://#{gem_repo1}"
+        gem "rack"
+      G
+      bundle "install"
+    end
+
+    subject { bundle "package --frozen" }
+
+    it "tries to install with frozen" do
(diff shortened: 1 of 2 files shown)


spec/commands/package_spec.rb
```diff
@@ -99,6 +99,32 @@
...
+
+    subject { bundle "package --frozen" }
+
+    it "tries to install with frozen" do
+      gemfile <<-G
+        source "file://#{gem_repo1}"
+        gem "rack"
+        gem "rack-obama"
+      G
+      subject
```
segiddins: This is supposed to fail, right? If so, `expect_err` and test the exitstatus?

RochesterinNYC: Yup, thanks for the catch.

RochesterinNYC:
PR is updated.

segiddins:
:+1: 
@indirect r?

indirect:
Awesome, thanks!

@homu r+

PR #4301, opened 2016-02-15
Add error with message/explanation for `bundle outdated` in a frozen state

RochesterinNYC:
Adds a helpful error message for when you run `bundle outdated` after having run `bundle install --deployment`.

```
You are trying to check outdated gems in deployment mode. Run `bundle outdated` elsewhere.

If this is a development machine, remove the /path/to/Gemfile freeze
by running `bundle install --no-deployment`.
```
- closes #4287

PR #4318, opened 2016-02-25
Cherry-pick behavior fix for `bundle outdated --minor` for `1-12-stable`

RochesterinNYC:
Related PR: #4309 
The commit about `bundle outdated` in a frozen context (from PR #4301) is also cherry-picked as the commits regarding `bundle outdated --minor` move around the specs from that frozen-related PR. 

Please let me know if only the `bundle outdated --minor` PR commits should be cherry-picked and the cherry-picked commits should just be edited/fixed before merge.

### DIFF
--- a/lib/bundler/cli/outdated.rb
+++ b/lib/bundler/cli/outdated.rb
@@ -10,6 +10,8 @@ module Bundler
     end
 
     def run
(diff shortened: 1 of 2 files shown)


segiddins:
@homu r+

PR #4321, opened 2016-02-25
[DSL] Set ruby version when using the gemspec method

segiddins:
- [x] Specs

\c @indirect


indirect:
@segiddins this is good, does it need a rebase for the tests to pass?

segiddins:
This breaks new tests, so I need to fix those.

segiddins:
@indirect remind me, is this something we still want for 2.0?

indirect:
Yeah, I think this is good!

indirect:
@homu r+

segiddins:
@homu r=indirect

PR #4341, opened 2016-03-05
[RubygemsIntegration] Add support for new activate_bin_path binstubs

segiddins:
This ought to fix the build!

\c @indirect

### DIFF
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -361,25 +361,31 @@ module Bundler
     # +specs+
     def replace_bin_path(specs)
       gem_class = (class << Gem; self; end)
-      redefine_method(gem_class, :bin_path) do |name, *args|
+
+      redefine_method(gem_class, :find_spec_for_exe) do |name, *args|
         exec_name = args.first
 
+        spec = if exec_name
+          specs.find {|s| s.executables.include?(exec_name) }
+        else
+          specs.find {|s| s.name == name }
+        end
+        raise(Gem::Exception, "can't find executable #{exec_name}") unless spec
+        raise Gem::Exception, "no default executable for #{spec.full_name}" unless exec_name ||= spec.default_executable
+        unless spec.name == name
+          warn "Bundler is using a binstub that was created for a different gem.\n" \
+            "This is deprecated, in future versions you may need to `bundle binstub #{name}` " \
+            "to work around a system/bundle conflict."
+        end
+        spec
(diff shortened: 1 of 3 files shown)


segiddins:
(caused by https://github.com/rubygems/rubygems/pull/1527)

segiddins:
@indirect r?

indirect:
@homu r:

indirect:
@homu r+

lib/bundler/rubygems_integration.rb
```diff
@@ -361,25 +361,31 @@ def stub_source_index(specs)
...
+          warn "Bundler is using a binstub that was created for a different gem.\n" \
+            "This is deprecated, in future versions you may need to `bundle binstub #{name}` " \
+            "to work around a system/bundle conflict."
         end
+        spec
+      end
+
+      redefine_method(gem_class, :bin_path) do |name, *args|
+        exec_name = args.first
+        return ENV["BUNDLE_BIN_PATH"] if exec_name == "bundle"
```
taoza: Looks like the new `Gem.activate_bin_path` will miss the chance to hit this early return and breaks this feature on bundle-exec's man page:
> make sure that it's still possible to shell out to bundle from inside a command invoked by bundle exec (using $BUNDLE_BIN_PATH)

Not sure what would be the best way to fix it :-/

segiddins:
@taoza please open a new issue with steps to reproduce the bug you're seeing

taoza:
@segiddins Working on it :)

lib/bundler/rubygems_integration.rb
chrismo: Looks like you were right :) We're about to make this change on this PR #4992. I ran across this while researching a write up of a fresh commit message. It's been a long and winding road to re-discover what you'd found in March.

tguo: @chrismo I'm feeling a bit guilty now of not being able to help fixing the issue a few months back.. But it was the first time I was reporting an issue with Rubygems and Bundler. There was a tendency to think I was missing something in the code and not fully understand the details. But thanks so much for the on https://github.com/bundler/bundler/pull/4992. Looking forward to the release!

chrismo: No worries, there are a lot of moving parts (as I continue to learn) and it can be a tough decision to know what matters and what will just be fine. At least you recorded this comment which helps confirm what we're seeing. :)

Issue #4334, opened 2016-03-04
Colorize gem updates


zedalaye:
Could it be possible to colorize gems updated using an existing gem version during ```bundle update``` like it is already done for updated gems when a new version has to be installed ?

```
Using webmock 1.24.2 (<green>was 1.22.6</green>)
<green>Installing dalli 2.7.6 (was 2.7.5)</green>
```

PR #4342, opened 2016-03-06
Colorize gem updates

bronzdoc:
closes https://github.com/bundler/bundler/issues/4334
alt tag

### DIFF
--- a/lib/bundler/ui/shell.rb
+++ b/lib/bundler/ui/shell.rb
@@ -17,6 +17,10 @@ module Bundler
         @warning_history = []
       end
 
+      def add_color(string, color)
+        @shell.set_color(string, color)
+      end
+
       def info(msg, newline = nil)
         tell_me(msg, nil, newline) if level("info")
       end
--- a/lib/bundler/source.rb
+++ b/lib/bundler/source.rb
@@ -18,11 +18,7 @@ module Bundler
         locked_spec = Bundler.locked_gems.specs.find {|s| s.name == spec.name }
         locked_spec_version = locked_spec.version if locked_spec
         if locked_spec_version && spec.version != locked_spec_version
-          message += if block_given?
-                       yield locked_spec_version
-                     else
-                       " (was #{locked_spec_version})"
-                     end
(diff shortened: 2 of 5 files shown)


seanlinsley:
There's this nice bit of code that already exists to colorize text: https://github.com/bundler/bundler/blob/master/lib/bundler/ui/shell.rb. Shouldn't we reuse that here?

The Bundler output is also theoretically supposed to be computer-parseable, which is why color can be disabled. So if this doesn't [follow the same pattern](https://github.com/bundler/bundler/blob/master/lib/bundler/ui/shell.rb#L12-L14), it might be considered a breaking change.

bronzdoc:
@seanlinsley i don't think i can use those shell methods for this specific task or maybe i could but will be uglier, i may create a helper in the shell ui to follow the pattern, something like:
```ruby
 #.../ui/shell.rb
def add_color(str, color)
  @shell.set_color(str, color)
end 
```    
```ruby
     #lib/bundler/source.rb

     def version_message(spec)

      if Bundler.locked_gems
        locked_spec = Bundler.locked_gems.specs.find {|s| s.name == spec.name }
        locked_spec_version = locked_spec.version if locked_spec
          message += " (#{Bundler.ui.add_color("was #{locked_spec_version}", :green)})"
        end
      end

      message
    end
```
will update in a bit thanks!

@indirect @segiddins thoughts? :grin:

segiddins:
@bronzdoc yeah that seems good

lib/bundler/ui/shell.rb
```diff
@@ -17,6 +17,10 @@ def initialize(options = {})
         @warning_history = []
       end
 
+      def add_color(string, color)
```
segiddins: also needs to be added to the other UI subclass, no?

bronzdoc: yup to ```ruby UI::Silent``` too, thanks.

segiddins:
👍🏻

segiddins:
@homu r+

PR #4345, opened 2016-03-08
[Rakefile] Run Travis tests in parallel

segiddins:
\c @indirect @RochesterinNYC 

this isn't working yet, but I need to study for finals so I thought I'd put this up. the first commit _basically_ works, the second less so

(Might want to use https://github.com/grosser/parallel_tests for the coordination)

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -125,9 +125,14 @@ begin
       releases = %w(v1.3.6 v1.3.7 v1.4.2 v1.5.3 v1.6.2 v1.7.2 v1.8.29 v2.0.14 v2.1.11 v2.2.5 v2.4.8 v2.6.1)
       (branches + releases).each do |rg|
         desc "Run specs with Rubygems #{rg}"
-        RSpec::Core::RakeTask.new(rg) do |t|
-          t.rspec_opts = %w(--format documentation --color)
-          t.ruby_opts  = %w(-w)
+        task rg do
+          patterns = Dir["spec/*"] - %w(spec/spec_helper.rb spec/support)
+          patterns.map do |pattern|
+            pattern = pattern.end_with?(".rb") ? pattern : "#{pattern}/{**/,*/,}*_spec.rb"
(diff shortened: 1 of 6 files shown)


segiddins:
running out of space 😅

Rakefile
```diff
@@ -125,9 +125,14 @@ begin
...
         desc "Run specs with Rubygems #{rg}"
-        RSpec::Core::RakeTask.new(rg) do |t|
-          t.rspec_opts = %w(--format documentation --color)
-          t.ruby_opts  = %w(-w)
+        task rg do
+          patterns = Dir["spec/*"] - %w(spec/spec_helper.rb spec/support)
+          patterns.map do |pattern|
+            pattern = pattern.end_with?(".rb") ? pattern : "#{pattern}/{**/,*/,}*_spec.rb"
+            Thread.new do
+              sh "rspec --pattern '#{pattern}'"
```
RochesterinNYC: is this running every single spec in its own thread? Hahah, might have to group specs for a set number of threads less memory or space limits get hit?

RochesterinNYC:
do we want to look into utilizing parallel rspec + separate file systems/dirs to achieve this rather than having every spec in its own thread?

Rakefile
segiddins: no no, its every subdirectory of specs

aledalgrande:
I would be surprised if this change wouldn't actually slow down tests because the VM keeps swapping. I had more success with tools like Knapsack.

segiddins:
Travis VMs should have enough memory to run multiple ruby processes?

aledalgrande:
Depends on the case, it could probably work for a gem, with Rails apps and fixtures definitely not. Either memory or disk speed was a bottleneck, but probably doesn't apply here: RSpec never goes over 100MB with Bundler.

segiddins:
Closing because this approach didn't work.

PR #4350, opened 2016-03-09
[RuboCop] Update to 0.38.0

segiddins:


segiddins:
@homu r+

lib/bundler/cli/outdated.rb
```diff
@@ -23,11 +23,11 @@ def run
       current_dependencies = {}
       Bundler.ui.silence { Bundler.load.dependencies.each {|dep| current_dependencies[dep.name] = dep } }
 
-      if gems.empty? && sources.empty?
+      definition = if gems.empty? && sources.empty?
```
indirect: I really dislike this style of assignment, and would strongly prefer that we have rubocop enforce that we never assign the results of an if/else. 🙃

PR #4352, opened 2016-03-09
Document change in --without behavior

### LINKED ISSUE #4351: Bundle without change in implicit behavior in v1.10.0
Prior to version 1.10.0 if you are installing using the explicit flag `--without` only the arguments provided will be used in the bundle command so if you

```
$ bundle _1.9.10_ install  --without development:test
# ...
Gems in the groups development and test were not installed.
```

Then it will only exclude your `development` and `test` groups, However with 1.10.0 any config in your `.bundle/config` will get merged in. So someone in your team checks in their bundle config that looks like this:

```
$ cat .bundle/config
---
BUNDLE_WITHOUT: unneeded
```

And you run this

```
$ bundle _1.10.0_ install  --without development:test
```

You exclude a group not present in the explicit without flag:

```
Gems in the groups development, test and unneeded were not installed.
```

I checked over the changelog and I haven't  yet figured out exactly where this change in implicit behavior came from (forgive me if I missed it in the changelog). Should we retroactively document this change or is it a bug that should be reverted?

cc/ @hone

schneems:
close #4351

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -230,6 +230,7 @@ Bugfixes:
   - allow `viz` to work with prereleases (#3621, #3217, @aprescott)
   - validate gemspecs used in path and git gems (#3639, @segiddins, @indirect)
   - stop printing config warnings when config is unchanged (#3649, @fotanus, @indirect)
+  - Without groups saved via `config` are no longer ignored when the `--without` flag is used
 
 ## 1.10.0.pre.2 (2015-05-07)
 


indirect:
Thanks!

@homu r+

segiddins:
You'll need to remove the `[ci skip]` as every commit must go through CI

schneems:
Done

segiddins:
@homu r+

PR #4353, opened 2016-03-10
Git repos should not be updated when --local is passed

chalkos:
* Running `bundle install --local` without having the gems locally fails, because it can not checkout the git repo
* After running `bundle install` at least once
  * following runs of `bundle install` complete successfully and do not checkout the git repositories
  * following runs of `bundle install --local` complete successfully and do not checkout the git repositories


segiddins:
@indirect this seems good to me, even though the specs are failing (Because the bundler 2 branch is failing). I'm :+1: too merge.

PR #4354, opened 2016-03-10
Fix the `Bundler::Fetcher::Downloader` rspec warning in test suite

RochesterinNYC:
Fixes the following warning in the test suite:
```
Bundler::Fetcher::Downloader
  fetch
    when the # requests counter is greater than the redirect limit
WARNING: An expectation of `:body` was set on `nil`. To allow
expectations on `nil` and suppress this message, set
`config.allow_expectations_on_nil` to `true`. To disallow
expectations on `nil`, set `config.allow_expectations_on_nil` to
`false`. Called from
/home/travis/build/bundler/bundler/spec/bundler/fetcher/downloader_spec.rb:19:in
`block (3 levels) in <top (required)>'.
```


segiddins:
@homu r+

PR #4357, opened 2016-03-11
Checking out missing git repos (but not being installed) - Fixing #3981

### LINKED ISSUE #3981: Bundle install with path and excluding a group fails when Gemfile.lock is present.
Here is my Gemfile and Gemfile.lock:

```ruby
#Gemfile
source 'https://rubygems.org'
gem 'json', github: 'flori/json', group: :development
```

```ruby
#Gemfile.lock
GIT
  remote: git://github.com/flori/json.git
  revision: f7394adf3a50631b443135a95ec1bc9d8b8f32ab
  specs:
    json (1.8.3)

GEM
  remote: https://rubygems.org/
  specs:

PLATFORMS
  ruby

DEPENDENCIES
  json!

BUNDLED WITH
   1.10.6
```

The following command fails if we don't have the gem from Github already checkout in `vendor/bundle`.

`bundle install --path=vendor/bundle --without development`

It works if the Gemfile.lock is not present. Removing the Gemfile.lock is a work around.

Shouldn't this work or is there anything I'm doing wrong?

asutoshpalai:
Fixes #3981.

As far as I investigated the problem is, being in `without` group, it is not marked in missing_spec and so it is not resolved while resolving specs at https://github.com/bundler/bundler/blob/master/lib/bundler/installer.rb#L183
and https://github.com/bundler/bundler/blob/master/lib/bundler/installer.rb#L195
but it is in the definition's index at
https://github.com/bundler/bundler/blob/master/lib/bundler/definition.rb#L209
leading to 
https://github.com/bundler/bundler/blob/master/lib/bundler/source/git.rb#L150
where it checks for the presence of the folder at 
https://github.com/bundler/bundler/blob/master/lib/bundler/source/path.rb#L135

Possible solutions are :
- add it to the list of missing_specs
- check while building the index in Definition class and don't add it.

### DIFF
--- a/spec/install/git_spec.rb
+++ b/spec/install/git_spec.rb
@@ -19,7 +19,7 @@ describe "bundle install" do
       build_git "foo"
 
       gemfile <<-G
-        gem "foo", :git =>"file://#{lib_path("foo-1.0")}", group: :development
+        gem "foo", :git => "file://#{lib_path("foo-1.0")}", :group => :development
       G
 
       lockfile <<-L
--- a/spec/install/git_spec.rb
+++ b/spec/install/git_spec.rb
@@ -38,6 +38,7 @@ describe "bundle install" do
       bundle "install --path=vendor/bundle --without development"
 
       expect(out).to include("Bundle complete!")
+      expect(vendored_gems("bundler/gems/foo-1.0-#{revision_for(lib_path("foo-1.0"))[0..11]}")).to be_directory
     end
   end
 end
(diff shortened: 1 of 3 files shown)


spec/install/git_spec.rb
```diff
@@ -14,5 +14,30 @@
       expect(out).to include("Using foo 1.0 from #{lib_path("foo")} (at master@#{revision_for(lib_path("foo"))[0..6]})")
       should_be_installed "foo 1.0"
     end
+
+    it "should check out git repos that are missing but not being installed" do
```
segiddins: could you please also test that the repo is being checked out?

asutoshpalai: Is checking that the directory is present enough?
I added that.

segiddins:
@homu r+

PR #4358, opened 2016-03-11
[Fetcher] Switch from bundler.rubygems.org to index.rubygems.org

segiddins:
@dwradcliffe @indirect by my understanding, this is what we need to do to get the new index behind fastly?


indirect:
@segiddins this is my understanding as well. I'm going to merge it and then start testing to prep for a 1.12 RC.

indirect:
@homu r+

PR #4378, opened 2016-03-18
Forward-port cert changes from 1-12 stable

indirect:

PR #4379, opened 2016-03-18
Update Bundler ssl certs to include new (fastly-related) Rubygems certs

RochesterinNYC:
Update Bundler ssl certs to match the Rubygems certs added in rubygems/rubygems#1548

- Improve clarity of ssl cert spec description


indirect:
Hmmmm... now I'm confused about why #4378 didn't fix this.

RochesterinNYC:
rubygems/rubygems#1548 added some extra ssl certs (ex. `BaltimoreCyberTrustRoot.pem`) to Rubygems master that #4378 doesn't add. We're probably in a situation where there are extra, unneeded certs in both Bundler and Rubygems at the moment.

The changes in this PR should unblock builds for the time being? I do think it's worth cleaning up both Rubygems and Bundler and making sure that they have only the necessary certs.

RochesterinNYC:
Closing in favor of #4380

PR #4380, opened 2016-03-18
Update HTTPS certificates

indirect:
Remove the certs we don't need, add the new cert for index.rubygems.org, and stop checking for equality with RubyGems certs because they're out of date now.


RochesterinNYC:
Added a commit to this that should fix the broken specs via changing them to expect the new subdirectory structure for ssl certs.

indirect:
@RochesterinNYC aha! thank you. :)

@homu r+

RochesterinNYC:
@indirect what is the mapping basis between the directories and certs? As in, what decides which cert goes into which directory (i.e `DigiCertHighAssuranceEVRootCA.pem` in the `rubygems.global.ssl.fastly.net` directory)?

Asking this because I need to create a PR to update this [`util/update_bundled_ca_certificates.rb`](https://github.com/rubygems/rubygems/blob/master/util/update_bundled_ca_certificates.rb) script in Rubygems after rubygems/rubygems#1555 and I don't see any kind of logical mapping in the code.

indirect:
@RochesterinNYC I used each cert to validate an HTTPS session to each domain, and put the cert that actually worked into the directory named after that domain. 😅

indirect:
It seems like the RubyGems script is certs that aren't needed—we don't need to save the certs that are returned by the server, we need to save the root cert that makes it possible to trust all the certs that are returned by the server.

PR #4386, opened 2016-03-22
[Git] Fix checking for equality when a version includes a -

### LINKED ISSUE #4385: Prerelease gems with `-` cause `bundle show` to fail
```
$ cat Gemfile
source "https://rubygems.org"
gem "rack-timeout", "0.4.0-beta.1", github: "heroku/rack-timeout"

$ bundle install
Fetching git://github.com/heroku/rack-timeout.git
Resolving dependencies...
Using bundler 1.12.0.rc
Using rack-timeout 0.4.0.pre.beta.1 from git://github.com/heroku/rack-timeout.git (at master@e6d7211)
Bundled gems are installed into ./.bundle.

$ bundle show
The git source git://github.com/heroku/rack-timeout.git is not yet checked out. Please run `bundle install` before trying to start your application

$ bundle exec echo what
what

$ bundle show
The git source git://github.com/heroku/rack-timeout.git is not yet checked out. Please run `bundle install` before trying to start your application
```

First, it should never be possible to have `install` succeed and then have `show` fail. It should be doubly-impossible to have `install` and `exec` both succeed and still have `show` fail. Lastly, Bundler says it's using `rack-timeout 0.4.0.pre.beta.1`. That's not a version that exists, because [the gemspec for rack-timeout](https://github.com/heroku/rack-timeout/blob/master/rack-timeout.gemspec#L5) has the version set to `0.4.0-beta.1`.

???

segiddins:
Closes #4385.

### DIFF
--- a/lib/bundler/source/git.rb
+++ b/lib/bundler/source/git.rb
@@ -25,7 +25,7 @@ module Bundler
         @ref        = options["ref"] || options["branch"] || options["tag"] || "master"
         @submodules = options["submodules"]
         @name       = options["name"]
-        @version    = Gem::Version.create(options["version"]).to_s
+        @version    = options["version"].to_s.strip.gsub("-", ".pre.")
 
         @copied     = false
         @local      = false
--- a/lib/bundler/source/git.rb
+++ b/lib/bundler/source/git.rb
@@ -25,7 +25,7 @@ module Bundler
         @ref        = options["ref"] || options["branch"] || options["tag"] || "master"
         @submodules = options["submodules"]
         @name       = options["name"]
-        @version    = options["version"]
+        @version    = Gem::Version.create(options["version"]).to_s
 
         @copied     = false
         @local      = false
--- a/spec/commands/show_spec.rb
+++ b/spec/commands/show_spec.rb
@@ -110,6 +110,17 @@ describe "bundle show" do
       bundle :show
       expect(out).to include("foo (1.0 #{sha[0..6]})")
     end
+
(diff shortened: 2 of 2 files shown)


segiddins:
@indirect r?

indirect:
IMO, if the Gemfile (and the gem's gemspec) say that the gem version is `1.0.0-beta.1`, then `bundle show` should also print the version `1.0.0-beta.1`. Why are we accepting a version number and installing it, but then printing a different version number?

segiddins:
Because that's how RubyGems interprets that version string as a `Gem::Version` object

segiddins:
@indirect ping

segiddins:
Ping

segiddins:
I'm going to merge this soon unless anyone objects

RochesterinNYC:
LGTM.

segiddins:
@homu r+

deivid-rodriguez:
Note that in the semver protocol, these are the [only possible prerelease specifications](https://semver.org/#spec-item-9). So this surprising behavior (your lock file or `bundle show` showing a different version string from the one you have declared) happens by default if you follow semver.

It seems to me that the only possible solution would be to break backwards compatibility in `rubygems` and start using "semver prereleases", since currently for `rubygems` any version string containing non-numeric characters [is a pre-release](https://github.com/rubygems/rubygems/blob/9054079ce8a1ad05dd206a3562339b74712c7fc6/lib/rubygems/version.rb#L289-L295).

PR #4389, opened 2016-03-23
Use Travis' cache by default for new gem

tricknotes:
> (Bundler caching is not yet enabled automatically)
> You can explicitly enable Bundler caching in your .travis.yml:
> ```
> language: ruby
> cache: bundler
> ```

ref: https://docs.travis-ci.com/user/caching/#Bundler


segiddins:
Given the default gitignore includes the lockfile, I don't think this change will have any effect

PR #4397, opened 2016-03-26
added required_ruby_version and require_relative to gemspec template

mfazekas:
require_relative is supported since 1.9.2 (which is EOL already). It think it's safe to use it in gem template.

### DIFF
--- a/lib/bundler/templates/newgem/newgem.gemspec.tt
+++ b/lib/bundler/templates/newgem/newgem.gemspec.tt
@@ -1,7 +1,5 @@
 # coding: utf-8
-lib = File.expand_path('../lib', __FILE__)
-$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
-require '<%=config[:namespaced_path]%>/version'
+require_relative 'lib/<%=config[:namespaced_path]%>/version'
 
 Gem::Specification.new do |spec|
   spec.name          = <%=config[:name].inspect%>
@@ -15,6 +13,7 @@ Gem::Specification.new do |spec|
 <%- if config[:mit] -%>
   spec.license       = "MIT"
 <%- end -%>
+  spec.required_ruby_version = Gem::Requirement.new(">= 1.9")
 
   # Prevent pushing this gem to RubyGems.org. To allow pushes either set the 'allowed_push_host'
   # to allow pushing to a single host or delete this section to allow pushing to any host.


lib/bundler/templates/newgem/newgem.gemspec.tt
```diff
@@ -1,7 +1,5 @@
 # coding: utf-8
-lib = File.expand_path('../lib', __FILE__)
-$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
-require '<%=config[:namespaced_path]%>/version'
+require_relative 'lib/<%=config[:namespaced_path]%>/version'
```
segiddins: This will break things of the version file requires anything else inside of lib

lib/bundler/templates/newgem/newgem.gemspec.tt
```diff
@@ -15,6 +13,7 @@ Gem::Specification.new do |spec|
 <%- if config[:mit] -%>
   spec.license       = "MIT"
 <%- end -%>
+  spec.required_ruby_version = Gem::Requirement.new(">= 1.9")
```
segiddins: IMHO if we're going to do this it should be 2.1, which is the oldest MRI version still receiving updates.

lib/bundler/templates/newgem/newgem.gemspec.tt
mfazekas: My understanding is that this is a template for new gems generated with ```bundle gem bundlname``` so i'd use the phrase _requires extra work_ vs _break things_.

I'd argue this is not common for version.rb to include from lib. I've checked on my installed gems:
```sh
find ~/.rvm/gems/ruby-2.3.0/gems/ -name 'version.rb' | wc -l
486
grep -ir '\brequire\b' ~/.rvm/gems/ruby-2.3.0/gems/ --include="version.rb"
```
show that such gems in my system are:
- heroku-3.42.22
- rbnacl-3.3.0
- sass-3.2.19

i'd say in spec cases it's needed the author should be able modify the gemspec.

lib/bundler/templates/newgem/newgem.gemspec.tt
mfazekas: i'd argue adding a required_ruby_version would be usefull as most new gems would not be run checked against older ruby versions. Yes 2.1 is probably a better default.

jbodah: Any chance this could be configurable? Not sure if the rest of this can be driven from the `.bundle/config`, but that would seem like a good place to have it

deivid-rodriguez:
@mfazekas Are you still interested in getting this in? The `require_relative` change has been incorporated, but you can rebase this PR to introduce the `required_ruby_version` requirement if you want to.

deivid-rodriguez:
Closing in favor of #7209.

PR #4403, opened 2016-03-27
Fix `bundle outdated` behavior with combinations of `--major`, `--minor`, and `--patch` flags

RochesterinNYC:
- Allow `bundle outdated` to handle all combinations of `--major`, `--minor`, and `--patch` flags
- Related to discussion at #4249


indirect:
Awesome, thanks!

@homu r+

PR #4404, opened 2016-03-28
Backport #4393 and #4403 to 1.12

indirect:

PR #4407, opened 2016-03-28
Backport #4380 (Updated certificates) to `1-12-stable`

RochesterinNYC:
- Needs to be merged in before #4404 so ssl cert specs will pass


indirect:
@homu r+

indirect:
The certificate manager spec is failing? That's weird. :/ Is it because we removed the code that uses the check?

RochesterinNYC:
@indirect https://github.com/bundler/bundler/commit/28b86ac11c77cb199d65dfe8d7253f77a32f9668 is breaking this. This commit is only on `1-12-stable`.

indirect:
@RochesterinNYC we specifically need `lib/bundler/ssl_certs/GlobalSignRoot.pem` in order to be able to secure connections to Fastly. Should we just add that cert to RubyGems so the test will pass again?

RochesterinNYC:
Is that related to or done by rubygems/rubygems#1555 in the capacity you're thinking?

indirect:
Ah! For some reason I thought that was already merged. Once that goes in, I'll have homu retry here. Thanks!

RochesterinNYC:
@indirect Okay, I'll update this PR to add a revert for that commit then?

indirect:
@RochesterinNYC now that I've merged the RubyGems PR, that commit should pass, right? You can add it back. :)

RochesterinNYC:
@indirect Oh okay, I think we have a miscommunication here. So, the failing spec was a `CertificateManager` __unit__ spec, (not the `CertificateManager` integration/'other' spec that actually checks the Rubygems ssl certs on master or etc.). I think I misunderstood your comment above regarding adding a cert to Rubygems so that this failing test would pass.

The failing unit spec is failing because the setup for that unit test assumes certain things about the `ssl_certs` directory structure that are no longer true and the setup operations are polluted by the fact that there are certs that fit the glob `**/*.pem` and `*.pem` under `ssl_certs`. Cleanest way to rectify this: I can make a quick PR to master that will fix the unit spec setup and hence the unit spec, and then I can cherry pick that commit back to this PR?

indirect:
@RochesterinNYC okay, sounds good!

indirect:
@homu r+

RochesterinNYC:
@indirect I just added the cherry-picked commit from #4409 to this commit. I think the Homu r+ is approving the commit just before that cherry-picked commit though.

indirect:
Damn race conditions. :)

@homu r+

PR #4412, opened 2016-03-29
[Path] Speed up bin generation when there are many files

### LINKED ISSUE #2846: Big gems from git hang during every install doing... something

segiddins:
Closes #2846, I presume.

\c @indirect

### DIFF
--- a/lib/bundler/source/path.rb
+++ b/lib/bundler/source/path.rb
@@ -186,6 +186,7 @@ module Bundler
         # Some gem authors put absolute paths in their gemspec
         # and we have to save them from themselves
         spec.files = spec.files.map do |p|
+          next p unless p =~ /\A#{Pathname::SEPARATOR_PAT}/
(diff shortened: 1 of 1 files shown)


lib/bundler/source/path.rb
```diff
@@ -186,6 +186,7 @@ def generate_bin(spec, disable_extensions = false)
         # Some gem authors put absolute paths in their gemspec
         # and we have to save them from themselves
         spec.files = spec.files.map do |p|
+          next p unless p =~ /\A#{Pathname::SEPARATOR_PAT}/
```
indirect: `SEPARATOR_PAT`? really?

segiddins: this is how `Pathname#relative?` is implemented

indirect: lolsob, alright

indirect:
@homu r+ rollup

PR #4413, opened 2016-03-29
[Builders] Dont rebuild gem index when unnecessary

segiddins:
This should speed up the specs slightly

\c @indirect


indirect:
@homu r+ rollup

PR #4416, opened 2016-03-29
[Settings] Use a custom serializer instead of relying upon the YAML module

segiddins:
¯\\\_(ツ)\_/¯ seems like a good idea

\c @RochesterinNYC @indirect

### DIFF
--- a/lib/bundler/settings.rb
+++ b/lib/bundler/settings.rb
@@ -201,12 +201,19 @@ module Bundler
         hash.delete(key) if value.nil?
         SharedHelpers.filesystem_access(file) do |p|
           FileUtils.mkdir_p(p.dirname)
-          require "bundler/psyched_yaml"
-          File.open(p, "w") {|f| f.puts YAML.dump(hash) }
+          p.open("w") {|f| f.write(serialize_hash(hash)) }
         end
       end
 
       value
+    end
+
+    def serialize_hash(hash)
+      yaml = String.new("---\n")
+      hash.each do |key, value|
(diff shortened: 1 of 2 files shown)


segiddins:
W00t this passes!

lib/bundler/settings.rb
```diff
@@ -201,14 +201,21 @@ def set_key(key, value, hash, file)
         hash.delete(key) if value.nil?
         SharedHelpers.filesystem_access(file) do |p|
           FileUtils.mkdir_p(p.dirname)
-          require "bundler/psyched_yaml"
```
RochesterinNYC: ooo this is the start of something beautiful.

lib/bundler/settings.rb
```diff
@@ -201,14 +201,21 @@ def set_key(key, value, hash, file)
...
         end
       end
 
       value
     end
 
+    def serialize_hash(hash)
+      yaml = String.new("---\n")
+      hash.each do |key, value|
+        yaml << key << ": " << value.to_s.gsub(/\s+/, " ").inspect << "\n"
```
RochesterinNYC: There aren't any current cases where the `.bundle/config` would have or require something more complex than just simple key --> value mapping is there?

segiddins: it can't read anything else, so no?

segiddins:
@indirect r?

segiddins:
@indirect ping

segiddins:
@homu r+

PR #4417, opened 2016-03-29
[Installer] Don't eval gemfile twice

### LINKED ISSUE #3096: Gemfile is evaluated twice when a lockfile exists
```Ruby
#Gemfile
puts "WUT"
```

no Gemfile.lock -> WUT
with Gemfile.lock -> WUTWUT

tracked it down to this guy: https://github.com/bundler/bundler/blob/master/lib/bundler/installer.rb#L63-L71

I guess it should reuse the definition from https://github.com/bundler/bundler/blob/master/lib/bundler/cli/install.rb#L77-L78

not only `bundle install` also bundle exec ruby -e 1 evaluates the Gemfile twice, so basically every bundle run could save a few miliseconds

segiddins:
Closes #3096 

\c @indirect

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -46,14 +46,16 @@ module Bundler
     def initialize(lockfile, dependencies, sources, unlock, ruby_version = nil, optional_groups = [])
       @unlocking = unlock == true || !unlock.empty?
 
-      @dependencies    = dependencies
-      @sources         = sources
-      @unlock          = unlock
-      @optional_groups = optional_groups
-      @remote          = false
-      @specs           = nil
-      @ruby_version    = ruby_version
+      @dependencies     = dependencies
+      @sources          = sources
+      @original_sources = sources.dup
+      @unlock           = unlock
+      @optional_groups  = optional_groups
+      @remote           = false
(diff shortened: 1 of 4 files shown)


segiddins:
I don't know how to make this work, it turns out `.dup`ing a SourceList doesn't do what we need, and because there's hidden mutable state in `Source` and we need a matching source because `@dependencies` maintains references to sources and :sob:

segiddins:
Can anyone think of any possible avenues to make this work?

indirect:
I have previously spent days fighting this and eventually given up. I think we need less hidden mutable state for it to be possible?

segiddins:
@indirect well... yes. But a rewrite of how source / index / definition work isn't really feasible given time constraints right now

colby-swandale:
Im happy to work on this @segiddins if you're on other things.

segiddins:
@colby-swandale up to you -- it might just be frustrating though, so I can't wholeheartedly suggest it :/

segiddins:
Done in 2.0

lynnco:
Closed via https://github.com/bundler/bundler/pull/4952

PR #4419, opened 2016-03-31
Suggest usage of `--full-index` flag in case of Dependency API failure

### LINKED ISSUE #4384: Errors while using the dependency API should suggest --full-index
If Bundler experiences 3 errors in a row while trying to contact the dependency API, it will raise and exit. The error message should suggest trying `bundle install --full-index`, which downloads a static file instead of trying to use the dynamic dependency API, and almost always works around the problem while the API is having issues.

RochesterinNYC:
- Closes #4384

\cc @indirect Could use some review on this as to if the appropriate error message has been changed.


segiddins:
:+1:

segiddins:
@homu r+

PR #4425, opened 2016-04-03
Print sanitized urls with removed credentials for git sources on `bundle install`

RochesterinNYC:
If your environment includes `user1:password1` for your `GITHUB_CREDENTIALS` and you have something like the following in your `Gemfile`:

```
git "https://#{ENV['GITHUB_CREDENTIALS']}:x-oauth-basic@github.com/company/private-repo" do
  gem 'be_excellent', ref: '83623'
end
```
the output for `bundle install` is:

```
bundle
Using be_excellent 0.0.1 from https://user1:password1@github.com/company/private-repo (at master@53534b2)
```

This PR sanitizes this output and removes the credentials so the output would be:

```
bundle
Using be_excellent 0.0.1 from https://github.com/company/private-repo (at master@53534b2)
```

- Also handles oauth tokens (ex. `https://oauth_token:x-oauth-basic@github.com/company/private-repo`)

Related to: bundler/bundler-features#111

### DIFF
--- a/lib/bundler/uri_credentials_filter.rb
+++ b/lib/bundler/uri_credentials_filter.rb
@@ -3,7 +3,7 @@ module Bundler
   module URICredentialsFilter
   module_function
 
-    def anonymized_uri(uri_to_anonymize)
+    def credential_filtered_uri(uri_to_anonymize)
       return uri_to_anonymize if uri_to_anonymize.nil?
       uri = uri_to_anonymize.dup
       uri = URI(uri.to_s) unless uri.is_a?(URI)
@@ -16,15 +16,16 @@ module Bundler
         end
         uri.password = nil
       end
-      uri
+      return uri if uri_to_anonymize.is_a?(URI)
+      return uri.to_s if uri_to_anonymize.is_a?(String)
     rescue URI::InvalidURIError # uri is not canonical uri scheme
       uri
     end
 
-    def credentials_filtered_string(str_to_filter, uri)
+    def credential_filtered_string(str_to_filter, uri)
       return str_to_filter if uri.nil? || str_to_filter.nil?
       str_with_no_credentials = str_to_filter.dup
-      anonymous_uri_str = anonymized_uri(uri).to_s
+      anonymous_uri_str = credential_filtered_uri(uri).to_s
       uri_str = uri.to_s
       if anonymous_uri_str != uri_str
         str_with_no_credentials = str_with_no_credentials.gsub(uri_str, anonymous_uri_str)
--- a/lib/bundler/uri_credentials_filter.rb
+++ b/lib/bundler/uri_credentials_filter.rb
@@ -7,7 +7,15 @@ module Bundler
       return uri_to_anonymize if uri_to_anonymize.nil?
       uri = uri_to_anonymize.dup
       uri = URI(uri.to_s) unless uri.is_a?(URI)
-      uri.user = uri.password = nil if uri.userinfo
+      if uri.userinfo
+        # oauth authentication
+        if uri.password == "x-oauth-basic" || uri.password == "x"
+          # URI as string does not display with password if no user is set
+          oauth_designation = uri.password
+          uri.user = oauth_designation
+        end
+        uri.password = nil
+      end
       uri
     rescue URI::InvalidURIError # uri is not canonical uri scheme
       uri
--- a/lib/bundler/uri_credentials_filter.rb
+++ b/lib/bundler/uri_credentials_filter.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+module Bundler
+  module URICredentialsFilter
+  module_function
+
+    def anonymized_uri(uri_to_anonymize)
+      return uri_to_anonymize if uri_to_anonymize.nil?
+      uri = uri_to_anonymize.dup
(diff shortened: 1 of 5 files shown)


lib/bundler/uri_credentials_filter.rb
```diff
@@ -0,0 +1,27 @@
...
+      str_with_no_credentials = str_to_filter.dup
+      anonymous_uri_str = anonymized_uri(uri).to_s
+      uri_str = uri.to_s
+      if anonymous_uri_str != uri_str
+        str_with_no_credentials = str_with_no_credentials.gsub(uri_str, anonymous_uri_str)
+      end
+      str_with_no_credentials
+    end
+
+    module_function :anonymized_uri, :credentials_filtered_string
```
segiddins: This can go at the top of the scope just like `private` would

lib/bundler/uri_credentials_filter.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+module Bundler
+  module URICredentialsFilter
+    def anonymized_uri(uri_to_anonymize)
+      return uri_to_anonymize if uri_to_anonymize.nil?
+      uri = uri_to_anonymize.dup
+      uri = URI(uri.to_s) unless uri.is_a?(URI)
+      uri.user = uri.password = nil if uri.userinfo
```
segiddins: IMHO it should leave the user?

RochesterinNYC: In your opinion, what's the value add/UX value in leaving the user?

RochesterinNYC: Asking because I feel that user still fits the category of a credential when it comes to authentication.

RochesterinNYC:
Updated the PR with the `module_function` change. Discussion still pending on the user as a credential aspect.

lib/bundler/uri_credentials_filter.rb
segiddins: I think some gem servers serve different content to each user at the endpoint

indirect: AFAIK, servers like GemFury serve different gems to different users. It's an edge case, but this seems worth it, unless the "user" is a password token of some kind? :/

RochesterinNYC: Hm, okay, fair enough. Any suggestions on how to handle the `oauth_token:x-oauth-basic@` format for authentication and removing the `oauth_token` part? Removing just the password on this credential would render it `oauth_token@` instead of `x-oauth-basic@` as the `URI` module detects `oauth_token` as the user. 

Is `oauth_token:x-oauth-basic@` a valid authentication usage/format for git sources for Bundler? I don't have the first-hand context on this and am basing this oauth example off of the discussion at bundler/bundler-features#111

If `x-oauth-basic` will always be the "URI password" part of the uri for oauth authentication, then I could handle both regular login credentials and oauth credentials. I'm seeing a lot of sources that support the idea that this is the right assumption, though I'm not 100% sure.

indirect: Ugh; that's a good point. I have seen basic auth with either 'x' or 'x-oauth-basic' as the password, so maybe we can check for that.

RochesterinNYC: Okay, that works. Although, I don't think there's a way to differentiate between a user using 'x' or 'x-oauth-basic' as their login/authentication password vs. them using that to specify they want to authenticate using oauth. Though I guess maybe we can just completely hide user and password sections if password is 'x' or 'x-oauth-basic'? This should cover the vast majority of credential screening cases.

RochesterinNYC: On second thought, I think filtering out the password by default, and the user section (actually oauth token) instead of password when the password is 'x' or 'x-oauth-basic' covers all edge cases. I didn't consider the idea that the gem or git server or etc. wouldn't really allow regular login credentials that include 'x' or 'x-oauth-basic' as the password, because the server would try to authenticate using oauth in these cases. (i.e. `username123:x-oauth-basic@` would be authenticated as oauth where `username123` is the oauth token.)

segiddins:
:+1:

RochesterinNYC:
This PR's been updated based on the discussion regarding showing user information. 

Description of the new behavior:

If your environment includes `user1:password1` for your `GITHUB_CREDENTIALS` and you have something like the following in your `Gemfile`:

```
git "https://#{ENV['GITHUB_CREDENTIALS']}:x-oauth-basic@github.com/company/private-repo" do
  gem 'be_excellent', ref: '83623'
end
```
the output for `bundle install` is:

```
bundle
Using be_excellent 0.0.1 from https://user1:password1@github.com/company/private-repo (at master@53534b2)
```

This PR sanitizes this output and removes the sensitive credentials (password or oauth token) so the output would be:

```
bundle
Using be_excellent 0.0.1 from https://user1@github.com/company/private-repo (at master@53534b2)
```

- Also handles oauth tokens:

(ex. `https://oauth_token:x-oauth-basic@github.com/company/private-repo` --> `https://x-oauth-basic@github.com/company/private-repo`)

(ex. `https://oauth_token:x@github.com/company/private-repo` --> `https://x@github.com/company/private-repo`)

segiddins:
@homu r+

PR #4439, opened 2016-04-14
Handle recursive bundle package loop when multiple gemspecs present

### LINKED ISSUE #4430: bundle package recursive loop when using multiple gemspecs
I am attempting to run 'bundle package' during a deploy of our repo, but when we do we get a recursion loop during the cp_r ending with the error: "Errno::ENAMETOOLONG: File name too long @ rb_sysopen - /Users/me/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/localizer/vendor/cache/support_rb-4a8c0775d668/.git/objects/pack/pack-13658d6ae3c302c3932bfb35ee087a28ed30533a.idx"

I have narrowed the problem down to our second gemspec in our repo that we use for a client library. When we leave it in bundler wants to package our repo inside of itself and it falls into an recursion loop until the filename gets too long and fails. Fortunately the second second '-client' gemspec is not necessary for deployments and 'bundle package' works when I comment out that line, but I have not found any way to 'bundle package' without commenting it out.

The ideal way to solve this is to have bundler look at the path of the gem it wants to package and reject it if it is the same as the project root directory. I welcome other suggestions for temporary fixes until then.

Example Gemfile:
```
source 'https://rubygems.org'
gemspec name: 'localizer'
gemspec name: 'localizer-clients'
...
```

RochesterinNYC:
`root_gem_names` should contain the names of the gems that the gemspecs in the directory are correlated with. Gems are packaged unless they're called `bundler` or their names match with the gems specified in the gemspecs.

- Handles multiple gemspecs in directory
- Handles case where gemspec file name does not match with the gem name specified in the gemspec
- Fixes #4430

### DIFF
--- a/lib/bundler/runtime.rb
+++ b/lib/bundler/runtime.rb
@@ -127,16 +127,16 @@ module Bundler
 
       Bundler.ui.info "Updating files in #{Bundler.settings.app_cache_path}"
 
-      # Do not try to cache specification for the gem described by the .gemspec
-      root_gem_name = nil
+      # Do not try to cache specification for the gem described any of the gemspecs
+      root_gem_names = nil
       if gemspec_cache_hash = Bundler.instance_variable_get(:@gemspec_cache)
-        gemspec = gemspec_cache_hash.values.first
-        root_gem_name = gemspec.name unless gemspec.nil?
+        gemspecs = gemspec_cache_hash.values
+        root_gem_names = gemspecs.map(&:name)
       end
 
       specs.each do |spec|
         next if spec.name == "bundler"
-        next if !Dir.glob("*.gemspec").empty? && spec.source.class == Bundler::Source::Path && root_gem_name && spec.name == root_gem_name
+        next if !Dir.glob("*.gemspec").empty? && spec.source.class == Bundler::Source::Path && root_gem_names.include?(spec.name)
         spec.source.send(:fetch_gem, spec) if Bundler.settings[:cache_all_platforms] && spec.source.respond_to?(:fetch_gem, true)
         spec.source.cache(spec, custom_path) if spec.source.respond_to?(:cache)
       end
--- a/spec/commands/package_spec.rb
+++ b/spec/commands/package_spec.rb
@@ -95,6 +95,51 @@ describe "bundle package" do
         end
       end
     end
+
+    context "with multiple gemspecs" do
+      before do
+        File.open(bundled_app("mygem.gemspec"), "w") do |f|
+          f.write <<-G
+            Gem::Specification.new do |s|
+              s.name = "mygem"
+              s.version = "0.1.1"
+              s.summary = ""
+              s.authors = ["gem author"]
+              s.add_development_dependency "nokogiri", "=1.4.2"
+            end
+          G
+        end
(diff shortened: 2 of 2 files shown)


lib/bundler/runtime.rb
```diff
@@ -127,16 +127,16 @@ def cache(custom_path = nil)
...
 
-      # Do not try to cache specification for the gem described by the .gemspec
-      root_gem_name = nil
+      # Do not try to cache specification for the gem described any of the gemspecs
+      root_gem_names = nil
       if gemspec_cache_hash = Bundler.instance_variable_get(:@gemspec_cache)
-        gemspec = gemspec_cache_hash.values.first
-        root_gem_name = gemspec.name unless gemspec.nil?
+        gemspecs = gemspec_cache_hash.values
+        root_gem_names = gemspecs.map(&:name) unless gemspecs.empty?
```
segiddins: the `unless` seems unnecessary to me?

segiddins:
👍 other than the one comment

lib/bundler/runtime.rb
RochesterinNYC: oh, `gemspec_cache_hash.values` will return an empty array if the hash is empty. Good call. Will remove the `unless`.

RochesterinNYC:
Thanks for the catch. Removed the extraneous `unless`.

spec/commands/package_spec.rb
```diff
@@ -95,6 +95,52 @@
...
+      end
+
+      it "caches all dependencies except bundler and the gemspec specified gems" do
+        gemfile <<-D
+          source "file://#{gem_repo1}"
+          gem 'rack'
+          gemspec :name => 'mygem'
+          gemspec :name => 'mygem_client'
+        D
+        sleep 10
```
segiddins: 🔥🔥🔥🔥

RochesterinNYC: Hm, not sure how that got in since I don't think it was there on the previous iterations of this PR. Major thanks for the catch though. Think it's worth adding a spec quality check for this?

indirect:
A quality spec that checks for focused tests, binding.pry, debugger, and sleep sounds awesome to me. :)

segiddins:
Awesome but separate :P

segiddins:
@homu r+

PR #4443, opened 2016-04-15
Update bundle install man page with --frozen option

Nitrodist:
This option has existed since v1.1 and we're currently on v1.11!


segiddins:
👍

segiddins:
@homu r+

PR #4445, opened 2016-04-16
Allow conflicts for gems resolved via `gemspec`

segiddins:
Needs to be cleaned up, but this demonstrates the general idea

### DIFF
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -67,16 +67,16 @@ module Bundler
             "#{file}. Make sure you can build the gem, then try again"
         end
 
+        @gemspecs << spec
+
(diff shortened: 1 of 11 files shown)


segiddins:
@indirect thoughts on this?

indirect:
I like refactoring out Source::Gemspec! If I am understanding the feature, I think I'm okay with it. Can you add brief documentation for what this change allows, with at least one example?

segiddins:
Where would documentation for this behavior go? In the Gemfile docs?

indirect:
wherever we document `gemspec` seems reasonable, so Gemfile(5)?

segiddins:
@indirect added docs

indirect:
@homu r+

Issue #4437, opened 2016-04-13
Pushing to 'allowed_push_host' without rubygems.org credentails fails


callahat:
Without a `rubygems_api_key` a gem cannot be published to the host specified by the `allowed_push_host`. I was able to verify that with a valid key in `~/.gem/credentials` I am able to publish instead to the geminabox server (and it doesn't wind up on rubygems.org). Line is here: https://github.com/bundler/bundler/blob/master/lib/bundler/gem_helper.rb#L95

Cohen-Carlisle:
It would be nice to have this addressed.
Its a bit scary when you are prompted for Rubygems credentials when you are trying to push to a private gemserver.
I can also envision a situation in which someone sets their rubygem creds to be able to push to a private gemserver and then later forgets to set `allowed_push_host` in some gem and ends up pushing potentially proprietary code to a public gemserver.
:+1:

Cohen-Carlisle:
It looks like simply moving the linked line down to after the `if @gemspec.respond_to?(:metadata)`'s `end` and checking `allowed_push_host` as well would work. I.e.,
```ruby
def rubygem_push(path)
  allowed_push_host = nil
  gem_command = "gem push '#{path}'"
  if @gemspec.respond_to?(:metadata)
    allowed_push_host = @gemspec.metadata["allowed_push_host"]
    gem_command += " --host #{allowed_push_host}" if allowed_push_host
  end
  unless allowed_push_host || Pathname.new("~/.gem/credentials").expand_path.file?
    raise "Your rubygems.org credentials aren't set. Run `gem push` to set them."
  end
  sh(gem_command)
  Bundler.ui.confirm "Pushed #{name} #{version} to #{allowed_push_host ? allowed_push_host : "rubygems.org."}"
end
```

segiddins:
@Cohen-Carlisle a PR for that would be 🚀

Cohen-Carlisle:
What branch(es?) should a PR addressing this issue be against?

segiddins:
Master

PR #4446, opened 2016-04-16
don't error on missing rubygem cred if custom host

Cohen-Carlisle:
only raise error asking user to set rubygem credentials during gem push
if `allowed_push_host` is not set.

Fixes https://github.com/bundler/bundler/issues/4437

Apologies if this is a half baked PR.
- I looked around to add tests for this, but there is little test coverage over this area I simply wasn't sure how to add a test around this particular change.
- I also wasn't sure where to put this in the `CHANGELOG` as the latest entry there is already released.

### DIFF
--- a/lib/bundler/gem_helper.rb
+++ b/lib/bundler/gem_helper.rb
@@ -92,14 +92,14 @@ module Bundler
   protected
 
     def rubygem_push(path)
-      unless  Pathname.new("~/.gem/credentials").expand_path.file?
-        raise "Your rubygems.org credentials aren't set. Run `gem push` to set them."
-      end
       allowed_push_host = nil
       gem_command = "gem push '#{path}'"
       if @gemspec.respond_to?(:metadata)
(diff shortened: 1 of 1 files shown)


segiddins:
👍🏻

Cohen-Carlisle:
Also had some thoughts while doing this:
- The gemserver I'm pushing to doesn't require credentials so I'm not sure what happens if a gemserver requires them and they aren't provided. We might want something to handle that?
- Sometimes my brain hurts when trying to read `unless foo || bar`. Should it be `if !foo && !bar`?

segiddins:
`gem push` ought to fail with a helpful error anyways? And feel free to change the conditional, I don't have much of a preference

Cohen-Carlisle:
I'm fine with the existing conditoinal. Is there anything else I should do for this PR?

segiddins:
@homu r+

PR #4448, opened 2016-04-18
Update Bundler ssl certs to match updated Rubygems ssl certs/structure

RochesterinNYC:
This PR should only be merged if and after rubygems/rubygems#1583 is merged. Reasoning behind the PR is originally stated there but is as follows:

The current arrangement of the ssl certs for Bundler (and Rubygems) was produced through manual testing and configured/arranged in bundler/bundler#4380 instead of programmatically. rubygems/rubygems#1583 updates the Rubygems util script for updating the ssl certs to programmatically write the certs to the appropriate nested directories. This PR ensures Bundler's certs match Rubygems' after they are programmatically generated/arranged instead of manually.


RochesterinNYC:
I think this PR can be merged in now that the related/above-mentioned Rubygems PR has been merged in. This should ensure the certs match between the two repos.

segiddins:
@homu r+

PR #4451, opened 2016-04-20
before/after install hooks support

lxsameer:
### Why:
Running a code block at specific moments of bundler activities like dependency installation is a feature which allows users to customize the their environment based on bundler and installed gems.

### What are the new features:
  * `after_install` method added to `Bundler` dsl
  * `before_install` method added to `Bundler` dsl

related to: https://github.com/bundler/bundler-features/issues/118

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.12.0.rc.4 (2016-04-20)
+
+Features:
+  - Support for `before_install` and `after_install` hooks added.
+	
 ## 1.12.0.rc.3 (2016-04-19)
 
 Bugfixes:
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -30,6 +30,7 @@ module Bundler
   autoload :GemHelper,              "bundler/gem_helper"
   autoload :GemHelpers,             "bundler/gem_helpers"
   autoload :Graph,                  "bundler/graph"
+  autoload :Hooks,                  "bundler/hooks"
   autoload :Index,                  "bundler/index"
   autoload :Installer,              "bundler/installer"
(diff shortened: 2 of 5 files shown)


segiddins:
I've implemented basically this before and we decided not to merge it

lxsameer:
Why? It's small, handy and harmless

segiddins:
https://github.com/bundler/bundler/pull/3471

RochesterinNYC:
@segiddins is there a past PR/issue that can be referenced?

@lxsameer I'm adding a few questions regarding use cases and why/if this is a necessary feature to the issue that you opened in bundler-features. Let's discuss them there.

indirect:
We have somewhat more notes written up here: https://trello.com/c/LJUJNMGN. RubyGems already has before-install hooks, and we already respect them, so the discussion is around implementing a more flexible hook system that has slightly more protection against things like exceptions inside hooks.

lxsameer:
@indirect This solution is very simple and easy to use specially for end users. Also exception handling can be added to this solution easily. I think the most important point in this solution is simplicity. there is almost no overhead and implementing new hook can be done in no time.

It's an easier solution rather than rubygems hook and suitable for end users for tooling and managing tasks.

indirect:
@lxsameer this gem already automatically generates ctags for gems as they are installed by Bundler, using the RubyGems install hooks that Bundler already supports: https://github.com/tpope/gem-ctags. Do you have any other use-cases for this kind of hook?

indirect:
Ugh, I left this comment on the PR when I meant to leave it on the feature request. Sorry, I'll go over there now...

lxsameer:
@homu done

PR #3842, opened 2015-07-17
Drop Travis builds for Ruby versions below 2.0.0

smlance:

PR #4456, opened 2016-04-22
do not log the credentials used to contact a gem server

sandlerr:
Currently a network error results in the credentials being output in deploy logs, for example.

Not sure if there is a better approach or this needs to be changed in more locations.

### DIFF
--- a/lib/bundler/errors.rb
+++ b/lib/bundler/errors.rb
@@ -30,7 +30,12 @@ module Bundler
   class GemspecError < BundlerError; status_code(14); end
   class InvalidOption < BundlerError; status_code(15); end
   class ProductionError < BundlerError; status_code(16); end
-  class HTTPError < BundlerError; status_code(17); end
+  class HTTPError < BundlerError
+    status_code(17)
+    def filter_uri(uri)
+      URICredentialsFilter.credential_filtered_uri(uri)
+    end
+  end
   class RubyVersionMismatch < BundlerError; status_code(18); end
   class SecurityError < BundlerError; status_code(19); end
   class LockfileError < BundlerError; status_code(20); end
--- a/lib/bundler/fetcher.rb
+++ b/lib/bundler/fetcher.rb
@@ -19,6 +19,7 @@ module Bundler
     # This is the error raised if OpenSSL fails the cert verification
     class CertificateFailureError < HTTPError
       def initialize(remote_uri)
+        remote_uri = filter_uri(remote_uri)
         super "Could not verify the SSL certificate for #{remote_uri}.\nThere" \
           " is a chance you are experiencing a man-in-the-middle attack, but" \
           " most likely your system doesn't have the CA certificates needed" \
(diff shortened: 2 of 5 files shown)


segiddins:
@RochesterinNYC didn't we recently merge a PR that had a method to do this?

segiddins:
This should use `URICredentialsFilter`, see https://github.com/bundler/bundler/pull/4425/

sandlerr:
Thanks for the feedback @segiddins ! I've pushed another commit using that module, but I noticed this behaviour in many subclasses of `HTTPError` in https://github.com/bundler/bundler/blob/master/lib/bundler/fetcher.rb

A more DRY solution for those error classes would probably involve an `alias` of the constructor.

segiddins:
👍 with tests

lib/bundler/uri_credentials_filter.rb
```diff
@@ -5,7 +5,7 @@ module URICredentialsFilter
 
     def credential_filtered_uri(uri_to_anonymize)
       return uri_to_anonymize if uri_to_anonymize.nil?
-      uri = uri_to_anonymize.dup
+      uri = uri_to_anonymize.clone
```
segiddins: this seems to have caused issues

segiddins:
Sorry, I didn't see you pushed new commits :(

lib/bundler/uri_credentials_filter.rb
RochesterinNYC: I'm generally against the principle of altering application code purely for the sake of testing. What necessitates this change from `dup` to `clone`? Is it possible to replace the mocks in whichever tests this is affecting with actual URI objects?

sandlerr:
Haven't got the patience to wait for them at the moment but assuming CI is green I think this is ready for another look.
Sorry to have taken so long to get back to this! I have to admit seeing the old Ruby versions still being supported made it less appealing to finish this off as it either means waiting for Rubies to install or CI runs to complete, so great that you've merged https://github.com/bundler/bundler/pull/3842 and hope that branch becomes master soon.

allenzhao:
I assume this is not related to what I'm doing at https://github.com/bundler/bundler/issues/4663?
And I think you should probably add assertion that check both the err and the out doesn't contain the password since the one I'm working with actually throws error with credentials in the `err`

sandlerr:
@allenzhao this seems like a separate instance of a similar problem to the one you're working on, so there shouldn't be any duplication of efforts. A key difference in the tests seems to be that the ones that are touched in this branch are unit tests - so I'm simply asserting on the value of an error message. The tests affected by your issue are more integration-level tests so they make assertions on the entire output of the program.
As far as this branch is concerned, I'm not sure if the eventual error message even goes to STDOUT or STDERR, and certainly haven't made any changes to the component that finally catches the errors as they bubble up the stack.

segiddins:
thanks @sandlerr !

sandlerr:
thanks for maintaining! @segiddins

RochesterinNYC:
? Was this merged without Homu?

segiddins:
@RochesterinNYC see slack, but homu is down

RochesterinNYC:
Gotcha, just saw the message.

PR #4465, opened 2016-04-25
Only show potential updates on same platform for `bundle outdated`

### LINKED ISSUE #4450: Bundle outdates gives incorrect results sometimes
I noticed this strange problem.

With this `Gemfile`

```ruby
source 'https://rubygems.org'

gem 'pry', github: 'pry'
```

all good...

```shell
$ bundle outdated 
Fetching git://github.com/pry/pry.git
Resolving dependencies...

Bundle up to date!
```

But with this other Gemfile

```ruby
source 'https://rubygems.org'

gem 'pry', github: 'pry'
gem 'travis', '~> 1.8'
```

`pry` gem gets incorrectly reported as outdated.

```shell
$ bundle outdated
Fetching git://github.com/pry/pry.git
Fetching version metadata from https://rubygems.org/...
Fetching dependency metadata from https://rubygems.org/..
Resolving dependencies...

  * pry (newest 0.10.3, installed 0.10.3 f5c97ca) in group "default"
  * typhoeus (newest 1.0.1, installed 0.8.0)
```

Refs: https://github.com/pry/pry/pull/1510

RochesterinNYC:
This fixes the following behavior:

Say you're using a gem `laduradura` on platform `java` with version `v1.0.1`. The latest `java` platform version of this gem is `v1.0.1` but the latest `ruby` platform version of this gem is `v1.0.2`. Running `bundle outdated` will currently tell you that you can update `laduradura` to `v1.0.2`.

This is a bad user experience as the user is given the suggestion to update `laduradura` to `v1.0.2`, despite this version being on a completely different platform than the one that the user is currently using `laduradura` on.

With this PR, `bundle outdated` will only report potential version updates to gems for the same platforms those gems are being used on.

- Fixes #4450

### DIFF
--- a/lib/bundler/cli/outdated.rb
+++ b/lib/bundler/cli/outdated.rb
@@ -50,11 +50,11 @@ module Bundler
         if options["strict"]
           active_spec = definition.specs.detect {|spec| spec.name == current_spec.name }
         else
-          active_spec = definition.index[current_spec.name].sort_by(&:version)
-          if !current_spec.version.prerelease? && !options[:pre] && active_spec.size > 1
-            active_spec = active_spec.delete_if {|b| b.respond_to?(:version) && b.version.prerelease? }
+          active_specs = definition.index[current_spec.name].sort_by(&:version)
+          if !current_spec.version.prerelease? && !options[:pre] && active_specs.size > 1
+            active_spec = active_specs.delete_if {|b| b.respond_to?(:version) && b.version.prerelease? }
           end
-          active_spec = active_spec.last
+          active_spec = active_specs.last
 
           if options[:major] || options[:minor] || options[:patch]
             update_present = update_present_via_semver_portions(current_spec, active_spec, options)
--- a/lib/bundler/cli/outdated.rb
+++ b/lib/bundler/cli/outdated.rb
@@ -48,9 +48,9 @@ module Bundler
         dependency = current_dependencies[current_spec.name]
 
         if options["strict"]
-          active_spec = definition.specs.detect {|spec| spec.name == current_spec.name }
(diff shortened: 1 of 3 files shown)


segiddins:
How does this handle when a gemfile includes multiple platforms?

RochesterinNYC:
Hm, as in the case where `laduradura` is being used in the same `Gemfile` under both `ruby` and `java` platforms? We would expect/want to see `bundle outdated` show that there is a new possible update for `laduradura`?

RochesterinNYC:
Confirmed that in the case of a gem `laduradura` being used for both `ruby` and `java` platforms, with an update available for the gem on `java` platform, `bundle outdated` will not show any potential updates. However, it will show an update available if the update is available on the `ruby` platform. I'll work on a fix/more robust behavior for `bundle outdated` regarding multiple platforms soon.

deivid-rodriguez:
@RochesterinNYC Do you know of any gems that actually release specific versions for specific platforms? I was thinking that all the "multiplatform gems" I've seen around release every version for every platform.

Maybe the current fix is good enough to improve the current situation?

RochesterinNYC:
I'm inclined to agree and advocate holding off on trying to fix a problem that we might not actually have. I can create an issue regarding it, but I think that should be separate from this PR/fix. Thoughts? @segiddins @indirect

segiddins:
@RochesterinNYC I actually see no issue with this PR

RochesterinNYC:
Okay, could you queue up Homu? I still don't have access yet/don't think we should r+ our own PRs?

segiddins:
@homu delegate=RochesterinNYC

RochesterinNYC:
Awesome, thanks. Guessing you can't add permanent access via comments to homu?

@homu r+

Issue #4472, opened 2016-04-27
Checksum of /versions does not match the checksum provided by server


perlun:
When using `bundler-1.12.0.rc.4`, I get this error when doing a `bundle install`:

```
$ bundle install
Fetching gem metadata from https://gem.fury.io/ecraft-gems/.....
Fetching version metadata from https://gem.fury.io/ecraft-gems/...
Fetching dependency metadata from https://gem.fury.io/ecraft-gems/..
Checksum of /versions does not match the checksum provided by server! Something
is wrong.
```

As can be seen, I am using a 3rd party gem source. (Gemfury)

When using the released `bundler 1.11.2`, it works properly:

```
$ bundle install
Warning: the running version of Bundler is older than the version that created t
he lockfile. We suggest you upgrade to the latest version of Bundler by running
`gem install bundler --pre`.
Fetching gem metadata from https://gem.fury.io/ecraft-gems/.....
Fetching version metadata from https://gem.fury.io/ecraft-gems/...
Fetching dependency metadata from https://gem.fury.io/ecraft-gems/..
Fetching gem metadata from https://gem.fury.io/ecraft-gems/.....
Fetching gem metadata from https://rubygems.org/...........
Fetching version metadata from https://gem.fury.io/ecraft-gems/...
Fetching version metadata from https://rubygems.org/...
Fetching dependency metadata from https://gem.fury.io/ecraft-gems/..
Fetching dependency metadata from https://rubygems.org/..
Resolving dependencies....
Installing rake 11.1.1
[...]
```

Any ideas what could be causing this? /cc @rallenecraft

indirect:
@perlun a bug in either the server or client for the new index. Thanks for reporting this!

indirect:
@perlun can you provide some more information, so that we can try to reproduce this issue? The output from `bundle install --verbose` would be very helpful. When I try to install gems from my own GemFury account, everything works fine.

indirect:
It would also be very helpful to see

indirect:
It would also be very helpful to see what, exactly, is being returned for you when you use your GemFury credentials to access the url `https://gem.fury.io/ecraft-gems/versions`. For my account, that is a 404. It seems that it must not be a 404 for your account, and I don't understand how that's happening...

mwrock:
[...]

PR #4477, opened 2016-04-29
[Updater] Raise a more helpful error on checksum mismatch

segiddins:
@indirect this will make diagnosing https://github.com/bundler/bundler/issues/4472 much easier.


indirect:
@homu r+

PR #4487, opened 2016-04-29
Show executable's man page when --help or -h is present

### LINKED ISSUE #4480: bundle exec CMD --help opens bundler man-page
When running `bundle exec rspec --help` on bundler 1.11.2 it displays the help output of RSpec, as expected. 

Running the same command on bundler 1.12.0 opens the bundler-exec man page and completely ignores the specified command.

For some reason it works for the `bundler` command, only `bundle` is affected.

Some more examples:

```bash
# these work as expected
bundle exec rspec --version
bundle exec rspec -- --version
bundle exec rspec
bundler exec rspec --help

# these don't work
bundle exec rspec --help
bundle exec rspec -- --help
bundle exec this_will_never_be_executed_anyway --help
bundle exec this_will_never_be_executed_anyway -- --help
```

To reproduce it:
```dockerfile
FROM ubuntu:16.04
RUN apt-get update && apt-get install -y ruby
RUN gem install bundler -v 1.12.0
```

```bash
docker build -t bundler-test .

# works as expected (says no Gemfile)
docker run bundler-test bundler exec foobar --help

# opens the man-page
docker run bundler-test bundle exec foobar --help
```

b-ggs:
Fixes #4480, where `bundle exec foo --help` shows `bundle exec`'s man instead of `foo`'s man.

@RochesterinNYC

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -437,22 +437,37 @@ module Bundler
     # into the corresponding `bundle help #{command}` call
     def self.reformatted_help_args(args)
       bundler_commands = all_commands.keys
-      help_flags = args.select {|a| %w(--help -h).include? a }
+      help_flags = %w(--help -h)
+      exec_commands = %w(exec e)
+      help_used = args.select {|a| help_flags.include? a}
+      exec_used = args.select {|a| exec_commands.include? a}
       command = args.select {|a| bundler_commands.include? a }
-      other_flags = args.select {|a| a.include? "--" } - help_flags
-      executables = args - command - help_flags - other_flags
-      new_args = []
-      if command.empty?
+      if exec_used.any? && help_used.any?
+        regex = %r{
+            ( # Matches `exec --help` or `exec --help foo`
+              (#{exec_commands.join("|")})
+              \s
+              (#{help_flags.join("|")})
+              (.+)*
+            )|
+            ( # Matches `--help exec` or `--help exec foo`
+              (#{help_flags.join("|")})
+              \s
+              (#{exec_commands.join("|")})
+              (.+)*
+            )
+          }x
+        arg_str = args.join(" ")
+        if arg_str =~ regex
+          ["help", "exec"]
+        else
+          args
+        end
+      elsif command.empty?
         abort("Could not find command \"#{args.join(" ")}\".")
-      elsif executables.any?
-        new_args << command.first
-        new_args += executables
-        new_args += help_flags
-        new_args += other_flags
       else
-        new_args += ["help", command.first]
+        ["help", command.first]
       end
-      new_args
     end
 
   private
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -437,7 +437,7 @@ module Bundler
     # into the corresponding `bundle help #{command}` call
     def self.reformatted_help_args(args)
       bundler_commands = all_commands.keys
-      help_flags = %w(--help -h)
+      help_flags = args.select {|a| %w(--help -h).include? a }
       command = args.select {|a| bundler_commands.include? a }
       other_flags = args.select {|a| a.include? "--" } - help_flags
       executables = args - command - help_flags - other_flags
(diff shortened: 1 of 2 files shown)


lib/bundler/cli.rb
```diff
@@ -437,13 +437,22 @@ def env
...
+      other_flags = args.select {|a| a.include? "--" } - help_flags
+      executables = args - command - help_flags - other_flags
+      new_args = []
       if command.empty?
         abort("Could not find command \"#{args.join(" ")}\".")
+      elsif executables.any?
+        new_args << command.first
+        new_args += executables
+        new_args << "--help"
+        new_args += other_flags
```
RochesterinNYC: This seems a bit overly complex, in that it's basically reparsing the entire command and putting it back together. I think simply making the `exec` command an exception for now is satisfactory and simplest?

RochesterinNYC:
Thanks for the PR and test coverage @b-ggs! Please squash these commits. You don't need to squash commits for PRs in general (provided each commit is a logical component of changes), but for Rubocop specific commits, please squash those down. Thanks!

lib/bundler/cli.rb
b-ggs: I assume this is how you're suggesting it would be done:
```
      if command.empty?
        abort("Could not find command \"#{args.join(" ")}\".")
      elsif command.include? "exec"
        args
      else
        ["help", command.first]
      end
```
However only making an exception for `exec` makes `bundle exec --help` throw `bundler: command not found: --help`.

RochesterinNYC: Okay that's a good point. I'd suggest adding a spec covering that case to the PR.  It wasn't clear from the specs you added why your implementation had to be as complex/specific as it was, but your explanation makes sense, great call.

b-ggs:
Added additional specs to cover the possible usages of `exec

spec/commands/exec_spec.rb
```diff
@@ -218,6 +218,57 @@
...
+    end
+
+    it "shows executable's man page when --help is after the executable" do
+      bundle "exec echo --help"
+      expect(out).to include("Usage: echo [SHORT-OPTION]... [STRING]...")
+    end
+
+    it "shows executable's man page when --help is between exec and the executable" do
+      bundle "exec --help echo"
+      expect(out).to include("Usage: echo [SHORT-OPTION]... [STRING]...")
```
RochesterinNYC: I think I would expect this one to show the help/man page for `bundle exec` as it's like you were running `bundle exec --help (echo)`. And the standalone command (without bundle exec) would essentially be just echo. I.e. running  `--help echo` is not a valid command and does not show the help for echo.

segiddins:
This also needs to handle `exec` being invoked with only a prefix, as in `bundle e`

b-ggs:
Because I was originally only checking the `args` array for the presence of `exec` and `--help` or `-h`, the following would occur...
```
$ bundle exec --help echo
Usage: echo [SHORT-OPTION]... [STRING]...  [...]
```
...since my implementation didn't take into account the position of the help tag.

To remedy this, I used a regex that matches `exec --help`, `exec --help foo`, `--help exec`, `--help exec echo`, and its variations<sup>[1]</sup> in order to point args formatted like this to the man for `bundle-exec`

Because of this, only `exec foo --help`, and its variations<sup>[1]</sup> would point to the executable `foo`'s man page.

[1] - Variations such as `e` instead of `exec` or `-h` instead of `--help`

--

I also opted to use `cat` instead of `echo` for the tests, since `echo` doesn't have `--no-color` flag support, was being passed during tests.
```
$ echo --help
Usage: echo [SHORT-OPTION]... [STRING]...  [...]
$ echo --help --no-color
--help --no-color
```
```
$ cat --help
Usage: cat [OPTION]... [FILE]...  [...]
$ cat --help --no-color
Usage: cat [OPTION]... [FILE]...  [...]
```
--
@RochesterinNYC @segiddins Thanks, guys!

RochesterinNYC:
@homu r+

RochesterinNYC:
@homu r+

RochesterinNYC:
This is good to merge. I don't have Homu access so I can't r+.

segiddins:
@homu r+

Issue #4488, opened 2016-04-29
bundle exec makes $0==__FILE__ comparison fail in v1.12.0


manlon:
I have a script that relies on the `$0 == __FILE__` trick to determine if it is being run at the top level:
```
$ cat test.rb
#!/usr/bin/env ruby
if $0 == __FILE__
  puts "ok #{$0}"
else
  puts "mismatch! #{$0} != #{__FILE__}"
end
```
With bundler v1.11.2
```
$ bundle exec ./test.rb                                                                                                                                       
ok ./test.rb
```
With 1.12.0
```
$ bundle exec ./test.rb
mismatch! ./test.rb != /home/vagrant/test.rb
````
I.e., `__FILE__` is expanded, breaking `bundle exec` for such scripts.

**Comment by segiddins** [comment]
Ah so we need to File.expand_path before setting `$0`. Should be an easy fix

**Comment by manlon** [comment]
Expanding $0 would fix my particular issue, but note the previous behavior was that the values are the same as the way the script is invoked from the command line (as it is for a ruby script without bundler). Not sure what else might depend on this behavior.

**Comment by segiddins** [comment]
Ah gotcha... it's because we're `load`ing the absolute path :/

**Comment by indirect** [comment]
I think we need the path to be relative from `Dir.pwd` at exec time?

**Comment by segiddins** [comment]
So I dug into this, and it only seems to happen when you invoke with a `./`. I have no idea how to fix it, since it's `Kernel.load` itself that's turning that into an absolute path

**Comment by segiddins** [comment]
This is pretty much unfixable, see https://github.com/bundler/bundler/issues/4491 for reasons why (also, discrepancies between ruby versions complicate things)

**Comment by yob** [comment]
We ran into this today too, so I thought I'd add another use case to the discussion.

We still use delayed job with a few rails apps, and use the process names in our monitoring setup to ensure the worker processes are running. The worker processes are started with `bundle exec rake jobs:work`

With bundler 1.11.2, the process name looked like this:

```
$ ps ux | grep rake
jh        1758 62.2  2.4 385816 202032 pts/3   Sl+  16:16   0:04 ruby /home/jh/.rbenv/versions/2.2.4/bin/rake jobs:work
```

With bundler 1.12.1, the process name looks like this:

```
$ ps ux | grep rake
jh        2036 78.8  2.1 356688 172660 pts/3   Sl+  16:16   0:03 /home/jh/.rbenv/versions/2.2.4/bin/rake
```
[...]

PR #4491, opened 2016-04-30
[Exec] Add specs for $0 and __FILE__ when loaded

segiddins:
Tests for #4488

### DIFF
--- a/spec/commands/exec_spec.rb
+++ b/spec/commands/exec_spec.rb
@@ -564,6 +564,42 @@ describe "bundle exec" do
       it_behaves_like "it runs"
     end
 
+    context "regarding $0 and __FILE__" do
+      let(:executable) { super() + <<-'RUBY' }
+
+        puts "$0: #{$0.inspect}"
+        puts "__FILE__: #{__FILE__.inspect}"
+      RUBY
+
+      let(:expected) { super() + <<-EOS.chomp }
+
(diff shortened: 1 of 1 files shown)


spec/commands/exec_spec.rb
```diff
@@ -457,5 +457,33 @@ def bin_path(a,b,c)
...
+
+      it_behaves_like "it runs"
+
+      context "when the path is relative" do
+        let(:path) { super().relative_path_from(Pathname.pwd) }
+
+        it_behaves_like "it runs"
+      end
+
+      xcontext "when the path is relative with a leading ./" do
```
segiddins: note this is currently disabled since I haven't had time to dig into MRI to see if this is even possible

indirect:
@segiddins rebase? I think we need this...

segiddins:
Sure, but I don't actually know what the fix will be, these are just failing specs awaiting an implementation

indirect:
do we have a way (maybe a flag to exec?) to opt out of exec using `load` for people who need $0 or __FILE__?

segiddins:
At the moment we do not, but it should be easy to add a flag for it, `Bundler.settings[:disable_exec_load]` or something

segiddins:
Rebase'd

indirect:
@homu r+

PR #4531, opened 2016-05-06
[CompactIndex] Fall back when the user's home directory is not writable

segiddins:


RochesterinNYC:
This looks good, though should we also think about how we might be able to allow remote CI systems or systems without `$HOME` set to use the compact index? Maybe allow for the specification of an alternate path, which would be `$HOME` by default. Fallback would only occur if this alternate path is not set and if `$HOME` is not set or writable? I don't hve the curremt context on the compact index integration code, though, so I don't know how big of a change this would be.

segiddins:
I'd rather just have them set HOME to something to keep things simple

RochesterinNYC:
Hm, I'm inclined to agree from a maintaining and bug surface area perspective, but I could also see why devs would want to use the compact index in CI systems but not have $HOME set. Theoretically, would running `HOME=/path/to/something bundle install` work without broken side effects?

indirect:
I feel weird about an implementation detail of the new index (writing files) forcing people to not use it. I'd strongly prefer that we either keep everything in memory and still use the new index or use Tmpdir if HOME is not set.

segiddins:
@indirect keeping everything in memory (and having no caching) will make performance far worse than the dependency API and probably worse than `--full-index`. I'm scared about using TMPDIR because of the possibility of untrustworthy data getting into there (& what happens if it gets wiped out mid-install?)

indirect:
Bleh. That's a totally valid point. Ultimately, I think we need a solution that allows us to never fall back, because we want to stop using the old index options. If this is the only way to make it work for now, that's fine—let's plan to come up with a solution that doesn't require falling back once we're under less pressure to fix a bug in the wild.

@homu r+

PR #4558, opened 2016-05-10
[Index] Ensure nil and "ruby" platforms are treated identically

### LINKED ISSUE #4557: Installing json 1.8.3 on Ruby 2.3.1 on Travis results in error.
For some reason Bundler `~> 1.12.0` can't install `json-1.8.3` on TravisCI. See https://travis-ci.org/rspec/rspec-mocks/jobs/128825648 as an example, the only way I've been able to get RSpecs build to pass is to pin bundler to the latest 1.11 release.

segiddins:
Fixes #4557.

@indirect @RochesterinNYC ideas on how to test this?

### DIFF
--- a/spec/bundler/index_spec.rb
+++ b/spec/bundler/index_spec.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+require "spec_helper"
+
+describe Bundler::Index do
+  let(:specs) { [] }
+  subject { described_class.build {|i| i.use(specs) } }
+
+  context "specs with a nil platform" do
+    let(:spec) do
+      Gem::Specification.new do |s|
+        s.name = "json"
+        s.version = "1.8.3"
+        allow(s).to receive(:platform).and_return(nil)
+      end
+    end
+    let(:specs) { [spec] }
+
+    describe "#search_by_spec" do
+      it "finds the spec when a nil platform is specified" do
+        expect(subject.search(spec)).to eq([spec])
+      end
+
+      it "finds the spec when a ruby platform is specified" do
+        query = spec.dup.tap {|s| s.platform = "ruby" }
+        expect(subject.search(query)).to eq([spec])
+      end
+    end
+  end
+end
--- a/lib/bundler/index.rb
+++ b/lib/bundler/index.rb
@@ -89,7 +89,7 @@ module Bundler
     alias_method :[], :search
 
     def <<(spec)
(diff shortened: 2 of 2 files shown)


indirect:
Generate gemspecs until you have situation like json 1.8.3 I guess? I dunno :/

segiddins:
Spec added

RochesterinNYC:
@segiddins That's along the lines of what I would have suggested. Nice :+1:

segiddins:
Worth a patch release for this?

mokhan:
Thanks for working on this @segiddins.

segiddins:
@indirect @RochesterinNYC r?

spec/bundler/index_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+    let(:specs) { [spec] }
+
+    describe "#search_by_spec" do
+      it "finds the spec when a nil platform is specified" do
+        expect(subject.search(spec)).to eq([spec])
+      end
+
+      it "finds the spec when a ruby platform is specified" do
+        query = spec.dup.tap {|s| s.platform = "ruby" }
```
RochesterinNYC: does this really belong in the `"specs with a nil platform"` context?

segiddins: ah but the spec itself has a nil platform, only the query doesn't

RochesterinNYC: Ah okay. Thanks for clearing that up. My point of confusion was because `describe "#method"` blocks are usually the top-level contexts.

segiddins:
Might it make sense to just use `spec.full_name` as the hash key there?

RochesterinNYC:
Storing the specs as a hash with just one level of full name as key and spec as the value instead of a nested hash with name as the first level and then the version and/or platform as second level?

segiddins:
@RochesterinNYC still storing it two levels, but just changing it to `[name][full_name]`

RochesterinNYC:
Either actually makes sense to me, or rather, they seem semantically equivalent. What do you see as the benefit(s) of switching to full name as the second-level hash key? More explicitness?

segiddins:
Not reinventing the wheel, mostly

RochesterinNYC:
Oh, so that the `|| ruby` code and string construction for the second hash key in the search methods wouldn't be needed. Yes, that makes sense. :+1:

segiddins:
@homu r+

PR #4559, opened 2016-05-10
[Travis] Test against Rubygems `2.6.4` and Ruby `2.3.1`

RochesterinNYC:
- Side note: this feels like something we might be able to automate if we had a CI pipeline that could do periodic resource checks (i.e. watch for new version of Rubygems or Ruby).


segiddins:
@homu r+

PR #4566, opened 2016-05-11
do not cleanup load path for jruby

### LINKED ISSUE #4565: Bundler::SharedHelpers.clean_load_path does cleanup the default load_path on jruby
when packing a jruby into jar the internal layout looks as such: let ROOT be the root of the jar.

JRUBY_HOME=ROOT
GEM_PATH=ROOT
GEM_HOME=ROOT

and the default load path of jruby is
`$JRUBY_HOME/lib/ruby/2.3/site_ruby, $JRUBY_HOME/lib/ruby/stdlib`

in such a setup `Bundler::SharedHelpers.clean_load_path` will wipe out the default load_path and you will see errors like:

```
LoadError: no such file to load -- bundler/dep_proxy
  block in expand_dependencies at uri:classloader:/gems/bundler-1.12.3/lib/bundler/definition.rb:630
                          each at org/jruby/RubyArray.java:1593
  block in expand_dependencies at uri:classloader:/gems/bundler-1.12.3/lib/bundler/definition.rb:629
                          each at org/jruby/RubyArray.java:1593
           expand_dependencies at uri:classloader:/gems/bundler-1.12.3/lib/bundler/definition.rb:626
         converge_locked_specs at uri:classloader:/gems/bundler-1.12.3/lib/bundler/definition.rb:594
                       resolve at uri:classloader:/gems/bundler-1.12.3/lib/bundler/definition.rb:195
                         specs at uri:classloader:/gems/bundler-1.12.3/lib/bundler/definition.rb:140
                     specs_for at uri:classloader:/gems/bundler-1.12.3/lib/bundler/definition.rb:185
               requested_specs at uri:classloader:/gems/bundler-1.12.3/lib/bundler/definition.rb:174
               requested_specs at uri:classloader:/gems/bundler-1.12.3/lib/bundler/environment.rb:19
                         setup at uri:classloader:/gems/bundler-1.12.3/lib/bundler/runtime.rb:14
                         setup at uri:classloader:/gems/bundler-1.12.3/lib/bundler.rb:95
                         <top> at uri:classloader:/json-bundler-warbler/bin/run.rb:3
                          load at org/jruby/RubyKernel.java:962
                         <top> at uri:classloader:/META-INF/main.rb:1
                       require at org/jruby/RubyKernel.java:944
                        (root) at uri:classloader:/META-INF/main.rb:1
                         <top> at uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib/rubygems/core_ext/kernel_require.rb:1
ERROR: org.jruby.embed.EvalFailedException: (LoadError) no such file to load -- bundler/dep_proxy
```

see https://gist.github.com/mkristian/5738370aba7d6cbe38d516cfdc07d992 for a complete example.

I did monkey patch with jruby-mains like this: https://github.com/jruby/jruby-mains/blob/master/src/main/resources/META-INF/monkey_patches.rb#L9 for some time now. but warbler does not have this monkey patch, so warblers packed jar files can fail when using bundler.

PR is coming

mkristian:
jruby does not add default gems explicitly to the load path and the cleaning
of the load path wipes out the default laod_path of jruby in cases where
JRUBY_HOME == GEM_HOME

fixes #4565

### DIFF
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -193,13 +193,19 @@ module Bundler
     def clean_load_path
       # handle 1.9 where system gems are always on the load path
       if defined?(::Gem)
-        me = File.expand_path("../../", __FILE__)
+        # JRuby sometime produces uri:classloader:/gems/bundler-1.12.3/lib
+        # paths and sometimes uri:classloader://gems/bundler-1.12.3/lib
+        # due to some file-path normalization not able to keep uri prefixes
+        # intact. i.e. when 'me' is 'uri:classloader://gems/bundler-1.12.3/lib'
+        # and 'uri:classloader:/gems/bundler-1.12.3/lib' is on the load path
+        # then bundle itself gets removed from the load path
+        me = File.expand_path("../../", __FILE__).sub('://', ':/')
         me = /^#{Regexp.escape(me)}/
 
         loaded_gem_paths = Bundler.rubygems.loaded_gem_paths
 
         $LOAD_PATH.reject! do |p|
-          next if File.expand_path(p) =~ me
+          next if File.expand_path(p).sub('://', ':/') =~ me
           loaded_gem_paths.delete(p)
         end
         $LOAD_PATH.uniq!


mkristian:
please advise on how and where I can add a spec for this !

lib/bundler/shared_helpers.rb
```diff
@@ -192,7 +192,8 @@ def set_rubylib
 
     def clean_load_path
       # handle 1.9 where system gems are always on the load path
-      if defined?(::Gem)
+      # jruby does not put any system gems explicitly on the load path
+      if defined?(::Gem) and not defined?(JRUBY_VERSION)
```
segiddins: we don't use keyword and or not

segiddins:
in the platform specs, you can simulat_jruby iirc

RochesterinNYC:
Ex. of simulation in a test spec: https://github.com/bundler/bundler/blob/master/spec/install/gemfile/platform_spec.rb#L44

mkristian:
thanks for the hints on testing this. but I realized that my first analyses of the problem was wrong. changed the fix but that new one I have no idea how to test it, since I would bundler running from within a jar-file.

please let me know if such a "hacky" is still OK !

segiddins:
Can you please describe how / why this fixes the issue? (And preferably in a comment in the code).  I'm also very reticent to merge bug fixes without test coverage, since it makes it so easy to re-break things in the future (especially when the fix is so hack)

mkristian:
@segiddins I understand your point and I thought I explained it already in the comment. obviously I did not succeed. just "normalize" the uri-like paths so compare with the next hits where needed.

segiddins:
https://travis-ci.org/bundler/bundler/jobs/129491788#L490-L507

mkristian:
I thought about this and it is not the right place to fix this problem of jruby. feels like starting to fill the ecosystem with workaround which should/could be fix inside jruby.  https://github.com/jruby/jruby/issues/3881

mkristian:
thanks any one to look at this but we are not going this route.

segiddins:
👍 thanks for digging in and making the PR, @mkristian !

kares:
@mkristian couldn't we get **uri:classloader://** to be consistent on JRuby? there's special handling on a few places for it so maybe we should just make sure its always "normalized" ... just throwing a dice here.

mkristian:
@kares totally agree with you, that was is the idea behind  jruby/jruby#3881 - not sure if I was so clear about this in this issue

kares:
@mkristian oh right - sorry. did not check the issue as I was sure its another one.

PR #4570, opened 2016-05-12
Remove warning by forcing class removal

JonRowe:
We were getting some warnings on appveyor (windows build) from bundler, see (https://ci.appveyor.com/project/rspec/rspec-mocks/build/581/job/cpejy4xymta3oj7e). I think this is because bundler wasn't removing the method before overwriting it

### DIFF
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -362,7 +362,7 @@ module Bundler
     def replace_bin_path(specs)
       gem_class = (class << Gem; self; end)
 
-      redefine_method(gem_class, :find_spec_for_exe) do |name, *args|
+      redefine_method(gem_class, :find_spec_for_exe, true) do |name, *args|
         exec_name = args.first
 
         spec = if exec_name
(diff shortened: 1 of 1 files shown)


segiddins:
Can you take a look at the CI failures?

lib/bundler/rubygems_integration.rb
```diff
@@ -461,8 +461,8 @@ def backport_spec_file
       end
     end
 
-    def redefine_method(klass, method, &block)
-      if klass.instance_methods(false).include?(method)
+    def redefine_method(klass, method, force_removal = false, &block)
+      if force_removal || klass.instance_methods.include?(method)
```
RochesterinNYC: how exactly does this logic work or how is it intended to work? Currently, if `force_removal` is `true`, then it'll try to invoke `remove_method` regardless of whether that method is defined or not?

JonRowe: It was intended to allow overriding the method check, as I believe this method wasn't being detected properly and thus not removed on windows, (generating the warning generated seen on our windows ci)

PR #4580, opened 2016-05-16
Improve resolver performance

segiddins:
Some not insignificant improvements found while developing https://github.com/CocoaPods/Molinillo/pull/40, which I absolutely cannot wait to 🚢 

\c @indirect

### DIFF
--- a/lib/bundler/remote_specification.rb
+++ b/lib/bundler/remote_specification.rb
@@ -66,6 +66,10 @@ module Bundler
       [@name, @version, @platform == Gem::Platform::RUBY ? -1 : 1]
     end
 
+    def to_s
+      "#<#{self.class} name=#{name} version=#{version} platform=#{platform}>"
+    end
+
   private
 
     def _remote_specification
--- a/lib/bundler/remote_specification.rb
+++ b/lib/bundler/remote_specification.rb
@@ -75,11 +75,7 @@ module Bundler
     end
 
     def method_missing(method, *args, &blk)
-      if Gem::Specification.new.respond_to?(method)
-        _remote_specification.send(method, *args, &blk)
-      else
-        super
-      end
+      _remote_specification.send(method, *args, &blk)
     end
   end
 end
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -225,7 +225,7 @@ module Bundler
           source.dependency_names = dependency_names.dup
           idx.add_source source.specs
           dependency_names -= pinned_spec_names(source.specs)
-          dependency_names.push(*source.unmet_deps).uniq!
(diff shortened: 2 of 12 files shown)


lib/bundler/remote_specification.rb
```diff
@@ -75,11 +75,7 @@ def _remote_specification
     end
 
     def method_missing(method, *args, &blk)
-      if Gem::Specification.new.respond_to?(method)
-        _remote_specification.send(method, *args, &blk)
-      else
-        super
-      end
+      _remote_specification.send(method, *args, &blk)
```
indirect: I'm pretty sure that simply calling `_remote_specification` will instantiate it, and this was trying to avoid instantiating it if it didn't respond to the method that was being called? but maybe we removed whatever made that relevant.

segiddins: https://github.com/bundler/bundler/commit/c563003292bd36dc647b41016fad93d5b7324a07 introduced it. We can also just use `Gem::Specification.instance_methods.any? {|m| m.to_sym == method.to_sym }` or something if it's actually important

indirect:
woohoo this looks great!

lib/bundler/remote_specification.rb
indirect: if it's faster without it and all the tests pass, I'm fine with dropping it.

segiddins:
Biggest issue now is https://github.com/bundler/bundler/pull/4580/commits/61f65828186e67c580518b1884d89869539d17f8, which doesn't play nice with https://github.com/rubygems/rubygems/issues/1617

segiddins:
@indirect what should I do about this?

indirect:
@segiddins implement a new `Gem::Specification#hash` so you can ship this? Or ship this with everything but that, work on a new `#hash` later, and land just that part after. Up to you.

indirect:
(For backwards-compat, you're going to have to support older versions of RubyGems that have unequal hashes regardless... implement that, and then let's speed up hash and use the new faster version when it's available.

segiddins:
@indirect so would that mean making grouping specs by version something that need to go in `rubygems_integration`? Joy.

segiddins:
@indirect this should be ready for review now

indirect:
@homu r+

segiddins:
Fixed the rubocop violation

segiddins:
@indirect should now be fixed!

indirect:
@homu r+

segiddins:
@indirect can you take a look at that one failure?

indirect:
@segiddins it looks like the changes to Molinillo are producing a more full conflict message, and the test just needs to be updated to expect both conflicts (gemfile and rails) in the output.

segiddins:
@homu r=indirect

PR #4581, opened 2016-05-16
Handle yanked gems, fix #4344

### LINKED ISSUE #4344: Handle yanked gems in a clear way
Today, when a gem is yanked, we throw the exact same error as when you manually enter a version number into your Gemfile that doesn't actually exist. If a gem is yanked, we have evidence that the version used to exist (in the lockfile), and so we should be able to print an error message that is much more informative, something along the lines of: 

> Your bundle is locked to foo-1.2.3, but that version could not be found in any of the sources listed in your Gemfile. If you haven't changed sources, that means the author of foo-1.2.3 has removed it. You'll need to update your bundle to a different version of foo that hasn't been removed in order to install.

allenzhao:
@RochesterinNYC 
I will add test later. Please review if I'm handling this correctly.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -137,8 +137,17 @@ module Bundler
     # @return [Bundler::SpecSet]
     def specs
       @specs ||= begin
-        specs = resolve.materialize(Bundler.settings[:cache_all_platforms] ? dependencies : requested_dependencies)
-
+        begin
+          specs = resolve.materialize(Bundler.settings[:cache_all_platforms] ? dependencies : requested_dependencies)
+        rescue GemNotFound => e # Handle yanked gem
+          gem_name, gem_version = extract_gem_info(e)
+          locked_gem = @locked_specs[gem_name].last
+          raise if locked_gem.nil? || locked_gem.version.to_s != gem_version
+          raise GemNotFound, "Your bundle is locked to #{locked_gem}, but that version could not " \
+                             "be found in any of the sources listed in your Gemfile. If you haven't changed sources, " \
+                             "that means the author of #{locked_gem} has removed it. You'll need to update your bundle " \
+                             "to a different version of foo that hasn't been removed in order to install."
+        end
         unless specs["bundler"].any?
           local = Bundler.settings[:frozen] ? rubygems_index : index
           bundler = local.search(Gem::Dependency.new("bundler", VERSION)).last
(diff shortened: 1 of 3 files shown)


segiddins:
@allenzhao the best way to tell if it's correct is to write tests ;)

lib/bundler/definition.rb
```diff
@@ -137,8 +137,17 @@ def resolve_remotely!
     # @return [Bundler::SpecSet]
     def specs
       @specs ||= begin
-        specs = resolve.materialize(Bundler.settings[:cache_all_platforms] ? dependencies : requested_dependencies)
-
+        begin
+          specs = resolve.materialize(Bundler.settings[:cache_all_platforms] ? dependencies : requested_dependencies)
+        rescue GemNotFound => e # Handle yanked gem
+          gem_info = extract_gem_info(e)
+          locked_gem = @locked_specs[gem_info[0]].last
```
RochesterinNYC: I'd like to see some more explicitness about what information about the gem you're extracting (maybe through variable assignment?). What is being pulled out of `gem_info`?

RochesterinNYC:
It generally looks like you're on the right track here. I'd have to agree with @segiddins on the point about tests. One way to approach bug fixes here is to write the failing test first and then write the code to make it pass. This is often a lot easier than doing local reproduction/manual testing or etc. to see if your code fixes the issue.

lib/bundler/definition.rb
allenzhao: You mean assign like this?
gem_name = gem_info[0]
gem_version = gem_info[1]
Becasue what I did in extract_gem_info is a regex match

allenzhao:
I think I'll have to close this and open a new one.

RochesterinNYC:
Why?

allenzhao:
Somehow the force pushed change is not shown here, I don’t know why

On 周二, 5月 24, 2016 at 12:34 上午, James Wen <notifications@github.com> wrote:
Why?

RochesterinNYC:
In the future, please let me know and let's try to work it out first, rather than creating duplicate PRs. For example, Github is undergoing service issues, so that might have been the root of you not seeing the force-pushed update. https://status.github.com/ --> 11:31 AM EDT (We are still working to resolve minor service issues. You may see delays in updates across the site.)

allenzhao:
Okay, I won't create another PR and will reopen it if GitHub works fine.

PR #4585, opened 2016-05-18
Use the full expanded path of passed `Gemfile` during `eval_gemfile`

### LINKED ISSUE #4584: Bundler fails to parse a gemspec line when it appears in a gemfile that is evaled inside of another gemfile
Bundler version: 1.12.4
OS: observed on OS X 10.10.5 AND on Ubuntu 14.04.1
Description:
Bundler fails to parse a gemspec line when it appears in a gemfile that is evaled inside of another gemfile
Reproducing: The minimum setup for reproducing this bug can be found in the simple example at https://github.com/stupakov/bundler-bug

Failure output:

```
[!] There was an error parsing `Gemfile`:
[!] There was an error parsing `Gemfile-other`: No such file or directory - gems/mygem/mygem.gemspec. Bundler cannot continue.

 #  from Gemfile-other:1
 >  gemspec path: 'gems/mygem'
. Bundler cannot continue.

 #  from /Users/myusername/workspace/bundler-bug/Gemfile:1
 >  eval_gemfile 'Gemfile-other'
```

RochesterinNYC:
- fixes #4584

### DIFF
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -34,7 +34,9 @@ module Bundler
     end
 
     def eval_gemfile(gemfile, contents = nil)
-      @gemfile = Pathname.new(gemfile).expand_path
+      expanded_gemfile_path = Pathname.new(gemfile).expand_path
+      @gemfile = expanded_gemfile_path
+      original_gemfile = @gemfile
       contents ||= Bundler.read_file(gemfile.to_s)
       instance_eval(contents, gemfile.to_s, 1)
     rescue Exception => e
@@ -43,6 +45,8 @@ module Bundler
         "`#{File.basename gemfile.to_s}`: #{e.message}"
 
       raise DSLError.new(message, gemfile, e.backtrace, contents)
+    ensure
+      @gemfile = original_gemfile
     end
 
     def gemspec(opts = nil)
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -34,7 +34,7 @@ module Bundler
     end
 
     def eval_gemfile(gemfile, contents = nil)
-      @gemfile = Pathname.new(gemfile)
+      @gemfile = Pathname.new(gemfile).expand_path
       contents ||= Bundler.read_file(gemfile.to_s)
       instance_eval(contents, gemfile.to_s, 1)
     rescue Exception => e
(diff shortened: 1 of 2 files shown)


segiddins:
So if we're accidentally trying to support eval_gemfile we probably should restore `@gemfile` in an `ensure`?

RochesterinNYC:
Ah, as in there are things in the class that count on `@gemfile` being a non-absolute path?

segiddins:
No I mean after the nested gemfile is eval'd we're back in the original gemfile.

RochesterinNYC:
Ah, gotcha. Added that to the PR. Good call.

lib/bundler/dsl.rb
```diff
@@ -34,7 +34,9 @@ def initialize
     end
 
     def eval_gemfile(gemfile, contents = nil)
-      @gemfile = Pathname.new(gemfile)
+      expanded_gemfile_path = Pathname.new(gemfile).expand_path
+      @original_gemfile ||= expanded_gemfile_path
```
segiddins: This should just be `original_gemfile = @gemfile`, no need for an ivar

RochesterinNYC: Initially, I thought that we wanted to restore `@gemfile` to the original, top-level `Gemfile` path after each `eval_gemfile` invocation. However, you're right, I think if we want to support multiple, recursive nesting of `eval_gemfile` invocations, we have to restore `@gemfile` to the Gemfile path specified for the recursive level above each `eval_gemfile` call. This way, if people are doing things like chaining > 2 Gemfiles and `eval_gemfile` calls (I hope not though...), `@gemfile` is set properly on each recursive level.

lib/bundler/dsl.rb
```diff
@@ -34,7 +34,9 @@ def initialize
     end
 
     def eval_gemfile(gemfile, contents = nil)
-      @gemfile = Pathname.new(gemfile)
+      expanded_gemfile_path = Pathname.new(gemfile).expand_path
+      @gemfile = expanded_gemfile_path
+      original_gemfile = @gemfile
```
segiddins: these two lines are backwards, no?

RochesterinNYC: Yes, you're right. I think what this is telling me is that we need a spec specifically to cover this case. Though, I can't really think of a functionally useful test case. As in, a test with multiple Gemfiles that emulates a real-world use case. If you know any off the top of your head, let me know. Otherwise, I'll look around Github/think one up.

RochesterinNYC:
Added an integration spec covering the `@gemfile` restoration code change.

segiddins:
Yay! 👍 
@homu r+

PR #4589, opened 2016-05-19
Manually set a process title when using load

yob:
With bundler 1.11.2, the process name for rake tasks include the task name:

    $ bundle exec rake foo:bar

    $ ps ux | grep rake
     user         1758 62.2  2.4 385816 202032 pts/3   Sl+  16:16   0:04 ruby /usr/bin/rake foo:bar

On bundler 1.12.0, the process name changed:

    $ bundle exec rake foo:bar
 
    $ ps ux | grep rake
     user         1758 62.2  2.4 385816 202032 pts/3   Sl+  16:16   0:04 /usr/bin/rake

The change in behaviour is caused by bundler 1.12 using `load` (instead of `exec`) where
possible, and manually using `$0=` to set the command name. Unfortunately, that also alters
the process title visible with ps, and using `Process.setproctitle` can help reverse that.

The tests are green on my ruby 2.3.1/linux system, but I haven't run them on older rubies or non-linux environments. I'm assuming they'll need some modifications to be maximally compatible, but let's see what happens.

See also: #4488

### DIFF
--- a/lib/bundler/cli/exec.rb
+++ b/lib/bundler/cli/exec.rb
@@ -57,6 +57,7 @@ module Bundler
       args.pop if args.last.is_a?(Hash)
       ARGV.replace(args)
       $0 = file
+      Process.setproctitle(process_title(file, *args))
       ui = Bundler.ui
       Bundler.ui = nil
       require "bundler/setup"
@@ -68,6 +69,14 @@ module Bundler
       Bundler.ui.error "bundler: failed to load command: #{cmd} (#{file})"
       backtrace = e.backtrace.take_while {|bt| !bt.start_with?(__FILE__) }
       abort "#{e.class}: #{e.message}\n  #{backtrace.join("\n  ")}"
+    end
+
+    def process_title(file, *args)
+      if args.empty?
+        file
+      else
+        "#{file} #{args.join(" ")}".strip
+      end
     end
 
     def ruby_shebang?(file)
--- a/spec/commands/exec_spec.rb
+++ b/spec/commands/exec_spec.rb
@@ -462,6 +462,7 @@ describe "bundle exec" do
       puts "EXEC: \#{caller.grep(/load/).empty? ? 'exec' : 'load'}"
       puts "ARGS: \#{$0} \#{ARGV.join(' ')}"
       puts "RACK: \#{RACK}"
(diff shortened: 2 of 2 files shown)


lib/bundler/cli/exec.rb
```diff
@@ -70,6 +71,14 @@ def kernel_load(file, *args)
       abort "#{e.class}: #{e.message}\n  #{backtrace.join("\n  ")}"
     end
 
+    def process_title(file, *args)
```
segiddins: no reason to splat `*args` here

lib/bundler/cli/exec.rb
```diff
@@ -70,6 +71,14 @@ def kernel_load(file, *args)
...
+      if args.empty?
+        file
+      else
+        "#{file} #{args.join(" ")}".strip
```
yob: This new process title doesn't exactly match the old one, but it's enough for our purposes. I'm not super familiar with the conventions and behaviour of ruby processes, so I'm happy to modify/tweak it further if need be

lib/bundler/cli/exec.rb
```diff
@@ -57,6 +57,7 @@ def kernel_load(file, *args)
       args.pop if args.last.is_a?(Hash)
       ARGV.replace(args)
       $0 = file
+      Process.setproctitle(process_title(file, *args))
```
RochesterinNYC: @yob I believe this is causing the CI failures as the `Process#setproctitle` method was first created/available in [Ruby `2.1.0`](http://ruby-doc.org/core-2.3.0/Process.html#method-c-setproctitle). As a side note, I have no idea why this is passing: https://travis-ci.org/bundler/bundler/jobs/131320827

yob:
Hi all,

I haven't had much time to re-visit this, and we've since found a workaround for our use case.

I'm happy to try and polish this into something that's mergable, but I'll need a bit more guidance on how to proceed. It seems `Process.setproctitle` won't meet backward compatibility requirements. Should I consider making the fix only apply to ruby >= 2.1?

indirect:
@yob we typically use checks like `if Process.respond_to?(:setproctitle)` when we need something to continue to work (without the new feature) on older Rubies.

segiddins:
I've manually addressed the comments on this in https://github.com/bundler/bundler/pull/4863

PR #4606, opened 2016-05-24
Handle yanked gems

### LINKED ISSUE #4344: Handle yanked gems in a clear way

allenzhao:
Close #4344

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -146,7 +146,7 @@ module Bundler
           raise GemNotFound, "Your bundle is locked to #{locked_gem}, but that version could not " \
                              "be found in any of the sources listed in your Gemfile. If you haven't changed sources, " \
                              "that means the author of #{locked_gem} has removed it. You'll need to update your bundle " \
-                             "to a different version of foo that hasn't been removed in order to install."
+                             "to a different version of #{locked_gem} that hasn't been removed in order to install."
         end
         unless specs["bundler"].any?
           local = Bundler.settings[:frozen] ? rubygems_index : index
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -137,8 +137,17 @@ module Bundler
     # @return [Bundler::SpecSet]
     def specs
       @specs ||= begin
-        specs = resolve.materialize(Bundler.settings[:cache_all_platforms] ? dependencies : requested_dependencies)
-
+        begin
+          specs = resolve.materialize(Bundler.settings[:cache_all_platforms] ? dependencies : requested_dependencies)
+        rescue GemNotFound => e # Handle yanked gem
+          gem_name, gem_version = extract_gem_info(e)
+          locked_gem = @locked_specs[gem_name].last
+          raise if locked_gem.nil? || locked_gem.version.to_s != gem_version
+          raise GemNotFound, "Your bundle is locked to #{locked_gem}, but that version could not " \
+                             "be found in any of the sources listed in your Gemfile. If you haven't changed sources, " \
+                             "that means the author of #{locked_gem} has removed it. You'll need to update your bundle " \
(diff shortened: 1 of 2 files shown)


spec/install/yanked_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+require "spec_helper"
+
+describe "handle yanked" do
+  before(:each) do
+    build_repo4 do
+      build_gem "foo", "9.0.0"
+    end
+  end
+
+  it "throws and error when there is gem yanked" do
```
segiddins: `an error`

spec/install/yanked_spec.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+require "spec_helper"
+
+describe "handle yanked" do
```
segiddins: `context "when installing a bundle that includes yanked gems"`

lib/bundler/definition.rb
```diff
@@ -137,8 +137,17 @@ def resolve_remotely!
...
+        begin
+          specs = resolve.materialize(Bundler.settings[:cache_all_platforms] ? dependencies : requested_dependencies)
+        rescue GemNotFound => e # Handle yanked gem
+          gem_name, gem_version = extract_gem_info(e)
+          locked_gem = @locked_specs[gem_name].last
+          raise if locked_gem.nil? || locked_gem.version.to_s != gem_version
+          raise GemNotFound, "Your bundle is locked to #{locked_gem}, but that version could not " \
+                             "be found in any of the sources listed in your Gemfile. If you haven't changed sources, " \
+                             "that means the author of #{locked_gem} has removed it. You'll need to update your bundle " \
+                             "to a different version of foo that hasn't been removed in order to install."
```
RochesterinNYC: it says "different version of foo" in the error message. I think this should be the gem name, right?

spec/install/yanked_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+require "spec_helper"
+
+context "when installing a bundle that includes yanked gems" do
+  before(:each) do
+    build_repo4 do
+      build_gem "foo", "9.0.0"
+    end
+  end
+
+  it "throws an error when there is gem yanked" do
```
RochesterinNYC: "when the original gem version is yanked"

spec/install/yanked_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    install_gemfile <<-G
+        source "file://#{gem_repo4}"
+        gem "foo", "10.0.0"
+    G
+
+    expect(out).to include("Your bundle is locked to foo (10.0.0)")
+  end
+
+  it "throws the original error when manually enter a version number that doesn't exist" do
```
RochesterinNYC: "when only gemfile specifies a gem version that doesn't exist"

RochesterinNYC:
@allenzhao made a few comments. Other than those small changes, this LGTM.

lib/bundler/definition.rb
allenzhao: Oh, I forgot to change that

RochesterinNYC:
@segiddins, can you r+ this or delegate me? Thanks.

segiddins:
@homu delegate=RochesterinNYC

RochesterinNYC:
@homu r+

Issue #4367, opened 2016-03-14
ThreadError: can't create Thread: Resource temporarily unavailable


hendranata:
I face a strange error.
but u can see my screenshot below

when i did "bundle"
it was showing an error
ThreadError: can't create Thread: Resource temporarily unavailable
http://s16.postimg.org/kyoo484gl/errorrrr.png

in this case, i suspect that the problem actually comming from memory, but I have check the free memory is around 2gb and swap 3gb..
but dont know why the error is still coming up.

anybody help me..
thanks

RochesterinNYC:
Please post your `Gemfile` and the results of `bundle env` (or post the full error report in this issue or a gist or something).

hendranata:
here is my Gemfile

```
source 'https://rubygems.org'

gem 'rails', '4.2.5.1'
gem 'bootstrap-sass', '~> 3.3', '>= 3.3.6'
gem 'sass-rails', '~> 5.0'
gem 'uglifier', '>= 1.3.0'
gem 'coffee-rails', '~> 4.1.0'
gem 'jquery-rails'
gem 'turbolinks'
gem 'jbuilder', '~> 2.0'
gem 'sdoc', '~> 0.4.0', group: :doc

group :development, :test do
  gem 'byebug'
end

group :development do
  gem 'web-console', '~> 2.0'
  gem 'sqlite3'
  gem 'spring'
end

group :production do
#  gem 'pg',             '0.17.1'
#  gem 'rails_12factor', '0.0.2'
 # gem 'puma',           '3.1.0'
end

```



and here is result of bundle env


`(app:2.1)plncikande@iix-ssd [~/app]# bundle env
Environment

    Bundler   1.12.0.rc
    Rubygems  2.2.5
    Ruby      2.1.8p440 (2015-12-16 revision 53160) [i686-linux-gnu]
    GEM_HOME  /home/plncikande/rubyvenv/app/2.1
    GEM_PATH  /home/plncikande/rubyvenv/app/2.1:/opt/alt/ruby21/lib/ruby/gems/2.1.0
    Git       not installed

Bundler settings

    orig_path
[...]

PR #4607, opened 2016-05-24
Clean up worker threads once done with them

will-in-wi:
Right now, the thread pools created by CompactIndex are not cleaned up once they are done. I assume that over time, they would be garbage collected, but in the meantime there could be 200+ threads running. Many shared hosts have fork bomb protection set up which kills Bundler.

This patch will clean up the threads as soon as they are done, keeping the total number of active threads at any one time to a minimum.

Fixes https://github.com/bundler/bundler/issues/4367

### DIFF
--- a/lib/bundler/fetcher/compact_index.rb
+++ b/lib/bundler/fetcher/compact_index.rb
@@ -89,7 +89,10 @@ module Bundler
               worker_name = "Compact Index (#{display_uri.host})"
               worker = Bundler::Worker.new(25, worker_name, func)
               inputs.each {|input| worker.enq(input) }
-              inputs.map { worker.deq }
+              res = inputs.map { worker.deq }
+              worker.stop
+
+              res
             end
           end
         end
--- a/lib/bundler/fetcher/compact_index.rb
+++ b/lib/bundler/fetcher/compact_index.rb
@@ -89,10 +89,7 @@ module Bundler
               worker_name = "Compact Index (#{display_uri.host})"
               worker = Bundler::Worker.new(25, worker_name, func)
               inputs.each {|input| worker.enq(input) }
-              res = inputs.map { worker.deq }
-              worker.stop
-
-              res
+              inputs.map { worker.deq }.tap { worker.stop }
             end
           end
         end
--- a/spec/bundler/fetcher/compact_index_spec.rb
+++ b/spec/bundler/fetcher/compact_index_spec.rb
@@ -4,7 +4,7 @@ require "spec_helper"
 describe Bundler::Fetcher::CompactIndex do
   let(:downloader)  { double(:downloader) }
   let(:remote)      { double(:remote, :cache_slug => "lsjdf") }
-  let(:display_uri) { URI("http://sample_uri.com") }
+  let(:display_uri) { URI("http://sampleuri.com") }
   let(:compact_index) { described_class.new(downloader, remote, display_uri) }
 
   # Testing private method. Do not commit.
--- a/spec/bundler/fetcher/compact_index_spec.rb
+++ b/spec/bundler/fetcher/compact_index_spec.rb
@@ -8,7 +8,7 @@ describe Bundler::Fetcher::CompactIndex do
   let(:compact_index) { described_class.new(downloader, remote, display_uri) }
 
   # Testing private method. Do not commit.
-  describe '#specs_for_names' do
+  describe "#specs_for_names" do
     it "has only one thread open at the end of the run" do
       compact_index.specs_for_names(["lskdjf"])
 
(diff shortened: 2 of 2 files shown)


lib/bundler/fetcher/compact_index.rb
```diff
@@ -89,7 +89,10 @@ def compact_index_client
               worker_name = "Compact Index (#{display_uri.host})"
               worker = Bundler::Worker.new(25, worker_name, func)
               inputs.each {|input| worker.enq(input) }
-              inputs.map { worker.deq }
```
segiddins: `inputs.map { worker.deq }.tap { worker.stop }`?

segiddins:
Awesome, could you add a test for this?

will-in-wi:
I'm not sure how to write such a test without testing implementation... This is a private method which is returning another object's instance, to which it is passing a lambda containing the offending code. So, I'd have to run a method which calls this method, and somehow execute the lambda, verifying that the total number of threads in Ruby at that point is less than N. This seems really brittle to me. To do this correctly would involve a deep refactor, and even then would be tricky to verify. Number of threads is basically an implementation detail...

will-in-wi:
I can't seem to find any existing tests for the `CompactIndex`. Do they exist behind a level of abstraction somewhere?

will-in-wi:
I created a test, but it has so many anti-patterns, it isn't even funny.

```diff --git a/spec/bundler/fetcher/compact_index.rb b/spec/bundler/fetcher/compact_index.rb
new file mode 100644
index 0000000..f32c0dc
--- /dev/null
+++ b/spec/bundler/fetcher/compact_index.rb
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+require "spec_helper"
+
+describe Bundler::Fetcher::CompactIndex do
+  let(:downloader)  { double(:downloader) }
+  let(:remote)      { double(:remote, cache_slug: 'lsjdf') }
+  let(:display_uri) { URI("http://sample_uri.com") }
+  let(:compact_index) { described_class.new(downloader, remote, display_uri) }
+
+  # Testing private method. Do not commit.
+  describe '#compact_index_client' do
+    it 'has only one open thread after execution' do
+      client = compact_index.send(:compact_index_client)
+      func = lambda { |object| return }
+      client.in_parallel.call(['test_input'], &func)
+
+      thread_count = Thread.list.select {|thread| thread.status == "run"}.count
+      expect(thread_count).to eq 1
+    end
+  end
+end
```

However, it fails without the patch and passes with the patch.

Do you have suggestions as to how I ought to approach this?

will-in-wi:
I'm taking a crack at refactoring this guy…

segiddins:
I'd be happy with an expectation that any instance of worker receives stop

will-in-wi:
Does `/vendor` in the repo mean anything in particular? It looks like this could be dramatically simplified (and made testable) by moving the lambda logic from `CompactIndex#compact_index_client` to `CompactIndexClient#dependencies`. Any reason not to do this refactor?

will-in-wi:
I just pushed a test for the threading issue.

will-in-wi:
I finally got all of the builds passing. Kudos to the Bundler devs who deal with the massive array of test targets.

segiddins:
Looks good to me! @indirect @RochesterinNYC worth a patch release?

spec/bundler/fetcher/compact_index_spec.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+require "spec_helper"
+
+describe Bundler::Fetcher::CompactIndex do
+  let(:downloader)  { double(:downloader) }
+  let(:remote)      { double(:remote, :cache_slug => "lsjdf") }
+  let(:display_uri) { URI("http://sampleuri.com") }
+  let(:compact_index) { described_class.new(downloader, remote, display_uri) }
+
+  # Testing private method. Do not commit.
```
RochesterinNYC: This comment is a bit confusing?

RochesterinNYC:
Yup, this definitely looks worth a patch release after merge to master.

segiddins:
@indirect r?

spec/bundler/fetcher/compact_index_spec.rb
will-in-wi: Good point. I missed removing that after I fixed the test to not be nearly so ugly.

will-in-wi:
Can someone kick the failing test? I think it is a fluke, given that it passed before I removed a comment.

RochesterinNYC:
Restarted the failing job.

will-in-wi:
Thanks! They all pass again.

segiddins:
@homu r+

indirect:
Hooray! Thanks everyone!

stalker37:
on rvm and ruby 2.2.4 and bundler-1.13.1 problem is alive. debian jessie

segiddins:
@stalker37 please open a new issue

PR #4608, opened 2016-05-24
Plugin system

asutoshpalai:
This is the working PR for plugin system for bundler.

A rough workflow shall be
- [x] Add functionality to install a plugin as gem from RubyGem source and also from git source
- [x] Add support for command line plugin
- [ ] Add support for source plugin
- [ ] Add support for life cycle hooks
- [ ]  Finalize all the functionalities

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -82,7 +82,7 @@ module Bundler
     end
 
     def self.handle_no_command_error(command, has_namespace = $thor_runner)
-      if ENV["BUNDLE_PLUGIN"] && Bundler::Plugin.command?(command)
+      if Bundler.settings[:plugin] && Bundler::Plugin.command?(command)
         return Bundler::Plugin.exec_command(command, ARGV[1..-1])
       end
 
@@ -437,7 +437,7 @@ module Bundler
       Env.new.write($stdout)
     end
 
-    if ENV["BUNDLE_PLUGIN"]
+    if Bundler.settings[:plugin]
       require "bundler/cli/plugin"
       desc "plugin SUBCOMMAND ...ARGS", "manage the bundler plugins"
       subcommand "plugin", Plugin
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -16,6 +16,7 @@ module Bundler
 
     def initialize(*args)
       super
+      Bundler.reset!
 
       custom_gemfile = options[:gemfile] || Bundler.settings[:gemfile]
       ENV["BUNDLE_GEMFILE"] = File.expand_path(custom_gemfile) if custom_gemfile && !custom_gemfile.empty?
@@ -441,7 +442,7 @@ module Bundler
       Env.new.write($stdout)
     end
 
-    if ENV["BUNDLE_PLUGIN"]
+    if Bundler.settings[:plugin]
       require "bundler/cli/plugin"
       desc "plugin SUBCOMMAND ...ARGS", "manage the bundler plugins"
       subcommand "plugin", Plugin
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -82,6 +82,10 @@ module Bundler
     end
 
     def self.handle_no_command_error(command, has_namespace = $thor_runner)
+      if Bundler::Plugin.command? command
+        return Bundler::Plugin.exec_command(command, ARGV[1..-1])
+      end
+
       return super unless command_path = Bundler.which("bundler-#{command}")
 
       Kernel.exec(command_path, *ARGV[1..-1])
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -441,7 +441,7 @@ module Bundler
       Env.new.write($stdout)
     end
 
-    if Bundler.settings[:plugin]
+    if ENV["BUNDLE_PLUGIN"]
       require "bundler/cli/plugin"
       desc "plugin SUBCOMMAND ...ARGS", "manage the bundler plugins"
       subcommand "plugin", Plugin
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -433,6 +433,10 @@ module Bundler
       Env.new.write($stdout)
     end
 
+    require "bundler/cli/plugin"
+    desc "plugin SUBCOMMAND ...ARGS", "manage the bundler plugins"
+    subcommand "plugin", Plugin
+
     # Reformat the arguments passed to bundle that include a --help flag
     # into the corresponding `bundle help #{command}` call
     def self.reformatted_help_args(args)
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -83,7 +83,7 @@ module Bundler
     end
 
     def self.handle_no_command_error(command, has_namespace = $thor_runner)
-      if Bundler.settings[:plugin] && Bundler::Plugin.command?(command)
+      if Bundler.settings[:plugins] && Bundler::Plugin.command?(command)
         return Bundler::Plugin.exec_command(command, ARGV[1..-1])
       end
 
@@ -442,7 +442,7 @@ module Bundler
       Env.new.write($stdout)
     end
 
-    if Bundler.settings[:plugin]
+    if Bundler.settings[:plugins]
       require "bundler/cli/plugin"
       desc "plugin SUBCOMMAND ...ARGS", "manage the bundler plugins"
       subcommand "plugin", Plugin
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -82,7 +82,7 @@ module Bundler
     end
 
     def self.handle_no_command_error(command, has_namespace = $thor_runner)
-      if Bundler::Plugin.command? command
+      if ENV["BUNDLE_PLUGIN"] && Bundler::Plugin.command?(command)
         return Bundler::Plugin.exec_command(command, ARGV[1..-1])
       end
 
@@ -437,9 +437,11 @@ module Bundler
       Env.new.write($stdout)
     end
 
-    require "bundler/cli/plugin"
-    desc "plugin SUBCOMMAND ...ARGS", "manage the bundler plugins"
-    subcommand "plugin", Plugin
+    if ENV["BUNDLE_PLUGIN"]
+      require "bundler/cli/plugin"
+      desc "plugin SUBCOMMAND ...ARGS", "manage the bundler plugins"
+      subcommand "plugin", Plugin
+    end
 
     # Reformat the arguments passed to bundle that include a --help flag
     # into the corresponding `bundle help #{command}` call
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -56,7 +56,7 @@ module Bundler
       # rubygems plugins sometimes hook into the gem install process
       Gem.load_env_plugins if Gem.respond_to?(:load_env_plugins)
 
-      Plugin.eval_gemfile(Bundler.default_gemfile) if ENV["BUNDLE_PLUGIN"]
+      Plugin.gemfile_install(Bundler.default_gemfile) if Bundler.settings[:plugin]
 
       definition = Bundler.definition
       definition.validate_ruby!
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -185,19 +185,27 @@ module Bundler
     end
 
     def normalize_settings
-      Bundler.settings[:path]     = nil if options[:system]
-      Bundler.settings[:path]     = "vendor/bundle" if options[:deployment]
-      Bundler.settings[:path]     = options["path"] if options["path"]
-      Bundler.settings[:path] ||= "bundle" if options["standalone"]
-      Bundler.settings[:bin]      = options["binstubs"] if options["binstubs"]
-      Bundler.settings[:bin]      = nil if options["binstubs"] && options["binstubs"].empty?
-      Bundler.settings[:shebang]  = options["shebang"] if options["shebang"]
-      Bundler.settings[:jobs]     = options["jobs"] if options["jobs"]
-      Bundler.settings[:no_prune] = true if options["no-prune"]
-      Bundler.settings[:no_install] = true if options["no-install"]
-      Bundler.settings[:clean]    = options["clean"] if options["clean"]
-      Bundler.settings.without    = options[:without]
-      Bundler.settings.with       = options[:with]
+      Bundler.settings[:path]                = nil if options[:system]
+      Bundler.settings[:path]                = "vendor/bundle" if options[:deployment]
+      Bundler.settings[:path]                = options["path"] if options["path"]
+      Bundler.settings[:path]              ||= "bundle" if options["standalone"]
+
+      Bundler.settings[:bin]                 = options["binstubs"] if options["binstubs"]
+      Bundler.settings[:bin]                 = nil if options["binstubs"] && options["binstubs"].empty?
+
+      Bundler.settings[:shebang]             = options["shebang"] if options["shebang"]
+
+      Bundler.settings[:jobs]                = options["jobs"] if options["jobs"]
+
+      Bundler.settings[:no_prune]            = true if options["no-prune"]
+
+      Bundler.settings[:no_install]          = true if options["no-install"]
+
+      Bundler.settings[:clean]               = options["clean"] if options["clean"]
+
+      Bundler.settings.without               = options[:without]
+      Bundler.settings.with                  = options[:with]
+
       Bundler.settings[:disable_shared_gems] = Bundler.settings[:path] ? true : nil
     end
 
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -56,7 +56,7 @@ module Bundler
       # rubygems plugins sometimes hook into the gem install process
       Gem.load_env_plugins if Gem.respond_to?(:load_env_plugins)
 
-      Plugin.gemfile_install(Bundler.default_gemfile) if Bundler.settings[:plugin]
+      Plugin.gemfile_install(Bundler.default_gemfile) if Bundler.settings[:plugins]
 
       definition = Bundler.definition
       definition.validate_ruby!
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -96,6 +96,8 @@ module Bundler
       # rubygems plugins sometimes hook into the gem install process
       Gem.load_env_plugins if Gem.respond_to?(:load_env_plugins)
 
+      Bundler::Plugin.eval_gemfile(Bundler.default_gemfile)
+
       definition = Bundler.definition
       definition.validate_ruby!
 
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -17,28 +17,9 @@ module Bundler
         end
       end
 
-      if options[:without] && options[:with]
-        conflicting_groups = options[:without] & options[:with]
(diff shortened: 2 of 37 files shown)


lib/bundler/cli/plugin.rb
```diff
@@ -0,0 +1,23 @@
...
+module Bundler
+  class CLI::Plugin < Thor
+    desc "install PLUGIN", "Install the plugin from the source"
+    long_desc <<-D
+      Install a plugin named PLUGIN wither from the rubygems source provided (with --source option) or from a git source provided with (--git option).
+    D
+    method_option "source", :type => :string, :default => nil, :banner =>
+      "URL of the RubyGems source to fetch the plugin from"
+    method_option "version", :type => :string, :default => nil, :banner =>
+      "The version of the plugin to fetch from"
```
segiddins: ~~from~~

lib/bundler/dsl.rb
```diff
@@ -19,7 +19,7 @@ def self.evaluate(gemfile, lockfile, unlock)
 
     def initialize
       @source               = nil
-      @sources              = SourceList.new
+      @sources            ||= SourceList.new
```
segiddins: why is this change necessary?

lib/bundler/plugin.rb
```diff
@@ -0,0 +1,89 @@
+# frozen_string_literal: true
+
+module Bundler
+  class Plugin
+    autoload :Dsl,        "bundler/plugin/dsl"
+    autoload :Index,      "bundler/plugin/index"
+    autoload :Installer,  "bundler/plugin/installer"
+    autoload :SourceList, "bundler/plugin/source_list"
+
+    class << self
```
segiddins: since this class can't be instantiated, maybe it should be a module?

lib/bundler/plugin/dsl.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module Bundler
+  # Dsl to parse the Gemfile looking for plugins to install
+  class Plugin::Dsl < Bundler::Dsl
+    alias_method :_gem, :gem # To use for plugin installation as gem
+
+    # So that we don't have to override all there methods to dummy ones
+    # explicitly.
+    # They will be handled by missing_methods
```
segiddins: `method_missing`

lib/bundler/plugin/dsl.rb
```diff
@@ -0,0 +1,27 @@
...
+    [:gemspec, :gem, :path, :install_if, :platforms, :env].each {|m| undef_method m }
+
+    def initialize
+      @sources = Plugin::SourceList.new
+
+      super
```
segiddins: `super` should be called first

lib/bundler/source.rb
```diff
@@ -7,14 +7,22 @@ class Source
 
     attr_accessor :dependency_names
 
+    def initialize
+      @options = {}
```
segiddins: this is a bit of a code smell to me

segiddins:
If at all possible, it'd be great to get rid of all of the `if @options[:plugin]` -- in my eyes, it looks like that's introducing a look of branching where it may not be entirely necessary

asutoshpalai:
`@options[:plugin]` stemmed out of my attempt to use the existing source classes for fetch and install the plugins.

While writing that, I tired real hard to avoid `@options[:plugin]`. But the alternate solutions leads to a lot of code duplication or a major refactor of the the source classes.

I am still in lookout for solution and would definitely appreciate a few pointers.

I expect that development of source plugins might show me some way out of it, but research on those lines up till doesn't seem to thought any light in this direction.

asutoshpalai:
@segiddins : fixed the issues you pointed out (including the code smell :smile_cat: ). Removed `@options[:plugin]`.

The command function is also ready.

lib/bundler/cli/install.rb
```diff
@@ -77,25 +49,15 @@ def run
...
-      Bundler.settings.with       = options[:with]
+      normalize_settings
+
       Bundler::Fetcher.disable_endpoint = options["full-index"]
-      Bundler.settings[:disable_shared_gems] = Bundler.settings[:path] ? true : nil
 
       # rubygems plugins sometimes hook into the gem install process
       Gem.load_env_plugins if Gem.respond_to?(:load_env_plugins)
 
+      Plugin.eval_gemfile(Bundler.default_gemfile) if ENV["BUNDLE_PLUGIN"]
```
segiddins: it looks like this is being called for side effects, which is confusing

lib/bundler/cli/install.rb
```diff
@@ -190,5 +154,64 @@ def check_for_options_conflicts
...
+
+      options[:with]    = with
+      options[:without] = without
+    end
+
+    def normalize_settings
+      Bundler.settings[:path]     = nil if options[:system]
+      Bundler.settings[:path]     = "vendor/bundle" if options[:deployment]
+      Bundler.settings[:path]     = options["path"] if options["path"]
+      Bundler.settings[:path] ||= "bundle" if options["standalone"]
```
segiddins: the alignment in this method is now off
also, it'd be nice to have a blank line between the different settings?

lib/bundler/cli/plugin.rb
```diff
@@ -0,0 +1,23 @@
...
+      "URL of the RubyGems source to fetch the plugin from"
+    method_option "version", :type => :string, :default => nil, :banner =>
+      "The version of the plugin to fetch"
+    method_option "git", :type => :string, :default => nil, :banner =>
+      "URL of the git repo to fetch from"
+    method_option "branch", :type => :string, :default => nil, :banner =>
+      "The git branch to checkout"
+    method_option "ref", :type => :string, :default => nil, :banner =>
+      "The git revision to check out"
+    def install(plugin)
```
segiddins: any reason this couldn't install multiple plugins at once?

lib/bundler/plugin.rb
```diff
@@ -0,0 +1,131 @@
...
+
+module Bundler
+  module Plugin
+    autoload :Api,        "bundler/plugin/api"
+    autoload :Dsl,        "bundler/plugin/dsl"
+    autoload :Index,      "bundler/plugin/index"
+    autoload :Installer,  "bundler/plugin/installer"
+    autoload :SourceList, "bundler/plugin/source_list"
+
+    PLUGIN_FILE_NAME = "plugin.rb".freeze
```
segiddins: `plugins.rb`?

lib/bundler/plugin.rb
```diff
@@ -0,0 +1,131 @@
...
+
+    @commands = {}
+
+    class << self
```
segiddins: since this is never instantiated, maybe it should be a module with `module_function` instead?

lib/bundler/plugin/api.rb
```diff
@@ -0,0 +1,54 @@
...
+  # previous behavior or if otherwise desired, we can reimplement(or implement)
+  # the method to preserve compatibility.
+  #
+  # To use this, either the class can inherit this class or use it directly.
+  # For example of both types of use, refer the file `spec/plugins/command.rb`
+  #
+  # To use it without inheriting, you will have to create an object of this
+  # to use the functions (except for declaration functions like command, source,
+  # and hooks).
+  class Plugin::Api
```
segiddins: `API`

lib/bundler/plugin/dsl.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module Bundler
+  # Dsl to parse the Gemfile looking for plugins to install
+  class Plugin::Dsl < Bundler::Dsl
```
segiddins: `DSL`

lib/bundler/plugin/dsl.rb
```diff
@@ -0,0 +1,26 @@
...
+      super
+      @sources = Plugin::SourceList.new
+    end
+
+    def plugin(name, *args)
+      _gem(name, *args)
+    end
+
+    def method_missing(name, *args)
+      # Dummy evaluation
```
segiddins: can we please only ignore methods defined on `Bundler::Dsl` ? otherwise this could lead to bugs in this class

lib/bundler/plugin/index.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+require "yaml"
+
+module Bundler
+  # Manages which plugins are installed and their sources. This also is supposed to map
+  # which plugin does what (currently the features are not implemented so this class is
+  # now a stub class).
+  class Plugin::Index
```
segiddins: please nest the nested class

lib/bundler/plugin/index.rb
```diff
@@ -0,0 +1,73 @@
...
+    # the functions of the plugin to existing maps and also the name to source location.
+    #
+    # @param [String] name of the plugin to be registered
+    # @param [String] path where the plugin is installed
+    # @param [Array<String>] commands that are handled by the plugin
+    def register_plugin(name, path, commands)
+      @plugin_paths[name] = path
+
+      common = commands & @commands.keys
+      raise "Command(s) #{common.join(", ")} are already registered" if common.any?
```
segiddins: this should raise a specific error

lib/bundler/plugin/index.rb
```diff
@@ -0,0 +1,73 @@
...
+    end
+
+  private
+
+    # Reads the index file from the directory and initializes the instance variables.
+    def load_index
+      SharedHelpers.filesystem_access(index_file, :read) do |index_f|
+        valid_file = index_f && index_f.exist? && !index_f.size.zero?
+        break unless valid_file
+        index = YAML.load_file(index_f)
```
segiddins: if this can _ever_ be used in the same process as `bundle exec`, we can't use YAML

lib/bundler/plugin/installer.rb
```diff
@@ -0,0 +1,102 @@
...
+    # gems to their paths
+    #
+    # @param specs to install
+    #
+    # @return [Hash] map of names to path where the plugin was installed
+    def install_from_spec(specs)
+      paths = {}
+
+      specs.each do |spec|
+        next if spec.name == "bundler"
```
segiddins: is this really necessary?

lib/bundler/cli/install.rb
asutoshpalai: I was not able to find any better place for it.

Can you please suggest some

lib/bundler/plugin.rb
asutoshpalai: `module_function` doesn't provide the ability to define private methods on the module. That leaves us with two options: 1. Give up on singleton pattern for Plugin, 2. Make the private methods public.

I am aware that [bbatsov/ruby-style-guide](https://github.com/bbatsov/ruby-style-guide#module-function) recomends it, but I don't find it reflected in [GitHub's](https://github.com/airbnb/ruby) or [Airbnb's](https://github.com/styleguide/ruby)

lib/bundler/plugin/installer.rb
asutoshpalai: I can't remember why I wrote this, maybe some reminiscent of some other installation flow I was trying out.

Anyways it is not necessary now, so removing it

lib/bundler/cli/plugin.rb
asutoshpalai: Only that, I was not able to come up with some flow to accept the version/ref for multiple plugins.

segiddins: use the same for all of them?

lib/bundler/cli/install.rb
segiddins: well `eval_gemfile` suggests it will return something, if it doesn't then that method should be renamed

lib/bundler/plugin/index.rb
```diff
@@ -23,7 +30,7 @@ def register_plugin(name, path, commands)
         @plugin_paths[name] = path
 
         common = commands & @commands.keys
-        raise "Command(s) #{common.join(", ")} are already registered" if common.any?
+        raise CommandConflict.new(name, common) if common.any?
```
segiddins: `unless common.empty?`

lib/bundler/cli/plugin.rb
asutoshpalai: or use the default and raise error when multiple plugins are specified and a version is given?

segiddins: nah since you might want multiple gems at the same version, like if they're all in a suite of gems

lib/bundler/cli.rb
```diff
@@ -82,6 +82,10 @@ def help(cli = nil)
     end
 
     def self.handle_no_command_error(command, has_namespace = $thor_runner)
+      if ENV["BUNDLE_PLUGIN"] && Bundler::Plugin.command?(command)
```
segiddins: could this please using `Bundler.settings`?

lib/bundler/cli.rb
```diff
@@ -437,6 +441,12 @@ def env
       Env.new.write($stdout)
     end
 
+    if ENV["BUNDLE_PLUGIN"]
```
segiddins: settings

lib/bundler/plugin/api.rb
```diff
@@ -0,0 +1,56 @@
...
+      def cache
+        Plugin.cache.join("plugins")
+      end
+
+      # A tmp dir to be used by plugins
+      #
+      # @param [String] name unique for the plugin or the purpose
+      # @return [Pathname] object for the new directory created
+      def tmp(name)
+        Bundler.tmp(File.join(["plugin", name]))
```
segiddins: I think this can just be `Bundler.tmp("plugins", name)`

lib/bundler/plugin/index.rb
```diff
@@ -0,0 +1,116 @@
...
+          "commands" => @commands,
+        }
+
+        SharedHelpers.filesystem_access(index_file) do |index_f|
+          FileUtils.mkdir_p(index_f.dirname)
+          File.open(index_f, "w") {|f| f.puts dump_yaml(index) }
+        end
+      end
+
+      def dump_yaml(hash)
```
segiddins: the YAML stuff should be refactored so it can be shared with settings, and unit tested

lib/bundler/plugin/installer.rb
```diff
@@ -0,0 +1,99 @@
...
+      def install(names, options)
+        version = options[:version] || [">= 0"]
+
+        if options[:git]
+          install_git(names, version, options)
+        elsif options[:source]
+          source = options[:source]
+          install_rubygems(names, version, source)
+        else
+          raise(ArgumentError, "You need to provide the source")
```
segiddins: shouldn't it default to `Gem.sources` if no source is explicitly provided?

segiddins:
This is looking really good! Just for fun, could you try adding a spec for using plugins in an inline gemfile?

lib/bundler/cli.rb
asutoshpalai: I did that in  4a1baea, but some specs were failing because when called here (during the loading of `Cli` class)  `Bundler.settings` gets initialized before `CLI` class gets initialized.

So some options like `--gemfile` don't get properly set. Initializing `Settings` calls `Bundler.root` which caches the root path determined without respecting `--gemfile` option (which is set in [cli#initialize](https://github.com/bundler/bundler/blob/master/lib/bundler/cli.rb#L21)).

So reverted it cf284c0. I tried digging but failed to fix the problem.

lib/bundler/plugin/api.rb
asutoshpalai: `Bundler.tmp` doesn't accept splatted arguments. But :+1: on making name splatted, we can use `join`

lib/bundler/plugin/index.rb
asutoshpalai: I had already refactored it, just hadn't pushed it yet. I was trying to make the specs pass with `Settings` using it.
Made the specs pass :grinning:

lib/bundler/cli.rb
asutoshpalai: We can allow `plugin` sub command and show message if called but not enabled.

asutoshpalai:
Added thee default gem source, spec for inline gemfile and shifted settings to the new serializer. 

The only thing that remains is use of `ENV` as commented above. I believe that we should allow the sub command and show message when used.

asutoshpalai:
Using `Bundler.settings` now :dart:

segiddins:
Cool! Can you just make sure `plugins` is now a bool key in Settings?

lib/bundler/plugin.rb
```diff
@@ -0,0 +1,156 @@
...
+    # @param [Hash] options various parameters as described in description
+    # @option options [String] :source rubygems source to fetch the plugin gem from
+    # @option options [String] :version (optional) the version of the plugin to install
+    def install(names, options)
+      paths = Installer.new.install(names, options)
+
+      save_plugins paths
+    rescue PluginError => e
+      paths.values.map {|path| Bundler.rm_rf(path) } if paths
+      Bundler.ui.error "Failed to install plugin #{name}: #{e.message}\n  #{e.backtrace.join("\n  ")}"
```
segiddins: here only the first  line of the backtrace will be indented

lib/bundler/plugin.rb
```diff
@@ -0,0 +1,156 @@
...
+    # @param [Pathname] gemfile path
+    def gemfile_install(gemfile = nil, &inline)
+      if block_given?
+        builder = DSL.new
+        builder.instance_eval(&inline)
+        definition = builder.to_definition(nil, true)
+      else
+        definition = DSL.evaluate(gemfile, nil, {})
+      end
+      return unless definition.dependencies.any?
```
segiddins: `return if definition.dependencies.empty?`

lib/bundler/plugin.rb
```diff
@@ -0,0 +1,156 @@
...
+
+    # Checks if any plugins handles the command
+    def command?(command)
+      !index.command_plugin(command).nil?
+    end
+
+    # To be called from Cli class to pass the command and argument to
+    # approriate plugin class
+    def exec_command(command, args)
+      raise UndefinedCommandError, "Command #{command} not found" unless command? command
```
segiddins: the interpolated command in backticks

lib/bundler/plugin.rb
```diff
@@ -0,0 +1,156 @@
...
+
+    # Checks if the gem is good to be a plugin
+    #
+    # At present it only checks whether it contains plugins.rb file
+    #
+    # @param [Pathname] plugin_path the path plugin is installed at
+    # @raise [Error] if plugins.rb file is not found
+    def validate_plugin!(plugin_path)
+      plugin_file = plugin_path.join(PLUGIN_FILE_NAME)
+      raise MalformattedPlugin, "#{PLUGIN_FILE_NAME} was not found in the plugin!" unless plugin_file.file?
```
segiddins: no `!`

lib/bundler/plugin/api.rb
```diff
@@ -0,0 +1,56 @@
...
+
+      # A tmp dir to be used by plugins
+      # Accepts names that get concatenated as suffix
+      #
+      # @return [Pathname] object for the new directory created
+      def tmp(*names)
+        Bundler.tmp(["plugin", *names].join("-"))
+      end
+
+      def method_missing(name, *args, &blk)
```
segiddins: should also override `respond_to_missing?`

segiddins:
This gets a 👍 from me to ship in 1.13 (but disabled by default). There are some scenarios I'd like to see tested better (plugins with dependencies for example), but that can be done in a later PR. Well done, @asutoshpalai! @fotanus, I'm going to merge this by EOD unless you have any objections?

fotanus:
I haven't check the code closely enough, but I'm with you just because you are the most able person to do such decisions! 

The design is quite good by what I saw in the tests, and deviate little from the original proposal, so I think we are on point.

segiddins:
@homu r+

PR #4610, opened 2016-05-24
Add standalone flag to binstubs

### LINKED ISSUE #4594: Standalone configuration is not acknowledged by `binstub` command
Bundler Version: 1.12.4

I noticed some oddities when trying to setup only select binstubs with a standalone configuration. Running `bundle install --binstubs --standalone` adds the following to the binstubs but for all gems:

```ruby
$:.unshift File.expand_path '../../bundle', __FILE__

require 'bundler/setup'
```

However, if the process is to first run `bundle install --standalone` then run `bundle binstub GEM` the produced binstub does not include the standalone path:

```ruby
require "pathname"
ENV["BUNDLE_GEMFILE"] ||= File.expand_path("../../Gemfile",
  Pathname.new(__FILE__).realpath)

require "rubygems"
require "bundler/setup"
```

In this reproduction the produced `.bundle/config` appears as:

```
---
BUNDLE_PATH: bundle
BUNDLE_DISABLE_SHARED_GEMS: true
```

b-ggs:
Add `--standalone` flag to `binstubs`

Fix #4594 

@RochesterinNYC @segiddins

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -229,6 +229,8 @@ module Bundler
       "Overwrite existing binstubs if they exist"
     method_option "path", :type => :string, :lazy_default => "bin", :banner =>
       "Binstub destination directory (default bin)"
+    method_option "standalone", :type => :array, :lazy_default => [], :banner =>
+      "Make gem/s work without the Bundler runtime"
     def binstubs(*gems)
       require "bundler/cli/binstubs"
       Binstubs.new(options, gems).run
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -230,7 +230,7 @@ module Bundler
     method_option "path", :type => :string, :lazy_default => "bin", :banner =>
       "Binstub destination directory (default bin)"
     method_option "standalone", :type => :array, :lazy_default => [], :banner =>
-      "Make gem/s work without the Bundler runtime"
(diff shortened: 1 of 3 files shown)


spec/commands/binstubs_spec.rb
```diff
@@ -145,6 +145,21 @@
     end
   end
 
+  context "--standalone" do
```
RochesterinNYC: I think this context could be more specific/accurate. Something along the lines of "after installing with --standalone"? You could also move the `install_gemfile` and `bundle "install --standalone"` invocations to a before block that goes with the context then.

lib/bundler/cli.rb
```diff
@@ -229,6 +229,8 @@ def show(gem_name = nil)
       "Overwrite existing binstubs if they exist"
     method_option "path", :type => :string, :lazy_default => "bin", :banner =>
       "Binstub destination directory (default bin)"
+    method_option "standalone", :type => :array, :lazy_default => [], :banner =>
+      "Make gem/s work without the Bundler runtime"
```
segiddins: `Make binstubs that can work without the Bundler runtime` ?

RochesterinNYC:
This LGTM. @segiddins you want to homu r?

segiddins:
@homu delegate=RochesterinNYC

RochesterinNYC:
@homu r+

PR #4611, opened 2016-05-25
Fix the combination of gem :require and env or install_if

BrianHawley:
If you guard the installation of a gem with `env` or `install_if`, and that gem declaration has a `:require` option specified, it tried to require the gem even though it hadn't been installed. It looks like a spot was missed when the `env` command was added in the first place.

### DIFF
--- a/lib/bundler/runtime.rb
+++ b/lib/bundler/runtime.rb
@@ -68,9 +68,9 @@ module Bundler
       groups = [:default] if groups.empty?
 
       @definition.dependencies.each do |dep|
-        # Skip the dependency if it is not in any of the requested
-        # groups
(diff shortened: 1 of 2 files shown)


BrianHawley:
I also made sure that the tests would cover the problem situation.

segiddins:
Woah, I had no idea we even had an `env` method!

BrianHawley:
It's particularly useful with variables like "EXECJS_RUNTIME", so you make sure that it doesn't try to install one of the runtime gems unless it's going to be used, so as to resolve platform-specific compiler issues. Or at least it would be, once it no longer crashes in the require phase.

segiddins:
@homu r+

PR #4617, opened 2016-05-26
Add logic to check if the issue is path source related

### LINKED ISSUE #2175: Deployment fails when using paths in Gemfile
```ruby
gem "my-gem", path => "../../my-gem"
```

after running bundle package --all && bundler install --local Gemfile.lock contains new path -- "vendor/cache/my-gem"

Gemfile still contains old path as expected.

bundle install --deployment will end with:

<pre>
You have added to the Gemfile:
* source: source at ../../my-gem

You have deleted from the Gemfile:
* source: source at vendor/cache/my-gem

You have changed in the Gemfile:
* couchrest from `source at ../../my-gem` to `no specified source`
</pre>

Is it normal behavior and I must edit paths in Gemfile on deployment?

allenzhao:
Close #2175 

**Todo:**

- [x] Add spec test (Ref #2183)

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -366,9 +366,9 @@ module Bundler
       deleted_deps = @locked_deps - @dependencies
 
       # Check if it is possible that the source is only changed thing
-      if (new_deps.empty? && deleted_deps.empty?) && (new_sources.any? && deleted_sources.any?)
-        new_sources.reject! {|source| source.instance_of?(Bundler::Source::Path) && source.path.exist? }
-        deleted_sources.reject! {|source| source.instance_of?(Bundler::Source::Path) && source.path.exist? }
+      if (new_deps.empty? && deleted_deps.empty?) && (!new_sources.empty? && !deleted_sources.empty?)
+        new_sources.reject! {|source| source.is_a_path? && source.path.exist? }
+        deleted_sources.reject! {|source| source.is_a_path? && source.path.exist? }
       end
 
       if @locked_sources != gemfile_sources
@@ -382,7 +382,9 @@ module Bundler
       end
 
       added.concat new_deps.map {|d| "* #{pretty_dep(d)}" } if new_deps.any?
-      deleted.concat deleted_deps.map {|d| "* #{pretty_dep(d)}" } if deleted_deps.any?
+      if deleted_deps.any?
+        deleted.concat deleted_deps.map {|d| "* #{pretty_dep(d)}" }
+      end
 
       both_sources = Hash.new {|h, k| h[k] = [] }
       @dependencies.each {|d| both_sources[d.name][0] = d }
@@ -533,7 +535,15 @@ module Bundler
 
     def converge_dependencies
       (@dependencies + @locked_deps).each do |dep|
-        dep.source = sources.get(dep.source) if dep.source
+        locked_source = @locked_deps.select {|d| d.name == dep.name }.last
+        # This is to make sure that if bundler is installing in deployment mode and
+        # after locked_source and sources don't match, we still use locked_source.
+        if Bundler.settings[:frozen] && !locked_source.nil? &&
+            locked_source.respond_to?(:source) && locked_source.source.instance_of?(Source::Path) && locked_source.source.path.exist?
+          dep.source = locked_source.source
+        elsif dep.source
+          dep.source = sources.get(dep.source)
+        end
       end
       Set.new(@dependencies) != Set.new(@locked_deps)
     end
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -358,10 +358,20 @@ module Bundler
       changed = []
 
       gemfile_sources = sources.lock_sources
-      if @locked_sources != gemfile_sources
-        new_sources = gemfile_sources - @locked_sources
-        deleted_sources = @locked_sources - gemfile_sources
 
+      new_sources = gemfile_sources - @locked_sources
+      deleted_sources = @locked_sources - gemfile_sources
+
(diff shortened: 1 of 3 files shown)


segiddins:
@homu delegate=RochesterinNYC

lib/bundler/definition.rb
```diff
@@ -533,7 +533,14 @@ def converge_sources
 
     def converge_dependencies
       (@dependencies + @locked_deps).each do |dep|
-        dep.source = sources.get(dep.source) if dep.source
+        locked_source = @locked_deps.select {|d| d.name == dep.name }.last
```
allenzhao: It there a better way to find if the only difference of two dependencies is the sources' path? I've used dependency's name to find (which I assume is the identifier)

RochesterinNYC: I don't believe so. `name` is really the only accessible, applicable attribute of the `dependency` object.

allenzhao:
@RochesterinNYC Please review ;)

lib/bundler/definition.rb
```diff
@@ -358,10 +358,20 @@ def ensure_equivalent_gemfile_and_lockfile(explicit_flag = false)
...
 
+      new_sources = gemfile_sources - @locked_sources
+      deleted_sources = @locked_sources - gemfile_sources
+
+      new_deps = @dependencies - @locked_deps
+      deleted_deps = @locked_deps - @dependencies
+
+      # Check if it is possible that the source is only changed thing
+      if (new_deps.empty? && deleted_deps.empty?) && (new_sources.any? && deleted_sources.any?)
+        new_sources.reject! {|source| source.instance_of?(Bundler::Source::Path) && source.path.exist? }
```
segiddins: please change these to be `is_a?` checks

lib/bundler/definition.rb
```diff
@@ -358,10 +358,20 @@ def ensure_equivalent_gemfile_and_lockfile(explicit_flag = false)
...
-        deleted_sources = @locked_sources - gemfile_sources
 
+      new_sources = gemfile_sources - @locked_sources
+      deleted_sources = @locked_sources - gemfile_sources
+
+      new_deps = @dependencies - @locked_deps
+      deleted_deps = @locked_deps - @dependencies
+
+      # Check if it is possible that the source is only changed thing
+      if (new_deps.empty? && deleted_deps.empty?) && (new_sources.any? && deleted_sources.any?)
```
segiddins: please use `!empty?` instead of `any?`

lib/bundler/definition.rb
```diff
@@ -371,14 +381,8 @@ def ensure_equivalent_gemfile_and_lockfile(explicit_flag = false)
...
 
-      new_deps = @dependencies - @locked_deps
-      deleted_deps = @locked_deps - @dependencies
-
       added.concat new_deps.map {|d| "* #{pretty_dep(d)}" } if new_deps.any?
-
-      if deleted_deps.any?
-        deleted.concat deleted_deps.map {|d| "* #{pretty_dep(d)}" }
-      end
+      deleted.concat deleted_deps.map {|d| "* #{pretty_dep(d)}" } if deleted_deps.any?
```
segiddins: the condition is unnecessary here

allenzhao: Oh, actually I just changed the style to be consistent with L384.

lib/bundler/definition.rb
allenzhao: Actually, I think is_a? won't work because there's a Git subclass of Path and we don't want that got handled.

segiddins: in that case, can we think of an instance method on `Source` we can make this?

lib/bundler/definition.rb
segiddins: the condition is unnecessary there as well

RochesterinNYC: @allenzhao you can make a separate refactor commit removing those conditionals. There's no harm in concatenating an empty array (if `deleted_deps` was empty).

allenzhao: Sure, as for now I'll just force push to revert back and keep these two lines as-is

spec/install/deploy_spec.rb
```diff
@@ -263,4 +263,28 @@
       should_be_installed "rack 1.0.0"
     end
   end
+
+  context "with path in Gemfile and packed" do
```
RochesterinNYC: I'd suggest splitting up and reformatting these new specs instead of having them all in one `it` block?

allenzhao: I thought the logic are nested so I put them all in one it block

spec/install/deploy_spec.rb
```diff
@@ -263,4 +263,28 @@
...
+      install_gemfile <<-G
+      gem "foo", :path => "#{lib_path("foo")}"
+      G
+
+      bundle :install
+      should_be_installed "foo 1.0"
+      bundle "package --all"
+      expect(bundled_app("vendor/cache/foo")).to be_directory
+
+      bundle "install --local"
```
RochesterinNYC: It'd be great to get some further assertions here (maybe on output?) rather than just exitstatus.

spec/install/deploy_spec.rb
RochesterinNYC: Fair enough. Fine as is.

segiddins:
@RochesterinNYC r?

spec/install/deploy_spec.rb
RochesterinNYC: Great, thanks for adding the extra assertions.

RochesterinNYC:
@homu r+

segiddins:
@homu retry

PR #4618, opened 2016-05-26
[WIP] Properly detect path spec changes to avoid needless re-resolve

jrafanie:
* Use sets to compare dependencies
* Exclude dev dependencies when comparing indexes for equality
* Create an index from the Gemfile.lock's `@locked_specs` instead of from current gemspecs and use this for comparison to the current gemspecs.

More details...

Fixes an issue where we can mistakenly "think" the path gems have changed,
causing Bundler::Resolver.resolve to run again, merging in the prior lockfile
results.  We think the path gems have changed because the dependencies comparison cared about order using array == array.  We don't care about order so we can use set equality.  Additionally, we weren't properly detecting changes to path gems because we compared the current path gemspecs to themselves instead of the path specs from the lockfile.

**Before (with lockfile)**

time bundle check: 1.3 seconds
time bin/rails r 1: 7.2 seconds
rails boot memory[1]: 140 MB

**After (with lockfile)**

time bundle check: 0.46 seconds
time bin/rails r 1: 6.5 seconds
rails boot memory[1]: 125 MB

[1] `RAILS_ENV=production bin/rails r 'top = `/usr/bin/top -l 1 -pid #{Process.pid} | grep -E "ruby.+M"`; memory = top.split[7].to_i; puts memory'`

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -482,7 +482,7 @@ module Bundler
         end
       end
 
-      !locked || unlocking || dependencies_for_source_changed?(locked) || source.specs != locked.specs
+      !locked || unlocking || dependencies_for_source_changed?(source) || specs_for_source_changed?(source)
     end
 
     def dependencies_for_source_changed?(source)
@@ -490,6 +490,13 @@ module Bundler
       locked_deps_for_source = @locked_deps.select {|s| s.source == source }
 
       Set.new(deps_for_source) != Set.new(locked_deps_for_source)
+    end
+
+    def specs_for_source_changed?(source)
+      locked_index = Index.new
+      locked_index.use(@locked_specs.select {|s| source.can_lock?(s) })
+
+      source.specs != locked_index
     end
 
     # Get all locals and override their matching sources.
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -216,9 +216,11 @@ module Bundler
       @resolve ||= begin
         last_resolve = converge_locked_specs
         if Bundler.settings[:frozen] || (!@unlocking && nothing_changed?)
+          Bundler.ui.debug("Found no changes, using resolution from the lockfile")
           last_resolve
         else
           # Run a resolve against the locally available gems
+          Bundler.ui.debug("Found changes from the lockfile, re-resolving dependencies")
           last_resolve.merge Resolver.resolve(expanded_dependencies, index, source_requirements, last_resolve, ruby_version)
         end
       end
@@ -487,7 +489,7 @@ module Bundler
       deps_for_source = @dependencies.select {|s| s.source == source }
       locked_deps_for_source = @locked_deps.select {|s| s.source == source }
 
-      deps_for_source != locked_deps_for_source
+      Set.new(deps_for_source) != Set.new(locked_deps_for_source)
     end
 
     # Get all locals and override their matching sources.
--- a/lib/bundler/index.rb
+++ b/lib/bundler/index.rb
@@ -134,7 +134,7 @@ module Bundler
     def ==(other)
       all? do |spec|
         other_spec = other[spec].first
-        other_spec && (spec.dependencies & other_spec.dependencies).empty? && spec.source == other_spec.source
+        other_spec && Set.new(spec.dependencies) == Set.new(other_spec.dependencies) && spec.source == other_spec.source
       end
     end
 
--- a/lib/bundler/index.rb
+++ b/lib/bundler/index.rb
@@ -134,8 +134,14 @@ module Bundler
     def ==(other)
       all? do |spec|
         other_spec = other[spec].first
-        other_spec && Set.new(spec.dependencies) == Set.new(other_spec.dependencies) && spec.source == other_spec.source
+        other_spec && dependencies_eql?(spec, other_spec) && spec.source == other_spec.source
       end
+    end
+
+    def dependencies_eql?(spec, other_spec)
+      deps       = spec.dependencies.select {|d| d.type != :development }
+      other_deps = other_spec.dependencies.select {|d| d.type != :development }
+      Set.new(deps) == Set.new(other_deps)
     end
 
     def add_source(index)
--- a/spec/bundler/definition_spec.rb
+++ b/spec/bundler/definition_spec.rb
@@ -3,13 +3,12 @@ require "spec_helper"
 require "bundler/definition"
 
 describe Bundler::Definition do
-  before do
-    allow(Bundler).to receive(:settings) { Bundler::Settings.new(".") }
-    allow(Bundler).to receive(:default_gemfile) { Pathname.new("Gemfile") }
-    allow(Bundler).to receive(:ui) { double("UI", :info => "", :debug => "") }
-  end
-
   describe "#lock" do
+    before do
+      allow(Bundler).to receive(:settings) { Bundler::Settings.new(".") }
+      allow(Bundler).to receive(:default_gemfile) { Pathname.new("Gemfile") }
+      allow(Bundler).to receive(:ui) { double("UI", :info => "", :debug => "") }
+    end
     context "when it's not possible to write to the file" do
       subject { Bundler::Definition.new(nil, [], Bundler::SourceList.new, []) }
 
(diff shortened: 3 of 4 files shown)


spec/install/gemfile/path_spec.rb
```diff
@@ -13,6 +13,21 @@
...
   end
 
+  it "resolves from prior lockfile for source path gems" do
+    build_lib "rack-obama", "1.0", :path => lib_path("same_path")
+    build_lib "net-ssh", "1.0",    :path => lib_path("same_path")
+
+    install_gemfile <<-G
+      path "#{lib_path("same_path")}/" do
+        gem 'rack-obama'
+        gem 'net-ssh'
```
jrafanie: @segiddins I tried a few different combinations of existing gems for the test suite but not every pair of path based gems cause it to fail.  I don't know why the array order are not consistent.  

Keep me honest here, I think I did this test correctly.

spec/install/gemfile/path_spec.rb
```diff
@@ -13,6 +13,21 @@
...
+      end
+    G
+
+    bundle :check, :env => { "DEBUG" => 1 }
+    expect(out).to match(/Resolved from lockfile/)
```
jrafanie: I added a debug message for me to test against but thought might be useful going forward in case we need to debug a future problem where one of the "changed?" checks seems to do the wrong thing for people.

lib/bundler/definition.rb
```diff
@@ -200,8 +200,10 @@ def resolve
       @resolve ||= begin
         last_resolve = converge_locked_specs
         if Bundler.settings[:frozen] || (!@unlocking && nothing_changed?)
+          Bundler.ui.debug("Resolved from lockfile")
           last_resolve
         else
+          Bundler.ui.debug("Resolving again")
```
jrafanie: I'm open to better wording here...

spec/install/gemfile/path_spec.rb
jrafanie: Do I need to cleanup these gems and paths when I'm done?  I'm afraid I might be introducing some global state causing other tests to fail.  Maybe you have some ideas?

segiddins: Nope, none of this is introducing any leaky state

lib/bundler/definition.rb
jrafanie: Note, this is the exact place I put `STDERR.puts` calls based on @segiddins guidance that if we have a lockfile, we shouldn't be entering the "Resolving again" condition.  I added this logging for the test specifically, but I also think it's useful if people report  "bundle exec XYZ is slow".

This is what it looks like:

```
$ DEBUG=1 bundle check
Resolved from lockfile
The Gemfile's dependencies are satisfied
```

Now, I added a gem to my Gemfile...

```
$ DEBUG=1 bundle check
Resolving again
Bundler can't satisfy your Gemfile's dependencies.
```

jrafanie:
I re-pushed with changes to rename specs to dependencies in the commit message since the change is on current and locked deps and not specs themselves.

jrafanie:
@segiddins With @fryguy's help, the tests should now be passing.

Thanks for the help tracking this down by saying "You know, you really shouldn't be resolving if you have a lockfile..."  👍

jrafanie:
@segiddins I can add a changelog once you review these changes, especially the new Bundler.ui messages

segiddins:
@jrafanie don't worry about the CHANGELOG, we compile that when doing releases. I'll try and review this more in-depth soon, I want to see if it makes sense to just make some of these properties sets to begin with for performance reasons

jrafanie:
> @jrafanie don't worry about the CHANGELOG, we compile that when doing releases. I'll try and review this more in-depth soon, I want to see if it makes sense to just make some of these properties sets to begin with for performance reasons

Good point.  Yes, we have SpecSets for some things and it felt like some of these could be Sets from the start but it looked like too much of a change for a surgical fix.  I want this in a patch release if possible 😉

segiddins:
Sorry to say but this is unlikely to go in a patch release -- we generally reserve patch fixes for bugs that were introduced in the latest release. Otoh, this would fit in nicely with the performance improvements I have open PRs for, so 1.13 shouldn't be a very long release cycle

jrafanie:
> Sorry to say but this is unlikely to go in a patch release -- we generally reserve patch fixes for bugs that were introduced in the latest release. Otoh, this would fit in nicely with the performance improvements I have open PRs for, so 1.13 shouldn't be a very long release cycle

I understand.  Thanks for the support.  🙇

jrafanie:
Marking as WIP, I am seeing a problem locally I need to verify before this should be merged.

jrafanie:
I'm stumped... it looks like this still enters the re-resolve path because the `source.specs != locked.specs` check [here](https://github.com/bundler/bundler/blob/7374a01a74bf7ee85619248db48504bc5ce3849e/lib/bundler/definition.rb#L456) is only true for if path indexes are == , where they're only equal [if index specs share no dependencies](https://github.com/bundler/bundler/blob/7374a01a74bf7ee85619248db48504bc5ce3849e/lib/bundler/index.rb#L138).

I need to investigate more.

segiddins:
@jrafanie I have no idea why that;'s there, it dates back to https://github.com/bundler/bundler/commit/d9c8f028524acd8cdf5951454d3a7f46d02d960a. Maybe try removing it and see if the specs still pass?

jrafanie:
Thanks @segiddins for the suggestion, I'll give it a try and see how it goes.

lib/bundler/definition.rb
```diff
@@ -200,8 +200,10 @@ def resolve
       @resolve ||= begin
         last_resolve = converge_locked_specs
         if Bundler.settings[:frozen] || (!@unlocking && nothing_changed?)
+          Bundler.ui.debug("Resolved from lockfile")
```
segiddins: `"Found no changes, using resolution from the lockfile"`

lib/bundler/definition.rb
segiddins: `"Found changes from the lockfile, re-resolving dependencies"`

lib/bundler/definition.rb
jrafanie: 👍 Thanks!  I'll be looking at this in the next few days.  I really want to track this thing down.

segiddins:
@homu r+

jrafanie:
Thanks @segiddins, with @fryguy's help, I'm only a few failures away. Getting there.

spec/install/gemfile/path_spec.rb
```diff
@@ -399,7 +433,7 @@
     end
 
     it "unlocks all gems when a child dependency gem is updated" do
-      build_lib "bar", "2.0", :path => lib_path("foo/bar")
+      build_lib "bar", "2.0", :path => lib_path("foo")
```
jrafanie: These two changes are questionable.  Without them, this test fails on the should_be_installed for bar, trying to require bar, which can't be found.

spec/install/gemfile/path_spec.rb
```diff
@@ -418,6 +452,7 @@
     end
 
     it "gets dependencies that are updated in the path" do
+      build_lib "rack", "1.0.0", :path => lib_path("foo")
```
jrafanie: This looks like a bad test that was somehow hidden by the dependency order causing a re-resolve.  We depend on rack, we're using a path in the Gemfile, and we weren't building rack in the path, so it can't be found.  I'm not sure how this passed previously, unless, like the above tests, the bug I'm fixing caused these bad tests to quietly pass.

spec/bundler/definition_spec.rb
```diff
@@ -6,7 +6,7 @@
   before do
     allow(Bundler).to receive(:settings) { Bundler::Settings.new(".") }
     allow(Bundler).to receive(:default_gemfile) { Pathname.new("Gemfile") }
-    allow(Bundler).to receive(:ui) { double("UI", :info => "") }
+    allow(Bundler).to receive(:ui) { double("UI", :info => "", :debug => "") }
```
jrafanie: We're adding debug messages so we need to update the double for that expectation.

jrafanie:
🙏   I'm hoping this is now green on travis.

jrafanie:
Please don't merge yet.  @fryguy and I are uncovering more bugs in the Path code after we fixed Bundler "always re-resolving with Path gems (when there are no changes)".  We'll try to see how we can fix them independently.

lib/bundler/definition.rb
```diff
@@ -480,14 +482,21 @@ def specs_changed?(source, &block)
         end
       end
 
-      !locked || unlocking || dependencies_for_source_changed?(locked) || source.specs != locked.specs
+      !locked || unlocking || dependencies_for_source_changed?(source) || specs_for_source_changed?(source)
```
jrafanie: This change, `dependencies_for_source_changed?(source)` seems to be more about readability.  `source` is the current Source(path, git, etc.) containing the gemspecs from the filesystem for our bundle.  `locked` is the locked Source(path, git, rubygems, etc.) for the same Source as defined in the Gemfile.lock.  `dependencies_for_source_changed?` uses the parameter to find dependencies from our lockfile or gemspecs on the filesystem for that source.  This should have no difference for a source coming from the lockfile or gemspec sources.

lib/bundler/definition.rb
```diff
@@ -480,14 +482,21 @@ def specs_changed?(source, &block)
...
     end
 
     def dependencies_for_source_changed?(source)
       deps_for_source = @dependencies.select {|s| s.source == source }
       locked_deps_for_source = @locked_deps.select {|s| s.source == source }
 
-      deps_for_source != locked_deps_for_source
+      Set.new(deps_for_source) != Set.new(locked_deps_for_source)
```
jrafanie: Do Set comparison since order doesn't matter.

lib/bundler/definition.rb
jrafanie: `source.specs != locked.specs` is the key problem.

`source.specs` takes our source of gemspecs and enumerates the specs for that source by finding all the gemspecs for that source from the filesystem, etc.

`locked.specs` takes our Gemfile.lock version of this same source and calls specs, which does the SAME THING.  It finds the gemspecs for the location specified from the Gemfile.lock, which is the same as the source, and gets the exact same list.  This doesn't actually produce the list of gemspecs from our lockfile.

lib/bundler/definition.rb
```diff
@@ -480,14 +482,21 @@ def specs_changed?(source, &block)
...
+    end
+
+    def specs_for_source_changed?(source)
+      locked_index = Index.new
+      locked_index.use(@locked_specs.select {|s| source.can_lock?(s) })
```
jrafanie: We've changed the logic to build an index using the already parsed `@locked_specs` and compare that to what the source currently has.

lib/bundler/definition.rb
```diff
@@ -480,14 +482,21 @@ def specs_changed?(source, &block)
...
+
+      source.specs != locked_index
```
jrafanie: I think there's still a bug here because `source.specs` for path gems includes development gems, while locked_specs does not.  I think we need to do something similar to how converge_locked_specs discards dev gems on line 635 below:

 [deps2 = other.dependencies.select {|d| d.type != :development }](https://github.com/bundler/bundler/pull/4618/files#diff-ec267c71117bf68f38c928a3691ba98cR635)  cc @fryguy

jrafanie: I solved this in the Index#== method by discarding dev dependencies for the comparison.  An alternative would be to build the index here with only non-dev dependencies but the locked_specs already has sources with dev dependencies in them...

lib/bundler/index.rb
```diff
@@ -134,10 +134,16 @@ def size
...
       all? do |spec|
         other_spec = other[spec].first
-        other_spec && (spec.dependencies & other_spec.dependencies).empty? && spec.source == other_spec.source
+        other_spec && dependencies_eql?(spec, other_spec) && spec.source == other_spec.source
       end
     end
 
+    def dependencies_eql?(spec, other_spec)
+      deps       = spec.dependencies.select {|d| d.type != :development }
+      other_deps = other_spec.dependencies.select {|d| d.type != :development }
```
jrafanie: Can we say that all index comparisons should disregard development dependencies?  I don't know.  How does it work when you have a gem you're developing, you have dev dependencies in your gemspec, you have `gemspec` in your Gemfile and you run bundle.  It installs dev dependencies now.  Does it use an index to do comparisons?

jrafanie: Note:  I don't think many things were relying on this equality method since the original `(spec.dependencies & other_spec.dependencies).empty?` seems to be completely wrong.  Indexes are only equal if they share NO dependencies?  I don't even know what that means.

jrafanie:
@segiddins @fryguy I think this is ready for review.  I kept the commits separate to help explain what @fryguy and I did.  Note, the Index#== change to remove dev dependencies for comparison of indexes fixes my issue with the locked index having dev deps while the source index does not.  But, I have no idea if that's a valid thing to do... tests are passing though.  Are there missing tests or anything else we can do?

segiddins:
@indirect r?

segiddins:
:shipit:

PR #4625, opened 2016-05-28
Fixes specifying ruby patch level in gemfile and gemspec

### LINKED ISSUE #4593: If gemspec specifies Ruby patch level, `bundle update` emits uninformative message
I'm writing my first gem. After `bundle gem {name}`, reading RubyGems' API, I supplemented `{name}.gemspec` suitably.

`gem build {name}.gemspec` succeeded. However, `bundle update` gave this mysterious error message:

    Could not find gem '{name}' in source at `.`.
    Source contains '{name}' at: 0.1.0 ruby
Currently, RubyGems' API states 'The Ruby version can be specified to the patch level'. It documents method [Gem::Specification#required_ruby_version=](http://docs.seattlerb.org/rubygems/Gem/Specification.html#method-i-required_ruby_version-3D) as:
```
required_ruby_version=(req)
The version of Ruby required by this gem. The ruby version can be specified to the patch-level:
$ ruby -v -e 'p Gem.ruby_version'
ruby 2.0.0p247 (2013-06-27 revision 41674) [x86_64-darwin12.4.0]
#<Gem::Version "2.0.0.247">
Because patch-level is taken into account[...]
```
Following the API, I had specified my Ruby version (including patch level) in `{name}.gemspec` as follows:

    spec.required_ruby_version = '2.3.1.112'
And I'd specified my Ruby version (including patch level) in `Gemfile` as follows:

    ruby '2.3.1', engine_version: '2.3.1', engine: 'ruby', patchlevel: '112'
Ultimately, divide-and-conquer determined Bundler seemed to be confused by the four-numbered Ruby version specification (in `{name}.gemspec`). And `rm -f Gemfile.lock; bundle update` worked if I dropped the final, patch number.

It wasn't (and isn't) obvious to me why these two Ruby version specifications don't match (for Bundler).

Perhaps (from Bundler) a more informative error message would help other first-gem writers, who follow RubyGems' API in this way?

I've created a minimal example (using current versions) for which Bundler generates the above error message:

```bash
$ pwd
/home/mark/develop

$ bundle gem bundle-gem-error
Creating gem 'bundle-gem-error'...
MIT License enabled in config
Code of conduct enabled in config
      create  bundle-gem-error/Gemfile
      create  bundle-gem-error/.gitignore
      create  bundle-gem-error/lib/bundle/gem/error.rb
      create  bundle-gem-error/lib/bundle/gem/error/version.rb
      create  bundle-gem-error/bundle-gem-error.gemspec
      create  bundle-gem-error/Rakefile
      create  bundle-gem-error/README.md
      create  bundle-gem-error/bin/console
      create  bundle-gem-error/bin/setup
      create  bundle-gem-error/.travis.yml
      create  bundle-gem-error/test/test_helper.rb
      create  bundle-gem-error/test/bundle/gem/error_test.rb
      create  bundle-gem-error/LICENSE.txt
      create  bundle-gem-error/CODE_OF_CONDUCT.md
Initializing git repo in /home/mark/develop/bundle-gem-error

$ nano bundle-gem-error/Gemfile
$ nano bundle-gem-error/bundle-gem-error.gemspec

$ cd bundle-gem-error
RVM used your Gemfile for selecting Ruby, it is all fine - Heroku does that too,
you can ignore these warnings with 'rvm rvmrc warning ignore /home/mark/develop/bundle-gem-error/Gemfile'.
To ignore the warning for all files run 'rvm rvmrc warning ignore allGemfiles'.

ruby-2.3.1,engine_version:2.3.1,engine:ruby-p112 is not installed.
To install do: 'rvm install ruby-2.3.1,engine_version:2.3.1,engine:ruby-p112'

$ pwd
/home/mark/develop/bundle-gem-error

$ ls
bin  bundle-gem-error.gemspec  CODE_OF_CONDUCT.md  Gemfile  lib  LICENSE.txt  Rakefile  README.md  test

$ gem build bundle-gem-error.gemspec
  Successfully built RubyGem
  Name: bundle-gem-error
  Version: 0.1.0
  File: bundle-gem-error-0.1.0.gem

$ gem env
RubyGems Environment:
  - RUBYGEMS VERSION: 2.6.4
  - RUBY VERSION: 2.3.1 (2016-04-26 patchlevel 112) [x86_64-linux]
  - INSTALLATION DIRECTORY: /home/mark/.rvm/gems/ruby-2.3.1
  - USER INSTALLATION DIRECTORY: /home/mark/.gem/ruby/2.3.0
  - RUBY EXECUTABLE: /home/mark/.rvm/rubies/ruby-2.3.1/bin/ruby
  - EXECUTABLE DIRECTORY: /home/mark/.rvm/gems/ruby-2.3.1/bin
  - SPEC CACHE DIRECTORY: /home/mark/.gem/specs
  - SYSTEM CONFIGURATION DIRECTORY: /home/mark/.rvm/rubies/ruby-2.3.1/etc
  - RUBYGEMS PLATFORMS:
    - ruby
    - x86_64-linux
  - GEM PATHS:
     - /home/mark/.rvm/gems/ruby-2.3.1
     - /home/mark/.rvm/gems/ruby-2.3.1@global
  - GEM CONFIGURATION:
     - :update_sources => true
     - :verbose => true
     - :backtrace => false
     - :bulk_threshold => 1000
  - REMOTE SOURCES:
     - https://rubygems.org/
  - SHELL PATH:
     - /home/mark/.rvm/gems/ruby-2.3.1/bin
     - /home/mark/.rvm/gems/ruby-2.3.1@global/bin
     - /home/mark/.rvm/rubies/ruby-2.3.1/bin
     - /usr/local/bin
     - /usr/bin
     - /bin
     - /usr/local/games
     - /usr/games
     - /home/mark/.rvm/bin

$ rvm --version
rvm 1.27.0 (latest) by Wayne E. Seguin <wayneeseguin@gmail.com>, Michal Papis <mpapis@gmail.com> [https://rvm.io/]

$ cat /etc/debian_version
8.4

$ lsb_release -da
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 8.4 (jessie)
Release:	8.4
Codename:	jessie

$ hostnamectl
   Static hostname: mycroftdebian
         Icon name: computer-vm
           Chassis: vm
        Machine ID: 942aa37ac16443c5b7f04bc7491aa97c
           Boot ID: d97dbd7eb01a42b1b6d5732f72228f3c
    Virtualization: kvm
  Operating System: Debian GNU/Linux 8 (jessie)
            Kernel: Linux 3.16.0-4-amd64
      Architecture: x86-64

$ cat /proc/version
Linux version 3.16.0-4-amd64 (debian-kernel@lists.debian.org) (gcc version 4.8.4 (Debian 4.8.4-1) ) #1 SMP Debian 3.16.7-ckt25-2 (2016-04-08)
```

chalkos:
- [x] add failing spec
- [x] add fix (with passing spec) and more specs to cover edge cases
- [x] ~~change error message to a more meaningful one~~ fixed the bug and a meaningful error message is now being shown

spec always passes on 2.2.0p0, probably because the patch version is 0

### DIFF
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -125,7 +125,7 @@ module Bundler
       def for?(platform, required_ruby_version)
         if spec = @specs[platform]
           if required_ruby_version && spec.respond_to?(:required_ruby_version) && spec_required_ruby_version = spec.required_ruby_version
-            spec_required_ruby_version.satisfied_by?(required_ruby_version.gem_version)
+            spec_required_ruby_version.satisfied_by?(required_ruby_version.to_gem_version_with_patchlevel)
           else
             true
           end
--- a/lib/bundler/ruby_version.rb
+++ b/lib/bundler/ruby_version.rb
@@ -100,6 +100,15 @@ module Bundler
       @ruby_version ||= RubyVersion.new(RUBY_VERSION.dup, RUBY_PATCHLEVEL.to_s, ruby_engine, ruby_engine_version)
     end
 
+    def to_gem_version_with_patchlevel
+      @gem_version_with_patch ||= begin
+        patch_number = @patchlevel ? Gem::Requirement.create(@patchlevel).requirements.first.last : 0
+        Gem::Requirement.create("#{@gem_version}.#{patch_number}").requirements.first.last
+      rescue BadRequirementError
+        Gem::Requirement.create("#{@gem_version}.0").requirements.first.last
+      end
+    end
+
   private
 
     def matches?(requirements, version)
--- a/lib/bundler/ruby_version.rb
+++ b/lib/bundler/ruby_version.rb
@@ -102,10 +102,9 @@ module Bundler
 
     def to_gem_version_with_patchlevel
       @gem_version_with_patch ||= begin
-        patch_number = @patchlevel ? Gem::Requirement.create(@patchlevel).requirements.first.last : 0
-        Gem::Requirement.create("#{@gem_version}.#{patch_number}").requirements.first.last
-      rescue BadRequirementError
-        Gem::Requirement.create("#{@gem_version}.0").requirements.first.last
+        Gem::Version.create("#{@gem_version}.#{@patchlevel}")
+      rescue ArgumentError
+        @gem_version
       end
     end
 
--- a/spec/install/gemspecs_spec.rb
+++ b/spec/install/gemspecs_spec.rb
@@ -46,4 +46,76 @@ describe "bundle install" do
     bundle :install, :artifice => "endpoint_marshal_fail" # force gemspec load
     should_be_installed "activesupport 2.3.2"
   end
+
+  context "when ruby version is specified in gemspec and gemfile" do
+    it "installs when patch level is not specified and the version matches" do
+      build_lib("foo", :path => tmp.join("foo")) do |s|
+        s.write "Gemfile", <<-G
+        source "file://#{gem_repo1}"
+        ruby '#{RUBY_VERSION}', :engine_version => '#{RUBY_VERSION}', :engine => 'ruby'
+        gemspec
+        G
+        s.required_ruby_version = RUBY_VERSION
+      end
+
+      Dir.chdir(tmp.join("foo")) do
+        bundle "install"
+        should_be_installed "foo 1.0"
+      end
+    end
+
+    it "installs when patch level is specified and the version still matches the current version" do
+      build_lib("foo", :path => tmp.join("foo")) do |s|
+        s.write "Gemfile", <<-G
+        source "file://#{gem_repo1}"
+        ruby '#{RUBY_VERSION}', :engine_version => '#{RUBY_VERSION}', :engine => 'ruby', :patchlevel => '#{RUBY_PATCHLEVEL}'
+        gemspec
+        G
+        s.required_ruby_version = "#{RUBY_VERSION}.#{RUBY_PATCHLEVEL}"
+      end
+
+      Dir.chdir(tmp.join("foo")) do
+        bundle "install"
+        should_be_installed "foo 1.0"
+      end
+    end
+
+    it "fails and complains about patchlevel on patchlevel mismatch" do
+      patchlevel = RUBY_PATCHLEVEL.to_i + 1
+      build_lib("foo", :path => tmp.join("foo")) do |s|
+        s.write "Gemfile", <<-G
+        source "file://#{gem_repo1}"
+        ruby '#{RUBY_VERSION}', :engine_version => '#{RUBY_VERSION}', :engine => 'ruby', :patchlevel => '#{patchlevel}'
+        gemspec
+        G
+        s.required_ruby_version = "#{RUBY_VERSION}.#{patchlevel}"
+      end
+
+      Dir.chdir(tmp.join("foo")) do
+        bundle "install"
+        expect(out).to include("Ruby patchlevel")
+        expect(out).to include("but your Gemfile specified")
+        expect(exitstatus).to eq(18) if exitstatus
+      end
+    end
+
+    it "fails and complains about version on version mismatch" do
+      version = Gem::Requirement.create(RUBY_VERSION).requirements.first.last.bump.version
+      build_lib("foo", :path => tmp.join("foo")) do |s|
+        s.write "Gemfile", <<-G
+        source "file://#{gem_repo1}"
+        ruby '#{version}', :engine_version => '#{version}', :engine => 'ruby'
+        gemspec
+        G
+        s.required_ruby_version = version
+      end
+
+      Dir.chdir(tmp.join("foo")) do
+        bundle "install"
+        expect(out).to include("Ruby version")
+        expect(out).to include("but your Gemfile specified")
+        expect(exitstatus).to eq(18) if exitstatus
+      end
+    end
+  end
 end
--- a/spec/install/gemspecs_spec.rb
+++ b/spec/install/gemspecs_spec.rb
@@ -49,73 +49,60 @@ describe "bundle install" do
 
   context "when ruby version is specified in gemspec and gemfile" do
     it "installs when patch level is not specified and the version matches" do
-      build_lib("foo", :path => tmp.join("foo")) do |s|
-        s.write "Gemfile", <<-G
-        source "file://#{gem_repo1}"
-        ruby '#{RUBY_VERSION}', :engine_version => '#{RUBY_VERSION}', :engine => 'ruby'
-        gemspec
-        G
+      build_lib("foo", :path => bundled_app) do |s|
         s.required_ruby_version = RUBY_VERSION
       end
 
-      Dir.chdir(tmp.join("foo")) do
-        bundle "install"
-        should_be_installed "foo 1.0"
-      end
+      install_gemfile <<-G
+        ruby '#{RUBY_VERSION}', :engine_version => '#{RUBY_VERSION}', :engine => 'ruby'
+        gemspec
+      G
+      should_be_installed "foo 1.0"
     end
 
     it "installs when patch level is specified and the version still matches the current version" do
-      build_lib("foo", :path => tmp.join("foo")) do |s|
-        s.write "Gemfile", <<-G
-        source "file://#{gem_repo1}"
-        ruby '#{RUBY_VERSION}', :engine_version => '#{RUBY_VERSION}', :engine => 'ruby', :patchlevel => '#{RUBY_PATCHLEVEL}'
-        gemspec
-        G
+      build_lib("foo", :path => bundled_app) do |s|
         s.required_ruby_version = "#{RUBY_VERSION}.#{RUBY_PATCHLEVEL}"
       end
 
-      Dir.chdir(tmp.join("foo")) do
-        bundle "install"
-        should_be_installed "foo 1.0"
-      end
+      install_gemfile <<-G
+        ruby '#{RUBY_VERSION}', :engine_version => '#{RUBY_VERSION}', :engine => 'ruby', :patchlevel => '#{RUBY_PATCHLEVEL}'
+        gemspec
+      G
+      should_be_installed "foo 1.0"
     end
 
     it "fails and complains about patchlevel on patchlevel mismatch" do
       patchlevel = RUBY_PATCHLEVEL.to_i + 1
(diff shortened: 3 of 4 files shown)


chalkos:
My findings so far:

(On brach master)
In Bundler::Resolver::SpecGroup#for? (bundler/resolver.rb) in `spec_required_ruby_version.satisfied_by?(required_ruby_version.gem_version)`, `spec_required_ruby_version` has a ruby version with the patch number, but `required_ruby_version.gem_version` only contains the version without patch number (is it present in `required_ruby_version.patch_version`).

@segiddins do you think `required_ruby_version.gem_version` should include the patch number? I can track down where it is being set and make sure the patch number gets included. I believe this would be the best way to fix this, but as it may also have unintended effects on other parts of the code I could also change `Bundler::Resolver::SpecGroup#for?` to use a patch version when it is present.

The `Gem::Version` assumes a patch number of 0, that is why the tests for `ruby 2.2.0p0` pass: because `Gem::Requirement#satisfied_by?` returns true for that case.

segiddins:
@chalkos let's just see if changing `gem_version` fixes things?

lib/bundler/resolver.rb
```diff
@@ -131,7 +131,7 @@ def source
       def for?(platform, required_ruby_version)
         if spec = @specs[platform]
           if required_ruby_version && spec.respond_to?(:required_ruby_version) && spec_required_ruby_version = spec.required_ruby_version
-            spec_required_ruby_version.satisfied_by?(required_ruby_version.gem_version)
+            spec_required_ruby_version.satisfied_by?(Gem::Version.new("#{required_ruby_version.gem_version}.#{required_ruby_version.patchlevel}"))
```
segiddins: `required_ruby_version.patchlevel` can be nil, and this should be a method on `Bundler::RubyVersion`

chalkos: Thanks. I tried changing `Bundler::RubyVersion#initialize` to always include the patch version in `@gem_version`, but some tests started failing. From your experience, do you think it would be better to do this or handle this specific case in `#for?` ?

Maybe in some parts of bundler it is preferable not to have the patch version?

segiddins: Maybe just a `gem_version_with_patch`?

chalkos: That seems cleaner and reusable. I'll do that.

chalkos:
I think this could use some more specs.
The new method in `Bundler::RubyVersion` could (needs to?) be tested.
Regarding #4593, I think some more tests could be written, specially for some edge cases.

lib/bundler/ruby_version.rb
```diff
@@ -100,6 +100,19 @@ def self.system
       @ruby_version ||= RubyVersion.new(RUBY_VERSION.dup, RUBY_PATCHLEVEL.to_s, ruby_engine, ruby_engine_version)
     end
 
+    def gem_version_with_patch
+      unless @gem_version_with_patch
+        @gem_version_with_patch = Gem::Requirement.create("#{@gem_version}.0").requirements.first.last
+        if @patchlevel
+          begin
```
segiddins: the `begin` is unnecessary

lib/bundler/ruby_version.rb
```diff
@@ -100,6 +100,19 @@ def self.system
       @ruby_version ||= RubyVersion.new(RUBY_VERSION.dup, RUBY_PATCHLEVEL.to_s, ruby_engine, ruby_engine_version)
     end
 
+    def gem_version_with_patch
+      unless @gem_version_with_patch
```
segiddins: `||=`

lib/bundler/ruby_version.rb
```diff
@@ -100,6 +100,19 @@ def self.system
...
+        @gem_version_with_patch = Gem::Requirement.create("#{@gem_version}.0").requirements.first.last
```
segiddins: this unconditionally creates an object that might not be used

lib/bundler/ruby_version.rb
chalkos: right! 👍

lib/bundler/ruby_version.rb
chalkos: it's meant to handle the `BadRequirementError` exception. But I'm not really sure this even works for that case. I found a better way to do this already

lib/bundler/ruby_version.rb
chalkos: right. my next commit will change this

segiddins:
@chalkos is this still WIP?

chalkos:
Yes. I still need to add more specs to cover edge cases and to RubyVersion's new method and change error message to a more meaningful one.

lib/bundler/ruby_version.rb
```diff
@@ -100,6 +100,15 @@ def self.system
       @ruby_version ||= RubyVersion.new(RUBY_VERSION.dup, RUBY_PATCHLEVEL.to_s, ruby_engine, ruby_engine_version)
     end
 
+    def to_gem_version_with_patchlevel
+      @gem_version_with_patch ||= begin
+        patch_number = @patchlevel ? Gem::Requirement.create(@patchlevel).requirements.first.last : 0
+        Gem::Requirement.create("#{@gem_version}.#{patch_number}").requirements.first.last
```
segiddins: this can just be `Gem::Version.create("#{@gem_version}.#{patch_number}")`

lib/bundler/ruby_version.rb
```diff
@@ -100,6 +100,15 @@ def self.system
...
+      rescue BadRequirementError
+        Gem::Requirement.create("#{@gem_version}.0").requirements.first.last
```
segiddins: this can just be `@gem_version`

chalkos: I wanted to give it a better name than that. `@gem_version` does not seem like "a representation of the RubyVersion as a Gem::Version object", which is what the instance variable represents.

segiddins: No i mean there's no reason to create a new version object in this case

chalkos: Sorry, I misunderstood and thought you were talking about something else.
You are correct, I'll change this to `@gem_version`

chalkos:
@segiddins after making the changes you suggested, would you like me to separate this in 2 or 3 commits? I think we could have one with the initial specs and then another one or two with the fix and the RubyVersion specs.

lib/bundler/ruby_version.rb
```diff
@@ -100,6 +100,15 @@ def self.system
       @ruby_version ||= RubyVersion.new(RUBY_VERSION.dup, RUBY_PATCHLEVEL.to_s, ruby_engine, ruby_engine_version)
     end
 
+    def to_gem_version_with_patchlevel
+      @gem_version_with_patch ||= begin
+        patch_number = @patchlevel ? Gem::Requirement.create(@patchlevel).requirements.first.last : 0
```
segiddins: why can't this just be `@patchlevel` ?

spec/install/gemspecs_spec.rb
```diff
@@ -46,4 +46,76 @@
...
+
+      Dir.chdir(tmp.join("foo")) do
+        bundle "install"
+        should_be_installed "foo 1.0"
+      end
+    end
+
+    it "installs when patch level is specified and the version still matches the current version" do
+      build_lib("foo", :path => tmp.join("foo")) do |s|
+        s.write "Gemfile", <<-G
```
segiddins: please use `install_gemfile` to install a gemfile, rather than writing it into the lib path

lib/bundler/ruby_version.rb
chalkos: I think it's exactly the same. I'll change it, yours is simpler.
The purpose is to avoid raising and catching an exception when `@patchlevel` is nil. Although the exception is still risen and caught if `@patchlevel` is negative or some other invalid value.

I think that performance-wise it better to handle these kind of situations using conditions instead of catching the exception. But if I were to fully validate the `@patchlevel`, it would be like duplicating the code in Gem::Version that already does that.

I think that the code that just catches the exception instead of using conditions would be easier to understand:

```
@gem_version_with_patch ||= begin
  Gem::Version.create("#{@gem_version}.#{@patchlevel}")
rescue BadRequirementError
  Gem::Version.create(@gem_version)
end
```

Which one would you consider to be better?

segiddins:
```ruby
@gem_version_with_patch ||= begin
  Gem::Version.create("#{@gem_version}.#{@patchlevel}")
rescue BadRequirementError
  @gem_version
end
```

spec/install/gemspecs_spec.rb
chalkos: I looked for a way to set `Gem::Specification#required_ruby_version` with `install_gemfile`, but I could not find one. Is it possible?

indirect: Setting the required ruby version here is correct, but you can use `install_gemfile` to create and install the Gemfile.

indirect: I would suggest something like this:

```ruby
build_lib("foo", :path => bundled_app) do |s|
  s.required_ruby_version = RUBY_VERSION
end

install_gemfile <<-G
  ruby '#{RUBY_VERSION}', :engine_version => '#{RUBY_VERSION}', :engine => 'ruby'
  gemspec
G
should_be_installed "foo 1.0"
```

segiddins:
@bundlerbot r+

indirect:
woo bundlerbot! :tada:

segiddins:
@chalkos see https://travis-ci.org/bundler/bundler/jobs/136049835#L3428-L3446

chalkos:
I'll rebase it to reproduce this on my end.

chalkos:
@segiddins turns out `ruby 2.4.0dev` has a `RUBY_PATCHLEVEL` of -1
For these versions you can't specify a patchlevel in the gemspec (because it would be -1 and bundler will complain about that). I suggest that affected tests are skipped in such cases.

segiddins:
@chalkos agreed

chalkos:
I rebased on master and added code that should skip the spec on ruby versions with negative patchlevel.
@segiddins is this supposed to fail? https://travis-ci.org/bundler/bundler/jobs/136242613

segiddins:
@homu r+

PR #4628, opened 2016-05-31
[Inline] Install when necessary

segiddins:
The RubyGems change just means setting remote! will always cause specs to be fetched remotely

\c @indirect


indirect:
yesssssssssssssssssssssssssssssssssssss

@homu r+

PR #4632, opened 2016-06-01
Add `bundle add` command to add gems into the Gemfile

### LINKED ISSUE #4546: Add an "add" command to add gems to the Gemfile
_From @ryanstout on October 31, 2013 22:48_

Similar to https://github.com/DruRly/gemrat and NPM, it would be really useful to be able to say

    bundle add nokogiri

and have the gem added to the end of the Gemfile, then bundled.  With a lot of projects, it seems like a very common workflow to add gems without version numbers.  (I see bundler has inject is included, but it requires a version number.  Also, add seems like a better name)

I can implement this as a pull request, I wanted to discuss it before I went to build it however.  (As suggested in the development guide)

_Copied from original issue: bundler/bundler-features#32_

b-ggs:
Continuing the work done to implement `bundle add GEM` from bundler/bundler#4546

To do:
- [x] Cherry-pick commit from @ognevsky 
- [x] Run specs
- [x] Use Bundler.definition to resolve `last_version_number`
- [x] Pattern `bundle add` implementation after `bundle inject`
- [x] Add option to specify added gem's `group`
- [x] Add option to specify added gem's `source`
- [x] Modify existing specs & add additional specs

@RochesterinNYC

---

Closes #4546


segiddins:
Is there any reason this couldn't just be a plugin?

indirect:
I actually want this on a regular basis, and have planned on adding it (when I get around to it 😝) for years... this is also a standard-ish feature in RubyGems, NPM, etc. I feel like it's got a pretty good case for being in core.

lib/bundler/adder.rb
```diff
@@ -0,0 +1,36 @@
...
+require "yaml"
+
+module Bundler
+  class Adder
+    def initialize(name, version)
+      @name = name
+      @version = version || last_version_number
+    end
+
+    def add
```
RochesterinNYC: I think all the current Bundler command classes follow an `initialize` + `run` pattern that's invoked like so: https://github.com/bundler/bundler/blob/master/lib/bundler/cli.rb#L175,  https://github.com/bundler/bundler/blob/master/lib/bundler/cli.rb#L204, etc.

lib/bundler/adder.rb
```diff
@@ -0,0 +1,36 @@
...
+        f.puts
+        f.puts output_line
+      end
+      Bundler.ui.confirm "Added to Gemfile: #{output_line}"
+    end
+
+  private
+
+    def last_version_number
+      uri = URI.parse("https://rubygems.org/api/v1/gems/#{@name}.yaml")
```
RochesterinNYC: Good for a baseline, but I believe there's existing Bundler code to handle these kinds of calls out to the API. (ex. how does `bundle outdated` do it?)

chrismo:
Should a source argument be supported? And if the Gemfile is organized in source blocks ... I guess this command wouldn't be able to add the line within an existing source block very reliably (esp. in the case with a mix of blocks and same-line and no specified source).

lib/bundler/adder.rb
chrismo: `outdated` does a resolution of the Definition and gets results via an Index I'm pretty sure, so it's relying on all that infrastructure to get through to fetchers -- maybe there'd be a way to bypass most of that and get to the Rubygems fetcher itself?

RochesterinNYC:
@chrismo, that's a fair point. For the baseline implementation, I might just expect a source argument to add the gem with the proper source explicitly at the end. I'm not sure the current Bundler classes have the ability to inject a gem into a source block for a Gemfile. Are there any use cases where the injection of the added gem into the source block is necessary as opposed to being a nice-to-have?

b-ggs:
@RochesterinNYC Made a few changes. Can you review this? :smile:

spec/commands/add_spec.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+require "spec_helper"
+
+describe "bundle add" do
+  before :each do
+    gemfile
+  end
+
+  context "when version number is set" do
```
RochesterinNYC: can you add a spec and feature coverage for `bundle add foo "<= 1.2.3"` (specifying a version operator in the constraint passed).

RochesterinNYC:
hey @b-ggs, some things that @indirect and I discussed:

There is a `bundle inject` command. `bundle add` will be like a more user-friendly, advanced version of `bundle inject`. You've already covered the two main UX differences:
- version is not required to be passed to `bundle add`
- the default version constraint is `~> <latest_version>`
However, it'd be nice if the `bundle add` and `bundle inject` commands could use the same underlying class (maybe the Injector https://github.com/bundler/bundler/blob/master/lib/bundler/injector.rb) as the overall functionality is very similar. Does that make sense?

lib/bundler/cli.rb
```diff
@@ -437,6 +437,12 @@ def env
       Env.new.write($stdout)
     end
 
+    desc "add GEM [VERSION]", "Add the named gem to the bottom of Gemfile"
```
RochesterinNYC: named --> specified

lib/bundler/cli/add.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+require "bundler/cli/common"
+require "net/http"
+require "yaml"
+
+module Bundler
+  class CLI::Add
```
RochesterinNYC: I'll hold off on more extensive review of this until `bundle add` and `bundle inject` are sharing similar functionality/code and internal interfaces.

b-ggs:
@RochesterinNYC Still need to add specs, but how's this?

b-ggs:
@RochesterinNYC Code Climate seems to be complaining that this is too similar to the code in `bundle inject`. Maybe I should just call `Inject.new([...]).run` from within `add.rb`? Do you have any suggestions for this?

RochesterinNYC:
A `--pre` flag sounds compelling.

Yup, I don't think the Add class needs to exist. Rather, inject with certain flags/options passed in could serve for both commands.

b-ggs:
@RochesterinNYC Removed the `Add` class, and moved everything into the `Inject` class. I also implemented the `--pre` flag for adding prerelease gems. Also, I added the appropriate specs. Could you check it out and let me know if I've missed anything? Thanks!

RochesterinNYC:
Great, what does the output for this look like? Additionally, does this print out the timestamp line in the Gemfile like `bundle inject` will? We'd want to avoid printing that line for `bundle add`.

b-ggs:
@RochesterinNYC It originally also appened the timestamp. To remove the timestamp, I had to edit the `Injector` class to add an option to omit the timestamp from the lines appended to the Gemfile.

Here's the output for this:
```
╭─boggs@archbook ~/dev/4632-bundle-add-test ‹ruby-2.3.0› ‹node-v4.4.1›
╰─➤  cat Gemfile
source 'https://rubygems.org'
╭─boggs@archbook ~/dev/4632-bundle-add-test ‹ruby-2.3.0› ‹node-v4.4.1›
╰─➤  dbundle add rack 1.5.0
Added to Gemfile:
  rack (= 1.5.0)
╭─boggs@archbook ~/dev/4632-bundle-add-test ‹ruby-2.3.0› ‹node-v4.4.1›
╰─➤  cat Gemfile
source 'https://rubygems.org'

gem 'rack', '= 1.5.0'
```

b-ggs:
I also made sure that it didn't affect how `bundle inject` appended lines to the Gemfile.

```
╭─boggs@archbook ~/dev/4632-bundle-add-test ‹ruby-2.3.0› ‹node-v4.4.1›
╰─➤  dbundle inject rack 1.5.0
Added to Gemfile:
  rack (= 1.5.0)
╭─boggs@archbook ~/dev/4632-bundle-add-test ‹ruby-2.3.0› ‹node-v4.4.1›
╰─➤  cat Gemfile
source 'https://rubygems.org'

# Added at 2016-06-16 12:36:59 +0800 by boggs:
gem 'rack', '= 1.5.0'
```

b-ggs:
@RochesterinNYC For some reason, if I don't include a version number, it seems to fetch gem metadata twice. I'll look into this.

```
╭─boggs@archbook ~/dev/4632-bundle-add-test ‹ruby-2.3.0› ‹node-v4.4.1›
╰─➤  cat Gemfile                                                                                                                    1 ↵
source 'https://rubygems.org'
╭─boggs@archbook ~/dev/4632-bundle-add-test ‹ruby-2.3.0› ‹node-v4.4.1›
╰─➤  dbundle add rack
Resolving dependencies...
Resolving dependencies...
Added to Gemfile:
  rack (= 1.6.4)
╭─boggs@archbook ~/dev/4632-bundle-add-test ‹ruby-2.3.0› ‹node-v4.4.1›
╰─➤  cat Gemfile
source 'https://rubygems.org'

gem 'rack', '= 1.6.4'
```

But, so far, do you have any comments regarding the implementation?

lib/bundler/cli/inject.rb
```diff
@@ -1,12 +1,14 @@
 # frozen_string_literal: true
+require "bundler/cli/common"
```
RochesterinNYC: What was this added for?

lib/bundler/cli/inject.rb
```diff
@@ -1,12 +1,14 @@
...
 module Bundler
   class CLI::Inject
-    attr_reader :options, :name, :version, :gems
-    def initialize(options, name, version, gems)
+    attr_reader :options, :name, :version, :gems, :command
+    def initialize(options, name, version, gems, command = "inject")
```
RochesterinNYC: Why can't we pass the intention to have timestamps printed as part of the options parameter instead of specifying which command is being invoked? It seems the only reason that command is being specified as `inject` or `add` is to determine whether add :remove_timestamp to the options hash or not.

lib/bundler/injector.rb
```diff
@@ -55,7 +56,9 @@ def new_gem_lines
     def append_to(gemfile_path)
       gemfile_path.open("a") do |f|
         f.puts
-        f.puts "# Added at #{Time.now} by #{`whoami`.chomp}:"
+        unless !!@options[:remove_timestamp]
```
RochesterinNYC: Why the double negation?

spec/commands/add_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+      bundle "add activesupport '> 2.3.5'"
+      expect(bundled_app("Gemfile").read).to match(/gem 'activesupport', '> 2.3.5'/)
+    end
+  end
+
+  context "when version number is not set" do
+    it "adds gem with last stable version" do
+      bundle "add activesupport"
+      expect(bundled_app("Gemfile").read).to match(/gem 'activesupport', '= 2.3.5'/)
```
RochesterinNYC: I thought we decided that the gem would be added with `~>` if no version was specified?

RochesterinNYC: @indirect I know we discussed making `bundle add` act almost the same as `bundle inject`, but if we're really going to push `bundle add` as a `bundle inject` with far improved developer UX, I think we should have `bundle add` add new gems to the Gemfile with `~>` constraint on the latest version if no version is passed into the command.

I.e.

`bundle add test`

The latest non-beta version of `test` is `v1.0.2`.
The latest pre-release version of `test` is `v1.0.3.rc1`.

`Gemfile`:
```
gem 'test', '~> 1.0.2'
```

`bundle add test --pre`

Gemfile:
```
gem 'test', '~> 1.0.3rc1'
```

RochesterinNYC:
Added some comments, I think `source` and `group` flags are also very useful for `bundle add` for specifying a source or group on the new `gem` line in the `Gemfile`.

lib/bundler/cli/inject.rb
b-ggs: Ah, I was previously using `select_spec` from `CLI::Common`. Will remove this line as this isn't needed anymore.

RochesterinNYC:
How is this progressing?

spec/commands/add_spec.rb
RochesterinNYC: Let's go ahead and use the twiddle-waka (`~>`) operator as the default instead of `=`.

b-ggs:
@RochesterinNYC I've yet to add specs for `--groups` and `--source`, but it's pretty much working now.

```
╭─boggs@archbook ~/dev/4632-bundle-add-test ‹ruby-2.3.0› ‹node-v4.4.1›
╰─➤  touch Gemfile
╭─boggs@archbook ~/dev/4632-bundle-add-test ‹ruby-2.3.0› ‹node-v4.4.1›
╰─➤  dbundle add thin 1.0 --groups foobar barbaz --source file:///home/boggs/dev/bundler/tmp/gems/remote1
Fetching source index from file:/home/boggs/dev/bundler/tmp/gems/remote1/
Resolving dependencies...
Added to Gemfile:
  thin (~> 1.0)
╭─boggs@archbook ~/dev/4632-bundle-add-test ‹ruby-2.3.0› ‹node-v4.4.1›
╰─➤  cat Gemfile

gem 'thin', '~> 1.0', :group => [:foobar, :barbaz], :source => 'file:///home/boggs/dev/bundler/tmp/gems/remote1'
```

b-ggs:
@RochesterinNYC I've added the appropriate specs. I cover too much on issues such as gem duplicates, etc. since `inject_spec` pretty much covers those kinds of issues.

As soon as the specs pass, I think this is good to go. Do you have any feedback or is there something you'd like me to add?

b-ggs:
@RochesterinNYC This is odd - Travis seems to have stopped running specs for this PR.

RochesterinNYC:
@b-ggs that is odd, try to force push the PR to update it and see if that triggers a new build?

segiddins:
@b-ggs travis is running right now....

b-ggs:
@segiddins @RochesterinNYC Yeah, it just showed up on my end. A few minutes ago it had only listed codeclimate as one of the checks, and was telling me that it had passed all 1 of 1 checks.

RochesterinNYC:
Great, I'll take another look at this and review later tonight.

spec/commands/add_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  context "when version number is not set" do
+    it "adds gem with last stable version" do
+      bundle "add activesupport"
+      expect(bundled_app("Gemfile").read).to include("gem 'activesupport', '~> 2.3.5'")
+    end
+
+    it "adds the gem with the last prerelease version" do
+      update_repo2 do
+        build_gem "activesupport", "3.0.0.beta"
```
RochesterinNYC: Can we also build `activesupport` gem version `3.0.0` for this spec?

lib/bundler/cli.rb
```diff
@@ -444,6 +444,16 @@ def env
       Env.new.write($stdout)
     end
 
+    desc "add GEM [VERSION]", "Add the specified gem to the bottom of Gemfile"
+    method_option "group", :type => :array, :desc => "Specify groups to add the gem in"
+    method_option "source", :type => :string, :desc => "Specify the gem's source"
+    method_option "pre", :type => :boolean, :default => false, :desc => "Check for newer pre-release gems"
+    method_option "append-timestamp", :type => :boolean, :default => false, :desc => "Append timestamp to Gemfile"
```
RochesterinNYC: I think we can just make the option `--timestamp` instead of `--append-timestamp`.

lib/bundler/cli.rb
```diff
@@ -444,6 +444,16 @@ def env
       Env.new.write($stdout)
     end
 
+    desc "add GEM [VERSION]", "Add the specified gem to the bottom of Gemfile"
+    method_option "group", :type => :array, :desc => "Specify groups to add the gem in"
```
RochesterinNYC: Let's also add aliases (ex. `-g` for `--group`) for all the options.

spec/commands/add_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  context "when group is set" do
+    it "adds the gem with the specified groups" do
+      bundle "add activesupport --group development test"
+      expect(bundled_app("Gemfile").read).to include("gem 'activesupport', '~> 2.3.5', :group => [:development, :test]")
+    end
+  end
+
+  context "when source is set" do
```
RochesterinNYC: Let's also have a spec or two for multiple flag options (ex. specifying `bundle add --group development test --source https://proxy.org`)

RochesterinNYC:
Added a few more comments/small feedback points. This looks really good and about ready :+1:

b-ggs:
@segiddins Oops, slight misunderstanding of the `--pre` flag. `gem install cocoapods --pre` gives the latest stable despite it not being a prerelease, since it's the latest available.

I've reverted back to how I was originally checking for prereleases, and have added the appropriate specs.

RochesterinNYC:
@segiddins @indirect can I have homu access on this PR?

segiddins:
@homu delegate=RochesterinNYC

RochesterinNYC:
@homu r+

PR #4640, opened 2016-06-02
[Lock] Add --add-platform

segiddins:
\c @indirect @RochesterinNYC


indirect:
@homu r+

lib/bundler/definition.rb
```diff
@@ -433,6 +432,11 @@ def validate_ruby!
       end
     end
 
+    def add_platform(platform)
+      @new_platform ||= @platforms.include?(platform)
```
segiddins: !

indirect: yeeeah that doesn't look right 😬

segiddins:
Fixed!

indirect:
@homu r+

PR #4643, opened 2016-06-05
[bundle] Automatically trampoline to postit

segiddins:
- [x] Specs

~~Except on bundle exec, since that would be too slow~~

~~If you want, I guess we could vendor postit and manually setup the `LOAD_PATH` and `-r` the full path to the vendored postit exe. Your call.~~ Done this

\c @indirect

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -199,6 +199,8 @@ module Bundler
       "Force downloading every gem."
     method_option "ruby", :type => :boolean, :banner =>
       "Update ruby specified in Gemfile.lock"
+    method_option "bundler", :type => :string, :lazy_default => "> 0.a", :banner =>
+      "Update the locked version of bundler"
     def update(*gems)
       require "bundler/cli/update"
       Update.new(options, gems).run
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -169,6 +169,7 @@ module Bundler
 
       # Set BUNDLE_GEMFILE
       ENV["BUNDLE_GEMFILE"] = find_gemfile.to_s
+      ENV["BUNDLER_VERSION"] = Bundler::VERSION
     end
 
     def set_path
--- a/spec/other/trampoline_spec.rb
+++ b/spec/other/trampoline_spec.rb
@@ -82,10 +82,10 @@ describe "bundler version trampolining" do
     it "fails gracefully when installing the bundler fails" do
       ENV["BUNDLER_VERSION"] = "9999"
       bundle "--version", :expect_err => true
-      expect(err).to eq(<<-E.strip)
+      expect(err).to start_with(<<-E.strip)
 Installing the inferred bundler version (= 9999) failed.
 If you'd like to update to the current bundler version (1.12.5) in this project, run `bundle update --bundler`.
-The error was: Unable to resolve dependency: user requested 'bundler (= 9999)'
+The error was:
       E
     end
   end
--- a/spec/other/trampoline_spec.rb
+++ b/spec/other/trampoline_spec.rb
@@ -61,6 +61,15 @@ describe "bundler version trampolining" do
   context "installing missing bundler versions", :realworld => true do
     before do
       ENV["BUNDLER_VERSION"] = "1.12.3"
+      if Bundler::RubygemsIntegration.provides?("< 2.6.4")
+        # necessary since we intall with 2.6.4 but the specs can run against
+        # older versions that match againt the "gem" invocation
+        %w(bundle bundler).each do |exe|
+          system_gem_path.join("bin", exe).open("a") do |f|
+            f << %(\ngem "bundler", ">= 0.a"\n)
+          end
+        end
+      end
     end
 
     it "guesses & installs the correct bundler version" do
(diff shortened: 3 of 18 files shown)


indirect:
I thought we already trampolined to the version in the lockfile? (I'm thinking maybe we should merge postit in and treat it as an internal library rather than a vendored external library?)

segiddins:
@indirect postit is what figures out what version is in the lockfile

segiddins:
This has now been completely rewritten, and only really needs extensive testing

\c @RochesterinNYC

RochesterinNYC:
I can take a look at adding specs.

segiddins:
no no you don't have to, I can write specs :P
just saying you might want to review it

lib/bundler/cli.rb
```diff
@@ -199,6 +199,8 @@ def install
       "Force downloading every gem."
     method_option "ruby", :type => :boolean, :banner =>
       "Update ruby specified in Gemfile.lock"
+    method_option "bundler", :type => :string, :lazy_default => "> 0.a", :banner =>
```
indirect: Not sure about defaulting people into a prerelease without them asking for it. Do you feel like that's okay to do?

indirect:
Just to make sure I'm understanding this... the explicit `bundle update --bundler` flag is so that you can short-circuit the trampoline? Some usage docs might be good here 😅

I was initially expecting that it would be possible to manage Bundler's version using the Gemfile and `gem "bundler"`, but of course as I'm typing this I realize that would break backwards compatibility :sob:.

The code looks much less horrible than I was expecting, so great job there! 😁

lib/bundler/cli.rb
segiddins: It'll only ever pin them to a pre-release if that's the running version of bundler (i.e. if they've installed a pre-release). Remember that RubyGems' default is to run rereleases for binaries

indirect: Oh yeah! Good point.

segiddins:
> Just to make sure I'm understanding this... the explicit bundle update --bundler flag is so that you can short-circuit the trampoline? Some usage docs might be good here 😅

It something that's easy to check for, and is a simple way to force `Definition#to_lock` to include the new bundler version :P

indirect:
@homu r+

segiddins:
@indirect dear God please don't merge this without tests

indirect:
lol whooooops

lib/bundler/shared_helpers.rb
```diff
@@ -169,6 +169,7 @@ def set_bundle_variables
 
       # Set BUNDLE_GEMFILE
       ENV["BUNDLE_GEMFILE"] = find_gemfile.to_s
+      ENV["BUNDLER_VERSION"] = Bundler::VERSION
```
RochesterinNYC: Should this also be an `if defined?(Bundler::VERSION)` kind of case?

lib/bundler/version.rb
```diff
@@ -7,5 +7,5 @@ module Bundler
   # We're doing this because we might write tests that deal
   # with other versions of bundler and we are unsure how to
   # handle this better.
-  VERSION = "1.12.5" unless defined?(::Bundler::VERSION)
+  VERSION = "1.12.9" unless defined?(::Bundler::VERSION)
```
RochesterinNYC: What's the reason for this version bump?

segiddins: 😁testing. it shouldn't be there

lib/bundler/shared_helpers.rb
segiddins: nah this is inside the bundler module

segiddins:
Basic specs added, what else would you like to see?

spec/other/trampoline_spec.rb
```diff
@@ -0,0 +1,118 @@
+# frozen_string_literal: true
+require "spec_helper"
+
+describe "bundler version trampolining" do
+  before do
+    # ENV["BUNDLE_DISABLE_POSTIT"] = nil
```
indirect: is this for debugging?

indirect:
I'd like to see a test around the error message if postit's automatic installation fails, but otherwise this is looking pretty good!

spec/other/trampoline_spec.rb
segiddins: 😭 yes. I'm tired

spec/other/trampoline_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+    before do
+      simulate_bundler_version("1.11.1") do
+        install_gemfile ""
+      end
+    end
+
+    it "updates to the specified version" do
+      # HACK: since no released bundler version actually supports this feature!
+      bundle "update --bundler=1.12.0", :expect_err => true
+      expect(out).to include("Unknown switches '--bundler=1.12.0'")
```
indirect: Shouldn't `bundle update --bundler=1.12.0` at least lock to Bundler 1.12.0, and then automatically install it and trampoline to it on the next execution?

segiddins:
@indirect done

spec/other/trampoline_spec.rb
segiddins: yes, but I couldn't figure out how to do that without rewriting how all of definition locking works and I'm lazy and it doesn't seem worth it

indirect: Does this test exist to verify that specified versions are not supported? I guess I'm confused by the spec string of "it updates to the specified version" combined with the expectation of "Unknown switch".

segiddins: I'd update it as soon as we have a released version that actually accepts the `--bundle` option

segiddins:
@indirect this should now be green 🎆

indirect:
:tada:

@homu r+

indirect:
@homu r+ one more time

PR #4645, opened 2016-06-06
[Installer] Fail fast when required_ruby(gems)?_version doesn't match

segiddins:
Let's finally put these to use!

### DIFF
--- a/lib/bundler/installer.rb
+++ b/lib/bundler/installer.rb
@@ -68,6 +68,7 @@ module Bundler
       end
 
       resolve_if_need(options)
+      ensure_specs_are_compatible!
       install(options)
 
       lock unless Bundler.settings[:frozen]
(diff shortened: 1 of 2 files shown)


segiddins:
@indirect @RochesterinNYC look good?

lib/bundler/installer.rb
```diff
@@ -156,6 +157,24 @@ def install(options)
       install_in_parallel jobs, options[:standalone], force
     end
 
+    def ensure_specs_are_compatible!
```
RochesterinNYC: I wonder if this would be useful to generalize interface wise and allow it to take in a specific ruby version and rubygems version as arguments. `ensure_specs_are_compatible(ruby_version, rubygems_version)`

RochesterinNYC:
This LGTM :+1:

segiddins:
@homu delegate=RochesterinNYC

RochesterinNYC:
@homu r+

indirect:
I think we need to show errors for this kind of issue in a different place... I took this PR and fixed a bug and added a test for what I mean at #4650.

indirect:
I'm not sure why homu failed, but I feel like this is an improvement over the current state of things, so I'm r+

@homu retry

indirect:
dammit homu :(

PR #4650, opened 2016-06-07
Respect required ruby versions

indirect:
This is an extension of #4645, fixing the version check to actually work, testing that the correct versions get installed as a result, and showing the kind of error message that would actually be useful (as opposed to "couldn't find that gem").

### DIFF
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -125,7 +125,13 @@ module Bundler
       def for?(platform, ruby_version)
         spec = @specs[platform]
         return false unless spec
-        return true if ruby_version.nil? || !spec.is_a?(EndpointSpecification)
+
+        return true if ruby_version.nil?
+        # Only allow endpoint specifications since they won't hit the network to
+        # fetch the full gemspec when calling required_ruby_version
+        return true if !spec.is_a?(EndpointSpecification) && !spec.is_a?(Gem::Specification)
+        return true if spec.required_ruby_version.nil?
+
         spec.required_ruby_version.satisfied_by?(ruby_version.to_gem_version_with_patchlevel)
       end
 
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -122,14 +122,11 @@ module Bundler
         @source ||= first.source
       end
 
-      def for?(platform, required_ruby_version)
-        if spec = @specs[platform]
-          if required_ruby_version && spec.respond_to?(:required_ruby_version) && spec_required_ruby_version = spec.required_ruby_version
-            spec_required_ruby_version.satisfied_by?(required_ruby_version.to_gem_version_with_patchlevel)
-          else
-            true
-          end
-        end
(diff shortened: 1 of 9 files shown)


lib/bundler/resolver.rb
```diff
@@ -129,13 +129,10 @@ def source
...
-          if required_ruby_version && spec.respond_to?(:required_ruby_version) && spec_required_ruby_version = spec.required_ruby_version
-            spec_required_ruby_version.satisfied_by?(required_ruby_version.gem_version)
-          else
-            true
-          end
-        end
+        return true unless required_ruby_version && Bundler.rubygems.has_required_ruby_version?
+
+        spec = @specs[platform]
+        spec.nil? ? true : spec.required_ruby_version.satisfied_by?(required_ruby_version.gem_version)
```
segiddins: This should return `false` when the spec is nil

lib/bundler/rubygems_integration.rb
```diff
@@ -169,6 +169,12 @@ def preserve_paths
       yield
     end
 
+    def has_required_ruby_version?
+      return @has_required_ruby_version if @calculated_has_required_ruby_version
```
segiddins: Can use `defined?` so we only need one ivar

lib/bundler/rubygems_integration.rb
```diff
@@ -169,6 +169,12 @@ def preserve_paths
       yield
     end
 
+    def has_required_ruby_version?
```
segiddins: this seems to be necessary -- even rubygems 1.3.4 has `required_rubygems_version`

indirect:
rebased on master, removed `respond_to?` check, fixed to return false on spec missing

segiddins:
@homu r+

segiddins:
This breaks a lot of things, somehow

indirect:
Ugh, apparently we still aren't correctly respecting required ruby versions. Starting to feel like 1.13 is going to be "1.12 except the features Andre was involved with actually work this time". :/

spec/support/builders.rb
```diff
@@ -368,6 +368,9 @@ def build_repo(path, &blk)
     end
 
     def update_repo(path)
+      if path == gem_repo1 && caller_locations.first.label != "build_repo"
+        raise "Updating gem_repo1 is unsupported -- use gem_repo2 instead"
```
indirect: How about "It's not possible to update gem_repo1 because that would cause side effects in every other test. Use gem_repo2 instead, which is rebuilt (if needed) for each test."

indirect:
looks great, I hope this works!

indirect:
@segiddins any interest in polishing off the last two failures here?

segiddins:
Oof I didn't see those, those might be hard to fix but I'll try!

segiddins:
failures caused by gem_repo2 having an extra version of rack. this should actually pass this time!

indirect:
hooray @homu r+

PR #4651, opened 2016-06-07
Resolve respecting Ruby versions without `ruby` in the Gemfile

indirect:
This is an extension of #4650 that adds the currently running Ruby as the Ruby that is being resolved against, making it possible to install (and lock to) usable versions of gems even in Gemfiles that don't explicitly declare a Ruby version.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -52,7 +52,7 @@ module Bundler
       @optional_groups = optional_groups
       @remote          = false
       @specs           = nil
-      @ruby_version    = ruby_version
+      @ruby_version    = ruby_version || Bundler::RubyVersion.system
 
       @lockfile_contents      = String.new
       @locked_bundler_version = nil
--- a/lib/bundler/installer.rb
+++ b/lib/bundler/installer.rb
@@ -68,6 +68,7 @@ module Bundler
       end
 
       resolve_if_need(options)
+      ensure_specs_are_compatible!
       install(options)
 
       lock unless Bundler.settings[:frozen]
@@ -154,6 +155,24 @@ module Bundler
       jobs = 1
       jobs = [Bundler.settings[:jobs].to_i - 1, 1].max if can_install_in_parallel?
       install_in_parallel jobs, options[:standalone], force
+    end
+
+    def ensure_specs_are_compatible!
+      system_ruby = Bundler::RubyVersion.system
+      rubygems_version = Gem::Version.create(Gem::VERSION)
+      specs.each do |spec|
+        if required_ruby_version = spec.required_ruby_version
+          unless required_ruby_version.satisfied_by?(system_ruby.gem_version)
+            raise InstallError, "#{spec.full_name} requires ruby version #{required_ruby_version}, " \
+              "which is incompatible with the current version, #{system_ruby}"
+          end
+        end
+        next unless required_rubygems_version = spec.required_rubygems_version
+        unless required_rubygems_version.satisfied_by?(rubygems_version)
+          raise InstallError, "#{spec.full_name} requires rubygems version #{required_rubygems_version}, " \
+            "which is incompatible with the current version, #{rubygems_version}"
+        end
+      end
     end
 
     def can_install_in_parallel?
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -129,13 +129,10 @@ module Bundler
       end
 
       def for?(platform, required_ruby_version)
-        if spec = @specs[platform]
-          if required_ruby_version && spec.respond_to?(:required_ruby_version) && spec_required_ruby_version = spec.required_ruby_version
-            spec_required_ruby_version.satisfied_by?(required_ruby_version.gem_version)
-          else
-            true
-          end
-        end
+        return true unless required_ruby_version && Bundler.rubygems.has_required_ruby_version?
+
+        spec = @specs[platform]
+        spec.nil? ? true : spec.required_ruby_version.satisfied_by?(required_ruby_version.gem_version)
       end
 
       def to_s
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -169,6 +169,12 @@ module Bundler
       yield
     end
 
+    def has_required_ruby_version?
+      return @has_required_ruby_version if @calculated_has_required_ruby_version
+      @calculated_has_required_ruby_version = true
+      @has_required_ruby_version = Gem::Specification.new.respond_to?(:required_ruby_version)
+    end
+
     def loaded_gem_paths
       # RubyGems 2.2+ can put binary extension into dedicated folders,
       # therefore use RubyGems facilities to obtain their load paths.
--- a/spec/install/gems/resolving_spec.rb
+++ b/spec/install/gems/resolving_spec.rb
@@ -105,20 +105,37 @@ describe "bundle install with gem sources" do
 
     describe "when some gems require a different version of ruby" do
       it "does not try to install those gems" do
-        pending "waiting for a rubygems index that includes ruby version"
-
         update_repo gem_repo1 do
           build_gem "require_ruby" do |s|
             s.required_ruby_version = "> 9000"
           end
         end
 
-        install_gemfile <<-G
+        install_gemfile <<-G, :artifice => "compact_index"
           source "file://#{gem_repo1}"
           gem 'require_ruby'
         G
 
         expect(out).to_not include("Gem::InstallError: require_ruby requires Ruby version > 9000")
+        expect(out).to include("require_ruby-1.0 requires ruby version > 9000, which is incompatible with the current version, #{Bundler::RubyVersion.system}")
+      end
+    end
+
+    describe "when some gems require a different version of rubygems" do
+      it "does not try to install those gems" do
+        update_repo gem_repo1 do
+          build_gem "require_rubygems" do |s|
+            s.required_rubygems_version = "> 9000"
+          end
+        end
+
+        install_gemfile <<-G, :artifice => "compact_index"
+          source "file://#{gem_repo1}"
+          gem 'require_rubygems'
+        G
+
+        expect(out).to_not include("Gem::InstallError: require_rubygems requires RubyGems version > 9000")
+        expect(out).to include("require_rubygems-1.0 requires rubygems version > 9000, which is incompatible with the current version, #{Gem::VERSION}")
       end
     end
   end
--- a/spec/install/gems/resolving_spec.rb
+++ b/spec/install/gems/resolving_spec.rb
@@ -102,9 +102,9 @@ describe "bundle install with install-time dependencies" do
     end
   end
 
-  describe "when a required ruby version" do
-    context "allows only an older version" do
-      it "installs the older version" do
+  describe "when an explicit ruby version" do
+    context "allows only an older gem version" do
+      it "installs the older gem version" do
         update_repo gem_repo1 do
           build_gem "rack", "9001.0.0" do |s|
             s.required_ruby_version = "> 9000"
@@ -122,7 +122,7 @@ describe "bundle install with install-time dependencies" do
       end
     end
 
-    context "allows no versions" do
+    context "allows no gem versions" do
       it "raises an error during resolution" do
         update_repo gem_repo1 do
           build_gem "require_ruby" do |s|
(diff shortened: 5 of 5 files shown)


segiddins:
👎🏻 for the same reasons I gave last year.

indirect:
@segiddins can you repeat your reasons? Sorry for the hassle, I just want to be really sure that I'm understanding your position.

Last year we didn't have the ability to lock to a Ruby version, and we didn't have a diagnostic error message. The big reason that I think we need to ship this, or something like this, is because users expect `gem install` and `bundle install` to respect the currently running gem version.

As far as I'm concerned, the `ruby` DSL method is a hack that we should get rid of, and handle Ruby versions like we're planning on handling Bundler versions with a locked version and an update flag. When combined with lockable Ruby versions and a resolver-style error message that clearly explains conflicts if they come up, I think this is the most-expected-least-surprising implementation for this feature.

Thoughts?

/cc @RochesterinNYC @TimMoore @andremedeiros

RochesterinNYC:
I'd like to see/read through the historical context on this if it's in an issue/PR or somewhere?

segiddins:
https://github.com/bundler/bundler/pull/4160

mvz:
@segiddins your objections in #4160 don't seem to take the use case of a gem that supports a range of ruby versions into account (as detailed in https://github.com/bundler/bundler/pull/4160#issuecomment-188674878). I currently handle this using conditionals in the Gemfile, which is not ideal. Any thoughts on that particular use case?

mvz:
@segiddins as a concrete case of my previous comment, my `memory_test_fix` gem supports Rails 4.1, 4.2 and 5.0. Builds are failing on Travis because Bundler won't install a compatible version of activesupport on older Rubies: https://travis-ci.org/mvz/memory_test_fix/jobs/142714833.

It seems to me the two use cases of Bundler are in conflict here:

* On the one hand, there is the 'app' use case, which requires `Gemfile.lock` to be checked into version control. This is the case where you want to be sure your installations will succeed in production, and this is the case you are defending.
* On the other hand, there is the 'gem' use case. This is the case where `Gemfile.lock` should not be checked in, and each developer or CI environment should be free to satisfy the dependencies in a way that matches the installed Ruby version and gems.

Bundler has always catered to both use cases, so I think the question is how can Bundler help the second use case without gem developers resorting to ad-hoc logic in their `Gemfile`s and without damaging the needs of the first use case?

segiddins:
@mvz in bundler 2, bundler will fall back to the ruby version requirements in the gemspec from the gemspec dsl call

mvz:
@segiddins Maybe I'm missing something, but I don't immediately see how that is going to help?

segiddins:
because it means bundler will resolve respecting the ruby versions your gem supports

mvz:
Ah, but that is not the problem. Instead, I need it resolve respecting the ruby versions other gems support together with the **currently running** Ruby version. In the concrete example given above, I need it to pick activerecord and friends < 5.0 on Ruby 2.1, and activerecord 5.0 on Ruby 2.2 and above.

I am currently experimenting with simply putting
```
ruby RUBY_VERSION
```
in the Gemfile. I think this does exactly what I want but it feels like a hack.

mvz:
> I think this does exactly what I want

No, it seems the current Bundler just ignores this :cry:.

segiddins:
ah but rubygems' own `gem install` doesnt do that, so installing from a clean ruby install would fail in that case :/

mvz:
> rubygems' own `gem install` doesnt do that

Indeed you are right :disappointed:. I guess I'll have to rethink what versions of Rails and Ruby my gem should support.

Anyway, @segiddins, thanks for your responses, they really helped clear things up for me.

indirect:
I _think_ we can address all of these concerns simultaneously by resolving against the currently running ruby version and then writing that ruby version into the lockfile, even if you don't have `ruby` in your Gemfile. It breaks backwards compatibility, so we would have to ship it as a config option until the next major release, but it seems to address both the app and gem use cases at the same time.

mvz:
@indirect there would still be the problem of `gem install` not taking the running ruby version into account when selecting dependencies.

indirect:
@mvz this is the Bundler issue tracker, we can fix `gem install` in a separate ticket on the RubyGems tracker. :)

ferventcoder:
I just read through #4160 - We are in a similar boat as https://github.com/bundler/bundler/pull/4160#issuecomment-188674878 where we test against multiple versions of Ruby to ensure compatibility.

Our use case is that we would expect different behavior based on the currently running Ruby version and we can't add the ruby version to the Gemfile as that would break support across multiple versions of Ruby (currently 1.9.x, 2.0.0, 2.1.x, etc). 

We are currently resolving to workarounds where we need to test against older versions of Ruby, and as more things come online with `ruby ~> 2.0`, we are going to be seeing this more and correcting it with a conditional as **each** gem gets updated (e.g. json_pure today on a .z release - https://github.com/puppetlabs/puppetlabs-dsc/pull/220/files#diff-8b7db4d5cc4b8f6dc8feb7030baa2478R48)

Would it be possible to take a second look at this? 

Similar to @indirect I find it surprising that Bundler doesn't natively support this option already. Since not everyone has the same use case and there is some consternation about this, allow it with a bundler configuration value?

indirect:
@ferventcoder this isn't abandoned, we are still working on it. In the meantime, you can likely handle your current needs with either `ruby File.read(".ruby-version")` or `ruby RUBY_VERSION` in your Gemfile.

ferventcoder:
> `ruby RUBY_VERSION`

Interesting. That does seem like a nice workaround (not sure how I missed that before).

ferventcoder:
> @ferventcoder this isn't abandoned, we are still working on it. In the meantime, you can likely handle your current needs with either ruby File.read(".ruby-version") or ruby RUBY_VERSION in your Gemfile.

Unfortunately this doesn't appear to be a valid solution.  `ruby RUBY_VERSION` doesn't do anything for limiting the resolved versions of other gems in the Gemfile.

Added `ruby RUBY_VERSION` as the first line in the Gemfile. Using Bundler 1.6.3.

~~~sh
$ bundle install
[..snip..]
Gem::InstallError: json_pure requires Ruby version ~> 2.0.
An error occurred while installing json_pure (2.0.2), and Bundler cannot continue.
Make sure that `gem install json_pure -v '2.0.2'` succeeds before bundling.
~~~

Keep in mind that json_pure comes in as a non-restricted dependency from a different Gem. I'm happy to raise another issue for this if this is a bug.

indirect:
@ferventcoder The first version of Bundler that is even able to respect required ruby versions is 1.12. Please test on the latest version before complaining about things that "don't work".

ferventcoder:
@indirect apologies. I thought I read in the documentation that it was [`v1.2` that respected the setting](http://bundler.io/v1.2/gemfile_ruby.html), but perhaps I read the documentation incorrectly.

indirect:
@ferventcoder The `ruby` Gemfile method was introduced in Bundler 1.2, but it was not until [the new index format](http://andre.arko.net/2014/03/28/the-new-rubygems-index-format/) (first supported in Bundler 1.12) that Bundler even knows whether a gem has a required ruby version or not. You will need to use Bundler 1.12+ at a minimum for Bundler to be able to respect `required_ruby_verison` in any gems.

ferventcoder:
@indirect I'll open a new ticket.

lamont-granquist:
this is causing fairly significant pain as gems like activesupport and rack drop support for ruby 2.1 and we're picking those gems up transitively.  we're having to add pins for the transitive deps.   we also want to drop support for 2.1 ourselves, but want users with existing working Gemfile's on ruby 2.1 to not be broken when we release gems which drop support.  telling everyone to run around and add `ruby "2.1.9"` to all their Gemfiles is less than pleasant.

and @segiddins the fallback to the gemspec isn't helpful because the workflow that people have is a typically application workflow where there is no gemspec and no published gem and no required_ruby_version.  mostly these are Gemfiles that wrap up an app and people are `bundle exec`'ing their tools.

lamont-granquist:
So, and I'd like to strenuously argue that the existing behavior is definitely wrong.

```ruby
 % cat Gemfile
source 'https://rubygems.org'

gem "activesupport"
% bundle --version
Bundler version 1.13.0.rc.1   # <--- master from yesterday
% bundle update
Resolving dependencies...
activesupport-5.0.0 requires ruby version >= 2.2.2, which is incompatible with the current version, ruby 2.1.10p492
```

Literally the only piece of information bundler has here about what ruby version I want is the ruby version that I'm running bundler from -- but it helpfully ignores that and attempts to create a bundle that can't install on the version of ruby that I'm running and it fails.

That looks to me like the definition of bad UX.

I would agree that explicit ruby lines in the Gemfile or a required_ruby_verison in the gemspec should always override automatically picking up the ruby version, and that maybe there should be a switch to turn this behavior off to deal with edge cases, but this is the common case, and bundler clearly gets it wrong and this affects users.

ferventcoder:
> telling everyone to run around and add ruby "2.1.9" to all their Gemfiles is less than pleasant.

And as you see from https://github.com/bundler/bundler/issues/4819, asking folks to add `ruby "2.1.9"` may not work until master gets released (looks like v1.13.0), which means there really isn't a pleasant workaround for this situation.

lamont-granquist:
Even on master, though, this behavior is fairly infuriating.   Its also not going to go away permanently, and this argument will recur predictably every time we go through another ruby deprecation cycle.  It'll quiet down again later in the year, but next year when ruby 2.2.x falls off support we'll be right back here again.

indirect:
@lamont-granquist you're pushing up pretty hard against the edge of the code of conduct right now. Feedback is helpful, telling us that you're infuriated by inexcusably bad UX is not okay and will result in a ban if you continue.

To reiterate: Since 1.0, it has been literally impossible (based on the information Bundler has from the server) to know that installing a gem would fail because of the current ruby version. That information was not available until the moment the gem was already downloaded for installation, and we failed as fast as we were able to.

Now that Bundler has ruby version information (a brand new development, after three years of work!), we can try to resolve this problem in a more user-friendly way. This PR is one option for solving that problem.

Open source means that you are empowered to contribute time or money to fixing problems. Contributing no money and no time, but angrily complaining that we haven't done enough free work for you yet? Not cool.

e2:
@indirect - I believe @lamont-granquist is just sharing his gut emotional response to suggest there are other users frustrated by having to deal with versioning issues instead of moving forward with getting more important work done.

Versioning is an advanced hardcore subject, so if users have to deal with it, it can be frustrating.

That's why you're such a hero to me, @indirect - these changes will be a huge boost to the Ruby community.

Thank you @lamont-granquist for your insights here. Just make sure than whenever you're describing your frustrations, you do it in a way where people don't see it as an attack. I know you just want to say how important these changes are (and how delighted you'll be once they're solved) - just make it clear that you're just describing your frustrations that you believe other users share. (And make it first clear that you aren't frustrated with anyone here and you appreciate the work already done here).

e2:
@segiddins - correct me if I'm wrong: I think the case you're arguing for is when the Gemfile.lock exists.

I believe a Ruby version should always be in a `Gemfile.lock`. It's like there's this idea in the Ruby community that a given gem should work on as many Ruby interpreter versions as possible.

Being a runtime language, Ruby makes this possible. But it's somehow become a requirement.

The biggest danger I see is: people having a crash during runtime due to running a bad Ruby and gem version combination. Gem library writers avoid specifying a `required_ruby_version` to avoid the flak they'd get from breaking CI builds (for example). That's not a good quality practice (failing CI builds are better than crashes in production).

Refusing to install on a Ruby that conflicts with a gem's required Ruby - that's huge progress already.

The biggest frustration: that users don't have a suggested way of dealing with an installation problem like this. They don't know where the fix should be, because they first assume that Bundler is working as designed. (Most people are oblivious to the work being done here right now).

I believe the current running Ruby is fundamental, because Ruby project workflows have usually no notion of "cross-compiling" (e.g. using Ruby 2.3.1 to install Ruby 2.1 dependencies).

I think Bundler should immediately abort if a `Gemfile.lock` was created with a different Ruby version. At least for now until there are "meta-Ruby" tools to do this. (How to define version Ruby constraints and test them to generate a project setup).

FWIW, I created https://github.com/e2/ruby_dep. One of the main drivers was to get people to upgrade Ruby, so that adding ruby_version_requirements would cause less havoc for ordinary users. In practices it breaks CI builds (the biggest reason people complain) and forces people to upgrade (another reason people complain). But in my opinion, those are minor issues compared to crashes in production.

I'd vote for easier Ruby upgrades instead of a smarter Bundler - any day.

I love you guys for moving this forward! 

Thanks @segiddins for tracking the edge cases - can't wait for Bundler 2.x!

And thanks @mzv for participating in this - the "app" vs " library" really helps understand and explain this better to everyone else.

And thanks @indirect for everything you do!

mvz:
With recent releases of Bundler (I suppose 1.13.x) I'm happy to report that putting `ruby RUBY_VERSION` in the Gemfile works very well and makes Bundler resolve respecting Ruby versions.

er0ck:
This seems to make a recent situation of ours a bit less painful so we don't have to pin tons of Gems (we have to test against ruby 1.8.7 for some legacy stuff).

It appears to breakdown for some Gems and i'm not clear exactly on why.  I'm assuming it's because some of our transitive deps don't specify `required_ruby_version` in some of their own releases.

e.g.
https://github.com/puppetlabs/puppetlabs-catalog_preview/pull/190

✋ @ferventcoder, we miss you!

PR #4654, opened 2016-06-07
Add resolver-style error messages for required Ruby conflicts

indirect:
*Proposal:* When a gem in the Gemfile requires a Ruby version different than the Gemfile's `ruby`, print a conflict error message that looks more or less exactly like any other resolution conflict error.

*Current behavior:* On `master` today, Bundler will ignore the gem's required_ruby_version until install-time, and raise an exception like "require_ruby-1.0 requires ruby version > 9000, which is incompatible with the current version`.

After #4650 has been merged, Bundler will correctly reject gems that require a different version of Ruby, and instead print the error "Could not find `required_ruby` in any of the sources in your Gemfile".

*Rationale for change:* I believe that error message is surprising and misleading. The source _does_ contain `required ruby`, and I think we'll get bug reports about how our error message is wrong.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -742,19 +742,10 @@ module Bundler
     # the metadata dependencies here
     def expanded_dependencies
       @expanded_dependencies ||= begin
-        ruby_versions = []
-        add_ruby_versions = proc do |ruby_version|
-          break unless ruby_version
-          if ruby_version.patchlevel
-            ruby_versions << ruby_version.to_gem_version_with_patchlevel
-          else
-            ruby_versions += ruby_version.versions.map {|v| "~> #{v}.0" }
-          end
-        end
-        add_ruby_versions.call(@ruby_version)
+        ruby_versions = concat_ruby_version_requirements(@ruby_version)
         if ruby_versions.empty? || !@ruby_version.exact?
-          add_ruby_versions.call(RubyVersion.system)
-          add_ruby_versions.call(locked_ruby_version_object) unless @unlock[:ruby]
+          concat_ruby_version_requirements(RubyVersion.system)
+          concat_ruby_version_requirements(locked_ruby_version_object) unless @unlock[:ruby]
         end
 
         metadata_dependencies = [
@@ -762,6 +753,22 @@ module Bundler
           Dependency.new("rubygems\0", Gem::VERSION),
         ]
         expand_dependencies(dependencies + metadata_dependencies, @remote)
+      end
+    end
+
+    def concat_ruby_version_requirements(ruby_version, ruby_versions = [])
+      return ruby_versions unless ruby_version
+      if ruby_version.patchlevel
+        ruby_versions << ruby_version.to_gem_version_with_patchlevel
+      else
+        ruby_versions.concat(ruby_version.versions.map do |version|
+          requirement = Gem::Requirement.new(version)
+          if requirement.exact?
+            "~> #{version}.0"
+          else
+            requirement
+          end
+        end)
       end
     end
 
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -246,7 +246,7 @@ module Bundler
         else
           # Run a resolve against the locally available gems
           Bundler.ui.debug("Found changes from the lockfile, re-resolving dependencies because #{change_reason}")
-          last_resolve.merge Resolver.resolve(expanded_dependencies, index, source_requirements, last_resolve, ruby_version, gem_version_promoter, additional_base_requirements_for_resolve)
+          last_resolve.merge Resolver.resolve(expanded_dependencies, index, source_requirements, last_resolve, gem_version_promoter, additional_base_requirements_for_resolve)
         end
       end
     end
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -343,8 +343,9 @@ module Bundler
       return unless @locked_ruby_version
       @locked_ruby_version_object ||= begin
         unless version = RubyVersion.from_string(@locked_ruby_version)
-          raise LockfileError, "Failed to create a `RubyVersion` object from " \
-            "`#{@locked_ruby_version}` found in #{@lockfile} -- try running `bundle update --ruby`."
+          raise LockfileError, "The Ruby version #{@locked_ruby_version} from " \
+            "#{@lockfile} could not be parsed. " \
+            "Try running bundle update --ruby to resolve this."
         end
         version
       end
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -62,6 +62,7 @@ module Bundler
       @specs           = nil
       @ruby_version    = ruby_version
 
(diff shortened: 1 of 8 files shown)


segiddins:
Depending on how high a priority this is, I can try and get to it by the start of next week?

segiddins:
```ruby
diff --git a/lib/bundler/dep_proxy.rb b/lib/bundler/dep_proxy.rb
index 998975b..05bcd38 100644
--- a/lib/bundler/dep_proxy.rb
+++ b/lib/bundler/dep_proxy.rb
@@ -13,6 +13,7 @@ module Bundler
     end
 
     def ==(other)
+      return dep == other if other.is_a?(Gem::Dependency)
       dep == other.dep && __platform == other.__platform
     end
 
index 918b3c9..d9bafca 100644
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -140,6 +140,10 @@ module Bundler
         @activated.map {|p| __dependencies[p] }.flatten
       end
 
+      def ruby_version_dependencies_for_activated_platforms
+        @activated.map {|p| Gem::Dependency.new(Gem::Platform::RUBY, @specs[p].required_ruby_version) }.compact
+      end
+
       def platforms_for_dependency_named(dependency)
         __dependencies.select {|_, deps| deps.map(&:name).include? dependency }.keys
       end
@@ -175,6 +179,7 @@ module Bundler
     def self.resolve(requirements, index, source_requirements = {}, base = [], ruby_version = nil)
       base = SpecSet.new(base) unless base.is_a?(SpecSet)
       resolver = new(index, source_requirements, base, ruby_version)
+      requirements += [Gem::Dependency.new("ruby", @ruby_version.gem_version)] if @ruby_version
       result = resolver.start(requirements)
       SpecSet.new(result)
     end
@@ -239,10 +244,21 @@ module Bundler
     include Molinillo::SpecificationProvider
 
     def dependencies_for(specification)
-      specification.dependencies_for_activated_platforms
+      deps = specification.dependencies_for_activated_platforms
+      deps.concat(specification.ruby_version_dependencies_for_activated_platforms)
+      deps
+    end
+
+    def ruby_specs_for(dependency)
+      return [] unless dependency.requirement.satisfied_by?(@ruby_version.gem_version)
+      [SpecGroup.new([Gem::Specification.new do |s|
+        s.name = dependency.name
+        s.version = @ruby_version.gem_version
+      end])]
     end
 
     def search_for(dependency)
+      return ruby_specs_for(dependency) if dependency.name == Gem::Platform::RUBY
       platform = dependency.__platform
       dependency = dependency.dep unless dependency.is_a? Gem::Dependency
       search = @search_for[dependency] ||= begin
```

indirect:
lol this isn't the start of next week! I like the idea, and would like to ship it in 1.13 if we can finish it in, say, less than 10-15 hours of work?

segiddins:
@indirect in your opinion, should we resolve for rubygems versions in addition to ruby versions?

indirect:
If we have it, I guess so? it seems much less critical, since (iirc) almost no gems in existence have required rubygems versions.

indirect:
So, having said that, let's punt on required rubygems version and wait to see if it's needed in real life before we do the work of implementing it, testing it, and resigning ourselves to supporting it. We have the data in the index if we need it.

indirect:
@segiddins rebase? I'd still like to ship this.

segiddins:
This requires more than just a rebase, it requires someone to actually implement it. So far, the branch is just your test case, and my very hacky patch is merely a comment :P

segiddins:
@indirect this is awful. please talk me out of this approach.

segiddins:
This might work now?

segiddins:
Hopefully green now

segiddins:
@indirect this is green on travis, please review?

lib/bundler/definition.rb
```diff
@@ -340,6 +339,17 @@ def locked_ruby_version
       end
     end
 
+    def locked_ruby_version_object
+      return unless @locked_ruby_version
+      @locked_ruby_version_object ||= begin
+        unless version = RubyVersion.from_string(@locked_ruby_version)
+          raise LockfileError, "Failed to create a `RubyVersion` object from " \
+            "`#{@locked_ruby_version}` found in #{@lockfile} -- try running `bundle update --ruby`."
```
indirect: I think the RubyVersion object information isn't helpful to end users reading this message—maybe something like 

> The Ruby version `#{@locked_ruby_version}` from #{@lockfile} could not be parsed. Try running `bundle update --ruby` to resolve this.

lib/bundler/resolver.rb
```diff
@@ -124,23 +124,17 @@ def source
...
-        return false unless spec
-
-        return true if ruby_version.nil?
-        # Only allow endpoint specifications since they won't hit the network to
-        # fetch the full gemspec when calling required_ruby_version
-        return true if !spec.is_a?(EndpointSpecification) && !spec.is_a?(Gem::Specification)
-        return true if spec.required_ruby_version.nil?
-
-        spec.required_ruby_version.satisfied_by?(ruby_version.to_gem_version_with_patchlevel)
+        !spec.nil?
```
indirect: looks like this entire method should probably be removed, and usages of `for?` should be replaced with `if @specs[platform]`?

lib/bundler/resolver.rb
```diff
@@ -163,6 +157,21 @@ def __dependencies
           dependencies
         end
       end
+
+      def metadata_dependencies(spec, platform)
+        return [] unless spec
+        # Only allow endpoint specifications since they won't hit the network to
+        # fetch the full gemspec when calling required_ruby_version
+        return [] if !spec.is_a?(EndpointSpecification) && !spec.is_a?(Gem::Specification)
```
indirect: seems like both guard clauses could be replaced by `return [] unless [EndpointSpecification, Gem::Specification].include?(spec.class)`, since that would also check for `NilClass`.

indirect:
I am super-impressed that you managed to make the existing infrastructure do so much of the work for this. :100: r+ once you feel the comments have been addressed.

lib/bundler/definition.rb
segiddins: 👍 looks way better. in reality this exception should never be hit and I think I copied it from above

lib/bundler/resolver.rb
segiddins: it's used in `search_for` so it needs to have a public interface. I can remove the second argument, however

lib/bundler/resolver.rb
segiddins: except that won't handle subclasses

segiddins:
@indirect r?

lib/bundler/resolver.rb
indirect: Sorry, I was kind of vague here.

Instead of this:

```ruby
def for(platform)
  spec = @specs[platform]
  !spec.nil?
end
```

How about this?

```ruby
def for?(platform)
  @specs[platform]
end
```

In the end I don't care enough to block on this, though.

indirect:
@homu r+

PR #4669, opened 2016-06-13
Invalid lockfile local should not block install or update.

drn:
Hey @indirect,

I finally got around to reproducing this issue ( https://github.com/bundler/bundler/pull/4172#issuecomment-185566638 ) in a minimal ruby project ( https://github.com/drn/bundler-locals-issue-project ). There are instructions in the README.

Basically both `bundle install` and `bundle update gem` raise the exception:
```
https://github.com/drn/bundler-locals-issue-gem (at TA-1000@d8a205d) is not yet checked
out. Run `bundle install` first.
```

I dug into the bundler source code and traced the exception to this line:
https://github.com/bundler/bundler/blob/master/lib/bundler/definition.rb#L483

The issue arises if there are no other dependencies in the project that are unlocking. All the other conditions are false, so source.specs != locked.specs triggers. source.specs returns a Bundler::Index instance. However, locked.specs raises an exception.

The exception that's raised is in https://github.com/bundler/bundler/blob/master/lib/bundler/source/git.rb#L192. The error message is:
The path/Users/darrencheng/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/bundler/gems/bundler-locals-issue-gem-d8a205d072a3does not exist., which in turn raises a GitError which isn't handled in the specs_changed? method.

The changes in this pull request patches the issue, but maybe there is a better fix for the issue? Potentially ensuring unlocking = true if a GitError is raised?

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -480,7 +480,16 @@ module Bundler
         end
       end
 
-      !locked || unlocking || dependencies_for_source_changed?(locked) || source.specs != locked.specs
+      return true unless locked
+      return true if unlocking
+      return true if dependencies_for_source_changed?(locked)
+      begin
+        locked_specs = locked.specs
+      rescue GitError
+        locked_specs = Bundler::Index.new
+      end
+      return true if source.specs != locked_specs
+      false
     end
 
     def dependencies_for_source_changed?(source)
--- a/spec/install/git_spec.rb
+++ b/spec/install/git_spec.rb
@@ -41,4 +41,43 @@ describe "bundle install" do
       expect(vendored_gems("bundler/gems/foo-1.0-#{revision_for(lib_path("foo-1.0"))[0..11]}")).to be_directory
     end
   end
+
+  context "local overrides" do
+    before do
+      build_git "rack", "0.8"
+
+      build_git "rack", "0.8", :path => lib_path("local-rack") do |s|
+        s.write "lib/rack.rb", "puts :LOCAL"
+      end
+
+      install_gemfile <<-G
+          source "file://#{gem_repo1}"
+          gem "rack", :git => "#{lib_path("rack-0.8")}", :branch => "master"
+      G
+
+      bundle %(config local.rack #{lib_path("local-rack")})
+    end
+
+    context "invalid revision and branch in gemfile.lock" do
+      it "should not explode on install" do
+        lockfile <<-G
+          GIT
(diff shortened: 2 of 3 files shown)


segiddins:
Please add a spec for this change?

drn:
Sure, I'll add one tonight

drn:
@segiddins - I've updated the commit to include specs that cover this issue. Let me know if you need anything else!

segiddins:
@indirect r?

lib/bundler/definition.rb
```diff
@@ -480,7 +480,16 @@ def specs_changed?(source, &block)
...
       end
 
-      !locked || unlocking || dependencies_for_source_changed?(locked) || source.specs != locked.specs
+      return true unless locked
+      return true if unlocking
+      return true if dependencies_for_source_changed?(locked)
+      begin
+        locked_specs = locked.specs
+      rescue GitError
+        locked_specs = Bundler::Index.new
```
drn: Can also just return true here

indirect:
I completely agree that "git repo is not checked out" is a bug, but I'm not sure that "git locals can be locked to non-existent branches and shas" should be the solution?

Why can't we get the gemspec from the locals repo, and validate against a branch or sha in the locals repo?

drn:
Hmm, maybe this is too broad of a patch.

To give you some background, my original use case is actually slightly different, though it's not reflected in the specs. I reproduced this error by having a lockfile that was pointing to an existent branch / sha. The install gemfile branch was then updated to another branch on a different sha in the git local.

I assumed that after I ran `bundle install` or `bundle update gem-name`, it would update the locked sha / branch to whatever the new branch was pointing to locally. However, I ran into the `Run 'bundle install' first` exception using both of the commands.

This pull solves my use case as well as the more general use case of when a lockfile is pointing to an invalid sha / branch.

I'm happy to help out with some code, but I'm not entirely sure what the desired behavior is. Can you clarify what the intended behavior is for these two commands:
* `bundle install`
  * Having a more descriptive exception raised that indicates that the locked sha / branch don't exist or are different than the current branch / head sha makes sense. Is that what you mean by adding validations?
* `bundle update gem-name`
  * Even if the lockfile sha is missing or the locked branch is different than the install gemfile branch, shouldn't the lockfile be updated to whatever the current head sha of the git local is?

indirect:
I think your usecase is now fixed by #4840! Thanks for reporting this.

PR #4674, opened 2016-06-16
[Plugin] Source plugins

asutoshpalai:
Adds source plugin. This is in continuation of #4608.


lib/bundler/dsl.rb
```diff
@@ -125,11 +125,26 @@ def gem(name, *args)
       @dependencies << dep
     end
 
-    def source(source, &blk)
-      source = normalize_source(source)
-      if block_given?
+    def source(source, *args, &blk)
+      options = args.last.is_a?(Hash) ? args.pop.dup : {}
+      options = normalize_hash(options)
+      if options && options.key?("type")
```
segiddins: options can't be nil here

lib/bundler/lockfile_parser.rb
```diff
@@ -22,14 +22,15 @@ class LockfileParser
...
     SPECS        = "  specs:".freeze
     OPTIONS      = /^  ([a-z]+): (.*)$/i
-    SOURCE       = [GIT, GEM, PATH].freeze
+    SOURCE       = [GIT, GEM, PATH, PLUGIN].freeze
 
     SECTIONS_BY_VERSION_INTRODUCED = {
       Gem::Version.create("1.0") => [DEPENDENCIES, PLATFORMS, GIT, GEM, PATH].freeze,
       Gem::Version.create("1.10") => [BUNDLED].freeze,
-      Gem::Version.create("1.12") => [RUBY].freeze,
+      Gem::Version.create("1.12") => [RUBY, PLUGIN].freeze,
```
segiddins: this should be 1.13 for PLUGIN

lib/bundler/plugin.rb
```diff
@@ -29,7 +31,7 @@ def install(names, options)
       save_plugins paths
     rescue PluginError => e
       paths.values.map {|path| Bundler.rm_rf(path) } if paths
-      Bundler.ui.error "Failed to install plugin #{name}: #{e.message}\n  #{e.backtrace.join("\n  ")}"
+      Bundler.ui.error "Failed to install plugin #{name}: #{e.message}\n  #{e.backtrace[0]}"
```
segiddins: why this change?

lib/bundler/plugin.rb
```diff
@@ -71,21 +73,49 @@ def add_command(command, cls)
...
+    # Checks if any plugin declares the source
+    def source?(name)
+      !index.source_plugin(name).nil?
+    end
+
+    # @return [Class] that handles the source. The calss includes API::Source
+    def source(name)
+      raise UnknownSourceError, "Source #{name} not found" unless source? name
+
+      load_plugin index.source_plugin name unless @sources.key? name
```
segiddins: please use parenthesis here

lib/bundler/plugin/api.rb
```diff
@@ -23,16 +23,28 @@ module Bundler
...
 
-      # The cache dir to be used by the plugins for persistance storage
+      # The plugins should declare that they provide a installation source
+      # through this helper.
+      #
+      # @param [String] the source type they provide
+      # @param [Class] (optional) class that handles the source. If not
+      #                 provided, the `self` class will be used.
+      def self.source(source, cls = self)
+        include Bundler::Plugin::API::Source
```
segiddins: why does this call include?

lib/bundler/plugin/api/source.rb
```diff
@@ -0,0 +1,102 @@
...
+          @dependency_names = []
+          @uri = opts["uri"]
+          @type = opts["type"]
+        end
+
+        def installed?
+          File.directory?(install_path)
+        end
+
+        def fetch_gemfiles
```
segiddins: what does this method do?

lib/bundler/plugin/api/source.rb
```diff
@@ -0,0 +1,102 @@
+# frozen_string_literal: true
+require "uri"
+require "digest/sha1"
+
+module Bundler
+  module Plugin
+    class API
+      module Source
```
segiddins: this looks like its a Class?

lib/bundler/lockfile_parser.rb
asutoshpalai: Actually there is a spec where `unknown_sections_in_lockfile` should return `nil`

lib/bundler/plugin.rb
asutoshpalai: You left a comment before merging the previous PR for this.

lib/bundler/plugin/api/source.rb
asutoshpalai: Sorry, forgot to document this class.

The reason this is module is a point where I was looking for suggestions.

My justification:

Our goal here is to provide some default implementations for many of source plugin methods and these implementations should be overridable so that if something doesn't work well for a source plugin they should be able to change it. Also at the same time all the methods of `API` should also be available.

I can see this being done in two ways:

1. (The one that's implemented here) **Runtime inheritance:** We have the implementations in a `module` (since `class` can't be inherited during runtime) and include that module **when a class registers for source plugin**. The advantage is that the only class that plugin writers will have to deal with is `API` and it does all the magic.

2. **Static inheritance:** We can make `API::Source` a subclass of `API` and the source plugins will inherit `API::Source` and the others can inherit `API` only. The advantage is that there is no confusion of runtime inheritance.

lib/bundler.rb
```diff
@@ -38,7 +39,6 @@ module Bundler
   autoload :MatchPlatform,          "bundler/match_platform"
   autoload :Mirror,                 "bundler/mirror"
   autoload :Mirrors,                "bundler/mirror"
-  autoload :Plugin,                 "bundler/plugin"
```
segiddins: why can't this be autoloaded any more?

lib/bundler/lockfile_parser.rb
```diff
@@ -22,14 +22,15 @@ class LockfileParser
     GIT          = "GIT".freeze
     GEM          = "GEM".freeze
     PATH         = "PATH".freeze
+    PLUGIN       = "PLUGIN".freeze
```
segiddins: maybe `"PLUGIN SOURCE"`?

lib/bundler/lockfile_parser.rb
```diff
@@ -116,14 +117,15 @@ def warn_for_outdated_bundler_version
...
+      GIT    => Bundler::Source::Git,
+      GEM    => Bundler::Source::Rubygems,
+      PATH   => Bundler::Source::Path,
+      PLUGIN => Bundler::Plugin,
     }.freeze
 
     def parse_source(line)
       case line
-      when GIT, GEM, PATH
+      when GIT, GEM, PATH, PLUGIN
```
segiddins: `when *source`

lib/bundler/plugin.rb
```diff
@@ -96,11 +127,12 @@ def installed?(plugin)
     # Post installation processing and registering with index
     #
     # @param [Hash] plugins mapped to their installtion path
-    def save_plugins(plugins)
+    # @param [Array<String>] names of auto added source plugins that can be ignored
+    def save_plugins(plugins, optionals = [])
```
segiddins: `optional_plugins`

lib/bundler/plugin/api.rb
```diff
@@ -51,6 +63,10 @@ def method_missing(name, *args, &blk)
         super unless Bundler.respond_to?(name)
         Bundler.send(name, *args, &blk)
       end
+
+      def respond_to_missing?(name, include_private = false)
+        Bundler.respond_to?(name) || super
```
segiddins: should also use the second param there

lib/bundler/plugin/api/source.rb
```diff
@@ -0,0 +1,157 @@
...
+      # expected to work. But, all source plugins have to override `fetch_gemfiles`
+      # and `install`. Defaults are also not provided for `remote!`, `cache!`
+      # and `unlock!`.
+      #
+      # The defaults shall work for most situations but nevertheless they can
+      # be (preferably should be) overridden as per the plugins' needs safely
+      # (as long as they behave as expected).
+      # On overriding `initialize` you should call super first.
+      module Source
+        attr_reader :uri, :options
```
segiddins: please document the attributes

lib/bundler/plugin/api/source.rb
```diff
@@ -0,0 +1,157 @@
...
+        end
+
+        # This is used by the default `spec` method to constructs the
+        # Specification objects for the gems and versions that can be installed
+        # by this source plugin.
+        #
+        # Note: If the spec method is overridden, this function is not necessary
+        #
+        # @return [Array<String>] paths of the gemfiles that can be installed
+        def fetch_gemfiles
```
segiddins: why do plugins need to concern themselves with gem files?

lib/bundler/plugin/api/source.rb
```diff
@@ -0,0 +1,157 @@
...
+        #   - The `loaded_from` for each of the specs shall be correct (it is
+        #   used to find the load path)
+        #
+        # @return [Bundler::Index] index containing the specs
+        def specs
+          files = fetch_gemfiles
+
+          Bundler::Index.build do |index|
+            files.each do |file|
+              next unless spec = Bundler.load_gemspec(file)
```
segiddins: so is `file` a gemfile or a gem spec?

lib/bundler/plugin/dsl.rb
```diff
@@ -12,9 +12,12 @@ class PluginGemfileError < PluginError; end
       # They will be handled by method_missing
       [:gemspec, :gem, :path, :install_if, :platforms, :env].each {|m| undef_method m }
 
+      attr_reader :auto_plugins
```
segiddins: please document this, I'm confused about what it is

lib/bundler.rb
asutoshpalai: https://github.com/bundler/bundler/pull/4674/files#diff-507c71857bb3b7292c61afb88df1c880L28
This change made starting of `bundler` without `API::Source` fail

lib/bundler/plugin/api/source.rb
asutoshpalai: My bad, they are `gemspec` files.

I guess I am seeing too much of `Gemfile` that I got confused.

lib/bundler/plugin/api/source.rb
asutoshpalai: `gemspec`, will fix that in next commit

lib/bundler/plugin/api.rb
```diff
@@ -40,7 +40,7 @@ def self.command(command, cls = self)
       # @param [Class] (optional) class that handles the source. If not
       #                 provided, the `self` class will be used.
       def self.source(source, cls = self)
-        include Bundler::Plugin::API::Source
+        cls.include Bundler::Plugin::API::Source
```
segiddins: this might need to be `cls.send :include`

lib/bundler/plugin/dsl.rb
```diff
@@ -12,6 +12,14 @@ class PluginGemfileError < PluginError; end
...
 
+      # This lists the plugins that was added automatically and not specified by
+      # the user.
+      #
+      # When we encounter :type attribute with a source block, we add a plugin
+      # by name bundler-source-<type> to list of plugins to be installed.
+      #
+      # These plugins are optional and are not installed when there is conflict
+      # with any other plugin.
       attr_reader :auto_plugins
```
segiddins: `inferred_plugins`?

lib/bundler/plugin.rb
```diff
@@ -128,9 +129,9 @@ def installed?(plugin)
     #
     # @param [Hash] plugins mapped to their installation path
     # @param [Array<String>] names of inferred source plugins that can be ignored
-    def save_plugins(plugins, optional_plugins = [])
-      plugins.each do |name, path|
-        path = Pathname.new path
+    def save_plugins(plugins, paths, optional_plugins = [])
```
segiddins: please doc `paths`

lib/bundler/plugin.rb
```diff
@@ -197,7 +197,16 @@ def load_plugin(name)
       # done to avoid conflicts
       path = index.plugin_path(name)
 
+      if insert_index = Bundler.rubygems.load_path_insert_index
```
segiddins: we should just be `#activate`ing the spec here, no?

lib/bundler/plugin/api.rb
```diff
@@ -59,6 +59,17 @@ def tmp(*names)
...
 
+      # Change directory helper that provides thread safety
+      #
+      # Sources should use this since, if possible, they are run in parallel
+      # threads
+      #
+      # @param [String] directory
+      # @param [Proc] block to execute
+      def chdir(dir, &blk)
+        SharedHelpers.chdir(dir, &blk)
```
segiddins: could just mix in shared helpers to the API?

lib/bundler/plugin.rb
segiddins: (or using `spec.add_self_to_load_path`?)

asutoshpalai: I tried to do that. But could find a way to get `spec` here.

Even if I store the spec's locations, `add_self_to_load_path` depended on `full_gem_path`, which itself depended on `source`.

The alternative to hard coded `lib` is, we can store the `full_required_paths` with index.

segiddins: we have to do something, since it is totally valid for a gem to have a require path that isn't lib

asutoshpalai: Yes, I am working on extending the serializer to with array. As soon as I get around that, will fix this.

allenzhao:
@asutoshpalai I was trying to edit `lib/bundler/errors.rb` and found that you've used a duplicated status code 23 in L44 which was already taken in L69

asutoshpalai:
@allenzhao Yes, right! Sorry, will fix that with the next push!

segiddins:
@asutoshpalai @indirect is this ready to merge?

asutoshpalai:
@segiddins I think it's now ready to merge

segiddins:
@asutoshpalai if you can get this passing, ill give it one last look over and then likely merge

asutoshpalai:
This shall pass, there is some problem with travis. It is giving error.

Since the last time it was passing, I just added some new specs that are
passing. (and another small change that I reverted)

asutoshpalai:
@segiddins I fixed the problem with the specs setup and they are passing now.

asutoshpalai:
Here is the path with which I fixed the specs
```diff
diff --git a/spec/support/rubygems_ext.rb b/spec/support/rubygems_ext.rb
index 95f596e..e45e69f 100644
--- a/spec/support/rubygems_ext.rb
+++ b/spec/support/rubygems_ext.rb
@@ -6,13 +6,15 @@ module Spec
   module Rubygems
     DEPS = begin
       deps = {
-        "fakeweb artifice rack compact_index" => nil,
+        "fakeweb artifice compact_index" => nil,
         "sinatra" => "1.2.7",
         # Rake version has to be consistent for tests to pass
         "rake" => "10.0.2",
         # 3.0.0 breaks 1.9.2 specs
         "builder" => "2.1.2",
         "bundler" => "1.12.0",
+        # 2.0.1 requires ruby >= 2.2.2
+        "rack" => "1.6.4",
       }
       # ruby-graphviz is used by the viz tests
       deps["ruby-graphviz"] = nil if RUBY_VERSION >= "1.9.3"
@@ -33,7 +35,7 @@ module Spec
         FileUtils.rm_rf(Path.base_system_gems)
         FileUtils.mkdir_p(Path.base_system_gems)
         puts "installing gems for the tests to use..."
-        DEPS.each {|n, v| install_gem(n, v) }
+        DEPS.sort {|a, _| a[1].nil? ? 1 : -1 }.each {|n, v| install_gem(n, v) }
         File.open(manifest_path, "w") {|f| f << manifest.join }
       end
```

segiddins:
@asutoshpalai I'll merge if you can rebase to get rid of the merge commit :)

segiddins:
@homu r+

PR #4676, opened 2016-06-16
Conservative updates

chrismo:
A port of bundler-patch to bundler proper.

Core team questions:
- [x] bundler-patch also works in vulnerable gem updates based on ruby-advisory-db content. I presume the core team would consider this out of scope for Bundler proper? I probably would, and there's no reason bundler-patch couldn't continue to exist as a plugin with this behavior.
- [x] "resolves foo only to latest patch - changing dependency case" - more comments on that spec. It's a weird edge case. Hopefully the specs and comments make it clear.
- [x] Name of new class: GemVersionPromoter. It's gone through a few renames already and every option I think of I hate now :)
- [x] The `--minimal` flag is also something bundler-patch has that has debatable value. Would love to hear opinions on it and could be easily swayed to remove it or keep it. If I'm on 1.0.1, and both 1.0.2 and 1.0.15 exist, `--minimal` will resolve to `1.0.2` instead of the latest `1.0.15`. Dunno if there's value in that.
- [x]  even without `--strict` mode, this code will _remove_ older versions from the options handed back to Molinillo (in `--patch` and `--minor` cases). This is a weird case, but I have seen `bundle update` proper regress a gem version backwards if its parent gem changed to an older dependency (I guess in a bug fix case). I'm ok with this, but it could be considered an inconsistency with the default `--major` case.
- [x] if we roll this out first as an undocumented feature, do we require a config option to toggle it on? We wouldn't have to have one, none of the new behavior will kick in until a new flag (`--patch` or `--minor`) is passed. 

---

Notes for issues to create that could be handled separately after this PR, esp. if we release undocumented/unsupported to begin with.

- Updating `outdated` to use the new flags and resolution
- adding some warnings when dependent updates go past --patch or --minor level (with an instruction to use --strict if they don’t want that)
- adding a warning when an unlocked gem doesn't move (possibly w/ a reason why?). the lack of feedback otherwise can be disconcerting, like "did I not do it right?"
- possibly adding a dry-run flag - though after looking at that today, that would make the most sense inside Installer, and make that flag an option for both install and update (drilling through all of the calls to push that option in looks like it would be annoying).
- man page documentation

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -208,12 +208,16 @@ module Bundler
       "Update ruby specified in Gemfile.lock"
     method_option "bundler", :type => :string, :lazy_default => "> 0.a", :banner =>
       "Update the locked version of bundler"
+
+    # MODO: hide these from help to release undocumented?
     method_option "patch", :type => :boolean, :banner =>
       "Prefer updating only to next patch version"
     method_option "minor", :type => :boolean, :banner =>
       "Prefer updating only to next minor version"
     method_option "major", :type => :boolean, :banner =>
       "Prefer updating to next major version (default)"
+    method_option "strict", :type => :boolean, :banner =>
+      "Do not allow any gem to be updated past latest --patch/--minor/--major"
     def update(*gems)
       require "bundler/cli/update"
       Update.new(options, gems).run
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -208,15 +208,13 @@ module Bundler
       "Update ruby specified in Gemfile.lock"
     method_option "bundler", :type => :string, :lazy_default => "> 0.a", :banner =>
       "Update the locked version of bundler"
-
-    # MODO: hide these from help to release undocumented?
-    method_option "patch", :type => :boolean, :banner =>
+    method_option "patch", :type => :boolean, :hide => true, :banner =>
       "Prefer updating only to next patch version"
-    method_option "minor", :type => :boolean, :banner =>
+    method_option "minor", :type => :boolean, :hide => true, :banner =>
       "Prefer updating only to next minor version"
-    method_option "major", :type => :boolean, :banner =>
+    method_option "major", :type => :boolean, :hide => true, :banner =>
       "Prefer updating to next major version (default)"
-    method_option "strict", :type => :boolean, :banner =>
+    method_option "strict", :type => :boolean, :hide => true, :banner =>
       "Do not allow any gem to be updated past latest --patch/--minor/--major"
     def update(*gems)
       require "bundler/cli/update"
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -208,6 +208,12 @@ module Bundler
       "Update ruby specified in Gemfile.lock"
     method_option "bundler", :type => :string, :lazy_default => "> 0.a", :banner =>
       "Update the locked version of bundler"
+    method_option "patch", :type => :boolean, :banner =>
+      "Prefer updating only to next patch version"
+    method_option "minor", :type => :boolean, :banner =>
+      "Prefer updating only to next minor version"
+    method_option "major", :type => :boolean, :banner =>
+      "Prefer updating to next major version (default)"
     def update(*gems)
       require "bundler/cli/update"
       Update.new(options, gems).run
--- a/lib/bundler/cli/update.rb
+++ b/lib/bundler/cli/update.rb
@@ -40,7 +40,7 @@ module Bundler
       end
 
       patch_level = [:major, :minor, :patch].detect {|v| options.keys.include?(v.to_s) }
-      Bundler.definition.update_opts.level = patch_level
+      Bundler.definition.dependency_search.level = patch_level
 
       Bundler::Fetcher.disable_endpoint = options["full-index"]
 
--- a/lib/bundler/cli/update.rb
+++ b/lib/bundler/cli/update.rb
@@ -39,9 +39,10 @@ module Bundler
         Bundler.definition(:gems => gems, :sources => sources, :ruby => options[:ruby])
       end
 
-      patch_level = [:major, :minor, :patch].detect {|v| options.keys.include?(v.to_s) }
+      patch_level = [:major, :minor, :patch].select {|v| options.keys.include?(v.to_s) }
+      raise ProductionError, "Provide only one of the following options: #{patch_level.join(", ")}" unless patch_level.length <= 1
       Bundler.definition.gem_version_promoter.tap do |gvp|
-        gvp.level = patch_level || :major
+        gvp.level = patch_level.first || :major
         gvp.strict = options[:strict]
       end
 
--- a/lib/bundler/cli/update.rb
+++ b/lib/bundler/cli/update.rb
@@ -3,7 +3,7 @@ module Bundler
   class CLI::Update
     attr_reader :options, :gems
     def initialize(options, gems)
-      @options = options
+      @options = options.tap {|o|p o}
       @gems = gems
     end
 
--- a/lib/bundler/cli/update.rb
+++ b/lib/bundler/cli/update.rb
@@ -39,7 +39,7 @@ module Bundler
         Bundler.definition(:gems => gems, :sources => sources, :ruby => options[:ruby])
       end
 
-      patch_level = [:major, :minor, :patch].detect { |v| options.keys.include?(v.to_s) }
+      patch_level = [:major, :minor, :patch].detect {|v| options.keys.include?(v.to_s) }
       Bundler.definition.update_opts.level = patch_level
 
       Bundler::Fetcher.disable_endpoint = options["full-index"]
--- a/lib/bundler/cli/update.rb
+++ b/lib/bundler/cli/update.rb
@@ -40,7 +40,10 @@ module Bundler
       end
 
       patch_level = [:major, :minor, :patch].detect {|v| options.keys.include?(v.to_s) }
-      Bundler.definition.gem_version_promoter.level = patch_level || :major
+      Bundler.definition.gem_version_promoter.tap do |gvp|
+        gvp.level = patch_level || :major
+        gvp.strict = options[:strict]
+      end
 
       Bundler::Fetcher.disable_endpoint = options["full-index"]
 
--- a/lib/bundler/cli/update.rb
+++ b/lib/bundler/cli/update.rb
@@ -40,7 +40,7 @@ module Bundler
       end
 
       patch_level = [:major, :minor, :patch].detect {|v| options.keys.include?(v.to_s) }
-      Bundler.definition.gem_version_promoter.level = patch_level
+      Bundler.definition.gem_version_promoter.level = patch_level || :major
 
       Bundler::Fetcher.disable_endpoint = options["full-index"]
 
--- a/lib/bundler/cli/update.rb
+++ b/lib/bundler/cli/update.rb
@@ -39,7 +39,8 @@ module Bundler
         Bundler.definition(:gems => gems, :sources => sources, :ruby => options[:ruby])
       end
 
-      Bundler.definition.update_opts.level = [:major, :minor, :patch].detect { |v| options.keys.include?(v) }
+      patch_level = [:major, :minor, :patch].detect { |v| options.keys.include?(v.to_s) }
+      Bundler.definition.update_opts.level = patch_level
 
       Bundler::Fetcher.disable_endpoint = options["full-index"]
 
--- a/lib/bundler/cli/update.rb
+++ b/lib/bundler/cli/update.rb
@@ -40,7 +40,7 @@ module Bundler
       end
 
       patch_level = [:major, :minor, :patch].detect {|v| options.keys.include?(v.to_s) }
-      Bundler.definition.dependency_search.level = patch_level
+      Bundler.definition.gem_version_promoter.level = patch_level
 
       Bundler::Fetcher.disable_endpoint = options["full-index"]
 
--- a/lib/bundler/cli/update.rb
+++ b/lib/bundler/cli/update.rb
@@ -3,7 +3,7 @@ module Bundler
   class CLI::Update
     attr_reader :options, :gems
     def initialize(options, gems)
-      @options = options.tap {|o|p o}
+      @options = options
       @gems = gems
     end
 
@@ -38,6 +38,8 @@ module Bundler
 
         Bundler.definition(:gems => gems, :sources => sources, :ruby => options[:ruby])
       end
+
+      Bundler.definition.update_opts.level = [:major, :minor, :patch].detect { |v| options.keys.include?(v) }
 
       Bundler::Fetcher.disable_endpoint = options["full-index"]
 
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -7,7 +7,7 @@ module Bundler
   class Definition
     include GemHelpers
 
-    attr_reader :dependencies, :platforms, :ruby_version, :locked_deps, :dependency_search
+    attr_reader :dependencies, :platforms, :ruby_version, :locked_deps, :gem_version_promoter
 
     # Given a gemfile and lockfile creates a Bundler definition
     #
@@ -94,7 +94,7 @@ module Bundler
       end
       @unlocking ||= @unlock[:ruby] ||= (!@locked_ruby_version ^ !@ruby_version)
 
-      @dependency_search = Resolver::DependencySearch.new(@locked_specs, @unlock[:gems])
+      @gem_version_promoter = GemVersionPromoter.new(@locked_specs, @unlock[:gems])
 
       current_platform = Bundler.rubygems.platforms.map {|p| generic(p) }.compact.last
       add_platform(current_platform)
@@ -223,7 +223,7 @@ module Bundler
         else
           # Run a resolve against the locally available gems
           Bundler.ui.debug("Found changes from the lockfile, re-resolving dependencies because #{change_reason}")
-          last_resolve.merge Resolver.resolve(expanded_dependencies, index, source_requirements, last_resolve, ruby_version, dependency_search)
+          last_resolve.merge Resolver.resolve(expanded_dependencies, index, source_requirements, last_resolve, ruby_version, gem_version_promoter)
         end
       end
     end
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -83,8 +83,6 @@ module Bundler
         @locked_sources = []
       end
 
-      @update_opts = Resolver::UpdateOptions.new(@locked_specs)
-
       @unlock[:gems] ||= []
       @unlock[:sources] ||= []
       @unlock[:ruby] ||= if @ruby_version && @locked_ruby_version
@@ -95,6 +93,8 @@ module Bundler
         @ruby_version.diff(locked_ruby_version_object)
       end
       @unlocking ||= @unlock[:ruby] ||= (!@locked_ruby_version ^ !@ruby_version)
+
+      @update_opts = Resolver::UpdateOptions.new(@locked_specs, @unlock[:gems])
 
       current_platform = Bundler.rubygems.platforms.map {|p| generic(p) }.compact.last
       add_platform(current_platform)
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -7,7 +7,7 @@ module Bundler
   class Definition
     include GemHelpers
 
-    attr_reader :dependencies, :platforms, :ruby_version, :locked_deps, :update_opts
+    attr_reader :dependencies, :platforms, :ruby_version, :locked_deps, :dependency_search
 
     # Given a gemfile and lockfile creates a Bundler definition
     #
@@ -94,7 +94,7 @@ module Bundler
       end
       @unlocking ||= @unlock[:ruby] ||= (!@locked_ruby_version ^ !@ruby_version)
 
-      @update_opts = Resolver::UpdateOptions.new(@locked_specs, @unlock[:gems])
+      @dependency_search = Resolver::DependencySearch.new(@locked_specs, @unlock[:gems])
 
       current_platform = Bundler.rubygems.platforms.map {|p| generic(p) }.compact.last
       add_platform(current_platform)
@@ -223,7 +223,7 @@ module Bundler
         else
           # Run a resolve against the locally available gems
           Bundler.ui.debug("Found changes from the lockfile, re-resolving dependencies because #{change_reason}")
-          last_resolve.merge Resolver.resolve(expanded_dependencies, index, source_requirements, last_resolve, ruby_version, update_opts)
+          last_resolve.merge Resolver.resolve(expanded_dependencies, index, source_requirements, last_resolve, ruby_version, dependency_search)
         end
       end
     end
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -94,7 +94,7 @@ module Bundler
       end
       @unlocking ||= @unlock[:ruby] ||= (!@locked_ruby_version ^ !@ruby_version)
 
-      @gem_version_promoter = GemVersionPromoter.new(@locked_specs, @unlock[:gems])
+      @gem_version_promoter = create_gem_version_promoter
 
       current_platform = Bundler.rubygems.platforms.map {|p| generic(p) }.compact.last
       add_platform(current_platform)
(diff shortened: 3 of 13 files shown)


lib/bundler/cli/update.rb
```diff
@@ -37,6 +37,9 @@ def run
         Bundler.definition(:gems => gems, :sources => sources, :ruby => options[:ruby])
       end
 
+      patch_level = [:major, :minor, :patch].detect { |v| options.keys.include?(v.to_s) }
```
segiddins: What if multiple are specified?

lib/bundler/definition.rb
```diff
@@ -7,7 +7,7 @@ module Bundler
   class Definition
     include GemHelpers
 
-    attr_reader :dependencies, :platforms, :ruby_version, :locked_deps
+    attr_reader :dependencies, :platforms, :ruby_version, :locked_deps, :update_opts
```
segiddins: `update_options`

lib/bundler/cli/update.rb
chrismo: should favor left to right (maj, min, pat) ... specifying multiple wouldn't make sense internally. dunno if it should favor the last one specified by the user? e.g. `--minor --patch` => `--patch` vs. `--patch --minor` => `--minor`

lib/bundler/definition.rb
chrismo: yeah, agree - that's probably going to be renamed anyway though, i've pushed more functionality into that internal class and it needs a better name. I'm thinking DependencySearch so far. #namingishard

chrismo:
After rubocop, seeing good test results. 1 failing that's intentional on my part, 2 that seem unrelated.

spec/resolver/basic_spec.rb
```diff
@@ -100,8 +100,82 @@
...
+      should_consv_resolve_and_include [:patch, :minimal], [], %w(foo-1.4.4 bar-2.0.4)
+    end
+
+    it "resolves foo only to latest patch - same dependency case" do
+      @locked = locked(%w(foo 1.3.7), %w(bar 2.0.3))
+      # bar is locked, and the lock holds here because the dependency on bar doesn't change on the matching foo version.
+      should_consv_resolve_and_include :patch, ["foo"], %w(foo-1.3.8 bar-2.0.3)
+    end
+
+    it "resolves foo only to latest patch - changing dependency case" do
```
chrismo: just thinking through this - here's the potential design flaw. In all other cases, `strict` means: ensure no _unlocking_ gem gets past the `patch`/`minor` threshold. And accordingly, not specifying `strict` means for _unlocking_ gems, if one of them has to go past the `patch`/`minor` threshold to satisfy the dependency tree, then ... ok.

But in THIS case, the 'not specifying `strict`' case means for _unlocking gems *and their dependencies*_ ... and thus having to use `strict` to get the expected behavior is a debatable flaw. 

Compare this case to the prior spec: "resolves foo only to latest patch - same dependency case" - that should highlight the inconsistency.

I'm not sure there's an easy code fix for this case, though.

chrismo:
Another discussion point - bundler-patch also works in vulnerable gem updates based on ruby-advisory-db content. I presume the core team would consider this out of scope for Bundler proper? I probably would, and there's no reason bundler-patch couldn't continue to exist as a plugin with this behavior.

spec/resolver/basic_spec.rb
```diff
@@ -100,8 +100,82 @@
...
+      should_consv_resolve_and_include [:minor, :strict], [], %w(foo-1.5.0 bar-2.1.1)
+    end
+
+    it "resolves all gems to latest minor minimal" do
+      # minimal is set, and it takes precedence over minor. not sure what is the PoLS in this case. Not sure
+      # if minimal is a great option in the first place. It exists to help a case where there are many, many
+      # versions and I'd rather go from 1.0.2 to 1.0.3 instead of 1.0.45. But, we could consider killing the
+      # minimal option altogether. If that's what you need, use the Gemfile dependency.
+      should_consv_resolve_and_include [:minor, :minimal], [], %w(foo-1.4.4 bar-2.0.4)
+    end
```
chrismo: The `--minimal` flag is also something bundler-patch has that has debatable value. Would love to hear opinions on it and could be easily swayed to remove it or keep it.

lib/bundler/definition.rb
```diff
@@ -94,7 +94,7 @@ def initialize(lockfile, dependencies, sources, unlock, ruby_version = nil, opti
       end
       @unlocking ||= @unlock[:ruby] || (!@locked_ruby_version ^ !@ruby_version)
 
-      @gem_version_promoter = GemVersionPromoter.new(@locked_specs, @unlock[:gems])
+      create_gem_version_promoter
```
segiddins: for testing purposes, maybe make the assignment here instead of in `create_gem_version_promoter` ?

chrismo: Do you mean inline #create_gem_version_promoter? constructor already feels rather fat to me - I wanted to encapsulate that in its own method with the additional logic to re-capture the locked_specs. I tried to find a way to re-use the @locked_specs or locked local var rather than re-executing, but it didn't seem clean to me with the various conditionals.

But if that's not what you mean, I'm not understanding your comment ... ?

segiddins: no, I mean have `@gem_version_promoter = create_gem_version_promoter`

chrismo: oh right - yeah, I'm down with that.

indirect:
- ruby-advisory-db: sounds like a great plugin. :)
 - changing dependency spec: I buy your argument based on the docs that if it's not in the Gemfile (and you don't pass any restriction options), you're implicitly declaring that you don't care about how it updates. This seems okay to me.
 - GemVersionPromoter: works for me.
 - `--minimal` flag: let's drop it and see if the masses demand it. if so we can always add it. :)

lib/bundler/gem_version_promoter.rb
```diff
@@ -0,0 +1,127 @@
...
+        # We only need the version here so it's ok to hard code this to the first instance.
+        locked_spec = @locked_specs[gem_name].first
+
+        if @strict
+          filter_dep_specs(dep_specs, locked_spec)
+        else
+          sort_dep_specs(dep_specs, locked_spec)
+        end.tap do |specs|
+          if ENV["DEBUG_PATCH_RESOLVER"] # MODO: proper debug flag name, check and proper debug output
+            STDERR.puts before_result
```
segiddins: please use the `Bundler.ui` debug methods here

chrismo: will do. I need to revisit that flag too, perhaps re-use `DEBUG_RESOLVER`.

chrismo:
@indirect thx for the feedback - added a couple of more to the checklist

segiddins:
* No need for an explicit feature flag, since the new behavior is still opt-in via the CLI options
* I think in strict mode, we might want to `>= ` the current version in the locked deps handed to molinillo, if that answers your question?

lib/bundler/cli.rb
```diff
@@ -206,6 +206,12 @@ def install
       "Update ruby specified in Gemfile.lock"
     method_option "bundler", :type => :string, :lazy_default => "> 0.a", :banner =>
       "Update the locked version of bundler"
+    method_option "patch", :type => :boolean, :banner =>
+      "Prefer updating only to next patch version"
```
segiddins: isn't this  `only allow updating gems' patch versions` ?

lib/bundler/cli/update.rb
```diff
@@ -37,6 +37,9 @@ def run
         Bundler.definition(:gems => gems, :sources => sources, :ruby => options[:ruby])
       end
 
+      patch_level = [:major, :minor, :patch].detect {|v| options.keys.include?(v.to_s) }
```
segiddins: should raise if more than one is specified

lib/bundler/gem_version_promoter.rb
```diff
@@ -0,0 +1,127 @@
+# frozen_string_literal: true
+module Bundler
+  class GemVersionPromoter
```
segiddins: since this is a new class, it would be nice to add API docs for it

lib/bundler/gem_version_promoter.rb
```diff
@@ -0,0 +1,127 @@
...
+
+    def level=(value)
+      v = begin
+        case value
+        when String, Symbol
+          value.to_sym
+        end
+      end
+
+      @level = [:major, :minor, :patch].include?(v) ? v : @level_default
```
segiddins: imho this should raise for an unknown level value

lib/bundler/gem_version_promoter.rb
```diff
@@ -0,0 +1,127 @@
...
+      before_result = "before sort_versions: #{debug_format_result(dep, dep_specs).inspect}"
+
+      result = @sort_versions[dep] ||= begin
+        gem_name = dep.name
+
+        # An Array per version returned, different entries for different platforms.
+        # We only need the version here so it's ok to hard code this to the first instance.
+        locked_spec = @locked_specs[gem_name].first
+
+        if @strict
```
segiddins: use the accessor method

lib/bundler/resolver.rb
```diff
@@ -172,14 +172,14 @@ def __dependencies
...
+    def self.resolve(requirements, index, source_requirements = {}, base = [], ruby_version = nil, gem_version_promoter = GemVersionPromoter.new)
       base = SpecSet.new(base) unless base.is_a?(SpecSet)
-      resolver = new(index, source_requirements, base, ruby_version)
+      resolver = new(index, source_requirements, base, ruby_version, gem_version_promoter)
       result = resolver.start(requirements)
       SpecSet.new(result)
     end
 
-    def initialize(index, source_requirements, base, ruby_version)
+    def initialize(index, source_requirements, base, ruby_version, gem_version_promoter = GemVersionPromoter.new)
```
segiddins: this doesn't need a default value

lib/bundler/resolver.rb
```diff
@@ -267,7 +268,9 @@ def search_for(dependency)
           []
         end
       end
-      search.select {|sg| sg.for?(platform, @ruby_version) }.each {|sg| sg.activate_platform!(platform) }
+      platform_results = search.select {|sg| sg.for?(platform, @ruby_version) }.each {|sg| sg.activate_platform!(platform) }
+      return platform_results if @gem_version_promoter.level == :major # default behavior
+      @gem_version_promoter.sort_versions(dependency, platform_results)
```
segiddins: can this be done inside the cache, instead of outside?

lib/bundler/gem_version_promoter.rb
```diff
@@ -0,0 +1,127 @@
...
+      filtered = specs.select {|s| s.first.version >= locked_version }
+
+      filtered.sort do |a, b|
+        a_ver = a.first.version
+        b_ver = b.first.version
+        case
+        when a_ver.segments[0] != b_ver.segments[0]
+          b_ver <=> a_ver
+        when !(@level == :minor) && (a_ver.segments[1] != b_ver.segments[1])
+          b_ver <=> a_ver
```
segiddins: note that the second segment can be a string instead of an integer

spec/support/indexes.rb
```diff
@@ -49,6 +49,23 @@ def gem(*args, &blk)
       build_spec(*args, &blk).first
     end
 
+    def locked(*args)
+      Bundler::SpecSet.new(args.map do |name, version|
+        gem(name, version)
+      end)
+    end
+
+    def should_consv_resolve_and_include(opts, unlock, specs)
```
segiddins: `should_conservatively_resolve_and_include`

lib/bundler/cli.rb
chrismo: there's a `--strict` flag I hadn't added here yet I just realized ... 

If `--strict` is also used, then the code will actually remove versions fed back to Molinillo, but that can result in no updates happening or VersionConflict cases that report a version can't be found that actually does exist in the index, causing a very weird "looks like a bug" experience. (hmmm ... while I did see VersionConflict cases early on, I haven't seen one in a while ... I may need to try and recreate that in a spec -- maybe that was due to an early bug and shouldn't be possible). 

So, not using `--strict` is the default and it means the code will only sort to the front (not filter out) the patch versions first - which favors getting something done, but it's potentially not _as_ conservative. 

One challenge still existing is trying to give better output in these conservative cases, because sometimes no change will happen at all and there's no feedback to the user trying to explain "well, you wanted to go up to 1.8.3 from 1.8.2, but 1.8.3 needs a minor increment and you fed it --strict so it couldn't do anything" - but I've no idea if that's even plausible. It wouldn't be too unlike the Conflict output ... but these aren't cases of a Conflict.

I did think also there could be a way to detect increments beyond what was requested and warn those somehow. 

#pascalsapology

lib/bundler/resolver.rb
chrismo: most likely. part of this is leftovers from how i had to hack on this from the outside in bundler-patch - part of this was due to a weird bug I never chased down where on large resolves sometimes bundler clears the result array in-place, corrupting the cache. it never matters in bundler due to the pre-existing search.select effectively duping the result ... but it mattered in bundler-patch which worksaround it by duping its result (here in the last line of `#sort_versions`). 

I did have an idea that freezing the result could flush out the (presumably) offending code, and then, yeah, the whole thing could go inside. Or the whole thing could go inside and then we #dup once right here.

lib/bundler/cli.rb
segiddins: > So, not using --strict is the default and it means the code will only sort to the front (not filter out) the patch versions first - which favors getting something done, but it's potentially not as conservative.

AHHH, this was the whole piece I was missing, sorry

lib/bundler/resolver.rb
segiddins: we already dup, effectively, two lines above with the `select`

indirect:
@chrismo: we want to stop allowing `update` to downgrade gems, so removing downgrades from new flags seems good to me. since this is only new flags, no need for a feature flag either.

lib/bundler/resolver.rb
chrismo: yeah, i wasn't sure if you wanted that part in, too---- well, you couldn't with the platform bit. i'll take a look.

lib/bundler/cli.rb
chrismo: no worries - option isn't listed there yet.

chrismo:
i imagine it's a rare case, but if i need to bugfix `foo` with a little new code _and_ fall back to an older version of dependent gem `bar` ... if we always eliminate older gems, there'd be no way to resolve that.

chrismo:
Thinking on this more - I think the best balance is: (a) do NOT filter older versions from the non-strict case, and (b) remove older versions from the `--strict` case. Otherwise, there's no way to resolve the rare but (IMO) legit case of regressing a dependent gem.

chrismo:
(For a real life case, bundler-patch itself started with slop v. 4, but that conflicted with any bundle including pry, which uses slop v. 3. So I released a newer version of bundler-patch that required an older version of slop).

chrismo:
There's also an existing test that tests a regression: https://github.com/bundler/bundler/blob/master/spec/commands/install_spec.rb#L155-L168 - from activesupport 2.3.5 back to 2.3.2

lib/bundler/gem_version_promoter.rb
```diff
@@ -1,9 +1,11 @@
 # frozen_string_literal: true
 module Bundler
+  # MODO: docs
```
segiddins: `MODO` ?

lib/bundler/resolver.rb
```diff
@@ -250,28 +250,34 @@ def search_for(dependency)
...
-        else
-          []
         end
+        # MODO: ideally trying to get GVP to sort by major. but some test cases aren't showing consistent order
+        # by default. (so we don't have to have this conditional here)
+        if @gem_version_promoter.level != :major
+          @gem_version_promoter.sort_versions(dependency, spec_groups)
+        else
+          spec_groups
+        end # .tap {|sg| STDERR.puts "sg>> #{sg.inspect}" }
```
segiddins: commented-out code

lib/bundler/resolver.rb
```diff
@@ -250,28 +250,34 @@ def search_for(dependency)
...
         end
-        if results.any?
-          nested = []
-          results.each do |spec|
-            version, specs = nested.last
-            if version == spec.version
-              specs << spec
-            else
-              nested << [spec.version, [spec]]
+        spec_groups = begin
```
segiddins: the `begin` isn't necessary, the `if/else` already returns a value

lib/bundler/gem_version_promoter.rb
chrismo: there's too many TODOs throughout the codebase, I'm using MODO (like 'chrisMO DO') to custom filter my editor down to ones that are 'mine'. it's goofy, i know.

spec/commands/update_spec.rb
```diff
@@ -464,7 +464,7 @@
         gem 'foo'
       G
 
-      bundle "update --patch"
+      bundle "update --patch", :env => { "DEBUG_RESOLVER" => true }
```
segiddins: should include `expect_err => true`

lib/bundler/resolver.rb
chrismo: yeah, I did that because without, then RubyMine indents the whole thing which makes Rubocop mad. So rather than figure out how to get RubyMine to make like Rubocop in this one case, I've taken to this. But it's sloppy, you're right.

lib/bundler/resolver.rb
chrismo: yeah, sorry - i'm playing with those lines right now - i knew i was coming back to that section next, didn't think about it before pushing that commit.

lib/bundler/resolver.rb
```diff
@@ -250,32 +250,30 @@ def search_for(dependency)
...
-          @gem_version_promoter.sort_versions(dependency, spec_groups)
+          nested.reduce([]) do |groups, (version, specs)|
+            next groups if locked_requirement && !locked_requirement.satisfied_by?(version)
+            groups << SpecGroup.new(specs)
+          end
         else
+          []
+        end
+        # GVP handles major itself, but it's still a bit risky to trust it with it
+        # until we get it settled with new behavior. For 2.x it can take over all cases.
```
segiddins: also a performance concern. also, looking from the code, the resorting might mess with source ordering?

chrismo: agree on perf concern, can you say moar words on source ordering?

segiddins: @indirect would know better than I, but I'm fairly certain that the order things are returned from the index search can be more than just by version

spec/resolver/basic_spec.rb
```diff
@@ -188,8 +188,45 @@
...
+
+        # base represents declared dependencies in the Gemfile that are still satisfied by the lockfile
+        @base = Bundler::SpecSet.new([])
+
+        # locked represents versions in lockfile
+        @locked = locked(%w(foo 1.4.3), %w(bar 2.2.3))
+      end
+
+      after :each do
+        ENV['DEBUG_RESOLVER'] = nil
```
segiddins: this should be necessary since the spec helper resets the env for every test

chrismo:
@segiddins the two failing tests are intentional at this point, I don't see why Molinillo gives up on 1.4.3 here: https://travis-ci.org/bundler/bundler/jobs/140130543#L3563

It hasn't been able to establish 1.4.4 as a viable option, but somehow when starting to consider 1.4.3 it says, oh wait, 1.4.4 and then gives up, even though -- I think -- 1.4.3 should work out here. I may try to crack open Molinillo and dive in at some point, but figured you could save me a ton of time :)

segiddins:
That looks like a Molinillo bug in either unwinding or reversing operations

segiddins:
@chrismo actually you're returning versions in the wrong order -- specs are considered last to first

chrismo:
This whole "conservative updates" thing functions by returning versions in specific orders to encourage Molinillo in the direction we prefer (:patch or :minor, etc). Does that interfere with certain optimization strategies?

chrismo:
specs here call that out: https://github.com/bundler/bundler/pull/4676/files#diff-ea30f9e81560dce1c6e2704d08767dbeR88

chrismo:
(all the other specs are passing fine - this new spec is a weirder case where the spec is expecting one version to revert to an older version than it started with).

segiddins:
So the order the resolver will try these specs in is
```
1.7.7
1.7.9
1.7.8
1.8.1
1.8.0
2.0.1
2.0.0
1.7.6
1.6.5
1.5.4
```

segiddins:
@chrismo after staring at it for a while, the test is wrong -- that is regressing the version of bar from what's locked

chrismo:
Yeah, the regression is intentional in this case - this spec is "could revert to a previous version level patch" - the comment for the context block says:
```ruby
    # Why would this happen in real life? If bar 2.2 has a bug that the author of foo wants to bypass
    # by reverting the dependency, the author of foo could release a new gem with an older requirement.
```
Current versions of Bundler can revert a gem to an older version under the right circumstances. Right now just trying to sort out of this is leaning towards 'bug' or 'why would we ever want this to work?' :)

segiddins:
@chrismo I thought that's what happens normally, but allowing anything to regress is blocked when using --strict?

chrismo:
ah right - this _is_ the strict case, I'd pasted the spec name from above, not this one: "will not revert to a previous version in strict mode level patch (FAILED - 1)" ... 

(sorry if the following is pedantic, trying to think it through out loud to make sure I'm not confusing myself)

```ruby
        @index = build_index do
          gem("foo", "1.4.3") { dep "bar", "~> 2.2" }
          gem("foo", "1.4.4") { dep "bar", "~> 2.1.0" }
          gem("foo", "1.5.0") { dep "bar", "~> 2.0.0" }
          gem "bar", %w(2.0.5 2.1.1 2.2.3)
        end

        @locked = locked(%w(foo 1.4.3), %w(bar 2.2.3))
```

first we attempt to update foo, and since we're patch level, it sorts it so 1.4.4 is considered first, then 1.4.3, (1.5.0 is also eliminated due to strict flag, but that's irrelevant here)
```ruby
before sort_versions: ["foo (>= 0)", [["1.4.3", []], ["1.4.4", []], ["1.5.0", []]], :patch, :strict]
 after sort_versions: ["foo (>= 0)", [["1.4.3", []], ["1.4.4", []]], :patch, :strict]
```

1.4.4 wants bar ~> 2.1, but we're locked at 2.2 and in strict mode so we remove every version outside the current level scope:
```ruby
before sort_versions: ["bar (~> 2.1.0)", [["2.1.1", []]], :patch, :strict]
 after sort_versions: ["bar (~> 2.1.0)", [], :patch, :strict]
```

So then it unwinds as it should and says it'll try 1.4.3, which should find bar 2.2, which will leave everything where it started with - which is fine - we requested `strict` ... but then molinillo gives up right at that point and that's what I'm confused by:
```
      Unwinding for conflict: bar (~> 2.1.0)
  Creating possibility state for foo (1 remaining)
    Attempting to activate foo (1.4.3)
    Found existing spec (foo (1.4.4))
    Creating possibility state for bar (~> 2.1.0) (0 remaining)
      Unwinding for conflict: bar (~> 2.1.0)
```
If it gave 1.4.3 a chance, it'd search_for bar ~> 2.2, find one and then be good.

segiddins:
Yeah, you're right. ¯\_(ツ)_/¯

chrismo:
@indirect @segiddins IMO, we can still go forward with an undocumented release with these two tests unresolved (har) (I can change them to pending to get the suite passing).

segiddins:
Cool, I'll try and review this next week!

segiddins:
👍 as long as this doesn't negatively effect performance too badly

chrismo:
yeup.

lib/bundler/cli/update.rb
segiddins: this still hasn't been addressed

segiddins:
@chrismo I'm happy to ship merge this &  ship it (undocumented) in 1.13 if you think it's ready (and have verified the performance isn't too bad)

chrismo:
Cool - I was presuming you would look at performance stuff, but happy to help with that, esp. if you have some worst case test scenarios with Molinillo or whatevs that I can check it out against.

segiddins:
We don't really have benchmarks for bundle update at the moment, but I'd just say pick a large, outdated gemfile, and play around with running `bundle update` with different combos of gem names and update flags and make sure nothing is pathological

chrismo:
ok

chrismo:
ok, performance looks good - here's a gist with some details. https://gist.github.com/chrismo/4b68d64db8d1d99245c817321b0b9bc5

segiddins:
@homu r+

segiddins:
@homu retry

segiddins:
@homu retry

chrismo:
I think I found the problem here. Should have commits soon.

chrismo:
@indirect or @segiddins - can you retry the auto-merge?

segiddins:
@homu r+

chrismo:
w00t - thx :)

PR #4688, opened 2016-06-19
[WIP] Rescue StandardError instead of Exception

dpulliam:
@indirect

Hey Y'all,

I wanted to try submitting my first bundler patch. I took a look at the PullReview for bundler and saw there was an issue with rescuing Exception in several files.  I went though and basically replaced Exception with StandardError. 

There were a situation where StandardError was not passing tests on its own. In the dsl.rb I needed to rescue both StandardError and SyntaxError. 

I noticed a couple tests continued to fail. They fail on both master and my branch so I am not sure if its my code or my environment. 

> rspec ./spec/runtime/setup_spec.rb:137 # Bundler.setup load order orders the load path correctly when there are dependencies
> rspec ./spec/commands/help_spec.rb:29 # bundle help simply outputs the txt file when there is no man on the path

I would love to hear any feedback y'all have on the PR or the failing tests. 

Thanks!

### DIFF
--- a/lib/bundler/cli/exec.rb
+++ b/lib/bundler/cli/exec.rb
@@ -63,7 +63,7 @@ module Bundler
       Kernel.load(file)
     rescue SystemExit
       raise
-    rescue Exception => e # rubocop:disable Lint/RescueException
+    rescue => e
       Bundler.ui = ui
       Bundler.ui.error "bundler: failed to load command: #{cmd} (#{file})"
       backtrace = e.backtrace.take_while {|bt| !bt.start_with?(__FILE__) }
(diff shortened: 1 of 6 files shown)


lib/bundler/cli/exec.rb
```diff
@@ -63,7 +63,7 @@ def kernel_load(file, *args)
       Kernel.load(file)
     rescue SystemExit
       raise
-    rescue Exception => e # rubocop:disable Lint/RescueException
```
dpulliam: Did we really want to ignore rubocop here?

dpulliam:
Looks like the travis build was successful. Not sure why the tests failed locally.

segiddins:
I think most of these are cases where we do want to rescue all exceptions, though

Issue #4695, opened 2016-06-21
Bundler 2.0 breaking changes


segiddins:
This is a list of breaking changes, mostly so I can ensure we have proper deprecation notices. But also generally helpful for release notes and blog posts, etc.
- Removal of `bundle_ruby` (in favor of `bundle platform --ruby`)
- Removal of `Bundler.clean_env` (in favor of `Bundler.original_env`)
- `:rubyforge`, `:gemcutter`, `:rubygems` sources
- Support for Ruby < 2.0
- Support for RubyGems < 2.4?
- global `path` sources
- multiple global gemfile sources
- [removal of `--frozen`](https://github.com/bundler/bundler/issues/4619)
- [inference of ruby version from `gemspec`](https://github.com/bundler/bundler/pull/4321)

LynnCo:
There's a (likely outdated) list of breaking changes on the Yellow board also https://trello.com/b/DioUaKgx/bundler

segiddins:
I can't get Trello to load at the moment ಠ_ಠ

LynnCo:
Thats fair, I want to delete it soon anyway. Here's that list:

- [ ] [Stop auto-remembering any command flags](https://trello.com/c/yGsPNDpg/48-stop-auto-remembering-any-command-flags)
- [ ] [Add docs to config for all current remembered flags](https://trello.com/c/Bcx15aFl/49-add-docs-to-config-for-all-current-remembered-flags)
- [ ] [Default installs to --path ./.bundle/gems](https://trello.com/c/K4zdOFl8/51-default-installs-to-path-bundle-gems)
- [ ] [Block older versions during `bundle update`](https://trello.com/c/hGXbxb0w/70-block-older-versions-during-bundle-update)
- [ ] [Change platform method: separate arch, version, and name](https://trello.com/c/3wbBxOLG/53-change-platform-method-separate-arch-version-and-name)
- [ ] [Give `gems.rb` and `gems.locked` higher priority than the Gemfile counterparts](https://trello.com/c/nzhMlstw/55-give-gems-rb-and-gems-locked-higher-priority-than-the-gemfile-counterparts)
- [ ] [Split gem server sources in the lock file](https://trello.com/c/JlXm2yZy/100-split-gem-server-sources-in-the-lock-file)
- [ ] [Reinstall gem when source changes](https://trello.com/c/kZtjlsld/80-reinstall-gem-when-source-changes)
[...]

PR #4702, opened 2016-06-22
Add machinery for printing major deprecations

segiddins:
First step towards handling all of #4695 

\c @RochesterinNYC @indirect

### DIFF
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -127,6 +127,15 @@ module Bundler
       ui.warn("[DEPRECATED FOR #{Bundler::VERSION.split(".").first.to_i + 1}.0] #{message}")
     end
 
+    def print_major_deprecations!
+      if RUBY_VERSION < "2"
+        major_deprecation("Bundler will only support ruby >= 2.0, you are running #{RUBY_VERSION}")
+      end
+      unless Bundler.rubygems.provides?(">= 2")
+        major_deprecation("Bundler will only support rubygems >= 2.4, you are running #{self.class.version}")
+      end
+    end
+
   private
 
     def find_gemfile
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -120,6 +120,13 @@ module Bundler
       namespace.const_get(constant_name)
     end
 
+    def major_deprecation(message)
+      return unless prints_major_deprecations?
+      @major_deprecation_ui ||= Bundler::UI::Shell.new("no-color" => true)
+      ui = Bundler.ui.is_a?(@major_deprecation_ui.class) ? Bundler.ui : @major_deprecation_ui
+      ui.warn("[DEPRECATED FOR #{Bundler::VERSION.split(".").first.to_i + 1}.0] #{message}")
+    end
+
   private
 
     def find_gemfile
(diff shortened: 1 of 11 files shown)


lib/bundler/shared_helpers.rb
```diff
@@ -120,6 +120,21 @@ def const_get_safely(constant_name, namespace)
...
+      ui = Bundler.ui.is_a?(@major_deprecation_ui.class) ? Bundler.ui : @major_deprecation_ui
+      ui.warn("[DEPRECATED FOR #{Bundler::VERSION.split(".").first.to_i + 1}.0] #{message}")
+    end
+
+    def print_major_deprecations!
+      if RUBY_VERSION < "2"
+        major_deprecation("Bundler will only support ruby >= 2.0, you are running #{RUBY_VERSION}")
+      end
+      return if Bundler.rubygems.provides?(">= 2")
+      major_deprecation("Bundler will only support rubygems >= 2.4, you are running #{self.class.version}")
```
indirect: I thought we said RubyGems 2.0 and above?

lib/bundler/shared_helpers.rb
```diff
@@ -207,6 +222,14 @@ def clean_load_path
       end
     end
 
+    def prints_major_deprecations?
+      require "bundler"
+      return false unless Bundler.settings[:major_deprecations]
```
indirect: I would strongly prefer that this be a constant, maybe even in `bundler/version.rb` so that we can only require the file with the constant. This isn't a setting we want to change often.

segiddins: > or rather `Bundler.settings[:major_deprecations]`
> ha
> sold

segiddins:
@indirect r?

indirect:
@homu r+

segiddins:
@homu r=indirect

segiddins:
@homu r=indirect

grosser:
Please reconsider removing clean_env ... it's not a nice api and has known drawbacks, but forcing others to re-implement it is not helping either.
There is no clean way of removing bundler from them ENV, but it is needed for some usecases like restarting an app with `exec` like puma does or shelling out with a clean environment from inside a `bundle exec` call ... 

@indirect

indirect:
@grosser why would you not use original_env instead?

grosser:
because it still includes BUNDLE_GEMFILE / RUBYLIB etc

grosser:
`bundle exec ruby -e 'puts Bundler::ORIGINAL_ENV'`

grosser:
the more common case would be `bundle exec rake` / `bundle exec puma` ... `bundle exec ruby` could be re-written to `ruby -rbundler/setup -e 'puts Bundler::ORIGINAL_ENV'` and would work ... but the others don't

indirect:
Ugh, that seems like a bug. :/ I agree that we should either fix original_env to provide access to the pre-bundler env or keep clean_env, though.

grosser:
agreed!

grosser:
... stupid idea to fix the bug: serialize ORIGINAL_ENV into `ENV['BUNDLE_ORIGINAL_ENV']` when calling bundle exec ...

segiddins:
@grosser please open a new issue to discuss this! This PR was merged a year ago and was simply porting things that were implemented on the 1-99-dev branch before

PR #4708, opened 2016-06-23
Print only version number from `bundler --version`

### LINKED ISSUE #4706: Print only version number from `bundler --version`
It's much easier for programs to consume Bundler's version number if only the version is printed. Instead of printing "Bundler version 2.0.0", let's just print "2.0.0".

(PRs for this issue should be made against the `2-0-dev` branch.)

chalkos:
Closes #4706


indirect:
Perfect, thanks! 👍 @homu r+

PR #4709, opened 2016-06-24
[Resolver] Pin transitive deps to relevant sources

segiddins:
This is a total hack but appears to work!

\c @indirect @TimMoore


indirect:
amazeballs

@homu r+

segiddins:
@indirect the tests probably won't all pass because of other issues on the branch that gets rid of the aggregate. I just want eyes in this change as it obviously has security implications

segiddins:
@indirect @TimMoore @chalkos this should now pass :D

PR #4711, opened 2016-06-24
Prefer spec name matches when searching for an exe

### LINKED ISSUE #4705: Gem.bin_path references local gem's bin directory
I believe issue #2644 may still exist within a very narrow scope.  I have the following set up:

- I'm using rbenv
- A project which uses a local gem via `bundle config local.GEM_NAME`
- binstubs generated in the project
- binstubs generated in the local gem
- I'm using Bundler version 1.12.5

When trying to run `bin/guard` I get the following error:

```
Bundler is using a binstub that was created for a different gem.
This is deprecated, in future versions you may need to `bundle binstub guard` to work around a system/bundle conflict.
```

I have verified the following:

- My PATH does not contain references to either bin directory (so `bin/guard` is indeed pointing at the correct binstub)
- The project's `bin/guard` reference the guard gem directly.  `load Gem.bin_path("guard", "guard")`
- `load Gem.bin_path("guard", "guard")` __doesn't__ reference the same directory that `bundler show guard` does.  Instead, it references the local gem's path.

segiddins:
Closes #4705.

### DIFF
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -362,19 +362,21 @@ module Bundler
     def replace_bin_path(specs)
       gem_class = (class << Gem; self; end)
 
-      redefine_method(gem_class, :find_spec_for_exe) do |name, *args|
+      redefine_method(gem_class, :find_spec_for_exe) do |gem_name, *args|
         exec_name = args.first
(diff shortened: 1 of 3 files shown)


segiddins:
@homu r+

PR #4712, opened 2016-06-24
[Definition] Add a #change_reason printed in debug

segiddins:
@indirect this might make your debugging a bit easier?

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -92,7 +92,7 @@ module Bundler
         end
         @ruby_version.diff(locked_ruby_version_object)
       end
-      @unlocking ||= @unlock[:ruby] || (!@locked_ruby_version ^ !@ruby_version)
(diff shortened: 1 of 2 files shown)


lib/bundler/definition.rb
```diff
@@ -462,6 +462,27 @@ def nothing_changed?
...
+        end.join(", ")
+        return "bundler is unlocking #{unlock_reason}"
+      end
+      [
+        [@source_changes, "the list of sources changed"],
+        [@dependency_changes, "the dependencies in your gemfile changed"],
+        [@new_platform, "you added a new platform to your gemfile"],
+        [@path_changes, "the gemspecs for path gems changed"],
+        [@local_changes, "the gemspecs for git local gems changed"],
+      ].select(&:first).map(&:last).join(", ")
```
jrafanie: Great idea @segiddins... @fryguy and I were basically doing this with puts debugging... I like making this change right in bundler so it's really obvious why we're re-resolving.  👍

indirect:
This is the best thing ever.

@homu r+

PR #4059, opened 2015-10-15
Add failing test for invalid warning

jbarreneche:
Failing test case for issue #4058

PR #4714, opened 2016-06-24
[2.0] Remove RubyGems Aggregate & support transitive source pinning

segiddins:
Closes https://github.com/bundler/bundler/issues/3671.
Closes https://github.com/bundler/bundler/pull/3696.
Closes #4059.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -668,7 +668,7 @@ module Bundler
       # Record the specs available in each gem's source, so that those
       # specs will be available later when the resolver knows where to
       # look for that gemspec (or its dependencies)
-      default = sources.rubygems_sources.last
+      default = sources.default_source
       source_requirements = { :default => default }
       dependencies.each do |dep|
         next unless source = dep.source || default
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -219,17 +219,6 @@ module Bundler
           dependency_names.push(*source.unmet_deps).uniq!
         end
 
-        memoize = proc do |&work|
-          ret = memoize
-          proc do
-            if memoize.equal?(ret)
-              ret = work.call
-            else
-              ret
-            end
-          end
-        end
-
         # Suppose the gem Foo depends on the gem Bar.  Foo exists in Source A.  Bar has some versions that exist in both
         # sources A and B.  At this point, the API request will have found all the versions of Bar in source A,
         # but will not have found any versions of Bar from source B, which is a problem if the requested version
@@ -238,8 +227,13 @@ module Bundler
         loop do
           idxcount = idx.size
           sources.all_sources.each do |source|
-            unmet_dependency_names = memoize.call do
-              dependency_names.+(idx.dependency_names).uniq unless idx.size > Source::Rubygems::API_REQUEST_LIMIT
+            names = :names # do this so we only have to traverse to get dependency_names from the index once
+            unmet_dependency_names = proc do
+              if names == :names
+                names = dependency_names.+(idx.dependency_names).uniq unless idx.size > Source::Rubygems::API_REQUEST_LIMIT
+              else
+                names
+              end
             end
             source.double_check_for(unmet_dependency_names, :override_dupes)
           end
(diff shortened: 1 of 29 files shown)


lib/bundler/definition.rb
```diff
@@ -218,15 +218,42 @@ def index
...
+        memoize = proc do |&work|
+          ret = memoize
+          proc do
+            if memoize.equal?(ret)
+              ret = work.call
+            else
+              ret
+            end
+          end
+        end
```
indirect: Is there an extremely good reason to make this a prod instead of a method? Procs are sooo much slower than method calls. Also, procs are much harder to read and understand than methods... like how i can't even tell if this has to be a prof instead of a method. 😝

spec/bundler/source_list_spec.rb
```diff
@@ -301,18 +304,17 @@
...
         Bundler::Source::Git.new("uri" => "git://third-git.org/path.git"),
         Bundler::Source::Path.new("path" => "/first/path/to/gem"),
         Bundler::Source::Path.new("path" => "/second/path/to/gem"),
         Bundler::Source::Path.new("path" => "/third/path/to/gem"),
-        Bundler::Source::Rubygems.new("remotes" => [
-          "https://duplicate-rubygems.org",
-          "https://first-rubygems.org",
-          "https://second-rubygems.org",
-          "https://third-rubygems.org",
-        ]),
```
indirect: If sources are searched from top to bottom, doesn't this mean that rubygems.org will not get prioritized over path/git gems that contain indirect dependencies? eg `gem "rails", github: "rails"` should use AR from git, not from RubyGems.org (and it shouldn't raise an error, I don't think.)

segiddins: this is only the order used when making the lockfile. I can add a spec for the rails case though!

lib/bundler/definition.rb
segiddins: I can actually inline it entirely -- I don't like methods because the can be used elsewhere :P

indirect: inline is fine, but you should get used to private methods :P

spec/bundler/source_list_spec.rb
indirect: Sounds good! I thought we had a spec for that type of case, but I'm not totally sure.

segiddins:
@indirect addressed feedback

spec/install/gemfile/sources_spec.rb
```diff
@@ -226,6 +226,122 @@
...
+                                          "actionpack 2.3.2" => "rails-git",
+                                          "activerecord 2.3.2" => "rails-git",
+                                          "activeresource 2.3.2" => "rails-git",
+                                          "activesupport 2.3.2" => nil,
+                                          "rails 2.3.2" => "rails-git",
+                                          "rake 10.0.2" => nil)
+        end
+      end
+    end
+
```
indirect: :100:

lib/bundler/dsl.rb
```diff
@@ -146,6 +146,16 @@ def git_source(name, &block)
...
+      unless block_given?
+        msg = "You can no longer specify a path source by itself. Instead, \n" \
+              "either use the :path option on a gem, or specify the gems that \n" \
+              "bundler should find in the path source by passing a block to \n" \
+              "the path method, like: \n\n" \
+              "  path 'dir/containing/rails' do\n" \
+              "    gem 'rails'\n" \
+              "  end"
+        raise DeprecatedError, msg
+      end
```
segiddins: yeah, it does what git does and uses and empty block to fetch the source when used as an option: https://github.com/bundler/bundler/blob/4327cfe160f45f1bb0f02b9ea8b7e006ec255e02/lib/bundler/dsl.rb#L343

indirect: 👍

indirect:
@homu r+

indirect:
Locally I get a failure at `./spec/install/gems/resolving_spec.rb:6`, when running on 68d2ba3. Maybe try `rm -rf tmp; rspec ./spec/install/gems/resolving_spec.rb:6`?

segiddins:
@indirect I can't repro the issue locally _at all_

segiddins:
@homu r=indirect

segiddins:
🎉🎉🎉🎉

PR #4715, opened 2016-06-24
Don't incorrectly relativize sibling `--path` with the same prefix

jenseng:
Given current working directory `/foo/app`, when you install with
`--path /foo/app_cache`, ensure that the messaging says:

    Bundled gems are installed into /foo/app_cache

instead of the incorrect:

    Bundled gems are installed into ._cache


indirect:
Oof. Thanks for the fix!

segiddins:
@homu r+

PR #4716, opened 2016-06-24
[Definition] Just search for changed specs in an exactly matching source

segiddins:
\c @indirect


indirect:
This fixes things for me! 🏆 @homu r+

PR #4720, opened 2016-06-25
Print an outdated warning when a project is locked to an old version of Bundler

### LINKED ISSUE #4707: Print an outdated warning (once) when a project is locked to an old version of Bundler
As of Bundler 1.13.pre.1, just installing new versions with `gem install bundler` does not mean that the new version will be used when you run `bundle install`. Instead, the version of Bundler listed in the Gemfile.lock will be used. To update the version of Bundler in the Gemfile.lock, and start using the newest version of Bundler possible, users need to run `bundle update --bundler`.

When a user runs `bundle install` on an application that is locked to an older version of Bundler than the version that is running, we want to print a warning like this:

> You're running Bundler 1.13.0 but this project uses 1.12.5. To update, run `bundle update --bundler`. You won't see this message again unless you upgrade to a newer version of Bundler.

When we print that message, we should save a local setting, with code something like `Bundler.settings[:warned_version] = Bundler::VERSION`. Then, we should only print the warning when `Bundler.settings[:warned_version].nil? || Bundler::VERSION > Bundler.settings[:warned_version]`.

b-ggs:
@indirect @RochesterinNYC 

I'm a bit stumped.

The original issue on bundler/bundler#4707 mentions that this warning should only be printed once unless the user updates their version of Bundler, and that the warned version should be stored in `Bundler.settings[:warned_version]`.

The issue is that, AFAIK, this file, `postit_trampoline.rb` is the last place where I can access both the `installed_version` (system version) and the `running_version` (project's version) of Bundler before it completely switches over to the `running_version`.

Since `Bundler.settings` still hasn't been initialized at this point, I can't store the `running_version` in `Bundler.settings[:warned_version]`.

Right now, this warning is shown every time a `bundle` command is issued in a project directory where `installed_version != running_version`.

Is there any other way I can store the warned version from this file?

---

Closes #4707

### DIFF
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -11,6 +11,8 @@ module Bundler
 
       warn_if_root
 
+      warn_if_outdated
+
       [:with, :without].each do |option|
         if options[option]
           options[option] = options[option].join(":").tr(" ", ":").split(":")
@@ -112,6 +114,19 @@ module Bundler
       Bundler.ui.warn "Don't run Bundler as root. Bundler can ask for sudo " \
         "if it is needed, and installing your bundle as root will break this " \
         "application for all non-root users on this machine.", :wrap => true
+    end
+
+    def warn_if_outdated
+      return if ENV["BUNDLE_POSTIT_TRAMPOLINING_VERSION"].nil?
+      installed_version = Gem::Version.new(ENV["BUNDLE_POSTIT_TRAMPOLINING_VERSION"])
+      running_version = Gem::Version.new(Bundler::VERSION)
+      if Bundler.settings[:warned_version].nil? || running_version > Gem::Version.new(Bundler.settings[:warned_version])
+        Bundler.settings[:warned_version] = running_version
+        Bundler.ui.warn "You're running Bundler #{installed_version} but this " \
+          "project uses #{running_version}. To update, run `bundle update " \
+          "--bundler`. You won't see this message again unless you upgrade " \
+          "to a newer version of Bundler.", :wrap => true
+      end
     end
 
     def confirm_without_groups
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -118,7 +118,7 @@ module Bundler
 
     def warn_if_outdated
       return if ENV["BUNDLE_POSTIT_TRAMPOLINING_VERSION"].nil?
-      installed_version = Gem::Version.new(ENV["BUNDLE_POSTIT_TRAMPOLINING_VERSION"])
+      installed_version = Gem::Version.new(ENV["BUNDLE_POSTIT_TRAMPOLINING_VERSION"].dup)
       running_version = Gem::Version.new(Bundler::VERSION)
       if Bundler.settings[:warned_version].nil? || running_version > Gem::Version.new(Bundler.settings[:warned_version])
         Bundler.settings[:warned_version] = running_version
--- a/lib/bundler/postit_trampoline.rb
+++ b/lib/bundler/postit_trampoline.rb
@@ -51,10 +51,9 @@ end
 
 if Gem::Requirement.new(">= 1.13.pre".dup).satisfied_by?(Gem::Version.new(running_version))
   ENV["BUNDLE_POSTIT_TRAMPOLINING_VERSION"] = installed_version.to_s
-elsif ARGV.any? {|a| %w(install i).include? a }
-  # TODO: Change this to warn in Bundler 2.0
+elsif ARGV.empty? || ARGV.any? {|a| %w(install i).include? a }
   puts <<-WARN.strip
-You're running Bundler #{installed_version} but this project uses #{running_version}. To update, run `bundle update --bundler`.\n
+You're running Bundler #{installed_version} but this project uses #{running_version}. To update, run `bundle update --bundler`.
   WARN
 end
 
--- a/lib/bundler/postit_trampoline.rb
+++ b/lib/bundler/postit_trampoline.rb
@@ -49,7 +49,7 @@ rescue LoadError, NameError
   nil
 end
 
-if Gem::Version.new(running_version) >= Gem::Version.new("1.13.0.rc.1".dup)
+if Gem::Requirement.new(">= 1.13.pre".dup).satisfied_by?(Gem::Version.new(running_version))
   ENV["BUNDLE_POSTIT_TRAMPOLINING_VERSION"] = installed_version.to_s
 elsif ARGV.any? {|a| %w(install i).include? a }
   # TODO: Change this to warn in Bundler 2.0
--- a/lib/bundler/postit_trampoline.rb
+++ b/lib/bundler/postit_trampoline.rb
@@ -49,9 +49,10 @@ rescue LoadError, NameError
   nil
 end
 
-if Gem::Version.new(running_version) >= Gem::Version.new("1.13.0.pre.1")
+if Gem::Version.new(running_version) >= Gem::Version.new("1.13.0.pre.1".dup)
   ENV["BUNDLE_POSTIT_TRAMPOLINING_VERSION"] = installed_version.to_s
 elsif ARGV.any? {|a| %w(install i).include? a }
+  # TODO: Change this to warn in Bundler 2.0
   puts <<-WARN.strip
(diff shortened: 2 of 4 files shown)


segiddins:
@b-ggs maybe stick it in an ENV var in the trampoline, `BUNDLE_POSTIT_TRAMPOLINING_VERSION`?

b-ggs:
@segiddins It doesn't save to the actual machine's `env` vars though? Every new `bundle` command still evaluates `ENV["BUNDLE_POSTIT_TRAMPOLINING_VERSION"]` as `nil`. Need it to persist over future `bundle` calls.

segiddins:
It'd let the actual bundler code access settings though and know that it had trampolined

b-ggs:
@segiddins @indirect I added different warning methods for both pre and post-`1.13.0.pre.1` Bundler versions.

For post-`1.13.0.pre.1`, I save the `running_version` of bundler in an `ENV` var, and let `CLI:Install` warn the user once that they're running an outdated version of Bundler. After which, `running_version` is saved into `Bundler.settings[:warned_version]`, and the warning will only be shown again if `running_version > Bundler.settings[:warned_version]`.

For pre-`1.13.0.pre.1`, since I can't change `CLI:Install` because it will trampoline to an older version of Bundler which doesn't have my warning implementation, it will always show the warning before trampolining over to an older version of Bundler.

What do you think of this?

b-ggs:
Welp, apparently putting the warning on everything broke some of the specs. Limited the warning to `bundle install` or `bundle i`.

lib/bundler/cli/install.rb
```diff
@@ -114,6 +116,20 @@ def warn_if_root
         "application for all non-root users on this machine.", :wrap => true
     end
 
+    def warn_if_outdated
+      return if ENV["BUNDLE_POSTIT_TRAMPOLINING_VERSION"].nil?
+      installed_version = ENV["BUNDLE_POSTIT_TRAMPOLINING_VERSION"]
+      installed_version = Gem::Version.new(installed_version)
```
RochesterinNYC: what's the reasoning behind this double assignment?

RochesterinNYC:
I'm not understanding something here, if this change/PR will be merged into master and released as part of `1.13.x`, how would the user ever have this PR's functionality and the `>= 1.13.0` checking code in their running Bundler unless they're running a Bundler that's >= `1.13.0`? Isn't it impossible for the user to have the `>= 1.13.0` check code in their running Bundler and have that `current running bundler version < 1.13.0` constraint apply to them?

b-ggs:
@RochesterinNYC Impossible how? If Bundler is set to trampoline to a version older than `1.13.0.pre.1`, it shows this warning before it jumps: https://github.com/bundler/bundler/pull/4720/files#diff-969f16d17fb6c5af3568a1d0391978bbR55 However, if it's set to trampoline to something newer than `1.13.0.pre.1`, it uses the warning implementation I've put in place in `CLI:Install`.

lib/bundler/cli/install.rb
b-ggs: This was an attempt to fix breaking specs on 1.9.3. I'll be removing the commit that did this in a few hours when I arrive at uni.

RochesterinNYC:
@b-ggs Okay, the piece of context I was missing is what "trampolining" is. Read up on this new feature and this makes sense to me now.

lib/bundler/postit_trampoline.rb
```diff
@@ -50,6 +50,14 @@
   nil
 end
 
+if Gem::Version.new(running_version) >= Gem::Version.new("1.13.0.pre.1")
+  ENV["BUNDLE_POSTIT_TRAMPOLINING_VERSION"] = installed_version.to_s
+elsif ARGV.any? {|a| %w(install i).include? a }
+  puts <<-WARN.strip
```
segiddins: please add a TODO here to change this to warn in bundler 2.0

spec/commands/install_spec.rb
```diff
@@ -485,4 +485,19 @@
                         "setting them for authentication.")
     end
   end
+
+  describe "warns user if Bundler is outdated" do
+    it "warns only once and is > 1.13.0.pre.1" do
+      gemfile <<-G
+        source "file://#{gem_repo1}"
+      G
+      ENV["BUNDLER_VERSION"] = "1.13.0.pre.1"
```
segiddins: please have this use `Bundler::VERSION`

lib/bundler/postit_trampoline.rb
```diff
@@ -50,6 +50,14 @@
   nil
 end
 
+if Gem::Version.new(running_version) >= Gem::Version.new("1.13.0.pre.1")
```
segiddins: this needs a `.dup` for ooooold rubygems versions

b-ggs:
@segiddins @RochesterinNYC Rebased from `master` and updated the version number to check against from `1.13.0.pre.1` to `1.13.0.rc.1`.

lib/bundler/postit_trampoline.rb
```diff
@@ -49,6 +49,15 @@ module BundlerVendoredPostIt; end
   nil
 end
 
+if Gem::Version.new(running_version) >= Gem::Version.new("1.13.0.rc.1".dup)
```
indirect: Pretty sure this should be `if Gem::Requirement.new(">= 1.13.pre").satisfied_by?(Gem::Version.new(running_version))`.

b-ggs:
@segiddins I take that the codeclimate issue will be ignored since it's just a TODO line?

RochesterinNYC:
I don't think we need to keep the `TODO` comment in. Feel free to create an issue for it if we're really apprehensive about forgetting to fix or address that `TODO`.

segiddins:
@indirect r?

lib/bundler/postit_trampoline.rb
```diff
@@ -49,6 +49,15 @@ module BundlerVendoredPostIt; end
   nil
 end
 
+if Gem::Requirement.new(">= 1.13.pre".dup).satisfied_by?(Gem::Version.new(running_version))
+  ENV["BUNDLE_POSTIT_TRAMPOLINING_VERSION"] = installed_version.to_s
+elsif ARGV.any? {|a| %w(install i).include? a }
+  # TODO: Change this to warn in Bundler 2.0
+  puts <<-WARN.strip
```
segiddins: the `strip` and the `\n` at the end seem at odds

lib/bundler/postit_trampoline.rb
```diff
@@ -49,6 +49,15 @@ module BundlerVendoredPostIt; end
   nil
 end
 
+if Gem::Requirement.new(">= 1.13.pre".dup).satisfied_by?(Gem::Version.new(running_version))
+  ENV["BUNDLE_POSTIT_TRAMPOLINING_VERSION"] = installed_version.to_s
+elsif ARGV.any? {|a| %w(install i).include? a }
```
segiddins: this doesn't handle the case of just calling `bundle`

lib/bundler/postit_trampoline.rb
```diff
@@ -49,6 +49,14 @@ module BundlerVendoredPostIt; end
   nil
 end
 
+if Gem::Requirement.new(">= 1.13.pre".dup).satisfied_by?(Gem::Version.new(running_version))
+  ENV["BUNDLE_POSTIT_TRAMPOLINING_VERSION"] = installed_version.to_s
+elsif ARGV.empty? || ARGV.any? {|a| %w(install i).include? a }
```
b-ggs: @segiddins It was passing fully previously (the only fail for the previous check was codeclimate because of the TODO). This is what made it fail. I was trying to add support for `bundle` command only. It broke a completely different spec, so I'm looking for another way to determine whether the user just runs `bundle` without relying on the `ARGV`.

b-ggs:
@segiddins The failure was caused by an outdated spec that simulated Bundler 1.12.0. The spec was expecting `out` to `eq(Bundler::VERSION)`, but `out` also has the warning message added in this PR.

I modified it to expect `out` to `include(Bundler::VERSION)` instead.

This is the only failure in the previous run, so this should be green and good to go.

segiddins:
@homu delegate=RochesterinNYC

RochesterinNYC:
@homu r+

PR #4727, opened 2016-06-27
Fixes INT signal for ruby executables

### LINKED ISSUE #4568: bundle exec signal trapping regression from v1.11.2 to v1.12.0
There seems to be a regression between bundler version 1.11.2 and 1.12.0 (also the latest version).

Easy way to reproduce:

Gemfile:
```rb
source 'https://rubygems.org'

ruby '2.3.0'

gem 'puma'
```

config.ru:
```rb
run -> env { [200, { 'Content-Type' => 'text/plain' }, ['pong']] }
```

Working scenario:
```sh
$ gem install bundler -v 1.11.2
$ bundle
$ bundle exec puma
Puma starting in single mode...
* Version 3.4.0 (ruby 2.3.0-p0), codename: Owl Bowl Brawl
* Min threads: 0, max threads: 16
* Environment: development
* Listening on tcp://0.0.0.0:9292
Use Ctrl-C to stop
^C- Gracefully stopping, waiting for requests to finish
=== puma shutdown: 2016-05-12 11:27:51 +0200 ===
- Goodbye!
$ echo $?
0
```

Not working scenario:
```sh
$ gem install bundler -v 1.12.0
$ bundle
$ bundle exec puma
Puma starting in single mode...
* Version 3.4.0 (ruby 2.3.0-p0), codename: Owl Bowl Brawl
* Min threads: 0, max threads: 16
* Environment: development
* Listening on tcp://0.0.0.0:9292
Use Ctrl-C to stop
^C$ echo $?
1
```

asutoshpalai:
Fixes #4568

### DIFF
--- a/lib/bundler/cli/exec.rb
+++ b/lib/bundler/cli/exec.rb
@@ -60,6 +60,7 @@ module Bundler
       ui = Bundler.ui
       Bundler.ui = nil
       require "bundler/setup"
+      trap("INT", "DEFAULT")
       Kernel.load(file)
     rescue SystemExit
       raise
--- a/lib/bundler/cli/exec.rb
+++ b/lib/bundler/cli/exec.rb
@@ -60,7 +60,7 @@ module Bundler
       ui = Bundler.ui
       Bundler.ui = nil
       require "bundler/setup"
-      trap("INT", "DEFAULT")
+      Signal.list.each {|signal| trap(signal, "DEFAULT") }
       Kernel.load(file)
     rescue SystemExit
       raise
--- a/lib/bundler/cli/exec.rb
+++ b/lib/bundler/cli/exec.rb
@@ -5,6 +5,8 @@ module Bundler
   class CLI::Exec
     attr_reader :options, :args, :cmd
 
+    RESERVED_SIGNALS = %w(SEGV BUS ILL FPE VTALRM KILL STOP).freeze
+
     def initialize(options, args)
       @options = options
       @cmd = args.shift
@@ -60,7 +62,8 @@ module Bundler
       ui = Bundler.ui
       Bundler.ui = nil
       require "bundler/setup"
-      Signal.list.each {|signal| trap(signal, "DEFAULT") }
+      signals = Signal.list.keys - RESERVED_SIGNALS
+      signals.each {|s| trap(s, "DEFAULT") }
       Kernel.load(file)
     rescue SystemExit
       raise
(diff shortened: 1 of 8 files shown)


lib/bundler/cli/exec.rb
```diff
@@ -60,6 +60,7 @@ def kernel_load(file, *args)
       ui = Bundler.ui
       Bundler.ui = nil
       require "bundler/setup"
+      Signal.list.each {|signal| trap(signal, "DEFAULT") }
```
segiddins: `Signal.list` is a hash, so could that be `|signal, _|`?

asutoshpalai: Yea! My bad!

asutoshpalai:
Is it fine to include the spec file in `exempt` list of `quality_spec` for `debug` mechanism?
The `sleep` is firing it off.

The alternative to `sleep` is an infinite loop with the risk that if the specs fail we will be left with a running ruby process on our computer even after `rspec` exits.

segiddins:
@asutoshpalai apply the following and add the magic comment at the end of the line with the sleep

```diff
diff --git a/spec/quality_spec.rb b/spec/quality_spec.rb
index d5178d7..57e737c 100644
--- a/spec/quality_spec.rb
+++ b/spec/quality_spec.rb
@@ -28,7 +28,9 @@ describe "The library itself" do
 
     failing_lines = []
     File.readlines(filename).each_with_index do |line, number|
-      failing_lines << number + 1 if line =~ debugging_mechanisms_regex
+      if line =~ debugging_mechanisms_regex && !line.end_with?("# ignore quality_spec")
+        failing_lines << number + 1
+      end
     end
 
     return if failing_lines.empty?
```

segiddins:
@asutoshpalai this seems to break some stuff on 1.8.7

asutoshpalai:
The variable arguments to yield are breaking some specs in 1.8.7. Will fix those specs with dummy arguments.

But the major problem is on 1.8.7 `popen3` doesn't have any method to obtain the `pid`, which in turn means we can't send any signal to the process. So the new spec can't work on 1.8.7.

segiddins:
that's fine, you can limit the spec to being ruby >= 1.9.3

spec/commands/exec_spec.rb
```diff
@@ -544,5 +544,32 @@ def bin_path(a,b,c)
...
+            STDOUT.flush
+            sleep 1 # ignore quality_spec
+            raise "Didn't receive INT at all"
+          end.join
+        rescue Interrupt
+          puts "foo"
+        end
+      RUBY
+
+      it do
```
segiddins: The spec needs a name

segiddins:
👍 other than the one comment

segiddins:
@homu r+

PR #4729, opened 2016-06-27
[WIP]Add pristine command and spec

allenzhao:
- [ ] [Spec] Add spec for pristine all gems
- [ ] [Spec] Add spec for pristine only a list of gems
- [ ] [Spec] Add spec for pristine skip a list of gems
- [ ] [Spec] Add spec for pristine extensions only
- [ ] [Spec] Add spec for pristine no-extensions only
- [ ] [Spec] Add spec for handling git-source gems
- [ ] [Logic] Add logic for handling git-source gems
- [x] [Logic] Add logic for handling path-source gems
- [x] [Logic] Add logic to compute gem list to run `gem pristine`
- [ ] [Logic] Add logic to handle extension related gems

### DIFF
--- a/lib/bundler/cli/pristine.rb
+++ b/lib/bundler/cli/pristine.rb
@@ -1,0 +1,69 @@
+# frozen_string_literal: true
+module Bundler
+  module CLI::Pristine
+    class ConflictingGems < PristineError; end
+    class AmbiguousOption < PristineError; end
+    attr_reader :options, :gem_list
+    def initialize(options, gem_list)
+      @options = options
+      @gem_list = gem_list
+    end
+
+    def run
+      # Determine if -extensions and --no-extensions or  are conflict
+      if options["extensions"] && options["no-extensions"]
+        Bundler.ui.error "You can't use --extensions and --no-extensions at the same time"
+        raise AmbiguousOption
+      end
+
+      # Raise error when Gemfile is not present
+      definition = begin
+        Bundler.definition
+      rescue GemfileNotFound
+        Bundler.ui.error "You can't pristine without a Gemfile. Please consider add Gemfile then run `bundle install` before pristine"
+        raise
+      end
+      # Raise error when lockfile is not present
+      unless Bundler.default_lockfile.exist?
+        Bundler.ui.error "You can't pristine without a Gemfile.lock. Please consider run `bundle install` before pristine"
+        raise GemfileLockNotFound, "Could not locate Gemfile.lock"
+      end
+
+      # Find the path gems and exclude them from calling gem pristine
+      path_gems, git_gems, pristine_gems = []
+      skip_gems = options[:skip]
+      # If gem list is empty, we need to add everything to pristine_gems, else it equals
+      if gem_list.empty?
+        pristine_gems = definition.calculate_full_gem_list
+      else
+        unless skip_gems.empty?
+          # Determine if there are any conflict groups
+          check_conflicting(gem_list, skip_gems)
+          pristine_gems = gem_list - skip_gems
+        end
+      end
+
+      if definition.any_path_sources?
+
+      end
+      # Find git sources gems and
+      git_sources = definition.git_sources if definition.any_git_sources?
+    end
+
+  private
+
+    # This method is used to compute what are the gems
+    def compute_gem_list(definition)
+      gem_list = definition if @gem_list.empty?
+    end
+
+    def check_conflicting(gem_list, skip_gems)
+      conflicting_gems = skip_gems & gem_list
+      unless conflicting_gems.empty?
+        Bundler.ui.error "You can't list a gem in both GEMLIST and --skip." \
+          "The offending gems are: #{conflicting_gems.join(", ")}."
+        raise ConflictingGems
+      end
+    end
+  end
+end
--- a/lib/bundler/cli/pristine.rb
+++ b/lib/bundler/cli/pristine.rb
@@ -1,9 +1,13 @@
 # frozen_string_literal: true
+require "bundler/cli/exec"
 module Bundler
-  module CLI::Pristine
-    class ConflictingGems < PristineError; end
-    class AmbiguousOption < PristineError; end
+  class CLI::Pristine
+    class ConflictingGems < PristineError
+    end
+    class AmbiguousOption < PristineError
+    end
     attr_reader :options, :gem_list
+
     def initialize(options, gem_list)
       @options = options
       @gem_list = gem_list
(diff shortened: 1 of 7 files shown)


lib/bundler/cli/pristine.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+module Bundler
+  class CLI::Pristine
```
segiddins: please nest the class inside the namespace

lib/bundler/cli/pristine.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+
+    def run
+      # Determine if there are any conflict groups
+      if !gem_list.empty? && !options[:skip].empty?
+        conflicting_gems = options[:skip] & gem_list
+        unless conflicting_gems.empty?
+          Bundler.ui.error "You can't list a gem in both GEMLIST and --skip." \
+          "The offending gems are: #{conflicting_gems.join(", ")}."
+          exit 1
```
segiddins: should raise an error instead of exiting

lib/bundler/cli/pristine.rb
```diff
@@ -0,0 +1,57 @@
...
+        end
+      end
+      # Determine if -extensions and --no-extensions or  are conflict
+      if options[:extensions] && options["no-extensions"]
+        Bundler.ui.error "You can't use --extensions and --no-extensions at the same time"
+        exit 1
```
segiddins: should raise an error instead of exiting

lib/bundler/cli/pristine.rb
```diff
@@ -0,0 +1,57 @@
...
+      end
+
+      definition = Bundler.definition
```
segiddins: this will already raise

lib/bundler/cli/pristine.rb
```diff
@@ -0,0 +1,57 @@
...
+      # Raise error when Gemfile is not present
+      begin
+        Bundler.default_gemfile
+      rescue GemfileNotFound
+        Bundler.ui.error "You can't pristine without a Gemfile. Please consider add Gemfile then run `bundle install` before pristine"
+        raise
+      end
+      # Raise error when lockfile is not present
+      unless Bundler.default_lockfile.exist?
```
segiddins: why can this run install implicitly and just pristine the installed gems?

spec/commands/pristine_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  context "when both the Gemfile and the Gemfile.lock is present" do
+    before :each do
+      install_gemfile <<-G
+        source "file://#{gem_repo1}"
+        gem "rack", "1.0.0"
+        gem "foo", "1.2.3"
+      G
+    end
+  end
+end
```
segiddins: missing trailing newline

lib/bundler/cli/pristine.rb
allenzhao: Didn't I already done the nesting?

segiddins: never mind, this seems to be how the other CLI classes do it

lib/bundler/cli/pristine.rb
allenzhao: I thought it should be the case that the user installs dependencies via bundler then change it, and then want to pristine so he/she could "revert back" to the original state. 
But on a second thought, it makes sense that the user changes local installed gems first then want to pristine.
I think a warning and ask if the user want to let us do the `bundle install` for him/her should be more flexible?

lib/bundler/errors.rb
```diff
@@ -42,6 +42,7 @@ class LockfileError < BundlerError; status_code(20); end
   class CyclicDependencyError < BundlerError; status_code(21); end
   class GemfileLockNotFound < BundlerError; status_code(22); end
   class PluginError < BundlerError; status_code(23); end
+  class PristineError < BundlerError; end
```
segiddins: needs a status code

lib/bundler/cli/pristine.rb
```diff
@@ -0,0 +1,85 @@
...
+      Bundler.ui.error "You can't list a gem in both GEMLIST and --skip." \
+          "The offending gems are: #{conflicting_gems.join(", ")}."
+      raise ConflictingGems
+    end
+  end
+
+  def pristine(gems, git_gems = nil, lazy_spec_provided = false)
+    gem_list = lazy_spec_provided ? gems.map(&:name) : gems
+    check_conflicting(gem_list, options[:skip])
+    pristine_gems = gem_list - options[:skip]
```
allenzhao: @RochesterinNYC 
To make line 53 & this line work I need to implement a "-" and a "&" on `SpecSet` since `gem_list` here could be an Array or an instance of `SpecSet` and options[:skip] would always be array. But I don't know if this would be somehow "bad practice" ? What's your opinion?

lib/bundler/cli/pristine.rb
allenzhao: Any thoughts on this?
@indirect @RochesterinNYC

lib/bundler/cli/pristine.rb
RochesterinNYC: Casting and then performing the operations could be a possibility? Why can `gem_list` be either an array or a SpecSet instead of always one or the other?

allenzhao: When it's not provided by the user, it's from the definition's locked_spec so it is a SpecSet and using a SpecSet rather than an array could provide the version information

RochesterinNYC: I don't know that a `-` and `&` method on the actual `SpecSet` itself is the right solution. Maybe write a method that handles the casting and comparison and then at the end, we can gauge whether it really belongs on the class itself?

segiddins:
@allenzhao what's the status on this?

lib/bundler/cli/pristine.rb
```diff
@@ -0,0 +1,98 @@
...
+      Bundler.ui.error "You can't list a gem in both GEMLIST and --skip." \
+          "The offending gems are: #{conflicting_gems.join(", ")}."
+      raise ConflictingGems
+    end
+
+    def pristine(gems, git_gems = nil, lazy_spec_provided = false)
+      gem_list = lazy_spec_provided ? gems.map(&:name) : gems
+      skip_gems = options[:skip]
+      pristine_gems = compute_pristine_gems(gem_list, skip_gems)
+      binding.pry
```
ruba-ruba: pry

allenzhao: Yes, it's still a WIP.

allenzhao:
@segiddins I haven't been able to pick up these PRs. I will manage to start working on them this week, but first I think I need some time to follow up on the current status of bundler dev.

segiddins:
Closed by #5503.

PR #4730, opened 2016-06-27
Fix spurious spec failures due to leaking Bundler.ui

segiddins:
Was leaking through the GemHelper unit tests


segiddins:
@homu r+

PR #4735, opened 2016-06-29
Make `bundler` and `bundle` executables have the same functionality

### LINKED ISSUE #4725: `bundler` and `bundle` executables should be the same functionally
The `bundler` executable (https://github.com/bundler/bundler/blob/master/exe/bundler) should functionally be the same as the `bundle` executable (https://github.com/bundler/bundler/blob/master/exe/bundle).

Ideally, this would be done in a way that doesn't involve duplication of code and doesn't require updating code in two places when a change is made to the `bundle` executable.

b-ggs:
Closes #4725

### DIFF
--- a/exe/bundler
+++ b/exe/bundler
@@ -1,22 +1,4 @@
 #!/usr/bin/env ruby
 # frozen_string_literal: true
 
-# Exit cleanly from an early interrupt
-Signal.trap("INT") { exit 1 }
-
-require "bundler"
-# Check if an older version of bundler is installed
-$LOAD_PATH.each do |path|
-  next unless path =~ %r{/bundler-0\.(\d+)} && $1.to_i < 9
-  err = String.new
-  err << "Looks like you have a version of bundler that's older than 0.9.\n"
-  err << "Please remove your old versions.\n"
-  err << "An easy way to do this is by running `gem cleanup bundler`."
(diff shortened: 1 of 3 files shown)


RochesterinNYC:
@segiddins @indirect quick sanity check, do we have existing test coverage for the `bundler` executable? Can't seem to find any from a preliminary search of the specs.

segiddins:
Maybe not? We should add a quick test that just does `bundler --version` or something?

b-ggs:
@segiddins @RochesterinNYC Added a spec for `bundler --version`!

b-ggs:
I've force-pushed this branch several times, all resulting in an errored check from Travis, which I highly doubt was caused by the code I've added. Are we currently experiencing some weird issues with Travis at the moment?

segiddins:
@b-ggs do the specs pass locally for you?

RochesterinNYC:
Looks like flakey behavior on Travis's part. Restarted the build.

segiddins:
@b-ggs this should pass if you rebase it

b-ggs:
@segiddins Gotcha. Rebased from `master`. Tests are running now. And yeah, it was passing for me locally.

segiddins:
@homu r+

PR #4740, opened 2016-07-01
[CLI] Print running command

segiddins:
This also has the nice side-effect of printing the bundler version ;)

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -16,6 +16,10 @@ module Bundler
       Bundler::SharedHelpers.print_major_deprecations!
     end
 
+    def self.dispatch(*)
+      super {|i| i.send(:print_command) }
+    end
+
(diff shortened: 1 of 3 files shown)


indirect:
@segiddins I think we only want to do this post-trampoline? or do you want to do it all the time?

segiddins:
Why not all the time? It makes the output more helpful when running with aliases or bundler from a script

indirect:
The CLI ethos is generally to do the work you asked for and only print information when it's unexpected or unknown. Because of that, I don't think we should echo the command you just ran back to STDOUT unless either 1) trampolining occurred or 2) --verbose mode is enabled.  FWIW, this would would resolve the second half of #4753 nicely.

segiddins:
@indirect done

indirect:
@homu r+

Issue #4540, opened 2016-05-09
Need to be able to push signed tags


lynnco:
_From @todb-r7 on June 23, 2015 17:12_

Currently, `rake release` pushes a version tag up to github in an unsigned way. Thus, it gets difficult to say with any certainty who pushed a tag on a project with multiple committers.

I couldn't find anything in the documentation about an environment variable or other option where I can specify a signing key for the version tag. Note, I'm not looking specifically for signing a **gem** (with a PEM certificate), I'm just looking to sign a **tag commit**.

All about signing commits: http://mikegerwitz.com/papers/git-horror-story

If I've just missed this in the docs, an RTFM with a pointer would help a ton.

Thanks!

_Copied from original issue: bundler/bundler-features#90_

LynnCo:
_From @indirect on June 23, 2015 17:18_

Bundler does not support signing commits or releases--if you would like to do so, I suggest overriding the Rake task for releasing.

LynnCo:
_From @todb-r7 on June 23, 2015 17:22_

> Bundler does not support signing commits or releases--if you would like to do so, I suggest overriding the Rake task for releasing.

Ah good, then this is a proper feature request then, and not merely a noob question.

I'd argue that it's important that Bundler allow people to do the secure thing as simple config option. Ideally, as a strongly encouraged option.

LynnCo:
_From @indirect on June 23, 2015 18:10_

Thanks for the feature request! You're the first person to ever request that Bundler support signing release tags, which is why I suggested that you could overwrite the Rake task to get the functionality you're asking for.

The Bundler gem functionality is deliberately designed to only cover the most common use-cases, and today signing commits is not a common use-case. We'll take this request under consideration for the future, but at present it's unlikely.

LynnCo:
_From @todb-r7 on June 23, 2015 18:25_

>  signing commits is not a common use-case

I bet if Bundler supported it, it'd become common. :)

LynnCo:
_From @indirect on June 23, 2015 18:29_

The Bundler gem functionality has, so far, merely implemented the most common practice for gems, rather than dictating what those practices should be. :)

LynnCo:
_From @Mange on December 16, 2015 7:57_

It should be mentioned that creating a signed tag is the "default" for `git`, and that it's mostly Github that has created the current trend of not signing tags.

Would a pull request be welcome, if I did one?

LynnCo:
[...]

PR #4743, opened 2016-07-02
Allow users to sign the version tag

eagletmt:
This patch implements #4540 .

### DIFF
--- a/lib/bundler/gem_helper.rb
+++ b/lib/bundler/gem_helper.rb
@@ -140,7 +140,7 @@ module Bundler
     end
 
     def tag_version
-      sh "git tag -a -m \"Version #{version}\" #{version_tag}"
+      sh "git tag -m \"Version #{version}\" #{version_tag}"
       Bundler.ui.confirm "Tagged #{version_tag}."
       yield if block_given?
     rescue


spec/bundler/gem_helper_spec.rb
```diff
@@ -253,6 +254,21 @@ def remove_push_guard(gemspec_content)
...
+        it "when user.signingkey is present" do
+          mock_build_message app_name, app_version
+          mock_confirm_message "Tagged v#{app_version} signed with ABCDEF01."
+          mock_confirm_message "Pushed git commits and tags."
+          expect(subject).to receive(:rubygem_push).with(app_gem_path.to_s)
+
+          Dir.chdir(app_path) { sys_exec("git push -u origin master", true) }
+
+          Dir.chdir(app_path) do
+            sys_exec("git config gpg.program #{File.expand_path("../../support/fake-gpg", __FILE__)}")
```
segiddins: could this use `Bundler.which("echo")`?

segiddins:
👍 other than the one tiny comment

spec/bundler/gem_helper_spec.rb
eagletmt: Fixed in https://github.com/bundler/bundler/pull/4743/commits/aae87129f60052bb45156474295db31aea015164

segiddins:
@indirect r?

rafaelfranca:
`user.signingkey` is not what usually used to know if a commit needs a sign or not. I believe `commit.gpgsign` is more accurate with the intentions to sign a tag. We could also use `tag.forceSignAnnotated` and just let git use its default behavior (that will be creating a annotated and signed that) not passing the `-a` option.

indirect:
Git [uses this setting to sign tags already](http://stackoverflow.com/a/20628522):
> git 2.9 (June 2016) will use user.signingKey to force signing annotated tags as well as commits: commit 61c2fe0.

@eagletmt since git already signs annotated tags if `user.signingKey` is present, why is this change to Bundler needed?

@rafaelfranca `user.signingkey` is the correct config setting according to the GitHub docs: https://help.github.com/articles/telling-git-about-your-gpg-key/. Setting `commit.gpgsign` merely tells git to automatically sign each commit you create (effectively adding `-s` to all invocations of `git commit`).

indirect:
Oops! I misunderstood the git commit, sorry.

@rafaelfranca yep, you're totally right. 😆

@eagletmt you can get this behavior by setting `git config tag.forcesignannotated true`. I think this means it does not need to be a setting in Bundler. Do you agree?

rafaelfranca:
> I think this means it does not need to be a setting in Bundler. Do you agree?

I think we still need to remove the `-a` option that Bundler adds because that option disable the signature.

```
tag.forceSignAnnotated
           A boolean to specify whether annotated tags created should be GPG signed. If --annotate is specified
           on the command line, it takes precedence over this option.
```

eagletmt:
I didn't know tag.forceSignAnnotated. Removing `-a` seems sufficient to me 😌 
We can remove `-a` without changing the current behavior because `-m` implies `-a`.

> If -m <msg> or -F <file> is given and -a, -s, and -u <keyid> are absent, -a is implied
> https://www.kernel.org/pub/software/scm/git/docs/git-tag.html

eagletmt:
totally changed my patch to simply removing `-a` option.

indirect:
👍 @homu r+

PR #4744, opened 2016-07-02
Fix test failures due to Rack 2.x

eagletmt:
- Rack 2.x requires Ruby version >= 2.2.2
- artifice doesn't support rack 2.x now.


segiddins:
@homu r+

PR #4745, opened 2016-07-02
Installing gems without version after the ones with version for tests

asutoshpalai:
`rake < 2` should be installed before `artifice`.

On `ruby 1.8.7` it fails randomly because of no specific order of hash traversal.

https://travis-ci.org/bundler/bundler/builds/141867428


segiddins:
@homu r+
Thanks!

PR #4746, opened 2016-07-03
Simplified Gemfile template

KrauseFx:


indirect:
@segiddins can you chime in on the frozen string comment? Since Gemfiles are ultimately processed by `instance_eval`, I'm pretty sure we don't actually need the comment, right?

segiddins:
@indirect I'd rather have the comment, I'm pretty sure it is taken into account, and will help people when migrating to ruby 3 or using the frozen string literal ruby opts

KrauseFx:
So should I close this PR? 
I wonder if we can delete the `# A sample Gemfile` still?

indirect:
I am fine with deleting `# A sample Gemfile`. I'm also pretty sure that `instance_eval` does not respect magic comment strings. If someone can figure that out one way or another, we'll know whether we need it in the template.

segiddins:
@indirect i verified yesterday that eval does indeed respect the magic string comment

indirect:
Closing this for now, but if you still want to delete the sample gemfile line feel free to update this PR and let us know.

KrauseFx:
@indirect awesome, thanks: https://github.com/bundler/bundler/pull/4880

PR #4756, opened 2016-07-05
Fix RubyGems printing packaging messages during the specs

### LINKED ISSUE #4755: Suppress gem building logs in test output
Recently, it appears that tests have started to print extremely verbose log output, like this:

```
WARNING:  bin/rackup is not executable
WARNING:  See http://guides.rubygems.org/specification-reference/ for help
  Successfully built RubyGem
  Name: rack
  Version: 1.2
  File: rack-1.2.gem
  Successfully built RubyGem
  Name: activesupport
  Version: 3.0
  File: activesupport-3.0.gem
```

This output is printed into the terminal even when running tests that pass on Travis. All that extra output is making it hard to find the actual output that we want, so it would be great if we could stop printing those messages.

segiddins:
Closes #4755.
My bad!


segiddins:
@homu r+

PR #4765, opened 2016-07-07
WIP: add command to check dynamic library linkage

mistydemeo:
This new command, linkage, checks for broken dynamic library links in C extensions.

I'd heard there was some interest in adding this functionality, so I thought I'd submit a preliminary PR for discussion. In my experience, broken dylib linkage is a common issue with C extensions, so I think having a good way to diagnose it would be valuable.

This command scans all of the specifications in the bundle for .bundle files in C extensions. If any of the dynamic libraries linked against no longer exist, bundler will report a message to the console and exit non-0.

TODOs:
* Add support for non-Darwin OSs
* Improve tests

A few questions:

* Is there a good way to mock functionality in the tests? Doing it in the standard rspec way isn't working since `bundle :command` runs in a subprocess. I'd like to be able to stub out stuff that actually checks dylibs and the like.
* Is making this a new command the right approach? I assumed this wouldn't be ideal to include in, say, `check` because it would slow it down.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -479,9 +479,6 @@ module Bundler
       Doctor scans the OS dependencies of each of the gems requested in the Gemfile. If
       missing dependencies are detected, Bundler prints them and exits status 1.
       Otherwise, Bundler prints a success message and exits with a status of 0.
-
-      The bundle's gem dependencies must all be installed to run this command; if
-      they are not, Bundler prints an error message and exits with a status of 2.
     D
     method_option "gemfile", :type => :string, :banner =>
       "Use the specified gemfile instead of Gemfile"
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -474,6 +474,24 @@ module Bundler
       Inject.new(options, name, version, gems).run
     end
 
+    desc "doctor [OPTIONS]", "Checks the bundle for common problems"
+    long_desc <<-D
+      Doctor scans the OS dependencies of each of the gems requested in the Gemfile. If
+      missing dependencies are detected, Bundler prints them and exits status 1.
+      Otherwise, Bundler prints a success message and exits with a status of 0.
+
+      The bundle's gem dependencies must all be installed to run this command; if
+      they are not, Bundler prints an error message and exits with a status of 2.
+    D
+    method_option "gemfile", :type => :string, :banner =>
+      "Use the specified gemfile instead of Gemfile"
+    method_option "quiet", :type => :boolean, :banner =>
+        "Only output warnings and errors."
+    def doctor
+      require "bundler/cli/doctor"
+      Doctor.new(options).run
+    end
+
     if Bundler.settings[:plugins]
       require "bundler/cli/plugin"
       desc "plugin SUBCOMMAND ...ARGS", "manage the bundler plugins"
--- a/lib/bundler/cli/doctor.rb
+++ b/lib/bundler/cli/doctor.rb
@@ -66,9 +66,9 @@ module Bundler
         not_installed = definition.missing_specs
         raise GemNotFound if not_installed.any?
       rescue GemNotFound
-        Bundler.ui.error "This bundle's gems must be installed to run this command."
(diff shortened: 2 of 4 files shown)


indirect:
Whoa, awesome! There is definitely interest in having something like this.

indirect:
1. I would suggest writing a CLI command class that contains all the code, and then unit-test invoking that class. The `bundle! :command` helper is really only for integration tests, and this command probably only needs a single one to make sure the entry in cli.rb is wired up to the command class.

2. Maybe the command could be named `doctor`, and we could try to diagnose other common problems? "You're running Spring, so changes to your Gemfile might not apply until you restart Spring.", that sort of thing.

mistydemeo:
> Maybe the command could be named doctor, and we could try to diagnose other common problems?

I think this is a great idea! Pushed a commit to rename it.

Glad to hear there's interest! I'll beef up the tests (thanks for the suggestion), and add support for `ldd`.

mistydemeo:
Looks like the one failure on Travis is unrelated to my PR? It succeeded on the other jobs.

segiddins:
@mistydemeo I restarted that build and it failed again. I'm happy to take a look at that Ruby/RubyGems combination locally if you'd like?

mistydemeo:
Looks like the test is flaky; my `File.exist?` stub is overaggressive and stomping on other invocations. May be order-dependent. Will fix!

mistydemeo:
Rebased on master, and made some extra changes:

* Since this will eventually have more checks, not being able to scan deps due to them not being installed is no longer fatal and exits 0.
* The dylib check now checks for the presence of a working `otool` or `ldd` as appropriate, and returns a stub result if they aren't present. Unfortunately, due to limitations of recent OS X versions, this may pop up a GUI window for users who a) have gems with C extensions installed, and b) do not have the CLT or Xcode installed. This should be niche, but could be irritating.
* Added `doctor` to the manpage.

Any other changes you'd like to see?

indirect:
Hooray! 🎉  🏆  I'm super excited about this, thank you so much!

@homu r+

lib/bundler/cli.rb
```diff
@@ -474,6 +474,24 @@ def add(name, version = nil, *gems)
...
     end
 
+    desc "doctor [OPTIONS]", "Checks the bundle for common problems"
+    long_desc <<-D
+      Doctor scans the OS dependencies of each of the gems requested in the Gemfile. If
+      missing dependencies are detected, Bundler prints them and exits status 1.
+      Otherwise, Bundler prints a success message and exits with a status of 0.
+
+      The bundle's gem dependencies must all be installed to run this command; if
+      they are not, Bundler prints an error message and exits with a status of 2.
```
mistydemeo: This is no longer correct, I should update these docs. Will pushing a new commit mess with the approval you just granted?

indirect:
@homu r-

lib/bundler/cli.rb
indirect: I _think_ that I have just cancelled the imminent merge of this PR. Go ahead and push another commit, and I'll kick off another build to get a final merge with green tests. Thanks for following up!

lib/bundler/cli/doctor.rb
```diff
@@ -0,0 +1,95 @@
...
+    LDD_REGEX = /\t\S+ => (\S+) \(\S+\)/
+
+    attr_reader :options
+
+    def initialize(options)
+      @options = options
+    end
+
+    def otool_available?
+      system("otool --version 2>&1 >/dev/null")
```
segiddins: we can't use `/dev/null` and need to use `Bundler::NULL` instead because of Windows

mistydemeo: Sure thing, though this line is only run on Darwin. Same with `ldd_available?` below.

mistydemeo:
Updated.

mistydemeo:
This only failed failed on one job, and it seems unrelated to my PR.

indirect:
Sometimes the suite is kind of flaky on old Rubies. :/ hopefully it will pass for merging!

@homu r+

PR #4770, opened 2016-07-09
[RubygemsIntegration] Add support for reversing hooks

segiddins:
This is a pre-req for being able to use bundler as the backend for RubyGems' Gemfile support, as RubyGems tests run in a single process and the hooks installed (particularly by `replace_entrypoints`) are rather destructive to those tests. More changes to come to get `Gem.use_gemdeps` using bundler (i.e. to get the RG suite green), but this is likely to be the grossest of them.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -7,7 +7,7 @@ module Bundler
   class Definition
     include GemHelpers
 
-    attr_reader :dependencies, :platforms, :ruby_version, :locked_deps, :gem_version_promoter
+    attr_reader :dependencies, :platforms, :ruby_version, :locked_deps, :gem_version_promoter, :requires
 
     # Given a gemfile and lockfile creates a Bundler definition
     #
@@ -106,6 +106,8 @@ module Bundler
       @source_changes = converge_sources
       @dependency_changes = converge_dependencies
       @local_changes = converge_locals
+
+      @requires = compute_requires
 
       fixup_dependency_types!
     end
@@ -215,7 +217,7 @@ module Bundler
     end
 
     def current_dependencies
-      dependencies.reject {|d| !d.should_include? }
+      dependencies.select(&:should_include?)
     end
 
     def specs_for(groups)
@@ -767,6 +769,17 @@ module Bundler
       # This method will extract the error message like "Could not find foo-1.2.3 in any of the sources"
       # to an array. The first element will be the gem name (e.g. foo), the second will be the version number.
       error.message.scan(/Could not find (\w+)-(\d+(?:\.\d+)+)/).flatten
+    end
+
+    def compute_requires
+      dependencies.reduce({}) do |requires, dep|
+        next requires unless dep.should_include?
+        requires[dep.name] = Array(dep.autorequire || dep.name).map do |file|
+          # Allow `require: true` as an alias for `require: <name>`
+          file == true ? dep.name : file
+        end
+        requires
+      end
     end
   end
 end
--- a/lib/bundler/rubygems_ext.rb
+++ b/lib/bundler/rubygems_ext.rb
@@ -16,7 +16,7 @@ module Gem
   class Specification
     attr_accessor :remote, :location, :relative_loaded_from
 
-    if instance_methods(false).include?(:source)
+    if instance_methods(false).map(&:to_sym).include?(:source)
       remove_method :source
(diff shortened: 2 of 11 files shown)


segiddins:
@indirect @RochesterinNYC please review, this is now green!

lib/bundler/definition.rb
```diff
@@ -768,5 +770,17 @@ def extract_gem_info(error)
       # to an array. The first element will be the gem name (e.g. foo), the second will be the version number.
       error.message.scan(/Could not find (\w+)-(\d+(?:\.\d+)+)/).flatten
     end
+
+    def compute_requires
+      dependencies.reduce({}) do |requires, dep|
+        next requires unless dep.should_include?
+        requires[dep.name] = Array(dep.autorequire || dep.name).each do |file|
```
segiddins: this needs to be `map`

lib/bundler/rubygems_ext.rb
```diff
@@ -16,8 +16,15 @@ module Gem
   class Specification
     attr_accessor :remote, :location, :relative_loaded_from
 
-    remove_method :source if instance_methods(false).include?(:source)
-    attr_accessor :source
+    if instance_methods(false).map(&:to_s).include?("source")
```
RochesterinNYC: What's the context behind the addition of the type conversion here? (mapping to an array of strings now instead of symbols)

lib/bundler/shared_helpers.rb
```diff
@@ -23,7 +23,7 @@ module SharedHelpers
     def default_gemfile
       gemfile = find_gemfile
       raise GemfileNotFound, "Could not locate Gemfile" unless gemfile
-      Pathname.new(gemfile)
+      Pathname.new(gemfile).untaint
```
RochesterinNYC: Why are the `untaint` invocations necessary for these methods?

spec/bundler/dsl_spec.rb
```diff
@@ -80,7 +80,7 @@
     it "handles syntax errors with a useful message" do
       expect(Bundler).to receive(:read_file).with("Gemfile").and_return("}")
       expect { subject.eval_gemfile("Gemfile") }.
-        to raise_error(Bundler::GemfileError, /There was an error parsing `Gemfile`: (syntax error, unexpected tSTRING_DEND|(compile error - )?syntax error, unexpected '}'). Bundler cannot continue./)
+        to raise_error(Bundler::GemfileError, /There was an error parsing `Gemfile`: (syntax error, unexpected tSTRING_DEND|(compile error - )?syntax error, unexpected '\}'). Bundler cannot continue./)
```
RochesterinNYC: Why does `}` need to be escaped?

spec/bundler/shared_helpers_spec.rb
```diff
@@ -209,6 +209,10 @@
   end
 
   describe "#set_bundle_environment" do
+    before do
+      ENV["BUNDLE_GEMFILE"] = "Gemfile"
```
RochesterinNYC: Isn't looking for a file called `Gemfile` the default? What's the impetus behind this setup?

lib/bundler/rubygems_ext.rb
segiddins: 1.8.7 returns strings

lib/bundler/shared_helpers.rb
segiddins: ruby 1.8.7 `$SAFE = 1` which RubyGems tests use

spec/bundler/dsl_spec.rb
segiddins: 1.8.7 gives a warning about it

spec/bundler/shared_helpers_spec.rb
segiddins: there isn't actually a gemfile there, it was just passing since there was an existing `@definition` set

lib/bundler/rubygems_ext.rb
indirect: Can we map to_sym so it's a no-op on the more common ruby versions?

indirect:
👍  this is good to go once comments are addressed.

segiddins:
@indirect done

indirect:
@homu r+

segiddins:
@homu retry

PR #4777, opened 2016-07-09
GVP - refactor move locked into sorting cases.

chrismo:
I've been changing bundler-patch to work with the code portions that
have been moved into Bundler, and having this extra bit of 'sorting'
code tacked on to the end of this case statement thwarts some ability to
override sorting behavior in a subclass.

It's also been a smell for quite a while. When initially working through
all of the sorting cases, I couldn't wrap my brain around how to do this
more complicated bit quickly enough and fell to the way it was before.

Doing it within the sorting cases isn't that difficult when approached
properly, so ... here it is.

### DIFF
--- a/lib/bundler/gem_version_promoter.rb
+++ b/lib/bundler/gem_version_promoter.rb
@@ -129,15 +129,10 @@ module Bundler
           b_ver <=> a_ver
         when !minor? && segments_do_not_match(:minor, a_ver, b_ver)
           b_ver <=> a_ver
+        when !unlocking_gem?(gem_name) && one_version_matches(locked_version, a_ver, b_ver)
+          sort_matching_to_end(locked_version, a_ver, b_ver)
         else
           a_ver <=> b_ver
-        end
-      end.tap do |result|
-        # default :major behavior in Bundler does not do this
-        unless major?
-          unless unlocking_gem?(gem_name)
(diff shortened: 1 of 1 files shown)


lib/bundler/gem_version_promoter.rb
```diff
@@ -129,16 +129,11 @@ def sort_dep_specs(spec_groups, locked_spec)
           b_ver <=> a_ver
         when !minor? && segments_do_not_match(:minor, a_ver, b_ver)
           b_ver <=> a_ver
+        when !unlocking_gem?(gem_name) && one_version_matches(locked_version, a_ver, b_ver)
```
segiddins: personally this would read cleaner to me as `(locked_version == a_ver || locked_version == b_ver)`

chrismo: part of what i don't like about this is passing a_ver and b_ver around all over (OOP smell). I have similar code in bundler-patch that I tried recently making ivars of these (@a_ver and @b_ver) so I don't have to pass them through to see if I like it -- then it just reads `one_version_matches(locked_version)` -- what would you think about that?

chrismo:
so - this is a bad idea turns out. sorting of course doesn't compare every element, so this can result in inconsistent output.

PR #4779, opened 2016-07-12
GemVersionPromoter refactorings

chrismo:
I've been changing bundler-patch to work with the code portions that
have been moved into Bundler, and the tap block at the end of
sort_dep_specs couldn't be overridden, so that's been moved into its own
method.

Another refactoring was to move local vars into instance vars to stop
passing them around. Going all the way to push that into a class seemed
a bit much at this stage.

### DIFF
--- a/lib/bundler/gem_version_promoter.rb
+++ b/lib/bundler/gem_version_promoter.rb
@@ -114,52 +114,56 @@ module Bundler
 
     def sort_dep_specs(spec_groups, locked_spec)
       return spec_groups unless locked_spec
-      gem_name = locked_spec.name
-      locked_version = locked_spec.version
+      @gem_name = locked_spec.name
+      @locked_version = locked_spec.version
 
-      spec_groups.sort do |a, b|
-        a_ver = a.version
-        b_ver = b.version
+      result = spec_groups.sort do |a, b|
+        @a_ver = a.version
+        @b_ver = b.version
         case
         when major?
-          a_ver <=> b_ver
-        when either_version_older_than_locked(locked_version, a_ver, b_ver)
-          a_ver <=> b_ver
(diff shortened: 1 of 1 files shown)


chrismo:
@segiddins I've caught up bundler-patch to be compatible with this - if yer cool with it, be nice to have maybe an rc2 release when convenient?

indirect:
@chrismo can you talk me through the decision to use ivars instead of passing the values being compared to the functions?

chrismo:
(a) reads a lot cleaner to me (b) ivars are shared state within an object ... preferably they'd all go into their own class (along the lines of https://sourcemaking.com/refactoring/introduce-parameter-object), but that seems overkill and a perf risk to be spawning an object per comparison. I could go further a create a single instance that was reused - but then I could just 're-use' the instance I'm already in.

It's an awkward in-between state, but I feel is a compromise between readability of all the various comparison cases, good OOP and perf issues. 

Happy to discuss other options.

indirect:
That's fair. This seems good to me! @segiddins r?

segiddins:
I personally don't like the pattern but it's @chrismo's Code so it's up to him

chrismo:
If you feel the style is too different from the rest of the codebase or just have a strong preference that would be a good enough argument for me to go back to passing the args (I don't plan on going anywhere, but y'all on core team have a greater responsibility to it ;) 

Thinking about it, there's a good chance I'll be in this code again soon, with a couple pending specs and/or once we get some more feedback - we could go as-is for now, and I'll tinker with it some more.

indirect:
I'm think I'm okay with trying it—I feel kind of weird about it, but it also seems frustrating to have to pass the arguments a thousand times.  ¯\_(ツ)_/¯

@homu r+

PR #4782, opened 2016-07-14
fix standalone executable when dest of symlink

terinjokes:
When an executable is the destination of a symlink `__FILE__` contains
the path to the symlink, rather than the actual path of the current
file.

`Pathname#realpath` resolves symlinks to get the actual location on disk.

### DIFF
--- a/spec/install/gems/standalone_spec.rb
+++ b/spec/install/gems/standalone_spec.rb
@@ -279,13 +279,26 @@ shared_examples "bundle install --standalone" do
     it "creates stubs that can be executed from anywhere" do
       require "tmpdir"
       Dir.chdir(Dir.tmpdir) do
-        expect(`#{bundled_app("bin/rails")} -v`.chomp).to eql "2.3.2"
+        sys_exec!(%(#{bundled_app("bin/rails")} -v))
+        expect(out).to eq("2.3.2")
       end
+    end
+
+    it "creates stubs that can be symlinked" do
+      pending "File.symlink is unsupported on Windows" if Bundler::WINDOWS
+
+      symlink_dir = tmp("symlink")
+      FileUtils.mkdir_p(symlink_dir)
+      symlink = File.join(symlink_dir, "rails")
+
+      File.symlink(bundled_app("bin/rails"), symlink)
+      sys_exec!("#{symlink} -v")
+      expect(out).to eq("2.3.2")
     end
 
     it "creates stubs with the correct load path" do
       extension_line = File.read(bundled_app("bin/rails")).each_line.find {|line| line.include? "$:.unshift" }.strip
-      expect(extension_line).to eq "$:.unshift File.expand_path '../../bundle', __FILE__"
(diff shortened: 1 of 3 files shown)


segiddins:
Can you add a spec covering the bug this fixes? Thanks!

terinjokes:
@segiddins Absolutely! Are there already specs for this part of the code? I looked for them, but I didn't find them, and I feel like I missed it.

segiddins:
I'll have a look next time I'm at my computer, but I feel we might have a standalone_specs.rb?

terinjokes:
@segiddins I've dded a spec to confirm that running a symlink works.

spec/install/gems/standalone_spec.rb
```diff
@@ -283,9 +283,16 @@
       end
     end
 
+    it "creates stubs that can be symlinked" do
+      require "tmpdir"
+      symlink = File.expand_path("rails", Dir.mktmpdir)
+      File.symlink(bundled_app("bin/rails"), symlink)
+      expect(`#{symlink} -v`.chomp).to eql "2.3.2"
```
segiddins:
```ruby
sys_exec!("#{symlink} -v")
expect(out).to eq("2.3.2")
```

spec/install/gems/standalone_spec.rb
```diff
@@ -283,9 +283,16 @@
       end
     end
 
+    it "creates stubs that can be symlinked" do
+      require "tmpdir"
+      symlink = File.expand_path("rails", Dir.mktmpdir)
```
segiddins: `symlink = tmp("symlink", "rails")` to ensure the file is under bundlers control

spec/install/gems/standalone_spec.rb
```diff
@@ -283,9 +283,16 @@
...
+      File.symlink(bundled_app("bin/rails"), symlink)
```
segiddins: this should fail gracefully if we're on a platform that doesn't support symlink, instead of causing a test failure

segiddins:
👍 from me other than the test comments

spec/install/gems/standalone_spec.rb
terinjokes: How? I tried catching the exception, but got errors from rubocop.

spec/install/gems/standalone_spec.rb
terinjokes: I'm paralleling the nearby test. Should I change it as well?

spec/install/gems/standalone_spec.rb
segiddins:
```ruby
pending "File.symlink is unsupported on windows" if Bundler::WINDOWS
```
at the top of the spec?

spec/install/gems/standalone_spec.rb
segiddins: sure!

terinjokes:
@segiddins sorry I was away. I'm currently addressing the issues you brought up.

indirect:
Thanks!

@homu r+

segiddins:
@homu retry

PR #4784, opened 2016-07-14
update path already exists error message to not be specific to symlinks

colby-swandale:
If i'm attempting to install my gems into a folder that conflicts with a file that already exists, Bundler will raise the following error: 
```
Could not install to path `vendor` because of an invalid symlink. Remove the symlink so the directory can be created.
```

But this error will be raised even if its just a regular ol file and not a symlink. IMO its a better user experience to drop the symlink mention and ask the user to simply move or rename the file.

Example:

```
Could not install to path `vendor` because a file already exists at that path. Either remove or rename the file so the directory can be created.
```


segiddins:
👍🏻

indirect:
@homu r+

PR #4787, opened 2016-07-15
Random bits of cleanup ¯\_(ツ)_/¯

segiddins:


segiddins:
This should pass now

lib/bundler/dsl.rb
```diff
@@ -332,7 +332,20 @@ def normalize_options(name, version, opts)
...
+                  val.call
+                when String
+                  begin
+                    eval(val, TOPLEVEL_BINDING) # rubocop:disable Lint/Eval
+                  rescue
+                    true
+                  end
+                else
+                  true
+                end
```
indirect: we didn't support evaling strings before, so I don't think this shouldn't be in a random cleanup—it should be in a PR, with a changelog entry and an addition to the docs for `install_if`. also I feel like an indented case statement like this is gross, can you extract the evaluation into a method?

indirect:
👍  once comments are addressed and the specs pass

lib/bundler/dsl.rb
segiddins: I can move this commit to a separate PR, sure. I think this was actually a remnant of me experimenting with something else

segiddins:
@indirect done, and fixed up a RubyGems warning as well

lib/bundler/dsl.rb
indirect: you should actually remove this now that it's in a separate PR 😁

indirect:
@homu r+

lib/bundler/dsl.rb
segiddins: I did?

segiddins:
@homu r=indirect

PR #4788, opened 2016-07-15
Project local plugins

asutoshpalai:
If the user is in an app directory the plugin is installed in `.bundle/` inside the app root.

If the user is in not in any app directory the plugin is installed in the `.bundle/` in the user's home.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -393,6 +393,7 @@ module Bundler
       rubygems.undo_replacements
       rubygems.reset
       @rubygems = nil
+      Plugin.reset!
     end
 
   private
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -389,11 +389,13 @@ module Bundler
       @locked_gems = nil
       @bundle_path = nil
       @bin_path = nil
+
+      Plugin.reset!
+
       return unless defined?(@rubygems) && @rubygems
       rubygems.undo_replacements
       rubygems.reset
       @rubygems = nil
-      Plugin.reset!
     end
 
   private
--- a/lib/bundler/plugin.rb
+++ b/lib/bundler/plugin.rb
@@ -146,7 +146,7 @@ module Bundler
 
     # Used by specs
     def reset!
-      instance_variables.each {|i| remove_instance_variable(i)}
+      instance_variables.each {|i| remove_instance_variable(i) }
 
       @sources = {}
       @commands = {}
--- a/lib/bundler/plugin.rb
+++ b/lib/bundler/plugin.rb
@@ -63,9 +63,25 @@ module Bundler
       @index ||= Index.new
     end
 
-    # The directory root to all plugin related data
+    # The directory root for all plugin related data
+    #
+    # Points to root in app_config_path if ran in an app else points to the one
+    # in user_bundle_path
     def root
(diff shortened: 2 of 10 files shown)


lib/bundler/plugin.rb
```diff
@@ -128,6 +144,14 @@ def installed?(plugin)
       Index.new.installed?(plugin)
     end
 
+    # Used by specs
+    def reset!
```
segiddins: `Bundler.reset!` should call this

segiddins:
This seems like it might be a bit confusing -- silently switching behavior based upon the current working director

asutoshpalai:
Any other suggestions for project level plugins.

An alternative I thought is to make `source` plugins project level and `command` plugins global.

segiddins:
@asutoshpalai maybe the way `bundle config` does it? defaults to local, has `--local` and `--global`?

asutoshpalai:
Yes, right!
But what about the cases when the user is not in any app directory and tries to install/use plugin (command plugins)? Should we fail or switch silently (or verbosly)?

segiddins:
@indirect thoughts?

indirect:
Source plugins should definitely be project level—they only apply to a single gemfile at a time.

My (admittedly vague) general idea was that command plugins would be global, and installed via `gem install` rather than listed in a gemfile—but I don't know how that interacts with command plugins that are used by a specific project. It seems like `bundle grep` should be a global plugin, while `bundle audit` should be a project-level plugin. Maybe we should support both, for different kinds of commands?

asutoshpalai:
How about this:
Everything thing is in general project level. When in a project **only commands** get _merged_ from global level (can't see how it harms since they don't modify bundler's flow).

When not in any project, the global commands are available.

indirect:
@segiddins r?

lib/bundler.rb
```diff
@@ -393,6 +393,7 @@ def reset!
       rubygems.undo_replacements
       rubygems.reset
       @rubygems = nil
+      Plugin.reset!
```
segiddins: this needs to be executed regardless of whether theres a `rubygems` or not

spec/bundler/plugin/index_spec.rb
```diff
@@ -75,6 +79,20 @@
     end
   end
 
+  describe "global index" do
+    before do
+      Dir.chdir tmp
```
segiddins: this should use the block form of `chdir`

spec/support/path.rb
```diff
@@ -81,12 +81,12 @@ def bundler_path
       Pathname.new(File.expand_path("../../../lib", __FILE__))
     end
 
-    def plugin_root(*args)
-      home ".bundle", "plugin", *args
+    def global_plugin_gems(*args)
```
segiddins: `gem` in the singular

spec/support/path.rb
```diff
@@ -81,12 +81,12 @@ def bundler_path
...
+      home ".bundle", "plugin", "gems", *args
     end
 
-    def plugin_gems(*args)
-      plugin_root "gems", *args
+    def local_plugin_gems(*args)
```
segiddins: singular `gem`

indirect:
@segiddins how's it looking now, r?

segiddins:
@homu r+

Issue #4586, opened 2016-05-18
bundle exec rake on jruby 1.7.19 running in --1.8 mode


ms-ati:
It appears that Bundler 1.12.4 cannot exec rake 10.5.0 on jruby 1.7.19 running in --1.8 mode. This happens to be the configuration used in Travis CI for `jruby-18mode`.

To reproduce locally:

```
rvm get stable
rvm install jruby-1.7.19
rvm use jruby-1.7.19@new-gemset-name --create
gem install rake -v 10.5.0
bundle exec rake
```

The result:

```
- What did you do?

  I ran the command `/Users/msiegel/.rvm/gems/jruby-1.7.19@docile/bin/bundle exec rake`

- What did you expect to happen?

  I expected Bundler to...

- What happened instead?

  Instead, what actually happened was...


Error details

    TypeError: can't convert Array into String
      org/jruby/RubyKernel.java:1766:in `exec'
      /Users/msiegel/.rvm/gems/jruby-1.7.19@docile/gems/bundler-1.12.4/lib/bundler/cli/exec.rb:44:in `kernel_exec'
      /Users/msiegel/.rvm/gems/jruby-1.7.19@docile/gems/bundler-1.12.4/lib/bundler/cli/exec.rb:26:in `run'
      /Users/msiegel/.rvm/gems/jruby-1.7.19@docile/gems/bundler-1.12.4/lib/bundler/cli.rb:304:in `exec'
      org/jruby/RubyKernel.java:2222:in `send'
      /Users/msiegel/.rvm/gems/jruby-1.7.19@docile/gems/bundler-1.12.4/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
      /Users/msiegel/.rvm/gems/jruby-1.7.19@docile/gems/bundler-1.12.4/lib/bundler/vendor/thor/lib/thor/invocation.rb:126:in `invoke_command'
      /Users/msiegel/.rvm/gems/jruby-1.7.19@docile/gems/bundler-1.12.4/lib/bundler/vendor/thor/lib/thor.rb:359:in `dispatch'
      /Users/msiegel/.rvm/gems/jruby-1.7.19@docile/gems/bundler-1.12.4/lib/bundler/vendor/thor/lib/thor/base.rb:440:in `start'
      /Users/msiegel/.rvm/gems/jruby-1.7.19@docile/gems/bundler-1.12.4/lib/bundler/cli.rb:11:in `start'
[...]

PR #4790, opened 2016-07-16
Improve exec on JRuby

ojab:
Unfortunately I can't get testsuite running locally, hope that travis will tell that changes are OK.
Fixes https://github.com/bundler/bundler/issues/4586.

I don't really know if test should be added for the last commit, please advise.

### DIFF
--- a/lib/bundler/cli/exec.rb
+++ b/lib/bundler/cli/exec.rb
@@ -25,7 +25,11 @@ module Bundler
       if bin_path = Bundler.which(cmd)
         return kernel_load(bin_path, *args) if ruby_shebang?(bin_path)
         # First, try to exec directly to something in PATH
-        kernel_exec([bin_path, cmd], *args)
+        if Bundler.current_ruby.jruby_18?
+          kernel_exec(bin_path, *args)
+        else
+          kernel_exec([bin_path, cmd], *args)
+        end
       else
         # exec using the given command
         kernel_exec(cmd, *args)
--- a/lib/bundler/cli/exec.rb
+++ b/lib/bundler/cli/exec.rb
@@ -77,6 +77,7 @@ module Bundler
     def ruby_shebang?(file)
       possibilities = [
         "#!/usr/bin/env ruby\n",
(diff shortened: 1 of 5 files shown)


ojab:
Ouch, JRuby is not tested on travis :/.
Anyway, manual testing of exec on JRuby works fine locally and changes are not too invasive, so hope that's ok.

spec/spec_helper.rb
```diff
@@ -85,10 +85,6 @@
   original_wd  = Dir.pwd
   original_env = ENV.to_hash
 
-  def pending_jruby_shebang_fix
```
segiddins: Why did you remove this?

ojab: Oops, doesn't commited spec/support/builders.rb changes, pushed fixed version.
I assume that `executables` is actually ruby scripts here regardless of RUBY_PLATFORM, so why not check `#/usr/bin/env jruby\n` shebang on JRuby.
These test are passing here w/ `ruby_shebang?` changes.

lib/bundler/cli/exec.rb
```diff
@@ -25,7 +25,11 @@ def run
       if bin_path = Bundler.which(cmd)
         return kernel_load(bin_path, *args) if ruby_shebang?(bin_path)
         # First, try to exec directly to something in PATH
-        kernel_exec([bin_path, cmd], *args)
+        if RUBY_PLATFORM == "java" && RUBY_VERSION < "1.9"
```
segiddins: `if Bundler.current_ruby.jruby_18?`

spec/support/builders.rb
```diff
@@ -534,8 +534,13 @@ def executables=(val)
         @spec.executables = Array(val)
         @spec.executables.each do |file|
           executable = "#{@spec.bindir}/#{file}"
+          shebang = if RUBY_PLATFORM == "java"
```
segiddins: `Bundler.current_ruby.jruby?`

segiddins:
@homu r+

PR #4793, opened 2016-07-18
Add spec for only updating from pinned source

### LINKED ISSUE #4631: Source pinning seems to be ignored during updates
Given a Gemfile where a gem is pinned to a secondary source:

```ruby
source "https://rubygems.org"
source "private" do
  gem "pony"
end
```

It appears that running `bundle update --source pony` (and possibly even just `bundle update`) will install the pony gem from RubyGems if the version available from RubyGems is higher than the version available from the private server.

b-ggs:
@indirect @RochesterinNYC 

---

Closes #4631


segiddins:
@homu r+

PR #4795, opened 2016-07-18
Add message filter to generate better search URL

### LINKED ISSUE #4771: Permissions errors generate a search URL with the user's local paths
As displayed by https://github.com/bundler/bundler/issues/4429, https://github.com/bundler/bundler/issues/4429, https://github.com/bundler/bundler/issues/4458 the error message for `Errno::EACCES: Permission denied @ rb_sysopen` often includes the users local directory on the first line (ex `/Users/username/.bundle/...`). This means that [the search url](https://github.com/bundler/bundler/blob/196594bb6992e01ee4654b970f486a37bf0e78fa/lib/bundler/friendly_errors.rb#93) won't return any results. Fixing this either by

* splitting that line on the `-`, or
* moving the user path to the second line of the error message

Would likely save a few users worth of issues.

(side note: I also made another issue https://github.com/bundler/bundler/issues/4647 about this type of error, but this one is simpler)

allenzhao:
Fixes #4771 
/c @RochesterinNYC @segiddins

### DIFF
--- a/lib/bundler/friendly_errors.rb
+++ b/lib/bundler/friendly_errors.rb
@@ -89,8 +89,10 @@ module Bundler
     end
 
     def issues_url(exception)
+      message = exception.message.lines.first.tr(":", " ").chomp
+      message = message.split("-").first if exception.is_a?(Errno)
       "https://github.com/bundler/bundler/search?q=" \
-        "#{CGI.escape(exception.message.lines.first.chomp)}&type=Issues"
+        "#{CGI.escape(message)}&type=Issues"
     end
   end
 
--- a/spec/bundler/friendly_errors_spec.rb
+++ b/spec/bundler/friendly_errors_spec.rb
@@ -2,6 +2,7 @@
 require "spec_helper"
 require "bundler"
 require "bundler/friendly_errors"
+require "cgi"
 
 describe Bundler, "friendly errors" do
   context "with invalid YAML in .gemrc" do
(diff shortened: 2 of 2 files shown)


segiddins:
Could you add a test for this fix?

allenzhao:
Tests seems passing ;)
@segiddins @RochesterinNYC Please review ;)

spec/bundler/friendly_errors_spec.rb
```diff
@@ -66,5 +66,20 @@
 
       expect(Bundler::FriendlyErrors.issues_url(exception)).to eq("https://github.com/bundler/bundler/search?q=First+line+of+the+exception+message&type=Issues")
     end
+
+    it "generates the url without colons" do
+      exception = Exception.new(<<END)
+Errno::EACCES: Permission denied @ dir_s_mkdir - /Users/foo/bar/
+END
+      expect(Bundler::FriendlyErrors.issues_url(exception)).not_to include("%3A")
```
RochesterinNYC: Can we include a positive assertion as well? Only testing for negatives/absence is a spec can be misleading/pass for the wrong reasons.

spec/bundler/friendly_errors_spec.rb
```diff
@@ -66,5 +66,20 @@
...
+    end
+
+    it "removes - for Errono::EACCES" do
+      exception = Exception.new(<<END)
+Errno::EACCES: Permission denied @ dir_s_mkdir - /Users/foo/bar/
+END
+      allow(exception).to receive(:is_a?).with(Errno::EACCES).and_return(true)
+      expect(Bundler::FriendlyErrors.issues_url(exception)).not_to include("/Users/foo/bar")
```
RochesterinNYC: Can we include a positive assertion here as well?

lib/bundler/friendly_errors.rb
```diff
@@ -89,8 +89,10 @@ def request_issue_report_for(e)
     end
 
     def issues_url(exception)
+      message = exception.message.lines.first.tr(":", " ").chomp
+      message = message.split("-").first if exception.is_a?(Errno::EACCES)
```
segiddins: this could happen on errors other than an `EACCESS` too, I believe, such as `ENOTSUP`

allenzhao: You mean to use `Errno` instead? I'm not following

segiddins: yes

allenzhao:
I've updated and the CI build shall pass this time. Please review again ;)
@segiddins @RochesterinNYC

spec/bundler/friendly_errors_spec.rb
```diff
@@ -66,5 +67,24 @@
...
       expect(Bundler::FriendlyErrors.issues_url(exception)).to eq("https://github.com/bundler/bundler/search?q=First+line+of+the+exception+message&type=Issues")
     end
+
+    it "generates the url without colons" do
+      exception = Exception.new(<<END)
+Errno::EACCES: Permission denied @ dir_s_mkdir - /Users/foo/bar/
+END
+      issues_url = Bundler::FriendlyErrors.issues_url(exception)
+      expect(issues_url).not_to include("%3A")
+      expect(issues_url).to include(CGI.escape("Errno  EACCES  Permission denied @ dir_s_mkdir"))
```
segiddins: instead to checking for `include`, can you just made these all equality tests?

segiddins:
@homu delegate=RochesterinNYC

RochesterinNYC:
@homu r+

Issue #4753, opened 2016-07-05
Trampoline version message UX problems


indirect:
Here's what I did:

```
$ echo 'source "https://rubygems.org"' > Gemfile

$ echo 'GEM
  remote: https://rubygems.org/
  specs:

PLATFORMS
  ruby

DEPENDENCIES

BUNDLED WITH
   1.12.0' > Gemfile.lock

$ dbundle install
Fetching: bundler-1.12.0.gem (100%)
You're running Bundler 1.13.0.rc.1 but this project uses 1.12.0. To update, run `bundle update --bundler`.
The Gemfile specifies no dependencies
Bundle complete! 0 Gemfile dependencies, 1 gem now installed.
Use `bundle show [gemname]` to see where a bundled gem is installed.

$ dbundle -v
1.12.0
```

After running `bundle install`, no output appeared for approximately 10 seconds. That's too long, and it makes it seem like Bundler has frozen. When you run `bundle install`, it should _immediately_ print a message like "Switching to locked Bundler version 1.12.0...". Then, if 1.12.0 still needs to be installed, we can optionally print something like "Downloading and installing Bundler 1.12.0...".

The later "You're running Bundler.." message should _not_ appear. We already promised (the first time they saw that message) that we wouldn't show that message again!

Finally, when you run `bundle -v` or `bundle version`, we need to print out the version of Bundler that is actually installed, not the version of Bundler that is in the lockfile. The output should be `1.13.0.rc.1`. If they need to know which version of Bundler this bundle is locked to, they can run `bundle show bundler`, and it will print a path that includes the version.

/cc @RochesterinNYC @b-ggs @segiddins

segiddins:
Why shouldn't bundle -v print out the version that will run?

segiddins:
@indirect I believe it prints twice since the trampoline is unable to access Bundler.settings -- it won't print twice when trampolining to a version >= 1.13.0.rc.1

segiddins:
I also personally think it shouldn't print anything different when installing, as that's simply just noise and is liable to break anything that depends upon bundler output

indirect:
@segiddins It's a non-starter to run `gem install bundler`, see "Installed Bundler 1.13" and then have `bundle -v` print an older version. If you run `bundle -v`, it should print the version of the gem that you have installed and that was invoked by the `bundle` command. If you want to also show the version that will be used to run other commands, we could print that as well, maybe something like `1.13.0 (Gemfile locked to 1.12.5)`.

[...]

PR #4804, opened 2016-07-24
Display message showing locked Bundler version is being installed

b-ggs:
This is for #4753.

[This](https://github.com/bundler/bundler/issues/4753#issuecomment-233786976) always showed the installing message whether the locked Bundler version was installed or not, so I moved the line into `lib/bundler/vendor/postit/lib/postit/installer.rb`.

I've also added the appropriate specs for this.

@RochesterinNYC @segiddins @indirect

### DIFF
--- a/lib/bundler/postit_trampoline.rb
+++ b/lib/bundler/postit_trampoline.rb
@@ -21,7 +21,7 @@ if ENV["BUNDLE_ENABLE_TRAMPOLINE"]
     begin
       installer = BundlerVendoredPostIt::PostIt::Installer.new(version)
       unless installer.installed?
-        warn "Installing locked Bundler version #{version}" 
+        warn "Installing locked Bundler version #{version}"
         installer.install!
       end
     rescue => e
--- a/lib/bundler/postit_trampoline.rb
+++ b/lib/bundler/postit_trampoline.rb
@@ -21,7 +21,7 @@ if ENV["BUNDLE_ENABLE_TRAMPOLINE"]
     begin
       installer = BundlerVendoredPostIt::PostIt::Installer.new(version)
       unless installer.installed?
-        warn "Installing locked Bundler version #{version}"
+        warn "Installing locked Bundler version #{version.to_s.gsub("= ", "")}..."
         installer.install!
       end
     rescue => e
--- a/lib/bundler/postit_trampoline.rb
+++ b/lib/bundler/postit_trampoline.rb
@@ -20,7 +20,10 @@ if ENV["BUNDLE_ENABLE_TRAMPOLINE"]
   if !version.satisfied_by?(installed_version)
     begin
       installer = BundlerVendoredPostIt::PostIt::Installer.new(version)
-      installer.install!
+      unless installer.installed?
+        warn "Installing locked Bundler version #{version}" 
+        installer.install!
+      end
     rescue => e
       abort <<-EOS.strip
 Installing the inferred bundler version (#{version}) failed.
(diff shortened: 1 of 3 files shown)


lib/bundler/vendor/postit/lib/postit/installer.rb
```diff
@@ -19,6 +19,7 @@ def installed?
 
     def install!
       return if installed?
+      puts "Installing locked Bundler version #{@bundler_version}"
```
segiddins: please move this into the postit trampoline, vendored code shouldn't be modified directly

lib/bundler/postit_trampoline.rb
```diff
@@ -19,7 +19,10 @@ module BundlerVendoredPostIt; end
 if !version.satisfied_by?(installed_version)
   begin
     installer = BundlerVendoredPostIt::PostIt::Installer.new(version)
-    installer.install!
+    unless installer.installed?
+      puts "Installing locked Bundler version #{version}"
```
segiddins: @indirect think this should go to stderr instead of stdout?

b-ggs:
@segiddins Done. I added a condition to see whether the locked Bundler version is installed before it shows the message and runs `installer.installed!`.

b-ggs:
@segiddins Gotcha. Also updated the appropriate specs to use `err`.

lib/bundler/postit_trampoline.rb
```diff
@@ -20,7 +20,7 @@ module BundlerVendoredPostIt; end
   begin
     installer = BundlerVendoredPostIt::PostIt::Installer.new(version)
     unless installer.installed?
-      puts "Installing locked Bundler version #{version}" 
+      STDERR.puts "Installing locked Bundler version #{version}"
```
segiddins: `warn`

b-ggs: Committed and squashed

segiddins:
@b-ggs you have to update `./spec/other/trampoline_spec.rb:85` as well

segiddins:
@indirect r?

b-ggs:
Rebased from upstream `master`.

RochesterinNYC:
@b-ggs Looks like a small formatting fix is needed (rubocop).

RochesterinNYC:
This LGTM.

lib/bundler/postit_trampoline.rb
```diff
@@ -20,7 +20,10 @@ module BundlerVendoredPostIt; end
   if !version.satisfied_by?(installed_version)
     begin
       installer = BundlerVendoredPostIt::PostIt::Installer.new(version)
-      installer.install!
+      unless installer.installed?
+        warn "Installing locked Bundler version #{version}"
```
indirect: We realize that the Gem::Version uses `=` to mean "an exact matching version", but most people don't, and I think end users will generally be confused by it.

Let's remove the `= ` from the beginning of the version, and add `...` to the end to indicate that it will take a moment to happen. So the message should look like: `Installing locked Bundler version 1.12.6...`

indirect:
This is looking good! One tiny change, and then I am r+.

@homu delegate=RochesterinNYC

lib/bundler/postit_trampoline.rb
b-ggs: Gotcha.

lib/bundler/postit_trampoline.rb
```diff
@@ -21,7 +21,7 @@ module BundlerVendoredPostIt; end
     begin
       installer = BundlerVendoredPostIt::PostIt::Installer.new(version)
       unless installer.installed?
-        warn "Installing locked Bundler version #{version}"
+        warn "Installing locked Bundler version #{version.to_s.gsub!("= ", "")}..."
```
segiddins: this needs to be `gsub` instead of `gsub!`

segiddins:
@homu r+

PR #4808, opened 2016-07-25
Add URI to http response output in debug mode

NickLaMuro:
Map request response to original URI when printing debug output for http responses for `bundler/fetcher/downloader`.

Overview
--------
I have recently been debugging some bundler stalling issues and found this addition to the debug output helpful when ruling out if a particular URL was timing out with the `CompactIndex` fetcher as it does it's requests over 25 threads and returns them asynchronously.  Getting a bunch of `HTTP 304 Not Modified` isn't as helpful if you are trying to determine how a thread is locking up.

Incidentally, this didn't end up helping solving some issues (further PRs will address this), but it seems like a simple debugging addition that could help rule out specific endpoint issues without requiring more [advanced debugging setups](https://gist.github.com/NickLaMuro/e923fc4e55307437a8f0e97ee6429410).

### DIFF
--- a/lib/bundler/fetcher/downloader.rb
+++ b/lib/bundler/fetcher/downloader.rb
@@ -14,7 +14,7 @@ module Bundler
         raise HTTPError, "Too many redirects" if counter >= redirect_limit
 
         response = request(uri, options)
-        Bundler.ui.debug("HTTP #{response.code} #{response.message}")
+        Bundler.ui.debug("HTTP #{response.code} #{response.message} #{uri}")
 
         case response
         when Net::HTTPSuccess, Net::HTTPNotModified
--- a/spec/bundler/fetcher/downloader_spec.rb
+++ b/spec/bundler/fetcher/downloader_spec.rb
@@ -31,7 +31,7 @@ describe Bundler::Fetcher::Downloader do
       let(:http_response) { Net::HTTPSuccess.new("1.1", 200, "Success") }
 
       it "should log the HTTP response code and message to debug" do
-        expect(Bundler).to receive_message_chain(:ui, :debug).with("HTTP 200 Success")
+        expect(Bundler).to receive_message_chain(:ui, :debug).with("HTTP 200 Success #{uri}")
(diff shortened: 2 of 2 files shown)


lib/bundler/fetcher/downloader.rb
```diff
@@ -14,7 +14,7 @@ def fetch(uri, options = {}, counter = 0)
         raise HTTPError, "Too many redirects" if counter >= redirect_limit
 
         response = request(uri, options)
-        Bundler.ui.debug("HTTP #{response.code} #{response.message}")
+        Bundler.ui.debug("HTTP #{response.code} #{response.message} #{uri}")
```
segiddins: maybe `HTTP GET #{uri} => #{response.code} #{response.message}` ?

NickLaMuro: @segiddins Works for me, and I can make that change.  The only problem I do see with it is the URIs plus query params can be quite lengthy, so putting the URI first might make it hard to quickly find 500 errors if the response code/message are at the end.  Thoughts?

segiddins: I'm thinking in terms of the message being long? ¯\_(ツ)_/¯ If in your experience it's been readable with the URL at the end as it is here, then lets :shipit:

NickLaMuro: Oh, heh.  Yeah, usually the message was something like "Not Modified", so the full thing was just "HTTP 304 Not Modified" (without the URI of course).

segiddins:
./spec/bundler/fetcher/downloader_spec.rb:35 needs to be updated, otherwise 👍 
(I thought `response.message` was `response.body` at first)

NickLaMuro:
@segiddins Sorry about that.  Like a Typical Developer™, I didn't run tests locally.  Fixed now.

To be fair, had I done so, I would have ran into the issues I have fixed in #4810.  `¯\_(ツ)_/¯`

segiddins:
@homu r+ thanks a bunch!

PR #4717, opened 2016-06-24
[GitProxy] Deinit submodules if they are not requested

segiddins:

PR #4810, opened 2016-07-26
Removes OSX/msysgit strings from Bundler::Source::Git::GitProxy#version

NickLaMuro:
Overview
--------

When running the previous version of `Bundler::Source::Git::GitProxy#version` output through `Gem::Version.create` when the git version is something like

    $ git --version
    git version 1.2.3 (Apple Git-22)

Ruby blows up with the error:

    ArgumentError:  Malformed version number string 1.2.3 (Apple Git-22)

The '(Apple Git-22)' and any additions that are added by `msysgit` (ex: `1.8.3.msysgit.0`) can be safely ignored as they are additions to the git version string for those specific versions, but should still line up with the core git version.

This regression was added in https://github.com/bundler/bundler/pull/4717, so this is not in a release of the gem yet, only in `1.13.0.rc1` installation.  This prevented:

* running tests on OSX or on any OS with a non-core git installation
* running `bundle install/update` with any git gem dependencies  in the Gemfile on OSX or any OS with a non-core git installation

This change simply regexps those version additions out, in addition to removing the `git version`, so only the version number (containing numbers or `.`s) will be parsed by this regexp.

Also, another method, `full_version` has been added to allow the `Bundler::Env` to report on the specific version of git being used, which is useful  when error reports are submitted.

Note about tests
----------------
The tests stub out any shell cmd and mocks a return value, which was easier to test the functionality of the regexp and requiring different git executable versions.  I also didn't add integration tests to address the regression caused by #4717, but simply tested that the versions parsed would work when sent through `Gem::Version.create`

This does mean I only tested a sample of git version types, so there might be something that I missed.  Also, the test for `full_version` working in reports was kinda jank, but it was the best I could do...

### DIFF
--- a/lib/bundler/env.rb
+++ b/lib/bundler/env.rb
@@ -75,7 +75,7 @@ module Bundler
     end
 
     def git_version
-      Bundler::Source::Git::GitProxy.new(nil, nil, nil).version
+      Bundler::Source::Git::GitProxy.new(nil, nil, nil).full_version
     rescue Bundler::Source::Git::GitNotInstalledError
       "not installed"
     end
--- a/lib/bundler/source/git/git_proxy.rb
+++ b/lib/bundler/source/git/git_proxy.rb
@@ -80,7 +80,7 @@ module Bundler
         end
 
         def version
-          git("--version").sub("git version", "").strip
+          git("--version").match(/(git version\s*)?((\.?\d+)+).*/)[2]
         end
 
         def checkout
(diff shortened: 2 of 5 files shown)


segiddins:
This fails rubocop: https://travis-ci.org/bundler/bundler/jobs/147334936#L501
`bin/rubocop -a`

NickLaMuro:
@segiddins Crap... sorry about that.  I thought the `spec/quality_spec.rb` covered the `rubocop` stuff as well.  Fixing.

segiddins:
<3
@homu r+

PR #4811, opened 2016-07-26
Fix random string printing inconsistencies and formatting

NickLaMuro:
Add dot output for CompactIndex fetcher

The `Bundler::Fetcher::CompactIndex`'s logging was missing the "dot logging" when not in debug mode, which is an assumed behavior based on the code in `lib/bundler/source/rubygems.rb`:

```ruby
api_fetchers.each do |f|
  Bundler.ui.info "Fetching gem metadata from #{f.uri}", Bundler.ui.debug?
  idx.use f.specs_with_retry(dependency_names, self)
  Bundler.ui.info "" unless Bundler.ui.debug? # new line now that the dots are over
end
```

While this isn't critical for `bundler` function properly by any stretch of the imagination, it provides a small bit of user feedback that requests are still continuing to be processed and `bundler` hasn't stalled.  It also maintains logging consistency between the different fetcher models.

Add newlines for `Bundler::Retry`
---------------------------------

This is very pedantic, but it makes it so that retry warning messages that used to look like this:

```
Fetching gem metadata from https://rubygems.org/....Retrying fetcher due to error (2/4): Error::OMGWatHappened LOL, no idea
.....
```

Now will look like this:

```
Fetching gem metadata from https://rubygems.org/....
Retrying fetcher due to error (2/4): Error::OMGWatHappened LOL, no idea.....
```

I think this reads a bit better and puts context to the "dots" so they now match up with the original attempt and the retries

Remove unneeded if statement
----------------------------

Ran across this while failing at writing tests for the above (for longer than I want to admit in public), but basically the `if` statement to determine whether or not to print "name" in the `Bundler.ui.warn` was not needed, as it never could be reached because of a short circuiting return statement prior to the print

```ruby
return true unless name
```

Have no idea how to test it anymore than we already are, so instead of embarrassing myself further, I moved on.

### DIFF
--- a/lib/bundler/fetcher/base.rb
+++ b/lib/bundler/fetcher/base.rb
@@ -36,6 +36,16 @@ module Bundler
       def api_fetcher?
         false
       end
+
+    private
+
+      def log_specs(debug_msg)
+        if Bundler.ui.debug?
+          Bundler.ui.debug debug_msg
+        else
+          Bundler.ui.info ".", false
+        end
+      end
     end
   end
 end
--- a/lib/bundler/fetcher/compact_index.rb
+++ b/lib/bundler/fetcher/compact_index.rb
@@ -36,7 +36,7 @@ module Bundler
         remaining_gems = gem_names.dup
 
         until remaining_gems.empty?
-          Bundler.ui.debug "Looking up gems #{remaining_gems.inspect}"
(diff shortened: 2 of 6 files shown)


segiddins:
<3 @homu r+

PR #4815, opened 2016-07-26
[Plugin] Life-cycle hooks

asutoshpalai:
Implements life-cycle hook plugins.

This shows only [`before-install-all` hook](https://github.com/bundler/bundler/pull/4815/files#diff-0964e829e3db904ba7935ca3a933f111R22), but other hooks can be easily added in the similar fashion.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -405,7 +405,6 @@ EOF
       rubygems.undo_replacements
       rubygems.reset
       @rubygems = nil
-      Plugin.reset!
     end
 
   private
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -405,6 +405,7 @@ EOF
       rubygems.undo_replacements
       rubygems.reset
       @rubygems = nil
+      Plugin.reset!
     end
 
   private
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -68,8 +68,6 @@ module Bundler
       definition = Bundler.definition
       definition.validate_ruby!
 
-      Plugin.hook("before-install-all", definition.dependencies)
-
       installer = Installer.install(Bundler.root, definition, options)
       Bundler.load.cache if Bundler.app_cache.exist? && !options["no-cache"] && !Bundler.settings[:frozen]
 
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -68,6 +68,8 @@ module Bundler
       definition = Bundler.definition
       definition.validate_ruby!
 
+      Plugin.hook("before-install-all", definition.dependencies)
+
       installer = Installer.install(Bundler.root, definition, options)
       Bundler.load.cache if Bundler.app_cache.exist? && !options["no-cache"] && !Bundler.settings[:frozen]
 
--- a/lib/bundler/plugin.rb
+++ b/lib/bundler/plugin.rb
@@ -18,6 +18,8 @@ module Bundler
 
     @commands = {}
     @sources = {}
+    @hooks = {}
+    @loaded = []
 
     # Installs a new plugin by the given name
     #
@@ -137,6 +139,19 @@ module Bundler
       src.new(locked_opts.merge("uri" => locked_opts["remote"]))
     end
 
+    def add_hook(event, &block)
+      (@hooks[event.to_s] ||= []) << block
+    end
+
+    def hook(event, *args)
+      plugins = index.hook_plugins(event)
+      return unless plugins
+
+      (plugins - @loaded).each {|name| load_plugin(name) }
+
+      @hooks[event].each {|blk| blk.call(*args) }
+    end
+
     # currently only intended for specs
     #
     # @return [String, nil] installed path
@@ -190,9 +205,11 @@ module Bundler
     def register_plugin(name, spec, optional_plugin = false)
       commands = @commands
       sources = @sources
+      hooks = @hooks
 
       @commands = {}
       @sources = {}
+      @hooks = {}
 
       load_paths = spec.load_paths
       add_to_load_path(load_paths)
@@ -208,12 +225,14 @@ module Bundler
         Bundler.rm_rf(path)
         false
       else
-        index.register_plugin name, path.to_s, load_paths, @commands.keys, @sources.keys
+        index.register_plugin name, path.to_s, load_paths, @commands.keys,
+          @sources.keys, @hooks.keys
         true
       end
     ensure
       @commands = commands
       @sources = sources
+      @hooks = hooks
     end
 
     # Executes the plugins.rb file
@@ -228,6 +247,8 @@ module Bundler
       add_to_load_path(index.load_paths(name))
 
       load path.join(PLUGIN_FILE_NAME)
+
+      @loaded << name
     rescue => e
       Bundler.ui.error "Failed loading plugin #{name}: #{e.message}"
       raise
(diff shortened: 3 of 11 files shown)


lib/bundler.rb
```diff
@@ -393,6 +393,7 @@ def reset!
       rubygems.undo_replacements
       rubygems.reset
       @rubygems = nil
+      Plugin.reset!
```
segiddins: this needs to go before the early return

lib/bundler/cli/install.rb
```diff
@@ -68,6 +68,8 @@ def run
       definition = Bundler.definition
       definition.validate_ruby!
 
+      Plugin.hook("before-install-all", definition.dependencies)
```
segiddins: this should go in the installer, rather than the CLI

lib/bundler/plugin.rb
```diff
@@ -121,13 +123,35 @@ def source_from_lock(locked_opts)
...
+    end
+
     # currently only intended for specs
     #
     # @return [String, nil] installed path
     def installed?(plugin)
       Index.new.installed?(plugin)
     end
 
+    def reset!
```
segiddins: maybe just call `reset!` instead of having the separate ivar initialization above?

lib/bundler/plugin.rb
```diff
@@ -121,13 +123,35 @@ def source_from_lock(locked_opts)
       src.new(locked_opts.merge("uri" => locked_opts["remote"]))
     end
 
+    def add_hook(event, &block)
+      (@hooks[event.to_s] ||= []) << block
```
segiddins: instead of this, maybe define hooks as `@hooks = Hash.new {|h, k| h[k] = [] }`?

lib/bundler/plugin.rb
```diff
@@ -18,6 +18,8 @@ class UnknownSourceError < PluginError; end
 
     @commands = {}
     @sources = {}
+    @hooks = {}
```
segiddins: `hooks_by_event`, maybe?

lib/bundler/plugin.rb
```diff
@@ -18,6 +18,8 @@ class UnknownSourceError < PluginError; end
...
+    @loaded = []
```
segiddins: `@loaded_plugin_names` ?

lib/bundler/plugin.rb
```diff
@@ -184,12 +210,14 @@ def register_plugin(name, spec, optional_plugin = false)
         Bundler.rm_rf(path)
         false
       else
-        index.register_plugin name, path.to_s, load_paths, @commands.keys, @sources.keys
+        index.register_plugin name, path.to_s, load_paths, @commands.keys,
```
segiddins: please parenthesize the `register_plugin` method call

lib/bundler/plugin/index.rb
```diff
@@ -87,6 +90,10 @@ def source_plugin(name)
         @sources[name]
       end
 
+      def hook_plugins(event)
+        @hooks[event]
```
segiddins: imo this should always return an array, even if it's empty

lib/bundler/plugin/index.rb
```diff
@@ -111,9 +119,10 @@ def load_index
       def save_index
         index = {
           "plugin_paths" => @plugin_paths,
-          "load_paths" => @load_paths,
-          "commands" => @commands,
-          "sources" => @sources,
+          "load_paths"   => @load_paths,
+          "commands"     => @commands,
+          "sources"      => @sources,
+          "hooks"        => @hooks,
```
segiddins: please sort the hash

lib/bundler/plugin.rb
```diff
@@ -121,13 +123,35 @@ def source_from_lock(locked_opts)
...
+      (@hooks[event.to_s] ||= []) << block
+    end
+
+    def hook(event, *args)
+      plugins = index.hook_plugins(event)
+      return unless plugins
+
+      (plugins - @loaded).each {|name| load_plugin(name) }
+
+      @hooks[event].each {|blk| blk.call(*args) }
```
segiddins: should this also allow passing a block to hooks?

lib/bundler/cli/install.rb
asutoshpalai: Should it be called before resolve or after resolve?

lib/bundler/plugin.rb
asutoshpalai: There where 3 places this hash was getting initialized, so I skipped this.

With calling `reset!` I think this can be used.

lib/bundler/cli/install.rb
segiddins: that I'm not sure on

lib/bundler/plugin/index.rb
asutoshpalai: Fixed with default block during initialization.

segiddins:
This needs to go behind the feature flag

asutoshpalai:
Since the hooks will be called from many places, I added the feature flag in the `Plugin#hook` method. This will make removing the flag easy.

indirect:
@segiddins r?

lib/bundler/plugin.rb
```diff
@@ -30,7 +38,7 @@ def install(names, options)
       save_plugins names, specs
     rescue PluginError => e
       specs.values.map {|spec| Bundler.rm_rf(spec.full_gem_path) } if specs
-      Bundler.ui.error "Failed to install plugin #{name}: #{e.message}\n  #{e.backtrace[0]}"
+      Bundler.ui.error "Failed to install plugin #{name}: #{e.message}\n  #{e.backtrace.join("\n")}"
```
segiddins: `.join("\n  ")`

lib/bundler/plugin.rb
```diff
@@ -166,9 +197,11 @@ def validate_plugin!(plugin_path)
     def register_plugin(name, spec, optional_plugin = false)
       commands = @commands
       sources = @sources
+      hooks = @hooks_by_event
 
       @commands = {}
       @sources = {}
+      @hooks_by_event = Hash.new {|h, k| h[k] = [] }
```
segiddins: this looks like it's duplicating the behavior of `reset!` ?

lib/bundler/plugin/index.rb
```diff
@@ -24,6 +24,7 @@ def initialize
         @plugin_paths = {}
         @commands = {}
         @sources = {}
+        @hooks = {}
```
segiddins: maybe make this `Hash.new {|h, k| h[k] = [] }` instead of all the `|| []` ?

asutoshpalai: Actually I did that and then reversed it (ebfebc5).

It was leading to empty hashes in the serialized index file for the hooks that are not registered but were being queried .

lib/bundler/plugin.rb
asutoshpalai: `reset!` also removes all other instance variables along with `@loaded_plugin_names` .  Calling `reset!` may have undesirable side effects.

Also I don't think this list will expand where as reset may expand.

lib/bundler/plugin/index.rb
segiddins: 👍 as long as there's a test for that scenario

indirect:
@segiddins is this ready now?

segiddins:
@homu r+

indirect:
@asutoshpalai can you rebase this so we can merge it? Thanks!

segiddins:
@homu r+

PR #4816, opened 2016-07-26
[Graph] Remove monkey-patching of Gem::Dependency

segiddins:


segiddins:
@indirect this now includes the new matcher goodness

indirect:
I like this! How do you feel about `expect(installed_bundle).to include("rack 1.0.0")`?

segiddins:
@indirect we need to use a custom matchers for the pretty error messages, so I don't think we can use include

indirect:
@segiddins oh right good point... I should have written more. I can see two things we want to do with this matcher, 1) check for installed/uninstalled gems in the bundler context (whether that is `--path` or `--system`) and 2) check for installed/uninstalled gems in the rubygems context (e.g. in `Gem.path`, regardless of Bundler config). I foresee these use cases for the matcher:

1. `expect(bundled_gems).to contain("rack 1.0.0")`
1. `expect(bundle_at(bundle_path)).to contain("rack 1.0.0")`
1. `expect(system_gems).to contain("rack 1.0.0")`

I think that `expect(the_bundle).to have_installed` reads kind of awkwardly, so I tried to come up with other options. That said, I'm not super attached to any of `bundled_gems`, `bundle_at`, `system_gems`, or `have_installed`... do you have naming opinions?

segiddins:
1,2 are already covered by this PR -- (`:bundle_dir` option)
3 shouldn't be too hard to do, but it doesn't exist yet and probably shouldn't block this work being merged

I personally like the way `expect(the_bundle)` reads, and the `have_installed` is a combination of the prior matcher name and the need to not conflict with any of the other RSpec matchers

indirect:
hooray :tada: @homu r+

segiddins:
@homu retry

PR #4822, opened 2016-07-27
Feature-flag the postit trampoline, disabled by default

### LINKED ISSUE #4821: Move trampolining behind a feature flag
New plan: since the compact index and ruby locking need some fixes to be fully stable and reliable, we're going to release master as 1.13 with both plugins and trampolining behind feature flags. (h/t @segiddins). That way everyone can use the improvements we've made over 1.12 without having to worry about issues from features that aren't fully finished yet. It's not set in stone, but I think the trampoline flag should be an env var named BUNDLE_ENABLE_TRAMPOLINE.

segiddins:
Set ENV["BUNDLE_ENABLE_TRAMPOLINE"] to enable it

Closes #4821

### DIFF
--- a/lib/bundler/postit_trampoline.rb
+++ b/lib/bundler/postit_trampoline.rb
@@ -1,62 +1,65 @@
 # frozen_string_literal: true
+if ENV["BUNDLE_ENABLE_TRAMPOLINE"]
 
-module BundlerVendoredPostIt; end
-require "bundler/vendor/postit/lib/postit"
-require "rubygems"
+  module BundlerVendoredPostIt; end
+  require "bundler/vendor/postit/lib/postit"
+  require "rubygems"
 
-environment = BundlerVendoredPostIt::PostIt::Environment.new([])
-version = Gem::Requirement.new(environment.bundler_version)
+  environment = BundlerVendoredPostIt::PostIt::Environment.new([])
+  version = Gem::Requirement.new(environment.bundler_version)
 
-installed_version =
-  if defined?(Bundler::VERSION)
-    Bundler::VERSION
-  else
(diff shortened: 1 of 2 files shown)


indirect:
@homu r+

Supernats:
@segiddins I realize I'm bringing this up many years after the PR has been merged and now that v2 is live, it's unlikely any changes near this section of code are desired. This is more of a question about what to expect from Bundler.

From what I can see in the patch as well as what I've been able to confirm on my local machine, in all pre-2 versions, it is only possible to enable the trampoline by setting the `BUNDLE_ENABLE_TRAMPOLINE` (later `BUNDLE_TRAMPOLINE_FORCE`) environment variables. If one then runs `bundle config`, a stanza for `trampoline_force` shows up, as would be expected. At this point, the trampoline executes as expected.

If one then sets `BUNDLE_TRAMPOLINE_FORCE` in the local bundle config and unsets the environment variable, `bundle config` still shows the same value for `trampoline_force`, just with a different source. At this point, if one tries any bundle command, the trampoline will not execute.

Here is my question: Why does this feature flag seemingly contradict the Bundler configuration norms set by [the `bundle config` page](https://bundler.io/v1.16/bundle_config.html)? Are there other settings that don't follow the lookup described by `bundle config`

In case you are curious about my use case, I was hoping to commit a `.bundle/config` to our repository to help standardize our Bundler version across developers without making sure everyone updated.

If this comes across as an attack, I apologize deeply and would love to know what specifically I've said to make it feel that way.

Thanks for all the work you and the rest of the Bundler team do to keep this critical piece of Ruby infrastructure working for the community.

segiddins:
This feature was reverted from Bundler and reimplemented in RubyGems 3

Supernats:
Ah, thanks for very much for the insight.

PR #4824, opened 2016-07-30
Allow signed tags on rake release

spk:
Removing `--annotate` option does not change default behaviour

Since git 2.9.0 its possible to configure signed tag with
`tag.forceSignAnnotated` to true

Cheers


segiddins:
@indirect r?

indirect:
It doesn't actually do anything, but sure. @homu r+

PR #4831, opened 2016-08-02
Allow installing / updating when everything necessary is available locally

segiddins:
This should ideally have some tests, but right now it's behind a feature flag


segiddins:
Specs added

indirect:
Awesome.

@homu r+

segiddins:
@homu retry

PR #4832, opened 2016-08-02
Remove Bundler::Environment

segiddins:
Pretty straightforward refactoring


lib/bundler/installer.rb
```diff
@@ -56,12 +61,13 @@ def self.install(root, definition, options = {})
     # Finally: TODO add documentation for how the standalone process works.
     def run(options)
       create_bundle_path
+      remove_0_9_environment_rb
```
indirect: please, please trash this 😬

indirect:
Pretty ridiculous that these classes only shared an `initialize` method 😛

lib/bundler/installer.rb
segiddins: there's a test for the 0.9 upgrade behavior, so to adhere to semver we should probably wait until 2.0 to do so

segiddins:
Fix'd

indirect:
@homu r+

PR #4834, opened 2016-08-02
[Install] Print errors at the end of installation instead of inline

segiddins:
This needs test coverage


segiddins:
This should no longer treat every install as an error

segiddins:
Fix'd

indirect:
@homu r+

PR #4836, opened 2016-08-03
Resolve for specific platforms

### LINKED ISSUE #4295: bundle package flag --all-platforms has no effect
When i run `bundle package --all-platforms` I would expect `vendor/cache` to be populated with `.gem` for all platforms, however I only get the current platform gem.

```
# bundle env
Environment

    Bundler   1.11.2
    Rubygems  2.5.1
    Ruby      2.3.0p0 (2015-12-25 revision 53290) [x86_64-linux]
    GEM_HOME  /usr/local/bundle
    GEM_PATH
    Git       2.1.4

Bundler settings

    path
      Set for the current user (/root/.bundle/config): "/usr/local/bundle"
    bin
      Set for the current user (/root/.bundle/config): "/usr/local/bundle/bin"
    silence_root_warning
      Set for the current user (/root/.bundle/config): "true"
    app_config
      Set via BUNDLE_APP_CONFIG: "/usr/local/bundle"

Gemfile

    source 'https://rubygems.org'

    gem 'google-protobuf', '3.0.0.alpha.5.0.3'

Gemfile.lock

    GEM
      remote: https://rubygems.org/
      specs:
        google-protobuf (3.0.0.alpha.5.0.3)

    PLATFORMS
      ruby

    DEPENDENCIES
      google-protobuf (= 3.0.0.alpha.5.0.3)

    BUNDLED WITH
       1.11.2
```

and

```
# bundle package --all-platforms --verbose
HTTP GET https://bundler.rubygems.org/api/v1/dependencies
HTTP 200 OK
Query List: ["google-protobuf"]
Query Gemcutter Dependency Endpoint API: google-protobuf
HTTP GET https://bundler.rubygems.org/api/v1/dependencies?gems=google-protobuf
HTTP 200 OK
Query List: []
Query List: []
Installing google-protobuf 3.0.0.alpha.5.0.3
0:  google-protobuf (3.0.0.alpha.5.0.3) from /usr/local/bundle/specifications/google-protobuf-3.0.0.alpha.5.0.3-x86_64-linux.gemspec
Using bundler 1.11.2
0:  bundler (1.11.2) from /usr/local/bundle/specifications/bundler-1.11.2.gemspec
Bundled gems are installed into /usr/local/bundle.
Updating files in vendor/cache
  * google-protobuf-3.0.0.alpha.5.0.3-x86_64-linux.gem
```

My ultimate desire would be to just get the `ruby` platform version and build from source, but I thought that `--all-platforms` would be the next best alternative. I couldn't find this reported elsewhere (but was mentioned in https://github.com/bundler/bundler-features/issues/4#issuecomment-76508244)

segiddins:
Closes #4295.

This will require adding a bunch of tests, as well as figuring out how to put this new behavior behind a feature flag (thus fixing all of the existing tests).

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -104,9 +104,7 @@ module Bundler
 
       @gem_version_promoter = create_gem_version_promoter
 
-      current_platform = Bundler.rubygems.platforms.last
-      add_platform(current_platform) if Bundler.settings[:specific_platform]
-      add_platform(generic(current_platform))
+      add_current_platform unless Bundler.settings[:frozen]
 
       @path_changes = converge_paths
       eager_unlock = expand_dependencies(@unlock[:gems])
@@ -503,6 +501,12 @@ module Bundler
     def remove_platform(platform)
       return if @platforms.delete(Gem::Platform.new(platform))
       raise InvalidOption, "Unable to remove the platform `#{platform}` since the only platforms are #{@platforms.join ", "}"
+    end
+
+    def add_current_platform
+      current_platform = Bundler.rubygems.platforms.last
+      add_platform(current_platform) if Bundler.settings[:specific_platform]
+      add_platform(generic(current_platform))
     end
 
     attr_reader :sources
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -70,7 +70,8 @@ module Bundler
       if lockfile && File.exist?(lockfile)
         @lockfile_contents = Bundler.read_file(lockfile)
         @locked_gems = LockfileParser.new(@lockfile_contents)
-        @platforms = @locked_gems.platforms
+        @locked_platforms = @locked_gems.platforms
+        @platforms = @locked_platforms.dup
         @locked_bundler_version = @locked_gems.bundler_version
         @locked_ruby_version = @locked_gems.ruby_version
 
@@ -91,6 +92,7 @@ module Bundler
         @locked_deps    = []
         @locked_specs   = SpecSet.new([])
         @locked_sources = []
+        @locked_platforms = []
       end
 
       @unlock[:gems] ||= []
@@ -102,8 +104,9 @@ module Bundler
 
       @gem_version_promoter = create_gem_version_promoter
 
-      current_platform = Bundler.rubygems.platforms.map {|p| generic(p) }.compact.last
-      add_platform(current_platform)
+      current_platform = Bundler.rubygems.platforms.last
+      add_platform(current_platform) if Bundler.settings[:specific_platform]
+      add_platform(generic(current_platform))
 
       @path_changes = converge_paths
       eager_unlock = expand_dependencies(@unlock[:gems])
@@ -413,6 +416,11 @@ module Bundler
       added =   []
       deleted = []
       changed = []
+
+      new_platforms = @platforms - @locked_platforms
+      deleted_platforms = @locked_platforms - @platforms
+      added.concat new_platforms.map {|p| "* platform: #{p}" }
+      deleted.concat deleted_platforms.map {|p| "* platform: #{p}" }
 
       gemfile_sources = sources.lock_sources
 
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -468,6 +468,11 @@ module Bundler
       raise ProductionError, msg if added.any? || deleted.any? || changed.any?
     end
 
+    def validate_runtime!
+      validate_ruby!
+      validate_platforms!
+    end
+
     def validate_ruby!
       return unless ruby_version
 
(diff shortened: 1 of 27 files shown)


lib/bundler/dependency.rb
```diff
@@ -52,7 +52,7 @@ class Dependency < Gem::Dependency
       :x64_mingw_20 => Gem::Platform::X64_MINGW,
       :x64_mingw_21 => Gem::Platform::X64_MINGW,
       :x64_mingw_22 => Gem::Platform::X64_MINGW,
-      :x64_mingw_23 => Gem::Platform::X64_MINGW
+      :x64_mingw_23 => Gem::Platform::X64_MINGW,
```
indirect: really? we're doing trailing commas now? 😝

lib/bundler/rubygems_ext.rb
```diff
@@ -167,6 +167,9 @@ class Platform
     MSWIN64 = Gem::Platform.new("mswin64") unless defined?(MSWIN64)
     MINGW = Gem::Platform.new("x86-mingw32") unless defined?(MINGW)
     X64_MINGW = Gem::Platform.new("x64-mingw32") unless defined?(X64_MINGW)
+    # DARWIN = Gem::Platform.new("universal-darwin") unless defined?(DARWIN)
+    # LINUX = Gem::Platform.new("x86-linux") unless defined?(LINUX)
+    # X64_LINUX = Gem::Platform.new("x86_64-linux") unless defined?(X64_LINUX)
```
indirect: commented out but included deliberately?

lib/bundler/source.rb
```diff
@@ -14,6 +14,7 @@ def unmet_deps
 
     def version_message(spec)
       message = "#{spec.name} #{spec.version}"
+      message += " (#{spec.platform})" if Bundler.ui.debug? && spec.platform != Gem::Platform::RUBY
```
indirect: we probably want this always, even when debug is off

lib/bundler/spec_set.rb
```diff
@@ -99,6 +99,17 @@ def materialize(deps, missing_specs = nil)
       SpecSet.new(materialized.compact)
     end
 
+    def materialized_for_all_platforms
```
indirect: is the all platforms here every platform currently listed in the gemfile?

lib/bundler/gem_helpers.rb
```diff
@@ -28,5 +28,38 @@ def generic_local_platform
...
+          1
+        end,
+
+        if spec_platform.version == user_platform.version
+          0
+        elsif spec_platform.version.nil?
+          1
+        else
+          2
+        end,
```
indirect: please turn each of these into comparison methods or something 😎

indirect:
omg this is the most hopeful I have felt about platforms in literally five years, no exaggeration. hooray :tada:

lib/bundler/dependency.rb
segiddins: we have been since introducing rubocop

lib/bundler/rubygems_ext.rb
segiddins: nope, part of the process to get to this point, it can be deleted

lib/bundler/source.rb
segiddins: 👍 I was just trying to avoid changing the output

lib/bundler/spec_set.rb
segiddins: it's just all the specs in the spec set, regardless of what platform they're for, in contrast to how `for` does platform filtering (and specifically different from how `materialize` calls `for` only for the current platform)

lib/bundler/dependency.rb
indirect: :sob:

segiddins: this way we don't have to touch that line when adding a new element to the array

indirect: what good is it to be in charge of a project if we adopt code style that clashes with my personal coding aesthetic 😝

indirect:
@homu r+

indirect:
@homu r-

indirect:
@homu r+

indirect:
@homu r-

indirect:
(Waiting to merge this until after we stable-branch 1.13)

mwrock:
This fixes #4896

lib/bundler/definition.rb
```diff
@@ -105,8 +107,9 @@ def initialize(lockfile, dependencies, sources, unlock, ruby_version = nil, opti
 
       @gem_version_promoter = create_gem_version_promoter
 
-      current_platform = Bundler.rubygems.platforms.map {|p| generic(p) }.compact.last
-      add_platform(current_platform)
+      current_platform = Bundler.rubygems.platforms.last
+      add_platform(current_platform) if Bundler.settings[:specific_platform]
```
segiddins: we'll probably need to special-case around when in deployment mode and not add any platforms and count on the platforms check in https://github.com/bundler/bundler/pull/4895

indirect:
@homu r+

segiddins:
@homu r- until I address my last comment

segiddins:
@indirect this should finally be mergable

indirect:
hooray 🏅 @homu r+

segiddins:
Rebase'd
@homu r=indirect

PR #4840, opened 2016-08-04
Unlock sources when a local override leads to changes

### LINKED ISSUE #4838: Changing dependencies in a gemspec causes "corrupt Gemfile.lock" :/
If you download [this zip file](http://files.arko.net/2e3E1p1Q1h1w/corrupt-lock-repro.zip), decompress it, cd into the directory, and run `./demo.sh`, you will get this error:

> Your Gemfile.lock is corrupt. The following gems are missing from the DEPENDENCIES section: 'twitter' 'oauth'.

The problem with that error is that all I did was add two gem dependencies to the gemspec of `lita-tweet`, which is depended on by `rubytogether-lita`. :/ I am pretty sure this should not trigger a corrupt lockfile error, and should instead install the newly-added gems and update the lockfile.

Maybe a GSoC student suitable bug? /cc @segiddins @RochesterinNYC

segiddins:
Closes #4838.
Requires tests

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -555,9 +555,10 @@ module Bundler
         end
       end
 
-      locals.any? do |source, changed|
+      sources_with_changes = locals.select do |source, changed|
         changed || specs_changed?(source)
-      end
+      end.map(&:first)
+      !sources_with_changes.each {|source| @unlock[:sources] << source.name }.empty?
     end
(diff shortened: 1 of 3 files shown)


indirect:
yay

segiddins:
Spec added

indirect:
@homu r+

Issue #4772, opened 2016-07-09
Change outdated to use the new bundle update flags and conservative resolution code


chrismo:
Task spawned from #4676

chrismo:
saw this PR which would cause conflicts if work is done at the same time. #4841

chrismo:
there's also `bundle show --outdated` -- same stuff?

PR #4841, opened 2016-08-05
Better gem outdated list

ryanfox1985:
Hi,

the current version of bundler is listing all gems with the group, I improve these list grouping the gems by group and not repeat the group every line.

Current:
```
  * aasm (newest 4.11.0, installed 3.4.0, requested = 3.4.0) in group "default"
  * activerecord-import (newest 0.15.0, installed 0.13.0) in group "default"
  * annotate (newest 2.7.1, installed 2.7.0) in group "development"
  * authority (newest 3.2.0, installed 3.0.0, requested ~> 3.0.0) in group "default"
  * awesome_print (newest 1.7.0, installed 1.2.0, requested ~> 1.2.0) in groups "development, test"
  * aws-sdk (newest 2.5.1, installed 1.60.2, requested ~> 1.60.2) in group "default"
  * bootstrap-sass (newest 3.3.7, installed 3.3.5.1, requested ~> 3.3.5.1) in group "default"
  * capybara (newest 2.7.1, installed 2.5.0, requested ~> 2.4) in group "test"
  * capybara-email (newest 2.5.0, installed 2.4.0) in group "test"
  * codeclimate-test-reporter (newest 0.6.0, installed 0.4.7) in group "test"
  * coffee-rails (newest 4.2.1, installed 4.0.1, requested ~> 4.0.0) in group "default"
  * countries (newest 1.2.5, installed 0.9.3, requested ~> 0.9.3) in group "default"
  * country_select (newest 2.5.2, installed 2.1.1) in group "default"
  * dalli (newest 2.7.6, installed 2.7.4) in group "default"

```

Better group:
```
$ bundle outdated

===== Group default =====
  * addressable (newest 2.4.0, installed 2.3.8)
  * puma (newest 3.6.0, installed 2.10.2, requested ~> 2.10.2)
  * rails-assets-Sortable (newest 1.4.2, installed 0.7.1, requested ~> 0.7.1)
  * rails-assets-jquery-file-upload (newest 9.12.5, installed 9.7.0, requested = 9.7.0)
  * rails-assets-jquery-ui (newest 1.12.0, installed 1.11.0, requested = 1.11.0)
  * rails-assets-superagent (newest 2.1.0, installed 0.21.0, requested ~> 0.21.0)
  * rollbar (newest 2.12.0, installed 2.7.1, requested ~> 2.0)
  * sidekiq-pro (newest 3.3.2, installed 3.3.1)
===== Group development =====
  * brakeman (newest 3.3.3, installed 3.1.0)
===== Group development, test =====
  * pry (newest 0.10.4, installed 0.10.3)
  * pry-doc (newest 0.9.0, installed 0.8.0)
  * rspec (newest 3.5.0, installed 3.3.0, requested ~> 3.0)
  * rspec-support (newest 3.5.0, installed 3.3.0)
  * rubocop-rspec (newest 1.4.0, installed 1.3.1)
===== Without group =====
  * babel-source (newest 5.8.35, installed 5.8.34)
  * execjs (newest 2.7.0, installed 2.6.0)
  * hashie (newest 3.4.4, installed 3.3.1)
  * ice_nine (newest 0.11.2, installed 0.11.1)
  * mini_portile2 (newest 2.1.0, installed 2.0.0)
  * minitest (newest 5.9.0, installed 5.8.4)
  * multi_json (newest 1.12.1, installed 1.11.2)
  * multipart-post (newest 2.0.0, installed 1.2.0)
  * rails-assets-blueimp-canvas-to-blob (newest 3.3.0, installed 2.1.1)
  * rails-assets-blueimp-load-image (newest 2.6.1, installed 1.13.0)
  * rails-assets-blueimp-tmpl (newest 3.4.0, installed 2.5.4)
  * rails-assets-jquery (newest 3.1.0, installed 2.1.4)
  * rake (newest 11.2.2, installed 10.5.0)
  * rdoc (newest 4.2.2, installed 4.2.0)
```

Regards

### DIFF
--- a/lib/bundler/cli/outdated.rb
+++ b/lib/bundler/cli/outdated.rb
@@ -4,6 +4,7 @@ require "bundler/cli/common"
 module Bundler
   class CLI::Outdated
     attr_reader :options, :gems
+
     def initialize(options, gems)
       @options = options
       @gems = gems
@@ -30,7 +31,7 @@ module Bundler
         Bundler.definition(:gems => gems, :sources => sources)
       end
 
-      definition_resolution = proc { options["local"] ? definition.resolve_with_cache! : definition.resolve_remotely! }
+      definition_resolution = proc { options[:local] ? definition.resolve_with_cache! : definition.resolve_remotely! }
       if options[:parseable]
         Bundler.ui.silence(&definition_resolution)
       else
@@ -38,8 +39,9 @@ module Bundler
       end
 
       Bundler.ui.info ""
+      outdated_gems_by_groups = {}
+      outdated_gems_list = []
 
-      out_count = 0
       # Loop through the current specs
       gemfile_specs, dependency_specs = current_specs.partition {|spec| current_dependencies.key? spec.name }
       [gemfile_specs.sort_by(&:name), dependency_specs.sort_by(&:name)].flatten.each do |current_spec|
@@ -47,7 +49,7 @@ module Bundler
 
         dependency = current_dependencies[current_spec.name]
 
-        if options["strict"]
+        if options[:strict]
           active_spec = definition.specs.detect {|spec| spec.name == current_spec.name && spec.platform == current_spec.platform }
         else
           active_specs = definition.index[current_spec.name].select {|spec| spec.platform == current_spec.platform }.sort_by(&:version)
@@ -67,46 +69,88 @@ module Bundler
         gem_outdated = Gem::Version.new(active_spec.version) > Gem::Version.new(current_spec.version)
         git_outdated = current_spec.git_version != active_spec.git_version
         if gem_outdated || git_outdated
-          unless options[:parseable]
-            if out_count == 0
-              if options["pre"]
-                Bundler.ui.info "Outdated gems included in the bundle (including pre-releases):"
-              else
-                Bundler.ui.info "Outdated gems included in the bundle:"
-              end
-            end
-          end
-
-          spec_version    = "#{active_spec.version}#{active_spec.git_version}"
-          current_version = "#{current_spec.version}#{current_spec.git_version}"
-          dependency_version = %(, requested #{dependency.requirement}) if dependency && dependency.specific?
-
+          groups = nil
           if dependency && !options[:parseable]
             groups = dependency.groups.join(", ")
-            pl = (dependency.groups.length > 1) ? "s" : ""
-            groups = " in group#{pl} \"#{groups}\""
           end
 
-          spec_outdated_info = "#{active_spec.name} (newest #{spec_version}, installed #{current_version}#{dependency_version})"
-          if options[:parseable]
-            Bundler.ui.info spec_outdated_info.to_s.rstrip
-          else
-            Bundler.ui.info "  * #{spec_outdated_info}#{groups}".rstrip
-          end
+          outdated_gems_list << { :active_spec => active_spec,
+                                  :current_spec => current_spec,
+                                  :dependency => dependency,
+                                  :groups => groups }
 
-          out_count += 1
+          outdated_gems_by_groups[groups] ||= []
+          outdated_gems_by_groups[groups] << { :active_spec => active_spec,
+                                               :current_spec => current_spec,
+                                               :dependency => dependency,
+                                               :groups => groups }
         end
+
         Bundler.ui.debug "from #{active_spec.loaded_from}"
       end
 
-      if out_count.zero?
+      if outdated_gems_list.empty?
         Bundler.ui.info "Bundle up to date!\n" unless options[:parseable]
       else
+        unless options[:parseable]
+          if options[:pre]
+            Bundler.ui.info "Outdated gems included in the bundle (including pre-releases):"
+          else
+            Bundler.ui.info "Outdated gems included in the bundle:"
+          end
+        end
+
+        options_include_groups = [:group, :groups].select {|v| options.keys.include?(v.to_s) }
+        if options_include_groups.any?
+          ordered_groups = outdated_gems_by_groups.keys.compact.sort
+          [nil, ordered_groups].flatten.each do |groups|
+            gems = outdated_gems_by_groups[groups]
+            contains_group = if groups
+              groups.split(",").include?(options[:group])
+            else
+              options[:group] == "group"
+            end
+
+            next if (!options[:groups] && !contains_group) || gems.nil?
+
+            unless options[:parseable]
+              if groups
+                Bundler.ui.info "===== Group #{groups} ====="
+              else
+                Bundler.ui.info "===== Without group ====="
+              end
+            end
+
+            gems.each do |gem|
+              print_gem(gem[:current_spec], gem[:active_spec], gem[:dependency], groups, options_include_groups.any?)
+            end
+          end
+        else
+          outdated_gems_list.each do |gem|
+            print_gem(gem[:current_spec], gem[:active_spec], gem[:dependency], gem[:groups], options_include_groups.any?)
+          end
+        end
+
         exit 1
       end
     end
 
   private
+
+    def print_gem(current_spec, active_spec, dependency, groups, options_include_groups)
+      spec_version = "#{active_spec.version}#{active_spec.git_version}"
+      current_version = "#{current_spec.version}#{current_spec.git_version}"
+      dependency_version = %(, requested #{dependency.requirement}) if dependency && dependency.specific?
+
+      spec_outdated_info = "#{active_spec.name} (newest #{spec_version}, installed #{current_version}#{dependency_version})"
+      if options[:parseable]
+        Bundler.ui.info spec_outdated_info.to_s.rstrip
+      elsif options_include_groups || !groups
+        Bundler.ui.info "  * #{spec_outdated_info}".rstrip
+      else
+        Bundler.ui.info "  * #{spec_outdated_info} in groups \"#{groups}\"".rstrip
+      end
+    end
 
     def check_for_deployment_mode
       if Bundler.settings[:frozen]
@@ -123,7 +167,6 @@ module Bundler
       active_major = active_spec.version.segments.first
 
       update_present = false
-
       update_present = active_major > current_major if options[:major]
 
       if !update_present && (options[:minor] || options[:patch]) && current_major == active_major
--- a/spec/commands/outdated_spec.rb
+++ b/spec/commands/outdated_spec.rb
@@ -72,6 +72,120 @@ describe "bundle outdated" do
     end
   end
 
+  describe "with --group option" do
+    def test_group_option(group = nil, gems_list_size = 1)
+      install_gemfile <<-G
+        source "file://#{gem_repo2}"
+
+        gem "weakling", "~> 0.0.1"
+        gem "terranova", '8'
+        group :development, :test do
+          gem "duradura", '7.0'
+          gem 'activesupport', '2.3.5'
+        end
+      G
+
+      update_repo2 do
+        build_gem "activesupport", "3.0"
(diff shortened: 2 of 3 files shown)


indirect:
Thanks for sending us a PR! The current output lists every gem alphabetically so that you can find a specific gem easily. If we separated each group, that would become impossible unless you already know beforehand exactly which group the gem is in. This could work as either an option (maybe `--groups`?) or as a plugin that adds a command (maybe `bundle outdated-groups`). Are you interested in doing one of those things?

ryanfox1985:
Hello @indirect,

thanks for the feedback! I think both of your proposals will duplicate and get dirty the code.
I understand that I modify the current list alphabetically and in my PR is listing in groups (alphabetically the gems inside).

In my case I have my Gemfile disordered and I check some repos and anybody is respecting the alphabetically in their Gemfile.

- [rails](https://github.com/rails/rails/blob/master/Gemfile)
- [rubocop](https://github.com/bbatsov/rubocop/blob/master/Gemfile)
- [annotate_models](https://github.com/ctran/annotate_models/blob/develop/Gemfile)

I don't understand: What advantage do you have with the outdated list alphabetically ordered? 

In my case my application doesn't have 100% of coverage. My updating process is checks the gems outdated and if the environment is development or test I try upgrade executing the test. If the environment is production I do more safe trying some parts of the application or adding more test.

With this PR I can check my outdated list by groups quickly and it saves space in the rows.

Regards.

ryanfox1985:
Sorry, rebasing the last changes from master I closed by error the pull request. XD

indirect:
I see that this change makes Bundler better for you, but not everyone agrees. It is not okay to force everyone to this new output, so this change is only okay as an option to outdated or as a separate command. If you do not want to make this an option, I am sorry but we cannot accept it.

I think this would be useful if it worked like `bundle outdated --groups test development` to only show outdated gems in the `test` and `development groups. To show all groups, you would run `bundle outdated --groups`.

ryanfox1985:
Hello @indirect,

I understand, I restored the outdated command and I added a option to outdated: 
`bundle outdated --groups`

But I have a problem with the option, I tried to add an method_option here:
https://github.com/AirRefund/bundler/blob/master/lib/bundler/cli.rb#L272
but it doesn't appear in `bundle outdated --help` and it not working.

Also I will restore the test and I will create a new one with `--groups`. 

Regards.

ryanfox1985:
Hello,

I already solved with aliases.

Regards.

ryanfox1985:
Hello @indirect,

I restored the outdated behavior `alphabetically order`. 
I added 2 options more:
- `--groups`: list the gems grouped by groups. Also the groups are ordered alphabetically.
- `--group`: list the gems of a specific group.

In both options the gems are ordered alphabetically inside the group.

Regards.

ryanfox1985:
Hello @chrismo, 

I don't understand where is the conflict.

Regards.

ryanfox1985:
Hello @chrismo,

executing `bundle show --outdated` I don't see the same output.

Regards.

chrismo:
@ryanfox1985 sorry haven't replied yet. 1.13 has new, un-official options in bundle update to restrict updates to only the latest `--patch` or `--minor` version available (see https://github.com/bundler/bundler-features/issues/122). `bundle outdated` had similarly named options added to it previously, but they don't behave in the same way. 

For example, if bundle outdated is going to show:

foo: 1.0.0 => 2.0.0
bar: 1.0.0 => 1.1.0
qux: 1.0.0 => 1.0.1

and you add the `--patch` flag, then it merely filters the results to only show:

qux: 1.0.0 => 1.0.1

rather than showing what could happen as a result of bundle update --patch:

foo: 1.0.0 => 1.0.3
bar: 1.0.0 => 1.0.9
qux: 1.0.0 => 1.0.1

#4772 is a task I made to update bundle outdated to have matching behavior, and when I saw this PR, I simply linked to it to remind me that this PR would be changing code in a similar place.

ryanfox1985:
Hello @chrismo,

I have some doubts in my code: 
1- I'm calling two times to `Bundle.load`:
https://github.com/AirRefund/bundler/blob/master/lib/bundler/cli/outdated.rb#L22-L43

2- I removed `options[:pre]`, is it already managed by `gem_version_promoter`?  https://github.com/bundler/bundler/pull/4841/commits/88748672fc2d455eb997e50bf5a98d1857894a32#diff-1dacde23b0f908c590b785187df11b00L55

3- Is it possible that sometimes the `git_version` is not present in the `active_spec`?
https://github.com/bundler/bundler/pull/4841/commits/88748672fc2d455eb997e50bf5a98d1857894a32#diff-1dacde23b0f908c590b785187df11b00R68

Regards.

chrismo:
First of all, @ryanfox1985, I hadn't seen you're already working on including the new conservative update behavior in here, so that's awesome :) thx!

1 - `Bundler.load` is memoized (like many `Bundler::*` calls) so it's safe to call multiple times (though the trade off for global state is annoying at times)

2 - I'm not seeing where you removed `options[:pre]`? But in any case, GemVersionPromoter doesn't concern itself with that option - that's only an option for this command I believe.

3 - I don't know much about git dependencies, sorry. :(

ryanfox1985:
Hello @chrismo,

can you review the code? and if it's ok I will squash my commits into 1.

Regards.

lib/bundler/cli/outdated.rb
```diff
@@ -30,6 +30,13 @@ def run
         Bundler.definition(:gems => gems, :sources => sources)
       end
 
+      patch_level = [:major, :minor, :patch].select {|v| options.keys.include?(v.to_s) }
+      Bundler.definition.create_gem_version_promoter([]).tap do |gvp|
+        gvp.level = patch_level.first || :major
+        gvp.strict = options[:strict]
+      end
+      Bundler.load
```
chrismo: We need to rebase to master and replace this section with `Bundler::CLI::Common.config_gem_version_promoter(Bundler.definition, options)`

This is going to cause a problem though which is a key part of this change: the meaning of the switches `--patch` / `--minor` / `--major` are now changing. Before they filtered the output, now they'll be influencing the resolution process.

There's only two ways here, though - we either CHANGE what they mean and rename the filtering ones to something like `--filter-patch` (which is breaking), or leave them as-is and introduce new flags (`--resolve-patch`, etc?) that are translated for GemVersionPromoter, which won't be breaking, but will be inconsistent with the flags for `bundle update`. 

My preference is to the former, those new flags weren't introduced too long ago, but I want @indirect's input on this.

lib/bundler/definition.rb
```diff
@@ -134,7 +134,7 @@ def fixup_dependency_types!
       end
     end
 
-    def create_gem_version_promoter
+    def create_gem_version_promoter(unlock_gems = nil)
```
chrismo: i don't see you using this anywhere? can you remove?

lib/bundler/definition.rb
```diff
@@ -144,7 +144,7 @@ def create_gem_version_promoter
         else
           @locked_specs
         end
-      GemVersionPromoter.new(locked_specs, @unlock[:gems])
+      GemVersionPromoter.new(locked_specs, unlock_gems || @unlock[:gems])
```
chrismo: ...and remove this too

lib/bundler/cli/outdated.rb
```diff
@@ -38,76 +45,129 @@ def run
...
+          if !current_spec.version.prerelease? && !options[:pre] && specs.size > 1
+            specs.delete_if {|b| b.respond_to?(:version) && b.version.prerelease? }
+          else
+            specs
           end
         end
 
-        next if active_spec.nil?
-
+        active_spec = Bundler.definition.gem_version_promoter.sort_versions(dependency, spec_groups).last
```
chrismo: I don't understand this - this is an internal part of the resolution that's already happened inside Definition. It shouldn't be necessary to _re_-sort the versions ... we should just get the resolved dependencies out of the Definition like it used to do.

spec/commands/outdated_spec.rb
```diff
@@ -70,6 +70,22 @@
...
+          gem 'activesupport', '2.3.5'
+        end
+      G
+
+      update_repo2 { build_gem "activesupport", "3.0" }
+
+      bundle "outdated --groups --verbose"
+      expect(out).to include("===== Group development, test =====")
+      expect(out).to include("activesupport (newest 3.0, installed 2.3.5, requested = 2.3.5)")
+    end
```
chrismo: My gut tells me this one spec isn't enough to cover all of the changes in flight here, but I don't have any specifics at the moment.

lib/bundler/cli/outdated.rb
segiddins: Are we _sure_ that will actually be a breaking change?

chrismo: Yeah, cuz the code that's now in Bundler::CLI::Common raises an error if more than one --patch/--minor/--major option is provided, but current outdated code allows you to pass as many as you want.

Because they do different things. See [this comment](https://github.com/bundler/bundler/pull/4841#issuecomment-251122728)

chrismo: The flags as passed to the GemVersionPromoter _change_ resolution. <=1.13.2 of bundle outdated, those flags merely filter the output, with no influence on resolution or what's displayed as newest.

segiddins: ah yeah thats a pretty big change :/

chrismo: It's unfortunate timing. They were only added in 1.12 ... I'm biased, but feel like the affect the flags have on resolution is of greater value, and would be better to keep in sync with how they work fed to the `bundle update` command as of 1.13, and we could add new --filter-patch flags to replace what they do currently. 

IIRC, there was some confusion at the time that Andre perhaps even thought these flags as of 1.12 would be affecting what the newest would resolve to, instead of just filtering the output, so - _if_ I'm remembering correctly, that's also influencing my opinion. 

But, I can be swayed by the breaking behavior argument.

chrismo: also, @ryanfox1985, after all this typing about _my_ stuff, it may make sense to revert this PR to _just_ the grouping code you're trying to contribute. I may be overloading this PR with one too many things.

indirect: IMO, this is what we were trying to do in 1.12, and failed to do because resolving is complicated. :/ I would be okay with this change on the grounds that the previous flags were documented so it sounded like they cause the new resolver aware behavior. 👍

chrismo:
@ryanfox1985 - the more I think about this, I don't think we should be trying to do two things at once here. If you can back out the changes around configuring the GemVersionPromoter, and just go back to your changes to add the group/groups options ... I can do the conservative updates in a separate PR that builds on this one.

Thx for trying to accommodate me and sorry for the hassle.

ryanfox1985:
Hi @chrismo,

I rollback the changes with gemversionpromoter and I created other branch for me to no losing the last changes (gemversionpromoter merged).

I created some test, can you review the changes? and If is ok I will squash my commits into one.

Regards.

chrismo:
looking it over

chrismo:
looks good to me 👍

chrismo:
shoot - in pushing #5056 which re-includes this commit, it lost track of the [prior CI build](https://travis-ci.org/bundler/bundler/builds/165402970) which already passed. It's building again, but ... it's not necessary. anyway, by morning should hopefully be all consistent.

chrismo:
@indirect - this is ready for @homu

segiddins:
@homu delegate=chrismo

chrismo:
@homu r+

PR #4842, opened 2016-08-05
[Setup] Only add bundler to the load path if we havent trampolined

segiddins:
@indirect this should fix the stack overflow bug!


indirect:
@homu r+

Issue #4843, opened 2016-08-05
BUNDLE_MIRROR__URL env var syntax not always valid


opiethehokie:
$ export BUNDLE_MIRROR__HTTPS://RUBYGEMS.ORG="http://rubygems-mirror.org"
-bash: export: `BUNDLE_MIRROR__HTTPS://RUBYGEMS.ORG=http://rubygems-mirror.org': not a valid identifier

$ bash --version
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin15)

\# export BUNDLE_MIRROR__HTTPS://RUBYGEMS.ORG="http://rubygems-mirror.org"
bash: export: `BUNDLE_MIRROR__HTTPS://RUBYGEMS.ORG=http://rubygems-mirror.org': not a valid identifier

\# bash --version
GNU bash, version 4.3.11(1)-release (x86_64-pc-linux-gnu)

indirect:
This is resolved by #4844, thank you for both the report and the PR!

PR #4844, opened 2016-08-05
enables BUNDLE_MIRROR__HOSTNAME environment variables

opiethehokie:
Proposed fix for issue #4843.

I would like to be able to do something like `BUNDLE_MIRROR__FOO__RUBYGEMS__ORG` and have it work the same way as `BUNDLE_MIRROR__HTTPS://FOO__RUBYGEMS__ORG/`

There was some discussion about it already [here](https://groups.google.com/forum/#!topic/ruby-bundler/QlhlTuxoDCc).

I'm sure there are currently some issues with my implementation as this was the first time I looked at the bundler code, but would you accept this change once they are resolved?

### DIFF
--- a/lib/bundler/settings.rb
+++ b/lib/bundler/settings.rb
@@ -289,11 +289,7 @@ module Bundler
     def self.normalize_uri(uri)
       uri = uri.to_s
       uri = "#{uri}/" unless uri =~ %r{/\Z}
-      uri = URI(uri)
-      unless uri.absolute?
-        raise ArgumentError, "Gem sources must be absolute. You provided '#{uri}'."
-      end
-      uri
+      URI(uri)
     end
   end
 end
--- a/lib/bundler/settings.rb
+++ b/lib/bundler/settings.rb
@@ -289,7 +289,11 @@ module Bundler
     def self.normalize_uri(uri)
       uri = uri.to_s
       uri = "#{uri}/" unless uri =~ %r{/\Z}
-      URI(uri)
+      uri = URI(uri)
+      unless uri.absolute?
+        raise ArgumentError, format("Gem sources must be absolute. You provided '%s'.", uri)
+      end
+      uri
     end
   end
 end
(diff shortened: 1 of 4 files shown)


lib/bundler/settings.rb
```diff
@@ -289,11 +289,7 @@ def load_config(config_file)
     def self.normalize_uri(uri)
       uri = uri.to_s
       uri = "#{uri}/" unless uri =~ %r{/\Z}
-      uri = URI(uri)
-      unless uri.absolute?
-        raise ArgumentError, "Gem sources must be absolute. You provided '#{uri}'."
-      end
-      uri
+      URI(uri)
```
indirect: I think this will completely change the way we validate any existing source URLs, and that's probably not okay. Can we do the host extraction in a different method so that we don't change the existing behavior?

indirect:
In your description for this PR, you talk about a mirror for an `https` source, but the code in this PR only works for `http` sources. Is that deliberate?

opiethehokie:
Thanks for taking a look at this, not sure I understand what you mean though. The test I added happened to be for `http` but it seems to be working for me with `https` sources as well. My intent was to have `FOO__RUBYGEMS__ORG` override both http://foo.rubygems.org and https://foo.rubygems.org.

lib/bundler/settings.rb
opiethehokie: Sure I'll look into that.

indirect:
@opiethehokie you're right! HTTPS does work. It's maybe a potential problem that you could use this option to change a source given as HTTPS in the Gemfile to another source that uses HTTP, but I guess if you're using hostname-based mirroring you'll have to know that is a possibility.

This patch looks good, and I'm happy to accept it as soon as it doesn't change source URL validation!

opiethehokie:
@indirect I added a couple more commits, hopefully this addresses your concerns about source URL validation

indirect:
@opiethehokie looks great, thanks!

@homu r+

PR #4845, opened 2016-08-05
[Install] Fetch all symbolic refs (not just branch heads)

jesusaurus:
The intention here is to allow installing a gem from git using a
ref of the form "refs/zuul/<uuid>" (which are used by the zuul
continuous-integration system). There are likely other use cases for
arbitrary refs.

An example Gemfile that would take advantage of this functionality can
be seen at [1]; however, it currently fails to fetch the existing ref
with the error seen at [2].

[1] https://review.openstack.org/#/c/351437/1/Gemfile
[2] http://logs.openstack.org/37/351437/1/check/gate-puppet-log_processor-puppet-lint/7be9b6a/console.html#_2016-08-05_02_06_48_423939

---
This change is [Reviewable](https://reviewable.io/reviews/bundler/bundler/4845)

### DIFF
--- a/lib/bundler/source/git/git_proxy.rb
+++ b/lib/bundler/source/git/git_proxy.rb
@@ -92,7 +92,7 @@ module Bundler
             return if has_revision_cached?
             Bundler.ui.info "Fetching #{URICredentialsFilter.credential_filtered_uri(uri)}"
             in_path do
-              git_retry %(fetch --force --quiet --tags #{uri_escaped_with_configured_credentials} "refs/heads/*:refs/heads/*")
+              git_retry %(fetch --force --quiet --tags #{uri_escaped_with_configured_credentials} "refs/*:refs/*")
             end
           else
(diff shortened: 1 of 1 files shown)


segiddins:
Could this have performance implications?

segiddins:
Also, could you please add a test confirming that non-symbolic refs work? Thanks!

indirect:
This would (for example) have a dramatic impact on any Github-hosted open-source project, since Github creates a named ref for every single PR that has ever been opened. Pulling all named refs will mean fetching a lot of commits that aren't in any branch in the repo. :/

jesusaurus:
I would really like to be able to install gems from arbitrary remote refs in my CI system, but I can't find a way to make this configurable to prevent the performance impact on Github projects.

segiddins:
I think this could happen if you only fetch the named refs asked for in a gemfile?

segiddins:
See https://github.com/bundler/bundler/pull/5624

PR #4849, opened 2016-08-08
[YAMLSerializer] Allow loading with CRLF line endings

### LINKED ISSUE #4435: .bundle/config gets written with CRLF line endings on Windows
This is a problem in my case, since `BUNDLE_PATH` will then contain a ^M at the end, so `bundler install` will create a folder with a ^M in it... :smile:

(This is when using `bundler` on Windows 10 with msys2)

I think the only sane approach would be to _always_ write to `.bundle/config` with LF-only line endings. That's the only universal approach that will work everywhere.

This has the downside that people using e.g. Notepad on Windows will not be able to read the file. I think we can live with that; you should be using better editors than Notepad anyway.

segiddins:
Closes #4435.

\c @asutoshpalai

### DIFF
--- a/lib/bundler/yaml_serializer.rb
+++ b/lib/bundler/yaml_serializer.rb
@@ -52,7 +52,7 @@ module Bundler
       stack = [res]
       last_hash = nil
       last_empty_key = nil
-      str.split("\n").each do |line|
+      str.split(/\r?\n/).each do |line|
         if match = HASH_REGEX.match(line)
           indent, key, _, val = match.captures
           key = convert_to_backward_compatible_key(key)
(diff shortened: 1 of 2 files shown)


indirect:
@segiddins red build

segiddins:
@indirect a spurious failure

indirect:
@homu r+

segiddins:
@homu retry

segiddins:
@homu retry

segiddins:
@homu retry

segiddins:
I have no clue why this keeps on happening 😭

indirect:
@homu retry

perlun:
@segiddins - thanks for sorting this one out! 👏

PR #4851, opened 2016-08-08
[RubyGemsGemInstaller] Validate checksums from the compact index

### LINKED ISSUE #4464: Use the gem hash to throw an exception on mismatch
From here https://github.com/bundler/bundler/issues/4461#issuecomment-214003067

The index format includes a hash of the gem contents, bundler can check those vs local to see if the content is mismatched

segiddins:
Closes #4464

### DIFF
--- a/lib/bundler/rubygems_gem_installer.rb
+++ b/lib/bundler/rubygems_gem_installer.rb
@@ -45,7 +45,7 @@ module Bundler
       case checksum.length
       when 64 then :hexdigest!
       when 44 then :base64digest!
-      else raise InstallError, "The given checksum for #{spec.full_name} is not a valid SHA256 hexdigest nor base64digest"
+      else raise InstallError, "The given checksum for #{spec.full_name} (#{checksum.inspect}) is not a valid SHA256 hexdigest nor base64digest"
       end
     end
   end
--- a/lib/bundler/rubygems_gem_installer.rb
+++ b/lib/bundler/rubygems_gem_installer.rb
@@ -12,5 +12,32 @@ module Bundler
     def check_executable_overwrite(filename)
       # Bundler needs to install gems regardless of binstub overwriting
     end
+
+    def pre_install_checks
+      super && validate_bundler_checksum(options[:bundler_expected_checksum])
+    end
+
+  private
+
+    def validate_bundler_checksum(checksum)
+      return true unless checksum
+      return true unless source = @package.instance_variable_get(:@gem)
+      return true unless source.respond_to?(:with_read_io)
+      digest = source.with_read_io do |io|
+        digest = Digest::SHA256.new
+        digest << io.read(16_384) until io.eof?
+        io.rewind
+        digest.base64digest!
+      end
+      unless digest == checksum
+        raise SecurityError,
+          "The checksum for the downloaded `#{spec.full_name}.gem` did not match " \
+          "the checksum given by the API. This means that the contents of the " \
+          "gem appear to be different from what was uploaded, and could be an indicator of a security issue.\n" \
+          "(The expected SHA256 checksum was #{checksum.inspect}, but the checksum for the downloaded gem was #{digest.inspect}.)\n" \
+          "Bundler cannot continue installing #{spec.name} (#{spec.version})."
+      end
+      true
+    end
   end
 end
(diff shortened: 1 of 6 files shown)


indirect:
woo @homu r+

indirect:
Ugh, of course now that I've approved it I realized that we should add a setting to disable this check, because some people will have existing gems with invalid checksums that they want to keep using.

indirect:
@homu r-

indirect:
I am super happy with this, though!

spec/support/artifice/compact_index.rb
```diff
@@ -78,7 +78,12 @@ def gems(gem_repo = GEM_REPO)
               reqs = d.requirement.requirements.map {|r| r.join(" ") }.join(", ")
               CompactIndex::Dependency.new(d.name, reqs)
             end
-            CompactIndex::GemVersion.new(spec.version.version, spec.platform.to_s, nil, nil,
+            checksum = begin
+                         Digest::SHA256.file("#{GEM_REPO}/gems/#{spec.original_name}.gem").base64digest
+                       rescue
+                         nil
+                       end
+            CompactIndex::GemVersion.new(spec.version.version, spec.platform.to_s, checksum, nil,
```
indirect: can we just `checksum = Digest::SHA256.file("#{GEM_REPO}/gems/#{spec.original_name}.gem").base64digest rescue nil`? lots of lines super far indented makes me feel gross. 😝

segiddins: I don't like the modifier `rescue` and rubocop agrees with me :P

segiddins:
@indirect added a setting to disable validation

segiddins:
This is failing because `index.rubygems.org` is inconsistent between hex & base64 encoding the checksums. Is it worth accounting for that or should we convert to one or the other server-side?

spec/support/artifice/compact_index.rb
indirect: how about `checksum = gem_checksum(spec)` then?

lib/bundler/rubygems_gem_installer.rb
```diff
@@ -12,5 +12,33 @@ def self.at(*args)
...
+        io.rewind
+        digest.base64digest!
+      end
+      unless digest == checksum
+        raise SecurityError,
+          "The checksum for the downloaded `#{spec.full_name}.gem` did not match " \
+          "the checksum given by the API. This means that the contents of the " \
+          "gem appear to be different from what was uploaded, and could be an indicator of a security issue.\n" \
+          "(The expected SHA256 checksum was #{checksum.inspect}, but the checksum for the downloaded gem was #{digest.inspect}.)\n" \
+          "Bundler cannot continue installing #{spec.name} (#{spec.version})."
```
indirect: Based on the principle of empowering users to resolve their own errors, how do you feel about also printing the full path, suggesting deleting the gem with the bad checksum, and explaining that if they are sure they want to install this gem despite the checksum not matching, they can `bundle config disable.checksum_validaiton true`?

lib/bundler/rubygems_gem_installer.rb
```diff
@@ -12,5 +12,33 @@ def self.at(*args)
...
     end
+
+    def pre_install_checks
+      super && validate_bundler_checksum(options[:bundler_expected_checksum])
+    end
+
+  private
+
+    def validate_bundler_checksum(checksum)
+      return true if Bundler.settings[:disable_checksum_validation]
```
indirect: Can we make the setting `disable.checksum_validation`?

indirect:
The config key should probably also to be added to the `bundle config` manpage section about available config keys.

indirect:
Running a data migration right now to replace all checksums in the bundler-api database with the longer hex versions. I'm torn about what we should put in the actual index files, though... using base64 checksums would decrease the file size. :/

lib/bundler/rubygems_gem_installer.rb
segiddins: none of the other bool keys have dots in them, and wouldn't that break doing the setting as an env var?

segiddins:
Rebased

indirect:
This should be good now that the checksums in the database are all consistent.

@homu r+

segiddins:
This should pass now

indirect:
@homu retry

indirect:
oops, @homu r+

Issue #4852, opened 2016-08-09
1.13: the bundling! it's happening!


indirect:
- [x] refactor spec output to remove expect_err
 - [x] allow compact index resolves without a network connection behind a feature flag
 - [x] move gem installation errors to the end of install output
 - [x] refactor Runtime and Environment subclasses away
 - [x] Address comments and merge "set process title when using load" https://github.com/bundler/bundler/pull/4589
 - [x] Add `bundle exec --no-load` for use cases that need `$0` or `__FILE__` https://github.com/bundler/bundler/pull/4869
 - [x] Add docs about how exec defaults to `load` because it's fast and allows disabling that
 - [x] create stable branch to unblock 1.14 work
 - [x] release a new pre
 - [x] wait... 2 days?
 - [x] release final

segiddins:
>  Rebase and merge resolver-style errors for Ruby version conflicts #4654
I can't guarantee this will happen quickly

> Add bundle exec --no-load for use cases that need $0 or __FILE__
I'd rather make it be in `Bundler.settings[:disable_exec_load]`

indirect:
> Rebase and merge resolver-style errors for Ruby version conflicts #4654

Punting is totally fine! I thought it was already done, sorry. 😬  I've moved it to the "any version" list.

> I'd rather make it be in Bundler.settings[:disable_exec_load]

Yeah, that's fine.

indirect:
@segiddins ping on the docs for load-exec? ↑

segiddins:
Done!

PR #4863, opened 2016-08-09
Manually set a process title when using load

segiddins:
Closes #4589.
See https://github.com/bundler/bundler/issues/4852.


indirect:
@homu r+

segiddins:
@homu retry

PR #4865, opened 2016-08-10
Remove "or available on this machine" from GemNotFound error in --path mode 

### LINKED ISSUE #4854: Remove "or available on this machine" from GemNotFound error in --path mode
As discussed in #4571, it's confusing that bundler says "or available on this machine" when the bundle is installed using `--path` mode. We should stop printing it under those circumstances. While we're at it, it would be cool to add "or in gems cached in vendor/cache" if the cache exists.

b-ggs:
Closes #4854.

@RochesterinNYC @indirect @segiddins

### DIFF
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -365,8 +365,9 @@ module Bundler
                        "Source does not contain any versions of '#{requirement}'"
                      end
         else
+          cache_message = Bundler.app_cache.exist? ? " or in gems cached in vendor/cache" : ""
           message = "Could not find gem '#{requirement}' in any of the gem sources " \
-            "listed in your Gemfile or available on this machine."
+            "listed in your Gemfile#{cache_message}."
         end
         raise GemNotFound, message
       end
--- a/spec/commands/exec_spec.rb
+++ b/spec/commands/exec_spec.rb
@@ -540,7 +540,7 @@ describe "bundle exec" do
 
       let(:exit_code) { Bundler::GemNotFound.new.status_code }
       let(:expected) { <<-EOS.strip }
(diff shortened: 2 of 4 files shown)


segiddins:
@homu delegate=RochesterinNYC

RochesterinNYC:
@b-ggs some specs seem to be failing. Looks like `spec/commands/lock_spec.rb:88` was also asserting on "available on this machine"

b-ggs:
I've already tried rebasing from `master`, but still keeps failing on that one spec - which is odd, since I didn't even touch anything related to that spec.

b-ggs:
@RochesterinNYC Could you re-run the specs for this? It might be an issue on Travis' end.

```
Failures:
  1) bundler plugin install shows proper message when gem in not found in the source
     Failure/Error: expect(out).to include("Could not find")
     
       expected "Fetching source index from file:/home/travis/build/bundler/bundler/tmp/gems/remote1/\nCould not locate Gemfile or .bundle/ directory" to include "Could not find"
       Diff:
       @@ -1,2 +1,3 @@
       -Could not find
       +Fetching source index from file:/home/travis/build/bundler/bundler/tmp/gems/remote1/
       +Could not locate Gemfile or .bundle/ directory
```

indirect:
Restarting the tests on Travis...

b-ggs:
For some reason, running `Bundler.app_cache.exist?` breaks something, causing the failure.

b-ggs:
Still looking, but is there any other way to check if cache exists?

segiddins:
@b-ggs it just looks like a different error message is being printed for the one test now, so the expectation can likely just be updated

indirect:
@b-ggs ping?

PR #4867, opened 2016-08-10
Add resiliency against case conflicts on a case-insensitive FS

### LINKED ISSUE #4864: Compact index case collisions
Related to rubygems/rubygems.org#1356, there is a problem compact index clients: if you install the gem `Rack`, Bundler will download the file from `/info/Rack` and save it to the filesystem. If you later install the gem `rack`, Bundler will download the file from `/info/rack` and save it to the filesystem. But most filesystems are not case-sensitive. So this means that the compact index client will be overwriting the cache file `/info/rack` back and forth every time one of these conflicting gems is installed. :/ We can fix this in the client by changing the filename for caching capitalized gem names.

I propose that we add a hash to the end of the cache file for any gem name that contains any characters that can't be represented on the filesystem:

```ruby
info_file = [info_file.gsub(/[^a-z0-9_-]/, ''), md5(info_file)].join("-") if info_file =~ /[^a-z0-9_-]/
```

segiddins:
Closes #4864

### DIFF
--- a/lib/bundler/vendor/compact_index_client/lib/compact_index_client.rb
+++ b/lib/bundler/vendor/compact_index_client/lib/compact_index_client.rb
@@ -60,9 +60,8 @@ class Bundler::CompactIndexClient
 private
 
   def update(local_path, remote_path)
-    return if @endpoints.include?(remote_path)
+    return unless @endpoints.add?(remote_path)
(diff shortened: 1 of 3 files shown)


indirect:
@homu r+

segiddins:
@homu retry

PR #4868, opened 2016-08-10
Rename all BUNDLE_ORIG_* env to BUNDLER_ORIG*

indirect:
Any enviroment variable that starts with BUNDLE_ will be treated as a
configurationg setting, printed by `bundle config`, and made available
internally via `Bundler.settings`. The ORIG_* environment variables are
actually just internal housekeeping to enable us to provide clean
non-bundled environments, and so they shouldn't show up as settings.

This change to the environment variable names makes sure it is still
clear where they are coming from, but no longer surfaces them via
config/settings.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -214,8 +214,8 @@ module Bundler
       Bundler::SharedHelpers.major_deprecation("`Bundler.clean_env` has weird edge cases, use `.original_env` instead")
       env = original_env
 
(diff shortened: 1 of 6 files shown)


indirect:
cc @segiddins

segiddins:
@homu r+

indirect:
homu :sob: @homu retry

PR #4869, opened 2016-08-10
Add a setting for disable_exec_load

segiddins:
See #4852


indirect:
Hmmm I was actually thinking of a flag on `exec`, like `exec --no-load`. This is good too, though. Document this setting in `man/bundle_config` and I think this is good!

segiddins:
I personally prefer settings to flags for things that are rarely needed

indirect:
I was thinking of the flag for the use case where one particular command only works with load disabled, but you don't want to lose the performance improvements generally. But shipping a setting seems fine for now, you can do a one-off using env.

indirect:
@homu r+

segiddins:
@homu r+

PR #4874, opened 2016-08-11
[Travis] Put the tmp directory in a ramdisk

### LINKED ISSUE #2865: Run Travis tests on an in-memory file system
/cc @bleything

segiddins:
Closes #2865 if it works

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 language: ruby
 script: rake spec:travis
 before_script:
+  - rm -rf tmp
+  - mkdir -p /var/ramfs/bundler/tmp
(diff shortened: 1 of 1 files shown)


segiddins:
So this a) breaks specs and b) is no faster

indirect:
It seems... somewhat faster? 12 hrs 22 sec vs other recent times of around 12 hrs 45 min makes it seem like it's around 5% faster. Failing tests is not great though. :( Probably things inside the tests that either don't know how to handle symlinks or permissions that are different on the ramdisk.

This is probably not worth pursuing if our main bottleneck is CPU time, though. :/

PR #4876, opened 2016-08-12
Upgrade thor to master version to test the new release

rafaelfranca:
I'm planing to release a new version of thor so I'm opening this pull request to test if the new release doesn't break bundler and later to update to the released version.

### DIFF
--- a/lib/bundler/vendor/thor/lib/thor/shell/basic.rb
+++ b/lib/bundler/vendor/thor/lib/thor/shell/basic.rb
@@ -34,6 +34,15 @@ class Bundler::Thor
         @padding = [0, value].max
       end
 
+      # Sets the output padding while executing a block and resets it.
+      #
+      def indent(count = 1, &block)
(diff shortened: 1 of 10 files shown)


segiddins:
This seems to have broken the `--no-color` flag being passed to `help`

rafaelfranca:
Thanks, I'll investigate it.

lib/bundler/vendor/thor/lib/thor/shell/basic.rb
```diff
@@ -34,6 +34,15 @@ def padding=(value)
         @padding = [0, value].max
       end
 
+      # Sets the output padding while executing a block and resets it.
+      #
+      def indent(count = 1, &block)
+        orig_padding = padding
+        self.padding = padding + count
+        yield
+        self.padding = orig_padding
```
segiddins: this looks like it should be in an `ensure`, if you want me to send a PR?

PR #4878, opened 2016-08-14
Change standalone flag of binstubs cmd to boolean

Zorbash:
This flag was introduced in #4610, declared with the same code as the flag found in the `install` command.

In the case of the `install` command, it makes sense to be an `Array` option, since
one can specify groups.

This doesn't hold true for the `binstubs` cmd where `--standalone` just stands to signify
whether the generated stubs will require the Bundler runtime or not.

In the current implementation even passing groups with the flag doesn't have any affect.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -255,7 +255,7 @@ module Bundler
       "Overwrite existing binstubs if they exist"
     method_option "path", :type => :string, :lazy_default => "bin", :banner =>
       "Binstub destination directory (default bin)"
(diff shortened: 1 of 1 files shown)


segiddins:
👍🏻 thanks!

segiddins:
@homu r+

segiddins:
@homu retry 😭

segiddins:
@homu retry

indirect:
@homu retry

segiddins:
@homu retry

PR #4882, opened 2016-08-15
Fix open default gem error

### LINKED ISSUE #4436: `bundle open` fails when the gem is a default gem
- What did you do?
  
  I ran the command `/Users/Glenn/.rvm/gems/ruby-2.3.0/bin/bundle open json`
  
  I expected Bundler to open a folder that exists, such as `~/.rvm/rubies/ruby-2.2.2/lib/ruby/gems/2.2.0/gems/json-1.8.3` or `~/.rvm/gems/ruby-2.2.4/gems/json-1.8.3` or wherever it happens to really exist on the system. Maybe it should point to the `~/.rvm/rubies/ruby-2.3.0/lib/ruby/2.3.0/json` as the native 2.3.0 location.
- What happened instead?
  
  Instead, what actually happened was this error template spit out:

Error details

```
Errno::ENOENT: No such file or directory @ dir_chdir - /Users/Glenn/.rvm/rubies/ruby-2.3.0/lib/ruby/gems/2.3.0/gems/json-1.8.3
  /Users/Glenn/.rvm/gems/ruby-2.3.0/gems/bundler-1.11.2/lib/bundler/cli/open.rb:16:in `chdir'
  /Users/Glenn/.rvm/gems/ruby-2.3.0/gems/bundler-1.11.2/lib/bundler/cli/open.rb:16:in `run'
  /Users/Glenn/.rvm/gems/ruby-2.3.0/gems/bundler-1.11.2/lib/bundler/cli.rb:313:in `open'
  /Users/Glenn/.rvm/gems/ruby-2.3.0/gems/bundler-1.11.2/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
  /Users/Glenn/.rvm/gems/ruby-2.3.0/gems/bundler-1.11.2/lib/bundler/vendor/thor/lib/thor/invocation.rb:126:in `invoke_command'
  /Users/Glenn/.rvm/gems/ruby-2.3.0/gems/bundler-1.11.2/lib/bundler/vendor/thor/lib/thor.rb:359:in `dispatch'
  /Users/Glenn/.rvm/gems/ruby-2.3.0/gems/bundler-1.11.2/lib/bundler/vendor/thor/lib/thor/base.rb:440:in `start'
  /Users/Glenn/.rvm/gems/ruby-2.3.0/gems/bundler-1.11.2/lib/bundler/cli.rb:10:in `start'
  /Users/Glenn/.rvm/gems/ruby-2.3.0/gems/bundler-1.11.2/exe/bundle:19:in `block in <top (required)>'
  /Users/Glenn/.rvm/gems/ruby-2.3.0/gems/bundler-1.11.2/lib/bundler/friendly_errors.rb:7:in `with_friendly_errors'
  /Users/Glenn/.rvm/gems/ruby-2.3.0/gems/bundler-1.11.2/exe/bundle:17:in `<top (required)>'
  /Users/Glenn/.rvm/gems/ruby-2.3.0/bin/bundle:23:in `load'
  /Users/Glenn/.rvm/gems/ruby-2.3.0/bin/bundle:23:in `<main>'
```

Environment

```
Bundler   1.11.2
Rubygems  2.5.1
Ruby      2.3.0p0 (2015-12-25 revision 53290) [x86_64-darwin15]
GEM_HOME  /Users/Glenn/.rvm/gems/ruby-2.3.0
GEM_PATH  /Users/Glenn/.rvm/gems/ruby-2.3.0:/Users/Glenn/.rvm/gems/ruby-2.3.0@global
RVM       1.26.11 (latest)
Git       2.6.4
rubygems-bundler (1.4.4)
```

allenzhao:
Close #4436
@indirect @segiddins @RochesterinNYC Please review
Also, is an update in documentation of `bundle open` needed?

### DIFF
--- a/spec/commands/open_spec.rb
+++ b/spec/commands/open_spec.rb
@@ -2,11 +2,13 @@
 require "spec_helper"
 
 describe "bundle open" do
-  before :each do
-    install_gemfile <<-G
-      source "file://#{gem_repo1}"
-      gem "rails"
-    G
+  unless example.metadata[:skip_before]
+    before :each do
+      install_gemfile <<-G
+        source "file://#{gem_repo1}"
+        gem "rails"
+      G
+    end
   end
 
   it "opens the gem with BUNDLER_EDITOR as highest priority" do
@@ -82,5 +84,14 @@ describe "bundle open" do
   it "opens the editor with a clean env" do
     bundle "open", :env => { "EDITOR" => "sh -c 'env'", "VISUAL" => "", "BUNDLER_EDITOR" => "" }
     expect(out).not_to include("BUNDLE_GEMFILE=")
+  end
+
+  it "throws proper error when trying to open default gem", :skip_before => true do
+    gemfile <<-G
+        gem "json"
+      G
+    bundle "open json", :env => { "EDITOR" => "echo editor", "VISUAL" => "echo visual", "BUNDLER_EDITOR" => "echo bundler_editor" }
+    expect(out).to include("Unable to open json")
+    expect(out).to include("because the directory it would normally be installed to does not exist. This could happen when you try to open a default gem.")
   end
 end
--- a/lib/bundler/cli/open.rb
+++ b/lib/bundler/cli/open.rb
@@ -13,12 +13,16 @@ module Bundler
     def run
       editor = [ENV["BUNDLER_EDITOR"], ENV["VISUAL"], ENV["EDITOR"]].find {|e| !e.nil? && !e.empty? }
       return Bundler.ui.info("To open a bundled gem, set $EDITOR or $BUNDLER_EDITOR") unless editor
-      path = Bundler::CLI::Common.select_spec(name, :regex_match).full_gem_path
-      Dir.chdir(path) do
-        command = Shellwords.split(editor) + [path]
-        Bundler.with_clean_env do
-          system(*command)
-        end || Bundler.ui.info("Could not run '#{command.join(" ")}'")
+      spec = Bundler::CLI::Common.select_spec(name, :regex_match)
+      begin
+        Dir.chdir(spec.full_gem_path) do |path|
+          command = Shellwords.split(editor) + [path]
+          Bundler.with_clean_env do
+            system(*command)
+          end || Bundler.ui.info("Could not run '#{command.join(" ")}'")
+        end
+      rescue Errno::ENOENT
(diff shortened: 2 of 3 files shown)


allenzhao:
Failure seems to be related to hard-coded version string...
Any idea of how to get the system's `json` version?

RochesterinNYC:
You could probably do something with `gem list`. Or alternatively just change the assertion to not look for an explicit json version in the expected message.

segiddins:
@homu delegate=RochesterinNYC

segiddins:
@RochesterinNYC r?

RochesterinNYC:
@homu r+

indirect:
@allenzhao could you take a look at the failure?

allenzhao:
@indirect I've updated and it should be fine now. Turns out to_s won't be auto called under 2.0.0

segiddins:
Failing RuboCop -- run `bin/rubocop -a` locally to correct it / see the failure

segiddins:
@allenzhao ping

allenzhao:
I'm confused by the output - maybe I'm writing the wrong test.
When Ruby version higher than 2.0, everything seems fine, but when Ruby version < 2.0, say 1.9.7, the to_s is not auto called on string interpolation. 
When I manually called `to_s`, it seems that above 2.0 rubocop will fail, but under 2.0 'json' is not native built in so the test case itself would fail.
Any ideas on a proper way to test this feature?

segiddins:
What do you mean `to_s is not auto called on string interpolation`?

spec/commands/open_spec.rb
```diff
@@ -2,11 +2,13 @@
...
-      source "file://#{gem_repo1}"
-      gem "rails"
-    G
+  unless example.metadata[:skip_before]
+    before :each do
+      install_gemfile <<-G
+        source "file://#{gem_repo1}"
+        gem "rails"
+      G
+    end
```
olleolleolle: The [test output in Travis](https://travis-ci.org/bundler/bundler/jobs/154594024#L399) seems to install _this_ Gemfile.

The `skip_before: true` metadata is ignored.

How about:

- `before(:each) do |example|` to have access to the `example` inside the `before`
- add a line inside the block: `next if example.metadata[:skip_before]`
- remove the wrapping `unless`/`end`

Issue #4778, opened 2016-07-09
bundler assumes ~/.bundle to be at /.bundle when $HOME is foul


schnittchen:
This happened to me today:

```
vagrant@vagrant:~$ HOME=foo bundle config --global path value
There was an error while trying to write to `/.bundle/config`. It is likely that you need to grant write permissions for that path.
```

I'd say that touching `/.bundle` is probably never a good idea, unless HOME actually is `/`.

schnittchen:
I even had a situation where bundler happily created /.bundle/config without error, because during provisioning of a vagrant machine as root, then using sudo, somehow privileges for writing to `/` were not properly dropped. Took me hours to figure out this was happening.

segiddins:
Hmm, it seems like maybe we could just raise an exception when `Dir.home` isn't writable by the current user?

schnittchen:
Sounds reasonable to me.

allenzhao:
I'll work on this
@segiddins When should be proper to check the `Dir.home`?  Right before writing to it?

segiddins:
@allenzhao honestly, on start up might be the best place? otherwise in all of the places that expand based on `~`, centralized in a `SharedHelpers` method?

Issue #4871, opened 2016-08-11
Tasks anyone can claim


indirect:
- [x] resurrect the unfinished `add` command #4901
- [ ] store a Gemfile checksum in Gemfile.lock, so we can skip the resolver when unchanged
- [x] Add a `bundle issue` command that prints instructions for reporting issues as well as the output from `env` and `doctor` #5505
- [x] add `--parsable` option to `config` #4913 https://github.com/bundler/bundler/issues/4919 #5306
- [x] fix build and merge OpenSSL version for `env` https://github.com/bundler/bundler/pull/4624
- [x] on gem install errors, print the last ten lines of make output if possible
- [x] parse settings keys on "." first so keys are always accurate https://github.com/bundler/bundler/issues/4830
- [x] setting to only install newer options during `update` https://github.com/bundler/bundler/issues/4511, https://github.com/bundler/bundler/pull/4893
- [x] Refactor RSpec matchers into proper classes https://github.com/bundler/bundler/issues/4806
- [x] Use `/tmp/bundler` for compact_index when Dir.home isn't writable https://github.com/bundler/bundler/issues/4792 #4819
- [x] Skip persisting global settings to disk if Dir.home is missing or unwritable #4918 
- [x] Fix running bundler/inline blocks twice https://github.com/bundler/bundler/issues/4837
- [ ] add support for the compact index to [Gemstash](https://github.com/bundler/gemstash)
- [x] Address comments and merge symlink support for standalone bins https://github.com/bundler/bundler/pull/4782
- [x] Retry downloading gems https://github.com/bundler/bundler/issues/4846
- [x] Refactor Definition to avoid `eval`ing Gemfile twice https://github.com/bundler/bundler/pull/4417 #4952
- [ ] Run Travis specs in parallel using `parallel_tests` https://github.com/bundler/bundler/pull/4345
- [x] Write an RFC for and/or implement a fix for  `platform`, which today means multiple conflicting things: gem architecture (eg x86_64), gem platform OS (eg java), engine version (eg mri_22). This may involve adding gemspec support for Ruby engine and version requirements?
- [x] Rebase and merge resolver-style errors for Ruby version conflicts https://github.com/bundler/bundler/pull/4654
- [ ] Add a `bundle plugin list` command to list the current plugins in the bundle. #5467 
- [ ] Add a `bundle plugin update` command to update the version of an installed plugin #5446
- [ ] Fix vendored git gems not being updated on `bundle update` #5423 #5398
[...]

Issue #4819, opened 2016-07-26
Bundler 1.12.5 does not appear to respect `ruby RUBY_VERSION` for dependencies of gems in Gemfile


ferventcoder:
I'm not sure if I'm doing something wrong here, if this is a bug or is by design.

Gemfile:

~~~ruby
ruby RUBY_VERSION
source ENV['GEM_SOURCE'] || 'https://rubygems.org'

gem 'puppet', :require => false
~~~

~~~sh
$ rbenv shell 1.9.3-p392
$ ruby --version
ruby 1.9.3p392 (2013-02-22 revision 39386) [x86_64-darwin12.3.0]
$ bundle --version
Bundler version 1.12.5
$ bundle install
Fetching gem metadata from https://rubygems.org/
Fetching version metadata from https://rubygems.org/
Resolving dependencies...
Using CFPropertyList 2.2.8
Using bundler 1.12.5
Installing json_pure 2.0.2

Gem::InstallError: json_pure requires Ruby version ~> 2.0.
Installing facter 2.4.6
An error occurred while installing json_pure (2.0.2), and Bundler cannot continue.
Make sure that `gem install json_pure -v '2.0.2'` succeeds before bundling.
~~~

ferventcoder:
Discovered as part of https://github.com/bundler/bundler/pull/4651#issuecomment-235382389

ferventcoder:
Is this a bug or by design?

segiddins:
Can you check if this is fixed on master? I've recently fixed several bugs in that area. Thanks!

lamont-granquist:
I'm seeing this fixed on master

ferventcoder:
Thanks @lamont-granquist - I haven't gotten back to test this yet. 👍

lamont-granquist:
Would definitely be good to see this getting released, everyone is deprecating 2.1.x support so we're getting a plague of failing bundle installs, and this would help some.

lynnco:
The relevant changes were released in 1.13, closing

Issue #4792, opened 2016-07-17
Use a tmp dir when `$HOME/.bundle/compact_index` fails


kcee403:
ERROR REPORT TEMPLATE -------------------------------------------------------
- What did you do?

  I ran the command `/usr/local/bin/bundle install`

- What did you expect to happen?

  I expected Bundler to...

- What happened instead?

  Instead, what actually happened was...

Error details

    Errno::EACCES: Permission denied @ dir_s_mkdir - /home/kaycee/.bundle/cache/compact_index/rubygems.org.443.29b0360b937aa4d161703e6160654e47/versions20160717-3214-1st3h5d
      /usr/lib/ruby/2.3.0/tmpdir.rb:86:in `mkdir'
      /usr/lib/ruby/2.3.0/tmpdir.rb:86:in `block in mktmpdir'
      /usr/lib/ruby/2.3.0/tmpdir.rb:130:in `create'
      /usr/lib/ruby/2.3.0/tmpdir.rb:86:in `mktmpdir'
      /var/lib/gems/2.3.0/gems/bundler-1.12.5/lib/bundler/vendor/compact_index_client/lib/compact_index_client/updater.rb:29:in `update'
      /var/lib/gems/2.3.0/gems/bundler-1.12.5/lib/bundler/vendor/compact_index_client/lib/compact_index_client.rb:64:in `update'
      /var/lib/gems/2.3.0/gems/bundler-1.12.5/lib/bundler/vendor/compact_index_client/lib/compact_index_client.rb:55:in `update_and_parse_checksums!'
      /var/lib/gems/2.3.0/gems/bundler-1.12.5/lib/bundler/fetcher/compact_index.rb:65:in `available?'
      /var/lib/gems/2.3.0/gems/bundler-1.12.5/lib/bundler/fetcher/compact_index.rb:15:in `call'
      /var/lib/gems/2.3.0/gems/bundler-1.12.5/lib/bundler/fetcher/compact_index.rb:15:in `block in compact_index_request'
      /var/lib/gems/2.3.0/gems/bundler-1.12.5/lib/bundler/fetcher.rb:154:in `use_api'
      /var/lib/gems/2.3.0/gems/bundler-1.12.5/lib/bundler/source/rubygems.rb:331:in `block in api_fetchers'
      /var/lib/gems/2.3.0/gems/bundler-1.12.5/lib/bundler/source/rubygems.rb:331:in `select'
[...]

PR #4886, opened 2016-08-16
Add check of Dir.home's permission

allenzhao:
Ref #4778

### DIFF
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 require "pathname"
 require "rubygems"
+require "etc"
 
 require "bundler/constants"
 require "bundler/rubygems_integration"
@@ -137,7 +138,11 @@ module Bundler
     end
 
     def check_home_dir_permissions
-      raise PathError, "There was an error while trying to use your home path #{Dir.home}" unless File.writable?(Dir.home)
+      message = "There was error while trying to use your home path:"
+      home_path = File.expand_path(Dir.home(Etc.getlogin))
+      message += "\n * Your home directory #{home_path} is not writable" unless File.writable?(home_path)
+      message += "\n * Your home directory #{home_path} is not a directory" unless File.directory?(home_path)
+      raise PathError, message unless File.writable?(home_path) && File.directory?(home_path)
     end
 
   private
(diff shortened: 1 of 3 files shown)


lib/bundler/shared_helpers.rb
```diff
@@ -136,6 +136,10 @@ def print_major_deprecations!
       major_deprecation("Bundler will only support rubygems >= 2.0, you are running #{Bundler.rubygems.version}")
     end
 
+    def check_home_dir_permissions
+      raise PathError, "There was an error while trying to use your home path #{Dir.home}" unless File.writable?(Dir.home)
```
segiddins: this should specifically mention that the home directory must exist & be writable

spec/bundler/shared_helpers_spec.rb
```diff
@@ -412,4 +412,11 @@ module TargetNamespace
       end
     end
   end
+
+  describe "#check_home_dir_permissions" do
+    it "raises a PathError" do
+      allow(File).to receive(:writable?).with(Dir.home).and_return(false)
```
segiddins: please also add a test for when `Dir.home` does not exist

segiddins:
@allenzhao ping?

indirect:
I think Bundler should just work, even if home directories are unwritable or do not exist. User-level settings should not be persisted, and cached files like the compact index and gem download cache should go into `/tmp/bundler` instead.

spec/bundler/shared_helpers_spec.rb
JuanitoFatas: Regarding to `Dir.home`, I found using user's login is more reliable:

```
# suppose ENV HOME is gone
> ENV.delete 'HOME'

> Dir.home
ArgumentError: couldn't find HOME environment -- expanding `~'

> require 'etc'; File.expand_path(Dir.home(Etc.getlogin))
> "/Users/Juan"
```

allenzhao:
@indirect So what should be the solution here?
@segiddins I've already added the test and updated according to @JuanitoFatas 's advice.
Thanks!

segiddins:
@allenzhao there are RuboCop failures in addition

indirect:
@allenzhao I added notes to the list ticket #4871:

> Use /tmp/bundler for compact_index when Dir.home isn't writable #4792 #4819
> Skip persisting global settings to disk if Dir.home is missing or unwritable #4918

I think it means we will want to fix this problem in a different way, with PRs that solve those two problems. I would be very happy to have you work on those, and if you would like me to write a more detailed description of how that should work, I am happy to do it, just let me know.

allenzhao:
@indirect Please do by opening a new issue, and I will close this PR.

PR #4891, opened 2016-08-18
Add FeatureFlag class

segiddins:
Main use will be code showing what is true setting vs what is flagged

Also can be used to develop for future major versions without needing a separate dev branch


indirect:
@homu r+

Issue #4511, opened 2016-05-03
--newer-only option for install/update


lynnco:
From: https://github.com/bundler/bundler-features/issues/73

Involves:

- [ ] setting a flag in 1.X
- [ ] changing the default `update` behavior in 2.0

segiddins:
```diff
diff --git a/lib/bundler/definition.rb b/lib/bundler/definition.rb
index 36bb9f2..f2f8f49 100644
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -249,7 +249,7 @@ module Bundler
         else
           # Run a resolve against the locally available gems
           Bundler.ui.debug("Found changes from the lockfile, re-resolving dependencies because #{change_reason}")
-          last_resolve.merge Resolver.resolve(expanded_dependencies, index, source_requirements, last_resolve, ruby_version, gem_version_promoter)
+          last_resolve.merge Resolver.resolve(expanded_dependencies, index, source_requirements, resolver_base_set(last_resolve), ruby_version, gem_version_promoter)
         end
       end
     end
@@ -633,6 +633,18 @@ module Bundler
       Set.new(@dependencies) != Set.new(@locked_deps)
     end
 
+    LockedSpec = Struct.new(:name, :version)
+
+    def resolver_base_set(converged_locked_specs)
+      base = converged_locked_specs.to_hash
+      @locked_specs.each do |spec|
+        if base[spec.name].empty?
+          base[spec.name] << LockedSpec.new(spec.name, ">= #{spec.version}")
+        end
+      end
+      base.values.flatten
+    end
+
     # Remove elements from the locked specs that are expired. This will most
     # commonly happen if the Gemfile has changed since the lockfile was last
     # generated
@@ -707,7 +719,6 @@ module Bundler
           source.unlock! if !diff.empty? && diff.any? {|s| s.source == source }
         end
       end
-
       resolve
     end
 
```

I believe this should be the gist of it, but it's untested

Issue #4856, opened 2016-08-09
2.0: a whole new world


indirect:
- [x] remove psyched_yaml https://github.com/bundler/bundler/issues/4825 https://github.com/bundler/bundler/pull/4916
- [x] remove .ruby-version support #4892 
- [x] remove `inject` command (deprecated in favor of `add`)
- [x] always resolve with auto-lock to RUBY_VERSION (as per #4853)
- [x] in --deployment mode, raise when running ruby doesn't match auto-locked ruby
- [x] documentation for managing ruby versions with bundler
- [x] remove --binstubs from the docs (code already removed)
- [x] default to only installing newer versions during `update` https://github.com/bundler/bundler/issues/4511 https://github.com/bundler/bundler/pull/4893
- [x] default to `cache --all-platforms` and remove flag #4918 
- [x] merge in Tim's spec for source issues https://github.com/bundler/bundler/pull/3696
- [x] merge in 1.13
- [x] merge in 1.14
- [x] take over master
- [x] ~enable trampolining to locked versions "~> 1"~

segiddins:
the pre-req for all of this work is for master to be merged back to development (again)

PR #4893, opened 2016-08-18
Block resolving to older versions during an update

segiddins:
This is currently behind the only_update_to_newer_versions setting

Closes https://github.com/bundler/bundler/issues/4511.

See https://github.com/bundler/bundler/issues/4856 and https://github.com/bundler/bundler/issues/4871


indirect:
@homu r+

Issue #4853, opened 2016-08-09
1.14: the one with the platforms, finally


indirect:
- [x] create stable branch after 1.13 final to unblock 2.0 merge
- [x] merge new platform support https://github.com/bundler/bundler/pull/4836
  - [x] bundler treats distinct platforms as identical (see also https://github.com/bundler/bundler/issues/4295, https://rubygems.org/gems/google-protobuf/versions)
  - [x] bundler seems to complain about broken lockfiles when they have more than one platform (???) https://github.com/bundler/bundler/issues/4620#issuecomment-225650438
  - [x] users on windows need to be able to deploy to heroku, which is currently impossible bc heroku needs ruby/source gems but windows devs lock to (one of) the windows binaries. maybe this is fixed by `lock --add-platform ruby`?
  - [x] Add an option to force Bundler to only use `Gem::Platform::RUBY` https://github.com/bundler/bundler/pull/4895 https://github.com/bundler/bundler/issues/4813
- [x] Create a (predictably-named) tmpdir if it's impossible to write the compact index to $HOME
- [x] final trampoline messaging PR
- [x] default enable trampolining to locked versions "~> 2" #4894 
- [x] env/setting to enable trampolining to any version (`BUNDLE_ENABLE_TRAMPOLINE`)
- [x] announcement/docs for platform support & commands
- [ ] announcement/docs for bundler locking option
- [ ] website guide on how to use bundler across platforms
- [ ] website guide on how to use bundler locking

segiddins:
> option to resolve respecting running ruby versions
I thought we agreed to move this to 2.0?

indirect:
> > option to resolve respecting running ruby versions
> I thought we agreed to move this to 2.0?

@segiddins I was under the impression that we agreed to delay raising exceptions until 2.0, but it's fully backwards-compatible to auto-lock and then print a warning when there's no `ruby` in the Gemfile and the running ruby doesn't match the locked ruby. Raising when auto-ruby doesn

segiddins:
Sure, but we'll need test coverage to fully flesh this out

indirect:
@segiddins btw, are you planning to write the docs in this checklist?

segiddins:
I had no immediate plans to, I figured I'd let someone else participate in the release :P

lynnco:
- [ ] add documentation for `force_ruby_platform` https://github.com/bundler/bundler/pull/5028

segiddins:
1.14 was released

PR #4894, opened 2016-08-18
[PostitTrampoline] Automatically enable when target version is >= 2

segiddins:
See https://github.com/bundler/bundler/issues/4853


indirect:
@homu r+

PR #4897, opened 2016-08-19
[Resolver] Ensure all platforms are activated for the activated spec group

### LINKED ISSUE #4896: `bundle lock` does not activate gems for current platform if it was activated under a different platform for a separate dependency
As part of our build process we `bundle lock` against `ruby` and then again against `x86-mingw32` and commit a unified lock file with aggregate gem requirements. This works well for us in most cases. However there is a specific edge case where this breaks.

Assume this minimal Gemfile:
```
source "https://rubygems.org"

gem "mixlib-shellout"
gem "gssapi"
```
Now if I `bundle lock` separately  on linux and windows, each platform generates an acceptable lockfile. However, if I have a windows generated lock file that looks like this:
```
GEM
  remote: https://rubygems.org/
  specs:
    ffi (1.9.14-x86-mingw32)
    gssapi (1.2.0)
      ffi (>= 1.0.1)
    mixlib-shellout (2.2.6-universal-mingw32)
      win32-process (~> 0.8.2)
      wmi-lite (~> 1.0)
    win32-process (0.8.3)
      ffi (>= 1.0.0)
    wmi-lite (1.0.0)

PLATFORMS
  x86-mingw32

DEPENDENCIES
  gssapi
  mixlib-shellout

BUNDLED WITH
   1.12.5
```
Then I run `bundle lock` on linux, I get:
```
GEM
  remote: https://rubygems.org/
  specs:
    ffi (1.9.14-x86-mingw32)
    gssapi (1.2.0)
      ffi (>= 1.0.1)
    mixlib-shellout (2.2.6)
    mixlib-shellout (2.2.6-universal-mingw32)
      win32-process (~> 0.8.2)
      wmi-lite (~> 1.0)
    win32-process (0.8.3)
      ffi (>= 1.0.0)
    wmi-lite (1.0.0)

PLATFORMS
  ruby
  x86-mingw32

DEPENDENCIES
  gssapi
  mixlib-shellout

BUNDLED WITH
   1.12.5
```
The key error here is the ommision of `ffi (1.9.14)`. I'm left with only the windows variant but `gssapi` should be activating one for the `ruby` platform.

First, I realize there is work going into 1.13 to improve platform related issues. I have tried this scenario with 1.13.0.RC.1 and this problem is still present. Second I have looked through bundler source and "think" I understand the issue and have a fix that remedies the problem. However as I'll discuss shortly, its probably not the right fix.

So this is definitely an edge case and with different requirements I get expected results. For example, if my Gemfile only had `gssapi` and did not include `mixlib-shellout`, I'd end up with both `ffi` platforms. Here is why the a Gemfile with `mixlib-shellout` (and I'm sure there are others) produces unexpected results. `mixlib-shellout` has two separate gemspecs - a [second windows specific one](https://github.com/chef/mixlib-shellout/blob/master/mixlib-shellout-windows.gemspec). Only its windows gemspec takes a dependency on `win32-process` which in turn takes a dependency on `ffi`. So the resolver resolves `mixlib-shellout` and rightly so activates no `ffi` on linux but does activate it for `windows` given a lock file with a mingw dependency. Later when it resolves `gssapi` on linux, the resolver produces a `SpecGroup` for `ffi` with an activated `ruby` platform. However, when the `Resolution` attempts to activate the `possibility`:
```
      # Attempts to activate the current {#possibility}
      # @return [void]
      def attempt_to_activate
        debug(depth) { 'Attempting to activate ' + possibility.to_s }
        existing_node = activated.vertex_named(name)
          if existing_node.payload
          debug(depth) { "Found existing spec (#{existing_node.payload})" }
          attempt_to_activate_existing_spec(existing_node)
        else
          attempt_to_activate_new_spec
        end
      end
```
Here `existing_node` points to the previously activated windows dependency and the ruby `possibility` is thrown away.

Changing:
```
if existing_node.payload
```
to:
```
if existing_node.payload && (possibility.activated - existing_node.payload.activated).empty?
```
fixes this and ensures if the current possibility contains active platforms not included in the existing possibiity, it will be activated. This produces the ` ffi (>= 1.0.1)` requirement previously missing and prevents `corrupted Gemfile.lock` errors on linux.

Now I think this is possibly not the "right" change since it is in the vendored `Molinillo` library. I'm honestly not familiar enough with these code bases to know if I should submit a PR and tests to that repo or if something should be changed in the way bundler consumes it. Like maybe keying off of something other than `name` (if thats even controlled on the bundler side).

Would really appretiate thoughts on this. Thanks!!

segiddins:
Closes #4896

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -816,7 +816,8 @@ module Bundler
     def additional_base_requirements_for_resolve
       return [] unless @locked_gems && Bundler.settings[:only_update_to_newer_versions]
       @locked_gems.specs.reduce({}) do |requirements, locked_spec|
-        requirements[locked_spec.name] = Gem::Dependency.new(locked_spec.name, ">= #{locked_spec.version}")
+        dep = Gem::Dependency.new(locked_spec.name, ">= #{locked_spec.version}")
+        requirements[locked_spec.name] = DepProxy.new(dep, locked_spec.platform)
         requirements
       end.values
     end
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -189,8 +189,11 @@ module Bundler
       @resolver = Molinillo::Resolver.new(self, self)
       @search_for = {}
       @base_dg = Molinillo::DependencyGraph.new
-      @base.each {|ls| @base_dg.add_vertex(ls.name, Dependency.new(ls.name, ls.version), true) }
-      additional_base_requirements.each {|d| @base_dg.add_vertex(d.name, d) }
+      @base.each do |ls|
+        dep = Dependency.new(ls.name, ls.version)
+        @base_dg.add_vertex(ls.name, DepProxy.new(dep, ls.platform), true)
+      end
+      additional_base_requirements.each {|d| @base_dg.add_vertex(d.name, DepProxy.new(d, Gem::Platform::RUBY)) }
       @ruby_version = ruby_version
       @gem_version_promoter = gem_version_promoter
     end
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -193,7 +193,7 @@ module Bundler
         dep = Dependency.new(ls.name, ls.version)
         @base_dg.add_vertex(ls.name, DepProxy.new(dep, ls.platform), true)
       end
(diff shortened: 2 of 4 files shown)


indirect:
r+ once this is green.

indirect:
@homu r+

segiddins:
This is actually green

indirect:
@homu retry

indirect:
@homu r+

PR #4903, opened 2016-08-21
Allow 1.8.7 to fail, test RubyGems 2.6.6

indirect:


indirect:
@segiddins I propose that there are not enough hours in the universe to retry desired mergers until they will have no segfaults. If we manage break 1.8.7 compatibility, we will fix it and release a bugfix version ASAP, but we need to be able to ship 1.14 ASAP first.

Rakefile
```diff
@@ -127,7 +127,7 @@ begin
       rubyopt = ENV["RUBYOPT"]
       # When editing this list, also edit .travis.yml!
       branches = %w(master)
-      releases = %w(v1.3.6 v1.3.7 v1.4.2 v1.5.3 v1.6.2 v1.7.2 v1.8.29 v2.0.14 v2.1.11 v2.2.5 v2.4.8 v2.6.4)
+      releases = %w(v1.3.6 v1.3.7 v1.4.2 v1.5.3 v1.6.2 v1.7.2 v1.8.29 v2.0.14 v2.1.11 v2.2.5 v2.4.8 v2.6.6)
```
segiddins: this needs to include 2.5.2 now

segiddins:
😭 I guess we have no choice unless travis fixes this on their end. My only stipulation is that we manually check those allowed 1.8.7 failures (on the latest master/stable CI run) before release

segiddins:
@homu r+

PR #4905, opened 2016-08-22
[WIP] Ask for permission when using passwordless sudo

### LINKED ISSUE #3006: ask for permission before using password-less sudo
I've been using bundler for years already in various scenarios (and BTW it's **awesome** ;-) but somehow this is the first time I noticed that if I am using the system-wide `/usr/bin/ruby` (e.g. `rvm use system`) and I have `sudoers` configured to allow running arbitrary commands as `root`, if I do `bundle install` as non-root, it will happily take advantage of that and start running a load of `sudo` commands to install gems system-wide into `/usr/lib64`.  It's of course possible I've misunderstood something, but this was a big WTF moment for me, since now my nice clean rpm-managed root filesystem was messed up with no warning at all.

I realise that by default, bundler tries installs gems to the same location as `gem install`, and that in the above case, that means system-wide.  I realise that there are totally valid scenarios where that is the required behaviour (e.g. since [supporting half-and-half system+user setups is a nightmare](https://github.com/bundler/bundler/issues/946#issuecomment-12483423)), and I'm totally OK with that.  I'm also OK with the general idea of bundler achieving that via `sudo`.  *However*, in the `NOPASSWD sudoers` case, this happens without any warning, and IMHO that's definitely not cool.  It's a bit like leaving your front door unlocked - yes, you are probably asking for trouble, but you still really hope that people will have the courtesy to knock before they come in ;-)

I have read #710, #890, #946 etc. and I'm aware that setting `GEM_HOME` to a writable location will prevent this.  But that's a bit like someone telling you "remember to lock the door when you leave!"  Yes most of the time you'll remember, but occasionally you'll screw up (e.g. a bug in your shell start-up scripts, or you unset it for some reason and forgot to reset it later) - and on those occasions bundler should still behave like the courteous guest who remembers to knock.

So I would suggest:

*   Before using `sudo`, bundler should do `sudo -l` to check whether it is allowed to do stuff without first prompting for the user password.
*   If `sudo -l` shows that a password is required, carry on with the existing behaviour, since `sudo` will naturally ensure that bundler is a polite guest.
*   Otherwise, *by default* prompt the user once per bundler run, asking for permission to use `sudo`.  I suggest the following prompt text:

```
bundler needs to use sudo in order to install gems to $GEM_HOME,
and has detected that this user is allowed to use `sudo` without
any password prompts (`NOPASSWD` in /etc/sudoers).

Are you happy for bundler to use sudo?

    n - no, quit (default)
    y - yes, allow bundler to use sudo just for this run
    A - always, configure bundler to never again ask permission,
        via: bundle config sudo.passwordless always
    N - never, configure bundler to never again attempt to use
        password-less sudo, via: 
        bundle config sudo.passwordless always

Enter your choice [nyAN] ?
```

So the `sudo.passwordless` option would default to `ask`, and in this case, when STDIN is not a tty, it would default to aborting without using `sudo` (since it would have run out of ways to get permission from the user).

Hope that makes sense.  Thanks a lot for listening!

b-ggs:
- [x] Add initial implementation
- [ ] Address additional issues

---
Closes #3006 
@RochesterinNYC

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -318,6 +318,7 @@ module Bundler
     end
 
     def sudo(str)
+      return unless settings.allow_sudo?
       SUDO_MUTEX.synchronize do
         prompt = "\n\n" + <<-PROMPT.gsub(/^ {6}/, "").strip + " "
         Your user account isn't allowed to install to the system Rubygems.
--- a/lib/bundler/settings.rb
+++ b/lib/bundler/settings.rb
@@ -178,7 +178,32 @@ module Bundler
     end
 
     def allow_sudo?
-      !@local_config.key?(key_for(:path))
+      is_nopasswd = `sudo -l` =~ /NOPASSWD/
+      allow_user = false || Bundler.settings[:allow_passwordless_sudo]
+      if is_nopasswd && !allow_user
+        resp = Bundler.ui.ask <<-PROMPT
+bundler needs to use sudo in order to install gems to $GEM_HOME,
+and has detected that this user is allowed to use `sudo` without
+any password prompts (`NOPASSWD` in /etc/sudoers).
+
+Do you want bundler to use sudo?
+
+    n - no, quit (default)
+    y - yes, allow bundler to use sudo just for this run
+    A - always, configure bundler to never again ask permission
+
+Enter your choice [nyA]:
+        PROMPT
+        if resp == "y"
+          allow_user = true 
+        elsif resp == "A"
+          allow_user = true 
+          Bundler.settings[:allow_passwordless_sudo] = true
+        else
+          raise BundlerError
+        end
+      end
+      !@local_config.key?(key_for(:path)) && allow_user
     end
 
     def ignore_config?


b-ggs:
I think it would be better to implement (and easier to test!) this even for users whose `sudo`s require passwords. Thoughts? @rochesterinnyc

RochesterinNYC:
Isn't the issue regarding circumstances where no password is required for `sudo` and hence the user has no warning at all that Bundler is trying to use sudo privileges?

segiddins:
Yes.

b-ggs:
Added an initial implementation for this. Still a WIP, but how is this looking so far?

By the way, not too sure how this would go down with Travis, as maybe its sudo has `NOPASSWD` set.

indirect:
The implementation looks okay in general, but I think this is handling the wrong problem—our goal is to let the user know we are going to use sudo (for example, if they have recently run a sudo command, and so sudo will not ask for their password again). In that case, we want to prompt with something like "Bundler needs to use `sudo` to install gems because the current user does not have permission."

I like the way this implementation offers y/n/a options. 👍

segiddins:
@homu r=RochesterinNYC

segiddins:
@homu r-

segiddins:
@homu delegate=RochesterinNYC
Clearly I need to get my coffee now

segiddins:
Closing in favor of https://github.com/bundler/bundler/pull/4956, which was merged

PR #4907, opened 2016-08-22
Add --only argument to bundle install

will-in-wi:
Implementation of an `--only` argument for `bundle install`.

https://github.com/bundler/bundler-features/issues/59

The reasoning here is that sometimes you want a single group or two, and not default. My particular use case is having a deployment group which contains everything needed for a CI server. To do this with existing functionality, I have to list all possible groups in `--without`, which can change and easily be forgotten.

Thanks for considering!

### DIFF
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -151,6 +151,7 @@ module Bundler
     def normalize_groups
       options[:with] &&= options[:with].join(":").tr(" ", ":").split(":")
       options[:without] &&= options[:without].join(":").tr(" ", ":").split(":")
+      options[:only] &&= options[:only].join(":").tr(" ", ":").split(":")
 
       check_for_group_conflicts_in_cli_options
 
@@ -165,8 +166,12 @@ module Bundler
       without |= Bundler.settings[:without].map(&:to_s)
       without -= options[:with] if options[:with]
 
+      only = options.fetch("only", [])
+      only |= Bundler.settings.only.map(&:to_s)
+
       options[:with]    = with
       options[:without] = without
+      options[:only] = only
     end
 
     def normalize_settings
@@ -189,12 +194,13 @@ module Bundler
 
       Bundler.settings.set_command_option_if_given :clean, options["clean"]
 
-      unless Bundler.settings[:without] == options[:without] && Bundler.settings[:with] == options[:with]
+      unless Bundler.settings[:without] == options[:without] && Bundler.settings[:with] == options[:with] && Bundler.settings[:only] == options[:only]
         # need to nil them out first to get around validation for backwards compatibility
         Bundler.settings.set_command_option :without, nil
         Bundler.settings.set_command_option :with,    nil
         Bundler.settings.set_command_option :without, options[:without] - options[:with]
         Bundler.settings.set_command_option :with,    options[:with]
+        Bundler.settings.set_command_option :only,    options[:only]
       end
 
       options[:force] = options[:redownload]
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -197,6 +197,8 @@ module Bundler
       "Exclude gems that are part of the specified named group."
     deprecated_option "with", :type => :array, :banner =>
       "Include gems that are part of the specified named group."
+    method_option "only", :type => :array, :banner =>
+      "Include only the gems that are part of the specified named group."
     map "i" => "install"
     def install
       require "bundler/cli/install"
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -920,6 +920,7 @@ module Bundler
     end
 
     def requested_groups
+      return Bundler.settings[:only] if Bundler.settings[:only].any?
       groups - Bundler.settings[:without] - @optional_groups + Bundler.settings[:with]
     end
 
--- a/lib/bundler/settings.rb
+++ b/lib/bundler/settings.rb
@@ -61,6 +61,7 @@ module Bundler
     ARRAY_KEYS = %w[
       with
       without
+      only
     ].freeze
 
     DEFAULT_CONFIG = {
--- a/man/bundle-install.ronn
+++ b/man/bundle-install.ronn
@@ -23,6 +23,7 @@ bundle-install(1) -- Install the dependencies specified in your Gemfile
                  [--trust-policy=POLICY]
                  [--with=GROUP[ GROUP...]]
                  [--without=GROUP[ GROUP...]]
+                 [--only=GROUP[ GROUP...]]
 
 ## DESCRIPTION
 
@@ -150,6 +151,11 @@ time `bundle install` is run, use `bundle config` (see bundle-config(1)).
   A space-separated list of groups referencing gems to skip during installation.
   If a group is given that is in the remembered list of groups given
   to --with, it is removed from that list.
+
+* `--only=<list>`:
+  A space-separated list of groups referencing gems to install. If groups are
+  given, they are the only groups installed. This overrides --with and
+  --without.
 
 ## DEPLOYMENT MODE
 
--- a/spec/install/gemfile/groups_spec.rb
+++ b/spec/install/gemfile/groups_spec.rb
@@ -70,6 +70,58 @@ RSpec.describe "bundle install with groups" do
     end
   end
 
+  describe "installing --only" do
+    describe "with gems assigned to a single group" do
+      before :each do
+        gemfile <<-G
+          source "file://#{gem_repo1}"
+          gem "thin"
+          group :emo do
+            gem "activesupport", "2.3.5"
+          end
+          group :debugging, :optional => true do
+            gem "rack"
+          end
+        G
+      end
+
+      it "installs gems from only the given optional group when that group is in only" do
+        bundle :install, :only => "debugging"
+        expect(the_bundle).not_to include_gems "thin"
+        expect(the_bundle).not_to include_gems "activesupport"
+        expect(the_bundle).to include_gems "rack 1.0.0"
+      end
+
+      it "installs gems from only the given groups when those group are in only" do
+        bundle :install, :only => "debugging emo"
+        expect(the_bundle).to include_gems "rack 1.0.0"
+        expect(the_bundle).to include_gems "activesupport 2.3.5"
+        expect(the_bundle).not_to include_gems "thin"
+      end
+
+      it "installs gems from only the given group when that group is in only" do
+        bundle :install, :only => "emo"
+        expect(the_bundle).not_to include_gems "rack"
+        expect(the_bundle).to include_gems "activesupport 2.3.5"
+        expect(the_bundle).not_to include_gems "thin"
+      end
+
+      it "installs gems from only the given optional group overriding with" do
+        bundle :install, :only => "debugging", :with => "emo"
+        expect(the_bundle).to include_gems "rack 1.0.0"
+        expect(the_bundle).not_to include_gems "activesupport"
+        expect(the_bundle).not_to include_gems "thin"
+      end
+
+      it "installs gems from only the given optional group ignoring without" do
+        bundle :install, :only => "debugging", :without => "debugging"
+        expect(the_bundle).to include_gems "rack 1.0.0"
+        expect(the_bundle).not_to include_gems "activesupport"
+        expect(the_bundle).not_to include_gems "thin"
+      end
+    end
+  end
+
   describe "installing --without" do
     describe "with gems assigned to a single group" do
       before :each do


lib/bundler/cli/install.rb
```diff
@@ -13,7 +13,7 @@ def run
 
       warn_if_outdated
 
-      [:with, :without].each do |option|
+      [:with, :without, :group].each do |option|
```
MOZGIII: Why is `:group` here?

will-in-wi: Heh, yup. Good catch. Fixed

segiddins:
How does this new setting interact with the existing `without` and `with` settings?

will-in-wi:
I have test cases demonstrating that. Both are ignored if `--only` is passed. It is a more specific setting than either, so neither really makes sense with it.

MOZGIII:
It overrides them, as I can see from the code.
Logically, the purpose is to install stuff from just that group(s) without writing everything else in `--without`.

MOZGIII:
I like this change a lot. Did code review, RTM I guess.

indirect:
Thanks for submitting a PR! I think it's important that using `only` when `with` or `without` are also set should print a warning informing the user that their settings will be ignored.

I'm also somewhat worried that this feature adds complexity for everyone while providing benefit to very few users. The specific use-case of "install only one group" can be solved many ways without changing Bundler. Can you help us understand why you feel this patch is useful to all Bundler users, rather than just your personal use-case?

MOZGIII:
I don't know a way to install only one group without explicitly enumerating every other group that you want to exclude. If there is one, please tell me how to do it.
When is it needed? Mainly for automation. The particular cases are when you want to install only a limited set of gems (as said - without explicitly stating which groups you don't need), and, either when you want them installed "for deployment" for some reason, or when you want to manage all gem dependencies in a `Gemfile` (for convenience maybe?), or both at the same time.

It is useful for installing linters, deploy utils and things like that in a CI environment.
In general, it's usually possible to install required gems separately via `gem install ...`, but then you have to manage them separately and can't easily install them in `--deployment` manner how bundler does.
Managing these "standalone" gems with bundler, of course, is a big plus, because you fix their versions, and do it alongwith the other gems.

will-in-wi:
Before working on this, I looked at several alternatives:

The simplest approach is to use `--without`. This involves repeating a complete list of all groups in the gemfile inside of a CI script. If a group is ever added to the gemfile, the committer will have to remember to add it as well in the CI script. This is error prone and an example of why DRY is a good thing.

Another approach would be to create a secondary gemfile and require it inside the first. This is unexpected for most developers, albeit somewhat self-descriptive. I personally consider it ugly.

Other options such as parsing the gemfile for groups and building the command in the first option, or using env variables and metaprogramming in the Gemfile present themselves, but it is somewhat self-evident as to why those aren't great options.

~15 people 👍-ed the idea as presented in https://github.com/bundler/bundler-features/issues/59.

There is no reasonable way of approaching this problem, which I have to imagine would be fairly common in complex enterprise environments.

indirect:
@will-in-wi sorry, I guess I wasn't clear when I said that there are other ways to accomplish this goal. The most straightforward way to accomplish this goal is [the `bundle-only` gem](https://github.com/MOZGIII/bundle-only). It seems like with that gem installed, you can run `bundle only ci`, and get exactly the outcome you're describing.

I'm looking for an explanation of why you think this need is widespread enough to merit an option in the core of Bundler, rather than an addon that anyone who wants it can use. Thanks!

will-in-wi:
I'll start my case by pointing out that the author of the bundle-only Gem, @MOZGIII, is lobbying for its inclusion. Would it strengthen the case if he pulled his Gem? 😉

I considered the Gem. I'm maintaining CI for ~40 projects which need this feature. I would not use this Gem because it doesn't have a maintenance record. I have no idea whether in a year or two a new version of Bundler would break any of the APIs and @MOZGIII might have moved on to Elixer or something. I don't use new Gems for core elements of my CI environment, since the pain of replacing them is high. If this doesn't go through, I'll probably eat the pain of maintaining two lists of possible groups rather than have another dependency to worry about everywhere.

Additionally, in an environment with split development/infrastructure (common in enterprise), Ruby and Bundler are frequently installed as root which the devs don't have access to. Right now, Bundler can usually be upgraded with minimal testing since the only API used is the CLI. If I were to pull in bundle-only, I would now have to test each version upgrade with that, too.

Finally, in the aforementioned split development/infrastructure environment, I'd have to get IT to take on the job of installing bundler-only and upgrading it everywhere, since I can't use Bundler to install it. Bundler is considered by us (and I presume by many others) to be essentially part of Ruby and installed and maintained as such. Adding another Gem to this list requires a long history of maintenance.

The added complexity to Bundler is minimal. Other than some settings boilerplate, docs, and tests, the real meat of the thing is a one-liner (lib/bundler/definition.rb:783).

I've pinged the ~15 other people on the originating thread (https://github.com/bundler/bundler-features/issues/59) and asked them to weigh in here if they still need the feature.

will-in-wi:
@MOZGIII: I hope my previous comment doesn't offend you. I'm not passing judgement on your work at all, this is how I would consider any Gem that needs to be this incredibly core to our systems.

Thanks for the review!

marioizquierdo:
This is very useful! Some of my projects have a few scripts/rake tasks that only need one group of the gems. The `--only` option would make it easier for other team mates that only need to install the repo in order to run those scripts. The same goes for having a deploy that only uses that subset (i.e. a host that only needs to run capistrano).

rymohr:
@will-in-wi thanks for putting this together. I would find this useful too.

Sure the `bundle-only` gem works but here's a few arguments for why this should be pulled into core:

1. Multiple people have expressed a need for this feature
2. Installing gems to augment bundler's behavior isn't a common practice. Most people view bundler as a standalone tool that just works.
3. The code to implement the feature is only a few lines
4. The change is backwards compatible
5. The `--only` logic is much simpler to reason about (see `requested_groups` below)

```rb
def requested_groups
  return Bundler.settings.only if Bundler.settings.only.any?
  groups - Bundler.settings.without - @optional_groups + Bundler.settings.with
end
```

MOZGIII:
@will-in-wi, that's a legit concern there, and I totally understand it. Really, there are no guarantees that my gem won't suddenly stop working, and I'm worried about that too (but if it was to stop working, someone would quickly point that out to me and we'd fix it, at least that was the plan). Actually, I was worried even before I made it, at the design phase. It is true: the `bundle-only` gem is a dirty/hacky solution to the problem that bundler had at the time I made it. I knew it would take a lot of time to get a change like this into the `bundler` itself, but I needed something quick.
Now it's time to debate and finally add this `--only` possibility (or not, if there is a real reason not to).

indirect:
Believe me, I understand the "but it's not that big of a change" argument. The reason I'm hesitant is that every additional option brings with it the possibility of breaking any other option or combination of options, and every one of those combinations has to be tested and maintained for years.

Asking us to accept this feature is also asking us to do work on your behalf for the entire future life of Bundler. If this option is only useful to a very small number of users (and yes, 15 or even 150 users is a very small percentage of users), then this feature is probably best handled by a plugin.

We are actively working to improve support for plugins, including declaring plugins in the Gemfile and having Bundle manage installing them. We would like to document and expose a public API that plugins can use to accomplish goals like this, so that everyone can have their specific needs met without imposing a maintenance cost on every single user.

MOZGIII:
@indirect I believe your estimate of users that need this feature is incorrect.

Other than that, a useful plugin system would probably do the job too, at least for me. Having the ability to implement something like that as a plugin, and define it's usage in `Gemfile` would also be great.

However, in this case, it seems to me that "core feature" implementation would be better that a plugin, just for the reasons above. Stability, maintenance, and all of that - plugins, as well as gem in general has lower change to be supported over time, so people like the author of this PR would still choose not to use them.

It is true that supporting this change over the years would take effort. But supporting something like bundler will anyways will take effort - with this changes or not. Really, changes are not that big, and I don't see a *significant* growth of complexity in supporting this code.

indirect:
@MOZGIII thanks for the feedback. can you say more about why you believe this feature has a very large audience? (I am pretty much right on the edge about this particular feature... it seems useful to some people, but not everyone, or even a majority of users).

MOZGIII:
@indirect I think this feature would be more useful because of how obvious it's use case is. Also, given that much people managed to somehow show interest of this feature on github, it is far more likely it would be used by many more than that. So, that said, I just feel like the estimate would be more like 15000, than 150 people, which, however, may still be insignificant to you. But that was my point.

By the way, other use case: when using vagrant or docker for development, you often want to install some limited set of gems, i.e. rubocop or other linters, and make sure their version is matches the one that is used by CI server. So I would make a `:host` group and install only it to easily setup my dev env.
This, IIRC, was the reason I originally needed this back at 2010.

By the way, I think this feature would be more popular than optional groups.

MOZGIII:
:lgtm:

---

Reviewed 5 of 6 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/bundler/bundler/4907#-:-KR_RNGYiRoWnQkM43Ji:bnfp4nl)*

RochesterinNYC:
I'm conflicted about this change. The use cases stated do actually seem individually solid (especially when it comes to CI and automation), but they also do feel like use cases that exist at the edge of Bundler's audience (i.e massive Gemfiles, that are complex not necessarily because they have a large number of gems, but because they have a large number of groups that are organized in a way that presumably make sense). I'd be interested in seeing references/links to a few of these kinds of Gemfiles.

For example, for a `Gemfile`, how often do new groups get added? Is it often enough and is the danger of forgetting to add a group to the `--without` blacklist catastrophic enough to warrant the future support load?

rymohr:
@RochesterinNYC all great points. In our case no, groups are not added often. We've pretty much settled on `default development test production enterprise`.

As far as support load goes though, `--only` isn't exactly a huge leap from `--with` and `--without`.

MOZGIII:
We don't add a lot of groups either. Only `:host` and `:ci` groups so far.
Host group is meant to install gem (rubocop i.e.) to the host machine when developing in the vagrant environment. CI group does the same for lightweight jobs, like static code analysis, on the build server.
Though, other groups (not to be used with `--only`) are added on per-environment basis. This, however, does not relate to the usecases we have here.
Unfortunately I can't share the actual `Gemfile` to demonstrate my point.

will-in-wi:
In my company's case, we usually have `test`, `development`, `common_dev`, `stage`, and `production`. I'm now introducing `deployment`, and I've also recently set up `profile` on a project.

The issue honestly isn't huge Gemfiles. Running `rails new --api appname` and then `bundle install --path vendor/` on Rails 5.0.0.1 results in a 50M folder. Without `--api`, it is 56M. 19 vanilla applications will take up 1G in just gem files files. However, I have quite a few apps that are larger than that. Nokogiri (part of core rails, and unneeded for deployment) is 24-25M. Eventmachine is 25M. Fog is 22M. FFI is 16M. This adds up rapidly. On our prod servers, we have about 2G in installed gems alone (after we clean up old, unused gems with `bundle clean`, it'll sometimes balloon to 8G unattended).

It's also a matter of speed. If CI is running against a new version of an application which happened to upgrade Nokogiri, the build time for that Gem is now involved in every process. If I can filter it out for CI steps on various servers that run Bundler Audit, Rubocop, Capistrano, Brakeman, et cetera, it'll speed up the overall build dramatically. Heck, if I had --only, I would make the rubocop CI build only install rubocop. I could have a couple dozen groups quite easily.

There are workarounds for all of this, certainly. However, if I have the ability to easily specify that a particular task should only install one group, then I will start using groups far more. They are actually made much, much more useful with `--only`.

dreyks:
Actually the CI use case is a winner for this option. I just need the capistrano gem and capistrano-related stuff

imagine-ecommerce:
I would also greatly appreciate the '--only' option. Our use case: In one of our Rails projects, we deploy to about 50 servers with Capistrano. Some are big servers that run our app, some are very small and have only one CPU core - their purpose is supplementary to our Rails app and they need only a subset of files and gems installed.

Every time we do a bundle update and the minor version number for i.e. libv8 changes, udating the gems and compiling native extensions can take up to several hours on our small servers!

This is completely unnecessary and I haven't found a clean way to get around this issue yet.

atodorov:
ping, any update on this ?  Code review is positive, CI checks all PASS and there is a great number of people interested in this feature (myself included) which means it is a common use-case rather than an edge case.

will-in-wi:
FYI: I just answered a question on Stack Overflow wherein the `--only` option would have made this easier for them: https://stackoverflow.com/questions/41385579/capistrano-rails-undefined-method-already-invoked-for-raketask-when-settin/41399845#41399845

Sija:
@segiddins ping?

segiddins:
We still haven't decided if this is a feature we want to support

Sija:
@segiddins even though there's whole thread of pleas, SO topic, many Google searches leading here and a bunch of devs asking for that on many other channels? Well... I guess you have a plenty of reason to click this "merge" button finally... What would be the downsides of having such feature? Can't see any, can you?

colby-swandale:
> @segiddins even though there's whole thread of pleas, SO topic, many Google searches leading here and a bunch of devs asking for that on many other channels? Well... I guess you have a plenty of reasons to click this "merge" button finally... What would be the downsides of having such feature? Can't see any, can you?

@Sija please watch your tone, you're coming off as very rude and entitled which is completely unfair to everyone that works on this project. Please be considerate and i remind you of our code of conduct.

Sija:
@colby-swandale Indeed! Sorry for that! leaving this issue unresolved, without any reply for almost half of a year tho' might come for some as rude too mind you.

MOZGIII:
What can we do at this point to get the issue solved? Should we reconsider the proposed feature implementation and design it better? Or discuss the intent of having the feature in the first place?

indirect:
The core of our hesitation to accept the PR boils down to this comment from @will-in-wi, when asked why he wants this feature to be in core rather than use the already-existing `bundle-only` gem:

> I would not use this Gem because it doesn't have a maintenance record. I have no idea whether in a year or two a new version of Bundler would break any of the APIs and @MOZGIII might have moved on to Elixer or something.

The reasons given for wanting this feature in core also perfectly describe why merging this PR will mean more work for the core team: if we merge, we are agreeing to maintain not just the code in this PR, but also every single way that the code in this PR interacts with every other piece of code in Bundler... forever.

On the other side, as noted by @Sija, there are "many" developers who all very much want this feature to exist—but none of them help the core team with the maintenance that keeps Bundler working for the literally millions of users who don't care about this feature and will never use it.

So from my perspective, this ticket can be summarized as some people telling me (and the rest of the core team) "my life would be better if you agreed to do work for me for free for as long as Bundler exists".

While we very much want to make Bundler as useful as possible for as many people as possible, supporting exponentially expanding feature interactions in order to help one one thousandth of a percent of the developers who use Bundler is a genuinely hard call, and we haven't made a final decision yet.

MOZGIII:
That's a great answer, thanks for making it clear.

I don't really know how to measure wheather a userbase of a project like bundler wiould or would not use a feature. I guess, the simplest way to know would be to just fork the project, extend it heacily with load of such "scary to maintain" for upstream features and see whether it's going to gain any sort of popularity. However, that would take some kind of effort. There are probably very few people that would give that effort, which is fully understandable. There is also the fact that bundler is deeply integrated in ruby world, which does not help with forks popularity. Bundler currently has 1782 forks, but I haven't ever used one.
Another option would be something like uservoice. Did you ever considered having one?

Bundler is a great project, and it's maintained by great people. We all appreciate your work.

PS: there is something that wierds me out in the argument about working for someone for free. From this perspective, it looks really bad and unfair. But the thing is you don't really have a choice. Whether you will or will not merge something, you will still effectively work for me (and every other bundler user) for free (in general) for as long as bundler exists. The only difference is wherether you do or do not make a life easier for someone who needs a particular feature. I appreciate you hard job and really like how stable bundler is, but I think that point of view may do more harm than good.

chrismo:
"Whether you will or will not merge something, you will still effectively work for me (and every other bundler user) for free (in general) for as long as bundler exists." 

Sure, but I think that ignores a previously discussed point: the question to the larger community is "what set of features do you want to be maintained by the core group for the length of the project." There's a limited amount of time, what do we all want worked on? Every feature we commit to supporting means less time for other things, etc.

MOZGIII:
@chrismo yes, it does purposely ignore it, furthermore: it does not impose any constraints on how the "work" is done and what is the result. Just like the original discussed phrase does. But I can see your point and it's absolutely valid.
I guess people in general don't know much about what bundler's core team is doing, and planning to do. Would be great if there was big arrow in the readme pointing to what's current in the mind of the core team. I really just don't know what you're doing guys, and probably it's like a 1000 times more important that what we're trying to merge here.
Speaking of other stuff, I've heard something about plugins in bundler, maybe we could implement this feature via plugin?
We have this feature implemented as an "independent" `bundler-only` gem, which relies heavily on bundler's internals and works using "unsafe" stuff, because those internal could we rewired differently in any new bundler version. If plugins support was to provide a safer and maintained API, that would really lower the need of adding this feature to the core, at least for my purposes.

will-in-wi:
I don't usually worry about such things because it doesn't really matter, but given this comment from @indirect :

> On the other side, as noted by @Sija, there are "many" developers who all very much want this feature to exist—but none of them help the core team with the maintenance that keeps Bundler working for the literally millions of users who don't care about this feature and will never use it.

The company I work for is a financial supporter of Ruby Together and by proxy Bundler development. We do "help the core team with the maintenance that keeps Bundler working for the literally millions of users who don't care about this feature and will never use it."

To be super, super clear: We would and will continue to support Ruby Together and Bundler financially regardless of the result of this ticket. It is not a factor at all in that decision. But given that the people-who-use-software-for-free-without-contributing-want-a-feature argument came up, I just wanted to point out that we put our money where our mouths are. We partially support all of this exactly because we use Ruby in an enterprise environment, it is critical and core to our infrastructure, and we've put countless hours into software which depends on the continued viability of Ruby, Rails, Bundler, et all. To this end, I put additional, company supported, time into a feature request and patch for bundler which will make our lives easier with our 100+ apps going to 30+ servers.

If this isn't merged because the core devs, who spend a heck of a lot more time on this codebase than I do and know a lot more than I do about this product, don't want to support it, more power to them. We financially support them because we trust their decision-making in cases like this.

will-in-wi:
I'll underscore my last comment: THANK YOU ALL FOR WORKING ON BUNDLER!!!!!!! This software is best of breed in the industry, IMO, and the reason it is is all of you who spend countless hours on it. I trust that ultimately you will make the right decision for the community and I'll work with or around whatever you decide. I benefit tremendously from your work, and can't say how many times I've been happy to work in a language with such a tool. I love Ruby for a lot of reasons, but Bundler is one of the primary ones.

indirect:
hey @will-in-wi, thanks for chiming in! I'm really sorry that what I said came across as aimed at you—in writing this PR, and in supporting Ruby Together, you've absolutely participated and helped with Bundler. You are great, and you have not at all given the impression that you are asking us to work for you for free. ❤️ 

My comment was intended to respond specifically to the statement "Well... I guess you have a plenty of reasons to click this "merge" button finally... What would be the downsides of having such feature? Can't see any, can you?" Claiming that there are no downsides is where the feeling of demanding free work comes in. There are in fact downsides, and we're still trying to figure out whether the upsides of bringing this into core outweigh the downsides or not.

@MOZGIII We've made some progress on the plugin system, but we don't really have anyone using it yet. If you're interested, I would be happy to worth with you to figure out what plugin APIs are needed for `bundler-only` to work.

MOZGIII:
@indirect Ok, lets try with plugin system. I guess using the plugin system here would benefit bundler even more. Hopefully, together we will be able to polish plugin system to suit the needs of any `bundler-only` like feature. I can only do it on my free time though.
No docs yet? I haven't found any.
Should we find some other place to work on this new project without distracting people here?

will-in-wi:
Thanks, @indirect. I totally understand pushing back against a feature due to the maintenance burden. It's funny, because I'm usually on the other end of this question. 😄

In my opinion, the changes to support this are minimal, and the maintenance burden similarly minimal. However, I totally recognize that this is an opinion, and one from someone who is not a Bundler developer. I'll try to make a more specific case as to why I think this can be considered a minimal change.

Six files are changed. Of these six, two are just argument handling, two are getters and setters, one is tests, and one line in one file is the actual logic.

75 new lines, and one changed. 52 of the new lines are tests. Of the remaining 23 lines, 8 are argument handling, 14 (and the changed line) are getters and setters, and one is the actual logic. Not counting tests, a 23 line change in a 88,816 line (Ruby code) is pretty miniscule (0.026%).

Of course, how much is changed is only a small part of this, what is changed is usually far more important.

What generally makes code a maintenance burden IMO is changing or adding to code which frequently changes as additional features are added (kind of a corollary to the Open/Closed principle).

With that in mind, I would consider the 52 lines of tests to have a near-zero (but not zero) maintenance burden. Unless the feature itself changes (which would likely be proof of its usefulness) or the tests change en-masse, the test code likely doesn't change.

The eight lines of argument handling code are also minimal impact, since they only have to change if the entire argument handling changes or the feature changes. In the former case (this also applies to the tests), because this feature isn't doing anything unusual or different and is just a variation on argument handling for `--with`, the added burden is theoretically minimal as there is very little additional work regardless of the existence of this feature.

The getter/setter logic is where we start adding complexity. One positive characteristic is that, similarly to the argument handling, the vast majority of this code is a copy/paste modification of `--with` and `--without`. So if other code around this changes, those features would change too, anyway. The additional cognitive load of managing this, over and above those other two features, is minimal. If desired, I might be able to refactor the `--with`, `--without`, and `--only` triad into fewer (more complex) lines of code because of the similarity.

The most complex change is probably the one-liner in `lib/bundler/definition.rb`. The scope of this method is fairly simple, it is determining the list of groups to be processed. The nature of this change is that it short-circuits the main logic and only returns its own list. This is likely to continue to be the algorithm, regardless of what other changes or additions are made. Even if the remaining logic becomes more complex, the `--only` feature will always short-circuit.

Does this make sense? Metrics like lines of code and such only go so far, but I'm trying to root this discussion in something concrete if possible.

Thanks for considering this!

MOZGIII:
Truly, the impact is small indeed, and the solution is elegant.
A proper plugin implementation though would probabaly lead to significantly bigger changes in core, and there is also the plugin code itself. It seems like plugin system is not mature yet, so there is also some maintenance cost for the plugin.
By the way, it's the unmanged possibility of changes in bundler internals that's worries me in `bundle-only` implementation. Plugin system is great, but it either needs a stable API for plugins to use, or some guarantees on the internals stability (with is not great for what I can tell, cause we usually want to change stuff) - that is if plugins were to anyhow monkeypatch the internals to do their job.
You would also have to install something other than "just bundler" for a CI job (in my case) to be able to use the feature. That's manageable, but annoying.

sekrett:
Let me say that CI is very popular nowadays. I would like to have this option in the bundler, not in an external gem.

dreyks:
any news? I think this should either be merged or closed with an explicit `won't implement`

cycomachead:
It would be helpful to have some clarity with the future of this.  I'm working with getting some devs setup with the local environments, and this would be a huge help. 

For me, I'd love to be able to easily too `bundle install --only linters`, which as you might except is a list of (mostly) pronto-related Gems to locally run our linters. 
We use Docker, primarily, but it's very very helpful to have developers run linters on their local machines. 
Without this patch, we need to specify `--without default development test production staging`, which while typeable, is much less clear about what's going on, is harder to remember, and liable to breaking if we need to add a new group to the gem file. Fortunately, that doesn't happen often, but when it does it would require updating docs and scripts and so on, whereas an `--only` situation would be much more stable. 

While I understand this is a subset of bundler users, as configurations expand over time and more of us are using various container-type tools, I think the use of groups only expands over time. 

Anyway, I just wanted to share that perspective and use case. Thank you all for the hard work and the maintenance of bundler. <3.

PR #4908, opened 2016-08-23
Add missing definition of `respond_to_missing?`

okkez:
Because we should overwrite this method when we overwrite
`method_missing`.

See http://ruby-doc.org/core-2.3.1/BasicObject.html#method-i-method_missing


segiddins:
Good catch!
@homu r+

PR #4917, opened 2016-08-25
[Settings] Allow temporarily overriding settings in-memory

segiddins:


indirect:
Can you explain the rationale for this?

segiddins:
Sure! We use settings in two ways -- one is user preferences, stored via `bundle config`, and the other is as global internal configuration. When attempting to refactor out `--frozen` for 2.0, I ran into issues getting rid of the auto-saved config for `Bundler.settings[:frozen]` because it's used in places where injecting the configuration value isn't really possible

indirect:
@homu r+

That's a good point about these being two separate things—maybe we will want to make them explicitly different in the future?

segiddins:
Blocked on https://github.com/bundler/bundler/pull/4928

segiddins:
@homu retry

PR #4918, opened 2016-08-25
[DSL] Warn on invalid options instead of raising

### LINKED ISSUE #4833: Allow forward compatibility with new options by printing a warning instead of an error
If future versions of Bundler add options or values (for `platform`, or `gem`, Bundler.settings, etc), older versions will currently raise an error like "not a valid option". That is technically true for that version of Bundler, but it's confusing if that option is a valid option in a later version.

It would be a better UX to print a warning on options that are not recognized, and suggest updating to a newer version of Bundler if the user needs that option/platform/setting/etc. That both helps users solve their own problem when an option isn't recognized, and makes it possible for older versions of Bundler to parse Gemfiles and print this warning.

segiddins:
Closes #4833, if we're sure this is what we want to do

### DIFF
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -382,20 +382,21 @@ module Bundler
       opts["optional"] ||= false
     end
 
-    def validate_keys(command, opts, valid_keys)
+    def validate_keys(command, opts, valid_keys, raise_on_invalid_keys = false)
       invalid_keys = opts.keys - valid_keys
-      if invalid_keys.any?
-        message = String.new
-        message << "You passed #{invalid_keys.map {|k| ":" + k }.join(", ")} "
-        message << if invalid_keys.size > 1
-                     "as options for #{command}, but they are invalid."
-                   else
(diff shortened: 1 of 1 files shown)


indirect:
I'm happy to discuss it... the reason we added raising on unknown values was to resolve the (many, many) support tickets about typos or stray spaces or quote mark problems. I feel like a warning helps those users just as much, without locking out users who knowingly want to use a Gemfile with an older Bundler?

I guess I'm also open to the argument that Bundler should refuse to run on a Gemfile with unknown values, but in that case I think I'd at least like to improve the error message to say something like "You may be able to resolve this by upgrading Bundler to the newest version." WDYT?

segiddins:
I'd personally prefer to keep it an error and say `You may be able to resolve this by upgrading Bundler to the newest version` or something like that -- that way, we don't have to worry about people ignoring the warnings and being confused by some flag or another not being respected

indirect:
@segiddins okay, I buy that. Let's just adjust the error, then.

Issue #4913, opened 2016-08-25
`bundle config NAME` is too verbose by default


BM5k:
And doesn't provide a non-verbose option.

```
Bundler version 1.12.5
ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-darwin16]
rubygems 2.6.6
```

> What you're trying to accomplish

I would like to query the value of a config setting.

> The command you ran

`bundle config my-setting`

> What you expected to happen

return the value associated with `my-setting`, possibly just the value, possibly as `KEY=VALUE`

> What actually happened

Information overload!

indirect:
I think this is a good point. We have added the `--parseable` option for this type of use, and I think it would be useful to add it here.

BM5k:
Yes, that would be perfect, since my end goal was for this to be machine readable.

lynnco:
Relevant PR https://github.com/bundler/bundler/pull/4919

colby-swandale:
Hi 👋 , We have added a `--parseable` option to `bundle config` in the latest release.

PR #4919, opened 2016-08-26
Feature: Add parseable option to bundle config command

JuanitoFatas:
This Pull Request implements #4913:

```
$ bundle config gem.coc --parseable
=> true
```
## 

Related: #4871

<details>
<summary>

Feedback for development of Bundler</summary>

Cannot use byebug / pry to dive into the code. 😢 
Cannot use puts / p to print something for debugging. 😢 
Currently use `Bundler.ui.info` to print something for debugging. 😅 
</details>

### DIFF
--- a/lib/bundler/cli/config.rb
+++ b/lib/bundler/cli/config.rb
@@ -21,6 +21,8 @@ module Bundler
         return
       end
 
+      return Bundler.ui.info(Bundler.settings[name]) if options[:parseable]
+
       unless valid_scope?(scope)
         Bundler.ui.error "Invalid scope --#{scope} given. Please use --local or --global."
         exit 1
--- a/spec/commands/config_spec.rb
+++ b/spec/commands/config_spec.rb
@@ -188,8 +188,32 @@ describe ".bundle/config" do
   end
 
   context "with --parseable option" do
-    it "returns value associated with the config" do
-      global_config "BUNDLE_GEM__COC" => "true"
+    it "returns empty when not set" do
+      bundle "config gem.coc", :parseable => true
+
+      expect(out).to eq ""
+    end
+
+    it "returns true when set" do
+      bundle "config gem.coc true"
+
+      bundle "config gem.coc", :parseable => true
+
+      expect(out).to eq "true"
+    end
+
+    it "returns empty string when unset" do
+      bundle "config gem.coc true"
+      bundle "config --delete gem.coc"
+
+      bundle "config gem.coc", :parseable => true
+
+      expect(out).to eq ""
+    end
+
+    it "returns default if global default presents when unset" do
+      bundle "config --delete gem.coc"
+      bundle "config --global gem.coc true"
 
       bundle "config gem.coc", :parseable => true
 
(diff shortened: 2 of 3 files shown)


indirect:
@JuanitoFatas I personally use `RUBYOPT=-rpry dbundle` to require pry and then run commands. Maybe we should improve the development documentation. :)

lib/bundler/cli/config.rb
```diff
@@ -21,24 +21,28 @@ def run
         return
       end
 
-      unless valid_scope?(scope)
-        Bundler.ui.error "Invalid scope --#{scope} given. Please use --local or --global."
-        exit 1
-      end
+      if options[:parseable]
+        Bundler.ui.info(Bundler.settings[name])
```
indirect: I think probably just return here instead of putting everything inside an `else` block.

JuanitoFatas: 👍 Thanks I've updated.

spec/commands/config_spec.rb
```diff
@@ -187,6 +187,16 @@
     end
   end
 
+  context "with --parseable option" do
+    it "returns value associated with the config" do
```
segiddins: please add a test for when the value is unset, and also when the value is unset but bundler has a default value

lib/bundler/cli/config.rb
```diff
@@ -21,6 +21,8 @@ def run
         return
       end
 
+      return Bundler.ui.info(Bundler.settings[name]) if options[:parseable]
```
segiddins: What happens if `parseable` is passed when setting a value?

JuanitoFatas: Did you mean doing something like:

```
$ bundle config --local parseable true
$ bundle config --global parseable true
```

I tried, this line of change, does not affect original functionality.

segiddins: is there a test for it?

JuanitoFatas: I'm sorry, added in 2ba6079. [The failure](https://travis-ci.org/bundler/bundler/jobs/160111059#L608) is fixed in #4990.

segiddins: Sorry, I meant `bundle config --local --parseable foo bar`

indirect:
@segiddins was this failure a new rails release breaking the edgecases spec?

lib/bundler/cli/config.rb
segiddins: I think this is the only blocker left for merging this

colby-swandale:
Is this being worked on @JuanitoFatas? I have a use case for this and would like to see this feature merged possibly in 1.14.0. I'm happy to finish it off if you want.

JuanitoFatas:
> I'm happy to finish it off if you want.

Please do, thank you!

segiddins:
@colby-swandale as 1-14-stable has already been branched, this won't make it in to 1.14

colby-swandale:
@segiddins thanks for letting me know. I'll still work on getting this merged in.

PR #4928, opened 2016-08-29
Update specs to use compact_index 0.11

segiddins:
Also removed the penalty for adding requirements to dependencies!

\c @indirect


segiddins:
@homu r+

PR #4938, opened 2016-08-31
Add some specs for Bundler::Settings#temporary

JuanitoFatas:
This PR adds a spec and moved an alias in Settings class.

followup of #4917.


segiddins:
Thanks!
@homu r+

PR #4943, opened 2016-09-02
[DSL] Suggest updating bundler on invalid options

segiddins:
See https://github.com/bundler/bundler/issues/4856 & https://github.com/bundler/bundler/pull/4918


segiddins:
@indirect r?

indirect:
@homu r+

PR #4944, opened 2016-09-02
[RubygemsExt] Add missing require rubygems/source

dekellum:
The change referenced below as released in 1.13.0.rc.2, may attempt to reference Gem::Source without it being loaded, resulting in:

~~~
  [!] There was an error parsing `Gemfile`:
  [!] There was an error while loading `elided.gemspec`: uninitialized constant Gem::Source. Bundler cannot continue.
~~~

Observed this on ruby 2.2.5 with stock rubygems 2.4.5 as well as upgraded rubygems 2.6.6. Add this require.

f9de70ee931ca4a8500916fa9480f6df6c062626  by @segiddins:
> [RubygemsExt] return Source::Installed from #source when appropriate

### DIFF
--- a/lib/bundler/rubygems_ext.rb
+++ b/lib/bundler/rubygems_ext.rb
@@ -8,6 +8,7 @@ end
 
 require "rubygems"
 require "rubygems/specification"
+require "rubygems/source"
 require "bundler/match_platform"
 
 module Gem
--- a/lib/bundler/rubygems_ext.rb
+++ b/lib/bundler/rubygems_ext.rb
@@ -8,7 +8,15 @@ end
 
 require "rubygems"
 require "rubygems/specification"
-require "rubygems/source"
+
+begin
+  # Possible use in Gem::Specification#source below and require
+  # shouldn't be deferred.
+  require "rubygems/source"
+rescue LoadError
+  # Not available before Rubygems 2.0.0, ignore
+end
+
 require "bundler/match_platform"
 
 module Gem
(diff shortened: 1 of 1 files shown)


dekellum:
Hmm... might not be that simple, as this suggests:

https://github.com/rubygems/rubygems/commit/d922a0f61

...that Gem::Source doesn't exist until rubygems 2.0.0?

segiddins:
Please add a test for the bug this is addressing -- thanks!

dekellum:
@segiddins I'll try to do that.  There are CI tests for Rubygems as far back as 1.6.2 however, and there is no  Gem::Source as referenced as far as I can tell in Rubygems versions before 2.0.0. Can you please comment on that?

segiddins:
The source method didn't exist before then either

dekellum:
Death by linter.  Wish I new how to make rubocop happy when intentionally swallowing a load error.

hsbt:
Hi, I faced same error with Ruby 2.2 and 2.1 on travis builds. see. https://travis-ci.org/ruby/rake/builds/158902251

indirect:
@homu r+

segiddins:
I don't think we should be merging this without a test

dekellum:
Sorry gents, I have not been able to reduce the error to a spec in bundler's environment. @hsbt's rake repro. gave me some hope, with a public code base, but again I can not seem to replicate the same (feature usage? load order?) in a bundler spec.  Thanks for the merge.

Apokly:
Have same error using ruby-2.3.0 and bundler-1.13.0 on continuous integration.

PR #4949, opened 2016-09-06
Document the default value for the cache_path setting

### LINKED ISSUE #4947: Documentation: specify what cache_path defaults to
It seems `cache_path` is the only forgotten one to define what it defaults to at https://bundler.io/v1.12/man/bundle-config.1.html#LIST-OF-AVAILABLE-KEYS.

segiddins:
Closes #4947.
Default taken from https://github.com/bundler/bundler/blob/96bdba7686ca39b41eebde7533148f92aee2260a/lib/bundler/settings.rb#L210


segiddins:
@homu r+

PR #4951, opened 2016-09-06
Fallback to a temp dir when the home directory is not usable

segiddins:
Closes https://github.com/bundler/bundler/issues/4778
This is an alternative to https://github.com/bundler/bundler/pull/4886
\c @allenzhao @indirect

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -143,8 +143,41 @@ module Bundler
       "#{Bundler.rubygems.ruby_engine}/#{Bundler.rubygems.config_map[:ruby_version]}"
     end
 
+    def user_home
+      @user_home ||= begin
+        home = Bundler.rubygems.user_home
+        warning = "Your home directory is not set properly:"
+        if home.nil?
+          warning += "\n * It is not set at all"
+        elsif !File.directory?(home)
+          warning += "\n * `#{home}` is not a directory"
+        elsif !File.writable?(home)
+          warning += "\n * `#{home}` is not writable"
+        else
+          return Pathname.new(home)
+        end
+
+        login = Etc.getlogin || "unknown"
+
+        tmp_home = Pathname.new(Dir.tmpdir).join("bundler", "home", login)
+        begin
+          SharedHelpers.filesystem_access(tmp_home, :write) do |p|
+            FileUtils.mkdir_p(p)
+          end
+        rescue => e
+          warning += "\n\nBundler also failed to create a temporary home directory at `#{tmp_home}`:\n#{e}"
+          raise warning
+        end
+
+        warning += "\n\nBundler will use `#{tmp_home}` as your home directory temporarily"
+
+        Bundler.ui.warn(warning)
+        tmp_home
+      end
+    end
+
     def user_bundle_path
-      Pathname.new(Bundler.rubygems.user_home).join(".bundle")
+      Pathname.new(user_home).join(".bundle")
     end
 
     def home
@@ -403,6 +436,7 @@ EOF
       @locked_gems = nil
       @bundle_path = nil
       @bin_path = nil
+      @user_home = nil
 
       Plugin.reset!
 
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -3,6 +3,8 @@ require "fileutils"
 require "pathname"
 require "rbconfig"
 require "thread"
+require "tmpdir"
+
 require "bundler/errors"
 require "bundler/environment_preserver"
 require "bundler/gem_remote_fetcher"
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -156,7 +156,7 @@ module Bundler
         elsif !File.writable?(home)
           warning += "\n * `#{home}` is not writable"
         else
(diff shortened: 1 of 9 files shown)


lib/bundler.rb
```diff
@@ -143,8 +145,41 @@ def ruby_scope
...
+        home = Bundler.rubygems.user_home
+        warning = "Your home directory is not set properly:"
+        if home.nil?
+          warning += "\n * It is not set at all"
+        elsif !File.directory?(home)
+          warning += "\n * `#{home}` is not a directory"
+        elsif !File.writable?(home)
+          warning += "\n * `#{home}` is not writable"
+        else
+          return Pathname.new(home)
```
ixti: This will interrupt `begin` block, return `Pathname.new(home)` and will leave `@user_home.nil?`

ixti: In other words, in case of happy path `@user_home` will remain `nil` and lines `149..159` will be executed every-time on `#user_home` call.

lib/bundler.rb
```diff
@@ -143,8 +145,41 @@ def ruby_scope
...
+        warning += "\n\nBundler will use `#{tmp_home}` as your home directory temporarily"
+
+        Bundler.ui.warn(warning)
+        tmp_home
+      end
+    end
+
     def user_bundle_path
-      Pathname.new(Bundler.rubygems.user_home).join(".bundle")
+      Pathname.new(user_home).join(".bundle")
```
ixti: `user_name` is `Pathname` already. So it should be safe to leave `user_home.join(".bundle")`

lib/bundler/ui/silent.rb
```diff
@@ -13,6 +17,7 @@ def confirm(message, newline = nil)
       end
 
       def warn(message, newline = nil)
+        @warnings |= [message]
```
indirect: Seems like this will mean that warning message strings will never be garbage collected?

lib/bundler/gem_helper.rb
```diff
@@ -98,7 +98,7 @@ def rubygem_push(path)
         allowed_push_host = @gemspec.metadata["allowed_push_host"]
         gem_command += " --host #{allowed_push_host}" if allowed_push_host
       end
-      unless allowed_push_host || Pathname.new("~/.gem/credentials").expand_path.file?
+      unless allowed_push_host || Bundler.user_home.join(".gem/credentials").file?
```
indirect: Does `expand_path` raise if HOME isn't set? If it doesn't raise, I think we shouldn't create or use the temporary directory when looking for gem creds.

indirect:
@segiddins ping

segiddins:
I took a day off of working, I'll answer tomorrow

lib/bundler/ui/silent.rb
ixti: Sorry, I accidentally removed @indirect's comment. Original comment was saying that (recovering from memories):

> Seems like this means message will never be garbage collected.

segiddins: until the `Silent` instance is deallocated, yes

lib/bundler/gem_helper.rb
segiddins: `in `expand_path': non-absolute home (ArgumentError)`

lib/bundler/ui/silent.rb
indirect: How do you feel about making this a `Set` instead of using the (esoteric and relatively confusing) `|=` operator?

lib/bundler/gem_helper.rb
indirect: Okay, as I suspected. That means we don't want to use `Bundler.user_home` here. In fact, this makes me think that all this logic should be inside the CompactIndex cache writer, and not inside `Bundler.user_home`, because we actively don't want to use a temp directory for either writing a settings file or looking for `.gem/credentials`. We specifically and only want to use a temp directory for the compact index files.

indirect:
This is my preferred implementation of a fix for #4778, once the comments are addressed. Thanks! 👍

lib/bundler/ui/silent.rb
indirect: How do you feel about using a `Set` instead of the (esoteric and relatively confusing) `|=` operator?

lib/bundler/gem_helper.rb
segiddins: Why don't we want to use a temp directory for writing a settings file? It would allow users without a home directory to have 'global' settings for at least a single usage session

lib/bundler/ui/silent.rb
segiddins: That's fine, I think I was just copying what the `Shell` class did

indirect:
Looks good. I didn't realize that `Dir.tmpdir` returned the same directory across processes! 👍 
@homu r+

brodock:
I would love to have an initial review here https://github.com/bundler/bundler/pull/6550 

It implements a workaround for more hardened installations like omnibus bundled software, where home might not be owned or writable by the user, but would still have required `.bundle` and `.gem` folders with correct permission.

PR #4952, opened 2016-09-07
[Installer] Skip eval’ing a Gemfile twice when there’s a lockfile

### LINKED ISSUE #3096: Gemfile is evaluated twice when a lockfile exists

segiddins:
Closes #3096. Closes #4417.

\c @indirect

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -169,6 +169,7 @@ module Bundler
 
     def missing_dependencies?
       missing = []
+      resolve.materialize(current_dependencies, missing)
       return false if missing.empty?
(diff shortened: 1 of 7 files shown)


indirect:
@homu r+

PR #4954, opened 2016-09-07
[Installer] Load gem plugins when installing

### LINKED ISSUE #2824: Bundler fails to load all rubygems plugins
When running with GEM_HOME set to a location and a rubygem plugin installed to that location the gem command will load the plugin, but bundler will not. 

This is because bundler calls Gem.load_env_plugins (in bundler::CLI) but not Gem.load_plugins, rubygems calls both (in rubygems/gem_runner.rb). The simple solution is to add a Gem.load_plugins call along side the Gem.load_env_plugins call in bundler::CLI. A better solution (although a far bigger job) would be to move all interaction with the rubygems libraries to Bundler::RubygemsIntegration and have it initialize the rubygems libs appropriately for everything. This would simplify things greatly, but I'd rather have some feedback before attempting the bigger change.

segiddins:
Closes #2824 I think

### DIFF
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -60,9 +60,6 @@ module Bundler
           "the --binstubs option will be removed in favor of `bundle binstubs`"
       end
 
-      # rubygems plugins sometimes hook into the gem install process
-      Gem.load_env_plugins if Gem.respond_to?(:load_env_plugins)
-
       Plugin.gemfile_install(Bundler.default_gemfile) if Bundler.feature_flag.plugins?
 
       definition = Bundler.definition
--- a/lib/bundler/cli/update.rb
+++ b/lib/bundler/cli/update.rb
@@ -54,9 +54,6 @@ module Bundler
 
       Bundler.settings[:jobs] = opts["jobs"] if opts["jobs"]
 
-      # rubygems plugins sometimes hook into the gem install process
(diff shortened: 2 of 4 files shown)


indirect:
@segiddins what's the difference between `load_plugins` and `load_env_plugins`, and why did you switch from one to the other?

segiddins:
Because that seemed to be what the issue was asking for

indirect:
@segiddins I am fairly certain it is not okay to remove `load_plugins`. Please either use both or research to figure out what the "right" (ha) answer is.

segiddins:
Where did I remove `load_plugins`? Honestly, if we don't know what www want to do, I'd rather not touch this. I was just making the exact change the linked issue asked for ¯\_(ツ)_/¯

indirect:
I'm not comfortable accepting this until we understand the difference between `load_plugins` and `load_env_plugins`, so I'm sure that this change doesn't break anything.

segiddins:
`load_plugins` loads all rubygems plugins and is what rubygems does when invoked via its own CLI, `load_env_plugins` just loads plugins that are already on the `$LOAD_PATH`

indirect:
@segiddins okay. I hope we don't regret this later when it turns out there are rubygems plugins that we don't actually want to load. 😅 

@homu r+

Issue #4966, opened 2016-09-11
Bundler cannot resolve right relative paths since version 1.11.2 (Test example attached)


the-teacher:
Hello, have no idea why it worked for me so long time and why I didn't meet it before, but looks like problem exists.

### My case

I have a rails based CMS. I use it for some projects. Some pojects based on this CMS have to use some specific gems. That is why in my `Gemfile` I use something like this:

```ruby
source 'https://rubygems.org'
gem 'rails'
eval_gemfile "./Gemfiles/#{ ENV['PROJECT_NAME'] }_specific_gems.rb"
```

Also, in my app I have
* a directory for specific Gemfiles (folder has name `Gemfiles`)
* a directory for specific gems (folder has name `gems`)

### The problem

Looks like since version `1.11.2` bundler cannot resolve right paths to my local gems.

When I use paths like this:

```ruby
gem 'gem1', path: './gems/gem1'
```

Bundler gives to me:

```
/PATH/TO/RELEASES/RAILS_CMS/current/Gemfiles/gems/gem1
```

But it should be `/PATH/TO/RELEASES/RAILS_CMS/current/gems/gem1`, and it was so before.

And when I use

```ruby
gem 'gem1', path: '../gems/gem1'
```

Bundler gives to me:

```
/PATH/TO/RELEASES/RAILS_CMS/gems/gem1
```

But actually I expected to get:

```ruby
/PATH/TO/RELEASES/RAILS_CMS/current/gems/gem1 # <= this one is right path, but it's impossible to get it somehow

/PATH/TO/RELEASES/RAILS_CMS/current/Gemfiles/gems/gem1 # <= this one is wrong 
/PATH/TO/RELEASES/RAILS_CMS/gems/gem1           # <= this one is wrong too
```

### TEST CASE

There is simple test case to reproduce my problem

https://github.com/the-forks/bundler_test_case

You can find steps to reproduce in a README.md

PS:
I tried to play with ENV `BUNDLE_GEMFILE`. It didn't help

segiddins:
This was an intentional change. Bundler now expands relative paths relative to the _gemfile being evaluated's directory_ rather than the parent gemfile.

The following diff fixes things!

```diff
diff --git a/Gemfiles/specific_1.rb b/Gemfiles/specific_1.rb
index a9c9f87..afd8c8a 100644
--- a/Gemfiles/specific_1.rb
+++ b/Gemfiles/specific_1.rb
@@ -1,2 +1,2 @@
[...]

PR #4976, opened 2016-09-13
[Path] Store lockfile paths relative to the root, not the gemfile path

segiddins:
Closes https://github.com/bundler/bundler/issues/4966, but i'm wondering if this will break anything else?

### DIFF
--- a/lib/bundler/source/path.rb
+++ b/lib/bundler/source/path.rb
@@ -46,7 +46,7 @@ module Bundler
 
       def to_lock
         out = String.new("PATH\n")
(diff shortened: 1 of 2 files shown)


the-teacher:
Thanks for a fast reaction. I hope it will solve the problem.

indirect:
@segiddins I feel like we go back and forth on this every teary or two, but if all the tests pass I guess we’re slowly approaching something that actually works?

indirect:
@homu r+

segiddins:
@homu retry

segiddins:
@homu retry

PR #4977, opened 2016-09-13
Fix #4934. Make GVP _after_ eager unlock.

### LINKED ISSUE #4934: Inconsistency with `bundle update` (major vs minor/patch)
Here's a [test case](https://github.com/marcandre/bundler-cases/blob/f1d57f49cc4871b5eda803825c649f982fd719ac/cases/conservative/very_basic.rb) aiming to illustrate the inconsistency with `bundle update foo` with different `--major`/`minor`/`patch`. The only available gem versions are patches releases, so all three options should all behave the same but they don't for [the `minor` and `patch` options](https://github.com/marcandre/bundler-cases/blob/f1d57f49cc4871b5eda803825c649f982fd719ac/cases/conservative/very_basic.rb#L43-L59).

I feel this inconsistency has to be bug. For sure, it makes it impossible to understand what is supposed to be going on.

Moved from https://github.com/chrismo/bundler-cases/issues/6

chrismo:
When Definition creates the GemVersionPromoter, it needs to do so
_after_ it's performed the eager_unlock so the GVP gets the correct list
of unlocked gems.

Prior to this fix, it had a stricter list of gems being updated with the
new `--patch` or `--minor` options used with `bundle update` and in some
cases would have inconsistent results when used without a conservative
switch or the `--major` option. See #4934 for details.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -102,13 +102,13 @@ module Bundler
       end
       @unlocking ||= @unlock[:ruby] ||= (!@locked_ruby_version ^ !@ruby_version)
 
-      @gem_version_promoter = create_gem_version_promoter
-
       add_current_platform unless Bundler.settings[:frozen]
 
       @path_changes = converge_paths
       eager_unlock = expand_dependencies(@unlock[:gems])
       @unlock[:gems] = @locked_specs.for(eager_unlock).map(&:name)
+
+      @gem_version_promoter = create_gem_version_promoter
 
       @source_changes = converge_sources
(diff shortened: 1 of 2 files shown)


segiddins:
@homu r+

segiddins:
@homu retry

chrismo:
thx @segiddins!

lynnco:
This PR seems to have been missed in the releases! Will have to investigate if its been shipped in `1.13.1` or `1.13.2`. If not if may require a `1.13.3` release

PR #4980, opened 2016-09-14
replicate cons. bundle install in update cmd

chrismo:
In the discussion on new 1.13 [Conservative Bundle
Updates](https://github.com/bundler/bundler-features/issues/122), some
users would like to have the same [Conservative
Updating](http://bundler.io/v1.12/man/bundle-install.1.html#CONSERVATIVE-UPDATING)
behavior available in `bundle install` when a declared dependency is
altered in the Gemfile, also available in the `bundle update` command.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -215,14 +215,16 @@ module Bundler
       "Update ruby specified in Gemfile.lock"
     method_option "bundler", :type => :string, :lazy_default => "> 0.a", :banner =>
       "Update the locked version of bundler"
-    method_option "patch", :type => :boolean, :hide => true, :banner =>
+    method_option "patch", :type => :boolean, :banner =>
       "Prefer updating only to next patch version"
-    method_option "minor", :type => :boolean, :hide => true, :banner =>
+    method_option "minor", :type => :boolean, :banner =>
       "Prefer updating only to next minor version"
-    method_option "major", :type => :boolean, :hide => true, :banner =>
+    method_option "major", :type => :boolean, :banner =>
       "Prefer updating to next major version (default)"
-    method_option "strict", :type => :boolean, :hide => true, :banner =>
+    method_option "strict", :type => :boolean, :banner =>
       "Do not allow any gem to be updated past latest --patch/--minor/--major"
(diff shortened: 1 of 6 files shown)


segiddins:
Is this approach preferable to adding a conservative option to the unlock hash for Bundler.definition?

chrismo:
It's easily debatable ... pushing another option into Definition seems to be bloating Definition, esp. for a piece of functionality (calculating eager_unlocking) that appears to be pretty specific to updating only. 

The other view is that it truly belongs in Definition -- that `bundle lock` possibly also cares about this functionality ... though one could step back and say `bundle lock` really shouldn't be implemented separately, but as an option of `update` and `install` ... ? 

After many thoughts thunk, I decided to pick one way to implement it, sleep on it and get some feedback.

chrismo:
@segiddins I went ahead and swapped out the design to just passing this through to Definition. My OOP OCD sort of still prefers it the other way, but it's a more intrusive/transitive change than doing the way you suggested.

chrismo:
@indirect also want some feedback on the new options naming here, additions to both bundle update and bundle lock.

indirect:
I think `--conservative` or `--lock-children` is pretty clear. There is literally a section in the `install` manpage named [CONSERVATIVE UPDATING](http://bundler.io/v1.3/man/bundle-install.1.html#CONSERVATIVE-UPDATING) so I think an option named `--conservative` that turns on that behavior is reasonably understandable/expected. 👍

chrismo:
cool. `--lock-children` could be confusing, IMO. I'll change it out to `--conservative`.

chrismo:
@indirect ready for review and @homu if you approve

indirect:
@chrismo looks great, thanks!

@homu r+

chrismo:
i canceled the homu build, realizing i hadn't squashed the commits

chrismo:
@indirect @homu ping

indirect:
@homu retry

chrismo:
Guessing this one test failure was a fluke? Looks unrelated, and only on one old matrix combo: https://travis-ci.org/bundler/bundler/builds/169640862

PR #4983, opened 2016-09-14
[RubygemsIntegration] Ensure redefined methods have the same visibility

### LINKED ISSUE #4975: Error in Bundler 1.13 with JRuby 1.7.x
Calling

Bundler.require

fails with

Gem Load Error is: private method `require' called for Kernel:Module
Backtrace for gem load error is:
/home/test/.gem/gems/bundler-1.13.0/lib/bundler/runtime.rb:91:in `require'
org/jruby/RubyArray.java:1613:in `each'
/home/test/.gem/gems/bundler-1.13.0/lib/bundler/runtime.rb:86:in `require'
org/jruby/RubyArray.java:1613:in `each'
/home/test/.gem/gems/bundler-1.13.0/lib/bundler/runtime.rb:75:in `require'
/home/test/.gem/gems/bundler-1.13.0/lib/bundler.rb:106:in `require'

The problem is in method reverse_rubygems_kernel_mixin from rubygems_integration.rb which is different from Bundler 1.12. When method implementation from 1.13 is replaced by implementation from 1.12, Bundler is working again.

It might be generic problem with Ruby 1.9 and not specifically JRuby 1.7.x, but I have no environment with MRI 1.9 to test it.

segiddins:
Closes #4975 

\c @indirect @headius 

I'd love to get this out as 1.13.2 asap, if it works

### DIFF
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -489,6 +489,7 @@ module Bundler
     end
 
     def redefine_method(klass, method, unbound_method = nil, &block)
+      visibility = method_visibility(klass, method)
       begin
         if (instance_method = klass.instance_method(method)) && method != :initialize
           # doing this to ensure we also get private methods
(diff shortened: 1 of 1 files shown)


headius:
Looks good. Better than my suggestion to just set everything public. This might also prevent future bugs from replaced methods being too visible.

headius:
@jayjlawrence Maybe you can verify that bundler works properly with this patch?

indirect:
this seems like a good solution to me 👍
@homu r+

jayjlawrence:
@headius confirmed branch succeeds with 1.7.24 and 1.7.26.

segiddins:
@homu r=indirect

PR #4990, opened 2016-09-15
Dynamically fetch expected rails version

segiddins:
This way, we don't have to update this expectation every time a new rails version comes out


segiddins:
@homu r+

PR #4701, opened 2016-06-22
Unset GEM_PATH with nil not empty string.

chrismo:
This should fix #4592, the tests all pass, but the line of code in
question goes back to 2010, so this sorta seems slightly dangerous, but
it's probable the circumstances of hitting this line in conjunction with
`bundle exec` is a combination that didn't exist prior to 1.12.x.

Issue #4592 has a full diagnosis, but the gist of it is this: if an
empty string is passed as the `GEM_PATH` to the subsequent process
launched by `bundle exec`, then if the `cmd` portion of `bundle exec` is
a ruby shebanged file, then if the current bundle install uses a local
path (`disable_shared_gems` is true) then it won't be able to find the
bundler gem at all because Bundler doesn't install itself into its own
Bundle, it's only installed in the system gems for the Ruby.

`nil` must be passed because the RubyGems code that sets up the
`GEM_PATH` does a conditional on the current `GEM_PATH` and empty string
evaluates to true, whereas `nil` evaluates to false. In the false case
the `GEM_PATH` is internally populated with the system gems path such
that the bundler gem can be found.

Issue #4602, opened 2016-05-23
Nested bundle exec support is broken on latest rubygems/bundler combo


SamSaffron:
@indirect asked me to open a new ticket on this: https://github.com/bundler/bundler/issues/4381#issuecomment-220893495

Bundler version: 1.12.4
Rubygems version: 2.6.4

```
docker exec -it --rm discourse/discourse:1.3.1 /bin/bash
cd /var/www/discourse
discourse@ubuntu-standalone:/var/www/discourse$ bundler --version
Bundler version 1.12.4
discourse@ubuntu-standalone:/var/www/discourse$ gem --version
2.6.4
discourse@ubuntu-standalone:/var/www/discourse$ ruby --version
ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-linux]
discourse@ubuntu-standalone:/var/www/discourse$ bundle exec bundle exec ls
bundler: failed to load command: bundle (/usr/local/bin/bundle)
Gem::Exception: can't find executable bundle
  /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.12.4/lib/bundler/rubygems_integration.rb:373:in `block in replace_bin_path'
  /usr/local/lib/ruby/site_ruby/2.3.0/rubygems.rb:278:in `activate_bin_path'
  /usr/local/bin/bundle:22:in `<top (required)>'
```

SamSaffron:
Confirmed issue, exact bit by bit same image with rubygems `2.5.1` all is working well and as expected

segiddins:
I'm unable to reproduce this locally, `bundle exec bundle exec ls` works in all of my projects

indirect:
@segiddins sounds like this might be linux-specific? All the examples I have seen are on Ubuntu.

indirect:
@SamSaffron I need better repro instructions, the commands you gave don't work for me:

```
$ docker exec -it --rm discourse/discourse:1.3.1 /bin/bash
flag provided but not defined: --rm
See 'docker exec --help'.
```

segiddins:
I don't have a linux machine, so someone else will have to debug then ¯\_(ツ)_/¯

indirect:
I'm perfectly happy to debug in a VM if I can figure out how to repro :P

SamSaffron:
@indirect oops was typing that from memory ... you want 

```
docker run -it --rm discourse/discourse:1.3.1 /bin/bash

% cd /var/www/discourse
[...]

Issue #4381, opened 2016-03-18
rubygems 2.6.2 breaks nested bundle exec support


taoza:
The issue can be reproduced with the latest [ruby:2.3.0-slim docker image](https://github.com/docker-library/ruby/blob/c88f3a67da720bfa9fb1717960d90fd5db11c757/2.3/slim/Dockerfile), which has a newly upgrade rubygems 2.6.2.

### Reproduce steps:
```shell
$  docker pull ruby:2.3.0-slim
2.3.0-slim: Pulling from ruby
d8bd0657b25f: Already exists
a582cd499e0f: Already exists
ae7c36464f9a: Already exists
9e853bd088c3: Already exists
29ba3bda4368: Already exists
8ddd74aaef1d: Already exists
2d6db04da3e0: Already exists
cf824e9b9462: Already exists
69f3ed685b86: Already exists
3396732706b8: Already exists
f0f8bb86227b: Already exists
0137e070fc88: Already exists
a1dea42157c8: Already exists
ca04fb901096: Already exists
b86b27ca1964: Already exists
41bb4e536f92: Already exists
Digest: sha256:927b3ffe7c47f830c2f955e28bb02e3aaca6cbecd71e534dc4fda02156eec589
Status: Image is up to date for ruby:2.3.0-slim
$  docker run -it ruby:2.3.0-slim bash
root@9a7a388a97ff:/# touch Gemfile
root@9a7a388a97ff:/# bundle install
The Gemfile specifies no dependencies
Resolving dependencies...
Bundle complete! 0 Gemfile dependencies, 1 gem now installed.
Bundled gems are installed into .usr/local/bundle.
root@9a7a388a97ff:/# bundle exec bundle -v
/usr/local/bin/bundle:22:in `load': cannot load such file -- /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.11.2/lib/gems/bundler-1.11.2/bin/bundle (LoadError)
        from /usr/local/bin/bundle:22:in `<main>'
```

### Possible cause:
A further look into bundler's binstub:
```shell
root@9a7a388a97ff:/# cat `which bundle`
#!/usr/local/bin/ruby
#
# This file was generated by RubyGems.
#
# The application 'bundler' is installed as part of a gem, and
# this file is here to facilitate running it.
#

require 'rubygems'

version = ">= 0.a"

if ARGV.first
  str = ARGV.first
  str = str.dup.force_encoding("BINARY") if str.respond_to? :force_encoding
[...]

PR #4992, opened 2016-09-15
Fix GEM_PATH regression in 1.13.

### LINKED ISSUE #4592: Could not find 'bundler'
Software
  * osx 10.11.4
  * rbenv 1.0.0
  * ruby 2.2.3p173 (2015-08-18 revision 51636) [x86_64-darwin15]
  * Rails 4.2.6

bundle exec gem list|grep guard

    guard (2.13.0)
    guard-compat (1.2.1)
    guard-rspec (4.6.5)

bundle exec guard -d

    RubyDep: WARNING: Your Ruby has security vulnerabilities! (To disable warnings, set RUBY_DEP_GEM_SILENCE_WARNINGS=1)
    RubyDep: WARNING: Your Ruby is: 2.2.3 (insecure). Recommendation: install 2.2.5 or 2.3.1. (Or, at least to 2.2.4 or 2.3.0)
    00:54:56 - DEBUG - Notiffany: gntp not available (Please add "gem 'ruby_gntp'" to your Gemfile and run your app with "bundle exec".).
    00:54:56 - DEBUG - Notiffany: growl not available (Please add "gem 'growl'" to your Gemfile and run your app with "bundle exec".).
    00:54:56 - DEBUG - Notiffany: terminal_notifier not available (Please add "gem 'terminal-notifier-guard'" to your Gemfile and run your app with "bundle exec".).
    00:54:56 - DEBUG - Notiffany: libnotify not available (Unsupported platform "darwin15.4.0").
    00:54:56 - DEBUG - Notiffany: notifysend not available (Unsupported platform "darwin15.4.0").
    00:54:56 - DEBUG - Notiffany: notifu not available (Unsupported platform "darwin15.4.0").
    00:54:56 - DEBUG - Command execution: {"ALTERNATE_EDITOR"=>"false"} emacsclient --eval '1'
    00:54:56 - DEBUG - Notiffany: emacs not available (Emacs client failed).
    00:54:56 - DEBUG - Notiffany: tmux not available (:tmux notifier is only available inside a TMux session.).
    00:54:56 - DEBUG - Notiffany: file not available (No :path option given).
    00:54:56 - DEBUG - Notiffany is using TerminalTitle to send notifications.
    00:54:56 - DEBUG - Command execution: hash stty
    00:54:56 - DEBUG - Guard starts all plugins
    00:54:56 - DEBUG - Hook :start_begin executed for Guard::RSpec
    00:54:56 - INFO - Guard::RSpec is running
    00:54:56 - DEBUG - Hook :start_end executed for Guard::RSpec
    00:54:56 - INFO - Guard is now watching at '/Users/lemon/Documents/chavis'
    00:54:56 - DEBUG - Start interactor
    00:54:56 - DEBUG - Command execution: stty -g 2>/dev/null
    [1] guard(main)>
    00:54:58 - DEBUG - Interactor was stopped or killed
    00:54:58 - DEBUG - Command execution: stty  2>/dev/null
    00:54:58 - INFO - Run all
    00:54:58 - DEBUG - Hook :run_all_begin executed for Guard::RSpec
    00:54:58 - INFO - Running all specs
    /Users/lemon/.rbenv/versions/2.2.3/lib/ruby/2.2.0/rubygems/dependency.rb:315:in `to_specs': Could not find 'bundler' (>= 0) among 98 total gem(s) (Gem::LoadError)
    Checked in 'GEM_PATH=/Users/lemon/Documents/chavis/vendor/bundle/ruby/2.2.0', execute `gem env` for more information
    	from /Users/lemon/.rbenv/versions/2.2.3/lib/ruby/2.2.0/rubygems/dependency.rb:324:in `to_spec'
    	from /Users/lemon/.rbenv/versions/2.2.3/lib/ruby/2.2.0/rubygems/core_ext/kernel_gem.rb:64:in `gem'
    	from /Users/lemon/.rbenv/versions/2.2.3/bin/bundle:22:in `<main>'
    00:54:58 - ERROR - Failed: "bundle exec rspec -f progress -r /Users/lemon/Documents/chavis/vendor/bundle/ruby/2.2.0/gems/guard-rspec-4.6.5/lib/guard/rspec_formatter.rb -f Guard::RSpecFormatter --failure-exit-code 2  spec" (exit code: 1)

    00:54:58 - DEBUG - Start interactor
    00:54:58 - DEBUG - Command execution: stty -g 2>/dev/null
gem env

    RubyGems Environment:
      - RUBYGEMS VERSION: 2.4.5.1
      - RUBY VERSION: 2.2.3 (2015-08-18 patchlevel 173) [x86_64-darwin15]
      - INSTALLATION DIRECTORY: /Users/lemon/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0
      - RUBY EXECUTABLE: /Users/lemon/.rbenv/versions/2.2.3/bin/ruby
      - EXECUTABLE DIRECTORY: /Users/lemon/.rbenv/versions/2.2.3/bin
      - SPEC CACHE DIRECTORY: /Users/lemon/.gem/specs
      - SYSTEM CONFIGURATION DIRECTORY: /Users/lemon/.rbenv/versions/2.2.3/etc
      - RUBYGEMS PLATFORMS:
        - ruby
        - x86_64-darwin-15
      - GEM PATHS:
         - /Users/lemon/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0
         - /Users/lemon/.gem/ruby/2.2.0
      - GEM CONFIGURATION:
         - :update_sources => true
         - :verbose => true
         - :backtrace => false
         - :bulk_threshold => 1000
         - :sources => ["https://gems.ruby-china.org/"]
      - REMOTE SOURCES:
         - https://gems.ruby-china.org/
      - SHELL PATH:
         - /Users/lemon/.rbenv/versions/2.2.3/bin
         - /usr/local/Cellar/rbenv/1.0.0/libexec
         - /Users/lemon/.rbenv/shims
         - /usr/local/bin
         - /usr/bin
         - /bin
         - /usr/sbin
         - /sbin

https://github.com/guard/guard-rspec/issues/369

### LINKED ISSUE #4974: Regression in disable_shared_gems behavior
Between the versions `v1.13.0.pre.1` and `v1.13.0.rc1` the behavior of `bundle install --deployment` has changed concerning the `disable_shared_gems` flag. Gems that are present in system rubygems are no longer installed into the local gems folder (`vendor/bundle`).

This behavior seems to be caused by change in `Bundler#configure_gem_path` of

    ENV["GEM_PATH"] = ""

to

    env["GEM_PATH"] = nil

introduced in commit 9a8ab67da6567c767307e8861893761542483ba0

I tested under 2.4.5.1 rubygems version.

I am sorry If that new behavior is an intended one and not a regression.

chrismo:
_nth time is a charm ... here's the latest latest latest commit message_:

Fix disable_shared_gems bug & keep nested exec
tl;dr `--deployment` will copy _all_ gems to the `--path` now and a
proper fix is in place for nested bundle execs when `--path` is set and
RubyGems >=2.6.2 is being used.

Fixes #4974.

There's two problems blended in here. Let's trace the events here from
the beginning, shall we?

First off, from probably the first versions of Bundler, when
`disable_shared_gems` is true, the `GEM_PATH` gets internally overridden
and initialized to an empty string. This is done to make sure that later
in the `Bundler.setup` process it's expanded to ONLY the Bundler `--path`
setting, otherwise it expands to include the system path.

In 1.12, `bundle exec` was changed to use `Kernel.load` in some cases,
and that new code path calls `require "bundler/setup"`.

Later, issue #4592 was filed, showing that in cases like `--deployment`
where `disable_shared_gems` is true, Bundler couldn't find itself,
because Bundler never lives in the `--path` but only in system gems. And
as it would later be discovered, was not a problem with RubyGems 2.6.1,
but was a problem with >=2.6.2 (and older RubyGems it would seem, though
those weren't as thoroughly investigated).

We fixed #4592 (see PR #4701) in 1.13.0 by changing the oooold code
initializing `GEM_PATH` to be initialized to `nil` instead of empty
string in all `disable_shared_gems` cases. But it created another bug,
filed as #4974, wherein system gems were no longer copied to the
`--path` when `disable_shared_gems` is true. In this commit here (#4992)
we've reverted the change so that `GEM_PATH` is now back to being
initialized to an empty string instead of `nil`.

That fixes #4974, but regresses #4592, and we needed a new way to fix
it.

After a few tortured attempts at this, I ran across issue #4602, a
similar report of nested bundle execs being broken, #4602 itself an
extension of #4381 reporting the same problem. It brought to light the
role the Rubygems version played in the problem.

When the `bundler` gem is installed and the wrapper is generated for any
gem executables, the contents of this wrapper are determined by the
Rubygems version. Up until RubyGems 2.6.1 the last line of this wrapper
calls `Gem.bin_path`. Bundler replaces the Rubygems implementation of
`Gem.bin_path` with its own, and has for a long time made a special
exception for the Bundler gem itself, short-circuiting with the contents
of a special ENV variable called `BUNDLE_BIN_PATH`.

In Rubygems 2.6.2, `bin_path` was superseded by a new
`Gem.activate_bin_path` method which did what `bin_path` did but also
activated the gem. Bundler 1.13 added support for this, but it didn't
include the same short-circuit for bundler itself. (Alert user @taoza
even noticed this here
fcaab35#r56665282).

This commit also includes that short circuit for Rubygems >=2.6.2 now
and nested bundle exec should continue to work.

### DIFF
--- a/spec/commands/exec_spec.rb
+++ b/spec/commands/exec_spec.rb
@@ -2,8 +2,9 @@
 require "spec_helper"
 
 describe "bundle exec" do
+  let(:system_gems_to_install) { %w(rack-1.0.0 rack-0.9.1) }
   before :each do
-    system_gems "rack-1.0.0", "rack-0.9.1"
+    system_gems(system_gems_to_install)
   end
 
   it "activates the correct gem" do
@@ -625,6 +626,28 @@ __FILE__: #{path.to_s.inspect}
 
         expect(out).to eq("foo")
       end
+    end
+  end
+
+  context "nested bundle exec" do
+    let(:system_gems_to_install) { super() << :bundler }
+    before do
+      gemfile <<-G
+        source "file://#{gem_repo1}"
+        gem "rack"
+      G
+      bundle :install, :system_bundler => true, :path => "vendor/bundler"
+    end
+
+    it "overrides disable_shared_gems so bundler can be found" do
+      file = bundled_app("file_that_bundle_execs.rb")
+      create_file(file, <<-RB)
+        #!#{Gem.ruby}
+        puts `bundle exec echo foo`
+      RB
+      file.chmod(0o777)
+      bundle! "exec #{file}", :system_bundler => true
+      expect(out).to eq("foo")
     end
   end
 end
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -443,7 +443,10 @@ EOF
     def configure_gem_path(env = ENV, settings = self.settings)
       blank_home = env["GEM_HOME"].nil? || env["GEM_HOME"].empty?
       if settings[:disable_shared_gems]
-        env["GEM_PATH"] = nil
+        # this needs to be empty string to cause
+        # PathSupport.split_gem_path to only load up the
+        # Bundler --path setting as the GEM_PATH.
+        env["GEM_PATH"] = ""
       elsif blank_home || Bundler.rubygems.gem_dir != bundle_path.to_s
         possibles = [Bundler.rubygems.gem_dir, Bundler.rubygems.gem_path]
         paths = possibles.flatten.compact.uniq.reject(&:empty?)
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -397,6 +397,17 @@ module Bundler
         spec
       end
 
+      redefine_method(gem_class, :activate_bin_path) do |name, *args|
+        exec_name = args.first
+        return ENV["BUNDLE_BIN_PATH"] if exec_name == "bundle"
+
+        # Copy of Rubygems activate_bin_path impl
+        requirement = args.last
+        spec = find_spec_for_exe name, exec_name, [requirement]
+        Gem::LOADED_SPECS_MUTEX.synchronize { spec.activate }
+        spec.bin_file exec_name
+      end
+
       redefine_method(gem_class, :bin_path) do |name, *args|
         exec_name = args.first
         return ENV["BUNDLE_BIN_PATH"] if exec_name == "bundle"
--- a/spec/bundler/bundler_spec.rb
+++ b/spec/bundler/bundler_spec.rb
@@ -143,12 +143,12 @@ describe Bundler do
 
   describe "configuration" do
     context "disable_shared_gems" do
-      it "should unset GEM_PATH with nil" do
+      it "should unset GEM_PATH with empty string" do
         env = {}
         settings = { :disable_shared_gems => true }
         Bundler.send(:configure_gem_path, env, settings)
         expect(env.keys).to include("GEM_PATH")
-        expect(env["GEM_PATH"]).to be_nil
+        expect(env["GEM_PATH"]).to eq ""
       end
     end
   end
--- a/spec/support/helpers.rb
+++ b/spec/support/helpers.rb
@@ -88,18 +88,28 @@ module Spec
 
       bundle_bin = options.delete("bundle_bin") || File.expand_path("../../../exe/bundle", __FILE__)
 
+      if system_bundler = options.delete(:system_bundler)
+        bundle_bin = "-S bundle"
+      end
+
       requires = options.delete(:requires) || []
       requires << File.expand_path("../fakeweb/" + options.delete(:fakeweb) + ".rb", __FILE__) if options.key?(:fakeweb)
       requires << File.expand_path("../artifice/" + options.delete(:artifice) + ".rb", __FILE__) if options.key?(:artifice)
       requires << "support/hax"
       requires_str = requires.map {|r| "-r#{r}" }.join(" ")
 
+      load_path = []
+      load_path << lib unless system_bundler
+      load_path << spec
+      load_path_str = "-I#{load_path.join(File::PATH_SEPARATOR)}"
+
       env = (options.delete(:env) || {}).map {|k, v| "#{k}='#{v}'" }.join(" ")
+      env["PATH"].gsub!("#{Path.root}/exe", "") if env["PATH"] && system_bundler
       args = options.map do |k, v|
         v == true ? " --#{k}" : " --#{k} #{v}" if v
       end.join
 
-      cmd = "#{env} #{sudo} #{Gem.ruby} -I#{lib}:#{spec} #{requires_str} #{bundle_bin} #{cmd}#{args}"
+      cmd = "#{env} #{sudo} #{Gem.ruby} #{load_path_str} #{requires_str} #{bundle_bin} #{cmd}#{args}"
       sys_exec(cmd) {|i, o, thr| yield i, o, thr if block_given? }
     end
     bang :bundle
@@ -242,11 +252,17 @@ module Spec
 
     def install_gems(*gems)
       gems.each do |g|
-        path = "#{gem_repo1}/gems/#{g}.gem"
+        path = if g == :bundler
+          Dir.chdir(root) { gem_command! :build, "#{root}/bundler.gemspec" }
+          bundler_path = root + "bundler-#{Bundler::VERSION}.gem"
+        else
+          "#{gem_repo1}/gems/#{g}.gem"
+        end
 
         raise "OMG `#{path}` does not exist!" unless File.exist?(path)
 
-        gem_command! :install, "--no-rdoc --no-ri --ignore-dependencies #{path}"
+        gem_command! :install, "--no-rdoc --no-ri --ignore-dependencies '#{path}'"
+        bundler_path && bundler_path.rmtree
       end
     end
 


lib/bundler/cli/exec.rb
```diff
@@ -81,6 +82,38 @@ def kernel_load(file, *args)
       abort "#{e.class}: #{e.message}\n  #{backtrace.join("\n  ")}"
     end
 
+    # This is pretty damned esoteric. Since the dawn of Bundler, when
```
segiddins: please remove the `damned`

lib/bundler/cli/exec.rb
```diff
@@ -81,6 +82,38 @@ def kernel_load(file, *args)
...
+    #
+    # We fixed this in 1.13.0 by removing GEM_PATH ever being set to
+    # empty string, and that fixed bundle exec and broke `--deployment`.
+    # So much for trusting a test suite. (shrug) I should know better.
+    #
+    # So, we're putting back the age-old empty string, and coping for
+    # the weirdness right here, with this method and its appropriate
+    # 27-to-1 ratio of doc lines to code lines.
+    def fixup_gem_path_env
+      ENV['GEM_PATH'] = nil if Bundler.settings[:disable_shared_gems]
```
segiddins: double quotes

lib/bundler/cli/exec.rb
chrismo: Yes, sorry - it was late at night.

lib/bundler/cli/exec.rb
chrismo: Yeah, I forgot to run rubocop, though I knew this wasn't done yet - no tests.

chrismo:
yeah, tests for the bundle exec, and tests for the disable_shared_gems putting _all_ gems into the --path, which said tests would have prevented this change in the first place.

indirect:
@chrismo awesomeeeeee thank you for tracking this down :bow: I'm very excited to get this fixed once we have tests. :)

chrismo:
due credit to @miros for isolating the problem commit in #4974 in the first place

chrismo:
The comments were a bit rambly and late-night stream of consciousness. Re-wrote them to try and be clearer.

Still need to research `bundle exec` tests -- there's actually plenty of coverage in spec/install/path_spec.rb, I don't know why CI continues to pass those specs when locally they clearly fail for me when the bug is in play.

chrismo:
This isn't working. I still can't get any spec within the bundler codebase itself to reliably recreate the problem, because it's a nested bundle exec that is the original problem ... and that's just a bit too bizarre for my brain to figure out. 

The spec_helper calls `Spec::Rubygems.setup` which sets up the `PATH` to be:

      ENV["PATH"] = ["#{Path.root}/exe", "#{Path.system_gem_path}/bin", ENV["PATH"]].join(File::PATH_SEPARATOR)

placing the local exe/bundle in the PATH, thwarting any attempt to recreate a nested bundle exec accurately which IRL needs to be hitting the Rubygems generated bin entry:

```
#!/usr/bin/env ruby
#
# This file was generated by RubyGems.
#
# The application 'bundler' is installed as part of a gem, and
#

require 'rubygems'
# ...snip...
load Gem.activate_bin_path('bundle', 'bundle', version)
```

So. I at least created [this bundler-case](https://github.com/chrismo/bundler-cases/blob/master/cases/issues/bundler-4592/bundler-4592.rb) based on the original #4592 bug report. This accurately shows the nested bundle exec problem in 1.12.5 and the fix in 1.13.1.

But the code HERE in this PR (#4992) - while it fixes #4974, it regresses the nested bundle exec fix (#4592).

The reason is also due to the nature of the 1.12.5 change to use Kernel.load (same process, I think) instead of Kernel.exec (forks a new process). There's so much global state (and a lot of it memoized) that's setup in `require "bundler/setup"` -- I can't find a way to create a workaround to fix up the pathing. It gets easily as complicated as the above situation just trying to recreate a failing spec. (Very happy to be wrong here, btw).


So so.

IMO, #4974 is a much worse problem than #4592. #4974 keeps `--deployment` from working properly (unless none of your bundled gems are installed as system gems). #4592 only breaks nested bundle exec only in a `--path` install situation. 

I'd rather release a 1.13.2 with #4974 fixed and intentionally regress #4592 until we can figure out how to properly support the nested `bundle exec` case.

chrismo:
proactive code review - I'm being a little sloppy on this PR, I'll rebase and remove the version change that snuck in and everything before for realz. Just pushing for comment on the substantive parts.

lib/bundler/cli/exec.rb
```diff
@@ -81,6 +82,34 @@ def kernel_load(file, *args)
...
+    #
+    # We fixed this (the bundle exec bug) in 1.13.0 by changing GEM_PATH to be
+    # initialized to nil instead of empty string in all cases. But it created
+    # another bug. We've reverted the change so that GEM_PATH is now back to
+    # being initialized, but still need to override how the GEM_PATH is set
+    # in this special case so that the bundler executable can be found.
+    def configure_gem_path_override
+      if Bundler.settings[:disable_shared_gems]
+        # TODO MODO hate this spooky at a distance global state voodoo
+        ENV["DO_NOT_LIMIT_GEM_PATH"] = "true"
```
segiddins: any reason this couldn't just check if `ENV["GEM_PATH"].nil?`

spec/commands/exec_spec.rb
```diff
@@ -591,4 +592,26 @@ def bin_path(a,b,c)
...
+    let(:system_gems_to_install) { super() << :bundler }
+    before do
+      gemfile <<-G
+        source "file://#{gem_repo1}"
+        gem "rack"
+      G
+      bundle :install, :system_bundler => true, :path => "vendor/bundler"
+    end
+
+    it "works with disable_shared_gems" do
```
segiddins: needs a more descriptive name

lib/bundler/cli/exec.rb
chrismo: this `configure_gem_path_override` method is now called _before_ `require "bundler/setup"` ... prior commit had me attempting to undo something that was done inside bundler/setup _after_ it was called ... had to change the approach since it was too late at that point.

chrismo:
Ok, I think this ready, presuming CI passes.

indirect:
@segiddins all your requested changes appear to be addressed in the final commit, even though they're still shown in the PR... is that normal? this seems sub-optimal. :/

chrismo:
dunno - prior versions of github, before the code review stuff, used to collapse comments on old commits. i'm not used to the new review flow ... maybe @segiddins needs to "sign off" on his comments in some way?

indirect:
yeah, I guess so. well, I'm happy with this, so @homu r+.

chrismo:
shoot. i also built a bundler-cases for this - and it's still failing. argh. looking ...

indirect:
heh. well, in that case... @homu r-

indirect:
@homu r=chrismo

@chrismo when you're ready, you can approve this for testing and merging with the comment `r+` directed at homu. once that's done, I'd like to release 1.13.2.

indirect:
...dammit, I can't remember the bot's syntax.

@homu r-
@homu delegate=chrismo

indirect:
@homu r-

chrismo:
bah, keep forgetting that one test I'd added that's still failing. fixing, new commit shortly

chrismo:
@homu retry

chrismo:
@homu r+

chrismo:
@homu retry

indirect:
ffffuuuu, sorry @chrismo

@homu retry

chrismo:
whew. and don't come back :)

PR #4994, opened 2016-09-16
[Definition] Avoid parsing lockfile twice on init

segiddins:
\c @chrismo


chrismo:
oh - yeah, good catch. thx

segiddins:
@homu r+

PR #4998, opened 2016-09-17
::Rake::CommandFailedError doesn't exist.

srbaker:


segiddins:
Thanks!

segiddins:
@homu r+

segiddins:
@homu retry

fgi:
I use Vlad 2.7.0 and just upgraded Bundler to 1.13.2.

This commit generates following error when bundler is loaded for a rake task:
```
$ rake -T
rake aborted!
NameError: uninitialized constant Vlad::CommandFailedError
/home/rails/.rvm/gems/ruby-2.3.1@my-app/gems/bundler-1.13.2/lib/bundler/deployment.rb:18:in `define_task'
```

`Vlad::CommandFailedError`has been removed from Vlad since v2.0.0. It does not exist anymore:
https://github.com/seattlerb/vlad/commit/548deb834e27683cde23897d22c9f0b4cd87cac0

But `Rake::CommandFailedError` does exist in a Vlad dependency gem from the same team called rake-remote_task: 
https://github.com/seattlerb/rake-remote_task

Bundler seems to work fine with Vlad without this change. Is there any particular raison for this commit to occur?
If not, would it be possible to revert it?

Thanks.

PR #5002, opened 2016-09-19
fix message parenthesis color

alepore:
Hello!
Using Bundler version 1.13.1 i noticed a small graphic inconsistency on parentheses color:

screen shot 2016-09-19 at 16 16 10

This commit resets the opening parenthesis color:

screen shot 2016-09-19 at 16 27 08

Is this the expected behaviour? The spec was only checking the closing parenthesis color.

p.s:
Imho green parentheses look better here.


segiddins:
👍

indirect:
@alepore I would prefer that both the open and close parens be green--can we do that instead? 😀  Thanks!

alepore:
👍 updated will green parens, looks better

indirect:
@homu r+

Issue #5001, opened 2016-09-19
Bundler will not find a gem after adding platform parameter to gem and then removing it


pavelbulanov:
Bundler v1.13.1

Steps to reproduce:
1) Add some gem to Gemfile, run "bundle install", gem with specific version added to Gemfile.lock
2) Add some platform (not matching your system) to that exact gem, e.g. adding "platform: :mswin". 
3) Run "bundle install" again. Gemfile.lock is updated, there is no longer gem name with version under GEM section, but it's still in DEPENDENCIES list (with no version as expected).
4) Remove platform parameter from the gem (assume it was added by mistake), run "bundle install". 

Result:
Gem is not restored - Gemfile.lock remains as is, with no gem name under GEM section, and with gem name still present under Dependencies. Gem is not factually added in the application with Bundle.require. Re-running bundle install multiple times or manually installing / removing gem doesn't help.

Interesting, running 'bundle show <gem_name>' shows the following (mini_racer is my gem):
"Could not find gem 'mini_racer'.
Did you mean mini_racer?"
Spelling is exactly the same for both occurrences.

Expected Result:
'bundle install" will "re-enable" the gem after platform condition was removed. 

Workaround - if I just remove the gem name from Gemfile completely, run bundle install - Gemfile.lock is updated and gem name is removed from Dependencies list. Re-adding gem in Gemfile again will turn it on.

segiddins:
So after digging into this, the issue appears to be the fact that the platform used for conditional installation isn't in the lockfile listed under `PLATFORMS`

pavelbulanov:
Inside Gemfile.lock, PLATFORMS only lists ruby, that's correct - after running "bundle install" or "bundle show gem_name" with the gem platform set to :mswin (tested with :jruby with same result, just in case)

PR #5003, opened 2016-09-19
[Definition] Print a helpful warning when a dependency is unused on a…

segiddins:
…ny platform

Should help address the confusion caused by https://github.com/bundler/bundler/issues/5001

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -808,8 +808,16 @@ module Bundler
       deps = []
       dependencies.each do |dep|
         dep = Dependency.new(dep, ">= 0") unless dep.respond_to?(:name)
-        next unless remote || dep.current_platform?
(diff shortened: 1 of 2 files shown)


indirect:
This is a good warning! That said, I'm not sure how this fixes the bug described in #5001. In that ticket, the Gemfile no longer contains a `platform` declaration at all, and the gem is still ignored. Seems like a different problem.

segiddins:
That's not fixable without completely re-writing large parts of definition.

indirect:
Assuming that is true, it still doesn't change the fact that this is solving a different problem than the one reported in #5001, does it?

segiddins:
No, but it might help address the confusion by printing a warning in that case (saying that something out of the ordinary is taking place)

indirect:
@homu r+

PR #4981, opened 2016-09-14
Fixed missing constant with rubygems integration

hsbt:
Fixes #5000.

I got uninitialized constant error with bundler-1.13.1 on travis. see https://travis-ci.org/ruby/rake/jobs/159734476

It's reproduce my local box with ruby/rake testing. I simply fixed this error.

Thank you.

PR #5010, opened 2016-09-21
[workaround] Use `require` instead of `autoload` for bundler/plugin/api

aycabta:
Please read https://github.com/bundler/bundler/pull/4981#issuecomment-248674155.

> But it's elusive reproduction (and lack of test case) makes me think that the problem is a bit deeper and may recur.

I think so. This Pull Request is just *workaround*.


indirect:
Merging this so that we at least have a fix for now. Hopefully #4981 will eventually produce a test case.

@homu r+

PR #5013, opened 2016-09-22
[Definition] Implicitly lock to running ruby version when there is no…

segiddins:
…ne in the gemfile

@indirect I think this does what you asked for in https://github.com/bundler/bundler/issues/4853?

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -100,7 +100,7 @@ module Bundler
       @unlock[:ruby] ||= if @ruby_version && locked_ruby_version_object
         @ruby_version.diff(locked_ruby_version_object)
       end
-      @unlocking ||= @unlock[:ruby] ||= (!@locked_ruby_version ^ !@ruby_version)
+      @unlocking ||= @unlock[:ruby]
 
       add_current_platform unless Bundler.settings[:frozen]
 
@@ -330,7 +330,6 @@ module Bundler
     end
 
     def locked_ruby_version
-      return unless ruby_version
       if @unlock[:ruby] || !@locked_ruby_version
         Bundler::RubyVersion.system
       else
@@ -386,10 +385,8 @@ module Bundler
         handled << dep.name
       end
 
-      if locked_ruby_version
-        out << "\nRUBY VERSION\n"
-        out << "   #{locked_ruby_version}\n"
-      end
+      out << "\nRUBY VERSION\n"
+      out << "   #{locked_ruby_version}\n"
 
       # Record the version of Bundler that was used to create the lockfile
       out << "\nBUNDLED WITH\n"
@@ -472,28 +469,34 @@ module Bundler
     end
 
     def validate_ruby!
-      return unless ruby_version
+      return unless ruby_version = self.ruby_version || locked_ruby_version_object
 
-      if diff = ruby_version.diff(Bundler::RubyVersion.system)
-        problem, expected, actual = diff
+      return unless diff = ruby_version.diff(Bundler::RubyVersion.system)
+      problem, expected, actual = diff
 
-        msg = case problem
-              when :engine
-                "Your Ruby engine is #{actual}, but your Gemfile specified #{expected}"
-              when :version
-                "Your Ruby version is #{actual}, but your Gemfile specified #{expected}"
-              when :engine_version
-                "Your #{Bundler::RubyVersion.system.engine} version is #{actual}, but your Gemfile specified #{ruby_version.engine} #{expected}"
-              when :patchlevel
-                if !expected.is_a?(String)
-                  "The Ruby patchlevel in your Gemfile must be a string"
-                else
-                  "Your Ruby patchlevel is #{actual}, but your Gemfile specified #{expected}"
-                end
-        end
+      implicit_ruby_version = self.ruby_version.nil?
+      specified = implicit_ruby_version ? "bundled is locked to" : "Gemfile specified"
 
-        raise RubyVersionMismatch, msg
+      msg = case problem
+            when :engine
+              "Your Ruby engine is #{actual}, but your #{specified} #{expected}"
+            when :version
+              return if implicit_ruby_version && ruby_version.to_gem_version_with_patchlevel < Bundler::RubyVersion.system.to_gem_version_with_patchlevel
+              "Your Ruby version is #{actual}, but your #{specified} #{expected}"
+            when :engine_version
+              "Your #{Bundler::RubyVersion.system.engine} version is #{actual}, but your #{specified} #{ruby_version.engine} #{expected}"
+            when :patchlevel
+              if !expected.is_a?(String)
+                "The Ruby patchlevel in your Gemfile must be a string"
+              else
+                "Your Ruby patchlevel is #{actual}, but your #{specified} #{expected}"
+              end
       end
+      msg += ". This may cause issues if there are any gems that were depending on that ruby version." if implicit_ruby_version
+
+      raise RubyVersionMismatch, msg if !implicit_ruby_version || Bundler.feature_flag.bundler_2_mode?
+
+      Bundler.ui.warn(msg)
     end
 
     # TODO: refactor this so that `match_platform` can be called with two platforms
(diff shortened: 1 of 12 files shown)


indirect:
@segiddins finally waded through everything else, phew. Here's a (not guaranteed complete) spec for this feature:

- [x] auto lock to running ruby version if ruby not in gemfile
- [x] if ruby isn't in the Gemfile, warn when running ruby version doesn't match locked ruby
- [ ] config to upgrade said warning to exception which will be default in 2.0
- [ ] use auto-locked version of Ruby to resolve respecting ruby version (repro below)

```
$ bundle init
$ echo "gem 'mime-types-data'" >> Gemfile
$ chruby 1.9.3
$ bundle install
Gem::InstallError: mime-types-data requires Ruby version >= 2.0.
An error occurred while installing mime-types-data (3.2015.1120), and Bundler cannot continue.
Make sure that `gem install mime-types-data -v '3.2015.1120'` succeeds before bundling.
```

segiddins:
@indirect that's because http://index.rubygems.org/info/mime-types-data is missing the required_ruby_version for 3.2015.1120

lib/bundler/definition.rb
```diff
@@ -472,28 +469,34 @@ def validate_runtime!
...
+            when :patchlevel
+              if !expected.is_a?(String)
+                "The Ruby patchlevel in your Gemfile must be a string"
+              else
+                "Your Ruby patchlevel is #{actual}, but your #{specified} #{expected}"
+              end
+      end
+      msg += ". This may cause issues if there are any gems that were depending on that ruby version." if implicit_ruby_version
+
+      raise RubyVersionMismatch, msg if !implicit_ruby_version || Bundler.feature_flag.bundler_2_mode?
```
segiddins: what would you like this config option to be called, @indirect ?

indirect:
@segiddins fuuuuuuuuu arghhhhhh okay. At least that will be fixed as soon as we switch the new index over to the Rails app:

```
$ curl -k  -H "Host: rubygems.org" https://54.186.104.15/info/mime-types-data
---
3.2015.1120 |checksum:2cc2d944557aacac49d15709247be3fba2028f65762bf474cea0212e80dfdf11,ruby:>= 2.0
3.2016.0221 |checksum:f6ce9f679d796f4d2647a93ca3d2c3bca4399924a744cdc3f5063a6b8195cf17,ruby:>= 2.0
3.2016.0521 |checksum:7542bccff06d70c4ad94d1cf10b7dac6bd89958356e5d0d7f6447168c819be12,ruby:>= 2.0
```

indirect:
Okay, found a use-case that I think doesn't work? Or I'm not understanding it. Lock to gems that work with Ruby 2.3, switch to Ruby 2.1, try to install. Fails with good error. Try to unlock and get working versions fails (this is surprising. why not?) Also, trying to explicitly update the lock to the current Ruby also fails, when it seems like it should not.

```
$ bundle init
$ echo "gem 'listen'" >> Gemfile
$ chruby 2.3
$ bundle install
$ chruby 2.1

$ dbundle install
Your Ruby version is 2.1.9, but your bundled is locked to 2.3.1. This may cause issues if there are any gems that were depending on that ruby version.
ruby_dep-1.5.0 requires ruby version >= 2.2.5, ~> 2.2, which is incompatible with the current version, ruby 2.1.9p490

$ dbundle update
Your Ruby version is 2.1.9, but your bundled is locked to 2.3.1. This may cause issues if there are any gems that were depending on that ruby version.
Fetching dependency metadata from https://rubygems.org/.
Resolving dependencies...
ruby_dep-1.5.0 requires ruby version >= 2.2.5, ~> 2.2, which is incompatible with the current version, ruby 2.1.9p490

$ dbundle update --ruby
Your Ruby version is 2.1.9, but your bundled is locked to 2.3.1. This may cause issues if there are any gems that were depending on that ruby version.
Fetching dependency metadata from https://rubygems.org/.
Resolving dependencies...
ruby_dep-1.5.0 requires ruby version >= 2.2.5, ~> 2.2, which is incompatible with the current version, ruby 2.1.9p490
```

I see three action items from this:

- [ ] the warning about mismatched ruby versions should explain how to update the ruby version
- [ ] `bundle update` should either succeed or explicitly tell the user that they need to run `update --ruby`
- [ ] `update --ruby` should definitely succeed

Vaguely related, where are the nice resolver-style ruby version conflict resolution messages? Didn't we merge that before this?

segiddins:
We did, I'm guessing this is erroring out somewhere in setup, and that it's skipping the resolver entirely (because the lockfile exists). I'll definitely try and dig in, thanks for the repro steps!

lib/bundler/definition.rb
indirect: Oh man, I completely misunderstood this question, sorry. How about `ruby.strict_lock`?

segiddins:
The `bundle update` succeeded for me, but `bundle update --ruby` fails. However, I'm not sure how to fix this. `bundle update --ruby` currently works by not unlocking any specs, by design -- this way, you can update _only_ the locked ruby. That does mean, however, there's no way to unlock both all specs and the ruby version, which is what is needed in the latter case

segiddins:
Nevermind, found a bug

segiddins:
Better error messages & warnings in place

indirect:
@segiddins okay, I think these are the remaining things:

- [ ] travis build is failing

- [ ] there is no config option to upgrade version mismatch warning to a raise

- [ ] warning is not printed during locked/current ruby mismatch

<details><summary>repro case</summary>
```
$ bundle init
$ echo "gem 'rack'" >> Gemfile
$ chruby 1.9
$ bundle install
$ chruby 2.3

$ bundle install --path .bundle
Using bundler 1.13.5
Installing rack 1.6.4
Bundled gems are installed into ./.bundle.

$ cat Gemfile.lock
GEM
  remote: https://rubygems.org/
  specs:
    rack (1.6.4)

PLATFORMS
  ruby

DEPENDENCIES
  rack

RUBY VERSION
   ruby 1.9.3p551

BUNDLED WITH
   1.13.5
```
</details>

- [ ] `bundle update` from the previous example does not succeed

<details><summary>repro case</summary>
```
$ bundle init
$ echo "gem 'listen'" >> Gemfile
$ chruby 2.3
$ bundle install
$ chruby 2.1

$ bundle install
Your Ruby version is 2.1.10, but your bundled is locked to 2.3.1. This may cause issues if there are any gems that were depending on that ruby version.
To update the locked ruby version, run `bundle update --ruby`.If that fails, run `bundle update` followed by `bundle update --ruby`.
Fetching dependency metadata from https://rubygems.org/.
ruby_dep-1.5.0 requires ruby version >= 2.2.5, which is incompatible with the current
version, ruby 2.1.10p492

$ bundle update
Your Ruby version is 2.1.10, but your bundled is locked to 2.3.1. This may cause issues if there are any gems that were depending on that ruby version.
To update the locked ruby version, run `bundle update --ruby`.If that fails, run `bundle update` followed by `bundle update --ruby`.
Fetching dependency metadata from https://rubygems.org/.
Could not find gem 'ruby (= 2.1.10.492, = 2.3.1.112)' in any of the gem sources listed
in your Gemfile or available on this machine.
```
</details>

- [ ] `bundle update` is broken in general

<details><summary>repro case</summary>
```
$ bundle init
$ echo "gem 'rack'" >> Gemfile
$ chruby 1.9
$ bundle install
$ chruby 2.3

$ bundle update
Could not find gem 'ruby (= 1.9.3.551, = 2.3.1.112)' in any of the gem sources listed in
your Gemfile or available on this machine.
```
</details>

Let me know if you notice anything else that doesn't seem to be working correctly 👍

segiddins:
I _think_ those two broken cases are the ones that can't be fixed until we can change the lockfile?

indirect:
@segiddins in your previous comment, you said the second repro case worked for you, so I think that  is a regression.

Both of the last two repros demonstrate a problem with the error messages: it is not an explanatory or useful error message to say `Could not find gem 'ruby (= 2.1.10.492, = 2.3.1.112)'`.

The last example is a set of gems that _should_ work. The version of rack that works on Ruby 1.9 also works on Ruby 2.3. It should be possible to upgrade the ruby version without unlocking any gems, but instead `bundle update` barfs.

segiddins:
`bundle update` doesn't unlock the ruby version, which we discussed not being able to change until 2.0

PR #5015, opened 2016-09-23
fixing NoMethodError on settings

m1k3:
When the Settings object is initialized with no root directory it cannot read the local_config. This used to not be a problem due to the fact that the `#load_config` method did not try to exit early. Since now it does it returns nil when the config is not present setting the @local_config instance variable to nil. The fix is to return an empty hash the same way the `#load_config` method would return if it encountered a problem later on.

The attached test proves that there is a problem and the fix makes the problem go away.

### DIFF
--- a/lib/bundler/settings.rb
+++ b/lib/bundler/settings.rb
@@ -296,7 +296,7 @@ module Bundler
     }xo
 
     def load_config(config_file)
-      return unless config_file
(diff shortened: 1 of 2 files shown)


segiddins:
Thanks! 👍

segiddins:
@homu r+

Issue #5030, opened 2016-09-29
bin/console works wrong in Mac finder


rikkimongoose:
Bundler: 1.12.5
MacOS: 10.11.6
Finder: 10.11.4

Autogenerated **bin/console** file in MacOS fails to execute if browse to it with MacOS Finder.

It fails with following error:

_/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/rubygems/core_ext/kernel_require.rb:55:in `require': cannot load such file --_

The reason is Finder executes all scripts in context of ~ directory. So, Dir.pwd contain "Users/username/". But, for example, File.dirname(**FILE**) stores right path.

My quick fix is adding the following line at **bin/console** file:

``` ruby
$LOAD_PATH.unshift File.expand_path('../../lib', __FILE__) # add this line for MacOS support

require "bundler/setup"
```

I guess, it should be fixed at https://github.com/bundler/bundler/blob/master/lib/bundler/templates/newgem/bin/console.tt

segiddins:
A PR making that change would be perfect, thanks!

colby-swandale:
This has not been active for a long time. If this is an issue that anyone would like to fix, then please open a PR. But i'm going to close this ticket.

PR #5032, opened 2016-09-29
* fix execution with another active directory

rikkimongoose:
Fix for the following issue: https://github.com/bundler/bundler/issues/5030

MacOS Finder (and I sure, that some other shells well) run the script setting ~ as his active directory. Running **bin/console** or **bin/setup** from it, you'll receive the following error, because they are unable to find /lib/:

_/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/rubygems/core_ext/kernel_require.rb:55:in `require': cannot load such file --_

This fix helps them to find /lib/ and execute well.

### DIFF
--- a/lib/bundler/templates/newgem/bin/setup.tt
+++ b/lib/bundler/templates/newgem/bin/setup.tt
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 set -euo pipefail
+cd -
 IFS=$'\n\t'
 set -vx
 
--- a/lib/bundler/templates/newgem/bin/setup.tt
+++ b/lib/bundler/templates/newgem/bin/setup.tt
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 set -euo pipefail
-cd -
+cd -P -- "$(dirname -- "$0")"
 IFS=$'\n\t'
 set -vx
 
--- a/lib/bundler/templates/newgem/bin/console.tt
+++ b/lib/bundler/templates/newgem/bin/console.tt
@@ -1,5 +1,9 @@
 #!/usr/bin/env ruby
 
+# help script to find ../lib/ if a shell executes it with another directory
+# set as active (like Mac OS Finder does)
+$LOAD_PATH.unshift File.expand_path('../../lib', __FILE__)
+
 require "bundler/setup"
 require "<%= config[:namespaced_path] %>"
 


lib/bundler/templates/newgem/bin/setup.tt
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 set -euo pipefail
+cd -
```
segiddins: why is this change necessary?

rikkimongoose: Because shells like MacOS Finder produce the same problem to setup as well. They execute it with ~ as active directory and bundle install command just can't find proper files.

With this line I say to script "please, set the directory you're in as the current one". I've found "cd -" one-liner a long time ago somewhere in bash manuals.

rikkimongoose: Checked it at my Ubuntu and found out that cd - is Mac-only shotcut.

Updated it with much more portable and accurate version, borrowed from here:

http://stackoverflow.com/questions/3349105/how-to-set-current-working-directory-to-the-directory-of-the-script

segiddins: `cd -` just changes to the last directory

rikkimongoose: Yes, it was my mistake.

Please, check the following commit, I've wrote a proper version of cd.

segiddins: I'm sorry but I still don't understand why changing directories at all is necessary here

rikkimongoose: Just try this on a Unix or Mac system:

```bash
mkdir ~/dev
cd ~/dev
bundle gem foo
# edit gemspec here, and after....
cd ~
# let's try to install
~/dev/foo/bin/setup
```

It will say "Could not locate Gemfile or .bundle/ directory", because active directory will be ~.

segiddins:
@indirect thoughts?

indirect:
This bug is legit, but I'm not sure yet if this patch is the best solution. I think the solution for `bin/console` is to instead add this line before `require 'bundler/setup'`:

```ruby
BUNDLE_GEMFILE = File.join("Gemfile", File.expand_path("..", __FILE__))
```

Bundler knows to add the `lib` from this gem to the loadpath, but it needs to be able to find the Gemfile for that to happen.

As for `bin/setup`, I'm not sure yet. It seems like getting the directory of the running script from bash is [super scary and unreliable](http://www.ostricher.com/2014/10/the-right-way-to-get-the-directory-of-a-bash-script/).

That said, I definitely think we should make sure that both `bin/console` and `bin/setup` work when executed from any working directory.

colby-swandale:
ping @rikkimongoose

esasse:
#5030 was closed and this PR is not active in a long time as well.

PR #5036, opened 2016-09-30
Fail fast in the specs when running from a path with special characters

segiddins:
\c @indirect

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -17,6 +17,10 @@ rescue LoadError
   abort "Run rake spec:deps to install development dependencies"
 end
 
+if File.expand_path(__FILE__) =~ %r{([^\w/\.])}
+  abort "The bundler specs cannot be run from a path that contains special characters (particularly #{$1.inspect})"
+end
+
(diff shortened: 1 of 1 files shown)


spec/spec_helper.rb
```diff
@@ -17,6 +17,10 @@
   abort "Run rake spec:deps to install development dependencies"
 end
 
+if File.expand_path(__FILE__) =~ %r{[^\w/\.]}
+  abort "The bundler specs cannot be run from a path that contains special characters"
```
hmistry: Nice and short. I suggest being more specific what special characters we're referring to. In this case we're looking for white spaces so lets say that as they can take action without having to looking it up. If down the road the list gets bigger, we can state the common ones and users can check the docs for a full list. I appreciate the quick fix!

indirect: 👍

segiddins: It now prints the offending character

hmistry: Nice... really nice!

indirect:
@homu r+

PR #5042, opened 2016-10-03
Remove build that always fails

indirect:
In retrospect, once this build started always failing, we should have just removed it rather than moved it to allowed failures. I guess I was hoping it would get better. :/


indirect:
@homu r+

Issue #4599, opened 2016-05-22
Error Installing Bundler - Parent Directory is World Writable but not Sticky


JARM876:
Hi, 

I received the following error when trying to install the bundler to use wpscan. Could anyone take a look and let me know what I'm doing wrong? I'm very new to all this. Thanks!

--- ERROR REPORT TEMPLATE -------------------------------------------------------
- What did you do?
  
  I ran the command `/Users/jodi-ann/.rvm/rubies/ruby-2.3.1/bin/bundle install`
- What did you expect to happen?
  
  I expected Bundler to...
- What happened instead?
  
  Instead, what actually happened was...

Error details

```
ArgumentError: parent directory is world writable but not sticky
  /Users/jodi-ann/.rvm/rubies/ruby-2.3.1/lib/ruby/2.3.0/tmpdir.rb:93:in `ensure in mktmpdir'
  /Users/jodi-ann/.rvm/rubies/ruby-2.3.1/lib/ruby/2.3.0/tmpdir.rb:95:in `mktmpdir'
  /Users/jodi-ann/.rvm/gems/ruby-2.3.1@global/gems/bundler-1.12.4/lib/bundler/vendor/compact_index_client/lib/compact_index_client/updater.rb:29:in `update'
  /Users/jodi-ann/.rvm/gems/ruby-2.3.1@global/gems/bundler-1.12.4/lib/bundler/vendor/compact_index_client/lib/compact_index_client.rb:64:in `update'
  /Users/jodi-ann/.rvm/gems/ruby-2.3.1@global/gems/bundler-1.12.4/lib/bundler/vendor/compact_index_client/lib/compact_index_client.rb:55:in `update_and_parse_checksums!'
  /Users/jodi-ann/.rvm/gems/ruby-2.3.1@global/gems/bundler-1.12.4/lib/bundler/fetcher/compact_index.rb:65:in `available?'
  /Users/jodi-ann/.rvm/gems/ruby-2.3.1@global/gems/bundler-1.12.4/lib/bundler/fetcher/compact_index.rb:15:in `call'
  /Users/jodi-ann/.rvm/gems/ruby-2.3.1@global/gems/bundler-1.12.4/lib/bundler/fetcher/compact_index.rb:15:in `block in compact_index_request'
  /Users/jodi-ann/.rvm/gems/ruby-2.3.1@global/gems/bundler-1.12.4/lib/bundler/fetcher.rb:154:in `use_api'
[...]

PR #4561, opened 2016-05-10
Safely store concurrent compact index downloads

domcleal:
When bundler is run concurrently using the same bundle dir in $HOME,
the versions file can be updated from two processes at once. The
download has been changed to a temporary file, which is securely moved
into place over the original.

If retrying the update operation, the original file is no longer
immediately deleted and instead a full download is performed, later
overwriting the original file if successful.

If two processes are updating in parallel, this should ensure the
original file isn't corrupted and that both processes succeed.

- Fixes #4519

---

This would be useful on 1.12.x if possible, since the new caching behaviour with a shared home directory is causing the errors described in #4519.

Issue #4519, opened 2016-05-03
Concurrent bundle install runs raise errors unlinking ~/.bundle/cache/compact_index/rubygems.org.443.[..]/versions


domcleal:
- What did you do?

  I ran the command `bundle install` in four parallel processes using different RVM gemsets (therefore different GEM_HOME), but under the same user account with a shared HOME environment variable.  No Gemfile.lock was present, it was a clean checkout of a project with only a Gemfile.

- What did you expect to happen?

  I expected Bundler to complete without error by either locking the files it's downloading to be safer with multiple processes, or to use caches that are unique per GEM_HOME.

- What happened instead?

  Instead, what actually happened was one of the `bundle install` commands failed intermittently (a very small percentage of the time) with an error deleting `~/.bundle/cache/compact_index/rubygems.org.443.29b0360b937aa4d161703e6160654e47/versions`.

  It appears that the compact index cache is shared for a whole user account without any locking, so multiple processes can try to download the same file at the same time and cause errors for each other.

Error details

    Errno::ENOENT: No such file or directory - /home/jenkins/.bundle/cache/compact_index/rubygems.org.443.29b0360b937aa4d161703e6160654e47/versions
      /usr/local/rvm/gems/ruby-2.0.0-p353@test_kafo_parsers_master_pull_request-0/gems/bundler-1.12.1/lib/bundler/vendor/compact_index_client/lib/compact_index_client/updater.rb:50:in `unlink'
      /usr/local/rvm/gems/ruby-2.0.0-p353@test_kafo_parsers_master_pull_request-0/gems/bundler-1.12.1/lib/bundler/vendor/compact_index_client/lib/compact_index_client/updater.rb:50:in `delete'
      /usr/local/rvm/gems/ruby-2.0.0-p353@test_kafo_parsers_master_pull_request-0/gems/bundler-1.12.1/lib/bundler/vendor/compact_index_client/lib/compact_index_client/updater.rb:50:in `update'
      /usr/local/rvm/gems/ruby-2.0.0-p353@test_kafo_parsers_master_pull_request-0/gems/bundler-1.12.1/lib/bundler/vendor/compact_index_client/lib/compact_index_client.rb:64:in `update'
      /usr/local/rvm/gems/ruby-2.0.0-p353@test_kafo_parsers_master_pull_request-0/gems/bundler-1.12.1/lib/bundler/vendor/compact_index_client/lib/compact_index_client.rb:55:in `update_and_parse_checksums!'
[...]

PR #5043, opened 2016-10-03
use /tmp for mktmpdir

indirect:
As we noticed in #4519, we need to use a temporary directory to hold
compact index downloads so that multiple processes don't write to the
same files at the same time and break everything.

The fix for that was #4561, which added temporary directories to hold
all files as they download, and then uses the (atomic) `FileUtils.cp` to
move the completed downloads into place, so there is never a point where
multiple processes are trying to write into the file at once.

Unfortunately, using `Dir.mktmpdir` requires that the parent directory
be _either_ world writable or sticky, but not both. Based on #4599, it
looks like it's common for home directories to be both world writable
and sticky. While that's a security problem by itself, it's not a big
concern for Bundler and the compact index. So we want to let users
continue to use Bundler, even with the compact index, without having to
change the permissions on their home directories.

This commit changes the `mktmpdir` call to create the temporary
directory inside the default OS tempdir, which is typically `/tmp` or
`/var/tmp` depending on distro. Since that directory is designed to hold
other temporary directories, that change should (theoretically) reduce
or eliminate the problem reported in #4599.

### DIFF
--- a/lib/bundler/compact_index_client/updater.rb
+++ b/lib/bundler/compact_index_client/updater.rb
@@ -27,7 +27,7 @@ module Bundler
       def update(local_path, remote_path, retrying = nil)
         headers = {}
 
-        Dir.mktmpdir(local_path.basename.to_s, local_path.dirname) do |local_temp_dir|
+        Dir.mktmpdir("bundler-compact-index-") do |local_temp_dir|
           local_temp_path = Pathname.new(local_temp_dir).join(local_path.basename)
 
           # download new file if retrying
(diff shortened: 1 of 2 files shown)


segiddins:
@indirect any chance for a test case on this?

indirect:
@segiddins yeah, working on it

indirect:
@segiddins test supplied, pls to review

segiddins:
did you confirm the test fails without the fix?

indirect:
@segiddins yes, I wrote a failing test against master and then committed this test on the branch after ensuring it changed to green.

segiddins:
@homu r+

indirect:
restarted the 1.8.7 build, didn't think to check if I actually broke 1.8 until after the log was gone >_>

PR #5045, opened 2016-10-04
Add conservative update options to bundle lock

### LINKED ISSUE #4912: Update bundle lock with conservative update behavior/options
Task spawned from #4676.

Suggested by @JuanitoFatas in https://github.com/livingsocial/bundler-patch/issues/34, we should investigate getting `bundle lock` in sync with `bundle update` `--patch` and `--minor` behavior.

chrismo:
Fix #4912.

The bundle lock command produces a lockfile same as bundle update but
won't install all the gems. If the `--update` option is fed to `bundle
lock` you can also specify the `--patch`, `--minor`, `--major` and
`--strict` options added to `bundle update` in 1.13.

### DIFF
--- a/lib/bundler/cli/common.rb
+++ b/lib/bundler/cli/common.rb
@@ -52,5 +52,14 @@ module Bundler
       message += "\nDid you mean #{suggestions}?" if suggestions
       message
     end
+
+    def self.config_gem_version_promoter(definition, opts)
+      patch_level = [:major, :minor, :patch].select {|v| opts.keys.include?(v.to_s) }
+      raise InvalidOption, "Provide only one of the following options: #{patch_level.join(", ")}" unless patch_level.length <= 1
+      definition.gem_version_promoter.tap do |gvp|
+        gvp.level = patch_level.first || :major
+        gvp.strict = opts[:strict]
+      end
+    end
   end
 end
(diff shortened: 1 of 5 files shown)


indirect:
I'm impressed by how little code this required. :)

chrismo:
Agreed! I decided to try and get restarted on these, this one seemed potentially daunting and it was nice that it was simple.

The negative-pants view on all this is there's a lot of dup spread out across lock.rb, install.rb, update.rb ... if that wasn't there, we could have had lock behavior on this 'for free' with the prior implementation.

But I'll squelch mr. negative-pants for right now :)

chrismo:
@indirect can you do that @homu voodoo that you do so well?

indirect:
@chrismo I'm excited about a potential future where there's even less duplication. :)

@homu r+

lib/bundler/cli/common.rb
```diff
@@ -52,5 +52,14 @@ def self.gem_not_found_message(missing_gem_name, alternatives)
       message += "\nDid you mean #{suggestions}?" if suggestions
       message
     end
+
+    def self.config_gem_version_promoter(definition, opts)
```
chrismo: yeah ... already merged though. I'll try to make a note to modify it along with the outdated work perhaps.

PR #5053, opened 2016-10-05
Changed the behavior of 'bundle clean --dry-run' to output the list regardless of path set or force option

### LINKED ISSUE #5027: Bundle clean --dry-run should print listing without --force when no path is set
Currently if `path` is not set, `bundle clean --dry-run` requires `--force` to print listing. It would be nice if it bundler followed the common paradigms where `--dry-run` prints a listing of what changes without `--force`, and `--force` is only required when you're prevented from executing the action for some reason and you need to override the failsafes and force the action.

Propose change current `bundle clean --dry-run --force` to `bundle clean --dry-run` for the case when `path` is not set.

hmistry:
Changed the behavior of 'bundle clean --dry-run' to output the list of gems bundle without having the local path set or providing the '--force' option. This change does not affect the actual behavior of 'bundle clean' which requires either the path being set or use of '--force'.

Closes #5027.

### DIFF
--- a/lib/bundler/cli/clean.rb
+++ b/lib/bundler/cli/clean.rb
@@ -8,7 +8,7 @@ module Bundler
     end
 
     def run
-      require_path_or_force
+      require_path_or_force unless options[:"dry-run"]
       Bundler.load.clean(options[:"dry-run"])
     end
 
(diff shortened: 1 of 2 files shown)


hmistry:
Closes #5027.

segiddins:
@homu r+

segiddins:
@homu retry

PR #5055, opened 2016-10-05
[Fetcher::Dependency] Sort gem names in query string

segiddins:
This might slightly help the fastly hit rate?

\c @indirect @dwradcliffe


segiddins:
@homu r+

lynnco:
This just needs to gem names to be sorted in the relevant spec https://github.com/bundler/bundler/blob/master/spec/bundler/fetcher/dependency_spec.rb#L271

Here's the fail https://travis-ci.org/bundler/bundler/jobs/165263641#L434-L435

segiddins:
Fixed, the tests were incorrectly passing in a nested array

@homu r+

PR #5056, opened 2016-10-06
PLACEHOLDER - Conservative updates on outdated

chrismo:
This is a branch from #4841 which is still in the final stages of review. At one point that PR contained the conservative update stuff, too, and there were some code review comments pertaining to breaking behavior that I wanted to copy over here, so I"m making this PR to capture that stuff so I can close out the review of the prior PR.

Once #4841 is merged, I'll rebase this PR.

- [ ] convert existing flags to --filter-*
- [ ] deal with strict flag
- [ ] document breaking change reasons?

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -269,6 +269,8 @@ module Bundler
       versions of the given gems. Prerelease gems are ignored by default. If your gems
       are up to date, Bundler will exit with a status of 0. Otherwise, it will exit 1.
     D
+    method_option "group", :aliases => "--group", :type => :string, :banner => "List gems from a specific group"
+    method_option "groups", :aliases => "--groups", :type => :boolean, :banner => "List gems organized by groups"
     method_option "local", :type => :boolean, :banner =>
       "Do not attempt to fetch gems remotely and use the gem cache instead"
     method_option "pre", :type => :boolean, :banner => "Check for newer pre-release gems"
--- a/lib/bundler/cli/outdated.rb
+++ b/lib/bundler/cli/outdated.rb
@@ -4,6 +4,7 @@ require "bundler/cli/common"
 module Bundler
   class CLI::Outdated
     attr_reader :options, :gems
+
     def initialize(options, gems)
       @options = options
       @gems = gems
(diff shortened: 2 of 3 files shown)


chrismo:
1.13 has new, un-official options in bundle update to restrict updates to only the latest --patch or --minor version available (see bundler/bundler-features#122). bundle outdated had similarly named options added to it previously, but they don't behave in the same way.

For example, if bundle outdated is going to show:

foo: 1.0.0 => 2.0.0
bar: 1.0.0 => 1.1.0
qux: 1.0.0 => 1.0.1

and you add the --patch flag, then it merely filters the results to only show:

qux: 1.0.0 => 1.0.1

rather than showing what could happen as a result of bundle update --patch:

foo: 1.0.0 => 1.0.3
bar: 1.0.0 => 1.0.9
qux: 1.0.0 => 1.0.1

#4772 is a task I made to update bundle outdated to have matching behavior, and when I saw this PR, I simply linked to it to remind me that this PR would be changing code in a similar place.

chrismo:
This is going to cause a problem though which is a key part of this change: the meaning of the switches --patch / --minor / --major are now changing. Before they filtered the output, now they'll be influencing the resolution process.

There's only two ways here, though - we either CHANGE what they mean and rename the filtering ones to something like --filter-patch (which is breaking), or leave them as-is and introduce new flags (--resolve-patch, etc?) that are translated for GemVersionPromoter, which won't be breaking, but will be inconsistent with the flags for bundle update.

My preference is to the former, those new flags weren't introduced too long ago, but I want @indirect's input on this.

chrismo:
The flags as passed to the GemVersionPromoter _change_ resolution. <=1.13.2 of bundle outdated, those flags merely filter the output, with no influence on resolution.

With these changes, `bundle outdated --minor` would be fed to the GemVersionPromoter and actually only resolve `foo` up to `1.1.0`.

chrismo:
from @indirect in [this comment](https://github.com/bundler/bundler/pull/4841#discussion_r82021277):

"IMO, this is what we were trying to do in 1.12, and failed to do because resolving is complicated. :/ I would be okay with this change on the grounds that the previous flags were documented so it sounded like they cause the new resolver aware behavior. 👍"

PR #5057, opened 2016-10-06
Improve -rbundler/setup performance

segiddins:
@indirect please run your benchmarks against this :)

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -665,11 +665,12 @@ module Bundler
     end
 
     def converge_dependencies
+      frozen_bundle = Bundler.settings[:frozen]
       (@dependencies + @locked_deps).each do |dep|
-        locked_source = @locked_deps.select {|d| d.name == dep.name }.last
+        locked_source = @locked_deps.find {|d| d.name == dep.name }
         # This is to make sure that if bundler is installing in deployment mode and
         # after locked_source and sources don't match, we still use locked_source.
-        if Bundler.settings[:frozen] && !locked_source.nil? &&
+        if frozen_bundle && !locked_source.nil? &&
             locked_source.respond_to?(:source) && locked_source.source.instance_of?(Source::Path) && locked_source.source.path.exist?
           dep.source = locked_source.source
         elsif dep.source
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -243,6 +243,11 @@ module Bundler
           Bundler.ui.debug("Found no changes, using resolution from the lockfile")
           last_resolve
         else
+          (sources.path_sources + sources.git_sources).each do |path_source|
+            path_source.specs.each do |path_spec|
+              Bundler.rubygems.validate(path_spec)
+            end
+          end
           # Run a resolve against the locally available gems
           Bundler.ui.debug("Found changes from the lockfile, re-resolving dependencies because #{change_reason}")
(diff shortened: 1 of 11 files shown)


lib/bundler/definition.rb
```diff
@@ -581,6 +586,7 @@ def dependencies_for_source_changed?(source, locked_source = source)
     end
 
     def specs_for_source_changed?(source)
+      return false if source.is_a?(Source::Path) && !source.has_spec_files? && Bundler.settings[:frozen]
```
segiddins: I think this was trying to address a different issue, and so it needs test coverage

lib/bundler/definition.rb
```diff
@@ -728,7 +735,7 @@ def converge_locked_specs
         # then we unlock it.
 
         # Path sources have special logic
-        if s.source.instance_of?(Source::Path) || s.source.instance_of?(Source::Gemspec)
+        if (s.source.instance_of?(Source::Path) || s.source.instance_of?(Source::Gemspec)) && (s.source.has_spec_files? || !Bundler.settings[:frozen])
```
segiddins: same thing, different issue & needs tests

lib/bundler/rubygems_ext.rb
```diff
@@ -167,7 +167,11 @@ def matches_spec?(spec)
   class Requirement
     # Backport of performance enhancement added to Rubygems 1.4
     def none?
-      @none ||= (to_s == ">= 0")
+      if @requirements.size == 1
```
segiddins: this is taking from the current RubyGems and avoids allocations

segiddins:
Failures appear to be only for Ruby < 2 and not related to the core of these changes

segiddins:
I can't reproduce the failures locally :(

segiddins:
@indirect if we can pinpoint which one of these caused the regression, I'd really like to ship the rest

esparta:
I was able to reproduce one of the failures. And in order to fix it, instead of this:
```diff
diff --git a/spec/runtime/setup_spec.rb b/spec/runtime/setup_spec.rb
index eab66e2..236f600 100644
--- a/spec/runtime/setup_spec.rb
+++ b/spec/runtime/setup_spec.rb
@@ -152,6 +152,7 @@ describe "Bundler.setup" do
         bundler_path.join("gems/bundler-#{Bundler::VERSION}/lib").to_s,
         tmp("rubygems/lib").to_s,
         root.join("../lib").expand_path.to_s,
+        root.join("../lib/bundler/vendor").expand_path.to_s,
       ]
       load_path.map! {|lp| lp.sub(/^#{system_gem_path}/, "") }
```

Should be this:

```diff
diff --git a/spec/runtime/setup_spec.rb b/spec/runtime/setup_spec.rb
index eab66e2..236f600 100644
--- a/spec/runtime/setup_spec.rb
+++ b/spec/runtime/setup_spec.rb
@@ -152,6 +152,7 @@ describe "Bundler.setup" do
         bundler_path.join("gems/bundler-#{Bundler::VERSION}/lib").to_s,
         tmp("rubygems/lib").to_s,
         root.join("../lib").expand_path.to_s,
+        root.join("../bundler/lib/bundler/vendor").expand_path.to_s,
       ]
       load_path.map! {|lp| lp.sub(/^#{system_gem_path}/, "") }
```

The main question is: are we expecting path be on `..lib/bundler/vendor`?  Because the 'fix' seems to easy.

indirect:
@segiddins ping

segiddins:
@indirect we still haven't narrowed down what exactly was causing the performance regression

PR #5060, opened 2016-10-06
Add more details on URI::InvalidComponentError

perlun:
This would have helped me when I had misunderstood the value of the `ENV['BUNDLE_GEM__FURY__IO']` variable used to provide details to private Rubygems sources. The job was running on Travis so it was hard to debug. (I had put the whole URL there, not just the username part)

Downside: it _will_ likely contain the secret text. Something like this: `bad component(expected user component): //<secret>@gem.fury.io/something-uber-cool/`

I will have total understanding if the PR is rejected for this exact reason. (What I did was to reproduce the problem locally, then add something like this in my local copy of bundler.)

### DIFF
--- a/lib/bundler/source/rubygems/remote.rb
+++ b/lib/bundler/source/rubygems/remote.rb
@@ -39,9 +39,9 @@ module Bundler
           end
 
           uri
-        rescue URI::InvalidComponentError
+        rescue URI::InvalidComponentError => e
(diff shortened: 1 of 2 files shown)


segiddins:
I don't think this change is going to be acceptable, exactly because it will print out private credentials.

perlun:
Thanks @segiddins, that's fine. The change is still here if someone happens to need it while debugging, like I did.

PR #5061, opened 2016-10-06
Conservative updates on outdated

### LINKED ISSUE #5065: outdated filter options don't work with --strict option
`--strict` only shows the latest version of a gem that still satisfies any requirement in the Gemfile. The addition of `--patch`/`--minor`/`--major` flags only works if `--strict` is not provided.

```
bundle outdated --patch

  * qux (newest 1.0.1, installed 1.0.0) in groups "dev, zep"


bundle outdated --patch --strict

  * bar (newest 1.1.0, installed 1.0.0) in group "foo_group"
  * foo (newest 2.0.0, installed 1.0.0) in group "foo_group"
  * qux (newest 1.0.1, installed 1.0.0) in groups "dev, zep"
```

I found this while working on #5061, so I can perhaps fix it up in that PR, too.

### LINKED ISSUE #5076: outdated shouldn't say "Bundle up to date!" if results are just filtered out.
See this case: https://github.com/chrismo/bundler-cases/blob/master/cases/issues/bundler-5076.rb

If your gem file only has a major increment available, and the `--minor` or `--patch` options are fed to it, it filters out the major update, but then still says, "Bundle up to date!" which isn't true.

chrismo:
Add conservative resolving behavior to outdated command.

- [x] convert existing flags to --filter-*
- [x] deal with strict flag
- [x] make a 2.0 issue to consider making strict flags more consistent
- [x] fix #5065 (outdated filter options don't work with `--strict')
- [x] fix #5076 (outdated shouldn't say "Bundle up to date!" if results are just filtered out.)
- [ ] document breaking change reasons? (_commit comment has something at least now_)
- [x] what about `bundle show --outdated`? (_it's a much simpler version ... prolly just going to leave it alone for now?_)

The flags as passed to the GemVersionPromoter _change_ resolution. <=1.13.2 of bundle outdated, those flags merely filter the output, with no influence on resolution.

With these changes, `bundle outdated --minor` would be fed to the GemVersionPromoter and actually only resolve `foo` up to `1.1.0`. 

This gist shows how it currently works, filtering the output: https://gist.github.com/chrismo/0bc6cfa00f539787101a9a2c900616d3

It's unfortunate timing. They were only added in 1.12 ... I'm biased, but feel like the affect the flags have on resolution is of greater value, and would be better to keep in sync with how they work fed to the `bundle update` command as of 1.13, and we could add new --filter-patch flags to replace what they do currently. 

IIRC, there was some confusion at the time that Andre perhaps even thought these flags as of 1.12 would be affecting what the newest would resolve to, instead of just filtering the output, so - _if_ I'm remembering correctly, that's also influencing my opinion. 

But, I can be swayed by the breaking behavior argument.

from @indirect in [this comment](https://github.com/bundler/bundler/pull/4841#discussion_r82021277):

"IMO, this is what we were trying to do in 1.12, and failed to do because resolving is complicated. :/ I would be okay with this change on the grounds that the previous flags were documented so it sounded like they cause the new resolver aware behavior. 👍"

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -268,6 +268,9 @@ module Bundler
       in the given source. Calling outdated with [GEM [GEM]] will only check for newer
       versions of the given gems. Prerelease gems are ignored by default. If your gems
       are up to date, Bundler will exit with a status of 0. Otherwise, it will exit 1.
+
+      For more information on conservative resolution options (--major, --minor,
+      --patch, --strict) see documentation on the same options on the update command.
     D
     method_option "group", :aliases => "--group", :type => :string, :banner => "List gems from a specific group"
     method_option "groups", :aliases => "--groups", :type => :boolean, :banner => "List gems organized by groups"
@@ -277,9 +280,14 @@ module Bundler
     method_option "source", :type => :array, :banner => "Check against a specific source"
     method_option "strict", :type => :boolean, :banner =>
       "Only list newer versions allowed by your Gemfile requirements"
-    method_option "major", :type => :boolean, :banner => "Only list major newer versions"
-    method_option "minor", :type => :boolean, :banner => "Only list minor newer versions"
-    method_option "patch", :type => :boolean, :banner => "Only list patch newer versions"
+    method_option "update-strict", :type => :boolean, :banner =>
+      "Strict conservative resolution, do not allow any gem to be updated past latest --patch/--minor/--major"
+    method_option "minor", :type => :boolean, :banner => "Prefer updating only to next minor version"
+    method_option "major", :type => :boolean, :banner => "Prefer updating to next major version (default)"
+    method_option "patch", :type => :boolean, :banner => "Prefer updating only to next patch version"
+    method_option "filter-major", :type => :boolean, :banner => "Only list major newer versions"
+    method_option "filter-minor", :type => :boolean, :banner => "Only list minor newer versions"
+    method_option "filter-patch", :type => :boolean, :banner => "Only list patch newer versions"
     method_option "parseable", :aliases => "--porcelain", :type => :boolean, :banner =>
       "Use minimal formatting for more parseable output"
     def outdated(*gems)
--- a/lib/bundler/cli/common.rb
+++ b/lib/bundler/cli/common.rb
@@ -53,13 +53,17 @@ module Bundler
       message
     end
 
-    def self.config_gem_version_promoter(definition, opts)
-      patch_level = [:major, :minor, :patch].select {|v| opts.keys.include?(v.to_s) }
+    def self.configure_gem_version_promoter(definition, options)
+      patch_level = patch_level_options(options)
       raise InvalidOption, "Provide only one of the following options: #{patch_level.join(", ")}" unless patch_level.length <= 1
       definition.gem_version_promoter.tap do |gvp|
         gvp.level = patch_level.first || :major
-        gvp.strict = opts[:strict]
+        gvp.strict = options[:strict] || options["update-strict"]
       end
+    end
+
+    def self.patch_level_options(options)
+      [:major, :minor, :patch].select {|v| options.keys.include?(v.to_s) }
     end
   end
 end
--- a/lib/bundler/cli/lock.rb
+++ b/lib/bundler/cli/lock.rb
@@ -25,7 +25,7 @@ module Bundler
       update = { :gems => update } if update.is_a?(Array)
       definition = Bundler.definition(update)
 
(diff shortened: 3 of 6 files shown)


chrismo:
My proactive PR #5056 was a fail since it had the same commit in it as #4841. Anyway - back at it here.

chrismo:
`--strict` on the `outdated` command was added in 1.5 ... been around a while. It's also the same name as a flag I've been using on the new conservative update options in `bundle update` - it further restricts resolution to not let any gem for any reason move up past the specified level (`--patch`/`--minor`). 

On `outdated` it means don't show the very latest available version of a gem, only show the very latest version that still satisfies the declared requirement. 

In that vein, perhaps best to add a new flag `--update-strict` to indicate using strict conservative update resolution behavior. Sucks that it can't match ... but best non-breaking option in 1.x?

For 2.0, it might make sense to get rid of `--strict` and report both versions at the same time, and re-use the flag for the update resolution behavior. (For 2.0, it'd be really great to overhaul the code, separate the data from the presentation and make more customizable options a la `git log PRETTY FORMATS` (think `strftime`))

@indirect @segiddins (or anyone else) any opinions?

chrismo:
another thing - the `--strict` flag as it works now has to be in effect for the `--patch`/`--minor`/`--major` flags to be useful at all. specifying `--patch` with the current non-strict behavior means it will ignore the result of the resolution and just show the latest available anyway. 

that said, a 3rd option would be to override the meaning of `--strict` in the case when `--patch` / `--minor` / `--major` are used ... but ... that seems overly clever and a bad idea in general.

segiddins:
> that said, a 3rd option would be to override the meaning of --strict in the case when --patch / --minor / --major are used ... but ... that seems overly clever and a bad idea in general.

honestly, that doesn't sound too awful to me? That way, it doesn't break the current meaning of `--strict`

chrismo:
well, thinking through this use case, I'm not so sure now:

- User is used to `--strict` on outdated, been using it forever.
- Upgrades to 1.1x.x - sees new conservative update options on outdated.
- Runs `bundle outdated --strict --patch` - using `--strict` out of habit, but now it affects resolution, and prevents any dependencies for any reason going past the latest patch version.
- Runs `bundle update -- patch` - skips using the `--strict` option because habitually thinks it's only an `outdated` option, and now the resolution is more lenient, and allows a dependency to go up a minor version to satisfy a change in a declared dependency going up a patch version.
- Files a ticket here saying this thing is broken - what `outdated` reported and what `update` did wasn't the same.
- I attempt to explain how we co-opted the `--strict` flag to mean something different and the user gives us a justified :| emoji.

chrismo:
I shouldn't keep noticing things - I've logged more Issues than code on this so far ;)

chrismo:
I hate adding slow integration tests to this. I may return after my conservative update adventure and refactor most of this stuff into classes where we can do thorough combo testing at a unit level and minimize the time consuming integration runs.

chrismo:
@indirect @segiddins this is ready for review - the failing tests are unrelated. Something about real world edge cases that Andre is aware of in Slack.

segiddins:
Seems good to me, but I'll let @indirect have a crack at a more in depth review if he wants :)

chrismo:
ping @indirect - when you have time.

indirect:
I am happy with this change. I'm also happy to deprecate `bundle show --outdated` and remove it in 2.0.

I think that it maybe makes sense in 2.0 to make strict the default mode? @chrismo, I feel like you have a much clearer understanding of user expectations around `update` and `outdated` than I do  right now, so I'm prepared to accept your recommendation. :)

indirect:
@homu r+

chrismo:
I was surprised at first to see --strict as an option ... thinking I'd probably want that knowledge ... but then ... if I don't have it clear in the output, I could see expectations varying among users. 

"I think that it maybe makes sense in 2.0 to make strict the default mode?" I think I'd recommend getting rid of the option, and including both into the output - show both latest according to declared requirement and latest regardless.

e.g. `gem 'rails', '~> 4.0'` shows both 4.2 (within req) and 5.0 (in general). Probably good to display what the requirement is as well in the output ... so it's there for the user to have handy without having to look it up.

I made #5081 for discussion on that. In fact, I'll go update it with that above detail.

segiddins:
@chrismo appears to fail on super-old rubygems

chrismo:
yes - that is sad-face. ... ah, it's an issue with older RubyGems and how `Gem::Dependency#specific?` must be implemented. Loosened up the matching string on those 3 specs.

chrismo:
that fixed it - @indirect can you re-homu?

chrismo:
@indirect @homu ping

indirect:
@homu retry

indirect:
@homu r+

PR #5064, opened 2016-10-08
change checksum warning to debug print

indirect:
This was super helpful when the server was continuously returning bad
checksums, but it’s a scary warning to see anytime we update the 
versions file. And we’re going to need to update the versions file at
least twice a year, so it seems like a good idea to head off users
worrying about a message that is actually things working as intended.


segiddins:
👍🏻

indirect:
@homu r+

Issue #5062, opened 2016-10-06
Clarify documentation around global sources


indirect:
As demonstrated by [this blog post](http://collectiveidea.com/blog/archives/2016/10/06/bundlers-multiple-source-security-vulnerability/), there is still a lot of end-user confusion about how to deal with the source issues originally revealed in http://bundler.io/blog/2014/08/14/bundler-may-install-gems-from-a-different-source-than-expected-cve-2013-0334.html.

A lot of that confusion is likely our fault as a team—we often weren't sure what was and wasn't possible even as we were trying to fix the problem. As the above-linked blog post mentions, the only way to be 100% safe in Bundler 1.x is to have no global sources.

The cross-source confusion is eliminated in the (as yet unreleased) Bundler 2.x series by a series of backwards-compatibility breaking changes to the format of the `Gemfile.lock` file and the way that Bundler handles gem sources internally. Once Bundler 2 is out, you'll be able to use one global source and additional non-global sources without worrying about any name conflicts.

As a result of these problems, let's try to make it clearer for users what they need to do:

- [ ] Update [the CVE warning post](http://bundler.io/blog/2014/08/14/bundler-may-install-gems-from-a-different-source-than-expected-cve-2013-0334.html) with clearer instructions around the possible problem and ways to avoid it
- [ ] Update the discussion of multiple sources in the Bundler docs to point out that this opens the possibility of name conflicts, and suggest using no global sources in Bundler 1.x Gemfiles.
- [ ] Update the existing warnings as needed to reflect the problems discussed in [the new blog post](http://collectiveidea.com/blog/archives/2016/10/06/bundlers-multiple-source-security-vulnerability/)

This is a good chance for anyone to contribute, even if they aren't familiar with the Bundler code, since the explanations and warnings have all already been written down at least once. 👍

sfcgeorge:
Nice clear explanation, thank you 🏌️ I'll aim to jump on some of those doc improvements then 🤓

lynnco:
This is still available for anyone who wants to address it

baweaver:
If no one claims it by tonight I'll get it done.

baweaver:
Ok then, dibs. I'll get out a short form PR and ask for reviews here in a bit. I'll likely be citing information from the above blog post.

lynnco:
[...]

PR #5074, opened 2016-10-11
Updated to use source blocks

baweaver:
As per issue #5062

Noted that this isn't currently updating the specs and probably needs some work. Call this a prelim draft of what may need to be updated. I welcome any suggestions on improving it.

### DIFF
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -100,7 +100,9 @@ module Bundler
         Bundler.ui.warn <<-WARN, :wrap => true
           Your Gemfile has no gem server sources. If you need gems that are \
           not already on your machine, add a line like this to your Gemfile:
-          source 'https://rubygems.org'
+          source 'https://rubygems.org' do
+            # Gems here
+          end
         WARN
       end
       raise e
--- a/man/gemfile.5.ronn
+++ b/man/gemfile.5.ronn
@@ -20,18 +20,19 @@ a number of methods used to describe the gem requirements.
 At the top of the `Gemfile`, add a line for the `Rubygems` source that contains
 the gems listed in the `Gemfile`.
 
-    source "https://rubygems.org"
+Global source lines are a security risk and should not be used as they can lead
+to gems being installed from unintended sources. If alternate sources are
+specified in the file, they will override the default global source even if
+they are only specified for a block or for a single gem. To avoid this use a
+source block that points to rubygems.org:
 
-It is possible, but not recommended as of Bundler 1.7, to add multiple global
-`source` lines. Each of these `source`s `MUST` be a valid Rubygems repository.
+    source 'https://rubygems.org' do
+      # Gems here
+    end
 
-Sources are checked for gems following the heuristics described in
-[SOURCE PRIORITY][]. If a gem is found in more than one global source, Bundler
-will print a warning after installing the gem indicating which source was used,
-and listing the other sources where the gem is available. A specific source can
-be selected for gems that need to use a non-standard repository, suppressing
-this warning, by using the [`:source` option](#SOURCE-source-) or a
-[`source` block](#BLOCK-FORM-OF-SOURCE-GIT-PATH-GROUP-and-PLATFORMS).
+The only way to properly map gems to a source is to use source blocks instead of
+global sources. Use of single line `source` specifications will open all gems in
+your Gemfile to a second source.
 
 ### CREDENTIALS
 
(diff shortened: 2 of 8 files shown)


segiddins:
This shouldn't be necessary if there's only a single RubyGems source in the Gemfile

lib/bundler/cli/install.rb
```diff
@@ -100,7 +100,9 @@ def run
         Bundler.ui.warn <<-WARN, :wrap => true
           Your Gemfile has no gem server sources. If you need gems that are \
           not already on your machine, add a line like this to your Gemfile:
-          source 'https://rubygems.org'
+          source 'https://rubygems.org' do
```
cben: Perhaps also reword above line from "add a line" to "wrap your Gemfile in a block like this" or something similar.

man/gemfile.5.ronn
```diff
@@ -20,18 +20,19 @@ a number of methods used to describe the gem requirements.
...
-`source` lines. Each of these `source`s `MUST` be a valid Rubygems repository.
-
-Sources are checked for gems following the heuristics described in
-[SOURCE PRIORITY][]. If a gem is found in more than one global source, Bundler
-will print a warning after installing the gem indicating which source was used,
-and listing the other sources where the gem is available. A specific source can
-be selected for gems that need to use a non-standard repository, suppressing
-this warning, by using the [`:source` option](#SOURCE-source-) or a
-[`source` block](#BLOCK-FORM-OF-SOURCE-GIT-PATH-GROUP-and-PLATFORMS).
+Global source lines are a security risk and should not be used as they can lead
```
cben: This doesn't mention single-gem `:source` options.
I think the audience the docs should be written for is not just for new people writing Gemfile from scratch, but people already having Gemfiles and trying to understand how they actually behave.

Worse, other places in the file do describe `:source` as something that sounds safe, eg.:

> This forces the gem to be loaded from this source and ignores any global sources

but [the blog post][1] specifically shows that:

```
source "http://public.org"
gem "foo"
gem "bar", source: "http://private.com"
gem "baz"
```
is unsafe.  I suppose all places mentioning `source` / `:source` should be touched.

Are `:source` exceptions safe if added inside a `source 'https://rubygems.org' do` block?  Are they always bad?

[1]: http://collectiveidea.com/blog/archives/2016/10/06/bundlers-multiple-source-security-vulnerability/

deivid-rodriguez:
I agree with @segiddins that we don't need this. I think using a single global default gem source is fine?

@indirect Do you have any thoughts?

indirect:
most Gemfiles only have one source, and don't need blocks. even if your Gemfile has more than one source, I think the generally ok way to go is one global source and other sources getting their own blocks. unless I'm missing something, anyway.

cben:
This was all intended to educate how to use multiple sources safely without hitting the bundler 1.x security bug with global sources:

https://bundler.io/blog/2014/08/14/bundler-may-install-gems-from-a-different-source-than-expected-cve-2013-0334.html
https://collectiveidea.com/blog/archives/2016/10/06/bundlers-multiple-source-security-vulnerability

A lot of time has passed since.

- Did 2.x fully solve the issue?
- Are all 1.x affected?
- Have everybody migrated to 2.x?  (I think not yet?)

Current docs https://bundler.io/man/gemfile.5.html#SOURCE do mention the issue:

> Using the :source option for an individual gem will also make that source available as a possible global source for any other gems which do not specify explicit sources. Thus, when adding gems with explicit sources, it is recommended that you also ensure all other gems in the Gemfile are using explicit sources.

(is this still true in 2.x?!)

above quote *technically* disclaims the issue but not in sufficient detail, doesn't explain the 1.x/2.x history, doesnt define what exactly "explicit sources" means, doesn't educate *how* to write a mixed-source Gemfile that actually means what a human reader would think it means.

Are there separate URLs for 1.x and 2.x docs?

deivid-rodriguez:
Yeah, a lot of time has passed indeed...

The definitions, as I understand them, are:

* Top level source: `source` without a block. For example, the omnipresent `source "https://rubygems.org"`.
* Explicit source: `source` with a block. For example, `source "http://my_gem_server" do gem "rails" end`.

I think the "Resolution" section is https://bundler.io/blog/2014/08/14/bundler-may-install-gems-from-a-different-source-than-expected-cve-2013-0334.html is not accurate and the Gemfile given as an example there is perfectly fine. The issue is when using multiple top level sources in the same Gemfile. When that happens, `bundler` will warn you, and will fail when bundler 3.0 is released.

deivid-rodriguez:
Ok, I now read the other link you posted and I see where this all comes from. We need to study this and give a proper answer to your questions. I'll try to investigate it and get back here when I have more answers.

Issue #5086, opened 2016-10-14
Version as an optional parameter for bundle inject 


fredrb:
As for now, `bundle inject` requires a `version` parameter with no default value. 

`Usage: "bundle inject GEM VERSION ..."`

From my point of view, it makes sense to run: 
`bundle inject 'rake'`
and get the latest version of the gem. 

Also, `CLI::Inject` class is already expecting `nil` as a feasible parameter for `version`. But this will never happend because `version` is specified as a required parameter. 
```ruby
    def initialize(options, name, version, gems)
      @options = options
      @name = name
      @version = version || last_version_number
      @group = options[:group]
      @source = options[:source]
      @gems = gems
    end

```

indirect:
I think this would be provided by https://github.com/bundler/bundler/issues/4901, so I will close this ticket in favor of that one.

Issue #4901, opened 2016-08-21
Fixes for `add` command


indirect:
- [ ] add an artifice test so we know it works against e.g. the dependency API
- [ ] get gem version list directly from source using fetcher
- [ ] either remove `--source` option or respect it when fetching the list of gem versions
- [ ] handle gem names that aren't present on rubygems.org

segiddins:
@b-ggs any interest in doing this?

shuboid:
Wanna contribute .. Leads will be appreciated.TIA

segiddins:
@shuboid the task items @indirect has posted above should be a good outline of the work that needs to be done

colby-swandale:
Hi @indirect,
I'm working on an implementation of this feature and i have a couple of questions. In terms of the interface is the idea you have along the lines of `bundler add pg '~> 1.5'`?  I assume that we allow users to specify the version and maybe also specify optional groups as a flag like `--group development` as well. Thoughts?

hmistry:
@shuboid Could you let me know if you're working on this feature? If you're not, then I'd like to work on it.

denniss:
Adding the RFC here: 
https://github.com/hmistry/rfcs/blob/bundle-add/text/0000-bundle-add.md

I'm currently giving this a try. Let me know if someone else is already working on this.

denniss:
I think this one can be closed.

segiddins:
Closed by https://github.com/bundler/bundler/pull/5486

PR #5087, opened 2016-10-14
Version as an optional parameter for bundle inject

fredrb:
This implements the suggestion in #5086


indirect:
I don't think this is a good implementation, because it will allow you to pass in non-existent versions that will immediately cause an exception. We already have a ticket open for the `add` command, with information on how we would like it to work: https://github.com/bundler/bundler/issues/4901

Would you be interested in helping us finish the `add` command? That would be super cool!

Issue #4754, opened 2016-07-05
The show command could be a lot better


indirect:
This is the help for the `bundle show` command:

```
$ bundle show --help
Usage:
  bundle show GEM [OPTIONS]

Options:
      [--paths=List the paths of all gems that are required by your gems.rb.], [--no-paths]   
      [--outdated=Show verbose output including whether gems are outdated.], [--no-outdated]  
      [--no-color], [--no-no-color]                                                           # Disable colorization in output
  -r, [--retry=NUM]                                                                           # Specify the number of times you wish to attempt network commands
  -V, [--verbose], [--no-verbose]                                                             # Enable verbose output mode

Description:
  Show lists the names and versions of all gems that are required by your gems.rb. Calling show with [GEM] will list the exact location of that gem on your machine.
```

We don't actually have `--no-paths`, `--no-outdated`, or `--no-verbose` options, so we should remove those if Thor will let us. Showing `--no-color` and `--no-no-color` is... pretty silly. Let's change that to (if we must) `--color` and `--no-color`, or better yet only `--no-color`.

Right now `bundle show` lists gem names, while `bundle show --paths` or `bundle show NAME` shows the paths to gems. Then, confusingly, `bundle show --outdated` shows detailed information including summaries, websites, and whether a gem is up to date or not. We'll probably want to keep the `bundle show --paths` and `bundle show NAME` commands around for the entire 2.x cycle so that tools will work with both Bundler 1 and Bundler 2, but we can make the user-facing commands better for sure.

Let's move `bundle show` to `bundle list` (with an alias of `bundle ls`). That seems to make sense, and gets us firmly into shared territory with `npm list`/`npm ls`.

Let's move `bundle show --outdated` to `bundle info NAME`, and completely remove the `--outdated` option. We already have an outdated command. (Both `brew` and `npm` use `info`. This seems like the right name.) T `bundle info` command without a gem name can either throw an error or print out bundle statistics, like number of Gemfile gems, total number of gems, and maybe... size of the bundle in MB? Sounds nice.

We can also add a `bundle info --path NAME` option to print out the path to a single gem. Later, when Bundler 1.x usage has dropped down to nearly nothing, we can deprecate `bundle show NAME` in favor of `bundle info --path NAME`.

segiddins:
@indirect there was a PR to fix the -no-no-color thing months ago that never got your r+

indirect:
[...]

Issue #5082, opened 2016-10-13
Remove bundle show --outdated for 2.0?


chrismo:
`bundle show --outdated` is a minimal, duplicate implementation of `bundle outdated` ... perhaps we should just remove that flag (and/or change it to redirect to the outdated command). See #5061.

chrismo:
ah, this is a dupe of a thing within #4754.

lynnco:
#4754 is a bit busy, so I think its reasonable to leave this as a single todo list item

segiddins:
`bundle show` is gone in 2.0

PR #5093, opened 2016-10-15
bundle info command

fredrb:
This PR refers to #4754. 

It introduces the command `bundle info GEM` that takes some responsabilities out of `bundle show`. 
### Example

`bundle info rake` will output information of a single gem, different from `bundle show --outdated` that outputs informataion of all gems:

```
* rake (11.3.0)
    Summary:  Rake is a Make-like program implemented in Ruby
    Homepage: https://github.com/ruby/rake
    Status:   Up to date
```

**Options:**
`--path` will output only gem's installation path:

```
$ bundle info rake --path 
/home/fredrb/.rvm/gems/ruby-2.3.1/gems/rake-11.3.0
```

---

As for now, the `GEM` parameter is not optional. I think this makes sense if there will be a `bundle list` command (or something similar) that will output the full list of gems. What do you guys think?

I also added a _deprecated warning_ in `bundle show --outdated` if that's fine.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -245,6 +245,14 @@ module Bundler
     # TODO: 2.0 remove `bundle list`
     map %w(list) => "show"
 
+    desc "info GEM [OPTIONS]", "Shows gem information"
+    method_option "path", :type => :boolean,
+                   :banner => "Show gem path"
+    def info(gem_name)
+      require "bundler/cli/info"
+      Info.new(options, gem_name).run
+    end
+
     desc "binstubs GEM [OPTIONS]", "Install the binstubs of the listed gem"
     long_desc <<-D
       Generate binstubs for executables in [GEM]. Binstubs are put into bin,
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -247,6 +247,7 @@ module Bundler
 
     desc "info GEM [OPTIONS]", "Shows gem information"
     method_option "path", :type => :boolean,
+                   :default => false,
                    :banner => "Show gem path"
     def info(gem_name)
       require "bundler/cli/info"
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -247,8 +247,8 @@ module Bundler
 
     desc "info GEM [OPTIONS]", "Shows gem information"
     method_option "path", :type => :boolean,
-                   :default => false,
-                   :banner => "Show gem path"
+                          :default => false,
+                          :banner => "Show gem path"
     def info(gem_name)
       require "bundler/cli/info"
       Info.new(options, gem_name).run
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -246,9 +246,6 @@ module Bundler
     map %w(list) => "show"
 
     desc "info GEM [OPTIONS]", "Shows gem information"
-    method_option "path", :type => :boolean,
-                          :default => false,
-                          :banner => "Show gem path"
     def info(gem_name)
       require "bundler/cli/info"
       Info.new(options, gem_name).run
--- a/lib/bundler/cli/info.rb
+++ b/lib/bundler/cli/info.rb
@@ -1,0 +1,59 @@
+# frozen_string_literal: true
+require "bundler/cli/common"
+
+module Bundler
+  class CLI::Info
+    attr_reader :gem_name, :options, :latest_specs
+    def initialize(options, gem_name)
+      @options = options
+      @gem_name = gem_name
+      @latest_specs = fetch_latest_specs if not options[:path]
+    end
+
+    def run
+      Bundler.ui.silence do
+        Bundler.definition.validate_runtime!
+        Bundler.load.lock
+      end
+
+      spec = Bundler::CLI::Common.select_spec(gem_name, :regex_match)
+      return unless spec
+
+      path = spec.full_gem_path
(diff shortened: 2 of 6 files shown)


fredrb:
Also: #5082

lynnco:
To clarify, this implements the `info` command in 1.x (from https://github.com/bundler/bundler/issues/4754), so that the `show` command can be removed in 2.0

segiddins:
@indirect please review?

indirect:
I am happy with this once the typo is fixed. It's a good start towards the `list` and `info` commands outlined in #4754. 👍

colby-swandale:
I have a few thoughts before this feature is merged. I think it will be a good idea to show the path as part of the output without needing to give the `--path` option, example being:
```
* rake (11.3.0)
    Summary:  Rake is a Make-like program implemented in Ruby
    Homepage: https://github.com/ruby/rake
    Status:   Up to date
    Path:     /home/fredrb/.rvm/gems/ruby-2.3.1/gems/rake-11.3.0
```
Seeing that `bundle show` outputs the path of the gem, users will be expecting `info` to do the same.

Also if this is intended to replace `bundle show` should we update the post install [success message](https://github.com/bundler/bundler/blob/master/lib/bundler/cli/install.rb#L79)?
 
    Use `bundle info [gemname]` to see where a bundled gem is installed.`

Let me know what everyone thinks.

Thanks.

segiddins:
@colby-swandale all points seem reasonable to me 👍

fredrb:
Agreed! Thanks for pointing out @colby-swandale.

fredrb:
@indirect @segiddins I think this is ready. I removed the `--path` argument and with that a lot of extra code could also be removed. Info command class is a bit simpler now.

lib/bundler/cli.rb
```diff
@@ -245,6 +245,12 @@ def show(gem_name = nil)
     # TODO: 2.0 remove `bundle list`
     map %w(list) => "show"
 
+    desc "info GEM [OPTIONS]", "Shows gem information"
```
segiddins: `Show information for the given gem` ?

lib/bundler/cli/info.rb
```diff
@@ -0,0 +1,47 @@
...
+    attr_reader :gem_name, :options
+    def initialize(options, gem_name)
+      @options = options
+      @gem_name = gem_name
+    end
+
+    def run
+      Bundler.ui.silence do
+        Bundler.definition.validate_runtime!
+        Bundler.load.lock
```
segiddins: should we really be locking here?

lib/bundler/cli/info.rb
```diff
@@ -0,0 +1,47 @@
...
+      end
+
+      spec = Bundler::CLI::Common.select_spec(gem_name, :regex_match)
+      return print_gem_info(spec) if spec
```
segiddins: the `return` here is unnecessary

lib/bundler/cli/info.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+  private
+
+    def print_gem_info(spec)
+      desc = "  * #{spec.name} (#{spec.version}#{spec.git_version})"
+      latest = fetch_latest_specs.find {|l| l.name == spec.name }
```
segiddins: should this match on platform as well?

lib/bundler/cli/info.rb
```diff
@@ -0,0 +1,47 @@
...
+      Bundler.ui.info <<-END.gsub(/^ +/, "")
+        #{desc}
+        \tSummary:  #{spec.summary || "No description available."}
```
segiddins: what would you think about not printing these fields if they're unset?

lib/bundler/cli/show.rb
```diff
@@ -17,6 +17,10 @@ def run
         Bundler.load.lock
       end
 
+      if options[:outdated]
+        Bundler.ui.warn("Deprecated: use `bundle info GEM` or `bundle outdated` instead of `bundle show --outdated`")
```
segiddins: I think this should use `SharedHelpers.major_deprecation`

spec/commands/info_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      install_gemfile <<-G
+        source "file://#{gem_repo1}"
+        gem "rails"
+      G
+    end
+
+    it "should print summary of rake gem" do
+      bundle "info rails"
+
+      expect(out).to include("rails")
```
segiddins: please test for the entirety of the output, instead of the many `include` tests

spec/commands/info_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      expect(out).to include("\tSummary:")
+      expect(out).to include("\tHomepage:")
+      expect(out).to include("\tStatus:")
+      expect(out).to include("\tPath:")
+    end
+
+    it "should create a Gemfile.lock if not existing" do
```
segiddins: why is this desired behavior?

fredrb:
@segiddins Thanks for the review. I'll check the points you mentioned.

colby-swandale:
@fredrb Are you still working on this? I would like to see this merged before the next release of Bundler. Let me know if you're not and i can put some time into this.

fredrb:
@colby-swandale I'm not working on it. Unfortunately I can't put the needed effort at the moment, I would appreciate a lot if you could pick this up!

Issue #5084, opened 2016-10-13
undefined method `start_with?' for nil:NilClass


mdumke:
--- ERROR REPORT TEMPLATE -------------------------------------------------------
- What did you do?

  I ran the command `/home/ubuntu/.rvm/gems/ruby-2.3.0@rails5/gems/bundler-1.13.3/exe/bundle exec spring binstub --all`

- What did you expect to happen?

  I expected Bundler to install the binstubs for a newly created rails5 project.

- What happened instead?

  Instead, what happened was the following error.

Error details

    NoMethodError: undefined method `start_with?' for nil:NilClass
      /home/ubuntu/.rvm/gems/ruby-2.3.0@rails5/gems/bundler-1.13.3/lib/bundler/cli/exec.rb:95:in `block in ruby_shebang?'
      /home/ubuntu/.rvm/gems/ruby-2.3.0@rails5/gems/bundler-1.13.3/lib/bundler/cli/exec.rb:95:in `any?'
      /home/ubuntu/.rvm/gems/ruby-2.3.0@rails5/gems/bundler-1.13.3/lib/bundler/cli/exec.rb:95:in `ruby_shebang?'
      /home/ubuntu/.rvm/gems/ruby-2.3.0@rails5/gems/bundler-1.13.3/lib/bundler/cli/exec.rb:26:in `run'
      /home/ubuntu/.rvm/gems/ruby-2.3.0@rails5/gems/bundler-1.13.3/lib/bundler/cli.rb:332:in `exec'
      /home/ubuntu/.rvm/gems/ruby-2.3.0@rails5/gems/bundler-1.13.3/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
      /home/ubuntu/.rvm/gems/ruby-2.3.0@rails5/gems/bundler-1.13.3/lib/bundler/vendor/thor/lib/thor/invocation.rb:126:in `invoke_command'
      /home/ubuntu/.rvm/gems/ruby-2.3.0@rails5/gems/bundler-1.13.3/lib/bundler/vendor/thor/lib/thor.rb:359:in `dispatch'
      /home/ubuntu/.rvm/gems/ruby-2.3.0@rails5/gems/bundler-1.13.3/lib/bundler/cli.rb:20:in `dispatch'
      /home/ubuntu/.rvm/gems/ruby-2.3.0@rails5/gems/bundler-1.13.3/lib/bundler/vendor/thor/lib/thor/base.rb:440:in `start'
[...]

PR #5100, opened 2016-10-19
Warn if executable in bundle exec is empty

bronzdoc:
closes https://github.com/bundler/bundler/issues/5084

### DIFF
--- a/lib/bundler/cli/exec.rb
+++ b/lib/bundler/cli/exec.rb
@@ -91,6 +91,12 @@ module Bundler
         "#!/usr/bin/env jruby\n",
         "#!#{Gem.ruby}\n",
       ]
+
+      if File.zero?(file)
+        Bundler.ui.warn "#{file} is empty"
+        return false
+      end
+
       first_line = File.open(file, "rb") {|f| f.read(possibilities.map(&:size).max) }
       possibilities.any? {|shebang| first_line.start_with?(shebang) }
     end
--- a/spec/commands/exec_spec.rb
+++ b/spec/commands/exec_spec.rb
@@ -514,6 +514,21 @@ describe "bundle exec" do
       end
     end
 
+    context "the executable is empty" do
+      let(:executable) { "" }
+
+      let(:exit_code) { 0 }
+      let(:expected) { "#{path} is empty" }
+      let(:expected_err) { "" }
+      if LessThanProc.with(RUBY_VERSION).call("1.9")
+        # Kernel#exec in ruby < 1.9 will raise Errno::ENOEXEC if the command content is empty,
+        # even if the command is set as an executable.
+        pending "Kernel#exec is different"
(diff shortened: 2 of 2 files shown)


segiddins:
👍 with a test, thanks!

bronzdoc:
Done @segiddins

bronzdoc:
Weird, all the tests pass in my local environment

bronzdoc:
This is failing in ruby 1.8 :unamused:

spec/commands/exec_spec.rb
```diff
@@ -514,6 +514,15 @@ def bin_path(a,b,c)
       end
     end
 
+    context "the executable is empty" do
+      let(:executable) { nil }
```
segiddins: this should be an empty string instead of `nil`

lib/bundler/cli/exec.rb
```diff
@@ -91,6 +91,12 @@ def ruby_shebang?(file)
         "#!/usr/bin/env jruby\n",
         "#!#{Gem.ruby}\n",
       ]
+
+      if File.zero?(file)
+        Bundler.ui.warn "#{file} is empty\n"
```
segiddins: the explicit `\n` shouldn't be necessary

bronzdoc:
Kernel.exec in 1.8 is raising `Errno::ENOEXEC` if the file is empty even if set as an executable, I'll figure out something tonight.

bronzdoc:
@segiddins I updated this PR let me know if this changes are ok with you

lib/bundler/cli/exec.rb
```diff
@@ -86,11 +87,14 @@ def process_title(file, args)
     end
 
     def ruby_shebang?(file)
+      return false if File.zero?(file)
```
segiddins: this will cause us to check `File.zero?` twice for every file?

spec/commands/exec_spec.rb
```diff
@@ -514,6 +514,19 @@ def bin_path(a,b,c)
...
     end
 
+    context "the executable is empty" do
+      let(:executable) { "" }
+
+      let(:exit_code) { 0 }
+      let(:expected) { "#{path} is empty" }
+      let(:expected_err) { "" }
+      if LessThanProc.with(RUBY_VERSION).call("1.9")
+        pending "Kernel.exec differ"
```
segiddins: please add a comment explaining what you mean by this

bronzdoc: will do

lib/bundler/cli/exec.rb
bronzdoc: Yeah... I went back to the first implementation, sorry for the mess :grimacing:

segiddins:
Thanks! @homu r+

PR #5104, opened 2016-10-20
Fixed bundler incorrectly printing "the gem may have been yanked" in local installs

### LINKED ISSUE #5022: Local installs should not print "this gem may have been yanked"
Bundler incorrecty prints the “this gem may have been yanked” error message when trying to use a gem from the bundle before the first `bundle install`. Here’s an example, from an application with a lock, checked out on to a new machine.

```
$ bundle exec rake
Your bundle is locked to rake (10.5.0), but that version could not be found in any of the sources listed in your Gemfile. If you haven't changed sources, that means the author of rake (10.5.0) has removed it. You'll need to update your bundle to a different version of rake (10.5.0) that hasn't been removed in order to install.
```

hmistry:
Closes #5022.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -173,7 +173,7 @@ module Bundler
         rescue GemNotFound => e # Handle yanked gem
           gem_name, gem_version = extract_gem_info(e)
           locked_gem = @locked_specs[gem_name].last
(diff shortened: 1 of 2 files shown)


segiddins:
Thanks!
@homu r+

PR #5106, opened 2016-10-21
[RubygemsIntegration] Temporarily make #gem public again

### LINKED ISSUE #5102: Bundler 1.13.4 and 1.13.5 private method `gem'
In bundler prior to 1.13 there was a public `gem` method in `gems/bundler-1.11.2/lib/bundler/rubygems_integration.rb:283`

In Bundler 1.13.5 it's private, which breaks existing applications using that declaration. 

For example that was used in Rails apps in the config section

```ruby
module Foo
  class Application < Rails::Application
    config.gem 'rake'
  end
end
```

We need to make this public again. If it needs to be private we should go through a deprecation cycle.

segiddins:
Closes #5102 

\c @indirect @schneems

### DIFF
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -353,6 +353,10 @@ module Bundler
 
           true
         end
+
+        # TODO: delete this in 2.0, it's a backwards compatibility shim
+        # see https://github.com/bundler/bundler/issues/5102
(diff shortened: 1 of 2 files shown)


indirect:
would you say that the comment is a... 2.0DO?

@homu r+

segiddins:
I can release this in a patch tomorrow

PR #5119, opened 2016-10-24
Improve gem checksum mismatch error message

### LINKED ISSUE #5075: Explain how to resolve checksum mismatches
In PR https://github.com/bundler/bundler/pull/4851, we started checking .gem file checksums against the checksum provided in the compact index. However, if the checksums don't match, this is potentially a breaking change.

Based on the principle of empowering users to resolve their own errors, we should print the full path to the failing .gem, suggest deleting the gem with the bad checksum, and explain that if they are sure they want to install this gem despite the checksum not matching, they can run `bundle config disable.checksum_validaiton true` to turn off checksum checking.

The error message to be changed is inside [rubygems_gem_installer.rb](https://github.com/bundler/bundler/blob/master/lib/bundler/rubygems_gem_installer.rb#L35-L39).

hmistry:
Improve gem checksum mismatch error message by telling users how to resolve it.

Old message:

```
The checksum for the downloaded `rack-1.0.0.gem` did not match the checksum given by the API. This means that the contents of the gem appear to be different from what was uploaded, and could be an indicator of a security issue.
(The expected SHA256 checksum was "abababababababababababababababababababababab", but the checksum for the downloaded gem was "BR6Oc6Gqnq93u1aIVX4m6DvTHaFgFxdYaLUfa4ekJKI=".)
Bundler cannot continue installing rack (1.0.0).
```

New message:

```
Bundler cannot continue installing rack (1.0.0).
The checksum for the downloaded `rack-1.0.0.gem` does not match the checksum given by the server. This means the contents of the downloaded gem is different from what was uploaded to the server, and could be a potential security issue.

To resolve this issue:
1. delete the downloaded gem located at: `full-path/gems/rack-1.0.0/rack-1.0.0.gem`
2. run `bundle install`

If you wish to continue installing the downloaded gem, and are certain it does not pose a security issue despite the mismatching checksum, do the following:
1. run `bundle config disable.checksum_validaiton true` to turn off checksum verification
2. run `bundle install`

(More info: The expected SHA256 checksum was "abababababababababababababababababababababab", but the checksum for the downloaded gem was "PbfaoemSetv76QdcxuoP0ggZ0TZhMELlMMWURJoX+dw=".)
```

Closes #5075

### DIFF
--- a/lib/bundler/rubygems_gem_installer.rb
+++ b/lib/bundler/rubygems_gem_installer.rb
@@ -31,20 +31,24 @@ module Bundler
         send(checksum_type(checksum), digest)
       end
       unless digest == checksum
-        raise SecurityError,
-          "Bundler cannot continue installing #{spec.name} (#{spec.version}).\n" \
-          "The checksum for the downloaded `#{spec.full_name}.gem` does not match " \
-          "the checksum given by the server. This means the contents of the downloaded " \
-          "gem is different from what was uploaded to the server, and could be a potential security issue.\n\n" \
-          "To resolve this issue:\n" \
-          "1. delete the downloaded gem located at: `#{spec.gem_dir}/#{spec.full_name}.gem`\n" \
-          "2. run `bundle install`\n\n" \
-          "If you wish to continue installing the downloaded gem, and are certain it does not pose a " \
-          "security issue despite the mismatching checksum, do the following:\n" \
-          "1. run `bundle config disable.checksum_validaiton true` to turn off checksum verification\n" \
-          "2. run `bundle install`\n\n" \
-          "(More info: The expected SHA256 checksum was #{checksum.inspect}, but the " \
-          "checksum for the downloaded gem was #{digest.inspect}.)\n" \
+        raise SecurityError, <<-MESSAGE
+          Bundler cannot continue installing #{spec.name} (#{spec.version}).
+          The checksum for the downloaded `#{spec.full_name}.gem` does not match \
+          the checksum given by the server. This means the contents of the downloaded \
+          gem is different from what was uploaded to the server, and could be a potential security issue.
+
+          To resolve this issue:
+          1. delete the downloaded gem located at: `#{spec.gem_dir}/#{spec.full_name}.gem`
+          2. run `bundle install`
+
+          If you wish to continue installing the downloaded gem, and are certain it does not pose a \
+          security issue despite the mismatching checksum, do the following:
+          1. run `bundle config disable.checksum_validaiton true` to turn off checksum verification
+          2. run `bundle install`
+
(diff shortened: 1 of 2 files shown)


segiddins:
@indirect r?

lib/bundler/rubygems_gem_installer.rb
```diff
@@ -31,12 +31,24 @@ def validate_bundler_checksum(checksum)
...
+          the checksum given by the server. This means the contents of the downloaded \
+          gem is different from what was uploaded to the server, and could be a potential security issue.
+
+          To resolve this issue:
+          1. delete the downloaded gem located at: `#{spec.gem_dir}/#{spec.full_name}.gem`
+          2. run `bundle install`
+
+          If you wish to continue installing the downloaded gem, and are certain it does not pose a \
+          security issue despite the mismatching checksum, do the following:
+          1. run `bundle config disable.checksum_validaiton true` to turn off checksum verification
```
olleolleolle: Same typo as in the test. `checksum_validaiton` => `checksum_validation`

hmistry:
@olleolleolle Good catch! Thanks.

hmistry:
Updated PR with typo fix and rebased to master.

lib/bundler/rubygems_gem_installer.rb
```diff
@@ -31,12 +31,24 @@ def validate_bundler_checksum(checksum)
         send(checksum_type(checksum), digest)
       end
       unless digest == checksum
-        raise SecurityError,
-          "The checksum for the downloaded `#{spec.full_name}.gem` did not match " \
-          "the checksum given by the API. This means that the contents of the " \
-          "gem appear to be different from what was uploaded, and could be an indicator of a security issue.\n" \
-          "(The expected SHA256 checksum was #{checksum.inspect}, but the checksum for the downloaded gem was #{digest.inspect}.)\n" \
-          "Bundler cannot continue installing #{spec.name} (#{spec.version})."
+        raise SecurityError, <<-MESSAGE
```
hmistry: Are you suggesting to remove all the indents for lines 35 to 51 or just perform a `lstrip` on the entire string?

segiddins: `.gsub(/^ +/, "")` should do it

hmistry: Doing something like that breaks the formatting because it either removes the empty newlines required for spacing out the message for readability, or it doesn't remove the leading spaces in the paragraphs. Heredoc actually seems to handle removing the leading spaces correctly but I can't be certain of all ruby versions and platforms... the test will fail if there are leading spaces in the paragraphs.

segiddins:
@bundlerbot r+

PR #5123, opened 2016-10-24
use Bundler.which to check for otool and ldd

colby-swandale:
Hi, there is some weird behaviour in `bundler doctor`. Bundler is checking for the existence of the `ldd` and `otool` commands but is not correctly redirecting the output of `otool`. Example being on the latest version of bundler:

```
› bundle doctor
llvm-otool(1): Apple Inc. version cctools-895
llvm-otool(1): Apple Inc. version cctools-895
llvm-otool(1): Apple Inc. version cctools-895
llvm-otool(1): Apple Inc. version cctools-895
llvm-otool(1): Apple Inc. version cctools-895
llvm-otool(1): Apple Inc. version cctools-895
```
 This is happening on Mac OS (10.12). I have refactored the check of these tool using the built in `command` function, this lets us check for the existence of these commands without actually having to execute them.

Thanks.

### DIFF
--- a/lib/bundler/cli/doctor.rb
+++ b/lib/bundler/cli/doctor.rb
@@ -14,11 +14,11 @@ module Bundler
     end
 
     def otool_available?
-      system("otool --version 2>#{Bundler::NULL} >#{Bundler::NULL}")
+      system("command -v otool >#{Bundler::NULL}")
     end
 
     def ldd_available?
-      !system("ldd --help 2>#{Bundler::NULL} >#{Bundler::NULL}").nil?
+      system("command -v ldd >#{Bundler::NULL}")
     end
 
     def dylibs_darwin(path)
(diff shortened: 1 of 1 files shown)


segiddins:
Maybe we should just use `Bundler.which` to check for command availability?

colby-swandale:
@segiddins Thats a fair point, my only issue with `Bundler.which` is that it will be assumed that `ldd` or `otool` is actually a file in the current users `$PATH` and not an alias/function . This probably won't be an issue for 99.99% of people (it is for me) but i'm fine to use `Bundler.which` instead.

segiddins:
Aliases / functional wouldn't be executed anyways, since bundler isn't a shell program

colby-swandale:
You're right! i have made the changes to use `Bundler.which` instead.

segiddins:
Thanks! @homu r+

PR #5124, opened 2016-10-24
add missing highlights to bundler's bundle man page

colby-swandale:
Hi, I noticed on the main man page for bundler that some of the listed categories where not being formatted correctly with the rest of the document. This change adds some backticks to add the highlights. The before image is on the right and the after image is on the left.

I'm not sure if those categories were missing their highlights on purpose but let me know if they were.

screen shot 2016-10-24 at 10 09 22 pm

Thanks!

### DIFF
--- a/man/bundle.ronn
+++ b/man/bundle.ronn
@@ -27,20 +27,20 @@ We divide `bundle` subcommands into primary commands and utilities.
 
 ## PRIMARY COMMANDS
 
-* [bundle install(1)][bundle-install]:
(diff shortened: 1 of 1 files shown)


lynnco:
Yea I can't tell if they were missing the highlights on purpose on not. Hopefully someone will know! If not I'll circle around and merge this in a few days

indirect:
Totally an accident. Thanks for fixing this!

@homu r+

PR #5133, opened 2016-10-27
fix license name

korobkov:
…to allow automatic license checks.

[skip ci]

### DIFF
--- a/bundler.gemspec
+++ b/bundler.gemspec
@@ -7,7 +7,7 @@ require "bundler/version"
 Gem::Specification.new do |s|
   s.name        = "bundler"
   s.version     = Bundler::VERSION
-  s.licenses    = ["MIT"]
+  s.license     = "MIT"
   s.authors     = ["André Arko", "Samuel Giddins"]
   s.email       = ["team@bundler.io"]
   s.homepage    = "http://bundler.io"


korobkov:
And please, release new gem version for this change to have effect…

indirect:
@koraktor according to the issue that you linked, we already have a correct and valid license name to allow automated checks. why do you think we should change from one valid way of declaring our license to a different valid way of declaring our license?

korobkov:
After analizing my Gemfile with [license_finder](https://github.com/pivotal/LicenseFinder) I've found several dependencies with unknown license status:
`bin/license_finder`
```
Dependencies that need approval:
attribute_normalizer, 1.2.0, unknown
bundler, 1.11.2, unknown
bundler-audit, 0.5.0, GPLv3
json, 1.8.3, unknown
unicorn-rails, 2.2.1, unknown
xpath, 2.0.0, unknown
```
I looked through their gemspecs and found one thing they all had in common: array-based license declaration style (licenses = ['MIT']).
While more than 100 other dependencies which were properly marked as MIT-licensed had another, plain license declaration style (license = 'MIT').
This makes me think it could be fixed just by changing this style.

See https://github.com/pivotal/LicenseFinder#a-plea-to-package-authors-and-maintainers

indirect:
That sounds like a bug in license finder. It is valid and correct to declare the license of a gem using `licenses = [”MIT”]`. Why not send a PR t o license finder instead of 6 existing and popular gems who are unlikely to do a release just to change the gemspec formatting?

korobkov:
While being right sintactically it's something wrong semantically.
Even if [rubygems specification[(http://guides.rubygems.org/specification-reference/) allows both forms, I don't see a point in array form for only one license.
Semantically it should be useful for dual- and multi-licensed gems.
The question seems for me like array[0] vs array.first.

@indirect , I see you trying hardly to be right. I'm not trying to convince you that you are wrong.
I'm asking about compatibility between bundler (and other gems with array-based declaration style) and license_finder.
If you think it should be fixed on their side, please, let them know.
May be your arguments could be able to turn them to your side…

korobkov:
P.S. @indirect, you already have this for bundle project:
https://github.com/bundler/bundler/blob/master/lib/bundler/templates/newgem/newgem.gemspec.tt#L16
:)

lynnco:
@korobkov it looks like we actually **cant** merge this since you told it to skip the CI, [which we explicitly instruct](https://github.com/bundler/bundler/blob/master/DEVELOPMENT.md#submitting-pull-requests) contributors not to do.

That makes this PR invalid, so unfortunately I'm going to have to close it.

Feel free to create a new PR with (a) thorough justification as to why you are making the change and (b) a bit more respect of the time and energy of your local dev team! ^__^

PR #5139, opened 2016-10-30
Move man pages to man folder

bitboxer:
The [gem-man](https://github.com/defunkt/gem-man) gem searches for the man pages in the man folder. Sadly bundler right now uses `lib/bundler/man` for the man pages.

This pr fixes this and also creates correct names for the man pages. A man page always needs the section number in the filename.

After this is merged, I will create a PR that changes the [bundler-site](https://github.com/bundler/bundler-site) repo to search in this new folder for the documentation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,8 @@
 
 # output from ronn
 /lib/bundler/man/
+man/*
+!man/*.ronn
 
 # rspec failure tracking
 .rspec_status
--- a/Rakefile
+++ b/Rakefile
@@ -238,14 +238,15 @@ begin
   require "ronn"
 
   namespace :man do
-    directory "lib/bundler/man"
+    directory "man"
 
     sources = Dir["man/*.ronn"].map {|f| File.basename(f, ".ronn") }
     sources.map do |basename|
       ronn = "man/#{basename}.ronn"
-      roff = "lib/bundler/man/#{basename}"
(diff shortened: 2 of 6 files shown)


bitboxer:
Correction: just tried the bundle-site script and it still works with my pr . So no work needed on that repo 👍

bitboxer:
Will change the code to make it compatible with Ruby 1.8.7

lynnco:
@bitboxer might there be any additional context on this besides a 5 year old plugin?

-- although interestingly, the `directory "lib/bundler/man"` [line](https://github.com/bundler/bundler/commit/b88d8a9a3fcfd49d78cfb021c501552a3573fb30#diff-52c976fc38ed2b4e3b1192f8a8e24cffR137) was also written 5 years ago, so there might not be any more needed context

bitboxer:
The few gems that actually have man pages follow the convention that they can be found in the `man` directory. They follow it because of that gem. And I want to make sure this gem also follows them. Otherwise these nice man files can only be used when entering `bundle help`. So it boils down to "convention over configuration(tm)" 😉 .

Disclaimer: I am writing a [new gem](https://github.com/bitboxer/manpages) that also uses this directory position to discover the files and exposes them to man during the gem installation process.

segiddins:
👍 from me.
@indirect r?

indirect:
looks good, thanks!

@homu r+

PR #5144, opened 2016-10-31
change license specification in gemspec

korobkov:
for compatibility with license_finder gem, allowing automatic license checks.

Bundler already proposes the same format in it's newgem.gemspec template:
https://github.com/bundler/bundler/blob/master/lib/bundler/templates/newgem/newgem.gemspec.tt#L16
Let use it for bundler itself too.


segiddins:
@homu r+

PR #5148, opened 2016-11-03
[LockfileParser] Stop requiring strscan

segiddins:
Introduced and seemingly never used by https://github.com/bundler/bundler/commit/ba579f4a3e525fdd78d89531291450b26a009b07


segiddins:
@indirect do you think anyone will say this is a regression?

indirect:
god I hope not

@homu r+

PR #5170, opened 2016-11-14
Support :mri_24 platform

amatsuda:
Current ruby trunk is versioned as "2.4.0".

% ruby -ve 'p RUBY_VERSION'
ruby 2.4.0dev (2016-11-15 trunk 56792) [x86_64-darwin15]
"2.4.0"

We'd like to run tests on some gems against ruby-trunk or 2.4 pre-releases, so here's a support for MRI 2.4.

(This is Ruby 2.4 version of #3865)


segiddins:
Thanks! @homu r+

segiddins:
@homu retry

Issue #5183, opened 2016-11-18
DSL feature request: platforms negation


steakknife:
### Current pain

```ruby
not_jruby = %i(mri rbx mswin mingw x64_mingw).freeze # brittle
# ...
gem "whatever", "~> 1.2.3", platforms: not_jruby
```

### Suggestion

```ruby
gem "whatever", "~> 1.2.3", platforms: {not: :jruby} # or `except:`, whatever makes the most sense
```

segiddins:
For now, `platforms: VALID_PLATFORMS.-(platforms_to_exclude)`

steakknife:
Resolved by #5192

segiddins:
We should determine whether or not this is a feature we want before closing the issue.

colby-swandale:
This has been quiet for a long time. Closing.

PR #5192, opened 2016-11-20
dsl: not platforms

skull-squadron:
Fixes/implements #5183

### DIFF
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -231,6 +231,11 @@ module Bundler
     end
     alias_method :platform, :platforms
 
+    def except_platforms(*except_platforms, &block)
+      platforms(VALID_PLATFORMS - except_platforms, &block)
+    end
+    alias_method :except_platform, :except_platforms
+
     def env(name)
       old = @env
       @env = name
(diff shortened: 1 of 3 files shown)


segiddins:
@indirect please share your thoughts on this PR / the linked-to issue

indirect:
I would probably be okay with this as long as the option takes the form `all_platforms_except: [:jruby]`, and if you pass both `platforms:` and `all_platforms_except:` it raises an InvalidOption exception. How do you feel about it, @segiddins?

segiddins:
I personally feel like it's DSL bloat, and the sort of thing that can cause bugs in combination with other options, just like how `with`/`without` can get confusing

indirect:
The bugs when used in combination issue is why I'm only okay with this if combining :platform and :except_platform raises an exception. The unfortunate reality of gems that work on any platform except the one makes me think this is probably worth accepting.

segiddins:
@indirect @colby-swandale we should probably come to a decision on this

colby-swandale:
I'm on the fence about this. Is there a particular example that would highlight the need for this feature?

colby-swandale:
Do we want to make a decision on this? @indirect @segiddins

PR #5193, opened 2016-11-20
[WIP] Parallel specs

colby-swandale:
HI. This is currently a work in progress but i want to share with everyone on what progress i have made so far.

This PR is continuing on from #4345. I have modified the `Rakefile` to run the `spec:rubygems:<rubygem>` rake task in 2 different processes, each process takes a equal amount of spec files to execute (this is obviously not the best way to balance the amount of work). There are specs that are not passing and some other tests are a bit flaky as well.

This change depends on the grosser/parallel gem, I'm not sure if its ok to bring in a new dev dependency. let me know if it's not and ill see if i can work around it.

Performance wise, i can finish the suite of specs in about ~20ish mins on my 2014 Macbook. This used to take ~38ish minutes so i think this is a good start. I will try and see if i can get this to run on Travis and link the result.

There are some tasks that needs to be completed before this can be merged
- The detailed error reports from failed specs is not being printed when i press Ctrl^C
- Some specs are not passing that need to be fixed

Finally, do we only want to run parallel specs on CI? (ie if i run `rake spec:rubygems:master` on my macbook, should that work as it currently is?)

Let me know what you think.

Thanks

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -10,6 +10,8 @@ else
   File.expand_path("tmp/rubygems")
 end
 
+CONCURRENT_TESTS = 2
+
 BUNDLER_SPEC = Gem::Specification.load("bundler.gemspec")
 
 def safe_task(&block)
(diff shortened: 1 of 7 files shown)


Rakefile
```diff
@@ -130,9 +132,21 @@ begin
...
+          require "parallel"
+          spec_search = File.join("spec", "**", "*_spec.rb")
+          specs = Dir.glob(spec_search)
+          spec_groups = specs.each_slice((specs.size / 2).round).to_a
+          Parallel.each(spec_groups, :in_processes => CONCURRENT_TESTS) do |proccess_specs|
+            cmd = "bin/rspec #{proccess_specs.join(" ")} --format progress --color"
+            open("|#{cmd}", "r") do |output|
+              loop do
+                read = output.readpartial(4096)
+                $stdout.print read
```
segiddins: will this lead to output being interleaved between the different processes?

colby-swandale: Sorry for not answering this sooner. The answer is yes but the stdout of each process will be read in chunks of 4096 bytes. So it should be kept the same as it currently is except on rare cases.

PR #5216, opened 2016-11-30
[CLI::Gem] When using rspec, configure persistence file

segiddins:
aka make my life easier when working with new gems.

### DIFF
--- a/lib/bundler/templates/newgem/spec/spec_helper.rb.tt
+++ b/lib/bundler/templates/newgem/spec/spec_helper.rb.tt
@@ -1,2 +1,11 @@
-$LOAD_PATH.unshift File.expand_path("../../lib", __FILE__)
+require "bundler/setup"
 require "<%= config[:namespaced_path] %>"
+
+RSpec.configure do |config|
+  # Enable flags like --only-failures and --next-failure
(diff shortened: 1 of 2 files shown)


lib/bundler/templates/newgem/spec/spec_helper.rb.tt
```diff
@@ -1,2 +1,11 @@
...
+require "bundler/setup"
 require "<%= config[:namespaced_path] %>"
+
+RSpec.configure do |config|
+  # Enable flags like --only-failures and --next-failure
+  config.example_status_persistence_file_path = ".rspec_status"
+
+  config.expect_with :rspec do |c|
+    c.syntax = :expect
+  end
```
indirect: this doesn't seem like configuring the persistence file...

segiddins: oops, copied from the bundler specs. suppose I should've put it in another commit, but I do think it's a good idea to include it

segiddins: @indirect do you want me to back this out, or what?

indirect:
@bundlerbot r+

segiddins:
@bundlerbot r=indirect

PR #5217, opened 2016-11-30
[RubyGems] Ignore specs missing extensions

### LINKED ISSUE #5210: Regenerating missing extensions on bundle install
Assuming my bundle path is `./vendor/bundle`:
```
$ bundle config
Settings are listed in order of priority. The top value will be used.
path
Set for the current user (/Users/kir/.bundle/config): "./vendor/bundle"

$ bundle -v
Bundler version 1.13.6
$ ruby -v
ruby 2.3.3p222 (2016-11-21 revision 56859) [x86_64-darwin16]
```

Imagine that my `vendor/bundle/extensions` somehow get corrupted:

```
$ rm -rf vendor/bundle/extensions
```

## Expected

I expect that `bundle install` would recover from that state.

## Actual

Now to get `vendor/bundle/extensions` recovered, I have to completely remove `vendor/bundle`. Only then Bundler will download all the gems from the scratch and rebuild the extensions.

## Proposal

We don't have to delete all gems and download them again to just rebuild the extensions.
Should we make `bundle install` check if extentions need to be built?
I'll be happy to work on that.

@indirect

segiddins:
Closes #5210 but I'm not entirely convinced that this is necessary

### DIFF
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -71,8 +71,13 @@ module Bundler
       spec.installed_by_version = Gem::Version.create(installed_by_version)
     end
 
-    def spec_missing_extensions?(spec)
-      !spec.respond_to?(:missing_extensions?) || spec.missing_extensions?
+    def spec_missing_extensions?(spec, default = true)
+      return spec.missing_extensions? if spec.respond_to?(:missing_extensions?)
+
+      return false if spec.respond_to?(:default_gem?) && spec.default_gem?
+      return false if spec.extensions.empty?
+
+      default
     end
 
     def path(obj)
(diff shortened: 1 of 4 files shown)


indirect:
@segiddins necessary in what sense? (Also, do you have a sense of the performance impact of this change on an app with 500 gems?)

segiddins:
Necessary in the sense of requiring the extension will still work if its installed globally, or fail otherwise, even without this change. And no, didn't profile it yet. Keep in mind that it can short-circuit in an RG with stubs for gems that don't include extensions

segiddins:
This is green, fyi

indirect:
@bundlerbot r+

kirs:
Thanks @segiddins!

Issue #5134, opened 2016-10-27
Error Report


tipyn:
--- ERROR REPORT TEMPLATE -------------------------------------------------------
- What did you do?
  
  I ran the command `/usr/local/bin/bundle install --deployment --verbose --without test --without development`
- What did you expect to happen?
  
  I expected Bundler to...
- What happened instead?
  
  Instead, what happened was...

Error details

```
Errno::ENOSPC: No space left on device @ fptr_finalize - /var/www/discourse/.bundle/config
  /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.13.6/lib/bundler/settings.rb:248:in `close'
  /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.13.6/lib/bundler/settings.rb:248:in `open'
  /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.13.6/lib/bundler/settings.rb:248:in `open'
  /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.13.6/lib/bundler/settings.rb:248:in `block in set_key'
  /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.13.6/lib/bundler/shared_helpers.rb:105:in `filesystem_access'
  /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.13.6/lib/bundler/settings.rb:245:in `set_key'
  /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.13.6/lib/bundler/settings.rb:77:in `[]='
  /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.13.6/lib/bundler/cli/install.rb:230:in `normalize_settings'
  /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.13.6/lib/bundler/cli/install.rb:54:in `run'
  /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.13.6/lib/bundler/cli.rb:189:in `install'
  /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.13.6/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
  /usr/local/lib/ruby/gems/2.3.0/gems/bundler-1.13.6/lib/bundler/vendor/thor/lib/thor/invocation.rb:126:in `invoke_command'
[...]

PR #5218, opened 2016-11-30
[Env] Print as GFM

segiddins:
I don't know if this is a good idea, but it might help increase the proportion of new issues we get that are properly formatted?

### DIFF
--- a/lib/bundler/friendly_errors.rb
+++ b/lib/bundler/friendly_errors.rb
@@ -58,18 +58,32 @@ module Bundler
     def request_issue_report_for(e)
       Bundler.ui.info <<-EOS.gsub(/^ {8}/, "")
         --- ERROR REPORT TEMPLATE -------------------------------------------------------
-        - What did you do?
+        # Error Report
+
+        ## Questions
+
+        Please fill out answers to these questions, it'll help us figure out
+        why things are going wrong.
+
+        - **What did you do?**
 
           I ran the command `#{$PROGRAM_NAME} #{ARGV.join(" ")}`
 
-        - What did you expect to happen?
+        - **What did you expect to happen?**
 
           I expected Bundler to...
 
-        - What happened instead?
+        - **What happened instead?**
 
           Instead, what happened was...
 
+        - **Have you tried any solutions posted on similar issues in our issue tracker, stack overflow, or google?**
+
+          I tried...
+
+        - **Have you read our issues document, https://github.com/bundler/bundler/blob/master/ISSUES.md?**
+
+          ...
 
         ## Backtrace
 
--- a/lib/bundler/friendly_errors.rb
+++ b/lib/bundler/friendly_errors.rb
@@ -71,12 +71,14 @@ module Bundler
           Instead, what happened was...
 
 
-        Error details
+        ## Backtrace
 
-            #{e.class}: #{e.message}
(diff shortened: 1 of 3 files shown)


indirect:
A little more detail would make it much, much easier to give feedback! Can you briefly explain what this changes, and and why you thought it was worth doing? How will this change result in more properly formatted issues?

segiddins:
Makes things render as https://gist.github.com/segiddins/7389e65645a4239e8f2de614fe7cdb63, and will be immune to users losing indentation, as exhibited in issues such as https://github.com/bundler/bundler/issues/5134 (see how the `Bundler settings` section isn't properly indented, making the whole issue harder to parse visually)

hmistry:
I really like this change and your reasoning. Having a good visual format greatly helps to highlight the information and makes it easier to read. Nice work @segiddins. 

I think we can further improve the formatting to create a consistent presentation of content. I propose the following changes:
- Have a H1 header specifying what this is
- Have a H2 title for the Questions so its clear from a information hierarchy
- Bold the questions to draw attention to them and visually separate it from the answers
- The answers need to visually grouped with its question and spaced between questions in order for it to be clear what a Q+A set is.
- Reword the questions to make it more palatable... (subjective)
- Added a 4th question to suggest users to try searching before blindly posting - I'm conflicted on this because it's a tradeoff between lots of issues support vs knowing the frequency of field failures
- Change the format for Gemfile and Gemfile.lock to normal text instead of code formatted. The "random" change in header font creates a visual discontinuity. Keep it normal text either H3 or H4.

I've only posted the portions of my proposed changes below to be incorporated in the above reports.

# Error Report

## Questions 
1. **Tell us what you did that resulted in this error?**
 I ran the command `/Users/segiddins/Development/OpenSource/bundler/exe/bundle info`

2. **What were you expecting to happen?**
I expected Bundler to do...

3. **Tell us what actually happened instead?**
Instead, what happened was...

4. **Have you tried any solutions posted on similar issues in our issue tracker, stack overflow, or google?**
I tried...

## Backtrace
```
something
```

## Gemfile

### Gemfile

```ruby
source 'https://rubygems.org'
gem 'json'
```

### Gemfile.lock

```
GEM
  remote: https://rubygems.org/
  specs:
    json (2.0.2)
```

segiddins:
Done!

hmistry:
@segiddins could you update the gist with your new changes so we can see what it looks like now?

segiddins:
Gist updated

lib/bundler/friendly_errors.rb
```diff
@@ -58,25 +58,41 @@ def exit_status(error)
...
+        - **What did you expect to happen?**
 
           I expected Bundler to...
 
-        - What happened instead?
+        - **What happened instead?**
 
           Instead, what happened was...
 
+        - **Have you tried any solutions posted on similar issues in our issue tracker, stack overflow, or google?
```
hmistry: Missing ** at end of question to format correctly. See the gist.

hmistry:
Except for the format miss, it looks good to me. I look forward to seeing this change. Good work @segiddins!  😃

indirect:
woo, @bundlerbot r+

indirect:
@bundlerbot retry

PR #5220, opened 2016-11-30
[Realworld] Use VCR for network requests

### LINKED ISSUE #5099: Stub the real world tests with vcr
The [real world tests](https://github.com/bundler/bundler/tree/master/spec/realworld) make need to make requests against external servers in order to pass. This leads to issues such as the tests breaking when a gem releases a new version, among other things. To circumvent this, the tests can be stubbed with [vcr](https://github.com/vcr/vcr).

See [this PR](https://github.com/how-is/how_is/pull/66/files) for an example vcr implementation

This involves:
- [ ] adding vcr to `bundler.gemspec`
- [ ] setting up vcr in `spec_helper.rb`
- [ ] wrapping the real world tests with vcr
- [ ] adding a script to reload the casettes into `rake release`

**! Important !** I need to completely rewrite the expectations of this feature to address https://github.com/bundler/bundler/issues/5099#issuecomment-254646393. Until then, the above details will be inaccurate

segiddins:
Closes #5099


segiddins:
Re-implemented, should pass now. Let me know if you want me to commit the recorded responses

lynnco:
I'm 👍 on committing the recorded responses

segiddins:
(I'm just worried about bloating the repo history by committing them and needing to amend / re-commit several times thereafter)

indirect:
git-lfs? submodule? I’m okay with checking them in if it means we can stop hitting the internet 😛

segiddins:
Submodules are terrible 😭 especially if someone from outside the org ever makes a PR that depends on changes in it

segiddins:
@indirect lfs doesn't work on directories

mvz:
I've been involved with two projects that used VCR, and in both cases it led to problems down the line. The root cause of this, IMHO, is that the recorded blob is a black box.

segiddins:
I actually rewrote the VCR functionality we need so we store HTTP requests and responses raw

indirect:
@segiddins I think I'm fine with this... how big are the responses?

segiddins:
```
$ du -hs spec/support/artifice/vcr_cassettes
 10M	spec/support/artifice/vcr_cassettes
```

colby-swandale:
Is there any roadblocks from getting this merged? i would love to see Travis builds not timing out ❤️

segiddins:
@colby-swandale I don't think this would change anything on Travis

segiddins:
@indirect any feedback on the size of the cassettes?

indirect:
@segiddins no real idea how to deal with the responses' size, tbh. this seems better than having a suite dependency on working internet and rubygems.org being up to me 😛 if you feel like this is ready to merge, I'm okay with it.

segiddins:
Let me commit the cassettes then

segiddins:
Cassettes added!

indirect:
Argh, I thought I merged this. Sorry @segiddins, can you rebase?

segiddins:
@indirect r?

chrismo:
Haven't looked at any details, but can the cassettes be re-recorded easily? IMO, this is the biggest mistake with VCR use ... possibly the same point @mvz is making.

segiddins:
There's a rake task to re-record

indirect:
@chrismo yeah, the biggest reason we've been looking at this is that a lot of the "realworld" tests are effectively resolver problems that we couldn't figure out how to reproduce with test gems. Since they are effectively resolver tests, we want them to guard against regressions, but we also want to be able to run the suite without talking to rubygems.org, and we want the tests to guard against regressions even if the gems available on rubygems.org stop surfacing those regressions. This is the best compromise I can think of. 👍 

@bundlerbot r+

chrismo:
`'cool' * 3`

PR #5221, opened 2016-12-01
[Thor] Update to v0.19.4

segiddins:
Closes #4876

@rafaelfranca

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -85,6 +85,9 @@ module Bundler
       end
     end
 
+    # Ensure `bundle help --no-color` is valid
+    all_commands["help"].disable_class_options = false
+
     def self.handle_no_command_error(command, has_namespace = $thor_runner)
       if Bundler.feature_flag.plugins? && Bundler::Plugin.command?(command)
         return Bundler::Plugin.exec_command(command, ARGV[1..-1])
--- a/lib/bundler/vendor/thor/lib/thor.rb
+++ b/lib/bundler/vendor/thor/lib/thor.rb
@@ -1,7 +1,7 @@
 require "set"
 require "bundler/vendor/thor/lib/thor/base"
 
-class Bundler::Thor # rubocop:disable ClassLength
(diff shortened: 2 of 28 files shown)


segiddins:
It appears Thor 0.19.4 breaks `--no-color` and `help` somehow? @indirect is that something that I need to work around?

segiddins:
@rafaelfranca can you confirm whether this breaking change was intended or not?

rafaelfranca:
No breaking changes were intended. I can take a look later.

segiddins:
👌🏻 I might also take a look, just wanted to check

segiddins:
This will probably pass now

lib/bundler/cli.rb
```diff
@@ -85,6 +85,9 @@ def help(cli = nil)
       end
     end
 
+    # Ensure `bundle help --no-color` is valid
+    all_commands["help"].disable_class_options = false
```
rafaelfranca: Hmm. This is concerning. Seems like the change was to make sure help works without the required arguments, but in fact the change disable the option parsing. I'll try to fix it in a way that we don't broke option parser in the next version.

rafaelfranca: See https://github.com/erikhuda/thor/pull/545

segiddins:
@rafaelfranca I'll check out whether that PR fixes the bundler specs tomorrow

lib/bundler/cli.rb
segiddins: Confirmed that PR also keeps the Bundler specs green

indirect:
cool, thanks!

@bundlerbot r+

lib/bundler/cli.rb
rafaelfranca: Great. I'll merge it and open a new PR removing this code.

segiddins: It'd be great to get that fix (and also my io-console PR, if possible), into a release that we could pull in. Thanks so much for your work on this!

PR #5222, opened 2016-12-01
[CompactIndex] Disable when openssl is in fips mode

### LINKED ISSUE #4989: FIPS compatibility
bundler will fail when run in FIPS mode because it uses MD5:

```
$ bundle
md5_dgst.c(80): OpenSSL internal error, assertion failed: Digest MD5 forbidden in FIPS mode!
Aborted
```

Two changes are needed to fix this (mostly in the vendor code):
- It must be possible to use an alternative to `Digest::MD5.hexdigest` using a cipher that is permitted by FIPS-140-2 (e.g., `Digest::SHA256.hexdigest`).  
  - This will affect the files:
    - `lib/bundler/vendor/compact_index_client/lib/compact_index_client/cache.rb`
    - `lib/bundler/vendor/compact_index_client/lib/compact_index_client/updater.rb`
    - `lib/bundler/vendor/thor/lib/thor/runner.rb`
    - `lib/bundler/source/rubygems/remote.rb`
- `require "digest/md5"'`  should not be required by default, since that will also fail on a system with FIPS mode enabled.
  - This affects the files:
    - `lib/bundler/vendor/compact_index_client/lib/compact_index_client/cache.rb`
    - `lib/bundler/vendor/thor/lib/thor/runner.rb`

segiddins:
Should close #4989

### DIFF
--- a/lib/bundler/fetcher/compact_index.rb
+++ b/lib/bundler/fetcher/compact_index.rb
@@ -124,7 +124,7 @@ module Bundler
       def md5_available?
         begin
           require "openssl"
-          return false if OpenSSL::OPENSSL_FIPS
+          return false if defined?(OpenSSL::OPENSSL_FIPS) && OpenSSL::OPENSSL_FIPS
         rescue LoadError
           nil
         end
--- a/lib/bundler/fetcher/compact_index.rb
+++ b/lib/bundler/fetcher/compact_index.rb
@@ -3,10 +3,10 @@ require "bundler/fetcher/base"
 require "bundler/worker"
 
 module Bundler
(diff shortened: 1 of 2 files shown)


segiddins:
Of course, Ruby < 2 doesn't have that constant

trevor-vaughan:
@segiddins Validated that this change works properly on a system with FIPS enabled.

As stated in #4989, it is a LOT slower when resolving dependencies but it does work correctly which should meet the requirements of anyone in a FIPS restricted environment.

breisig:
@segiddins Any update of getting this merged?

segiddins:
@indirect r?

indirect:
@bundlerbot r+

PR #5240, opened 2016-12-08
`bundle update` post install messages

smathy:
Currently post install messages are only displayed for `bundle install` not for `bundle update`.  This PR fixes that so they're displayed for both, and includes a small refactor of some duplicated functionality between `install` and `update`

### DIFF
--- a/spec/update/gems/post_install_spec.rb
+++ b/spec/update/gems/post_install_spec.rb
@@ -11,9 +11,9 @@ describe "bundle update" do
       gem 'thin'
     G
 
-    bundle "config #{config}" if config
+    bundle! "config #{config}" if config
 
-    bundle :install
+    bundle! :install
   end
 
   shared_examples "a config observer" do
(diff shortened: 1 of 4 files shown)


spec/update/gems/post_install_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+  before do
+    gemfile <<-G
+      source "file://#{gem_repo1}"
+      gem 'rack', "< 1.0"
+      gem 'thin'
+    G
+
+    bundle "config #{config}" if config
+
+    bundle :install
```
segiddins: please use `bundle!`

segiddins:
Thank you very much @smathy! 👍  from me!

spec/update/gems/post_install_spec.rb
smathy: Oh thanks, I hadn't noticed that method.  I put it as a separate commit, not sure what the preference is here :)

smathy:
Np @segiddins, really appreciate the time you spend to make bundler great.

segiddins:
@bundlerbot r+

smathy:
Thanks again.

PR #5241, opened 2016-12-08
Prefer require_relative to simplify loading

ioquatix:
Using `require_relative` avoids modifying `$LOAD_PATH`. It's simpler and more robust.

### DIFF
--- a/lib/bundler/templates/newgem/newgem.gemspec.tt
+++ b/lib/bundler/templates/newgem/newgem.gemspec.tt
@@ -1,7 +1,5 @@
 # coding: utf-8
-lib = File.expand_path('../lib', __FILE__)
-$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
-require '<%=config[:namespaced_path]%>/version'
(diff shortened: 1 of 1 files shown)


segiddins:
This is incompatible with 1.8.7

ioquatix:
Is that a problem?

segiddins:
Yes, bundler still supports ruby 1.8.7

ioquatix:
Okay, I can understand that.

However, on the other hand, 1.8.7 has been deprecated for ages. When will bundler drop support for it?

segiddins:
When we release bundler 2.0

ioquatix:
Okay, I see there is a dev branch, would it make sense to merge this into that?

segiddins:
2-0-dev

segiddins:
I think it's a good idea! It'd also encourage people to keep the version file as simple as possible by default, which is definitely something I like!

ioquatix:
Yeah I agree with that simplicity is good especially for something gets repeated over and over again.

PR #5247, opened 2016-12-11
Add a new line for before priring BundlerError errors

colby-swandale:
This PR is fixing a problem where the warnings and errors on a `bundle install` using an invalid source are not being put on a newline correctly.
```
› bundle install
Fetching source index from https://www.google.com/

Retrying fetcher due to error (2/4): Bundler::HTTPError Could not fetch specs from https://www.google.com/
Retrying fetcher due to error (3/4): Bundler::HTTPError Could not fetch specs from https://www.google.com/
Retrying fetcher due to error (4/4): Bundler::HTTPError Could not fetch specs from https://www.google.com/Could not fetch specs from https://www.google.com/
``` 

This is happening because the new lines for the warnings are being printed before the error itself. This behaviour is happening due to another UI problem, see https://github.com/bundler/bundler/commit/2ac26bc88ba30b28aba6e11c9b2d04cdb0276a25. 

Maybe a better way of fixing this problem is to add a new line to the exception message itself rather then what i have submitted. Its a bit of a weird one.

Let me know.

### DIFF
--- a/lib/bundler/retry.rb
+++ b/lib/bundler/retry.rb
@@ -43,7 +43,10 @@ module Bundler
 
     def fail_attempt(e)
       @failed = true
-      raise e if last_attempt? || @exceptions.any? {|k| e.is_a?(k) }
+      if last_attempt? || @exceptions.any? {|k| e.is_a?(k) }
+        Bundler.ui.info "" unless Bundler.ui.debug?
+        raise e
+      end
       return true unless name
       Bundler.ui.info "" unless Bundler.ui.debug? # Add new line incase dots preceded this
       Bundler.ui.warn "Retrying #{name} due to error (#{current_run.next}/#{total_runs}): #{e.class} #{e.message}", Bundler.ui.debug?
--- a/spec/bundler/retry_spec.rb
+++ b/spec/bundler/retry_spec.rb
@@ -65,10 +65,10 @@ describe Bundler::Retry do
       end
     end
 
-    context "with debugging on" do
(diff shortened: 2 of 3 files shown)


colby-swandale:
I have just realised this is really not the best way, Other errors just look really ugly because now they have 2 new lines and i can't just add a new newline to the exception because then we end up with:

```
Retrying fetcher due to error (2/4): Bundler::HTTPError
Could not fetch specs from https://www.google.com/
Retrying fetcher due to error (3/4): Bundler::HTTPError
Could not fetch specs from https://www.google.com/
Retrying fetcher due to error (4/4): Bundler::HTTPError
Could not fetch specs from https://www.google.com/Could not fetch specs from https://www.google.com/
```

segiddins:
@colby-swandale can you share a failing test for the output you'd rather see?

colby-swandale:
@segiddins i have replaced the PR with a failing spec. I can't find an existing spec that tests an exact output, so the spec looks a bit ugly with that string. Let me know if there is there is a better way to format it.

segiddins:
@colby-swandale looks perfect to me!

segiddins:
@colby-swandale pushed my fix

segiddins:
Ugh I pushed something from another experiment accidentally, will clean that up

segiddins:
Fixed

colby-swandale:
Thanks @segiddins, it's really appreciated.

segiddins:
@bundlerbot r+

PR #5255, opened 2016-12-14
Support test-unit when create new gem

okkez:


okkez:
I've created gem with `bundle gem` sometimes, but I could not create new gem with test-unit using `bundle gem` command.

In previous version,

1. Run `bundle gem -t minitest`
1. Fix generated files by hand :cry:

or

1. Run `bundle gem -t none`
1. Add files related to test-unit and edit some generated files to use test-unit :cry:

segiddins:
We really don't want to be adding more options for `bundle gem`, and there have been several issues discussing similar changes that we have decided not to pursue

okkez:
Thank you for quick response!
Sorry for missing past discussions :bow:

Can I extend `bundle gem` command by external gem?
Does bundler have such API?
If no, can I add such API to bundler?

indirect:
@okkez sorry, we do not want to expand `bundle gem` any more. It is too much to add everything that everyone wants. But please use another tool designed for extendability, like [Ore](https://github.com/ruby-ore/ore). Thanks!

PR #5256, opened 2016-12-15
Do not fetch development dependencies

Paxa:
If I add dependency from github then bundle will fetch it's `development_dependencies`. This fix prevent it


segiddins:
Please add test coverage for this change

segiddins:
Ugh I forced push `origin/master` to `Paxa/master` which closed the PR so I can't actually push up my tests. I'll make a new PR :(

Paxa:
Thank you

PR #5260, opened 2016-12-17
Improve -rbundler performance

segiddins:
Along with https://github.com/rubygems/rubygems/pull/1801, speeds up `-rbundler` by ~5x

Profiled via

```ruby
# frozen_string_literal: true

REQUIRE_TIMES = {} # rubocop:disable Style/MutableConstant
module Kernel
  require "benchmark"
  alias_method :require_before_req, :require
  def require(path)
    ret = nil
    Benchmark.realtime do
      ret = require_before_req(path)
    end.tap do |t|
      if ret
        from = begin
                 caller[2]
               rescue
                 nil
               end
        ::REQUIRE_TIMES["#{path} : #{from}"] ||= t
      end
    end
    ret
  end
end

at_exit { puts REQUIRE_TIMES.to_a.sort_by(&:last).reverse.map {|k, v| "#{k} =>\n\t#{v}" } }
```


segiddins:
I don't know why only that one matrix entry is deadlocking 😭 I'll have to try and reproduce locally

segiddins:
Reproduced locally, it's dependent upon the specific ruby/rubygems version combo, and I have no idea why it's happening.  (╯°□°）╯︵ ┻━┻

segiddins:
This should fix it

segiddins:
this is green now, other than the FIXME

@indirect r?

indirect:
@bundlerbot r+

segiddins:
@bundlerbot retry

PR #5261, opened 2016-12-18
Fix the specs on Ruby 2.4

segiddins:


segiddins:
Ugh my regexp is wrong, needs to capture the whole line

segiddins:
@bundlerbot r+

PR #5267, opened 2016-12-21
[ParallelInstaller] Allow installing with corrupted lockfiles

segiddins:
- [x] Specs

### DIFF
--- a/lib/bundler/installer/parallel_installer.rb
+++ b/lib/bundler/installer/parallel_installer.rb
@@ -78,6 +78,8 @@ module Bundler
       [Bundler.settings[:jobs].to_i - 1, 1].max
     end
 
+    attr_reader :size
+
     def initialize(installer, all_specs, size, standalone, force)
       @installer = installer
       @size = size
--- a/lib/bundler/installer/parallel_installer.rb
+++ b/lib/bundler/installer/parallel_installer.rb
@@ -47,21 +47,24 @@ module Bundler
       # sure needed dependencies have been installed.
       def dependencies_installed?(all_specs)
         installed_specs = all_specs.select(&:installed?).map(&:name)
-        dependencies(all_specs.map(&:name)).all? {|d| installed_specs.include? d.name }
+        dependencies.all? {|d| installed_specs.include? d.name }
       end
 
       # Represents only the non-development dependencies, the ones that are
       # itself and are in the total list.
-      def dependencies(all_spec_names)
+      def dependencies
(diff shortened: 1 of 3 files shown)


indirect:
This looks good to me! :+1:

lib/bundler/installer/parallel_installer.rb
```diff
@@ -131,6 +135,12 @@ def handle_error
       raise Bundler::InstallError, errors.map(&:to_s).join("\n\n")
     end
 
+    def check_for_corrupt_lockfile
+      return unless @specs.any? {|s| s.corrupt_lockfile?(@specs) }
+      Bundler.ui.warn "Using 1 thread to install specs instead of #{@size} due to lockfile corruption" unless @size == 1
```
indirect: Let's merge the warning in `corrupt_lockfile?` into this one, with wording something like:

> Your Gemfile.lock was created by an old Bundler that left some things out. Because of the missing `DEPENDENCIES`, we can only install gems one at a time, instead of installing #{@size} at a time.
> You can fix this by adding the missing gems to your `Gemfile`, running `bundle install`, and then removing the gems from your `Gemfile`. The missing gems are: #{missing.map(&:name).join(", ").

indirect: That will also mean returning the array of missing names from the method above, I think?

segiddins:
@indirect except it's not actually a bug with the old bundler but rather with rubygems.org...

segiddins:
That changed eating also means we can't print which gem has the dependencies causing the problems.

indirect:
@segiddins I didn't say "an old Bundler that had a bug", I said "an old Bundler that left some things out". I didn't also include "because of a bug in RubyGems.org" because that doesn't make anything about the problem clearer.

Moving the warning up into the check so we can keep the context is fine, just merge the warnings into one place. 😛

segiddins:
Does this look like a good warning?

```
Your lockfile was created by an old Bundler that left some things out.
Because of the missing DEPENDENCIES, we can only install gems one at a time, instead of installing 9 at a time.
You can fix this by adding the missing gems to your Gemfile, running bundle install, and then removing the gems from your Gemfile.
The missing gems are:
* archive-tar-minitar, hoe, rcov depended upon by mime-types
```

indirect:
This looks great.

indirect:
@bundlerbot r+

trevorturk:
Thank you @segiddins 💝 🎉

PR #5270, opened 2016-12-22
Use correct hook name in comment

pixeltrix:
I checked `Bundler::Source::Path::Installer#post_install` - it's definitely `pre_install` and not `post_install` twice though I wasn't sure given the method name 😄


segiddins:
@bundlerbot r+

segiddins:
@bundlerbot retry

PR #5276, opened 2016-12-24
Document 1.14 platform changes

segiddins:
See https://github.com/bundler/bundler/issues/4853

@indirect feel free to add any other docs you think might be helpful to this branch

### DIFF
--- a/man/bundle-config.ronn
+++ b/man/bundle-config.ronn
@@ -156,13 +156,19 @@ learn more about their operation in [bundle install(1)][bundle-install].
 * `timeout` (`BUNDLE_TIMEOUT`):
    The seconds allowed before timing out for network requests. Defaults to `10`.
 * `force_ruby_platform` (`BUNDLE_FORCE_RUBY_PLATFORM`):
-   Force bundler to install the `ruby` platform version of gems, instead of
-   platform-specific gems.
+   Ignore the current machine's platform and install only `ruby` platform gems.
+   As a result, gems with native extensions will be compiled from source.
 * `specific_platform` (`BUNDLE_SPECIFIC_PLATFORM`):
    Allow bundler to resolve for the specific running platform and store it in
    the lockfile, instead of only using a generic platform.
+   A specific platform is the exact platform triple reported by
+   `Gem::Platform.local`, such as `x86_64-darwin-16` or `universal-java-1.8`.
+   On the other hand, generic platforms are those such as `ruby`, `mswin`, or
+   `java`. In this example, `x86_64-darwin-16` would map to `ruby` and
+   `universal-java-1.8` to `java`.
 * `disable_checksum_validation` (`BUNDLE_DISABLE_CHECKSUM_VALIDATION`):
-   Ignore the checksums of downloaded gems returned by rubygems sources.
+   Allow installing gems even if they do not match the checksum provided by
+   RubyGems.
 
 In general, you should set these settings per-application by using the applicable
 flag to the [bundle install(1)][bundle-install] or [bundle package(1)][bundle-package] command.
--- a/man/bundle-config.ronn
+++ b/man/bundle-config.ronn
@@ -155,6 +155,9 @@ learn more about their operation in [bundle install(1)][bundle-install].
    The number of redirects allowed for network requests. Defaults to `5`.
 * `timeout` (`BUNDLE_TIMEOUT`):
    The seconds allowed before timing out for network requests. Defaults to `10`.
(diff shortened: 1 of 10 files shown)


man/bundle-config.ronn
```diff
@@ -155,6 +155,14 @@ learn more about their operation in [bundle install(1)][bundle-install].
    The number of redirects allowed for network requests. Defaults to `5`.
 * `timeout` (`BUNDLE_TIMEOUT`):
    The seconds allowed before timing out for network requests. Defaults to `10`.
+* `force_ruby_platform` (`BUNDLE_FORCE_RUBY_PLATFORM`):
+   Force bundler to install the `ruby` platform version of gems, instead of
+   platform-specific gems.
+* `specific_platform` (`BUNDLE_SPECIFIC_PLATFORM`):
+   Allow bundler to resolve for the specific running platform and store it in
+   the lockfile, instead of only using a generic platform.
```
indirect: What is a specific running platform? What is a generic platform? What distinguishes them? Can we list them, or at least give examples of each type?

man/bundle-config.ronn
```diff
@@ -155,6 +155,14 @@ learn more about their operation in [bundle install(1)][bundle-install].
    The number of redirects allowed for network requests. Defaults to `5`.
 * `timeout` (`BUNDLE_TIMEOUT`):
    The seconds allowed before timing out for network requests. Defaults to `10`.
+* `force_ruby_platform` (`BUNDLE_FORCE_RUBY_PLATFORM`):
+   Force bundler to install the `ruby` platform version of gems, instead of
+   platform-specific gems.
```
indirect: I feel like it's valuable to let the user know what the results of this setting will be. How do you feel about saying something like "Ignore the current machine's platform and install only `ruby` platform gems. As a result, gems with native extensions will be compiled from source."

man/bundle-config.ronn
```diff
@@ -155,6 +155,14 @@ learn more about their operation in [bundle install(1)][bundle-install].
...
+* `specific_platform` (`BUNDLE_SPECIFIC_PLATFORM`):
+   Allow bundler to resolve for the specific running platform and store it in
+   the lockfile, instead of only using a generic platform.
+* `disable_checksum_validation` (`BUNDLE_DISABLE_CHECKSUM_VALIDATION`):
+   Ignore the checksums of downloaded gems returned by rubygems sources.
```
indirect: Is the only practical effect of this setting to allow installing gems even if they do not match the checksum provided by RubyGems? If so, let's say that instead. We want to document the results of the settings whenever we can. 👍

man/bundle-lock.ronn
```diff
@@ -78,6 +79,14 @@ For instance, you only want to update `nokogiri`, run `bundle lock --update noko
...
 gems that you specified locked to the versions in the `Gemfile.lock`.
 
+## SUPPORTING OTHER PLATFORMS
+
+If you want your bundle to support platforms other than the one you're running
+locally, you can run `bundle lock --add-platform PLATFORM` to add PLATFORM to
+the lockfile, force bundler to re-resolve and consider the new platform when
+picking gems, all without needing to have a machine that matches PLATFORM handy
+to install those platform-specific gems on.
+
```
indirect: Can we list all of the possible values for `PLATFORM` here, or at least provide a link to a full list?

segiddins: It's anything that's a valid `Gem::Platform`, so I guess just link to those docs?

man/bundle-config.ronn
segiddins: 👍

segiddins:
Attempted to address the feedback

spec/quality_spec.rb
```diff
@@ -191,6 +191,58 @@ def check_for_specific_pronouns(filename)
     expect(error_messages.compact).to be_well_formed
   end
 
+  it "documents all used settings" do
+    exemptions = %w(
```
segiddins: @indirect do you think any of these should be documented right now? If so, I'm happy to do it

segiddins:
Done.
@bundlerbot r=indirect

spec/quality_spec.rb
indirect: I am okay with none of them being documented right now. I would like to document all of them eventually, so we should probably create a meta-ticket that lists all of them so that we can track getting them documented.

segiddins: yup, as soon as this PR gets into master I was planning on making an issue linking to the list

segiddins:
@bundlerbot retry

PR #5277, opened 2016-12-24
Don't include dev deps in Index#dependency_names

segiddins:
See #5256

### DIFF
--- a/lib/bundler/index.rb
+++ b/lib/bundler/index.rb
@@ -109,15 +109,18 @@ module Bundler
 
     # returns a list of the dependencies
     def unmet_dependency_names
-      names = dependency_names
-      names.delete_if {|n| n == "bundler" }
-      names.select {|n| search(n).empty? }
+      dependency_names.select do |name|
+        name != "bundler" && search(name).empty?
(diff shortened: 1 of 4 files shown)


segiddins:
@indirect r?

indirect:
LGTM. 👍 Why do we need this change? What issue does it resolve?

lib/bundler/index.rb
```diff
@@ -109,14 +109,19 @@ def each(&blk)
 
     # returns a list of the dependencies
     def unmet_dependency_names
-      names = dependency_names
-      names.delete_if {|n| n == "bundler" }
-      names.select {|n| search(n).empty? }
+      dependency_names.select do |name|
+        name != "bundler" && search(n).empty?
```
Paxa: Tests fails because can't fine variable `n`

segiddins:
@indirect it just potentially saves downloading gems

segiddins:
@bundlerbot r+

segiddins:
@bundlerbot retry

Paxa:
Thank you

PR #5279, opened 2016-12-24
[GemInstaller] Allow installing extensions in parallel

### LINKED ISSUE #4684: Remove the global lock around building gems
> We use a global monitor to synchronize building gems. In older versions of RubyGems, build options were set in a global variable, so we have to do that to make installing gems thread-safe.
> 
> In more recent versions of RubyGems, those options can be passed to each individual `Gem::Install` instance. On those versions of RubyGems, we should be able to remove the global lock and allow multiple gems to be built at the same time. Test carefully, because we may also need to fix thread-safety bugs in `Gem::Install` for this to work.

segiddins:
I believe this closes #4684

\c @lynnco

### DIFF
--- a/lib/bundler/installer/gem_installer.rb
+++ b/lib/bundler/installer/gem_installer.rb
@@ -52,12 +52,12 @@ module Bundler
     end
 
     def install
-      spec.source.install(spec, :force => force, :ensure_builtin_gems_cached => standalone)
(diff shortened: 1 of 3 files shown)


segiddins:
@indirect r?

indirect:
@bundlerbot r+

PR #5297, opened 2017-01-01
[GitProxy] Support branches containing shell metacharacters

### LINKED ISSUE #5295: branch named with leading #hash signs aren't bundle-able
Faraday Middleware uses a branch naming convention for PRs of "#ISSUE/fix-name", which just led me to find something unexpected when pointing my Gemfile to a patch. If I use `:branch => ‘#149/redirect-url-fragment’` in my Gemfile, Bundler (1.13.7) errors with 

    fatal: Needed a single revision
    Revision #149/redirect-url-fragment does not exist in the repository github.com/lostisland/far….
    Maybe you misspelled it?

If I instead point directly to `:ref => '6e04ea6'`, the SHA of the only commit on that branch, Bundler bundles normally.

To isolate the error, I tried branch names with just a slash, just a hash, and the hash in different places. Slashes in branch names worked for me, and putting the hash at the end of the branch name also worked, but having a hash at the beginning of the branch name seems to confuse bundler:

    # THESE BRANCH NAMES WILL BUNDLE:
    gem 'faraday_middleware', :git => 'https://github.com/jm3/faraday_middleware', 
      :branch => 'foo/bar'
    gem 'faraday_middleware', :git => 'https://github.com/jm3/faraday_middleware', 
      :branch => 'hash#'
    
    # THESE BRANCH NAMES WILL NOT BUNDLE:
    gem 'faraday_middleware', :git => 'https://github.com/jm3/faraday_middleware', 
      :branch => '#hash'
    gem 'faraday_middleware', :git => 'https://github.com/lostisland/faraday_middleware', 
      :branch => '#149/redirect-url-fragment'

It's not a huge issue, but I like the clear naming convention that Faraday Middleware's using, so maybe this could be supported in a future release. The above are valid branch names, btw, if you want to test this.

segiddins:
Closes #5295

### DIFF
--- a/lib/bundler/source/git/git_proxy.rb
+++ b/lib/bundler/source/git/git_proxy.rb
@@ -180,7 +180,7 @@ module Bundler
 
         def find_local_revision
           allowed_in_path do
-            git("rev-parse --verify #{ref}", true).strip
+            git("rev-parse --verify '#{ref}'", true).strip
           end
         end
 
--- a/spec/install/gemfile/git_spec.rb
+++ b/spec/install/gemfile/git_spec.rb
@@ -209,6 +209,37 @@ describe "bundle install with git sources" do
     end
   end
 
+  describe "when specifying a branch" do
+    let(:branch) { "branch" }
+    let(:repo) { build_git("foo").path }
+    before(:each) do
+      update_git("foo", :path => repo, :branch => branch)
+    end
+
+    it "works" do
+      install_gemfile <<-G
+        git "#{repo}", :branch => #{branch.dump} do
+          gem "foo"
+        end
+      G
+
+      expect(the_bundle).to include_gems("foo 1.0")
+    end
+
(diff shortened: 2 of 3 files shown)


indirect:
What if the branch name contains a single quote? Why are we quoting user input instead of using `ShellWords.escape` here?

segiddins:
@indirect isn't that only 1.9+?

indirect:
@segiddins doesn't seem like it? http://www.rubydoc.info/stdlib/shellwords/1.8.7/Shellwords

indirect:
@segiddins looks fine to me:

```
$ chruby 1.8.7
$ irb
irb(main):001:0> require 'shellwords'
=> true
irb(main):003:0> Shellwords.shellescape "foo'bar"
=> "foo\\'bar"
```

segiddins:
@indirect r?

indirect:
@bundlerbot r+

PR #5306, opened 2017-01-05
Bundler config parseable flag

colby-swandale:
This PR is continuing on from #4919. This adds an option to `bundle config` called `parseable` that prints the value of config without the verbose. Example:

    $ bundle config foo --parseable
    bar

Closes #4919


lib/bundler/cli/config.rb
```diff
@@ -32,6 +32,8 @@ def run
         return
       end
 
+      return Bundler.ui.info(Bundler.settings[name]) if options[:parseable] && args.empty?
```
segiddins: This should go in the existing `args.empty?` conditional

segiddins:
`bundle config --parseable` still isn't parseable

segiddins:
Maybe something like a TSV or CSV? just some sort of structured output. otherwise, using `--parseable` in those circumstances should raise an exception

chrismo:
bundle outdated has a parseable option ... i don't think it's either TSV or CSV, just w/o headers and maybe fixed-width? Can't recall off-hand, but probably good to go for consistency between the two?

segiddins:
Sounds reasonable!

segiddins:
Ping on this?

colby-swandale:
Sorry! I've been working on this on and off. Ill give this attention this week.

colby-swandale:
So i have been continuing to give this some thought and i have come to the conclusion that for `bundle config --parseable` the output will need to be something on the lines of:

```
myconfig: local=foo global=bar
myconfig2: local=foo global=bar
```

A few reasons why:
 - compatibility and no surprise with the current output of `bundle config`
 - 1 config per line which works the best for utilities like grep, awk, sed etc

segiddins:
For parseable, would it maybe be better to only print the setting with the highest priority that bundler will actually use?

colby-swandale:
@segiddins yes absolutely but the current behaviour for `bundle config` prints each config value separately which i don't want to deviate from ie:
```
Settings are listed in order of priority. The top value will be used.

foo
Set for your local app (/Users/c/Desktop/Projects/testapp/.bundle/config): "bar2"
Set via BUNDLE_FOO: "bar4"
Set for the current user (/Users/c/.bundle/config): "bar3"
```

I wanted to keep the behaviour of `--parseable` in line with the output `bundle config` and not surprise users with a completely different piece of behaviour on top what's currently there.

For Bundler 2 i was thinking we could drop the "verbosity" of `bundle config` to make it behave like `--parseable` and only show the config with the highest priority, ie:
```
> bundler2 config
foo=bar
```

segiddins:
Given `bundle config --parseable` doesn't exist yet, we can choose the best behavior for it, rather than deferring to existing output

colby-swandale:
@indirect can i just get your thoughts about the past few comments to see if you agree with @segiddins ?

indirect:
Yeah, I think for machine-readable output the useful thing to print is the actual settings! And since it's a new feature, I'm happy to jump straight to the ideal behavior. 👍

colby-swandale:
thanks @indirect, @segiddins ill update that behaviour then.

colby-swandale:
well nothing different from existing behaviour, i have a test for this at https://github.com/bundler/bundler/pull/5306/files#diff-0bbd2388477a861ecd2eb9aed2ff40a7R140

spec/commands/config_spec.rb
```diff
@@ -102,6 +136,12 @@
       run "puts Bundler.settings['local.foo']"
       expect(out).to eq(File.expand_path(Dir.pwd + "/.."))
     end
+
+    it "saves with parseable option" do
+      bundle "config --global --parseable foo value"
```
colby-swandale: What output are you expecting?

spec/commands/config_spec.rb
```diff
@@ -147,6 +187,13 @@
       run "puts Bundler.settings['local.foo']"
       expect(out).to eq(File.expand_path(Dir.pwd + "/.."))
     end
+
+    it "can be deleted with parseable option" do
+      bundle "config --local foo value"
+      bundle "config --delete --parseable foo"
```
colby-swandale: What output are you expecting?

spec/commands/config_spec.rb
segiddins: probably the old value, since that's what we print in the non-parsable case?

spec/commands/config_spec.rb
segiddins: probably the old value, since that's what we print in the non-parsable case?

colby-swandale:
```
› bundle config
Settings are listed in order of priority. The top value will be used.
foo
Set for the current user (/Users/c/.bundle/config): "bar"

› bundle config --delete foo
```

i can't see Bundler printing any output when i delete a config

segiddins: 👍 in that case, let's test that the output is empty

spec/commands/config_spec.rb
segiddins: (i.e. it's `You are replacing the current local value of foo, which is currently nil`)

colby-swandale: Yep, gotcha!

spec/commands/config_spec.rb
colby-swandale: 👍

segiddins:
@indirect r?

spec/commands/config_spec.rb
```diff
@@ -102,6 +136,22 @@
...
+      bundle "config --global --parseable foo value"
+      run "puts Bundler.settings['foo']"
+      expect(out).to eq("value")
+    end
+
+    context "when replacing a current value with the parseable flag" do
+      before { bundle "config --global foo value" }
+      it "prints the current value in a parseable format" do
+        bundle "config --global --parseable foo value2"
+        expect(out).to eq "foo=value"
```
segiddins: why does this print the key?
`bundle config --parseable foo` doesn't print the key, only the value

indirect:
I think maybe in the interests of consistency (and parsability), every single output from `--parsable` should be the same format? @colby-swandale @segiddins, how do you feel about always printing `key=value`, and treating any passed key names as filters on the list that is printed, rather than changing the output? Similarly, I think I would expect setting a new value with `--parsable` to result in printing `key=newvalue`, as confirmation that the value was set.

segiddins:
> Similarly, I think I would expect setting a new value with --parsable to result in printing key=newvalue, as confirmation that the value was set

this is different from what `bundle config` does now, but seems reasonable?

> every single output from --parsable should be the same format

yeah, I can get behind this

segiddins:
```diff
diff --git a/lib/bundler/cli/config.rb b/lib/bundler/cli/config.rb
index 8ad7f12fa..ff031899a 100644
--- a/lib/bundler/cli/config.rb
+++ b/lib/bundler/cli/config.rb
@@ -33,7 +33,12 @@ module Bundler
       end
 
       if args.empty?
-        return Bundler.ui.info(Bundler.settings[name]) if options[:parseable]
+        if options[:parseable]
+          if value = Bundler.settings[name]
+            Bundler.ui.info("#{name}=#{value}")
+          end
+          return
+        end
 
         confirm(name)
         return
diff --git a/spec/commands/config_spec.rb b/spec/commands/config_spec.rb
index 030a85ff6..33d43b2a8 100644
--- a/spec/commands/config_spec.rb
+++ b/spec/commands/config_spec.rb
@@ -139,6 +139,7 @@ RSpec.describe ".bundle/config" do
 
     it "saves with parseable option" do
       bundle "config --global --parseable foo value"
+      expect(out).to eq("")
       run "puts Bundler.settings['foo']"
       expect(out).to eq("value")
     end
@@ -256,14 +257,14 @@ RSpec.describe ".bundle/config" do
       bundle "config foo local"
       bundle "config foo --parseable"
 
-      expect(out).to eq "local"
+      expect(out).to eq "foo=local"
     end
 
     it "can print global config" do
       bundle "config --global bar value"
       bundle "config bar --parseable"
 
-      expect(out).to eq "value"
+      expect(out).to eq "bar=value"
     end
 
     it "preferes local config over global" do
@@ -271,7 +272,7 @@ RSpec.describe ".bundle/config" do
       bundle "config --global bar value"
       bundle "config bar --parseable"
 
-      expect(out).to eq "value2"
+      expect(out).to eq "bar=value2"
     end
   end
 
```

segiddins:
@indirect I think that does what you want?

indirect:
Not positive, but the tests make it look like it does? @colby-swandale, feel free to add this to your PR and then show us some example invocations and output!

colby-swandale:
```
› dbundler config --parseable foo bar
foo=bar
› dbundler config --parseable foo bar2
foo=bar2
› dbundler config --parseable foo
foo=bar2
› dbundler config --parseable foo1 bar
foo1=bar
› dbundler config --parseable
foo=bar2
foo1=bar
› dbundler config --parseable --delete foo
› dbundler config --parseable foo
```

indirect:
looks good to me! 👍  do any of you have an idea of whether or not we should exit > 0 if the user tries to get an unset config? or should we exit 0 with no output?

segiddins:
Exit 0 with no output seems reasonable to me?

colby-swandale:
i agree, exit 0 with empty output.

colby-swandale:
@segiddins r?

segiddins:
@bundlerbot r+

PR #5308, opened 2017-01-07
Call out the costs and ask profitable Rubyists to give back

indirect:


bundler.gemspec
```diff
@@ -32,4 +32,6 @@ Gem::Specification.new do |s|
   s.bindir        = "exe"
   s.executables   = %w(bundle bundler)
   s.require_paths = ["lib"]
+
+  s.post_install_message = "Bundler and RubyGems.org are free for anyone to use, but maintaining them costs more than $20,000 USD every month. If you make a profit using Ruby, help us cover those costs so that we can keep the gem ecosystem free for everyone: https://rubytogether.org/join"
```
segiddins: Line is too long

hmistry: I feel the message is too forward and could use a little more tact. Something like "Bundler and RubyGems.org is maintained through funds raised by Rubytogether.org. Please help us keep the gem ecosystem alive for you and others to use by a one time donation or becoming a member, see https://rubytogether.org/join".
It maybe a bit long but for this particular message, it is far more important to have a clear message with the right tone.

indirect: @hmistry I also prefer the more subtle type of message, but my experience when talking to Ruby devs is that they say things like "wow I had no idea it was so expensive, or that RubyGems is growing so much it is in danger unless it gets more funding". If saying "please help us" doesn't include explicit mentions of the high cost or that this funding is the only thing making it possible at all, it seems like most if not all devs will just assume everything is fine and ignore it. :/

hmistry: @indirect I totally understand where you're coming as I've followed things this year and knew what you were trying to do with that message. I agree that my example is not as striking as it could be but the tone of the current message can definitely rub people the wrong way and reflect poorly on Rubytogether.org. 
I've had a pretty bad day today, but if you're up for it, give me a couple days and see if I can help you come up with something better with a stronger appeal factor.

segiddins:
@bundlerbot r+

PR #5319, opened 2017-01-12
[Rakefile] Add release-related tasks

segiddins:
This adds some scripts that I've developed to make releasing easier to the repo, so that way they're not just sitting around in random gists. It also mitigates the risk of me accidentally `git clean -dfx`-ing an uncommitted script in `bin`, which I've totally done before.


indirect:
hooray @bundlerbot r+

PR #5321, opened 2017-01-12
Adds *.gem to the default git ignore for a new gem.

sblackstone:
Gemspecs by default use

     spec.files         = `git ls-files -z`.split("\x0").reject do |f|

So, if you create a gem file it will get sucked into spec.files.

Suppose you then try to use this development gem from another project in a Gemfile using path:

    gem 'thingy', path: /some/other/dir

`bundle install` then fails because the Gemfile itself is being included.

### DIFF
--- a/lib/bundler/templates/newgem/gitignore.tt
+++ b/lib/bundler/templates/newgem/gitignore.tt
@@ -7,6 +7,7 @@
 /pkg/
 /spec/reports/
 /tmp/
+*.gem
 <%- if config[:ext] -%>
 *.bundle
 *.so


segiddins:
Given `pkg` is already being ignored, I'm not sure what this change is really buying is?

sblackstone:
Perhaps I'm just using antiquated techniques?   I'm building gems with gem build whatever.gemspec and then manually pushing them a geminabox server.

Is there a different build command such that the binary ends up being created in pkg/ ?

simi:
@sblackstone there are [gem rake tasks](https://github.com/bundler/bundler/blob/4be855932695fc1fbff677687e7e54102982083d/lib/bundler/gem_tasks.rb) being required in default generated [Rakefile](https://github.com/bundler/bundler/blob/4be855932695fc1fbff677687e7e54102982083d/lib/bundler/templates/newgem/Rakefile.tt#L1). You can run `rake -T` to see all tasks. To build gem, you can use `rake build` AFAIK.

sblackstone:
Ahh-ha. 

Ok then, this patch is unneeded!

Thanks..
Stephen

PR #5331, opened 2017-01-15
Re-work RSpec configuration to be more CI-friendly

segiddins:
- Fail fast when there are a bunch of failures. Especially since holding all of the failure output in memory makes the tests hella slow
- Disable RSpec monkey patching
- Disable the same by default when generating new gems


indirect:
@bundlerbot r+

PR #5335, opened 2017-01-19
[SandaloneInstaller] Warn if different ruby used

brugidou:
The goal is to avoid confusion when user is running
`bundle/bundler/setup.rb` with a different `ruby_version` than the one
used to install gems.

I just did a rough patch. I would like to discuss to get something cleaner since:

* the `setup.rb` file is currently version agnostic, maybe some users are installing gems with different ruby versions in `bundle/ruby/` so that their app can run with multiple ruby versions?
* a warning in the output might not be the cleanest solution, failure is a possibility.
* we could also test for the existence of directoris for all the paths in `setup.rb`, I am afraid that this might cause issues or unnecessary warnings.

### DIFF
--- a/lib/bundler/installer/standalone.rb
+++ b/lib/bundler/installer/standalone.rb
@@ -14,6 +14,11 @@ module Bundler
         file.puts "# ruby 1.8.7 doesn't define RUBY_ENGINE"
         file.puts "ruby_engine = defined?(RUBY_ENGINE) ? RUBY_ENGINE : 'ruby'"
         file.puts "ruby_version = RbConfig::CONFIG[\"ruby_version\"]"
+        file.puts %(install_ruby_version = "#{RbConfig::CONFIG["ruby_version"]}")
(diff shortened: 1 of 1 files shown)


segiddins:
I'm not sure I understand what problem this is solving, can you please explain what issue this PR is addressing?

colby-swandale:
ping @brugidou

brugidou:
I want to avoid confusion when the setup.rb file is run with a different ruby version than the one used during the `bundle install --standalone`.

The use case I have in mind is that I use `bundle install --standalone` for myapp, package the whole directory in a tarball and ship it to a server to run my app with `ruby -r ./bundle/bundler/setup.rb bin/myarr`.

If the server is running a different ruby version than the one I used to build the tarball, it won't work and it's hard to debug.

segiddins:
In that case, I think you should be adding the ruby version you need to your gemfile -- by default, bundles should be able to work across multiple ruby versions

brugidou:
OK I understand

PR #5336, opened 2017-01-19
[ParallelInstaller] Print out why a gem needed to be installed if ins…

### LINKED ISSUE #5078: Feature request/proposal: `bundle why gem`
Idea is borrowed from newly arrived YARN package manager. Shown, for example, [in this twit](https://twitter.com/_developit/status/785872165723738116)

Idea is, something like that:

```
bundle why gemname
# output 1:
gemname 1.14 installed because required explicitly in Gemfile, group :development
# output 2:
gemname 1.14 installed because required by othergemname 2.03 (Gemfile, group :testing)
# more complicated:
gemname 1.14 installed because required by othergemname 2.03 => thirdone 0.19 => fourthgem 0.0.1 (Gemfile, group :runtime)
# and this one
gemname 1.14 installed because required by othergemname 2.03 (Gemfile, group :testing) and thirdone 0.19 (Gemfile, group :runtime)
```

Utterly naive implementation (not doing all what is proposed, just showing it is possible; also, really quickly dropped in bad Ruby) [shown here](https://gist.github.com/zverok/cb4aab711fa02ab3c4fbe13f15540921)

If an idea itself is acceptable (and I am not missing it is already possible/done), I can prepare proper PR with good design, specs and stuff.

segiddins:
…tallation fails

I think this will close #5078 by making it easier to understand why a gem needed to be installed, especially in the case when the user can't just look in the lockfile


lib/bundler/installer/gem_installer.rb
```diff
@@ -12,6 +12,7 @@ def initialize(spec, installer, standalone = false, worker = 0, force = false)
     end
 
     def install_from_spec
+      raise "install failed" if spec.name == "thor"
```
segiddins: I probably should remove this!

segiddins:
@indirect please review?

colby-swandale:
ping @indirect

indirect:
This looks good. Thanks!

@bundlerbot r+

indirect:
@segiddins ping

segiddins:
Updated the rubocop todo, should pass now. @indirect r?

indirect:
@bundlerbot r+

PR #5342, opened 2017-01-21
[SharedHelpers] Use block.call instead of yield to avoid a stack cons…

segiddins:
…istency error on Ruby 2.2.2

Using `yield` would cause a crash in the ruby VM due to calls sharing state
and leading to a count mismatch. Using block.call avoids that issue

Fixes https://github.com/bundler/bundler/issues/5341
Manually tested on Ruby 2.2.2 and 2.2.3, since they're not in the test matrix (only the latest 2.2.x is)

### DIFF
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -102,8 +102,10 @@ module Bundler
     #   end
     #
     # @see {Bundler::PermissionError}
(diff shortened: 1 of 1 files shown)


segiddins:
@bundlerbot r+

indirect:
Nice (and fast!) work. 👍

Issue #4988, opened 2016-09-15
Move some help files from bundler/bundler site to man/ ?


lynnco:
An problem illuminated by this issue https://github.com/bundler/bundler/issues/4940 is that it's very confusing whether a particular piece of documentation should be in this repo, or in bundler/bundler-site. Many of those pages might be better if they were inside of  bundler/bundler's `man/`.

Good candidates?
- https://github.com/bundler/bundler-site/blob/master/source/v1.13/bundle_clean.html.haml
- https://github.com/bundler/bundler-site/blob/master/source/v1.13/bundle_check.html.haml

segiddins:
Yes, all the command docs should be manpages

feministy:
i moved over everything that was missing but had a page in bundler site in #5348 - if y'all wouldn't mind giving it a once over that would be aces. thanks!

segiddins:
Done in https://github.com/bundler/bundler/pull/5348

PR #5348, opened 2017-01-22
create man pages for missing commands

feministy:
creates man pages for all of the commands we have on `bundler-site` that don't currently have man pages. this amounts to new pages for: 

* check
* clean
* init
* inject
* show
* open 
* viz

i followed the formatting for other man pages as best as i could, and added everything to the `index.txt` file. i previewed them all individually and they look 🙆 !

---

*notes*

* addresses #4988 
* do we need to remove the pages for the smaller commands from `bundler-site` before this is merged? generally i think the more complicated ones should stay (install, etc)

### DIFF
--- a/man/bundle-check.1.ronn
+++ b/man/bundle-check.1.ronn
@@ -1,0 +1,24 @@
+bundle-check(1) -- Verifies if dependencies are satisfied by installed gems
+===========================================================================
+
+## SYNOPSIS
+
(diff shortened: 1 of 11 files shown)


segiddins:
Spec failures: https://travis-ci.org/bundler/bundler/jobs/194303060#L416

feministy:
ohh thanks @segiddins! i will get to fixing those ASAP!

indirect:
hooray @bundlerbot r+

Issue #5349, opened 2017-01-23
Conflicting platform dependencies still not resolving.


nomoon:
Re #5340 closed as fixed in 1.14.2, I'm still getting the same error.

I seem to still be getting this error with Bundler 1.14.2 and JRuby (ruby-concurrency/thread_safe#29)

## Environment

```
Bundler   1.14.2
Rubygems  2.6.8
Ruby      2.3.1p0 (2016-11-09 revision 54768) [java]
GEM_HOME  /Users/nomoon/.rvm/gems/jruby-9.1.6.0@thread_safe
GEM_PATH  /Users/nomoon/.rvm/gems/jruby-9.1.6.0@thread_safe:/Users/nomoon/.rvm/gems/jruby-9.1.6.0@global
RVM       1.28.0 (latest)
Git       2.11.0
Platform  universal-java-1.8
rubygems-bundler (1.4.4)
```

## Bundler settings

```
build.eventmachine
  Set for the current user (/Users/nomoon/.bundle/config): "--with-cppflags=-I/usr/local/opt/openssl/include"
gem.test
  Set for the current user (/Users/nomoon/.bundle/config): "rspec"
gem.coc
  Set for the current user (/Users/nomoon/.bundle/config): "true"
gem.mit
  Set for the current user (/Users/nomoon/.bundle/config): "true"
local.readthis
  Set for the current user (/Users/nomoon/.bundle/config): "/Users/nomoon/LocalDev/readthis"
force_ruby_platform
  Set for the current user (/Users/nomoon/.bundle/config): "false"
```

### Gemfile

```ruby
source 'https://rubygems.org'

gemspec

group :development, :test do
  gem 'rspec', '~> 3.2.0'
  gem 'simplecov', '~> 0.9.2', :require => false
  if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.0.0')
    gem 'term-ansicolor', '~> 1.3.2', :require => false
    gem 'tins', '~> 1.6.0', :require => false
  end
  gem 'coveralls', '~> 0.7.11', :require => false
end

group :documentation do
  gem 'countloc', '~> 0.4.0', :platforms => :mri, :require => false
  gem 'yard', '~> 0.8.7.6', :require => false
  gem 'inch', '~> 0.5.10', :platforms => :mri, :require => false
  gem 'redcarpet', '~> 3.2.2', platforms: :mri # understands github markdown
end
```

### Gemfile.lock

```
PATH
  remote: .
  specs:
[...]

PR #5354, opened 2017-01-23
[Definition] Avoid re-resolving when a gemspec has dev deps

segiddins:
Inspired by #5349.

Since dev deps are added with `type: :development`, they are `!=` to the deps retrieved from the lockfile, which have no type. This compares the deps ignoring type completely


indirect:
@bundlerbot r+

PR #5363, opened 2017-01-24
Permissive temporary home directory

svoop:
We're deploying to a server with several Ruby applications each with it's own user. However, the home directory of these users is *not* writable by the users.

Since Bundler 1.14.x, a temporary home directoy in the `tmpdir` is used as a fallback. This works well for only one user, but will fail with more than one:

1. user1 execs `bundle`
2. `/tmp/bundler/home/user1` is created usually with 0755 modes
3. user2 execs `bundle`
4. `/tmp/bundler/home/user2` cannot be created since user2 can't write to `home`

This PR adds a `chmod` to apply mode 0777 to `home`. It rescues to nil in order to silently fail on step 4 (user2 cannot chmod `home` since it's owned by user1) as well as in case of any other problem performing chmod. Since we're in a fallback here, I guess it's okay to do this. However, if this is too much code smell, a `unless File.exist?` will do the trick as well.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -146,34 +146,37 @@ module Bundler
     def user_home
       @user_home ||= begin
         home = Bundler.rubygems.user_home
-        warning = "Your home directory is not set properly:"
-        if home.nil?
-          warning += "\n * It is not set at all"
+
+        warning = if home.nil?
+          "Your home directory is not set."
         elsif !File.directory?(home)
-          warning += "\n * `#{home}` is not a directory"
+          "`#{home}` is not a directory."
         elsif !File.writable?(home)
-          warning += "\n * `#{home}` is not writable"
+          "`#{home}` is not writable."
+        end
+
+        if warning
+          user_home = tmp_home_path(Etc.getlogin, warning)
+          Bundler.ui.warn "#{warning}\nBundler will use `#{user_home}' as your home directory temporarily.\n"
+          user_home
         else
-          return @user_home = Pathname.new(home)
+          Pathname.new(home)
         end
-
-        login = Etc.getlogin || "unknown"
-
-        tmp_home = Pathname.new(Dir.tmpdir).join("bundler", "home", login)
-        begin
-          SharedHelpers.filesystem_access(tmp_home, :write) do |p|
-            FileUtils.mkdir_p(p)
-          end
-        rescue => e
-          warning += "\n\nBundler also failed to create a temporary home directory at `#{tmp_home}`:\n#{e}"
-          raise warning
-        end
-
-        warning += "\n\nBundler will use `#{tmp_home}` as your home directory temporarily"
-
-        Bundler.ui.warn(warning)
-        tmp_home
       end
+    end
+
+    def tmp_home_path(login, warning)
+      login ||= "unknown"
+      path = Pathname.new(Dir.tmpdir).join("bundler", "home")
+      SharedHelpers.filesystem_access(path) do |tmp_home_path|
+        unless tmp_home_path.exist?
+          tmp_home_path.mkpath
+          tmp_home_path.chmod(0o777)
+        end
+        tmp_home_path.join(login).tap(&:mkpath)
+      end
+    rescue => e
+      raise "#{warning}\nBundler also failed to create a temporary home directory at `#{path}':\n#{e}"
     end
 
     def user_bundle_path
--- a/spec/bundler/bundler_spec.rb
+++ b/spec/bundler/bundler_spec.rb
@@ -170,4 +170,43 @@ EOF
       end
     end
   end
+
+  describe "#user_home" do
+    context "home directory is set" do
+      it "should return the user home" do
+        path = "/home/oggy"
+        allow(Bundler.rubygems).to receive(:user_home).and_return(path)
+        allow(File).to receive(:directory?).with(path).and_return true
+        allow(File).to receive(:writable?).with(path).and_return true
+        expect(Bundler.user_home).to eq(Pathname(path))
+      end
+    end
+
+    context "home directory is not set" do
+      it "should issue warning and return a temporary user home" do
+        allow(Bundler.rubygems).to receive(:user_home).and_return(nil)
+        allow(Etc).to receive(:getlogin).and_return("USER")
+        allow(Dir).to receive(:tmpdir).and_return("/TMP")
+        allow(FileTest).to receive(:exist?).with("/TMP/bundler/home").and_return(true)
+        expect(FileUtils).to receive(:mkpath).with("/TMP/bundler/home/USER")
+        message = <<EOF
+Your home directory is not set.
+Bundler will use `/TMP/bundler/home/USER' as your home directory temporarily.
+EOF
+        expect(Bundler.ui).to receive(:warn).with(message)
+        expect(Bundler.user_home).to eq(Pathname("/TMP/bundler/home/USER"))
+      end
+    end
+  end
+
+  describe "#tmp_home_path" do
+    it "should create temporary user home" do
+      allow(Dir).to receive(:tmpdir).and_return("/TMP")
+      allow(FileTest).to receive(:exist?).with("/TMP/bundler/home").and_return(false)
+      expect(FileUtils).to receive(:mkpath).once.ordered.with("/TMP/bundler/home")
+      expect(FileUtils).to receive(:mkpath).once.ordered.with("/TMP/bundler/home/USER")
+      expect(File).to receive(:chmod).with(0o777, "/TMP/bundler/home")
+      expect(Bundler.tmp_home_path("USER", "")).to eq(Pathname("/TMP/bundler/home/USER"))
+    end
+  end
 end
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -103,6 +103,14 @@ RSpec.configure do |config|
 
   config.before :all do
     build_repo1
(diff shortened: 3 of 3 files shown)


segiddins:
This will need test coverage

svoop:
@segiddins Can you give me a hint as to where to put the test? I've been looking for a location where the current code is covered, but couldn't find any.

segiddins:
You could put it in a unit test in spec/bundler

svoop:
@segiddins Okidoke. What do you think about the inline rescue, keep it (requires a RuboCop exception) or silence it with a slightly more verbose `exists?` or `writable?` condition?

segiddins:
You can also just do it in a `begin...rescue nil end`

svoop:
@segiddins All done: Unit tests added and passing, thanks for your hints!

segiddins:
:rocket: 

Look good to you too, @indirect ?

colby-swandale:
@segiddins Forgive my naiveness if it may apply but why couldn't we just set`tmp_home = Pathname.new(Dir.tmpdir)` without creating the sub-folders?

segiddins:
@colby-swandale excellent question, to which I don't know the answer :(

indirect:
@colby-swandale isn’t it so that settings will persist across commands? Regular tmpdirs are deleted when then process exits, right?

@segiddins I feel like there’s probably a better design for this feature that would keep this problem from existing,  but I don’t have time to discover that design today. 🚀

colby-swandale:
@segiddins i just realized that `Dir.tmpdir` on Linux is always `/tmp` so it makes sense to put everything into a sub-folder for that usecase.

@indirect are we storing settings in tmpdirs? In [this case](https://github.com/bundler/bundler/blob/master/lib/bundler.rb#L165) we're not creating a tmpdir so the folder will persist until the machine is restarted.

indirect:
@colby-swandale ahh, good point.

@svoop how would you feel about this code to create the temporary directory?

```ruby
        tmp_homes = Pathname.new(Dir.tmpdir).join("bundler", "home")
        tmp_userhome = tmp_homes.join(login)
        begin
          SharedHelpers.filesystem_access(tmp_home, :write) do |p|
            tmp_homes.mkpath
            tmp_homes.chmod(0o777)
            tmp_userhome.mkpath
          end
        [...]
```

colby-swandale:
@indirect `mkdir_p` has the option to specify a mode if that helps 

```
tmp_home = Pathname.new(Dir.tmpdir).join("bundler", "home", login)
begin
  SharedHelpers.filesystem_access(tmp_home, :write) do |p|
    FileUtils.mkdir_p(p, mode: => 0o777)
  end
...
```

svoop:
@indirect I'd be very happy with your code, however, RuboCop wouldn't – he's not a big fan of inline rescues. And I'm not sure, nested begin-rescue blocks would do readability a big favor. IS adding a RuboCop exception an option?

@colby-swandale I'd prefer not to make the whole path `0x777`. Okay, it would boil down the LOC needed, but even though this usually happens inside `/tmp`, we should only be as permissive as necessary IMO.

svoop:
@indirect Can you comment on my question in the previous comment? Thanks!

indirect:
@svoop sorry, I wasn't trying to say we should use an inline rescue, I was trying to ask about putting all of the directory creation and permission changes into the same place, using Pathname. Create a new method to hold the rescue block and it will suddenly be much more readable. :)

svoop:
@indirect Ah, okay, I'll give it a shot.

svoop:
@indirect Better [like this](https://github.com/svoop/bundler/blob/05c700282b65207338be80982d27fbfed4242c4f/lib/bundler.rb#L146)? (Don't mind the failing Ruby 1.8.7 test, I'll fix it tomorrow after I got that methuselah Ruby to install.)

---

Off-thread: Sometimes, Rubocop can be a gadfly. Personally, in this situation, I'd have preferred:

    warning = case
      when home.nil? then "Your home directory is not set."
      when !File.directory?(home) then "`#{home}` is not a directory."
      when !File.writable?(home) then "`#{home}` is not writable."
    end

Over:

    warning = if home.nil?
      "Your home directory is not set."
    elsif !File.directory?(home)
      "`#{home}` is not a directory."
    elsif !File.writable?(home)
      "`#{home}` is not writable."
    end

segiddins:
Looks good to me :)

lib/bundler.rb
```diff
@@ -146,34 +146,35 @@ def ruby_scope
...
-        warning += "\n\nBundler will use `#{tmp_home}` as your home directory temporarily"
-
-        Bundler.ui.warn(warning)
-        tmp_home
+    def tmp_home_path(login, warning)
+      login ||= "unknown"
+      path = Pathname.new(Dir.tmpdir).join("bundler", "home")
+      SharedHelpers.filesystem_access(path) do |p|
+        p.mkpath
+        p.chmod(0o777) unless p.world_writable?
```
segiddins: `world_writable?` doesn't seem to be available on 1.8.7

svoop: That's what it looks like, but [it should be](https://ruby-doc.org/stdlib-1.8.7/libdoc/pathname/rdoc/Pathname.html#method-i-world_writable-3F). (I'll have to look for my very old Mac, installing Ruby 1.8.7 on my more recent one is way too much hassle.)

segiddins: It's not listed in https://ruby-doc.org/core-1.8.7/FileTest.html though

svoop: Feels like time travel: `Pathname#world_writable?` esists in Ruby 1.8.7, but it delegates to `FileTest` where it is not defined. I'll have to uglify things further with something like `p.stat.mode`.

Just curious: Why does Bundler still support such long dead Rubies?

segiddins: Because of semver -- we will drop old rubies with Bundler 2

indirect:
Yeah, this is great, thanks! @bundlerbot r+

svoop:
@segiddins Do you have an idea where [these failures](https://travis-ci.org/bundler/bundler/builds/197754028) against Ruby 1.8.7 are coming from? I'm a little lost here in lack of a local Ruby 1.8.7 environment. (I have tried to get it going for the past two hours.)

svoop:
I finally managed to get a Ruby 1.8.7 environment working on an old Mac and I'm able reproduce the problem Travis faces. All tests pass when they are run one spec file at a time, but the entire suite fails. It seems as if `expect(FileUtils).to receive(:mkpath) ...` somehow bleeds across specs, whatever that means. :smile:

@segiddins @indirect Could this be an RSpec bug on Ruby 1.8.7? In any case, I'm running out of tricks here. Ideas welcome!

segiddins:
I'll take a look into it when I get a chance, no worries if you want to leave it here

svoop:
@segiddins Thanks! I'll gladly do the work to fix this, just need a little help here :smile:

lib/bundler.rb
```diff
@@ -146,34 +146,37 @@ def ruby_scope
...
-        tmp_home
+    def tmp_home_path(login, warning)
+      login ||= "unknown"
+      path = Pathname.new(Dir.tmpdir).join("bundler", "home")
+      SharedHelpers.filesystem_access(path) do |p|
+        unless p.exist?
+          p.mkpath
+          p.chmod(0o777)
+        end
+        p.join(login).tap(&:mkpath)
```
olleolleolle: In order to avoid confusion with the method `Kernel#p`, the `p` block-local variable could be renamed.

Example name:

- temporary_home_path

segiddins:
Opened https://github.com/rspec/rspec-mocks/issues/1143

segiddins:
This is fixed on `rspec-mocks` master, in the meantime you can manually `FileUtils.module_eval { alias mkpath mkdir_p; module_function :mkpath }` in an `after` block and annotate it as a `HACK`

lib/bundler.rb
svoop: Alright... even though I doubt anyone would actually confuse it with `Kernel#p`.

svoop: Btw, there are other LOCs in Bundler featuring `p.` as well.

olleolleolle: Not arguing with that. This is just me, a random passer-by saying "Defining a local variable `p` shadows `Kernel#p`."

segiddins: It also messes with my editor's syntax highlighting :P

svoop:
@segiddins Thanks a bunch for fixing Rspec!
@indirect All green now. Besides the Rspec hack, I've only changed the variable names as suggested by @olleolleolle since your review.

segiddins:
@bundlerbot r=indirect

lib/bundler.rb
```diff
@@ -146,34 +146,37 @@ def ruby_scope
...
-
-        Bundler.ui.warn(warning)
-        tmp_home
+    def tmp_home_path(login, warning)
+      login ||= "unknown"
+      path = Pathname.new(Dir.tmpdir).join("bundler", "home")
+      SharedHelpers.filesystem_access(path) do |tmp_home_path|
+        unless tmp_home_path.exist?
+          tmp_home_path.mkpath
+          tmp_home_path.chmod(0o777)
```
jlambert121: Is this a valid filesystem mod?

segiddins: yup, 777 is the most permissive permissions bits

jlambert121: Oh - octal, it was the `o` I was looking at.

PR #5374, opened 2017-01-26
Bundle info command

colby-swandale:
This is a continuation of #5093 

This PR adds a new command called `info` that is nearly identical to the `show` command, specifically `show <gem> --verbose`.

Example:
```
› dbundler info rack
  * rack (2.0.1)
	Summary: a modular Ruby webserver interface
	Homepage: http://rack.github.io/
	Path: /Users/colby/.gem/ruby/2.4.0/gems/rack-2.0.1
```

There is also a option called `--path` that only prints the path to the gem
```
› dbundler info rack --path
/Users/colby/.gem/ruby/2.4.0/gems/rack-2.0.1
```

One noticeable difference between `info` and `show --verbose` is that i have removed the `outdated` functionality. This was made for several reasons:
* There are currently problems with that functionality which have been raised in #5375
* We have a dedicated command called `outdated` for the user to get information about the outdated status of gems.
* The outdated status requires an active internet connection and time to download the necessary information. I think not having to download anything and limiting `info` to just read information it already has would make for a more responsive and better user experience.

### DIFF
--- a/lib/bundler/cli/info.rb
+++ b/lib/bundler/cli/info.rb
@@ -1,0 +1,59 @@
+# frozen_string_literal: true
+require "bundler/cli/common"
+
+module Bundler
+  class CLI::Info
+    attr_reader :gem_name, :options, :latest_specs
+    def initialize(options, gem_name)
+      @options = options
+      @gem_name = gem_name
+      @latest_specs = fetch_latest_specs if not options[:path]
+    end
+
+    def run
+      Bundler.ui.silence do
+        Bundler.definition.validate_runtime!
+        Bundler.load.lock
+      end
+
+      spec = Bundler::CLI::Common.select_spec(gem_name, :regex_match)
+      return unless spec
+
+      path = spec.full_gem_path
+      if options[:path]
+        return Bundler.ui.info(path)
+      end
+
+      unless File.directory?(path)
+        Bundler.ui.warn("The gem #{gem_name} has been deleted. It was installed at:")
+      end
+
+      return print_gem_info spec
+    end
+
+    private
+
+    def print_gem_info spec
+      desc = "  * #{spec.name} (#{spec.version}#{spec.git_version})"
+      latest = latest_specs.find {|l| l.name == spec.name }
+      Bundler.ui.info <<-END.gsub(/^ +/, "")
+        #{desc}
+        \tSummary:  #{spec.summary || "No description available."}
+        \tHomepage: #{spec.homepage || "No website available."}
+        \tStatus:   #{outdated?(spec, latest) ? "Outdated - #{spec.version} < #{latest.version}" : "Up to date"}
+      END
+    end
+
+    def fetch_latest_specs
+      definition = Bundler.definition(true)
+      definition.resolve_remotely!
+      definition.specs
+    end
+
+    def outdated?(current, latest)
+      return false unless latest
+      Gem::Version.new(current.version) < Gem::Version.new(latest.version)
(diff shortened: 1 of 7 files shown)


lib/bundler/cli/info.rb
```diff
@@ -0,0 +1,33 @@
...
+module Bundler
+  class CLI::Info
+    attr_reader :gem_name, :options
+    def initialize(options, gem_name)
+      @options = options
+      @gem_name = gem_name
+    end
+
+    def run
+      spec = Bundler::CLI::Common.select_spec(gem_name, :regex_match)
```
segiddins: Needs to return if spec is nil

lib/bundler/cli/info.rb
```diff
@@ -0,0 +1,33 @@
...
+      return print_gem_path(spec) if @options[:path]
+      print_gem_info(spec)
+    end
+
+    private
+
+    def print_gem_path(spec)
+      Bundler.ui.info spec.full_gem_path
```
segiddins: Does this properly handle default gems?

colby-swandale:
The issues i found with `bundle show --outdated` are in #5375

lib/bundler/cli/info.rb
colby-swandale: nope, i'll get right on that!

colby-swandale: @segiddins do you expect something like `bundle info rake` to work like?
```
› dbundler info rake
  * rake (12.0.0)
	Summary: Rake is a Make-like program implemented in Ruby
	Homepage: https://github.com/ruby/rake
	Path: /Users/colby/.rubies/ruby-2.4.0/lib/ruby/gems/2.4.0/gems/rake-12.0.0
```

segiddins: Yup!

colby-swandale:
@segiddins sorry for bugging you again, is there are better way to check for a default gem? I'm fairly unfamiliar with the rubygems api. Thanks in advance.

```diff
+      begin
+        gem = Gem::Specification.find_by_name(gem_name)
+       spec = gem if gem.default_gem?
+      rescue Gem::MissingSpecError
+        nil
+      end
```

segiddins:
Why not ask the selected spec if it's a default gem instead of searching separately?

colby-swandale:
I'm probably missing something here. The issue I'm having is that the gems that the user can perform `bundle info <gem>` on is limited to the definition ie: I can't `bundle info rack` unless the definition has `rack` as well.

The change i made checks to see if there are any gems installed in the environment with the given name and is also a default gem before it checks the definition. This is because `Bundler::CLI::Common.self.select_spec` will print an error when the gem cannot be found. Without this change i get:

```
› dbundler info rdoc
Could not find gem 'rdoc'.
Did you mean rack?
```

Hopefully that makes sense. Do we want to have the behavior mentioned above? let me know 😃

spec/commands/info_spec.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+require "spec_helper"
+
+describe "bundle info" do
```
segiddins: `RSpec.describe`

colby-swandale:
@segiddins could you take another look again. Thanks in advance.

segiddins:
looks good to me!

@indirect r?

indirect:
This is awesome, thanks @colby-swandale! @bundlerbot r+

PR #5382, opened 2017-01-30
[FriendlyErrors] Update ISSUES.md URL

segiddins:
This URL no longer exists after the docs reorganization

\c @feministy 

I'd like to release this as 1.14.x so people aren't frustrated when opening new issues


segiddins:
@bundlerbot r+

feministy:
😑 can't believe i missed this!! good catch!

PR #5387, opened 2017-01-31
[Definition] Remove no-op code in expanded_deps

chrismo:
While troubleshooting my bundler-fixture gem, which hacks up a
Definition class with a fake Index, and needed updating for 1.14
support for ruby and RubyGems version in the Definition index, I
noticed this block of code appears to not do anything, since a 2nd
argument is not being passed into `concat_ruby_version_requirements`
method.

The specs that were altered in the same commit that added these lines
pass with or without the lines of code, so I'm committing this up to do
a full Travis run to see if there are any other consequences. If not,
then we either can remove these lines, or need to hook them up properly
and add proper failing specs.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -789,10 +789,10 @@ module Bundler
     def expanded_dependencies
       @expanded_dependencies ||= begin
         ruby_versions = concat_ruby_version_requirements(@ruby_version)
-        if ruby_versions.empty? || !@ruby_version.exact?
-          concat_ruby_version_requirements(RubyVersion.system)
-          concat_ruby_version_requirements(locked_ruby_version_object) unless @unlock[:ruby]
-        end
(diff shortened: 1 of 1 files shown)


chrismo:
@segiddins dunno if i'll get time soon to dive into this, to figure out if this code should stay and be connected properly, and what the failing spec would be for it - or if this code can just be removed.

segiddins:
Let's rip it out for now

segiddins:
Or change it to `concat_ruby_versions(version, ruby_versions)` which is what I think it _should_ be

segiddins:
@chrismo which approach would you prefer?

segiddins:
@chrismo ping

colby-swandale:
ping @chrismo

esasse:
Closing due to no response from the author.

Issue #5235, opened 2016-12-07
'bundle exec' loads gemified openssl/io-console from outside the restricted gem set


rhenium:
`bundle exec` loads io/console and openssl before restricting load path and causes two different versions of the library to be loaded, if the latest installed version is not specified in Gemfile.

At the moment of Ruby 2.4, only io/console and openssl are affected because other stdlibs loaded by `bundle exec` aren't (yet?[1]) converted into default (!= bundled) gems.

Steps to reproduce (with io/console, but can be read as openssl if you use Ruby >= 2.4.0-preview1):

1. Install Ruby 2.3.3 which includes io-console 0.4.5 as a default gem and Bundler 1.13.6.
2. Run `gem install io-console` to install io-console 0.4.6. `gem list` shows `io-console (0.4.6, 0.4.5)`.
3. Setup an app with:

    Gemfile:

    ```
    source "https://rubygems.org"
    ```

    main.rb:

    ```
    puts "before require: #{$LOADED_FEATURES.grep(/console/)}"
    require "io/console"
    puts "after require: #{$LOADED_FEATURES.grep(/console/)}"
    ```

4. Run `bundle exec ruby -w main.rb`. The output shows that two versions are loaded:

    ```
    before require: ["/home/k/.rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/io-console-0.4.6/lib/io/console.so"]
    /home/k/.rbenv/versions/2.3.3/lib/ruby/2.3.0/x86_64-linux/io/console.so: warning: method redefined; discarding old raw
    /home/k/.rbenv/versions/2.3.3/lib/ruby/2.3.0/x86_64-linux/io/console.so: warning: method redefined; discarding old raw!
    /home/k/.rbenv/versions/2.3.3/lib/ruby/2.3.0/x86_64-linux/io/console.so: warning: method redefined; discarding old cooked
    /home/k/.rbenv/versions/2.3.3/lib/ruby/2.3.0/x86_64-linux/io/console.so: warning: method redefined; discarding old cooked!
    /home/k/.rbenv/versions/2.3.3/lib/ruby/2.3.0/x86_64-linux/io/console.so: warning: method redefined; discarding old getch
    /home/k/.rbenv/versions/2.3.3/lib/ruby/2.3.0/x86_64-linux/io/console.so: warning: method redefined; discarding old echo=
[...]

PR #5388, opened 2017-02-01
Defer requiring rubygems/spec_fetcher until it becomes necessary

rhenium:
Avoid conflict between two versions of openssl gem on 'bundle exec' if
a newer and non-default version of openssl gem is installed to the
system. rubygems/spec_fetcher loads openssl through resolv and
securerandom when running with Ruby <= 2.4.

This is not a proper fix for #5235 as other standard libraries that
Bundler currently loads will be gemified as well in Ruby >= 2.5,
however, this will fix openssl's case.

Reference: https://github.com/bundler/bundler/issues/5235
Fixes: https://github.com/ruby/openssl/issues/103

---

Reproduce:

1. Install Ruby 2.4.0 (comes with openssl 2.0.2)
2. Run `gem install openssl` to install openssl 2.0.3
3. Create a Gemfile and run `bundle install`
4. Run `bundle exec ruby -e'require "openssl"'

### DIFF
--- a/spec/runtime/setup_spec.rb
+++ b/spec/runtime/setup_spec.rb
@@ -1080,8 +1080,8 @@ RSpec.describe "Bundler.setup" do
     end
   end
 
-  describe "when Psych is not in the Gemfile", :ruby => "~> 2.2" do
-    it "does not load Psych" do
+  describe "with gemified standard libraries" do
+    it "does not load Psych", :ruby => "~> 2.2" do
       gemfile ""
       ruby <<-RUBY
         require 'bundler/setup'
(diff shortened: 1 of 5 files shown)


segiddins:
Awesome! Any way we could add tests for this?

rhenium:
Just pushed new commit.

segiddins:
Nice! Can you also add a eat that `Gem.loaded_specs` is empty, so we notice when more of the stdlib gets gem-ified? Thanks so much!

rhenium:
That really should be tested, but it fails because io-console is still loaded unconditionally without rubygems/rubygems@c65afec704a2 that is not yet in current releases of RubyGems.

spec/runtime/setup_spec.rb
```diff
@@ -1080,19 +1080,16 @@ def lock_with(ruby_version = nil)
...
-      expect(pre_bundler).to eq("undefined")
-      expect(post_bundler).to match(/\d+\.\d+\.\d+/)
-    end
+  it "does not load gemified standard libraries" do
+    stdlibs = %w(Psych OpenSSL)
+    install_gemfile ""
+    ruby! <<-RUBY
+      require "bundler/setup"
+      #{stdlibs.inspect}.each do |const|
+        puts Object.const_defined?(const) ? const : "undefined"
```
segiddins: can we also verify that requiring still works, like the deleted test did? Thanks!

segiddins: ping on this?

rhenium: (Sorry for taking so long to get back to you.) Pushed updated commit. I had to restore the psych part because of the Ruby version constraint.

segiddins:
Thanks!
@bundlerbot r+

PR #5393, opened 2017-02-02
Fix misleading warning message when a dependency is unused on any platform

5t111111:
The current warning message for a dep with empty platform is misleading/confusing. You see a sort of the below warning when you `bundle install`,

```
The dependency tzinfo-data (>= 0) will be unused by any of the platforms Bundler is installing for. Bundler is installing for ruby but the dependency is only for x86-mingw32, x86-mswin32, x64-mingw32, java. To add those platforms to the bundle, run `bundle lock --add-platform mingw, mswin, x64_mingw, jruby`.
```

However, there two issues exist in this message "run `bundle lock --add-platform mingw, mswin, x64_mingw, jruby`":

- You must not have comma (,) for separators
  - these commas will be considered as comma itself. see [here](https://github.com/bundler/bundler/blob/242fcbcb2d15d39c63e3441756b77fbfcc1b73b2/lib/bundler/cli/lock.rb#L34-L41)
- mingw, mswin, x64_mingw, jruby are not valid platforms
  - platforms must be specified as the formats that rubygems can understand, like (x86-mingw32, x86-mswin32, x64-mingw32, java)

This fixes these issues.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -829,7 +829,8 @@ module Bundler
             "The dependency #{dep} will be unused by any of the platforms Bundler is installing for. " \
             "Bundler is installing for #{@platforms.join ", "} but the dependency " \
             "is only for #{dep.platforms.map {|p| Dependency::PLATFORM_MAP[p] }.join ", "}. " \
-            "To add those platforms to the bundle, run `bundle lock --add-platform #{dep.platforms.join ", "}`."
+            "To add those platforms to the bundle, " \
+            "run `bundle lock --add-platform #{dep.platforms.map {|p| Dependency::PLATFORM_MAP[p] }.join " "}`."
         end
         platforms.each do |p|
           deps << DepProxy.new(dep, p) if remote || p == generic_local_platform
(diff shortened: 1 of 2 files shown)


lib/bundler/definition.rb
```diff
@@ -829,7 +829,8 @@ def expand_dependencies(dependencies, remote = false)
             "The dependency #{dep} will be unused by any of the platforms Bundler is installing for. " \
             "Bundler is installing for #{@platforms.join ", "} but the dependency " \
             "is only for #{dep.platforms.map {|p| Dependency::PLATFORM_MAP[p] }.join ", "}. " \
-            "To add those platforms to the bundle, run `bundle lock --add-platform #{dep.platforms.join ", "}`."
+            "To add those platforms to the bundle, " \
+            "run `bundle lock --add-platform #{dep.platforms.map {|p| Dependency::PLATFORM_MAP[p] }.join " "}`."
```
segiddins: Maybe extract the array from map into a variable since it's also used on the line above?

segiddins:
Awesome, thanks!

lib/bundler/definition.rb
5t111111: alright, in a sec 😉

segiddins:
Thanks! @bundlerbot r+

PR #5396, opened 2017-02-02
lock: make {add,remove}-platform default to the local platform

sschuberth:
We need to use "lazy_default" instead of "default" as otherwise thor does
not properly return the default value for array types. Guard against the
options being "nil" by wrapping them in "Array()".

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -479,9 +479,9 @@ module Bundler
       "the path the lockfile should be written to"
     method_option "full-index", :type => :boolean, :default => false, :banner =>
       "Fall back to using the single-file index of all gems"
-    method_option "add-platform", :type => :array, :default => [], :banner =>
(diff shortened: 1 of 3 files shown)


segiddins:
I don't think we want a default at all here

sschuberth:
Mind elaborating why not?

segiddins:
Because the user should be specifying exactly which platforms should be removed -- defaulting to the local platform seems arbitrary and confusing to me

sschuberth:
> which platforms should be removed

You explicitly say "removed" only. Does that mean you're fine with defaulting to the local platform for *adding* platforms?

> defaulting to the local platform seems arbitrary

I really don't see how that's arbitrary. It's not that we're rolling the dice. If it's in question which platform to use, using the local one seems quite natural to me.

segiddins:
I'd rather not have defaults for either

sschuberth:
Ok, well, then I' close this.

PR #5399, opened 2017-02-03
Set build args properly for Rubygems >=2.0

khrvi:
There is an issue with applying 'build' args that were set through 'bundler config' for Rubygems >=2.0. The default 'install_with_build_args' set them properly https://github.com/bundler/bundler/blob/master/lib/bundler/rubygems_integration.rb#L256 but existing one doesn't set at all.

### DIFF
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -724,10 +724,6 @@ module Bundler
       def repository_subdirectories
         Gem::REPOSITORY_SUBDIRECTORIES
       end
-
-      def install_with_build_args(args)
-        yield
-      end
     end
 
     # RubyGems 2.1.0


segiddins:
Please open an issue with details of the problem you're seeing rather than just deleting code?

khrvi:
Added issue - https://github.com/bundler/bundler/issues/5401.
It could be possible that I missed an idea of overriding method 'install_with_build_args' for RubyGems 2.0 ... if so please let me know and I'll modify a fix.

segiddins:
The idea is that rubygems 2.0+ allow setting build args directly on the gem installer, so we could avoid mutexing installation

khrvi:
Well, I definitely can install gem using 'gem install' with '--with-cppflags' but it sounds like a unrelated to bundler config issue. It seems to me that mutex usage for concurrent gem installation still reasonable because of an args that user can set in the bundler config.
Please share your thoughts and thank you for quick response.

segiddins:
https://github.com/bundler/bundler/issues/5410 should address this

PR #5403, opened 2017-02-04
Allow to set float number as mirror timeout.

seban:
I noticed some inconsistency in mirror timeout setting during playing
with gemstash. Default value is `0.1`, but when I set timeout value to
`0.5` it behaved as with `0` value. It set proper timeout only when I
set `1` as timeout value. I think it is also not proper behaviour with
`0.1` default and while setting cast it to Fixnum.

### DIFF
--- a/lib/bundler/mirror.rb
+++ b/lib/bundler/mirror.rb
@@ -81,7 +81,7 @@ module Bundler
         when false, "false"
           @fallback_timeout = 0
         else
-          @fallback_timeout = timeout.to_i
+          @fallback_timeout = timeout.to_f
         end
         @valid = nil
       end
--- a/spec/bundler/mirror_spec.rb
+++ b/spec/bundler/mirror_spec.rb
@@ -34,6 +34,11 @@ RSpec.describe Bundler::Settings::Mirror do
     expect(mirror.fallback_timeout).to eq(0)
   end
 
+  it "takes a float number as a fallback_timeout" do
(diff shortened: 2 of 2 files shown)


segiddins:
I think the issue with this might be that Settings only stores ints and not floats? But I could be wrong

segiddins:
Yeah, `Settings` only stores ints for number fields, so I think in the place this is set, it wont work

seban:
@segiddins could you explain it more to me? In `lib/bundler/mirror.rb` there is defined `Bundler::Settings::Mirror` class defined. Why `Settings` would be a problem in this case. Is this connected with config defined in `bundle config mirror.https://rubygems.org.fallback_timeout 0.5` way?

segiddins:
Because `bundle config` reads and writes to the `Settings` class, where it appears we only store integers and not floats

olleolleolle:
Can this finding be turned into clarified documentation, if it can not be allowed to have floats in settings?

segiddins:
@olleolleolle sure!

seban:
@olleolleolle @segiddins I can adjust documentation in this case. bundler, bundler-site and gemstash need to be adjusted or something else?

segiddins:
Just `man/bundle-config` should be necessary -- it's where all the settings are documented

segiddins:
Closing in favor of https://github.com/bundler/bundler/pull/5417 -- thanks for bringing this us @seban !

PR #5405, opened 2017-02-05
CI against Ruby 2.1.10

koic:
Ruby 2.1.10 was released.

https://www.ruby-lang.org/en/news/2016/04/01/ruby-2-1-10-released


segiddins:
Thanks!
@bundlerbot r+

Issue #5401, opened 2017-02-03
Not applicable bundler config build args for Rubygems >=2.0


khrvi:
There is an issue with applying 'build' args that were set through 'bundler config' for Rubygems >=2.0. The default 'install_with_build_args' set them properly https://github.com/bundler/bundler/blob/master/lib/bundler/rubygems_integration.rb#L256 but existing one doesn't set at all.

## Environment
Bundler config
```
bundle config
Settings are listed in order of priority. The top value will be used.
build.eventmachine
Set for the current user (/Users/slava/.bundle/config): "--with-cppflags=-I/usr/local/opt/openssl/include"
```
Gemfile
`gem 'eventmachine', '1.0.0.11',  :git => 'git@github.com:private_repo/eventmachine.git', :tag => '1.0.0.11'`

## Backtrace
bundle -V (version 1.13.7)
```
Using eventmachine 1.0.0.11 from git@github.com:private_repo/eventmachine.git (at 1.0.0.11@e965d81)
Building native extensions with: '--with-cppflags=-I/usr/local/opt/openssl/include'
This could take a while...
```
bundle -V (version 1.14.3)
```
Using eventmachine 1.0.0.11 from git@github.com:private_repo/eventmachine.git (at 1.0.0.11@e965d81)
Building native extensions.  This could take a while...
.....
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.
 /Users/slava/.rbenv/versions/2.1.2/bin/ruby extconf.rb
checking for rb_trap_immediate in ruby.h,rubysig.h... no
checking for rb_thread_blocking_region()... yes
checking for inotify_init() in sys/inotify.h... no
checking for __NR_inotify_init in sys/syscall.h... no
checking for writev() in sys/uio.h... yes
checking for rb_enable_interrupt()... no
checking for rb_thread_check_ints()... yes
checking for rb_time_new()... yes
checking for sys/event.h... yes
checking for sys/queue.h... yes
creating Makefile

make "DESTDIR=" clean

make "DESTDIR="
compiling binder.cpp
In file included from binder.cpp:20:
./project.h:110:10: fatal error: 'openssl/ssl.h' file not found
#include <openssl/ssl.h>
         ^
1 error generated.
make: *** [binder.o] Error 1

make failed, exit code 2

```

As it shown above cpp flags for compiling native extension weren't applied for 'bundler' version 1.14.x. The latest working version is '1.13.7'

segiddins:
[...]

PR #5410, opened 2017-02-08
Properly set native extension build args for git gems

segiddins:
Closes https://github.com/bundler/bundler/pull/5399
Closes https://github.com/bundler/bundler/issues/5401


indirect:
@bundlerbot r+

Issue #5426, opened 2017-02-13
fatal: No live threads left. Deadlock?


bradly-swart:
--- ERROR REPORT TEMPLATE -------------------------------------------------------
# Error Report

## Questions

Please fill out answers to these questions, it'll help us figure out
why things are going wrong.

- **What did you do?**

  I ran the command `/Users/bswart/.rbenv/versions/2.3.1/bin/bundle `

- **What did you expect to happen?**

  I expected Bundler to install gems

- **What happened instead?**

  Instead, what happened was the last gem to install,` guard-rspec 4.2.10` , failed with the error attached, fatal: No live threads left. Deadlock?

- **Have you tried any solutions posted on similar issues in our issue tracker, stack overflow, or google?**

 Nothing on google.

- **Have you read our issues document, https://github.com/bundler/bundler/blob/master/doc/contributing/ISSUES.md?**

  ...

## Backtrace

```
fatal: No live threads left. Deadlock?
  /Users/bswart/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/bundler-1.14.4/lib/bundler/worker.rb:43:in `pop'
  /Users/bswart/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/bundler-1.14.4/lib/bundler/worker.rb:43:in `deq'
  /Users/bswart/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/bundler-1.14.4/lib/bundler/installer/parallel_installer.rb:131:in `process_specs'
  /Users/bswart/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/bundler-1.14.4/lib/bundler/installer/parallel_installer.rb:102:in `call'
  /Users/bswart/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/bundler-1.14.4/lib/bundler/installer/parallel_installer.rb:77:in `call'
  /Users/bswart/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/bundler-1.14.4/lib/bundler/installer.rb:199:in `install_in_parallel'
  /Users/bswart/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/bundler-1.14.4/lib/bundler/installer.rb:166:in `install'
[...]

Issue #5339, opened 2017-01-21
Errors installing facter 2.4.6 with bundler/inline


chrismo:
As a part of the coverage script I put together for #5332, one of the gems it attempted to install was 'facter' and its latest version '2.4.6'. The exceptions only happen with bundler/inline. No exceptions happen using bundler normally. There are two versions of the exception, one for 1.13.7 and another for 1.14.0.pre.2.

With this script:
```ruby
require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'
  gem 'facter', '2.4.6'
end
```
1.13.7 results in:
```
/Users/chrismo/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/bundler-1.13.7/lib/bundler/resolver.rb:366:in `block in verify_gemfile_dependencies_are_found!': Could not find gem 'facter (= 2.4.6)' in any of the gem sources listed in your Gemfile or available on this machine. (Bundler::GemNotFound)
	from /Users/chrismo/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/bundler-1.13.7/lib/bundler/resolver.rb:341:in `each'
	from /Users/chrismo/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/bundler-1.13.7/lib/bundler/resolver.rb:341:in `verify_gemfile_dependencies_are_found!'
	from /Users/chrismo/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/bundler-1.13.7/lib/bundler/resolver.rb:199:in `start'
	from /Users/chrismo/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/bundler-1.13.7/lib/bundler/resolver.rb:181:in `resolve'
	from /Users/chrismo/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/bundler-1.13.7/lib/bundler/definition.rb:250:in `resolve'
	from /Users/chrismo/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/bundler-1.13.7/lib/bundler/definition.rb:174:in `specs'
	from /Users/chrismo/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/bundler-1.13.7/lib/bundler/installer.rb:171:in `ensure_specs_are_compatible!'
[...]

PR #5427, opened 2017-02-13
Fail gracefully when installing a spec where the API is missing deps

segiddins:
Fixes https://github.com/bundler/bundler/issues/5426
Closes https://github.com/bundler/bundler/issues/5339

### DIFF
--- a/lib/bundler/endpoint_specification.rb
+++ b/lib/bundler/endpoint_specification.rb
@@ -91,6 +91,11 @@ module Bundler
     end
 
     def __swap__(spec)
+      if (extra_deps = spec.runtime_dependencies.-(dependencies)) && extra_deps.any?
+        raise APIResponseMismatchError,
+          "Downloading #{full_name} revealed dependencies not in the API (#{extra_deps.map(&:to_s).join(", ")})." \
+          "\nInstalling with `--full-index` should fix the problem."
+      end
       @remote_specification = spec
     end
 
--- a/lib/bundler/endpoint_specification.rb
+++ b/lib/bundler/endpoint_specification.rb
@@ -91,7 +91,8 @@ module Bundler
     end
 
     def __swap__(spec)
-      if (extra_deps = spec.runtime_dependencies.-(dependencies)) && extra_deps.any?
+      without_type = proc {|d| Gem::Dependency.new(d.name, d.requirements_list) }
+      if (extra_deps = spec.runtime_dependencies.map(&without_type).-(dependencies.map(&without_type))) && extra_deps.any?
         raise APIResponseMismatchError,
           "Downloading #{full_name} revealed dependencies not in the API (#{extra_deps.map(&:to_s).join(", ")})." \
           "\nInstalling with `--full-index` should fix the problem."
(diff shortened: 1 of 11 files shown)


indirect:
@segiddins can you explain how this fix is related to the problem in the linked ticket?

segiddins:
Sure -- instead of deadlocking in the parallel installer, we print out that the API is returning messed up data (because the __swap__ happens after downloading the full spec) and error out suggesting using `--full-index`.

indirect:
cool, sounds good to me. maybe we could add some debug-only output to let us figure out what API responses are bad, so we can fix them?

segiddins:
Sure! I'll add that tomorrow -- debug or verbose output?

segiddins:
@indirect done

lib/bundler/installer/gem_installer.rb
```diff
@@ -16,7 +16,7 @@ def install_from_spec
       Bundler.ui.debug "#{worker}:  #{spec.name} (#{spec.version}) from #{spec.loaded_from}"
       generate_executable_stubs
       return true, post_install_message
-    rescue Bundler::InstallHookError, Bundler::SecurityError
+    rescue Bundler::InstallHookError, Bundler::SecurityError, APIResponseMismatchError
```
olleolleolle: Can they all be in the "short form", without the namespace wrapped around them?

```ruby
rescue InstallHookError, SecurityError, APIResponseMismatchError
```

segiddins: they could, just didn't feel like changing the other ones since they were already there

segiddins:
Doesn't actually fix the linked-to issue :(

segiddins:
OK, this should fix both the original issue and the first case I fixed, which is when the API has the wrong dependencies. @bradly-swart want to give this a go?

bradly-swart:
I'll give it a whirl, thank you.

indirect:
@segiddins I'm a little bit worried about the possible performance implications of doing this many checks on every single swap action—do you have any idea what the performance implications of this change are?

lib/bundler/endpoint_specification.rb
```diff
@@ -91,6 +91,13 @@ def _local_specification
     end
 
     def __swap__(spec)
+      without_type = proc {|d| Gem::Dependency.new(d.name, d.requirements_list.sort) }
+      if (extra_deps = spec.runtime_dependencies.map(&without_type).-(dependencies.map(&without_type))) && extra_deps.any?
```
ivanovaleksey: Could you please explain to me what does `.-()` mean? 
Is it array difference [method](https://ruby-doc.org/core-2.3.3/Array.html#method-i-2D)?
Are there some pitfalls of plain `[] - []`?

agis-: @ivanovaleksey It's the same thing. `[] - []` is really `[].-([])`.

ivanovaleksey: Yes, I see. Is just matter of style? No any special caveats?

segiddins:
> do you have any idea what the performance implications of this change are?

Given `#__swap__` is only ever caller in `Source::Rubygems#install` (and only then if the spec comes from a remote source), I don't think a map and array difference will have a noticeable performance impact (especially compared to the surrounding network & disk IO)

indirect:
@segiddins sold. let's do it. 👍

@bundlerbot r+

segiddins:
@bundlerbot retry

segiddins:
@bundlerbot retry

PR #5436, opened 2017-02-16
Use empty array when `spec_settings` returns `nil`

okkez:
If use bundler 1.14.4 with rubygems 2.6.8 and
`Bundler.settings["build.#{spec.name}"]` returns `nil` then we get
error "can't modify frozen literal string" from [rubygems](https://github.com/rubygems/rubygems/blob/v2.6.8/lib/rubygems/ext/rake_builder.rb#L13).
RubyGems 2.6.8 is the default version for Ruby2.4.0.

See also https://github.com/sickill/rainbow/issues/48

### DIFF
--- a/lib/bundler/installer/gem_installer.rb
+++ b/lib/bundler/installer/gem_installer.rb
@@ -52,7 +52,7 @@ module Bundler
     end
 
     def install
-      spec.source.install(spec, :force => force, :ensure_builtin_gems_cached => standalone, :build_args => [spec_settings])
+      spec.source.install(spec, :force => force, :ensure_builtin_gems_cached => standalone, :build_args => Array(spec_settings))
     end
 
     def install_with_settings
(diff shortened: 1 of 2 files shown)


okkez:
I could not add spec for this PR.
Could you add spec for this PR or how to add spec for this PR?

segiddins:
This was a bug in RubyGems that we fixed with the 2.6.10 release

okkez:
@segiddins I know that.
But I think that Ruby2.4.0 bundles RubyGems 2.6.8 and many CI services (such as CircleCI etc.) use RubyGems 2.6.8 or earlier with latest bundler. This change can avoid unexpected difficult "can't modify frozen literal string" errors on such CI services.

On the other hand, `build_args: [nil]` if `spec_settings` returns `nil`.
And `Gem::Ext::RakeBuilder.build` expects `args` does not include `nil`.
Bundler calls `Gem::Ext::RakeBuilder.build`with wrong usage.
This is a potential bug.

segiddins:
If you can write a test for this, I think that'd be acceptable -- a unit test inside `spec/bundler` would be sufficient

segiddins:
Failing on older rubygems: https://travis-ci.org/bundler/bundler/jobs/202845177#L405-L415

okkez:
Thank you for comment.

segiddins:
This is wonderful -- thanks so much!

@bundlerbot r+

PR #5439, opened 2017-02-16
Fix Github username for README.md contrib URL

### LINKED ISSUE #5438: Invalid Github URL in README
Bundler version 1.14.3

With using
```
bundle gem example
```
The `README.md` file under contributions shows as:

```
https://github.com/Daniel P. Clark/example
```

This is taken from `git config --get user.name` (I believe)

The command that should be used is `git config --get github.user` which would result in:

```
https://github.com/danielpclark/example
```

danielpclark:
This fixes #5438 where the wrong git config variable was being
retrieved.

This was using `git config user.name` rather than `git config github.user` in URL in `README.md`

### DIFF
--- a/lib/bundler/cli/gem.rb
+++ b/lib/bundler/cli/gem.rb
@@ -29,7 +29,8 @@ module Bundler
       constant_name = name.gsub(/-[_-]*(?![_-]|$)/) { "::" }.gsub(/([_-]+|(::)|^)(.|$)/) { $2.to_s + $3.upcase }
       constant_array = constant_name.split("::")
 
-      git_user_name = `git config user.name`.chomp
+      git_author_name = `git config user.name`.chomp
+      github_username = `git config github.user`.chomp
       git_user_email = `git config user.email`.chomp
 
       config = {
@@ -39,13 +40,13 @@ module Bundler
         :makefile_path    => "#{underscored_name}/#{underscored_name}",
         :constant_name    => constant_name,
         :constant_array   => constant_array,
-        :author           => git_user_name.empty? ? "TODO: Write your name" : git_user_name,
+        :author           => git_author_name.empty? ? "TODO: Write your name" : git_author_name,
         :email            => git_user_email.empty? ? "TODO: Write your email address" : git_user_email,
         :test             => options[:test],
         :ext              => options[:ext],
         :exe              => options[:exe],
         :bundler_version  => bundler_dependency_version,
-        :git_user_name    => git_user_name.empty? ? "[USERNAME]" : git_user_name
+        :github_username  => github_username.empty? ? "[USERNAME]" : github_username
       }
       ensure_safe_gem_name(name, constant_array)
 
--- a/spec/commands/newgem_spec.rb
+++ b/spec/commands/newgem_spec.rb
@@ -25,6 +25,7 @@ RSpec.describe "bundle gem" do
 
   def gem_skeleton_assertions(gem_name)
     expect(bundled_app("#{gem_name}/#{gem_name}.gemspec")).to exist
+    expect(bundled_app("#{gem_name}/README.md")).to exist
     expect(bundled_app("#{gem_name}/Gemfile")).to exist
     expect(bundled_app("#{gem_name}/Rakefile")).to exist
     expect(bundled_app("#{gem_name}/lib/test/gem.rb")).to exist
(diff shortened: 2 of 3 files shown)


lib/bundler/cli/gem.rb
```diff
@@ -29,7 +29,7 @@ def run
       constant_name = name.gsub(/-[_-]*(?![_-]|$)/) { "::" }.gsub(/([_-]+|(::)|^)(.|$)/) { $2.to_s + $3.upcase }
       constant_array = constant_name.split("::")
 
-      git_user_name = `git config user.name`.chomp
+      git_user_name = `git config github.user`.chomp
```
segiddins: Can we rename to GitHub username?

danielpclark: I'm working on it

spec/commands/newgem_spec.rb
```diff
@@ -116,19 +119,20 @@ def gem_skeleton_assertions(gem_name)
...
         expect(bundled_app("test_gem/README.md").read).not_to include("[USERNAME]")
+        expect(bundled_app("test_gem/README.md").read).to include("github.com/bundleuser")
       end
     end
 
-    context "git config user.name is absent" do
+    context "git config gihhub.user is absent" do
       before do
-        `git config --unset user.name`
+        `git config --unset github.user`
```
segiddins: `sys_exec("git config --unset github.user")`

spec/commands/newgem_spec.rb
```diff
@@ -116,19 +119,20 @@ def gem_skeleton_assertions(gem_name)
...
       end
 
       it "contribute URL set to git username" do
         expect(bundled_app("test_gem/README.md").read).not_to include("[USERNAME]")
+        expect(bundled_app("test_gem/README.md").read).to include("github.com/bundleuser")
       end
     end
 
-    context "git config user.name is absent" do
+    context "git config gihhub.user is absent" do
```
segiddins: can we please also keep a test that runs with `user.name` unset?

segiddins: `github`

danielpclark: That's already accounted for 4 times.

https://github.com/danielpclark/bundler/blob/54de051a2e24bddefa85bbb1ffadadacd6800f2d/spec/commands/newgem_spec.rb#L56 ,
https://github.com/danielpclark/bundler/blob/54de051a2e24bddefa85bbb1ffadadacd6800f2d/spec/commands/newgem_spec.rb#L67 ,
https://github.com/danielpclark/bundler/blob/54de051a2e24bddefa85bbb1ffadadacd6800f2d/spec/commands/newgem_spec.rb#L253 , https://github.com/danielpclark/bundler/blob/54de051a2e24bddefa85bbb1ffadadacd6800f2d/spec/commands/newgem_spec.rb#L543

spec/commands/newgem_spec.rb
danielpclark: This exact syntax with `git` is used 5 times in this file.  Is there a new policy written somewhere for this change to be enacted?

olleolleolle: @danielpclark The cool thing about `sys_exec` is that in `spec_helpers/support.rb` it is defined like this:

```ruby
    def sys_exec(cmd)
        yield stdin, stdout, wait_thr if block_given?
        stdin.close

        @exitstatus = wait_thr && wait_thr.value.exitstatus
        @out = Thread.new { stdout.read }.value.strip
        @err = Thread.new { stderr.read }.value.strip
      end

      (@all_output ||= String.new) << [
        "$ #{cmd.to_s.strip}",
        out,
        err,
        @exitstatus ? "# $? => #{@exitstatus}" : "",
        "\n",
      ].reject(&:empty?).join("\n")

      @out
    end
    bang :sys_exec

```

As you can see, some of the noteworthy output-management makes easily-checkable, standardized strings out of the I/O.

There are other helper methods for specs that check that standardized output.

So, using `sys_exec` smooths the path for the RSpec usage in the project.

danielpclark: @olleolleolle Alright, that sounds reasonable.  Is it cool if I modify all 5 lines in this file from back-ticks to this as a second commit?  Or should I keep the scope of the PR just to my changes.

olleolleolle: @danielpclark Here is my take. Code collaboration is difficult, at all times, and it's not interactive, and we all have to wait on each other. Timezones and life and everything. But we can do it!

To get _this_ change merged: focus on the immediate feedback.

If you only change the file where you _have_ to, a second, clean-up-only PR with only "go from backticks to `sys_exec` everywhere in this file" could be a way to get this fix (and the next one) quicker to "done".

Keep changes easy to review.

_Good luck, and thanks for doing this!_

olleolleolle:
@danielpclark Oh, the `gihhub.user` typo is still in there.

danielpclark:
@olleolleolle Got it.

segiddins:
Thanks so much!
@bundlerbot r+

PR #5440, opened 2017-02-16
Enable compact index when OpenSSL FIPS mode is enabled but not active

### LINKED ISSUE #5433: Bundler disabling Compact Index on all FIPS-supported OpenSSL installations
Since #5222, Bundler is disabling Compact Index on any system where OpenSSL is compiled with _support_ for FIPS mode, regardless of whether or not FIPS mode is currently _enabled_ on the process (e.g., by setting `OpenSSL.fips_mode = true` in Ruby).

Compare the following:

FIPS mode supported but not enabled:

```
$ ruby -ropenssl -rdigest/md5 -e "p OpenSSL::OPENSSL_FIPS; p Digest::MD5.hexdigest('foo')"
true
"acbd18db4cc2f85cedef654fccc4a4d8"
```

FIPS mode supported and enabled:

```
$ ruby -ropenssl -rdigest/md5 -e "p OpenSSL::OPENSSL_FIPS; OpenSSL.fips_mode = true; p Digest::MD5.hexdigest('foo')"
true
md5_dgst.c(83): OpenSSL internal error, assertion failed: Digest MD5 forbidden in FIPS mode!
Aborted
```

This bug is affecting current versions of Amazon Linux, which have `OpenSSL 1.0.1k-fips` installed by default.

wjordan:
Fixes #5433. Since there is no easy accessor in Ruby to detect whether or not FIPS mode is currently active, the best approach I could come up with is to `fork` a separate process and attempt to generate a build MD5 object as a test of whether MD5 module is currently available.

Because `fork` approach won't work on some platforms (JRuby, Windows etc), `md5_supported?` returns `false` on any platforms where FIPS mode is enabled and `Process.respond_to?(:fork)` is `false`.

I've added a spec that simulates behavior when OpenSSL FIPS mode is active - an error message is output to STDERR and the process is killed with the `ABRT` signal.

### DIFF
--- a/lib/bundler/fetcher/compact_index.rb
+++ b/lib/bundler/fetcher/compact_index.rb
@@ -122,14 +122,24 @@ module Bundler
       end
 
       def md5_available?
+        return true unless fips_enabled? && Process.respond_to?(:fork)
+        pid = fork do
+          $stderr.reopen(File.new("/dev/null", "w"))
+          require "digest/md5"
+          Digest::MD5.new
+          exit
+        end
+        Process.wait pid
+        $?.success?
+      end
+
+      def fips_enabled?
         begin
           require "openssl"
-          return false if defined?(OpenSSL::OPENSSL_FIPS) && OpenSSL::OPENSSL_FIPS
         rescue LoadError
           nil
         end
-
-        true
+        defined?(OpenSSL::OPENSSL_FIPS) && OpenSSL::OPENSSL_FIPS
       end
     end
   end
--- a/lib/bundler/fetcher/compact_index.rb
+++ b/lib/bundler/fetcher/compact_index.rb
@@ -122,24 +122,13 @@ module Bundler
       end
 
       def md5_available?
-        return true unless fips_enabled? && Process.respond_to?(:fork)
-        pid = fork do
-          $stderr.reopen(File.new("/dev/null", "w"))
-          require "digest/md5"
-          Digest::MD5.new
-          exit
-        end
-        Process.wait pid
-        $?.success?
-      end
-
-      def fips_enabled?
-        begin
-          require "openssl"
-        rescue LoadError
-          nil
-        end
-        defined?(OpenSSL::OPENSSL_FIPS) && OpenSSL::OPENSSL_FIPS
+        require "openssl"
+        OpenSSL::Digest::MD5.digest("")
+        true
+      rescue LoadError
+        true
+      rescue OpenSSL::Digest::DigestError
+        false
       end
     end
   end
--- a/spec/bundler/fetcher/compact_index_spec.rb
+++ b/spec/bundler/fetcher/compact_index_spec.rb
@@ -3,8 +3,8 @@ require "spec_helper"
 
 RSpec.describe Bundler::Fetcher::CompactIndex do
   let(:downloader)  { double(:downloader) }
-  let(:remote)      { double(:remote, :cache_slug => "lsjdf") }
   let(:display_uri) { URI("http://sampleuri.com") }
+  let(:remote)      { double(:remote, :cache_slug => "lsjdf", :uri => display_uri) }
   let(:compact_index) { described_class.new(downloader, remote, display_uri) }
 
   before do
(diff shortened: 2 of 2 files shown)


colby-swandale:
How does this behave on Windows environments?

wjordan:
I believe that `md5_supported?` will continue to return `false` on windows environments, assuming that `Process.respond_to?(:fork)` returns `false` in that environment. However, because Bundler's tests don't cover the Windows platform (at least doesn't appear to at first glance) that isn't something I can easily confirm.

lib/bundler/fetcher/compact_index.rb
```diff
@@ -122,14 +122,24 @@ def call(path, headers)
       end
 
       def md5_available?
+        return true unless fips_enabled? && Process.respond_to?(:fork)
+        pid = fork do
```
segiddins: I don't think forking on every Bundler process is going to be acceptable performance-wise  (in addition to all the platforms that aren't fork-safe) -- is there no way to determine in-process if the MD5 digest is available?

wjordan: - Platforms that aren't fork-safe will not reach this line of code because `Process.respond_to?(:fork)` will be `false`.
- Performance-wise, I expect a process fork to perform better than disabling the compact-index entirely, so for the case where FIPS mode is available but not active this should be a performance improvement. The tradeoff is that this causes the case where FIPS mode is active to perform  the fork.
- As far as I can tell, there is no way via Ruby's stdlib to determine in-process if the MD5 digest is available. The process is aborted as soon as OpenSSH's MD5 module is invoked, so in order to detect this failure without aborting the parent process, it's necessary to perform the test on a child process.

wjordan:
I don't think this PR is a greata solution, but it demonstrates the best I can do to extend #5222's detection-based approach to the underlying FIPS compatibility issue (#4989). I would actually prefer to see #5222 reverted and pursue an alternate solution to FIPS mode support in Bundler.

colby-swandale:
The solution to #5222 is the best we have currently, some users were unable to use Bundler when FIPS was enabled so loosing Compact Index is a suitable compromise. I'll ask the OpenSSL gem maintainers if we can work to improve the API to better detect FIPS.

colby-swandale:
Looking at the [OpenSSL gem src](https://github.com/ruby/ruby/blob/trunk/ext/openssl/ossl.c#L408) there is a method called `.fips_mode` that looks like it can be used to check if FIPS is actually enabled.

simi:
@colby-swandale it was introduced in Ruby 2.0.

rhenium:
Alternatively OpenSSL::Digest which uses the EVP interface can be used.

~~~ruby
def md5_available?
  return true unless defined?(OpenSSL)

  OpenSSL::Digest::MD5.digest("")
  true
rescue OpenSSL::Digest::DigestError
  false
end
~~~

colby-swandale:
@rhenium Does that catch the C error that gets printed? See #5222 
```
md5_dgst.c(80): OpenSSL internal error, assertion failed: Digest MD5 forbidden in FIPS mode!
```

rhenium:
> @rhenium Does that catch the C error that gets printed? See #5222

Yes. That code will not reach MD5_Init() that aborts the process.

wjordan:
Thanks @rhenium and @simi! I didn't know about those two Ruby alternatives. I've updated the PR to use `OpenSSL::Digest::MD5` to test MD5 availability- this gives a more general MD5 check that doesn't require Bundler to fork a separate process or even reference the FIPS-mode API directly.

spec/bundler/fetcher/compact_index_spec.rb
```diff
@@ -26,17 +26,47 @@
...
+
+        context "when FIPS-mode is active" do
+          before do
+            allow(OpenSSL::Digest::MD5).to receive(:digest).
+              and_raise(OpenSSL::Digest::DigestError)
+            allow(Digest::MD5).to receive(:new) do
+              # OpenSSL writes to STDERR and kills the current process with SIGABRT
+              # when FIPS mode prevents MD5 from being used.
+              $stderr.write "Digest MD5 forbidden in FIPS mode!"
+              Process.kill("ABRT", Process.pid)
```
segiddins: Is this still necessary?

wjordan: No, this is not necessary anymore for the spec to run correctly, I can remove.

segiddins:
Looks good to me!
@indirect r?

indirect:
This is awesome teamwork, everyone. Thanks so much!

@bundlerbot r+

PR #5443, opened 2017-02-17
[Resolver] Improve sorting of deps with 0 or 1 possibility

segiddins:
As the added comment says, this will force all dependencies where there are only 0 or 1 total specs with that name to be resolved first. Right now, they sort ~last, which leads to poor performance.

(This was one way of fixing some recent resolver issues, even though we've done the proper fixes in Molinillo for them)

### DIFF
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -329,6 +329,12 @@ module Bundler
 
   private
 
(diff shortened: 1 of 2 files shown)


indirect:
I like this a lot, and I believe I have seen many cases of frustrating resolutions or weird conflicts that this will improve.

@bundlerbot r+

PR #5450, opened 2017-02-19
[StubSpecification] Avoid loading all installed gemspecs

segiddins:
Call through to the stub instead of materializing the full remote_spec since all gems added to the local index are asked whether they are missing extensions or not

### DIFF
--- a/lib/bundler/stub_specification.rb
+++ b/lib/bundler/stub_specification.rb
@@ -15,6 +15,13 @@ module Bundler
       _remote_specification.to_yaml
     end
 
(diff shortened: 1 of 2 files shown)


indirect:
Whoa, nice catch! Also, great work adding a spec for this so we can’t accidentally do it again later. 👍

@bundlerbot r+

segiddins:
I might need to limit that test to only new rubygems versions

segiddins:
@bundlerbot r=indirect

PR #5454, opened 2017-02-21
Add a spec for not activating any gems during Bundler.setup

segiddins:
This is failing on ruby 2.5, since more of the stdlib is being gemified (fileutils in particular)

We can wait to merge until it passes if you'd like, @indirect, but this will likely also require changes in RubyGems

### DIFF
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -746,7 +746,13 @@ module Bundler
       end
 
       def backport_ext_builder_monitor
-        require "rubygems/ext"
+        # So we can avoid requiring "rubygems/ext" in its entirety
+        Gem.module_eval <<-RB, __FILE__, __LINE__ + 1
+          module Ext
+          end
+        RB
+
+        require "rubygems/ext/builder"
 
         Gem::Ext::Builder.class_eval do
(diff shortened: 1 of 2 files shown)


segiddins:
To ensure that Bundler can continue with the promise of providing the same set of gems across machines, we can't activate any gems, because then the user can't use different versions from what ships with their particular ruby (and their lockfile wont actually be respected)

indirect:
So my first guess at a solution is to vendor our own fileutils using Automatiek... what are you thinking?

segiddins:
Uh... That sounds janky? Especially since it isn't in its own repo yet. I was thinking we might just be able to avoid using it in code paths that require invokes?

indirect:
If that's possible, it sounds better. 👍

segiddins:
This should pass on everything but ruby-head now, want to see what's failing there

segiddins:
@indirect what do you want to do with this PR? Add exemptions for ruby-head so that entry stays green, and merge? Or merge as-is?

indirect:
@segiddins since it passed, I'm okay with merging it now... I guess it at least acts as an early warning system for future breaking changes to ruby head?

@bundlerbot r+

PR #5459, opened 2017-02-21
Avoid undefined Bundler::Plugin::API::Source exception

### LINKED ISSUE #5409: Uninitialized constant Bundler::Plugin::API::Source.
Seems this issue has reappeared in bundler 1.14.x in combination with 2.4.x rubygems. See:

https://travis-ci.org/dekellum/syncwrap/jobs/197069321#L202

Its not an issue with either rubygems 2.6.x (2.6.6 in one env.), or with bundler 1.13.7 as I'm able to demonstrate here:

https://travis-ci.org/dekellum/syncwrap/jobs/197079264

I think this is very similar to reports and PR #4981.

If the rake project can reproduce the issue, we don't know via travis, given that its pegged to bundler 1.13.x:

https://github.com/ruby/rake/blob/34f5e2e27d2c49431b7c84d0787eaf6afcddf794/.travis.yml

Cc: @hsbt 

But here are some other reports of this with bundler 1.14.3:

* skywinder/github-changelog-generator#474
* tmuxinator/tmuxinator#503

segiddins:
Fixes #5409, with a test!

### DIFF
--- a/lib/bundler/plugin/api/source.rb
+++ b/lib/bundler/plugin/api/source.rb
@@ -287,6 +287,12 @@ module Bundler
         def root
           Bundler.root
         end
+
+        # @private
+        # Returns true
+        def bundler_plugin_api_source?
+          true
+        end
       end
     end
   end
--- a/lib/bundler/rubygems_ext.rb
+++ b/lib/bundler/rubygems_ext.rb
@@ -42,7 +42,11 @@ module Gem
     attr_writer :full_gem_path unless instance_methods.include?(:full_gem_path=)
 
     def full_gem_path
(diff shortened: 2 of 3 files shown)


lib/bundler/plugin/api/source.rb
```diff
@@ -287,6 +287,12 @@ def gem_install_dir
         def root
           Bundler.root
         end
+
+        # @private
+        # Returns true
```
indirect: lol this comment 😂

indirect:
hooray, @bundlerbot r+

abhitrivedi:
Thanks! Is there a date when 1.14.6 will be available on rubygems?

segiddins:
Probably sometime this week

PR #5461, opened 2017-02-22
[LazySpecification] Handle when platform is nil in __materialize__

segiddins:
Because a nil platform is the same as RUBY


indirect:
Any chance we could get away with `sp = search.platform || RUBY; gp = platform || RUBY` instead of creating new objects?

segiddins:
@indirect I guess? I'll have to look into whether those platforms are always strings versus always platform objects versus mixed

indirect:
@segiddins ah, ok. I'm fine with this for now. 👍 

@bundlerbot r+

PR #5462, opened 2017-02-22
[Source::RubyGems] Print that we are fetching a gem

segiddins:
This is super helpful when you're on slow cafe wifi and downloading an 8MB gem such as nokogiri


segiddins:
@indirect r?

indirect:
Hmmmm, yeah this seems reasonable. @bundlerbot r+

segiddins:
@bundlerbot retry

Issue #5458, opened 2017-02-21
`bundle gem` fails when git is not available


jsch0221:
Unfortunately, an unexpected error occurred, and Bundler cannot continue.

First, try this link to see if there are any existing issue reports for this error:
https://github.com/bundler/bundler/search?q=No+such+file+or+directory+-+git&type=Issues

If there aren't any reports for this error yet, please create copy and paste the report template above into a new issue. Don't forget to anonymize any private data! The new issue form is located at:
https://github.com/bundler/bundler/issues/new
[sa16032@localhost ~]$ /usr/local/bin/bundle gem lemonade_stand
Creating gem 'lemonade_stand'...
--- ERROR REPORT TEMPLATE -------------------------------------------------------
# Error Report

## Questions

Please fill out answers to these questions, it'll help us figure out
why things are going wrong.

  I ran the command `/usr/local/bin/bundle gem lemonade_stand`

  I expected Bundler to create lemonade_stand/Gemfile, Rakefile, LICENSE.txt, READM.md, gitignore, lemonade_stand.gemspec,    lemonade_stand/lib/lemonade_stand.rb, version.rb, and  initialize git repo in /users/

  Instead, what happened was 'Unexpected error' and an error-report template

  I tried running the command as root, but there was no change. 

## Backtrace
```
Errno::ENOENT: No such file or directory - git
  /usr/local/share/gems/gems/bundler-1.14.4/lib/bundler/cli/gem.rb:32:in ``'
  /usr/local/share/gems/gems/bundler-1.14.4/lib/bundler/cli/gem.rb:32:in `run'
  /usr/local/share/gems/gems/bundler-1.14.4/lib/bundler/cli.rb:433:in `run'
  /usr/local/share/gems/gems/bundler-1.14.4/lib/bundler/vendor/thor/lib/thor/invocation.rb:126:in `invoke_command'
  /usr/local/share/gems/gems/bundler-1.14.4/lib/bundler/vendor/thor/lib/thor.rb:359:in `dispatch'
  /usr/local/share/gems/gems/bundler-1.14.4/lib/bundler/cli.rb:20:in `dispatch'
  /usr/local/share/gems/gems/bundler-1.14.4/lib/bundler/vendor/thor/lib/thor/base.rb:440:in `start'
  /usr/local/share/gems/gems/bundler-1.14.4/lib/bundler/cli.rb:11:in `start'
[...]

PR #5466, opened 2017-02-22
`bundle gem` check if git is installed

Shekharrajak:
Fixes https://github.com/bundler/bundler/issues/5458

### DIFF
--- a/lib/bundler/cli/gem.rb
+++ b/lib/bundler/cli/gem.rb
@@ -29,9 +29,11 @@ module Bundler
       constant_name = name.gsub(/-[_-]*(?![_-]|$)/) { "::" }.gsub(/([_-]+|(::)|^)(.|$)/) { $2.to_s + $3.upcase }
       constant_array = constant_name.split("::")
 
-      git_author_name = `git config user.name`.chomp
-      github_username = `git config github.user`.chomp
-      git_user_email = `git config user.email`.chomp
+      git_installed = Bundler.git_present?
+
+      git_author_name = git_installed ? `git config user.name`.chomp : ""
+      github_username = git_installed ? `git config github.user`.chomp : ""
+      git_user_email = git_installed ? `git config user.email`.chomp : ""
 
       config = {
         :name             => name,
--- a/spec/commands/newgem_spec.rb
+++ b/spec/commands/newgem_spec.rb
@@ -147,7 +147,7 @@ RSpec.describe "bundle gem" do
 
     context "git is not installed in the system" do
       before do
-        `rm "#{ENV["GIT_CONFIG"]}"` if File.exist?(ENV["GIT_CONFIG"])
+        FileUtils.rm ENV["GIT_CONFIG"].to_s if File.exist?(ENV["GIT_CONFIG"])
         reset!
         in_app_root
         bundle "gem #{gem_name}"
--- a/spec/commands/newgem_spec.rb
+++ b/spec/commands/newgem_spec.rb
@@ -144,6 +144,23 @@ RSpec.describe "bundle gem" do
         expect(bundled_app("test_gem/README.md").read).not_to include("github.com/bundleuser")
       end
     end
+
+    context "git is not installed in the system" do
+      before do
+        `rm "#{ENV["GIT_CONFIG"]}"` if File.exist?(ENV["GIT_CONFIG"])
+        reset!
+        in_app_root
+        bundle "gem #{gem_name}"
+        remove_push_guard(gem_name)
+      end
+
+      it "contribute URL set to [USERNAME]" do
+        expect(bundled_app("test_gem/README.md").read).to include("[USERNAME]")
+        expect(bundled_app("test_gem/README.md").read).not_to include("github.com/bundleuser")
+      end
+
+      it_should_behave_like "git config is absent"
+    end
   end
 
   it "generates a valid gemspec" do


lib/bundler/cli/gem.rb
```diff
@@ -29,6 +29,11 @@ def run
       constant_name = name.gsub(/-[_-]*(?![_-]|$)/) { "::" }.gsub(/([_-]+|(::)|^)(.|$)/) { $2.to_s + $3.upcase }
       constant_array = constant_name.split("::")
 
+      if !Bundler.git_present?
```
hmistry: use `unless` instead of `if !`

lib/bundler/cli/gem.rb
```diff
@@ -29,6 +29,11 @@ def run
...
+        Bundler.ui.error "You need to install git to be able to use bundle gem."
```
colby-swandale: Running `bundle gem` without git installed should work as expected for the user. We should just be setting empty values for `author`, `username` and `email` if we cannot source it from git.

colby-swandale:
We also need a spec testing this behaviour.

lib/bundler/cli/gem.rb
hmistry: @colby-swandale if that's the case, then this is not the right fix. I see your point in allowing it but it's not clear to me from bundler doc's if `bundle gem` is allowed to work without `git` or it requires `git`. Let's clarify this so the correct fix is implemented.

@indirect @segiddins any thoughts on git requirement for `bundle gem`?

Shekharrajak:
Thanks @colby-swandale  @hmistry for review. I have edited the PR. Now `git_author_name`, `git _username`, `git _user_email` will be set to empty string when git is not installed. 

So in the created `mygem.gemspec` , these lines will be present(similar as when git author, email is absent)

```
  spec.authors       = ["TODO: Write your name"]
  spec.email         = ["TODO: Write your email address"]
```

and in `README.md` file `[USERNAME]` will be present instead of github username. (Means this : `https://github.com/[USERNAME]/mygem` ).

I am still finding a way to write spec for this.

lib/bundler/cli/gem.rb
Shekharrajak: `git ` is needed only for fetching `username` `user_email` `git_author_name`, to write them in generated` mygem.gemspec` file and `README.md` file, which can be edited by the user afterward. 

So now I think it must not stop generating files (When git is not installed then just replace `username`, `email` with `["TODO: Write your name/email"]` in generated `mygem.gemspec` file)

spec/commands/newgem_spec.rb
```diff
@@ -144,6 +144,23 @@ def gem_skeleton_assertions(gem_name)
...
+        `rm "#{ENV["GIT_CONFIG"]}"` if File.exist?(ENV["GIT_CONFIG"])
+        reset!
+        in_app_root
+        bundle "gem #{gem_name}"
+        remove_push_guard(gem_name)
+      end
+
+      it "contribute URL set to [USERNAME]" do
+        expect(bundled_app("test_gem/README.md").read).to include("[USERNAME]")
+        expect(bundled_app("test_gem/README.md").read).not_to include("github.com/bundleuser")
```
colby-swandale: You probably will want to test the `gemspec` as well to make sure we're setting sensible defaults.

spec/commands/newgem_spec.rb
```diff
@@ -144,6 +144,23 @@ def gem_skeleton_assertions(gem_name)
...
+      end
+
+      it_should_behave_like "git config is absent"
```
Shekharrajak: @colby-swandale this line is using [shared example](https://github.com/bundler/bundler/pull/5466/files/4dec0e6f2c980c2b21fc96c987fbda5c6c351093#diff-3fbe0dfd8b4d505d057982532848c10eR66) to check the `gemspec` file. It must put `"TODO: Write your name"` , `"TODO: Write your email` in the place of author name, email (same as when git config is unset).

spec/commands/newgem_spec.rb
Shekharrajak: It is in [line 162](https://github.com/bundler/bundler/pull/5466#discussion_r102673468)

Shekharrajak:
Thanks @hmistry  for review.

spec/commands/newgem_spec.rb
```diff
@@ -144,6 +144,23 @@ def gem_skeleton_assertions(gem_name)
         expect(bundled_app("test_gem/README.md").read).not_to include("github.com/bundleuser")
       end
     end
+
+    context "git is not installed in the system" do
+      before do
+        `rm "#{ENV["GIT_CONFIG"]}"` if File.exist?(ENV["GIT_CONFIG"])
```
segiddins: please use `FileUtils` to `rm`

segiddins:
@colby-swandale want to re-review this?

colby-swandale:
Sorry, i didn't mean to give the impression to not merge this PR without my requested changes, only men't as suggestions. I'll keep that in mind in the future @segiddins

spec/commands/newgem_spec.rb
segiddins: This is the one blocker to merging

Shekharrajak: I will edit this PR soon.

segiddins:
@bundlerbot r+

Issue #4881, opened 2016-08-15
bundle outdated --source not working as expected


Aqualon:
The help for `bundle outdated` states:

`[--source=Check against a specific source]`

I expect that I could use this to check outdated gems not against our internal rubygems host defined in my Gemfile, but against rubygems.org by executing `bundle outdated --source rubygems.org`. 

But according to the output it still connects to our internal gem server and not `rubygems.org`.

Do I use it wrong, is this a bug or does the option mean something different than I would expect it to mean?

segiddins:
@Aqualon is there a Gemfile & Gemfile.lock we could use to reproduce the problem? Thanks!

Aqualon:
Since I can't give you our internal server, I confirmed it also happens other way around.

Example Gemfile:
```
source 'http://rubygems.org'

gem 'rack', '~> 1.6'
```

Calling `bundle outdated --source localhost` should try to access localhost, but only accesses rubygems.org

```
$ bundle outdated --source localhost
Fetching gem metadata from http://rubygems.org/
Fetching version metadata from http://rubygems.org/
Resolving dependencies...

Outdated gems included in the bundle:
  * rack (newest 2.0.1, installed 1.6.4, requested ~> 1.6) in group "default"
```

segiddins:
👍 good catch. It turns out we don't even have any specs for `bundle outdated --source`, so we'll need to work on defining what the expected behavior is for that option

Aqualon:
The main use case for me would be, that I can keep our internal gem server in the Gemfile and for checking if there are newer versions on rubygems.org, I could just add an option and would not have to edit the Gemfile.

But looking at the description of `--source` for `bundle update`, it may actually mean something different:
```
--source=<name>
  The  name of a :git or :path source used in the Gemfile(5). For instance, with a :git source
  of http://github.com/rails/rails.git, you would call bundle update --source rails
```

So maybe it's a feature request to add something like `--remote` option, that let's you specify the remote it checks for newer versions?

rrosenblum:
I have also run into this same issue. My use case sounds the same. We have a private gemserver that our `Gemfile` references. I would like to be able to check what is outdated against `https://rubygems.org'. My current work around is to have a second `Gemfile` that sources from RubyGems and evals in our main `Gemfile`.

Gemfile
```ruby
source <private_gem_server>

gem ...
```

Gemfile.rubygems
```ruby
source 'https://rubygems.org'

eval_gemfile 'Gemfile'
```

[...]

PR #5471, opened 2017-02-24
fix --source option not working for outdated and update commands

colby-swandale:
Relates to #4881

This fixes 2 related problems for the `outdated` and `update` commands in Bundler. Both commands have an option called `source` where you can specify a rubygems server not in the Gemfile for updated/outdated to use.

Currently the option for both commands is not working. See example:
```
source 'https://rubygems.org'

gem 'rack', '~> 1.6.5'
```
```
› bundle outdated --source https://mytestserver.com
Resolving dependencies...

  * rack (newest 2.0.1, installed 1.6.5, requested ~> 1.6.5) in groups "default"
```
```
› bundle update --source https://mytestserver.com
Resolving dependencies...
Using rack 1.6.5
Using bundler 1.14.0
Bundle updated!
Gems in the group test were not installed.
```

For both examples it should have been fetching the index/updating the gem from https://mytestserver.com.

This fix looks like part of a larger design issue, [specifically here](https://github.com/bundler/bundler/blob/master/lib/bundler.rb#L124).  Maybe work on that in the future.

I'm also not fully convinced of my patch in `dsl.rb` is the best solution for now, so i'm very open to suggestions.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -114,8 +114,8 @@ module Bundler
         @unlock[:gems] = @locked_specs.for(eager_unlock).map(&:name)
       end
 
-      if @unlock[:sources]
-        @unlock[:sources].each {|s| @sources.add_rubygems_remote(s) } if @unlock.is_a?(Hash) && @unlock[:sources]
+      if @unlock.is_a?(Hash) && @unlock[:sources]
+        @unlock[:sources].each {|s| @sources.add_rubygems_remote(s) }
       end
 
       @gem_version_promoter = create_gem_version_promoter
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -14,7 +14,8 @@ module Bundler
       :locked_gems,
       :platforms,
       :requires,
-      :ruby_version
+      :ruby_version,
+      :sources
     )
 
     # Given a gemfile and lockfile creates a Bundler definition
@@ -111,6 +112,10 @@ module Bundler
       unless @unlock[:lock_shared_dependencies]
         eager_unlock = expand_dependencies(@unlock[:gems])
         @unlock[:gems] = @locked_specs.for(eager_unlock).map(&:name)
+      end
+
+      if @unlock[:sources]
+        @unlock[:sources].each {|s| @sources.add_rubygems_remote(s) } if @unlock.is_a?(Hash) && @unlock[:sources]
       end
 
       @gem_version_promoter = create_gem_version_promoter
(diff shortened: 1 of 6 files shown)


spec/commands/update_spec.rb
```diff
@@ -635,6 +635,26 @@
...
+    it 'updates gems using the given source' do
+      gemfile <<-G
+        source "file://#{gem_repo1}"
+        gem 'rack'
+      G
+
+      bundle "install"
+
+      bundle "update --source file://#{gem_repo2}"
+      expect(out).to include "Fetching source index from file:#{gem_repo2}/"
```
colby-swandale: I realise a better test for this would be to actually check `the_bundle` instead

lib/bundler/dsl.rb
```diff
@@ -197,6 +197,7 @@ def github(repo, options = {})
     end
 
     def to_definition(lockfile, unlock)
+      unlock[:sources].each { |s| @sources.add_rubygems_remote(s) } if unlock.is_a?(Hash) && unlock[:sources]
```
segiddins: Maybe this belongs in the definition initializer?

lib/bundler/definition.rb
```diff
@@ -113,6 +114,10 @@ def initialize(lockfile, dependencies, sources, unlock, ruby_version = nil, opti
         @unlock[:gems] = @locked_specs.for(eager_unlock).map(&:name)
       end
 
+      if @unlock[:sources]
+        @unlock[:sources].each {|s| @sources.add_rubygems_remote(s) } if @unlock.is_a?(Hash) && @unlock[:sources]
```
olleolleolle: You already check `if unlock[:sources]` in the surrounding `if` block. The last bit of the suffix `if` can be removed, right?

spec/commands/outdated_spec.rb
```diff
@@ -595,6 +595,25 @@ def test_group_option(group = nil, gems_list_size = 1)
     it_behaves_like "patch version updates are detected"
   end
 
+  context "with --soruce option" do
```
olleolleolle: Typo: `--soruce` => `--source`

lib/bundler/definition.rb
```diff
@@ -113,6 +114,10 @@ def initialize(lockfile, dependencies, sources, unlock, ruby_version = nil, opti
         @unlock[:gems] = @locked_specs.for(eager_unlock).map(&:name)
       end
 
+      if @unlock.is_a?(Hash) && @unlock[:sources]
```
olleolleolle: Come to think of this line, line 112 with its `unless @unlock[:lock_shared_dependencies]` would have already failed if the `:[]` method was not available.

colby-swandale:
I have been asked why i closed this. It's because i misunderstood the behaviour of the `--source` option and this PR would have introduce undesirable behaviour.

mswiszcz:
@colby-swandale could you recreate this and add additional option as proposed in #4881 ? like `--remote` - it would be really useful for me as well

Issue #5447, opened 2017-02-18
Add `plugin uninstall` command


kerrizor:
`bundle plugin` currently only support a single command -- `install` It should also offer an `uninstall` option, to save people from resorting to mucking around with their `~/.bundle/plugins` directory by hand.

Shekharrajak:
I am interested to work on this issue. I think to uninstall plugin, we need to remove `commands`, `load_paths ` , `plugin_paths `  from the `./bundle/plugin/index` file.

PR #5473, opened 2017-02-25
plugin uninstall 

Shekharrajak:
Fixes https://github.com/bundler/bundler/issues/5447


lib/bundler/plugin.rb
```diff
@@ -41,6 +41,20 @@ def install(names, options)
       Bundler.ui.error "Failed to install plugin #{name}: #{e.message}\n  #{e.backtrace.join("\n ")}"
     end
 
+    # Uninstalls the given plugins
+    #
+    # @param [Array<String>] names the name of plugin to be uninstalled
```
olleolleolle: The list of names is plural? Perhaps `the name of plugin to be uninstalled` -> `the names of plugins to be uninstalled`?

Shekharrajak:
I will add test cases soon.

lib/bundler/cli/plugin.rb
```diff
@@ -19,5 +19,10 @@ class CLI::Plugin < Thor
     def install(*plugins)
       Bundler::Plugin.install(plugins, options)
     end
+
+    desc "uninstall PLUGINS", "Uninstall the plugins"
```
olleolleolle:
```ruby
     long_desc <<-D
       Uninstall given list of plugins. 
    D
```

lib/bundler/plugin.rb
Shekharrajak: Thanks I will edit it.

lib/bundler/cli/plugin.rb
Shekharrajak: Thanks I will add it.

lib/bundler/plugin.rb
```diff
@@ -41,6 +41,20 @@ def install(names, options)
...
+    def uninstall(names)
+      names.each do |name|
+        begin
+          uninstalled = index.remove_plugin(name)
+          Bundler.ui.info "Uninstalled plugin #{name}" if uninstalled
+        rescue => e
```
colby-swandale: I would avoid using a catchall `rescue` because Bundler has the error template.

lib/bundler/plugin/index.rb
```diff
@@ -106,6 +106,26 @@ def hook_plugins(event)
         @hooks[event] || []
       end
 
+      def remove_plugin(name)
+        begin
+          _plugin_path = plugin_path(name)
+        rescue TypeError => e
```
colby-swandale: I'm not sure why you're catching a `TypeError`. Maybe check if the path exists?

lib/bundler/plugin/index.rb
```diff
@@ -106,6 +106,26 @@ def hook_plugins(event)
...
+          Bundler.ui.error "plugin path of `#{name}` not found. \n" \
+          "It seems plugin `#{name}` is not installed. \n"\
+          "Error msg :#{e.message}"
+          exit 1
+        end
+
+        Bundler.rm_rf(_plugin_path)
+        @plugin_paths.delete_if{|k,_| k == name}
```
colby-swandale: I suggest encapsulating this logic into `plugin/index.rb`

lib/bundler/plugin.rb
Shekharrajak: Thanks, now I am using `rescue PluginError => e`

lib/bundler/plugin/index.rb
Shekharrajak: Yes, we can check using `index.installed?(name)` . Thanks for pointing it out.

lib/bundler/plugin/index.rb
Shekharrajak: I have added `remove_plugin_paths` method in `plugin/index.rb`. Thanks.

lib/bundler/cli/plugin.rb
hmistry: Is `PLUGINS` an optional param or do I need to list them?

My mental model when I see this command is to associate it as the opposite to `bundle install` for gems, as such it's not a natural naming convention. It would be better to associate this command more to `bundle plugins` like `bundle plugins uninstall` or `bundle plugins --uninstall` or something that tells the user this is for plugins and not gems. Making this distinction is important for good UX of bundler.

The short description can be improved like "Uninstalls the specified plugins".

The long description "Uninstall" needs to be "Uninstalls". Can you elaborate what happens if no `PLUGINS` is specified, i.e. will it remove all plugins? Is a list of plugins required? Anything else that a user should know about this command.

Shekharrajak: @hmistry , the command will be 
`bundle plugin uninstall plugin_name1 plugin_name2 ` . Right now it is not installing anything if `bundle plugin uninstall` is executed. I think there must be an option `--all` to uninstall all the plugins. I will edit the PR accordingly.

Shekharrajak:
I have added `--all` option. So `bundle plugin uninstall --all` will uninstall all the installed plugins. If no plugin is installed then it does nothing. 

I will add test case for the new options in next commit.

spec/bundler/plugin_spec.rb
```diff
@@ -80,6 +80,38 @@
...
+    it "remove plugin" do
+      allow(index).to receive(:remove_plugin).
+        with("new-plugin").once
+      allow(index).to receive(:installed?).
+        with("new-plugin").once
+      subject.uninstall ["new-plugin"], opts
+    end
+
+    context "multiple plugins" do
+      it do
```
colby-swandale: i don't encourage using `it` blocks without a description of what you're testing.

spec/bundler/plugin_spec.rb
```diff
@@ -80,6 +80,38 @@
...
+          with(["new-plugin", "another-plugin"]) do
+          {
+            "new-plugin" => spec,
+            "another-plugin" => spec2,
+          }
+        end.once
+
+        allow(index).to receive(:remove_plugin).twice
+        allow(index).to receive(:installed?).twice
+        subject.uninstall ["new-plugin", "another-plugin"], opts
```
colby-swandale: what are you testing for in this spec?

spec/bundler/plugin_spec.rb
```diff
@@ -80,6 +80,38 @@
     end
   end
 
+  describe "uninstall command" do
+    before do
+      allow(installer).to receive(:uninstall).with(["new-plugin"]) do
+        { "new-plugin" => spec }
+      end
+    end
+    let(:opts) { {} }
```
colby-swandale: i prefer to have space between `before` and `let` blocks

lib/bundler/cli/plugin.rb
```diff
@@ -19,5 +19,15 @@ class CLI::Plugin < Thor
...
+    long_desc <<-D
+      Uninstalls given list of plugins. To uninstall all the plugins, use -all option.
+   D
+    method_option "all", :type => :string, :default => nil, :banner =>
```
asutoshpalai: How about making `:type` as `:boolean`?

lib/bundler/plugin.rb
```diff
@@ -41,6 +41,32 @@ def install(names, options)
...
+      elsif names.empty?
+        Bundler.ui.error "No plugin to uninstall. Specify at least 1 plugin to uninstall.\n"\
+          "Use --all option to uninstall all the installed plugins."
+        exit 1
+      end
+      names.each do |name|
+        begin
+          if index.installed?(name)
+            uninstalled = index.remove_plugin(name)
+            Bundler.ui.info "Uninstalled plugin #{name}" if uninstalled
```
asutoshpalai: What if uninstall fails?

lib/bundler/plugin/index.rb
```diff
@@ -26,6 +25,7 @@ def initialize
         @sources = {}
         @hooks = {}
         @load_paths = {}
+        @installed_plugins = []
```
asutoshpalai: `@plugin_paths.keys` shall handle this. I don't think there is any need to explicitly store it.

lib/bundler/plugin/index.rb
```diff
@@ -106,6 +107,50 @@ def hook_plugins(event)
         @hooks[event] || []
       end
 
+      def all_plugins
+        @installed_plugins
```
asutoshpalai: `@plugin_paths.keys` ?

spec/bundler/plugin/index_spec.rb
```diff
@@ -105,6 +105,39 @@
     end
   end
 
+  describe "remove plugin" do
+    let(:plugin_path) { lib_path("new-plugin").to_s }
+    before do
+      index.remove_plugin_paths("new-plugin")
```
asutoshpalai: Shouldn't this be `index.remove_plugin` ?

asutoshpalai: Also, I can't find where you are registering the plugin with the index?

Instead of creating `:plugin_path` just override the `:sources`, `:commands` and `:hooks` with some values and then try removing it.

spec/bundler/plugin_spec.rb
```diff
@@ -80,6 +80,38 @@
     end
   end
 
+  describe "uninstall command" do
+    before do
+      allow(installer).to receive(:uninstall).with(["new-plugin"]) do
```
asutoshpalai: `installer` here is `Plugin::Installer` which is not being used for `uninstall`. So, I think we can skip this.

lib/bundler/plugin/index.rb
```diff
@@ -106,6 +107,50 @@ def hook_plugins(event)
...
+      end
+
+      def get_commands_of(name)
+        @commands.select {|_, v| v == name }
+      end
+
+      def remove_installed_plugin(name)
```
asutoshpalai: Shouldn't this and other remove helpers be made private?

segiddins:
@Shekharrajak ping on the comments?

Shekharrajak:
I will look into this soon.

colby-swandale:
ping @Shekharrajak

Shekharrajak:
If anyone trying to work on this issue, then feel free to take. If I see no one is currently working on this issue, I will do required changes after few weeks (I am really not able to manage time for this, because of college exams and gsoc project).

esasse:
Hey @Shekharrajak, do you still intend to work on this PR?

deivid-rodriguez:
Closing as superseded by #7409.

PR #5479, opened 2017-02-27
Update `Gemfile` docs' link targets for `:source`

myersjustinc:
These two links don't go anywhere in the browser, most likely because 01fc34bebd128d758733261f16d9e9a8523d81d4 changed the text of those (and other) headers.

As far as I know (based on a quick search with `/\w+-)/`), these are the only two links with this problem.

### DIFF
--- a/man/gemfile.5.ronn
+++ b/man/gemfile.5.ronn
@@ -30,7 +30,7 @@ Sources are checked for gems following the heuristics described in
 will print a warning after installing the gem indicating which source was used,
 and listing the other sources where the gem is available. A specific source can
 be selected for gems that need to use a non-standard repository, suppressing
-this warning, by using the [`:source` option](#SOURCE-source-) or a
+this warning, by using the [`:source` option](#SOURCE) or a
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 1.14.6 (unreleased)
+
+Documentation:
+
+  - Fix links about `:source` in `Gemfile` man page
```
segiddins: we don't document changes in the changelog until the release is made

myersjustinc: Ah, thanks for clarifying. Saw a mention of it in [the pull request guidelines](https://github.com/bundler/bundler/blob/master/doc/development/PULL_REQUESTS.md#changelogmd), so I went ahead and added it.

Should I remove that from this branch? And if so, would you prefer for me to do that in another commit or to amend my existing commit?

olleolleolle: (Amend the commit.)

myersjustinc: Amended!

segiddins:
Thanks so much!
@bundlerbot r+

PR #5486, opened 2017-03-01
Add command implementation

denniss:
@indirect  @segiddins Starting out with implementation of add command. Not very sure of what the differences are with `inject` other than this will have no ui and run install

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -268,6 +268,20 @@ module Bundler
       Binstubs.new(options, gems).run
     end
 
+    desc "add GEM VERSION", "Add gem to Gemfile and run bundle install"
+    long_desc <<-D
+      Adds the specified gem to Gemfile (if valid) and run 'bundle install' in one step.
+    D
+    method_option "version", :aliases => "-v", :type => :string
+    method_option "group", :aliases => "-g", :type => :string
+    method_option "source", :aliases => "-s", :type => :string
+
+    map "a" => "add"
+    def add(gem_name)
+      require "bundler/cli/add"
+      Add.new(options.dup, gem_name).run
+    end
+
     desc "outdated GEM [OPTIONS]", "list installed gems with newer versions available"
     long_desc <<-D
       Outdated lists the names and versions of gems that have a newer version available
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -276,7 +276,6 @@ module Bundler
     method_option "group", :aliases => "-g", :type => :string
     method_option "source", :aliases => "-s", :type => :string
 
-    map "a" => "add"
     def add(gem_name)
       require "bundler/cli/add"
       Add.new(options.dup, gem_name).run
--- a/lib/bundler/cli/add.rb
+++ b/lib/bundler/cli/add.rb
@@ -11,7 +11,14 @@ module Bundler
 
     def run
       version = @options[:version].nil? ? nil : @options[:version].split(",").map(&:strip)
-      dependency = Bundler::Dependency.new(@gem_name, version, @options)
+
+      begin
+        dependency = Bundler::Dependency.new(@gem_name, version, @options)
+      rescue Gem::Requirement::BadRequirementError => e
+        Bundler.ui.error(e.message)
+        return
+      end
+
       Injector.inject([dependency], :conservative_versioning => @options[:version].nil?) # Perform conservative versioning only when version is not specified
       Installer.install(Bundler.root, Bundler.definition)
     end
--- a/lib/bundler/cli/add.rb
+++ b/lib/bundler/cli/add.rb
@@ -12,12 +12,12 @@ module Bundler
     def run
       version = @options[:version].nil? ? nil : @options[:version].split(",").map(&:strip)
 
-      begin
-        dependency = Bundler::Dependency.new(@gem_name, version, @options)
-      rescue Gem::Requirement::BadRequirementError => e
-        Bundler.ui.error(e.message)
-        return
+      unless version.nil?
+        version.each do |v|
+          raise InvalidOption, "Invalid gem requirement pattern '#{v}'" unless Gem::Requirement::PATTERN =~ v.to_s
+        end
       end
+      dependency = Bundler::Dependency.new(@gem_name, version, @options)
 
       Injector.inject([dependency], :conservative_versioning => @options[:version].nil?) # Perform conservative versioning only when version is not specified
       Installer.install(Bundler.root, Bundler.definition)
--- a/lib/bundler/cli/add.rb
+++ b/lib/bundler/cli/add.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+require "bundler/cli/common"
+
+module Bundler
+  class CLI::Add
+    def initialize(options, gem_name)
+      @gem_name = gem_name
+      @options = options
+      @options[:group] = @options[:group].split(",").map(&:strip) if !@options[:group].nil? && !@options[:group].empty?
+    end
+
+    def run
+      dependency = Bundler::Dependency.new(@gem_name, @options[:version], @options)
+      Injector.inject([dependency], :conservative_versioning => @options[:version].nil?) # Perform conservative versioning only when version is not specified
+      Installer.install(Bundler.root, Bundler.definition)
+    end
+  end
+end
--- a/lib/bundler/cli/add.rb
+++ b/lib/bundler/cli/add.rb
@@ -10,7 +10,9 @@ module Bundler
     end
 
     def run
-      dependency = Bundler::Dependency.new(@gem_name, @options[:version], @options)
+      dependency = Bundler::Dependency.new(@gem_name, nil, @options)
+      dependency.instance_variable_set(:@requirement, Gem::Requirement.new(@options[:version].split(",").map(&:strip))) unless @options[:version].nil?
+
       Injector.inject([dependency], :conservative_versioning => @options[:version].nil?) # Perform conservative versioning only when version is not specified
       Installer.install(Bundler.root, Bundler.definition)
     end
--- a/lib/bundler/cli/add.rb
+++ b/lib/bundler/cli/add.rb
@@ -10,9 +10,8 @@ module Bundler
     end
 
     def run
-      dependency = Bundler::Dependency.new(@gem_name, nil, @options)
-      dependency.instance_variable_set(:@requirement, Gem::Requirement.new(@options[:version].split(",").map(&:strip))) unless @options[:version].nil?
-
+      version = @options[:version].nil? ? nil : @options[:version].split(",").map(&:strip)
+      dependency = Bundler::Dependency.new(@gem_name, version, @options)
       Injector.inject([dependency], :conservative_versioning => @options[:version].nil?) # Perform conservative versioning only when version is not specified
       Installer.install(Bundler.root, Bundler.definition)
     end
--- a/lib/bundler/injector.rb
+++ b/lib/bundler/injector.rb
@@ -69,12 +69,7 @@ module Bundler
         end
 
         if d.groups != Array(:default)
-          group =
-              if d.groups.size == 1
-                ", :group => #{d.groups.inspect}"
-              else
-                ", :groups => #{d.groups.inspect}"
-              end
+          group = d.groups.size == 1 ? ", :group => #{d.groups.inspect}" : ", :groups => #{d.groups.inspect}"
         end
 
         source = ", :source => \"#{d.source}\"" unless d.source.nil?
--- a/lib/bundler/injector.rb
+++ b/lib/bundler/injector.rb
@@ -76,8 +76,9 @@ module Bundler
                 ", :groups => #{d.groups.inspect}"
               end
         end
-        
-        source = ", :source => '#{d.source}'" unless d.source.nil?
+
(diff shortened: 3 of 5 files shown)


segiddins:
Can we please add tests for all of the use-cases outlined in the [RFC](https://github.com/bundler/rfcs/pull/4/files)?

indirect:
Cool, thanks for the initial implementation! As @segiddins noted, we'll want to expand the tests to cover the various examples mentioned in the RFC. It doesn't have to be done by you in this PR, but we'll also want to add a man page and full documentation for the add command before we ship it in a release. Good work getting things started. 👍

denniss:
Are you guys okay with 

`bundle add "rails" -v="~> 5.0.0" -s="https://gems.example.com" -g="development"`

Instead of (as specified in the RFC)

`bundle add "rails" -v "~> 5.0.0" -s "https://gems.example.com" -g "development"`

Note the equal sign. Any strong preference over the others?

segiddins:
I think Thor should handle making both work?

denniss:
You're right never tried the one without equal. Thanks

denniss:
@indirect @segiddins Can you guys elaborate more on the **conservative version update** requirement. May be an example and explanation of the motivation?

segiddins:
https://github.com/bundler/rfcs/blob/master/text/0000-bundle-add.md#conservative-version-addition-in-gemfile

denniss:
What do you guys think?

lib/bundler/cli.rb
```diff
@@ -266,6 +266,20 @@ def binstubs(*gems)
...
 
+    desc "add GEM VERSION", "Add gem to Gemfile and run bundle install"
+    long_desc <<-D
+      Adds the specified gem to Gemfile (if valid) and run 'bundle install' in one step.
+    D
+    method_option "version", :aliases => "-v", :type => :string
+    method_option "group", :aliases => "-g", :type => :string
+    method_option "source", :aliases => "-s", :type => :string
+
+    map "a" => "add"
```
segiddins: I think this is unnecessary

lib/bundler/injector.rb
```diff
@@ -47,17 +48,31 @@ def inject(gemfile_path, lockfile_path)
 
   private
 
-    def new_gem_lines
+    def conservative_version(version)
```
segiddins: this needs to return a pre version if the current version is a pre-release

spec/commands/add_spec.rb
```diff
@@ -28,6 +30,18 @@
       expect(bundled_app("Gemfile").read).to match(/gem 'baz', '~> 1.2'/)
       expect(the_bundle).to include_gems "baz 1.2.3"
     end
+
+    it "version requirement becomes ~> major.minor.patch.pre when resolved version is < 1.0" do
```
denniss: @segiddins Is this the correct expectation when a pre-release version is resolved?

segiddins: either this or you could just blindly add a `.a` after the numeric segments

denniss:
@segiddins When you have a moment, I think this one is ready for another review.

lib/bundler/injector.rb
```diff
@@ -47,17 +48,33 @@ def inject(gemfile_path, lockfile_path)
...
+
+    def build_gem_lines(conservative_versioning)
       @new_deps.map do |d|
         name = "'#{d.name}'"
-        requirement = ", '#{d.requirement}'"
+
+        requirement = if conservative_versioning
+          ", '#{conservative_version(@definition.specs[d.name][0])}'"
+        else
+          ", #{d.requirement.as_list.map {|k| "'#{k}'" }.join(", ")}"
```
segiddins: this could be `d.requirement.as_list.map(&:dump).join(", ")`

denniss: fixed

denniss:
@segiddins I don't think I have anything more to add to this. When you get a chance, I think this one is ready for another look. Thanks!

lib/bundler/injector.rb
```diff
@@ -27,17 +27,18 @@ def inject(gemfile_path, lockfile_path)
...
 
       # add new deps to the end of the in-memory Gemfile
-      builder.eval_gemfile("injected gems", new_gem_lines) if @new_deps.any?
+      # Set conservative versioining to false because we want to let the resolver resolve the version first
+      builder.eval_gemfile("injected gems", build_gem_lines(false)) if @new_deps.any?
 
       # resolve to see if the new deps broke anything
-      definition = builder.to_definition(lockfile_path, {})
-      definition.resolve_remotely!
+      @definition = builder.to_definition(lockfile_path, {})
```
segiddins: why is it necessary to change this to an ivar?

lib/bundler/injector.rb
```diff
@@ -47,17 +48,33 @@ def inject(gemfile_path, lockfile_path)
...
+
+      prerelease_suffix = version.to_s.gsub(version.release.to_s, "") if version.prerelease?
+      "~> #{segments[0..seg_end_index].join(".")}#{prerelease_suffix}"
+    end
+
+    def build_gem_lines(conservative_versioning)
       @new_deps.map do |d|
-        name = "'#{d.name}'"
-        requirement = ", '#{d.requirement}'"
+        name = "\"#{d.name}\""
```
segiddins: please use `.dump` instead of manually escaping strings

denniss: I need to reference this in `def build_gem_lines(conservative_versioning)`

lib/bundler/injector.rb
denniss: I need to reference this in `def build_gem_lines(conservative_versioning)`

lib/bundler/injector.rb
denniss: done

lib/bundler/cli/add.rb
```diff
@@ -0,0 +1,19 @@
...
+    def initialize(options, gem_name)
+      @gem_name = gem_name
+      @options = options
+      @options[:group] = @options[:group].split(",").map(&:strip) if !@options[:group].nil? && !@options[:group].empty?
+    end
+
+    def run
+      version = @options[:version].nil? ? nil : @options[:version].split(",").map(&:strip)
+      dependency = Bundler::Dependency.new(@gem_name, version, @options)
+      Injector.inject([dependency], :conservative_versioning => @options[:version].nil?) # Perform conservative versioning only when version is not specified
```
hmistry: Here you've coupled `bundle add` functionality to `bundle inject` internally. This is ok until we need to diverge the 2 sets of functionalities. I presume the changes you made to `Injector` class does not change the behavior or have any side effects on `bundle inject` and it's designated behavior.

When I wrote the RFC on this feature, it was made clear to me that `bundle inject`'s behavior should remain unchanged.

spec/commands/add_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  describe "with --source" do
+    it "adds dependency with specified source" do
+      bundle "add 'foo' --source='file://#{gem_repo2}'"
+      expect(bundled_app("Gemfile").read).to match(%r{gem "foo", "~> 2.0", :source => 'file:\/\/#{gem_repo2}'})
+      expect(the_bundle).to include_gems "foo 2.0"
+    end
+  end
+end
```
hmistry: Overall these tests look good. 

I don't see any tests for the short form options i.e. `-v, -g, -s`. It would be good to add a test for them incase there's a future regression.

Also, it's probably good to do a combo options command test for sanity check e.g. `bundle add foo --version=2.0 --group=test --source=file://...`.

Lastly, I don't see any negative testing like what happens if I tried to `bundle add` a gem that's not in the repo (think misspelled), or a bad version format, or unreachable source, or ... Hopefully it fails gracefully with good error messages.

segiddins: @denniss if you could address these testing comments, I think this should be good to go!

denniss: Thanks for the feedback guys and good point on the negative testing @hmistry

lib/bundler/cli/add.rb
```diff
@@ -11,7 +11,14 @@ def initialize(options, gem_name)
 
     def run
       version = @options[:version].nil? ? nil : @options[:version].split(",").map(&:strip)
-      dependency = Bundler::Dependency.new(@gem_name, version, @options)
+
+      begin
+        dependency = Bundler::Dependency.new(@gem_name, version, @options)
+      rescue Gem::Requirement::BadRequirementError => e
+        Bundler.ui.error(e.message)
+        return
```
segiddins: this should raise `InvalidOption` and give more context about what's invalid

denniss: Shouldn't the messages passed by `Gem::Requirement::BadRequirementError` be descriptive enough? It would output something like: `Illformed requirement ["~> 1 . 0"]`

spec/commands/add_spec.rb
```diff
@@ -79,4 +79,28 @@
...
+
+  it "using combination of short form options works like long form" do
+    bundle "add 'foo' -s='file://#{gem_repo2}' -g='development' -v='~>1.0'"
+    expect(bundled_app("Gemfile").read).to match(%r{gem "foo", "~> 1.0", :group => \[:development\], :source => 'file:\/\/#{gem_repo2}'})
+    expect(the_bundle).to include_gems "foo 1.1"
+  end
+
+  it "shows error message when version is not formatted correctly" do
+    bundle "add 'foo' -v='~>1 . 0'"
+    expect(out).to match('Illformed requirement ["~>1 . 0"]')
```
hmistry: Change to a user friendly error... something like "Invalid version specified. Please enter a valid version."

spec/commands/add_spec.rb
```diff
@@ -79,4 +79,28 @@
...
+  end
+
+  it "shows error message when gem cannot be found" do
+    bundle "add 'werk_it'"
+    expect(out).to match("Could not find gem 'werk_it' in any of the gem sources listed in your Gemfile.")
```
hmistry: Does this error happen if I specified the source in the add command? If so, then we should have a more generic error like `"Could not find gem 'werk_it' in any of the gem sources specified. Please check if it's a valid gem and whether it's available in the sources specified."`

spec/commands/add_spec.rb
```diff
@@ -79,4 +79,28 @@
...
+  end
+
+  it "shows error message when source cannot be reached" do
+    bundle "add 'baz' --source='http://badhostasdf'"
+    expect(out).to include("Could not reach host badhostasdf. Check your network connection and try again.")
```
hmistry: Improve error - `"Could not reach host 'badhostasdf'. Check your network connection is working and whether the source url is valid and try again."`

spec/commands/add_spec.rb
```diff
@@ -79,4 +79,28 @@
...
+
+    bundle "add 'baz' --source='file://does/not/exist'"
+    expect(out).to include("Could not fetch specs from file://does/not/exist/")
```
hmistry: Use consistent user terminology - `"Could not fetch gem from file://does/not/exist/"`.
User doesn't know what a spec is, only gems.

hmistry:
@denniss Thanks for implementing feature. I had it 70% done last year but didn't get chance to finish it. Glad to see it completed. 😃

lib/bundler/cli/add.rb
```diff
@@ -0,0 +1,26 @@
...
+      @options = options
+      @options[:group] = @options[:group].split(",").map(&:strip) if !@options[:group].nil? && !@options[:group].empty?
+    end
+
+    def run
+      version = @options[:version].nil? ? nil : @options[:version].split(",").map(&:strip)
+
+      begin
+        dependency = Bundler::Dependency.new(@gem_name, version, @options)
+      rescue Gem::Requirement::BadRequirementError => e
```
segiddins: btw, this constant doesn't exist on old rubygems versions, hence the error on CI

segiddins:
This will need a rebase, we try to avoid putting large merge commits into history

segiddins:
@indirect please review? I'd like to merge this once the few outstanding comments are addressed

denniss:
Apologies for being unresponsive. I will try to get to these ASAP.

indirect:
Thanks! and no worries, life happens. 🙂

spec/commands/add_spec.rb
denniss: I am piggy-backing on the functionality provided by `Bundler::Definition#resolve_remotely!`. The error message comes from that method.

denniss:
@segiddins Fixed the `Gem::Requirement::BadRequirementError`, ci reported a failure but can't figure out what's wrong. The detail is green. Can we try re-running?

spec/commands/add_spec.rb
denniss: Similar, just piggybacking on existing method call.

lib/bundler/cli/add.rb
denniss: Fixed this.

spec/commands/add_spec.rb
denniss: Added a test for this.

segiddins:
This is wonderful, thank you _so_ much.

@bundlerbot r+

spec/commands/add_spec.rb
hmistry: @denniss I see. Then never mind.

Issue #5009, opened 2016-09-21
Remove global `frozen` when running `install --no-deployment`


trevorturk:
This is a minor issue that I'm happy to work up a pull request for, but I'm wondering if anyone has opinions on how best to proceed.

The issue is that `bundle install --no-deployment` doesn't remove the global `frozen` flag if present. This can be worked around easily by using `bundle config --delete frozen` but Bundler suggests the problem will be solved by running `bundle install --no-deployment`.

Here's an example of how I ran into this issue:

```
$ bundle config --global frozen 1

$ bundle env
    Bundler   1.13.1
    Rubygems  2.2.5
    Ruby      2.1.10p492 (2016-04-01 revision 54464) [x86_64-darwin15.0]
    Git       2.10.0

Bundler settings

    frozen
      Set for the current user (/Users/trevorturk/.bundle/config): "1"

$ bundle install

You are trying to install in deployment mode after changing
your Gemfile. Run `bundle install` elsewhere and add the
updated Gemfile.lock to version control.

If this is a development machine, remove the /Users/trevorturk/Code/clearbit/company/Gemfile freeze 
by running `bundle install --no-deployment`.

$ bundle install --no-deployment

[...same error...]

$ bundle config --delete frozen
$ bundle install

[...success...]
```

I believe the best fix would be to delete the local and global setting via: https://github.com/bundler/bundler/blob/master/lib/bundler/cli/install.rb#L48-L52 which might mean adjusting `Bundler::Settings#delete` to also delete global settings here https://github.com/bundler/bundler/blob/master/lib/bundler/settings.rb#L99-L101 or to introduce a new `Bundler::Settings#delete_global` or such and delete the global setting when install is called with `--no-deployment`. 

Another possible fix might be adjusting the error message to suggest running `bundle config --delete frozen` instead of `bundle install --no-deployment` but that would leave out the `options[:system] = true` tweak, and I'm not quite sure if that would cause problems.

segiddins:
The only thing that should be touching the global config is `bundle config --global`, in my opinion. I think maybe the solution here is to just check where the `frozen` setting is coming from, and suggest deleting it if it's not coming from the app config?

trevorturk:
You can see from the code snippet I pasted above that I ran `bundle config --global frozen 1` -- this was part of my attempt to debug #5006 because that's what the `Dockerfile` I inherited contained. 

[...]

PR #5502, opened 2017-03-06
Suggest different command depending on where 'frozen' config is set

denniss:
Fixes: https://github.com/bundler/bundler/issues/5009

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -409,8 +409,10 @@ module Bundler
              "updated #{Bundler.default_lockfile.relative_path_from(SharedHelpers.pwd)} to version control."
 
       unless explicit_flag
+
+        suggested_command = Bundler.settings.locations("frozen")[:global] == "1" ? "bundle config --delete frozen" : "bundle install --no-deployment"
         msg << "\n\nIf this is a development machine, remove the #{Bundler.default_gemfile} " \
-               "freeze \nby running `bundle install --no-deployment`."
+               "freeze \nby running `#{suggested_command}`."
       end
 
       added =   []
--- a/spec/commands/update_spec.rb
+++ b/spec/commands/update_spec.rb
@@ -159,7 +159,15 @@ RSpec.describe "bundle update" do
       bundle "update"
 
       expect(out).to match(/You are trying to install in deployment mode after changing.your Gemfile/m)
+      expect(out).to match(/freeze \nby running `bundle install --no-deployment`./m)
       expect(exitstatus).not_to eq(0) if exitstatus
+    end
+
+    it "should suggest different command when frozen is set globally" do
+      bundler "config --global frozen 1"
+      bundle "update"
+      expect(out).to match(/You are trying to install in deployment mode after changing.your Gemfile/m)
+      expect(out).to match(/freeze \nby running `bundle config --delete frozen`./m)
     end
   end
 


lib/bundler/definition.rb
```diff
@@ -409,8 +409,10 @@ def ensure_equivalent_gemfile_and_lockfile(explicit_flag = false)
              "updated #{Bundler.default_lockfile.relative_path_from(SharedHelpers.pwd)} to version control."
 
       unless explicit_flag
+
+        suggest_command = Bundler.settings.locations("frozen")[:global] == "1" ? "bundle config --delete frozen" : "bundle install --no-deployment"
```
segiddins: Also, won't we need a `--global` if it is indeed set globally?

denniss: That's one other thing that I saw. I need to re-verify this but it looks like `bundle config --delete frozen` deletes BOTH local and global config.

denniss: May be this is expected?

`➜  bundler-dev cat ~/.bundle/config
---
BUNDLE_DISABLE_CHECKSUM_VALIDATION: "true"
BUNDLE_FROZEN: "1"
➜  bundler-dev cat .bundle/config
---
BUNDLE_PATH: "vendor/bundle"
BUNDLE_DISABLE_SHARED_GEMS: "true"
BUNDLE_FROZEN: "1"
➜  bundler-dev dbundle config --delete frozen
➜  bundler-dev cat ~/.bundle/config
---
BUNDLE_DISABLE_CHECKSUM_VALIDATION: "true"
➜  bundler-dev cat .bundle/config
---
BUNDLE_PATH: "vendor/bundle"
BUNDLE_DISABLE_SHARED_GEMS: "true"
➜  bundler-dev`

denniss: Fixed

segiddins: uh, I'm not entirely sure? @indirect ?

indirect: IIRC, deleting both was our way of punting on the issue. It seems like terrible UI to require specific combinations of multiple flags. So far, it has worked. 😬  We might need to change it if there is an actual usecase that requires deleting one and not the other, though.

segiddins: Cool, just wanted to check

denniss:
@segiddins My bad for force-pushing. `bundle config --delete` appears to delete both local and global config.

```
➜  bundler-dev cat ~/.bundle/config
---
BUNDLE_DISABLE_CHECKSUM_VALIDATION: "true"
BUNDLE_FROZEN: "1"
➜  bundler-dev cat .bundle/config
---
BUNDLE_PATH: "vendor/bundle"
BUNDLE_DISABLE_SHARED_GEMS: "true"
BUNDLE_FROZEN: "1"
➜  bundler-dev dbundle config --delete frozen
➜  bundler-dev cat ~/.bundle/config
---
BUNDLE_DISABLE_CHECKSUM_VALIDATION: "true"
➜  bundler-dev cat .bundle/config
---
BUNDLE_PATH: "vendor/bundle"
BUNDLE_DISABLE_SHARED_GEMS: "true"
```

denniss:
@segiddins I also don't have anything more to add to this one. When you have a moment, can you please take another look?

segiddins:
Excellent!
@bundlerbot r+

segiddins:
@bundlerbot retry

trevorturk:
🤘

Issue #4509, opened 2016-05-03
Add bundle pristine command


lynnco:
From: https://github.com/bundler/bundler-features/issues/5

allenzhao:
I'm trying to implement this feature, and after reading discussion @ bundler/bundler-features#5, and this feature can be implemented by adding support for git gems, and it supports other types of gems by just invoking ` bundle exec gem pristine` right? @indirect @segiddins @RochesterinNYC

RochesterinNYC:
@allenzhao that is correct, though gems that are specified via `path` will need to be handled as well.

indirect:
On an installed gem, we can re-install from the .gem file, and on a git repo we can `checkout --force`, but for a path we don't have any original files to restore—I think calling `pristine` on a path gem should print an error and exit.

RochesterinNYC:
Hm, so say for a `Gemfile` with 100 regular and git gems and 1 path gem, how will the user use `bundle pristine` on that `Gemfile`? I think we should somehow support this kind of use case without requiring the user to temporarily remove or comment out the path gem and run `bundle pristine`.

RochesterinNYC:
Actually, any reason we wouldn't just print a warning (ex. `"path-gem" gem specified by path "../../path-gem" could not be restored as path gems cannot be restored`) rather than erroring out?

indirect:
Oh yeah! If you're `pristine`ing more than one gem, just a warning seems fine. I was thinking of a single gem. :)

RochesterinNYC:
Oh, gotcha. Yup, agree on the erroring out for the case of single `path` gem `pristine`ing.

allenzhao:
I was going over gem pristine's code and [command options](http://guides.rubygems.org/command-reference/#gem-pristine) and wondering if we need everything similar or just a few?
Any thoughts/comments?

RochesterinNYC:
Don't think we need `-​-all` as invoking `bundle pristine` without a gem name passed will basically be invoking on all the gems in the bundle.
Not sure how the `--version` indicator would apply as the `Gemfile.lock` would already indicate the locked version to restore.
I see the value in the others.

RochesterinNYC:
@segiddins @indirect input on this would be helpful as I'm not as familiar with Rubygems as y'all.

lynnco:
ping @segiddins @indirect

> I was going over gem pristine's code and command options and wondering if we need everything similar or just a few? Any thoughts/comments?

[...]

PR #5503, opened 2017-03-07
Implementation of `bundle pristine`

denniss:
Initial implementation of bundle pristine

Addressing:
https://github.com/bundler/bundler/issues/4509
https://github.com/bundler/bundler-features/issues/5

### DIFF
--- a/lib/bundler/cli/pristine.rb
+++ b/lib/bundler/cli/pristine.rb
@@ -7,6 +7,10 @@ module Bundler
       ::Bundler.load.specs.each do |spec|
         gem_name = "#{spec.name} (#{spec.version}#{spec.git_version})"
 
+        if spec.platform != Gem::Platform::RUBY
+          gem_name += " (#{spec.platform.nil? ? "No platform" : spec.platform})"
+        end
+
         case spec.source
           when Source::Rubygems
             cached_gem = spec.cache_file
--- a/lib/bundler/cli/pristine.rb
+++ b/lib/bundler/cli/pristine.rb
@@ -7,26 +7,24 @@ module Bundler
       ::Bundler.load.specs.each do |spec|
         gem_name = "#{spec.name} (#{spec.version}#{spec.git_version})"
 
-        if spec.source.instance_of? Source::Path
-          ::Bundler.ui.warn("Cannot pristine #{gem_name}. Gem is sourced from local path.")
-          next
+        case spec.source
+          when Source::Rubygems
+            cached_gem = spec.cache_file
+            unless File.exist?(cached_gem)
+              # TODO: Refetch from ruby gem?
+              ::Bundler.ui.error("Failed to pristine #{gem_name}. Cached gem #{cached_gem} does not exist.")
+              next
+            end
+
+            spec.source.install(spec, :force => true)
+          when Source::Git
+            git_source = spec.source
+            git_source.remote!
+            git_source.install(spec, :force => true)
+          else
+            ::Bundler.ui.warn("Cannot pristine #{gem_name}. Gem is sourced from local path.")
         end
 
-        if spec.source.instance_of? Source::Rubygems
-          cached_gem = spec.cache_file
-          unless File.exist?(cached_gem)
-            # TODO: Refetch from ruby gem?
-            ::Bundler.ui.error("Failed to pristine #{gem_name}. Cached gem #{cached_gem} does not exist.")
-            next
-          end
-
-          spec.source.install(spec, :force => true)
-
-        elsif spec.source.instance_of? Source::Git
-          git_source = spec.source
-          git_source.remote!
-          git_source.install(spec, :force => true)
-        end
       end
     end
   end
--- a/lib/bundler/cli/pristine.rb
+++ b/lib/bundler/cli/pristine.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+require "bundler/cli/common"
+
+module Bundler
+  class CLI::Pristine
+    def run
+
+      ::Bundler.load.specs.each do |spec|
+
+        gem_name = "#{spec.name} (#{spec.version}#{spec.git_version})"
+
+        if spec.source.is_a?(Source::Path)
+          ::Bundler.ui.warn("Cannot pristine #{gem_name} Gem is sourced from path.")
+          next
+        end
+
+        if spec.source.is_a?(Source::Rubygems)
+          cached_gem = spec.cache_file
+          unless File.exists?(cached_gem)
+            ::Bundler.ui.error("Failed to pristine #{gem_name}. Cached gem #{cached_gem} does not exist.")
+            next
+          end
+
+          installer = Gem::Installer.at(cached_gem,
+                                        :wrappers => true,
+                                        :force => true,
+                                        :install_dir => spec.base_dir,
+                                        :build_args => spec.build_args)
+          installer.install
+
+        elsif spec.source.is_a?(Source::Git)
+
+        end
+
+
+      end
+    end
+  end
+end
--- a/lib/bundler/cli/pristine.rb
+++ b/lib/bundler/cli/pristine.rb
@@ -6,16 +6,12 @@ module Bundler
     def run
       ::Bundler.load.specs.each do |spec|
         gem_name = "#{spec.name} (#{spec.version}#{spec.git_version})"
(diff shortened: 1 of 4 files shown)


lib/bundler/cli/pristine.rb
```diff
@@ -9,30 +9,29 @@ def run
...
 
-          installer = Gem::Installer.at(cached_gem,
-                                        :wrappers => true,
-                                        :force => true,
-                                        :install_dir => spec.base_dir,
-                                        :build_args => spec.build_args)
-          installer.install
-
-        elsif spec.source.is_a?(Source::Git)
+          Gem::Installer.at(cached_gem,
```
segiddins: `spec.source.install(spec, :force => true)`

lib/bundler/cli/pristine.rb
```diff
@@ -9,30 +9,29 @@ def run
...
+                            :wrappers => true,
+                            :force => true,
+                            :install_dir => spec.base_dir,
+                            :build_args => spec.build_args).install
 
+        elsif spec.source.instance_of? Source::Git
+          `git -C "#{spec.full_gem_path}" checkout --force`
```
segiddins: `spec.source.install(spec, :force => true)`

simi:
What about to wrap [RubyGems pristine code](https://github.com/rubygems/rubygems/blob/0749715e4bd9e7f0fb631a352ddc649574da91c1/lib/rubygems/commands/pristine_command.rb#L83) first into separated reusable class and just reuse it here for per RubyGem?.

I can help here if needed.

denniss:
@simi That was I first thought but there seems to be a cleaner approach as suggested by @segiddins

lib/bundler/cli/pristine.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+require "bundler/cli/common"
+
+module Bundler
+  class CLI::Pristine
+    def run
+      ::Bundler.load.specs.each do |spec|
+        gem_name = "#{spec.name} (#{spec.version}#{spec.git_version})"
+
+        if spec.source.instance_of? Source::Path
```
segiddins: please change this to switch over `spec.source`, first matching `Rubygems`, then `Git`, then an `else` that prints the warning

spec/commands/pristine_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      gem "bar", :path => "#{lib_path("bar")}"
+    G
+
+    bundle "install"
+  end
+
+  context "when sourced from Rubygems" do
+    it "reverts using cached .gem file" do
+      spec = Bundler.definition.specs["weakling"].first
+      changes_txt = "#{spec.full_gem_path}/lib/changes.txt"
```
segiddins: `changes_txt = Pathname.new(spec.full_gem_path).join("lib/changes.txt")`

spec/commands/pristine_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+
+      FileUtils.touch(changes_txt)
+      expect(File.exist?(changes_txt)).to be_truthy
```
segiddins: `expect(changes_txt).to be_file`

segiddins: and the same thing on all the other expectations as well

spec/commands/pristine_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    end
+  end
+
+  context "when sourced from git repo" do
+    it "reverts by resetting to current revision`" do
+      spec = Bundler.definition.specs["foo"].first
+      changed_file = "#{spec.full_gem_path}/lib/foo.rb"
+      diff = "#Pristine spec changes"
+
+      `echo '#Pristine spec changes' >> #{changed_file}`
```
segiddins: please use `File.open` instead of a shell here

spec/commands/pristine_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      expect(File.read(changed_file)).to include(diff)
+
+      bundle "pristine"
+      expect(File.read(changed_file)).to_not include(diff)
+    end
+  end
+
+  context "when sourced from path" do
```
segiddins: please also test when using `gemspec` in the Gemfile

lib/bundler/cli/pristine.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+require "bundler/cli/common"
+
+module Bundler
+  class CLI::Pristine
+    def run
+      ::Bundler.load.specs.each do |spec|
+        gem_name = "#{spec.name} (#{spec.version}#{spec.git_version})"
```
segiddins: also include the platform, if it is not `Gem::Platform::RUBY` or `nil`

lib/bundler/cli/pristine.rb
```diff
@@ -0,0 +1,33 @@
...
+        if spec.source.instance_of? Source::Path
+          ::Bundler.ui.warn("Cannot pristine #{gem_name}. Gem is sourced from local path.")
+          next
+        end
+
+        if spec.source.instance_of? Source::Rubygems
+          cached_gem = spec.cache_file
+          unless File.exist?(cached_gem)
+            # TODO: Refetch from ruby gem?
+            ::Bundler.ui.error("Failed to pristine #{gem_name}. Cached gem #{cached_gem} does not exist.")
```
segiddins: this should also cause the command to exit with a non-zero status code, imho

spec/commands/pristine_spec.rb
denniss: I have added a few more specs to handle this case. Can you take a look ? Thanks!

lib/bundler/cli/pristine.rb
denniss: Done

colby-swandale:
@denniss can you fix the breaking spec please 😄

denniss:
@colby-swandale Tried running the spec with `ruby-1.9.3-p551` and the spec is passing. Can you try re-running Travis build?

colby-swandale:
Sorry for the delay, i did re-run that job but it appears there is actually a problem with that version of ruby.

segiddins:
It looks like it's only failing on very old RubyGems versions. You can use `RGV=VERSION bin/rspec` to test against them locally

lib/bundler/cli/pristine.rb
```diff
@@ -7,12 +7,12 @@ def run
       ::Bundler.load.specs.each do |spec|
         gem_name = "#{spec.name} (#{spec.version}#{spec.git_version})"
 
-        if spec.source.instance_of? Source::Path
-          ::Bundler.ui.warn("Cannot pristine #{gem_name}. Gem is sourced from local path.")
-          next
+        if spec.platform != Gem::Platform::RUBY
+          gem_name += " (#{spec.platform.nil? ? "No platform" : spec.platform})"
```
segiddins: a nil platform is the same as `Gem::Platform::RUBY`

lib/bundler/cli/pristine.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+require "bundler/cli/common"
+
+module Bundler
+  class CLI::Pristine
+    def run
+      ::Bundler.load.specs.each do |spec|
+        gem_name = "#{spec.name} (#{spec.version}#{spec.git_version})"
+
+        if spec.platform != Gem::Platform::RUBY
```
segiddins: `if !spec.platform.nil? || spec.platform == Gem::Platform::RUBY`

denniss: Do you mean
`if !spec.platform.nil? && spec.platform != Gem::Platform::RUBY` ?

So that it would only show message for non-ruby platform gems?

denniss:
Looks like `RGV=v1.5.3`'s `Gem::Installer#install` does not `rm -rf` the gem folder. The later versions do. I added a line to manually `FileUtils.rm_rf` before calling install.

lib/bundler/cli/pristine.rb
segiddins: yeah

segiddins:
This is _great_

@bundlerbot r+

indirect:
woo 🎉🎉🎉

PR #5504, opened 2017-03-07
[CLI] Dont reset every time the CLI is initialized

### LINKED ISSUE #5500: Nested calls to bundle exec resolve xmlrpc incorrectly
This is an odd edge case that occurs in `cucumber-rails` integration tests, where `bundle exec rake cucumber` is called inside an integration test that is itself run through `bundle exec cucumber`.

Here is a gist demonstrating the original case:
https://gist.github.com/mvz/39a2408005f16174f5831eec4011910b

@rhenium reduced this to a simple demonstration here: https://github.com/ruby/openssl/issues/103#issuecomment-283293385

Here is a gist with that demonstration as a Dockerfile:
https://gist.github.com/mvz/2b603766f5de391f5c23393f766146c8

In that gist, I expect the output of `bundle exec ./a.rb` and `bundle exec bundle exec ./a.rb` to be the same, but it is not.

segiddins:
Fixes #5500 

- [x] Tests

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -442,6 +442,12 @@ EOF
     end
 
     def reset!
+      reset_paths!
+      Plugin.reset!
+      reset_rubygems!
+    end
+
+    def reset_paths!
       @root = nil
       @settings = nil
       @definition = nil
@@ -451,9 +457,9 @@ EOF
       @bundle_path = nil
       @bin_path = nil
       @user_home = nil
+    end
 
-      Plugin.reset!
-
+    def reset_rubygems!
       return unless defined?(@rubygems) && @rubygems
       rubygems.undo_replacements
       rubygems.reset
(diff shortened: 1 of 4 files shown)


segiddins:
This is particularly hard to write a spec for since it depends on default gems who's source is the the ruby stdlib dir... ideas are welcome

segiddins:
@indirect @colby-swandale any idea how to test this?

indirect:
@segiddins can we maybe look in the ruby stdlib dir for a json file and only run the test (against the json gem) if it's there?

colby-swandale:
Sorry 😞 I'm still learning about Bundler's internals and I'm not sure how i would go testing this myself.

segiddins:
It's OK, I managed to get a test working 😄

segiddins:
@indirect r?

indirect:
looks good! @bundlerbot r+

mvz:
Thanks, @segiddins!

PR #5505, opened 2017-03-07
Add `bundle issue` command that prints instructions for reporting issues

jonathanpike:
This is my stab at adding a `bundle issue` command, as listed in the list of tasks anyone can claim (#4871).  This is also my first contribution (of what I hope will be many) to Bundler!  

I hope that this is useful, and please let me know how to improve it!

### DIFF
--- a/lib/bundler/cli/issue.rb
+++ b/lib/bundler/cli/issue.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+require "rbconfig"
+
+module Bundler
+  class CLI::Issue
+    def run
+      Bundler.ui.info <<-EOS.gsub(/^ {8}/, "")
+        Did you find an issue with Bundler? Before filing a new issue,
+        be sure to check out these resources:
+
+        1. Check out our troubleshooting guide for quick fixes to common issues:
+        https://github.com/bundler/bundler/blob/master/doc/TROUBLESHOOTING.md
+
+        2. Instructions for common Bundler uses can be found on the documentation
+        site: http://bundler.io/
+
+        3. Information about each Bundler command can be found in the Bundler
+        man pages: http://bundler.io/man/bundle.1.html
+
+        Hopefully the trouble shooting steps above resolved your problem!  If things
+        still aren't working the way you expect them to, please let us know so
+        that we can diagnose and hopefully fix the problem you're having. Please
+        view the Filing Issues guide for more information:
+        https://github.com/bundler/bundler/blob/master/doc/contributing/ISSUES.md
+
+      EOS
+
+      Bundler.ui.info Bundler::Env.new.report
+
+      Bundler.ui.info "\n## Bundle Doctor"
+      doctor
+    end
+
+    def doctor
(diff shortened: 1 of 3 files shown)


colby-swandale:
Thanks! This looks promising but checkout [bundler/rfcs#5](https://github.com/bundler/rfcs/pull/5) for ideas we have for the `issue` command.

indirect:
This is awesome, thanks for sending it! Thanks for submitting this! I think this is a great first step 👍

@colby-swandale @segiddins I'd like to propose that we accept many small PRs that each handle one aspect of the overall issues RFC, rather than requiring a single giant PR that implements the entire thing. How do you feel about that?

colby-swandale:
@indirect makes sense, i can count about 4 different RFCS inside rfc#5

jonathanpike:
@colby-swandale thanks for pointing out rfc#5.  Prior to submitting this PR, I did look around for further discussion about this feature but I guess I missed this.  

From what the rfc lays out, it looks like the final version of bundle issue is more of an automated issue reporting solution, but that isn't fully discussed yet.  Since a lot of it looks still up for discussion, is there anything else that would be helpful for me to add to this PR to make it closer to what the final version should be?

colby-swandale:
Checkout the `bundle env` command, we probably will want to include the contents of the `Gemfile` and `Gemfile.lock`

jonathanpike:
@colby-swandale I'm already including `Env#report`, but I didn't provide the `print_gemfile` option.  I'll update that. Anything else?

colby-swandale:
@indirect @segiddins  Do we want to update `ISSUES.md` to inform users to use `bundler issue` instead from now on?

indirect:
@colby-swandale I think we'll want to wait to update the docs until this is merged and in a release, but it seems good to do it then 👍

jonathanpike:
@colby-swandale Let me know if there's anything else I can add/change to move this PR along.  I appreciate your help!

lib/bundler/cli/issue.rb
```diff
@@ -0,0 +1,40 @@
...
+        1. Check out our troubleshooting guide for quick fixes to common issues:
+        https://github.com/bundler/bundler/blob/master/doc/TROUBLESHOOTING.md
+
+        2. Instructions for common Bundler uses can be found on the documentation
+        site: http://bundler.io/
+
+        3. Information about each Bundler command can be found in the Bundler
+        man pages: http://bundler.io/man/bundle.1.html
+
+        Hopefully the trouble shooting steps above resolved your problem!  If things
```
colby-swandale: `troubleshooting`

lib/bundler/cli/issue.rb
```diff
@@ -0,0 +1,40 @@
...
+        still aren't working the way you expect them to, please let us know so
+        that we can diagnose and hopefully fix the problem you're having. Please
```
colby-swandale: i would suggest something like `that we can diagnose and help fix the problem you're having.`

colby-swandale:
Thank you!

@bundlerbot r+

Issue #4600, opened 2016-05-22
Add subcommands to `config`


indirect:
Today, it's confusing that `config` lets you pass the flags `local`, `global`, and `delete` in any combination.
It would be an improvement to the interface of `config` if Bundler 2 supported a slightly different set of commands, e.g.:

```
# Setting
$ bundle config set foo bar
# Unsetting
$ bundle config unset foo bar
# Local-only
$ bundle config set foo bar --local
# Global-only
$ bundle config set foo bar --global
```

denniss:
@indirect may I give this one a shot?

indirect:
yes, definitely! sounds awesome :)

indirect:
I think starting an implementation is a good way to figure out if there are issues that we need to discuss. :)

denniss:
@indirect I am using existing `config_spec` to test my implementation of this feature. While looking through the existing specs, I found this one to be somewhat confusing: https://github.com/bundler/bundler/blob/master/spec/commands/config_spec.rb#L217

Specifically, shouldn't the line:
```
run "puts Bundler.settings.send(:global_config_file).read"
```
be 

```
run "puts Bundler.settings.send(:local_config_file).read"
```

Since the previous command is calling `bundle config foo '1'`.

indirect:
@denniss yup, that looks like a mistake. No idea how it's passing. 😮

denniss:
I think that actually might be the expected behavior. Although, I find it kind of confusing. From `bundle help config`

>  Executing bundle config <name> <value> will set that configuration to the value specified for all bundles executed as the current user. The configuration will be stored in ~/.bundle/config. If name already is set, name will be overridden and user will be warned.

I'd like to propose a couple change to the behavior of the config command. 
- I feel like global configuration should only be changed when `--global` is specified. 
- Setting a value where there are spaces involve should require quotes. 
E.g. 
`bundle set config build.mysql --with-mysql-config=/usr/local/mysql/bin/mysql_config`
Should be
`bundle set config build.mysql '--with-mysql-config=/usr/local/mysql/bin/mysql_config'`

May be @segiddins can chime in on this?

segiddins:
Those are both breaking changes, which we wouldn't be able to ship in Bundler 1.x

indirect:
@segiddins yes, this entire PR is already in 2.0 territory. But I'm comfortable with starting work on that now, I think.

indirect:
@denniss I like both of those changes, I think 👍

[...]

PR #5507, opened 2017-03-07
Implementation of new config subcommands

denniss:
Implement:
https://github.com/bundler/bundler/issues/4600


spec/commands/config2_spec.rb
```diff
@@ -83,13 +83,155 @@
...
+      File.open(bundled_app(".bundle/config"), "w") do |f|
+        f.write 'BUNDLE_FOO: "$BUILD_DIR"'
+      end
+
+      bundle "config2 set bar baz"
+      run "puts Bundler.settings.send(:local_config_file).read"
+
+      # Starting in Ruby 2.1, YAML automatically adds double quotes
+      # around some values, including $ and newlines.
+      expect(out).to include('BUNDLE_FOO: "$BUILD_DIR"')
```
segiddins: we use our own YAML serializer, so this shouldn't be version-dependent?

colby-swandale:
@denniss can you fix the rubocop feedback reported in travis please 😄

denniss:
@indirect Can you take a peek at this? I started this in response to https://github.com/bundler/bundler/issues/4600 . I used the existing `config_spec.rb` as a starting point and to pass the implementation of the new command (currently called `config2` and the spec is in `config2_spec.rb`).

colby-swandale:
ping @indirect

spec/commands/config2_spec.rb
```diff
@@ -0,0 +1,286 @@
...
+
+      bundle "config2 --global set foo global"
+      expect(out).to match(/Your application has set foo to "local"/)
+
+      run "puts Bundler.settings[:foo]"
+      expect(out).to eq("local")
+    end
+
+    it "can be unset" do
+      bundle "config2 unset --global foo global"
```
indirect: looks like this was supposed to be a set?

denniss:
Changed the specs to not pass `--local` by default. Also fixed the mistake you found @indirect

spec/commands/config2_spec.rb
segiddins: ping on this comment?

denniss: Removed the ruby version specifier

denniss:
Currently new implementation is aliased as `config2`. I'd like to remove the existing `config` command and renamed `config2` to `config` if you guys think this looks okay.

indirect:
Since this is a breaking change, we can merge it to master behind a feature flag that switches the new implementation on. Then, after we ship 2.0 and the feature flag is always on, we can delete the old implementation. 👍

denniss:
@indirect @segiddins While working through merge conflight, I was wondering: If the same key exist for environment variable (`ENV['BUNDLE_BAR']) and global (~/.bundle/config) which has higher precedence? Currently environment variable has higher precedence. Is this correct? Does it matter?

segiddins:
> Currently environment variable has higher precedence. Is this correct? Does it matter?
yes, and yes

segiddins:
@denniss what's the status of this PR?

denniss:
I won't be able to touch this for another 2 weeks. Busy with life at the
moment. Mich apologies.

segiddins:
https://github.com/bundler/bundler/pull/5981 is an alternative to this PR

segiddins:
Superseded by #5981

PR #5508, opened 2017-03-08
Fix installation from gemfile shared across different directories

faucct:


spec/commands/install_spec.rb
```diff
@@ -506,4 +506,24 @@
...
+        BUNDLE_GEMFILE: "../Gemfile"
+      YAML
+      gemfile <<-RUBY
+        gem 'foo', path: 'foo'
+      RUBY
+
+      bundle "install"
+      in_app_root_custom "app" do
+        bundle "install"
+        expect(out).to include("Bundle complete!")
```
segiddins: please also add `expect(the_bundle).to include_gem "foo 1.0"`?

faucct: I believe I am having some troubles with that: it seems that the match you are suggesting to add does not care about `.bundle/config` and tries to validate the empty `app/Gemfile`.
The only reason I have created this empty Gemfile is to make `app` directory the bundler root, not the `app/..`.

faucct: `Bundler.root` should probably be changed instead of creating this empty `Gemfile`: it should choose the first directory containing `Gemfile`, `gems.rb` or `.bundle`.

faucct: I did an attempt to improve `Bundler.root` and added the check you've suggested.

lib/bundler/shared_helpers.rb
```diff
@@ -19,6 +19,16 @@ def requirement
 
 module Bundler
   module SharedHelpers
+    def default_root
+      given_gemfile = ENV["BUNDLE_GEMFILE"]
+      return Pathname.new(given_gemfile).dirname if given_gemfile && !given_gemfile.empty?
+      search_up('') do |path|
+        if path.join("Gemfile").file? || path.join("gems.rb").file? || path.join(".bundle").directory?
```
segiddins: this will return `~` if the user is not in a bundle, which I don't think we want?

faucct: Okay, so I should put the check from `SharedExamples#default_bundle_dir` here, right?

segiddins: that should work, yeah!

lib/bundler/shared_helpers.rb
```diff
@@ -19,6 +19,16 @@ def requirement
 
 module Bundler
   module SharedHelpers
+    def default_root
+      given_gemfile = ENV["BUNDLE_GEMFILE"]
+      return Pathname.new(given_gemfile).dirname if given_gemfile && !given_gemfile.empty?
+      search_up("") do |path|
```
segiddins: this still can return `~`

segiddins:
@faucct ping on my last comment about `~`?

colby-swandale:
ping @faucct

faucct:
Hello, had no time to finish the thing yet. Will be done in couple weeks.

colby-swandale:
ping @faucct is this still being worked on?

lib/bundler/shared_helpers.rb
faucct: I have rebased the thing and it looks like my changes here weren't necessary anymore, so I have removed them.

faucct:
Finally rebased. Please, review this once again.

faucct:
ping

deivid-rodriguez:
Any news @faucct? I copied the test you added to both 2-0-stable and current master (in both 2.x and 3.x mode), and it passed in all cases, so we really nee a description of what this is fixing. Thanks!

faucct:
I guess this can be closed if the tests are passing.

deivid-rodriguez:
Ok, if you run into related trouble, make sure to let us know!

PR #5509, opened 2017-03-08
use yellow color when version appears to be moving backwards

### LINKED ISSUE #5506: More semantic colors in bundle update
One thing I have noticed in recent versions of bundler is helpful colors in the command line. Overall I think this is great. However I noticed that the comparison of old gems to new ones is green _even when the gem version is regressing_. Here’s a screenshot:

screen shot 2017-03-07 at 11 18 46 am

Notice that all of these gems have gone backwards!

I would suggest a color like yellow/amber for all decrementing gems, as this will (I think appropriately) direct the user’s attention to these items and make sure that there are not any unintentional regressions.

Happy to submit a PR if there is interest in this.

brchristian:
This is very much intended as a work in progress (no tests!) and proof-of-concept .
Closes #5506.

In cases where the _dependencies_ change, it is not always possible to tell if they are going forwards or backwards -- for instance a change from `> 2.0` to `< 2.8` is neither obviously a forward or a backward change.

Hence the behavior here is to always default to the green color unless there are two specific version numbers and the new one is strictly less than the old one, in which case we use the yellow color.

I’d love any code review and any suggestions for the right place to add some tests for this. I think the tests should end up being quite simple -- we can check for `\e[32m` (green) versus `e[33m` (yellow) in the strings sent to the shell, and that’s easy enough -- but I’m simply not familiar enough with the test suite to know the appropriate place and syntax to do that. Suggestions welcome!

### DIFF
--- a/lib/bundler/source.rb
+++ b/lib/bundler/source.rb
@@ -42,13 +42,17 @@ module Bundler
   private
 
     def version_color(spec_version, locked_spec_version)
-      earlier_version?(spec_version, locked_spec_version) ? :yellow : :green
+      if Gem::Version.correct?(spec_version) && Gem::Version.correct?(locked_spec_version)
+        # display yellow if there appears to be a regression
+        earlier_version?(spec_version, locked_spec_version) ? :yellow : :green
+      else
+        # default to green if the versions cannot be directly compared
+        :green
+      end
     end
 
     def earlier_version?(spec_version, locked_spec_version)
       Gem::Version.new(spec_version) < Gem::Version.new(locked_spec_version)
-    rescue ArgumentError
-      false
     end
   end
 end
--- a/lib/bundler/source.rb
+++ b/lib/bundler/source.rb
@@ -20,19 +20,11 @@ module Bundler
         locked_spec = Bundler.locked_gems.specs.find {|s| s.name == spec.name }
         locked_spec_version = locked_spec.version if locked_spec
         if locked_spec_version && spec.version != locked_spec_version
-          message += Bundler.ui.add_color(" (was #{locked_spec_version})", earlier_version?(spec.version, locked_spec_version) ? :yellow : :green)
+          message += Bundler.ui.add_color(" (was #{locked_spec_version})", version_color(spec.version, locked_spec_version))
         end
       end
 
       message
-    end
-
-    def earlier_version?(spec_version, locked_spec_version)
-      begin
-        Gem::Version.new(spec_version) < Gem::Version.new(locked_spec_version)
-      rescue ArgumentError
-        false
-      end
     end
 
     def can_lock?(spec)
@@ -45,6 +37,18 @@ module Bundler
 
     def inspect
       "#<#{self.class}:0x#{object_id} #{self}>"
+    end
+
+  private
+
+    def version_color(spec_version, locked_spec_version)
+      earlier_version?(spec_version, locked_spec_version) ? :yellow : :green
+    end
+
+    def earlier_version?(spec_version, locked_spec_version)
+      Gem::Version.new(spec_version) < Gem::Version.new(locked_spec_version)
+    rescue ArgumentError
+      false
     end
   end
 end
--- a/lib/bundler/source.rb
+++ b/lib/bundler/source.rb
@@ -30,7 +30,7 @@ module Bundler
     def earlier_version?(spec_version, locked_spec_version)
       begin
         Gem::Version.new(spec_version) < Gem::Version.new(locked_spec_version)
-      rescue
+      rescue ArgumentError
         false
       end
     end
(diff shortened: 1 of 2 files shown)


lib/bundler/source.rb
```diff
@@ -20,13 +20,21 @@ def version_message(spec)
...
         end
       end
 
       message
     end
 
+    def earlier_version?(spec_version, locked_spec_version)
+      begin
+        Gem::Version.new(spec_version) < Gem::Version.new(locked_spec_version)
+      rescue
```
colby-swandale: Preferred if you don't catch all exceptions

lib/bundler/source.rb
```diff
@@ -20,13 +20,21 @@ def version_message(spec)
         locked_spec = Bundler.locked_gems.specs.find {|s| s.name == spec.name }
         locked_spec_version = locked_spec.version if locked_spec
         if locked_spec_version && spec.version != locked_spec_version
-          message += Bundler.ui.add_color(" (was #{locked_spec_version})", :green)
+          message += Bundler.ui.add_color(" (was #{locked_spec_version})", earlier_version?(spec.version, locked_spec_version) ? :yellow : :green)
```
colby-swandale:
```
earlier_version?(spec.version, locked_spec_version) ? :yellow : :green
```

We can probably extract this logic into a private method that returns the `:color`

brchristian:
@colby-swandale Thanks for the review. I’ve extracted the color logic into a private method as you suggested, and have restricted the `rescue` to only catch `ArgumentError`, which is the relevant error in this case.

lib/bundler/source.rb
```diff
@@ -38,5 +38,17 @@ def include?(other)
...
+
+  private
+
+    def version_color(spec_version, locked_spec_version)
+      earlier_version?(spec_version, locked_spec_version) ? :yellow : :green
+    end
+
+    def earlier_version?(spec_version, locked_spec_version)
+      Gem::Version.new(spec_version) < Gem::Version.new(locked_spec_version)
+    rescue ArgumentError
```
segiddins: why are we rescuing here?

brchristian: @segiddins We’re turning the `version` strings in the parameters into `Gem::Version` so that we can use the `<` operator (and not have to write multi-decimal-point comparison logic from scratch).

However, some of the time when I was testing it, the strings would be dependencies rather than versions -- e.g., `>= 2.0.0` rather than `2.1.0`. In those cases, `Gem::Version.new` will throw an exception, and we simply catch that and default to not adjusting the default color.

If there’s a cleaner implementation, I’m certainly open to that. Hope this makes the existing implementation make a bit more sense in the meantime.

segiddins: Why would these be dependencies as opposed to versions? Since the parameter name is `version` -- if it's not possible to get the actual version, I'd prefer we check `Gem::Version.correct?` instead of using exceptions for control flow

colby-swandale: pring @brchristian

brchristian: @segiddins I can confirm that despite being named `spec.version`, the `version_message(spec)` method does get passed things that are not proper versions. For instance, in running `spec/bundler/source_spec.rb`, it frequently gets passed in an object for which `spec.version` is `">= 1.6"`.

I didn’t know about `Gem::Version.correct?` but that is much better than catching an exception, I totally agree.

Added a commit here that made that change.

This leaves two issues remaining:

1. Ideally we only want to even call `earlier_version?` when comparing actual concrete version numbers. I’m still learning the Bundler codebase and don’t know offhand how I would do this.

2. We need some specs to positively assert cases where we want/expect the amber color to be present. Again, I’m not very familiar with the Bundler code so I would love any help here if possible.

colby-swandale:
@segiddins does this look good to you?

segiddins:
Still needs a test, otherwise looks 💯

brchristian:
@segiddins I’m not sure yet how to set up the right test . . . would it go in [`bundler/source_spec.rb`](https://github.com/bundler/bundler/blob/master/spec/bundler/source_spec.rb)?

At a glance, I couldn’t find tests pertaining to the "Using ____ (was ____)" statements, which is where we should be adding tests. Once we find the right place to add them, the actual tests themselves should be quite simple. We’ll just check the output string for yellow or green as appropriate.

segiddins:
Yup, see line 64 of that file for an example

brchristian:
👍 Tests added for two conditions: an apparent regression (yellow) and an apparent forward change (green). All tests pass and things are looking good.

screen shot 2017-03-28 at 5 20 37 pm

screen shot 2017-03-28 at 5 21 27 pm


Woohoo!

One lingering question for me is that I’m not completely certain that the yellow coloring will work in cases where the _entire_ line shows in green. For instance, something like this:

screen shot 2017-03-28 at 5 19 14 pm

Not sure yet that cases like these will work. They may, actually, but I wasn’t sure how to test for that particular case... But maybe that’s a corner case anyway.

segiddins:
Looks great!
@bundlerbot r+

PR #5520, opened 2017-03-17
Only uninstall plugins that we attempted to install

kerrizor:
Previously, when encountering any kind of exception when installing or
registering a plugin, we would uninstall _everything_ in `specs`, which included
not just every installed plugin but Bundler itself 😱  

This patch prevents this from happening, by only attempting to delete code 
that is a) part of the installation attempt and b) NOT a currently registered
command. This second restriction is to avoid a broken state,  in the case
where we attempt to install the same plugin twice. In that case, if we uninstall
the code for the plugin, but leave the command registered, this causes
attempts to reinstall the plugin to fail, as the command is still in the
registry, and attempts to USE the command will fail because the code behind
it has been destroyed.

Paired with @roseaboveit

### DIFF
--- a/lib/bundler/plugin.rb
+++ b/lib/bundler/plugin.rb
@@ -38,7 +38,7 @@ module Bundler
       save_plugins names, specs
     rescue PluginError => e
       if specs
-        specs_to_delete = specs.select {|k, _v| names.include?(k) && !index.commands.values.include?(k) }
+        specs_to_delete = Hash[specs.select {|k, _v| names.include?(k) && !index.commands.values.include?(k) }]
         specs_to_delete.values.each {|spec| Bundler.rm_rf(spec.full_gem_path) }
       end
 
--- a/lib/bundler/plugin.rb
+++ b/lib/bundler/plugin.rb
@@ -37,7 +37,11 @@ module Bundler
 
       save_plugins names, specs
     rescue PluginError => e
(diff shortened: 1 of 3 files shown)


segiddins:
Looks good with a test!

kerrizor:
@segiddins basic test for a happy path added :)

lib/bundler/plugin.rb
```diff
@@ -37,7 +37,11 @@ def install(names, options)
 
       save_plugins names, specs
     rescue PluginError => e
-      specs.values.map {|spec| Bundler.rm_rf(spec.full_gem_path) } if specs
+      if specs
+        specs_to_delete = specs.select {|k, _v| names.include?(k) && !index.commands.values.include?(k) }
+        specs_to_delete.values.map {|spec| Bundler.rm_rf(spec.full_gem_path) }
```
segiddins: this should be `.each` since the return value isn't used

kerrizor: Nice catch!

kerrizor:
@segiddins I think we're good here. I'm a little shaky on 4022510 but it seems to be working, and tests are green.

segiddins:
Awesome!
@bundlerbot r+

segiddins:
@bundlerbot retry

kerrizor:
Looks like a timeout error in Travis?

colby-swandale:
@bundlerbot retry

colby-swandale:
@bundlerbot retry

PR #5524, opened 2017-03-19
Fix `bundle gem` command without having git

colby-swandale:
So the `gem` command is still breaking on master if `git` is not installed on the machine. This is because we're not checking if git is installed before we init a new git repo.

The test for #5466 is also not actually testing if git is not installed, it's only removing the git config which we already have tests for as well.

@segiddins Do you know a way i can stub `Bundler.git_present?` in the sub-process or have `git` appear to not be present in rspec?

### DIFF
--- a/spec/commands/newgem_spec.rb
+++ b/spec/commands/newgem_spec.rb
@@ -169,9 +169,13 @@ RSpec.describe "bundle gem" do
   context "when git is not avaiable" do
     let(:gem_name) { "test_gem" }
 
+    # This spec cannot have `git` avaiable in the test env
     before do
-      allow(Bundler).to receive(:git_present?).and_return(false)
-      bundle "gem #{gem_name}"
+      bundle_bin = File.expand_path("../../../exe/bundle", __FILE__)
+      load_paths = [lib, spec]
+      load_path_str = "-I#{load_paths.join(File::PATH_SEPARATOR)}"
+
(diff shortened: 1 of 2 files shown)


spec/commands/newgem_spec.rb
```diff
@@ -158,22 +158,32 @@ def gem_skeleton_assertions(gem_name)
...
-      it "contribute URL set to [USERNAME]" do
-        expect(bundled_app("test_gem/README.md").read).to include("[USERNAME]")
-        expect(bundled_app("test_gem/README.md").read).not_to include("github.com/bundleuser")
-      end
+  context "when git is not avaiable" do
+    let(:gem_name) { "test_gem" }
 
-      it_should_behave_like "git config is absent"
+    before do
+      allow(Bundler).to receive(:git_present?).and_return(false)
```
segiddins: maybe modify `$PATH` to remove whatever contains `git`?

colby-swandale: I tried looking into this quickly, but the biggest problem was that the `ruby` binary was in the same part of `$PATH` along with `git`.

segiddins:
Awesome!
@bundlerbot r+

colby-swandale:
@bundlerbot retry

colby-swandale:
@bundlerbot retry

colby-swandale:
@bundlerbot r+

PR #5527, opened 2017-03-21
try a version that uses the local cache on install

phoet:
i'm using `bundler/inline` quite heavily to configure dependencies for scripts in a rails project.

the inlining is great as i don't have pollute the apps dependency namespace for the scripts.

there are still a lot of transitive dependencies that could just be used from what is already installed.

when setting the `install` param to `true`, inline checks the rubygems repo every time. that is not really fast, especially from the eu or without internet. it also installs more things than necessary as it will use the best/latest compatible dependency. 

passing the `local` flag as in `bundle install --local` would save me from a lot of those problems.

i did not actually try this change (i would probably need to manually install my bundler fork as a gem?) i hope CI will complain about any issues 😸

### DIFF
--- a/lib/bundler/inline.rb
+++ b/lib/bundler/inline.rb
@@ -52,7 +52,7 @@ def gemfile(install = false, options = {}, &gemfile)
   definition = builder.to_definition(nil, true)
   def definition.lock(*); end
   definition.validate_runtime!
-  
+
   missing_specs = proc do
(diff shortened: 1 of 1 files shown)


segiddins:
Looks good with test coverage!

phoet:
i think this actually does not do what i want...

PR #5531, opened 2017-03-23
remove Rails 2.3 from ISSUES.md guide

colby-swandale:
Rails 2.3 has been EOL since 2013 and i have not seen any issue currently in Bundler's open issues or recently closed that is dealing with it. So i propose we remove it as it no longer has any value.


segiddins:
👍🏻

colby-swandale:
@bundlerbot r+

PR #5539, opened 2017-03-28
Change LockFileParse dependencies class variable from an array to a hash

jules2689:
What this addresses
---
The LockFileParser dependenies class variable is used throughout definition
using a series of select and find statements. While measuring Bundler performance, it was
noticed that pieces that interacted with @locked_deps took a fair chunk of time.

This chart shows some timings for `Definition.new` which is called during `require 'bundler/setup'`

image

** Note: The time scale is a little off because `TracePoint` slows it down, but this does show us a slow point we can optimize.

Diving into that a bit more, we can see the following timings

image

Again, despite the timings being off due to TracePoint, we can see that `locked_source = @locked_deps.select {|d| d.name == dep.name }.last (run 112812 times) :a1, 0.001, 0.182` was run upwards of 113K times for the Shopify application. Each one of these runs would run in `O(n)` time since we are literally iterating the array.

Looking at `LockFileParser`, we can see that the `dependencies` can easily be changed to a hash.

How this addresses it
---
This change makes the dependencies class variable a hash, which reduces the lookup
from `O(n)` to `O(1)` for each usage. Overall, the operation was `O(n^2)` as we iterated the array for each entry in the array. Now it's `O(n)` :tada:

In the Shopify application, this equated to a 20ms performance gain on each bundler/setup.

Some numbers
---
In particular, the `converge_dependencies` method in `definition.rb` took, on average, 24 ms for an application the size of Shopify. By simply changing the `dependencies` class variable, this was knocked down to 12ms. There was also a savings of a few ms on other method calls, such as `converge_locked_specs`.

Overall, I saw the time to `require 'bundler/setup'` from from 744ms to 724ms for a savings of 20ms.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -81,7 +81,7 @@ module Bundler
           @locked_sources = @locked_gems.sources
         else
           @unlock         = {}
-          @locked_deps    = []
+          @locked_deps    = {}
           @locked_specs   = SpecSet.new([])
           @locked_sources = []
         end
@@ -89,7 +89,7 @@ module Bundler
         @unlock         = {}
         @platforms      = []
         @locked_gems    = nil
-        @locked_deps    = []
+        @locked_deps    = {}
         @locked_specs   = SpecSet.new([])
         @locked_sources = []
         @locked_platforms = []
@@ -132,7 +132,7 @@ module Bundler
       # with a mismatch on #type.
       # Test coverage to catch a regression on this is in gemspec_spec.rb
       @dependencies.each do |d|
-        if ld = @locked_deps.find {|l| l.name == d.name }
+        if ld = @locked_deps[d.name]
           ld.instance_variable_set(:@type, d.type)
         end
       end
@@ -429,8 +429,8 @@ module Bundler
       new_sources = gemfile_sources - @locked_sources
       deleted_sources = @locked_sources - gemfile_sources
 
-      new_deps = @dependencies - @locked_deps
-      deleted_deps = @locked_deps - @dependencies
+      new_deps = @dependencies - @locked_deps.values
+      deleted_deps = @locked_deps.values - @dependencies
 
       # Check if it is possible that the source is only changed thing
       if (new_deps.empty? && deleted_deps.empty?) && (!new_sources.empty? && !deleted_sources.empty?)
@@ -455,7 +455,7 @@ module Bundler
 
       both_sources = Hash.new {|h, k| h[k] = [] }
       @dependencies.each {|d| both_sources[d.name][0] = d }
-      @locked_deps.each  {|d| both_sources[d.name][1] = d.source }
+      @locked_deps.each  {|name, d| both_sources[name][1] = d.source }
 
       both_sources.each do |name, (dep, lock_source)|
         next unless (dep.nil? && !lock_source.nil?) || (!dep.nil? && !lock_source.nil? && !lock_source.can_lock?(dep))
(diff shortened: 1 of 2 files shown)


lib/bundler/definition.rb
```diff
@@ -709,12 +709,8 @@ def converge_locked_specs
       # and Gemfile.lock. If the Gemfile modified a dependency, but
       # the gem in the Gemfile.lock still satisfies it, this is fine
       # too.
-      locked_deps_hash = @locked_deps.inject({}) do |hsh, dep|
-        hsh[dep] = dep
-        hsh
-      end
       @dependencies.each do |dep|
-        locked_dep = locked_deps_hash[dep]
```
segiddins: this might still need to be keyed off the dependency instead of just the name, it might be responsible for the test failures

jules2689: Actually, it seems the test failure is related to the method `dependencies_for_source_changed?`.

`locked_deps_for_source = @locked_deps.select {|_, dep| dep.source == locked_source }` fails, but `locked_deps_for_source = @locked_deps.values.select {|dep| dep.source == locked_source }` does not. I failed to take into account that we need an array of sources after the fact to turn into a set.

jules2689: Just pushed a fix for that :)

segiddins:
(Off topic, but what are you using to profile? That graph looks very nice :) )

jules2689:
@segiddins it's a little snippet of code that I use to generate text for [mermaid](http://knsv.github.io/mermaid/). You can find that code [here](http://notes.jnadeau.ca/computers/ruby/auto_generated_gantt_charts/), which can be generated into charts using [their live editor](http://knsv.github.io/mermaid/live_editor/). It also supports flowcharts, sequence diagrams, and a few others.

This is another example I used to keep track of slow methods (you'll see some PRs for the others hopefully soon!)

image

I also use these snippets for quick and dirty timing analysis.

To time individual lines on a per-call basis:
```ruby
    def _t(label)
      t = Process.clock_gettime(Process::CLOCK_MONOTONIC)
      ret = yield
      puts "#{label} #{Process.clock_gettime(Process::CLOCK_MONOTONIC) - t}"
      ret
    end
```

`_t('my label') { puts "hello world" }` for example:
```shell
irb(main):007:0> _t('my label') { puts "hello world" }
hello world
my label 1.999997766688466e-05
```

or for all calls aggregated, I use this:
```ruby
$agg = {}
at_exit do
  $agg.each do |k, v|
    puts "#{k} #{v}"
  end
end

# Aggregate the time it takes to run a block of code.
# Returns the value of yield.
# At exit, it will output the aggregated time.
def _ta(label)
  $agg[label] ||= 0
  t = Process.clock_gettime(Process::CLOCK_MONOTONIC)
  ret = yield
ensure
  $agg[label] += (Process.clock_gettime(Process::CLOCK_MONOTONIC) - t)
  ret
end
```

```shell
irb(main):026:0> 1000.times { _ta('my label') { 100 * 100 } }
=> 1000
irb(main):027:0> exit
my label 0.00041399989277124405
```

lib/bundler/definition.rb
jules2689: Oh, I see what you mean. There were more failures on CI than I had locally :(

jules2689: Actual fix is incoming.

jules2689: @segiddins I ended up opting for the next line `locked_dep = nil unless locked_dep == dep` to compensate for the inequality. This allows us to use the `O(1)` lookup of the `@locked_deps` hash without creating an entirely new hash.

segiddins: If it works (aka the tests all pass), then it seems good to me!

segiddins:
Amazing.
@bundlerbot r+

PR #5542, opened 2017-03-28
[Resolver] Only include minimal set of conflicting deps in conflict message

segiddins:
This implements a feature that @indirect thought up last week.
Previously, when a resolver conflict occurred, bundler would print _all_ the "conflicting" dependencies, even if some of them were not an issue! No longer. Now, we'll compute the minimal set of conflicting dependencies and only print out the requirement trees for those, which should hopefully make conflict messages less confusing and more actionable for users.

I have some tests for the `VersionRanges` module that I'll push up in a few minutes, but a test for the feature itself is already included.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -55,6 +55,7 @@ module Bundler
   autoload :StubSpecification,      "bundler/stub_specification"
   autoload :UI,                     "bundler/ui"
   autoload :URICredentialsFilter,   "bundler/uri_credentials_filter"
+  autoload :VersionRanges,          "bundler/version_ranges"
 
   class << self
     attr_writer :bundle_path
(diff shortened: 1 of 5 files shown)


segiddins:
Spec added

segiddins:
Should pass now, @indirect

indirect:
I have no idea how you implemented this so quickly, but it looks great. 😄👍

segiddins:
The flight home from Hong Kong was _long_

segiddins:
Should now pass on 1.8.7

segiddins:
@indirect @colby-swandale r?

colby-swandale:
wow this is awesome! 👍 from me

indirect:
@bundlerbot r+

segiddins:
@bundlerbot retry

PR #5545, opened 2017-03-29
Document that --retry works for update.

coderanger:
Also makes the --jobs doc a little closer to the copy from install.

### DIFF
--- a/man/bundle-update.ronn
+++ b/man/bundle-update.ronn
@@ -55,7 +55,7 @@ gem.
   Specify the number of jobs to run in parallel. The default is `1`.
 
 * `--retry=[<number>]`:
-Retry failed network or git requests for <number> times.
+  Retry failed network or git requests for <number> times.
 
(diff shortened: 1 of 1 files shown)


man/bundle-update.ronn
```diff
@@ -51,8 +51,11 @@ gem.
 * `--full-index`:
   Fall back to using the single-file index of all gems.
 
-* `--jobs`:
-  Specify the number of jobs to run in parallel.
+* `--jobs=[<number>]`:
+  Specify the number of jobs to run in parallel. The default is `1`.
+
+* `--retry=[<number>]`:
+Retry failed network or git requests for <number> times.
```
segiddins: alignment doesn't match the other options

segiddins:
Wonderful, thanks so much!

@bundlerbot r+

segiddins:
@bundlerbot retry

PR #5546, opened 2017-03-29
Change `definition#converge_dependencies` comparison to an iterator for performance

jules2689:
What this addresses
---
Follow up for #5539 

Previously, a proc was used to extract name and requirements from dependencies. This was
done twice, `2*O(n)`, and compared as `Set` objects.

This was identified as slow:

image

** Again, note that tracepoint slows the results in the chart down.

Without `TracePoint`, we run at about 13-15ms.

How this fixes it
---
Now that we use a hash for `@locked_deps`, we can
iterate one of the arrays (`@dependencies`) and compare against entries in `@locked_deps` with `O(1)`
access. This results in `O(n)` access.

The result is that `converge_dependencies`, which used to take 15ms, now takes about 2-3ms.

image

** Again, note that tracepoint slows the results in the chart down.

Considerations
---
Equality between 2 dependencies, as defined [here for =~](https://docs.ruby-lang.org/en/2.4.0/Gem/Dependency.html), shows us that equality means:

> This dependency will match if the name matches the other's name, and other has only an equal version requirement that satisfies this dependency.

To make sure we do not regress, we need to check the name and the requirements. Looking at the source code, we can actually just compare the deps as that doesnt take into account `type`.

Running the added test from #5354, it passed. So these assumptions seem to hold.

cc @segiddins as you wrote #5354 and reviewed #5539

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -695,8 +695,17 @@ module Bundler
           dep.platforms.concat(@platforms.map {|p| Dependency::REVERSE_PLATFORM_MAP[p] }.flatten(1)).uniq!
         end
       end
-      dependency_without_type = proc {|d| Gem::Dependency.new(d.name, *d.requirement.as_list) }
-      Set.new(@dependencies.map(&dependency_without_type)) != Set.new(@locked_deps.values.map(&dependency_without_type))
+
+      # We want to know if all match, but don't want to check all entries
+      # This means we need to return false if any dependency doesn't match
+      # the lock or doesn't exist in the lock.
+      @dependencies.any? do |dependency|
+        locked_dep = @locked_deps[dependency.name]
+        next true if locked_dep.nil?
+        # We already know the name matches from the hash lookup
+        # so we only need to check the requirement now
+        dependency.requirement != locked_dep.requirement
+      end
     end
 
     # Remove elements from the locked specs that are expired. This will most


segiddins:
👍🏻

hmistry:
@jules2689 This is great! Side question: how are you generating this performance charts?

jules2689:
@hmistry heh, segiddins had the same question. [This comment](https://github.com/bundler/bundler/pull/5539#issuecomment-289875292) explains my methodology.

lib/bundler/definition.rb
```diff
@@ -695,8 +695,14 @@ def converge_dependencies
...
       end
-      dependency_without_type = proc {|d| Gem::Dependency.new(d.name, *d.requirement.as_list) }
-      Set.new(@dependencies.map(&dependency_without_type)) != Set.new(@locked_deps.values.map(&dependency_without_type))
+
+      # We want to know if all match, but don't want to check all entries
+      # This means we need to bail if any dependency doesn't match the lock.
+      @dependencies.any? do |dependency|
+        locked_dep = @locked_deps[dependency.name]
+        next true if locked_dep.nil?
+        !(dependency =~ locked_dep)
```
segiddins: maybe do `dependency.requirement != locked_dep.requirement`?

jules2689: Thank you. That certainly helped.

lib/bundler/definition.rb
```diff
@@ -695,8 +695,17 @@ def converge_dependencies
...
       end
-      dependency_without_type = proc {|d| Gem::Dependency.new(d.name, *d.requirement.as_list) }
-      Set.new(@dependencies.map(&dependency_without_type)) != Set.new(@locked_deps.values.map(&dependency_without_type))
+
+      # We want to know if all match, but don't want to check all entries
+      # This means we need to return false if any dependency doesn't match
+      # the lock or doesn't exist in the lock.
+      @dependencies.any? do |dependency|
+        locked_dep = @locked_deps[dependency.name]
+        next false if locked_dep.nil?
```
segiddins: should this be `next true`?

jules2689: 🤦‍♂️ yes, you are right. I keep confusing myself for some reason.

segiddins:
Awesome, thanks so much @jules2689!
If you'd like, you're welcome to join the [bundler slack](http://slack.bundler.io/)!
@bundlerbot r+

segiddins:
@bundlerbot retry

PR #5554, opened 2017-03-31
[Source::Git] Only validate specs once (on download)

segiddins:
Instead of validating every time we setup, only do so when we re-serialize the spec

\c @jules2689

### DIFF
--- a/lib/bundler/source/git.rb
+++ b/lib/bundler/source/git.rb
@@ -234,6 +234,7 @@ module Bundler
           # The gemspecs we cache should already be evaluated.
           spec = Bundler.load_gemspec(spec_path)
           next unless spec
+          Bundler.rubygems.validate(spec)
           File.open(spec_path, "wb") {|file| file.write(spec.to_ruby) }
         end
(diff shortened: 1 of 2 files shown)


lib/bundler/source/git.rb
```diff
@@ -234,6 +234,7 @@ def serialize_gemspecs_in(destination)
           # The gemspecs we cache should already be evaluated.
           spec = Bundler.load_gemspec(spec_path)
           next unless spec
+          Bundler.rubygems.validate_spec(spec)
```
jules2689: `validate_spec` -> `validate`

jules2689:
This is great. It reduces the `bundler/setup` time for the Shopify application by 20+%. :tada:

lib/bundler/source/git.rb
segiddins: :head: :desk:

segiddins:
@indirect r?

indirect:
@bundlerbot r+

PR #5556, opened 2017-03-31
Sam tries to make bundler fast part 4372894632785

segiddins:


lib/bundler/lockfile_parser.rb
```diff
@@ -171,43 +171,45 @@ def parse_source(line)
       end
     end
 
-    NAME_VERSION = '(?! )(.*?)(?: \(([^-]*)(?:-(.*))?\))?'.freeze
-    NAME_VERSION_2 = /^ {2}#{NAME_VERSION}(!)?$/
-    NAME_VERSION_4 = /^ {4}#{NAME_VERSION}$/
-    NAME_VERSION_6 = /^ {6}#{NAME_VERSION}$/
+    NAME_VERSION = /^( {2}| {4}| {6})(?! )(.*?)(?: \(([^-]*)(?:-(.*))?\))?(!)?$/
```
jules2689: Could I recommend this syntax:
```ruby
%r{
  ^( {2}| {4}| {6})              # Match 2,4, or 6 spaces
    (?! )                        # Match ...
    (.*?)
    (?: \(([^-]*)
    (?:-(.*))?\))?(!)?
$}x
```

The `x` at the end of that lets you add comments and split up the pieces in newlines, indent, etc. I've otherwise found regexes difficult to reason when coming from an out of context position.

Similarly, I tend to leave a comment above the regex indicating what it _should_ match. This lets someone easily match between the regex, the comments, and the example - and easily see if a change is needed.

jules2689: Also, for below, would it make sense to use named capture groups? That way we don't have to just match at `$1`, it could be `match['name']`, `match['pinned']`

lib/bundler/definition.rb
```diff
@@ -464,7 +447,8 @@ def ensure_equivalent_gemfile_and_lockfile(explicit_flag = false)
         changed << "* #{name} from `#{gemfile_source_name}` to `#{lockfile_source_name}`"
       end
 
-      msg << "\n\n#{change_reason.split(", ").join("\n")}\n"
+      reason = change_reason
+      msg << "\n\n#{reason.split(", ").join("\n")}\n" unless reason.strip.empty?
```
jules2689: Out of curiosity, was there a reason to split this line up?

lib/bundler/definition.rb
```diff
@@ -696,16 +684,24 @@ def converge_dependencies
...
-      @dependencies.any? do |dependency|
-        locked_dep = @locked_deps[dependency.name]
-        next true if locked_dep.nil?
+      @dependencies.each do |dependency|
+        unless locked_dep = @locked_deps[dependency.name]
+          changes = true
+          next
+        end
+
+        locked_dep.instance_variable_set(:@type, dependency.type)
```
jules2689: Would it make sense to transfer part of the comment from `fixup_dependency_types` so there is context as to why we call `instance_variable_set`?

lib/bundler/definition.rb
segiddins: yeah, it fixes output

lib/bundler/definition.rb
segiddins: the specs will break without it ¯\_(ツ)_/¯ but sure, I could add a comment

jules2689:
Just for clarity, my comments were more suggestions to make things a little easier to follow for future readers. The code overall looks pretty straight forward. Testing now

jules2689:
🎩  nothing broke when I ran this for the shopify application!

segiddins:
And did it makes bundler faster?

lib/bundler/lockfile_parser.rb
segiddins: Is there a performance difference between named and unnamed capture groups?

jules2689:
Before
---
| measure | result |
| --- | --- |
| std. dev | 0.04323717102 |
| min	| 0.684562 |
| max	| 0.856426 |
| median	| 0.756842 |
| average |	0.7595205714 |

After
---
| measure | result |
| --- | --- |
| std. dev | 0.05084804717 |
| min	| 0.648624 |
| max	| 0.829057 |
| median	| 0.71773 |
| average | 0.7183865714 |

image

What I take from this is that the time it takes for this branch varies a little more before, but on average and most of the time it takes about 40-50ms less :tada:

segiddins:
> but on average and most of the time it takes about 40-50ms less

I'll take it!
I'll address the comments soon.

segiddins:
Done!
@indirect r?

indirect:
this looks good! @bundlerbot r+

Issue #5530, opened 2017-03-22
bundle config local: doesn't use local repo


perlun:
I am trying to use the `bundle config local` approach to be able to work with a local copy of a gem I am authoring.

Here is the command I've used:

```shell
$ bundle config local.ecraft-web_api ~/git/ecraft/ecraft-web_api/
```

However, `bundle install` doesn't list this path in its output, and `bundle update` doesn't seem to take it in as well.

```
Using ecraft-web_api 2.0.3
```

I added an exception to the first line of the gem, which is not raised, so I'm quite sure it doesn't use my local version of the gem.

My Gemfile (except):

```ruby
source 'https://private-gem-url/' do
  gem 'ecraft-web_api', branch: 'feature/add-authorization-support'         # Version specified in gemspec
end
```

Here is my Bundler and Ruby version:

```shell
$ bundle -v
Bundler version 1.14.6
$ ruby -v
ruby 2.3.3p222 (2016-11-21 revision 56859) [x86_64-darwin16]
```

segiddins:
You need to use it as a git gem rather than from a RubyGems source, but we definitely should be printing an error for using `branch` with a RubyGems source

perlun:
> You need to use it as a git gem rather than from a RubyGems source

Ah, OK! The use case I have is it is developed as a git gem, but every now and then being published as a Rubygem. But when making changes to the gem, it's nice to be able to tweak the gem locally and test it in its dependant.

Since the GitHub repo in question is private, it makes it harder for CI etc if I would specify the gem source to be the GitHub repo, the CI user will not have access to that repo.

segiddins:
Closed by https://github.com/bundler/bundler/issues/5559

perlun:
Thanks @segiddins and @colby-swandale!

PR #5559, opened 2017-04-03
print an error message when a non-git gem is given a `branch` option

colby-swandale:
When the user supplies the `branch` option to a non-git gem an error message should be printed.

Relates to #5530.

### DIFF
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -385,7 +385,8 @@ module Bundler
     def validate_keys(command, opts, valid_keys)
       invalid_keys = opts.keys - valid_keys
 
-      if opts["branch"] && !(opts["git"] || opts["github"])
+      git_source = opts.keys & @git_sources.keys.map(&:to_s)
+      if opts["branch"] && !(opts["git"] || opts["github"] || git_source.any?)
         raise GemfileError, %(The `branch` option for `#{command}` is not allowed. Only gems with a git source can specify a branch)
       end
 
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -321,10 +321,6 @@ module Bundler
 
       normalize_hash(opts)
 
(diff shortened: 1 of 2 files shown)


lib/bundler/dsl.rb
```diff
@@ -321,6 +321,10 @@ def normalize_options(name, version, opts)
 
       normalize_hash(opts)
 
+      if opts["branch"] && !(opts["git"] || opts["github"])
```
segiddins: Why not do this in `validate_keys`?

colby-swandale: I was being lazy and really should have checked. Thanks for the feedback!

spec/bundler/dsl_spec.rb
```diff
@@ -144,6 +144,11 @@
       expect { subject.gem(:foo) }.
         to raise_error(Bundler::GemfileError, /You need to specify gem names as Strings. Use 'gem "foo"' instead/)
     end
+
+    it "rejects branch option on non-git gems" do
+      expect { subject.gem("foo", :branch => "test") }.
+        to raise_error(Bundler::GemfileError, /The `branch` option for `gem "foo"` is not allowed. Only gems with a git source can specify a branch/)
```
segiddins: quotes don't match, hence the test failure

spec/bundler/dsl_spec.rb
```diff
@@ -144,6 +144,11 @@
       expect { subject.gem(:foo) }.
         to raise_error(Bundler::GemfileError, /You need to specify gem names as Strings. Use 'gem "foo"' instead/)
     end
+
+    it "rejects branch option on non-git gems" do
```
segiddins: please also test that specifying a branch when using a user-defined git source still passes? (or make sure such a test exists?)

segiddins:
@homu r+

segiddins:
@bundlerbot r+

PR #5565, opened 2017-04-06
dsl parsing performance improvement

colby-swandale:
i noticed in the Dsl that we're storing the `gem` dependencies in an array and performing certain operations with it like: `@dependencies.find {|d| d.name == dep.name }`. This can be improved and simplified with using a hash map instead.

Before:
```
› ruby -I ~/Desktop/Projects/bundler/lib benchmark.rb
  2.130000   0.020000   2.150000 (  2.166532)
› ruby -I ~/Desktop/Projects/bundler/lib benchmark.rb
  2.160000   0.030000   2.190000 (  2.204423)
› ruby -I ~/Desktop/Projects/bundler/lib benchmark.rb
  2.130000   0.030000   2.160000 (  2.184379)

```  

After:
```
› ruby -I ~/Desktop/Projects/bundler/lib benchmark.rb
  1.880000   0.020000   1.900000 (  1.919140)
› ruby -I ~/Desktop/Projects/bundler/lib benchmark.rb
  1.890000   0.020000   1.910000 (  1.933193)
```

Script: `ruby -I ~/Desktop/Projects/bundler/lib benchmark.rb`
```
require 'bundler'
require 'bundler/dsl'
require 'benchmark'

dsl = Bundler::Dsl.new
puts Benchmark.measure { 100.times { dsl.eval_gemfile('Gemfile') } }
```

Gemfile: https://gist.github.com/colby-swandale/5d04d5d72b6847bc242eeef87f9035b7

It's not a drastic improvement but i think is worth having.

### DIFF
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -21,7 +21,7 @@ module Bundler
       @source               = nil
       @sources              = SourceList.new
       @git_sources          = {}
-      @dependencies         = []
+      @dependencies         = {}
       @groups               = []
       @install_conditionals = []
       @optional_groups      = []
@@ -94,10 +94,10 @@ module Bundler
       dep = Dependency.new(name, version, options)
 
       # if there's already a dependency with this name we try to prefer one
-      if current = @dependencies.find {|d| d.name == dep.name }
+      if current = dependencies[dep.name]
         if current.requirement != dep.requirement
           if current.type == :development
-            @dependencies.delete current
+            dependencies.delete current
           else
             return if dep.type == :development
             raise GemfileError, "You cannot specify the same gem twice with different version requirements.\n" \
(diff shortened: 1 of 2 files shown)


segiddins:
Why did you close this?

colby-swandale:
Hey, basically it's because i really didn't do my research enough before making this PR. i was having a look at the failing specs, specifically `require_spec.rb`:

```
  describe "a gem with different requires for different envs" do
      before(:each) do
        build_gem "multi_gem", :to_system => true do |s|
          s.write "lib/one.rb", "puts 'ONE'"
          s.write "lib/two.rb", "puts 'TWO'"
        end

        install_gemfile <<-G
          gem "multi_gem", :require => "one", :group => :one
          gem "multi_gem", :require => "two", :group => :two
        G
      end

      it "requires both with Bundler.require(both)" do
        run "Bundler.require(:one, :two)"
        expect(out).to eq("ONE\nTWO")
      end
```

And noticed something i didn't know about Bundler, you can specify a gem multiple times with different options such as `require` and `group`. My change kinda breaks down for this behaviour and i didn't want to make anything more complicated for not so much benefit.

segiddins:
👍🏻

PR #5568, opened 2017-04-07
[StubSpecification] Avoid loading the full spec when possible

segiddins:
This should save us the hit of needing to `eval` all the gemspecs we use when the user has a modern (>= 2.3) RubyGems

### DIFF
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -320,18 +320,22 @@ module Bundler
       end
     end
 
+    def binstubs_call_gem?
+      true
+    end
+
     def replace_gem(specs, specs_by_name)
       reverse_rubygems_kernel_mixin
 
-      executables = nil
+      executables = specs.map(&:executables).flatten if binstubs_call_gem?
 
       kernel = (class << ::Kernel; self; end)
       [kernel, ::Kernel].each do |kernel_class|
         redefine_method(kernel_class, :gem) do |dep, *reqs|
-          executables ||= specs.map(&:executables).flatten
-          if executables.include? File.basename(caller.first.split(":").first)
+          if executables && executables.include?(File.basename(caller.first.split(":").first))
             break
           end
+
           reqs.pop if reqs.last.is_a?(Hash)
 
           unless dep.respond_to?(:name) && dep.respond_to?(:requirement)
@@ -795,6 +799,12 @@ module Bundler
         Bundler.ui = nil
         activated_spec_names = runtime.requested_specs.map(&:to_spec).sort_by(&:name)
         [Gemdeps.new(runtime), activated_spec_names]
+      end
+
+      if provides?(">= 2.5.2")
+        def binstubs_call_gem?
+          false
+        end
       end
     end
   end
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -437,8 +437,10 @@ module Bundler
         # Copy of Rubygems activate_bin_path impl
         requirement = args.last
         spec = find_spec_for_exe name, exec_name, [requirement]
-        Gem::LOADED_SPECS_MUTEX.synchronize { spec.activate }
-        spec.bin_file exec_name
+
+        gem_bin = File.join(spec.full_gem_path, spec.bindir, exec_name)
+        gem_from_path_bin = File.join(File.dirname(spec.loaded_from), spec.bindir, exec_name)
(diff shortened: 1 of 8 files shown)


lib/bundler/stub_specification.rb
```diff
@@ -9,23 +9,86 @@ def self.from_stub(stub)
...
+
+    def activated=(activated)
+      stub.instance_variable_set(:@activated, activated)
+    end
+
+    def default_gem
+      stub.default_gem
+    end
+
+    # This is what we do in bundler/rubygems_ext
```
segiddins: might be worth extracting these two methods into a module we can `include` instead of the copy-paste

lib/bundler/stub_specification.rb
```diff
@@ -9,23 +9,86 @@ def self.from_stub(stub)
...
+
   private
 
     def _remote_specification
-      stub.to_spec
+      @_remote_specification ||= begin
+        rs = stub.to_spec
+        if rs.equal?(self) # happens when to_spec gets the spec from Gem.loaded_specs
+          rs = Gem::Specification.load(loaded_from)
+          stub.instance_variable_set(:@spec, rs)
```
segiddins: probably worth moving this into `Bundler.rubygems`

lib/bundler/source/git.rb
```diff
@@ -302,6 +303,13 @@ def fetch
 
       # no-op, since we validate when re-serializing the gemspec
       def validate_spec(_spec); end
+
+      if defined?(::Gem::StubSpecification)
```
segiddins: might be worth adding `Bundler.rubygems.includes_stubs?` ?

lib/bundler/rubygems_integration.rb
```diff
@@ -323,11 +323,12 @@ def reverse_rubygems_kernel_mixin
     def replace_gem(specs, specs_by_name)
       reverse_rubygems_kernel_mixin
 
-      executables = specs.map(&:executables).flatten
+      executables = nil
 
       kernel = (class << ::Kernel; self; end)
       [kernel, ::Kernel].each do |kernel_class|
         redefine_method(kernel_class, :gem) do |dep, *reqs|
+          executables ||= specs.map(&:executables).flatten
```
segiddins: I'm not happy with this change, I'd like to investigate if this check is even necessary. It was introduced 7 years ago without any test coverage...

segiddins: it isn't needed

lib/bundler/rubygems_integration.rb
```diff
@@ -795,6 +802,12 @@ def use_gemdeps(gemfile)
         activated_spec_names = runtime.requested_specs.map(&:to_spec).sort_by(&:name)
         [Gemdeps.new(runtime), activated_spec_names]
       end
+
+      if provides?(">= 2.5.2")
```
jules2689: Would it be worth a comment on this one?

segiddins: I thought it was rather self explanitory -- on 2.5.2+, bindstubs don't call `gem`. What sort of comment were you looking for?

jules2689: Oh, nvm, it just seemed like a sot of edge case. nvm my comment

segiddins: this entire file is "hacks to support 7 years of rubygems versions", in that context nothing is an edge case :P

jules2689:
So here is an overview. I ran this 30 times on the Shopify application:

image

|         | 1.14.6  | master  | seg-stubs-no-load-full-spec | 
|----------|---------|---------|-----------------------------| 
| std. dev | 0.05631 | 0.03713 | 0.02711                     | 
| min      | 0.68303 | 0.50974 | 0.36393                     | 
| max      | 0.95849 | 0.63679 | 0.47162                     | 
| median   | 0.73591 | 0.54985 | 0.39107                     | 
| average  | 0.74377 | 0.55676 | 0.39813                     | 

With this change, we vary less than half of the amount as before. This makes the timings _less volatile_. The new maximum is less than the minimum on master (by about 40ms) and less than the minimum on 1.14.6 (by 210ms). The average time is ~160ms less than on master, and 350ms less than on 1.14.6.

segiddins:
This is now passing?
Please review, @indirect?

indirect:
This is awesome. Thanks so much, @jules2689 and @segiddins!

indirect:
@bundlerbot r+

PR #5570, opened 2017-04-08
[Worker] Allow interrups to exit the process even after the threads are disposed

segiddins:
This means that SIGINT during, say, dependency resolution will still cleanly exit the process.


indirect:
good catch! @bundlerbot r+

Issue #183, opened 2010-03-05
Add the ability to add a secondary Gemfile (that can be global to the system for example)


carllerche:
- the information in the secondary gemfile are lower priority
- Do not include the secondary gemfile in a lock
- the secondary Gemfile should not effect the resolve of the main Gemfile. It is a superset on top of it.
- Add the ability to specify this secondary gemfile with an ENV variable, .bundle/config, or flag.

CodyBrouwers:
For anyone that ends up coming across this thread here what I've done, that seems to work but does still have one flaw:

1. Create a Gemfile.local file and add:
   ```ruby
   # Include the regular Gemfile
    eval File.read('Gemfile')
   ```
2. Add any other gems you want to use to the ` group :development[, :test]` (I wouldn't suggest doing this for any other environment)
3. Run `bundle install --gemfile=Gemfile.local`
4. You will now have a Gemfile.local.lock
5. You can now use the gem by running `BUNDLE_GEMFILE=Gemfile.local bundle exec {{gem command}}`
6. Or to make Bundler prioritize Gemfile.local gems, you can specify in `{{app root}}/.bundle/config`:
    `BUNDLE_GEMFILE: "Gemfile.local"`

The flaw in this is that since there are two Gemfile lock files, there isn't a way to ensure consistent versions between the two so something to aware of. You can at least check for inconsistencies by not adding any additional gems to the Gemfile.local and comparing the two lock files, so it's not a dealbreaker for me.

indirect:
Thanks for reporting back! I'm glad you've found something that works for your needs. 👍

rlue:
> The flaw in this is that since there are two Gemfile lock files, there isn't a way to ensure consistent versions between the two so something to aware of. 
I use the following bash function to keep my lockfiles in sync:

```bash
function bundle() {
  bundle="$(type -P bundle)"

  if [ -r "$BUNDLE_GEMFILE" ] && [ -r Gemfile ] &&
     [ "$BUNDLE_GEMFILE" != Gemfile ] && 
     [[ "$1" =~ ^(|install|update)$ ]]; then 
    BUNDLE_GEMFILE=Gemfile "$bundle" "$@"
    cp Gemfile.lock "${BUNDLE_GEMFILE}.lock"
  fi

  "$bundle" "$@"
}
```

bmulholland:
[...]

PR #5584, opened 2017-04-13
[WIP] Experimental Gemfile checksum feature

hmistry:
Generate checksum for Gemfile and save to lock file. Use the previous…ly saved checksum in the lock file to compare against the current Gemfile checksum to determine if there were any changes made. If no changes, then skip resolution.

This is a quick hack to evaluate if there is any performance benefit of this feature. For a single rails gem Gemfile, it saved 130ms on my local machine (0.430s to 0.3s) to run bundle install.

Submitting PR for feedback (@indirect , @segiddins) and other possible performance tests to evaluate further.

Issue #4871

### DIFF
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -30,6 +30,7 @@ module Bundler
       @ruby_version         = nil
       @gemspecs             = []
       @gemfile              = nil
+      @checksum             = nil
       add_git_sources
     end
 
@@ -39,6 +40,7 @@ module Bundler
       @gemfile = expanded_gemfile_path
       contents ||= Bundler.read_file(gemfile.to_s)
       instance_eval(contents.dup.untaint, gemfile.to_s, 1)
+      @checksum = generate_checksum(contents)
     rescue Exception => e
       message = "There was an error " \
         "#{e.is_a?(GemfileEvalError) ? "evaluating" : "parsing"} " \
@@ -197,7 +199,7 @@ module Bundler
     end
 
     def to_definition(lockfile, unlock)
-      Definition.new(lockfile, @dependencies, @sources, unlock, @ruby_version, @optional_groups)
+      Definition.new(lockfile, @dependencies, @sources, unlock, @ruby_version, @optional_groups, @checksum)
     end
 
     def group(*args, &blk)
@@ -447,6 +449,10 @@ The :#{name} git source is deprecated, and will be removed in Bundler 2.0. Add t
       "#{repo_string}"
     end
       EOS
+    end
+
+    def generate_checksum(gemfile_source)
+      Digest::MD5.hexdigest(gemfile_source)
     end
 
     class DSLError < GemfileError
--- a/lib/bundler/installer.rb
+++ b/lib/bundler/installer.rb
@@ -75,6 +75,10 @@ module Bundler
         return
       end
 
+
+      # EXPERIMENT: Evaluate benefit if we skip resolution by comparing Gemfile checksum against the checksum saved in the lockfile.
+      return if Definition.build(Bundler.default_gemfile, Bundler.default_lockfile, nil).checksum_matches?
+
       resolve_if_need(options)
       ensure_specs_are_compatible!
       install(options)
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -51,7 +51,7 @@ module Bundler
     #   to be updated or true if all gems should be updated
     # @param ruby_version [Bundler::RubyVersion, nil] Requested Ruby Version
     # @param optional_groups [Array(String)] A list of optional groups
-    def initialize(lockfile, dependencies, sources, unlock, ruby_version = nil, optional_groups = [])
+    def initialize(lockfile, dependencies, sources, unlock, ruby_version = nil, optional_groups = [], checksum = nil)
       @unlocking = unlock == true || !unlock.empty?
 
       @dependencies    = dependencies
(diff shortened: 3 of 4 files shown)


indirect:
@hmistry thanks! I'm looking forward to seeing how this turns out.

@jules2689 I'm interested to see if this works for you, and if it saves any time on exec or install.

colby-swandale:
@indrect IIRC we were talking about this a while back and there was talk that this could be an issue for Gemfiles that has ruby that is evaluated? Does this need to have an opt-out/opt-in flag?

lib/bundler/installer.rb
```diff
@@ -75,6 +75,10 @@ def run(options)
         return
       end
 
+
+      # EXPERIMENT: Evaluate benefit if we skip resolution by comparing Gemfile checksum against the checksum saved in the lockfile.
+      return if Definition.build(Bundler.default_gemfile, Bundler.default_lockfile, nil).checksum_matches?
```
segiddins: this will also skip installation even if none of the gems are actually installed.... I think maybe you might've meant something like the following?

```diff
diff --git a/lib/bundler/installer.rb b/lib/bundler/installer.rb
index cbfdddeaf..6474d182f 100644
--- a/lib/bundler/installer.rb
+++ b/lib/bundler/installer.rb
@@ -212,6 +212,8 @@ module Bundler
     end
 
     def resolve_if_need(options)
+      return if @definition.nothing_changed?
+
       if !options[“update”] && !options[:inline] && Bundler.default_lockfile.file?
         local = Bundler.ui.silence do
           begin
```

jules2689:
> Gemfiles that has ruby that is evaluated

That would be a problem for us, yea. We do some juggling magic to support 2 Rails versions at a time.

Could this be used for the `bundle check` command maybe?

---

> I'm interested to see if this works for you, and if it saves any time on exec or install.

@indirect I'm happy to test this out, but it needs a rebase on master. It's missing the work that we've put in. I rebased locally to get the numbers below.

|   | 1.14.6 | master | this branch |
| --- | --- | --- | --- |
| `bundle install` with all deps met | 3.67s | 3.4s | 0.67s |
| `bundle install` with a missing dep | 15-16s | 15-16s  | ERROR |


The big difference I noticed is that there is less output, which is likely saving a lot of time so :100:

#### ERROR

```bash
~/src/github.com/Shopify/shopify(master*) ➜ time bundle install
Could not find json_schema-0.14.3 in any of the sources

real	0m0.639s
user	0m0.542s
sys	0m0.057s
```

jules2689:
> Gemfiles that has ruby that is evaluated

Actually, looking at the way we generate the checksum we can likely hit most use cases by supporting a separate gemfile lockfile (shopify's use case) via an env var*, combine that with an opt-out and we should be good.

`*` We have a single Gemfile so that people don't forget to add stuff, and resolve to separate lockfiles.

lib/bundler/dsl.rb
```diff
@@ -449,6 +451,10 @@ def warn_deprecated_git_source(name, repo_string)
       EOS
     end
 
+    def generate_checksum(gemfile_source)
+      Digest::MD5.hexdigest(gemfile_source)
```
jules2689: We can't use MD5 because of FIPS :( https://github.com/bundler/bundler/issues/4989

jules2689:
Another thing to note here is that gemfiles can source other gemfiles, which has been suggested in many instances such as https://github.com/bundler/bundler/issues/183#issuecomment-285773467

lib/bundler/installer.rb
hmistry: I already mentioned to @segiddins but for documentation purposes, this did not result in any time improvement.

hmistry:
@jules2689 and @colby-swandale thanks for all your feedback. This quick hack does not account for all the edge cases and was simply done to evaluate what performance benefit we could get from this approach. 
Looking at the initial data, thanks to @jules2689, there is a good gain so now the challenge will be to see how to address each of the edge cases and trade off with the time improvement.

colby-swandale:
ping @hmistry is still being worked on?

hmistry:
@colby-swandale I haven't had a chance to work on this since then. It's more of an exploration to see if there's any improvement. Is this something needed soon?

colby-swandale:
No but let us know if you're not working on this anymore.

agrim123:
@hmistry Are you still working on this? If not, I would be happy to continue the work!

segiddins:
I had something a bit more advanced in https://github.com/bundler/bundler/compare/seg-definition-from-lockfile, you might want to take a look at that as well

hmistry:
@agrim123 No, I haven't had time to continue working on this. Please go ahead.

agrim123:
@segiddins Can you explain what is `optional_groups` [here](https://github.com/bundler/bundler/compare/seg-definition-from-lockfile?diff=unified&name=seg-definition-from-lockfile#diff-ec267c71117bf68f38c928a3691ba98cR19) means exactly? Is it related to checksum part?
Is there a need to [store](https://github.com/bundler/bundler/compare/seg-definition-from-lockfile?diff=unified&name=seg-definition-from-lockfile#diff-3919b62708ad380340b355b09ccd7b8cR91) as key-value pair for Gemfile, when the only checksum of the corresponding gemfile matters, as did [here](https://github.com/bundler/bundler/pull/5584/files?diff=split#diff-ec267c71117bf68f38c928a3691ba98cR410)? I didn't quite understand this.

segiddins:
It’s needed because some people use dynamic ruby code in their gemfile, so simply checking the checksum matches is insufficient to guarantee a semantic match

agrim123:
@segiddins  I had one more doubt.
If I add a gem manually to the `gems.rb` and then run `bundle install`, then the  variable `static_gemfile` does not gets updated and thus no `GEMFILE CHECKSUMS` section is created. When do we want to  save the checksum?

segiddins:
As things are on my branch, you need to manually call`_static_gemfile!()` in your Gemfile to opt-in to the behavior

PR #5587, opened 2017-04-15
Dsl improvments

colby-swandale:
This PR contains a couple of small improvements to help remove complexity in the DSL. Refer to the commit message for context.

PS. Don't merge this in yet as i might have a few more commits to come.


lib/bundler/dsl.rb
```diff
@@ -46,7 +44,7 @@ def eval_gemfile(gemfile, contents = nil)
 
       raise DSLError.new(message, gemfile, e.backtrace, contents)
     ensure
-      @gemfile = original_gemfile
+      @gemfile = gemfile
```
segiddins: this is incorrect, it needs to reference the value of `@gemfile` before the method was invoked

colby-swandale: yea, learned that the hard way 😆

lib/bundler/dsl.rb
```diff
@@ -216,9 +217,9 @@ def group(*args, &blk)
       args.each { @groups.pop }
     end
 
-    def install_if(*args, &blk)
```
segiddins: Please keep the block arg? That way calling without a block will have a better error message

colby-swandale: Should this have a proper error message then if the user doesn't supply a block?

colby-swandale: @segiddins ?

segiddins: nvm, they give the same error

segiddins:
@bundlerbot r+

PR #5589, opened 2017-04-15
[Env] Default to always printing gemfile & gemspecs

segiddins:
This will save a lot of time spent asking for them, and is something I thought we'd already done

### DIFF
--- a/lib/bundler/cli/issue.rb
+++ b/lib/bundler/cli/issue.rb
@@ -26,7 +26,7 @@ module Bundler
 
       EOS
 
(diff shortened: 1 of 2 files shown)


colby-swandale:
This is another nitpick and not necessary to your changes but while you're changing `Env` i would like to change the way we call the report from `Bundler::Env.new.report` to just `Bundler::Env.report`.

Is this a problem because backwards compatibility? Any thoughts?

segiddins:
That should probably be in a separate PR

segiddins:
@bundlerbot r+

PR #5590, opened 2017-04-17
move implementation of Bundler::Env to class methods

colby-swandale:
This is to a comment i made in #5589 

Currently to print the Bundler env report you need to need to invoke it
via:

  Bundler::Env.new.report

This change removes the need to instantiate an instance and just call
a class method of the same name:

  Bundler::Env.report

### DIFF
--- a/lib/bundler/env.rb
+++ b/lib/bundler/env.rb
@@ -4,11 +4,11 @@ require "bundler/source/git/git_proxy"
 
 module Bundler
   class Env
-    def write(io)
+    def self.write(io)
       io.write report
     end
 
-    def report(options = {})
+    def self.report(options = {})
       print_gemfile = options.delete(:print_gemfile) { true }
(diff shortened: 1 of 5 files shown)


lib/bundler/env.rb
```diff
@@ -4,89 +4,91 @@
 
 module Bundler
   class Env
-    def write(io)
-      io.write report
-    end
+    class << self
+      def write(io)
```
segiddins: Can you change these to `def self.` to avoid reopening the class and thus the indentation? That way we can avoid the massive diffs

colby-swandale: Something i have not considered before. Thanks

segiddins:
👍 other than the whitespace

segiddins:
Normally unnecessary, but would you mind squashing just the once?

colby-swandale:
no worries

segiddins:
@bundlerbot r+

Issue #5594, opened 2017-04-18
Gems defined with extensions dirs conflict with diff versions


jules2689:
Problem:
---
- Using ruby version X, two projects define some gem at differing versions.
- Stub Specification does not pass extension dir, which causes issues.

We saw this as

```
** Notice: CityHash was not loaded. **

For optimal performance, use of the cityhash gem is recommended.

Run the following command, or add it to your Gemfile:

gem install cityhash
```

## Environment

```
Bundler   1.15.0.pre.1
Rubygems  2.6.11
Ruby      2.3.3p222 (2016-11-21 revision 56859) [x86_64-darwin16]
GEM_HOME  /Users/ryanmcilmoyl/.gem/ruby/2.3.3
GEM_PATH  /Users/ryanmcilmoyl/.gem/ruby/2.3.3:/opt/rubies/2.3.3/lib/ruby/gems/2.3.0
Git       2.11.0 (Apple Git-81)
Platform  x86_64-darwin-16
```

## Bundler settings

```
jobs
  Set for your local app (/Users/ryanmcilmoyl/src/github.com/Shopify/shopify/.bundle/config): "4"
```

## Relevant Gemfile

Shopify Project:

`gem 'cityhash', git: 'csfrancis/cityhash', sha: 'XYZ'`

Another Project

`gem 'cityhash'`

segiddins:
> see lib/bundler/rubygems_ext.rb, we need that impl of `extension_dir` for the stub spec, just replacing `rg_extension_dir` with `stub.extension_dir`

jules2689:
To replicate:

Gemfile
```ruby
source 'https://rubygems.org'
gem 'cityhash', git: 'https://github.com/csfrancis/cityhash.git', ref: '3cfc7d01f333c01811d5e834f1495eaa29f87c36'
```

test.rb
```ruby
require 'bundler/setup'
puts $:.grep(/cityhash/)
```

Which outputs
```shell
/Users/juliannadeau/.gem/ruby/2.3.3/bundler/gems/cityhash-3cfc7d01f333/lib
/Users/juliannadeau/.gem/ruby/2.3.3/bundler/gems/extensions/x86_64-darwin-16/2.3.0-static/cityhash-0.6.0
```

That `cityhash-0.6.0` should be a hash

jules2689:
[...]

PR #5595, opened 2017-04-18
Apply source and git_version to stub spec extension

jules2689:
Fixes https://github.com/bundler/bundler/issues/5594

`stub_specification` doesn't know what `full_name` should be because we don't have a record of the source, so it resolves to `cityhash-0.6.0` rather than `cityhash-3cfc7d01f333`

This makes `StubSpecification` aware of the source to we can still avoid loading the full spec when needed. This is done by bringing remote_specs `attr_accessor :source` to `stub_spec`, and assigning source to the `stub`.

I am probably missing some edge cases to check for though 🙈  Need to write specs.

cc @segiddins

### DIFF
--- a/lib/bundler/stub_specification.rb
+++ b/lib/bundler/stub_specification.rb
@@ -11,6 +11,18 @@ module Bundler
 
     attr_accessor :stub, :ignored
 
+    def source=(source)
+      super
+      # Stub has no concept of source, which means that extension_dir may be wrong
+      # This is the case for git-based gems. So, instead manually assign the extension dir
+      if source.respond_to?(:extension_dir_name)
+        path = File.join(stub.extensions_dir, source.extension_dir_name)
+        stub.extension_dir = File.expand_path(path)
+      else
+        stub.extension_dir = nil
+      end
+    end
+
     def to_yaml
       _remote_specification.to_yaml
     end
--- a/lib/bundler/stub_specification.rb
+++ b/lib/bundler/stub_specification.rb
@@ -11,15 +11,17 @@ module Bundler
 
     attr_accessor :stub, :ignored
 
-    def source=(source)
-      super
-      # Stub has no concept of source, which means that extension_dir may be wrong
-      # This is the case for git-based gems. So, instead manually assign the extension dir
-      if source.respond_to?(:extension_dir_name)
-        path = File.join(stub.extensions_dir, source.extension_dir_name)
-        stub.extension_dir = File.expand_path(path)
-      else
-        stub.extension_dir = nil
+    # Pre 2.2.0 did not include extension_dir
+    # https://github.com/rubygems/rubygems/commit/9485ca2d101b82a946d6f327f4bdcdea6d4946ea
+    if Bundler.rubygems.provides?(">= 2.2.0")
+      def source=(source)
+        super
+        # Stub has no concept of source, which means that extension_dir may be wrong
+        # This is the case for git-based gems. So, instead manually assign the extension dir
+        if source.respond_to?(:extension_dir_name)
+          path = File.join(stub.extensions_dir, source.extension_dir_name)
+          stub.extension_dir = File.expand_path(path)
+        end
       end
     end
 
--- a/lib/bundler/stub_specification.rb
+++ b/lib/bundler/stub_specification.rb
@@ -18,10 +18,9 @@ module Bundler
         super
         # Stub has no concept of source, which means that extension_dir may be wrong
         # This is the case for git-based gems. So, instead manually assign the extension dir
-        if source.respond_to?(:extension_dir_name)
-          path = File.join(stub.extensions_dir, source.extension_dir_name)
-          stub.extension_dir = File.expand_path(path)
-        end
+        return unless source.respond_to?(:extension_dir_name)
+        path = File.join(stub.extensions_dir, source.extension_dir_name)
+        stub.extension_dir = File.expand_path(path)
       end
     end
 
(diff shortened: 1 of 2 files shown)


lib/bundler/rubygems_ext.rb
```diff
@@ -23,6 +23,21 @@
 module Gem
   @loaded_stacks = Hash.new {|h, k| h[k] = [] }
 
+  class StubSpecification
```
segiddins: Stub specification isn't always defined, and I'd like us to avoid adding even more re-opening of classes we don't own

jules2689: Our spec doesn't call `extension_dir` at all since stub is what the `extension_dir` is called on. So `extension_dir` needs to be defined on `Gem::StubSpecification`

jules2689: Oh wait, I see a way around it

jules2689:
@segiddins I manually assigned `extension_dir` instead since it is an `attr_writer` on `Gem::BasicSpecification` :+1:

jules2689:
For the specs, I'm going to write a unit test for this, but first I'll have to merge #5594 otherwise my file will conflict

lib/bundler/stub_specification.rb
```diff
@@ -9,7 +9,20 @@ def self.from_stub(stub)
       spec
     end
 
-    attr_accessor :stub, :ignored
+    attr_accessor :stub, :ignored, :source
+
+    def source=(source)
```
segiddins: This will warn, need to make source a reader

segiddins:
👍 with test coverage, which I'm happy to write if you'd like

segiddins:
```diff
diff --git a/lib/bundler/stub_specification.rb b/lib/bundler/stub_specification.rb
index 257a47222..0876d7c6a 100644
--- a/lib/bundler/stub_specification.rb
+++ b/lib/bundler/stub_specification.rb
@@ -68,6 +68,13 @@ module Bundler
       stub.raw_require_paths
     end
 
+    def source=(source)
+      super
+      if source.respond_to?(:extension_dir_name)
+        stub.extension_dir = File.expand_path(File.join(stub.extensions_dir, source.extension_dir_name))
+      end
+    end
+
   private
 
     def _remote_specification
diff --git a/spec/install/gemfile/git_spec.rb b/spec/install/gemfile/git_spec.rb
index d2d46279a..8843fb7eb 100644
--- a/spec/install/gemfile/git_spec.rb
+++ b/spec/install/gemfile/git_spec.rb
@@ -1030,6 +1030,11 @@ RSpec.describe "bundle install with git sources" do
         puts FOO
       R
       expect(out).to eq("YES")
+
+      run! <<-R
+        puts $:.grep(/ext/)
+      R
+      expect(out).to eq(Pathname.glob(system_gem_path("bundler/gems/extensions/**/foo-1.0-*")).first.to_s)
     end
 
     it "does not use old extension after ref changes" do
```

segiddins:
just confirmed that diff fixes the issue with cityhash

jules2689:
Updated to use the diff. I opted to keep the attr_reader and move the `source` to the super class (stub_spec) still. I also opted to keep the extension_dir set to nil if it doesn't respond because that will [reset the cache in rubygems](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/basic_specification.rb#L90-L92)

lib/bundler/remote_specification.rb
```diff
@@ -12,7 +12,7 @@ class RemoteSpecification
 
     attr_reader :name, :version, :platform
     attr_writer :dependencies
-    attr_accessor :source, :remote
```
segiddins: You can't do this, it breaks thing. All remote specs have a source

jules2689: ugh, for some reason I thought this was a subclass, but it's the other way around 🙈

jules2689: fixed.

lib/bundler/stub_specification.rb
```diff
@@ -11,6 +11,18 @@ def self.from_stub(stub)
...
 
+    def source=(source)
+      super
+      # Stub has no concept of source, which means that extension_dir may be wrong
+      # This is the case for git-based gems. So, instead manually assign the extension dir
+      if source.respond_to?(:extension_dir_name)
+        path = File.join(stub.extensions_dir, source.extension_dir_name)
+        stub.extension_dir = File.expand_path(path)
+      else
+        stub.extension_dir = nil
```
segiddins: Why is assigning nil advantageous?

jules2689: I commented here: https://github.com/bundler/bundler/pull/5595#issuecomment-295017826

I can add a comment if you'd like.

segiddins: There's no reason to reset the cache since none of the values it sepends on have changed

jules2689: I was more thinking forwards for posterity, if some one sets a source on this that doesn't support extension_dir_name then it might cause an issue.

lib/bundler/stub_specification.rb
```diff
@@ -11,6 +11,21 @@ def self.from_stub(stub)
...
+    def source=(source)
+      super
+      # Stub has no concept of source, which means that extension_dir may be wrong
+      # This is the case for git-based gems. So, instead manually assign the extension dir
+      if source.respond_to?(:extension_dir_name)
+        path = File.join(stub.extensions_dir, source.extension_dir_name)
+
+        # Pre 2.2.0 did not include extension_dir
+        # https://github.com/rubygems/rubygems/commit/9485ca2d101b82a946d6f327f4bdcdea6d4946ea
+        if Bundler.rubygems.provides?(">= 2.2.0")
```
segiddins: Let's wrap the whole method def into the conditional for performance?

segiddins:
`bin/rubocop`

segiddins:
@bundlerbot r+

Issue #5599, opened 2017-04-19
duplicate dependencies in Gemfile.lock on bundle update


embark:
# Error Report
## Questions

**What did you do?**
I ran the commands:
- with the Gemfile and environment listed below
- `bundle`
- `gem uninstall responders`
- `bundle update easypost`

**What did you expect to happen?**
I expected Bundler to create a `Gemfile.lock` after `bundle update easypost` that was identical to the `Gemfile.lock` right after `bundle`

**What happened instead?**
Instead, Bundler duplicated `responders` in the gemfile:
```
...
     responders (2.3.0)
       railties (>= 4.2.0, < 5.1)
+    responders (2.3.0)
+      railties (>= 4.2.0, < 5.1)
     rest-client (2.0.1)
       http-cookie (>= 1.0.2, < 2.0)
...
```
**NOTE: this only happened when using a block with the `source`, not if the `source` was just listed at the top of the `Gemfile`.**

**Have you tried any solutions posted on similar issues in our issue tracker, stack overflow, or google?**
Couldn't find any similar issues, but might have missed it.

## Environment

```
Bundler   1.14.6
Rubygems  2.5.2
Ruby      2.3.4p301 (2017-03-30 revision 58214) [x86_64-darwin16]
GEM_HOME  ~/.rvm/gems/ruby-2.3.4
GEM_PATH  ~/.rvm/gems/ruby-2.3.4:~/.rvm/gems/ruby-2.3.4@global
RVM       1.27.0 (latest)
Git       2.11.0
Platform  x86_64-darwin-16
```

## Bundler settings

```
build.puma
  Set for the current user (~/.bundle/config): "--with-opt-dir=/usr/local/opt/openssl"
build.eventmachine
  Set for the current user (~/.bundle/config): "--with-opt-dir=/usr/local/opt/openssl"
```

### Gemfile

```ruby
source 'https://rubygems.org/' do
  gem 'easypost'
  gem 'responders'
end
```

### Gemfile.lock

```
GEM
  remote: https://rubygems.org/
  specs:
    actionpack (5.0.2)
      actionview (= 5.0.2)
      activesupport (= 5.0.2)
      rack (~> 2.0)
      rack-test (~> 0.6.3)
      rails-dom-testing (~> 2.0)
      rails-html-sanitizer (~> 1.0, >= 1.0.2)
    actionview (5.0.2)
      activesupport (= 5.0.2)
      builder (~> 3.1)
      erubis (~> 2.7.0)
[...]

PR #5600, opened 2017-04-20
Avoid duplicating specs in the lockfile after updating w/ the gem uninstalled

segiddins:
Closes https://github.com/bundler/bundler/issues/5599

I decided to not compare using `full_name` in `SpecSet#merge` for the sake of performance, since `#full_name` is uncached

### DIFF
--- a/lib/bundler/lazy_specification.rb
+++ b/lib/bundler/lazy_specification.rb
@@ -29,7 +29,7 @@ module Bundler
       @name          = name
       @version       = version
       @dependencies  = []
-      @platform      = platform
+      @platform      = platform || Gem::Platform::RUBY
       @source        = source
       @specification = nil
     end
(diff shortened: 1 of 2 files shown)


lib/bundler/lazy_specification.rb
```diff
@@ -29,7 +29,7 @@ def initialize(name, version, platform, source = nil)
       @name          = name
       @version       = version
       @dependencies  = []
-      @platform      = platform
+      @platform      = platform || Gem::Platform::RUBY
```
embark: I don't have any context in this codebase, so I have no idea -- is this a safe default assumption to make?

segiddins: yes, a `nil` platform is treated as ruby

segiddins:
@bundlerbot r+

PR #5601, opened 2017-04-20
Use `Gem::Specification.load` to activate gems in `bundler/setup` standalone script

### LINKED ISSUE #2720: Standalone bundle does not setup Gem::Specifications dir correctly.
I'm using bundler (with the `--standalone` flag) in a non-rails project so that I can package the application in a tarball and run it without depending on bundler being installed on the target machines (I am assured of a consistent ruby environment).

Towards this end, I run bundler with the following invocation `bundle --standalone --gemfile=<path-to-gemfile> --without <groups-to-exclude>`

This results in the appropriate `$PROJ_ROOT/bundle/bundler/setup.rb` file being created, which when required will setup the $LOAD_PATH to include all the necessary gems from the corresponding `$PROJ_ROOT/bundle/ruby/<ver>/gems` directory.

However, it does not set the `Gem::Specifications.dir=$PROJ_ROOT/bundle/ruby/<ver>/specifications`. 

This causes a problem in the following situation:
1. The bundle contains activerecord (4.0.0)
2. If you try to use it with the sqlite3 adapter, activerecord will call `gem 'sqlite3', '~>1.3.6'`
3. Rubygems will try to look at its specification dir to see if it can satisfy the request
4. A `Gem::LoadError` is raised.

It seems to me that the easiest (and correct?) thing to do here is for the synthesised `bundler/setup.rb` to also setup the `Gem::Specifications.dir = #{path}/../#{ruby_engine}/#{ruby_version}` as it does for the gems (this is possible since the packaged standalone bundle has these available in an adjacent dir).

It is possible that I may be completely missing something here (and possibly using the `--standalone` mode incorrectly), however with this fix in place (in my application code) things seem to work. Please correct me if I'm wrong.

I am running:
bundler v.1.3.5 on ruby 1.9.3. (2011-10-30 patchlevel0)

I'll be glad to provide any additional information requested. If the correctness of the analysis can be verified, I'd be happy to open a PR for this.

joshuaspence:
Fixes #2720.

### DIFF
--- a/lib/bundler/installer/standalone.rb
+++ b/lib/bundler/installer/standalone.rb
@@ -30,10 +30,9 @@ module Bundler
           gemspec = gem_path(spec.loaded_from, spec).sub(version_dir, replacement)
           full_gem_path = gem_path(spec.full_gem_path, spec).sub(version_dir, replacement)
 
-          file.puts %(Gem::Specification.load("\#{path}/#{gemspec}").instance_eval do |spec|)
+          file.puts %(Gem::Specification.load("\#{path}/#{gemspec}").tap do |spec|)
           file.puts %(  spec.full_gem_path = "\#{path}/#{full_gem_path}")
-          file.puts "  spec.activate"
-          file.puts "end"
+          file.puts "end.activate"
         end
       end
     end
--- a/lib/bundler/installer/standalone.rb
+++ b/lib/bundler/installer/standalone.rb
@@ -15,23 +15,30 @@ module Bundler
         file.puts "ruby_engine = defined?(RUBY_ENGINE) ? RUBY_ENGINE : 'ruby'"
         file.puts "ruby_version = RbConfig::CONFIG[\"ruby_version\"]"
         file.puts "path = File.expand_path('..', __FILE__)"
-        paths.each do |path|
-          file.puts %($:.unshift "\#{path}/#{path}")
+
+        @specs.each do |spec|
+          next if spec.name == "bundler"
+
+          # Ensure that all `require_paths` exist.
+          Array(spec.require_paths).map do |path|
+            gem_path(path, spec)
+          end
+
+          # This is a static string intentionally. It's interpolated at a later time.
+          replacement = '#{ruby_engine}/#{ruby_version}'
+
+          gemspec = gem_path(spec.loaded_from, spec).sub(version_dir, replacement)
+          full_gem_path = gem_path(spec.full_gem_path, spec).sub(version_dir, replacement)
+
+          file.puts %(Gem::Specification.load("\#{path}/#{gemspec}").instance_eval do |spec|)
+          file.puts %(  spec.full_gem_path = "\#{path}/#{full_gem_path}")
+          file.puts "  spec.activate"
+          file.puts "end"
         end
       end
     end
 
   private
-
-    def paths
-      @specs.map do |spec|
-        next if spec.name == "bundler"
-        Array(spec.require_paths).map do |path|
-          gem_path(path, spec).sub(version_dir, '#{ruby_engine}/#{ruby_version}')
-          # This is a static string intentionally. It's interpolated at a later time.
-        end
-      end.flatten
-    end
 
     def version_dir
       "#{Bundler::RubyVersion.system.engine}/#{RbConfig::CONFIG["ruby_version"]}"
--- a/lib/bundler/installer/standalone.rb
+++ b/lib/bundler/installer/standalone.rb
@@ -31,7 +31,11 @@ module Bundler
           full_gem_path = gem_path(spec.full_gem_path, spec).sub(version_dir, replacement)
 
           file.puts %(Gem::Specification.load("\#{path}/#{gemspec}").tap do |spec|)
-          file.puts %(  spec.full_gem_path = "\#{path}/#{full_gem_path}")
+          file.puts %(  if spec.respond_to? :full_gem_path=)
+          file.puts %(    spec.full_gem_path = "\#{path}/#{full_gem_path}")
+          file.puts %(  else)
+          file.puts %(    spec.instance_variable_set(:@full_gem_path, "\#{path}/#{full_gem_path}"))
+          file.puts %(  end)
           file.puts "end.activate"
         end
       end
--- a/spec/install/gems/standalone_spec.rb
+++ b/spec/install/gems/standalone_spec.rb
@@ -29,6 +29,22 @@ RSpec.shared_examples "bundle install --standalone" do
       expect(out).to eq(expected_gems.values.join("\n"))
     end
 
+    it "loads gems correctly" do
+      testrb = String.new <<-RUBY
+        $:.unshift File.expand_path("bundle")
+        require "bundler/setup"
+
+      RUBY
+      expected_gems.each do |k, _|
+        testrb << "\nputs (Gem.loaded_specs['#{k}'] || Gem::Specification.new).name"
+      end
+      Dir.chdir(bundled_app) do
+        ruby testrb, :no_lib => true
+      end
+
+      expect(out).to eq(expected_gems.keys.join("\n")), out
+    end
+
     it "works on a different system" do
       FileUtils.mv(bundled_app, "#{bundled_app}2")
 
@@ -75,10 +91,19 @@ RSpec.shared_examples "bundle install --standalone" do
       G
     end
 
-    it "generates a bundle/bundler/setup.rb with the proper paths", :rubygems => "2.4" do
-      extension_line = File.read(bundled_app("bundle/bundler/setup.rb")).each_line.find {|line| line.include? "/extensions/" }.strip
-      expect(extension_line).to start_with '$:.unshift "#{path}/../#{ruby_engine}/#{ruby_version}/extensions/'
-      expect(extension_line).to end_with '/very_simple_binary-1.0"'
+    it "loads gems correctly", :rubygems => "2.4" do
+      testrb = String.new <<-RUBY
+        $:.unshift File.expand_path("bundle")
+        require "bundler/setup"
+
+        require 'very_simple_binary'
+        puts (Gem.loaded_specs['very_simple_binary'] || Gem::Specification.new).name
+      RUBY
+      Dir.chdir(bundled_app) do
+        ruby testrb, :no_lib => true
+      end
+
+      expect(out).to eq('very_simple_binary')
     end
   end
 
--- a/spec/install/gems/standalone_spec.rb
+++ b/spec/install/gems/standalone_spec.rb
@@ -92,18 +92,16 @@ RSpec.shared_examples "bundle install --standalone" do
     end
 
     it "loads gems correctly", :rubygems => "2.4" do
-      testrb = String.new <<-RUBY
-        $:.unshift File.expand_path("bundle")
+      testrb = <<-RUBY
+        $:.unshift "#{bundled_app("bundle")}"
         require "bundler/setup"
 
-        require 'very_simple_binary'
-        puts (Gem.loaded_specs['very_simple_binary'] || Gem::Specification.new).name
+        require 'very_simple_binary_c'
+        puts VerySimpleBinaryInC
       RUBY
-      Dir.chdir(bundled_app) do
-        ruby testrb, :no_lib => true
-      end
+      ruby testrb, :no_lib => true
 
-      expect(out).to eq('very_simple_binary')
+      expect(out).to eq("VerySimpleBinaryInC")
     end
   end
 
--- a/lib/bundler/rubygems_ext.rb
+++ b/lib/bundler/rubygems_ext.rb
@@ -115,6 +115,24 @@ module Gem
       dependencies - development_dependencies
     end
 
+    # This method doesn't exist before RubyGems 1.8.
+    unless method_defined?(:activate)
+      def activate
+        raise_if_conflicts
+
+        return false if Gem.loaded_specs[self.name]
+
+        activate_dependencies
+        add_self_to_load_path
+
+        Gem.loaded_specs[self.name] = self
+        @activated = true
+        @loaded = true
+
+        return true
+      end
+    end
+
   private
 
     def dependencies_to_gemfile(dependencies, group = nil)


segiddins:
Please add a test for this? Thanks!

colby-swandale:
Thanks for opening this PR and helping make Bundler better. On top of the missing spec Rubocop has also detected a few linting problems, you can run it locally via `bin/rubocop`  👍

joshuaspence:
Thanks for the feedback, I'll fix the linting issues and add tests now.

joshuaspence:
Now with tests!

lib/bundler/installer/standalone.rb
```diff
@@ -15,24 +15,31 @@ def generate
...
-          file.puts %($:.unshift "\#{path}/#{path}")
-        end
-      end
-    end
 
-  private
+        @specs.each do |spec|
+          next if spec.name == "bundler"
+
+          # Ensure that all `require_paths` exist.
```
joshuaspence: I'm not sure if this code is important, but a test fails without it so I kept it.

spec/install/gems/standalone_spec.rb
```diff
@@ -67,21 +83,6 @@
     include_examples "common functionality"
   end
 
-  describe "with gems with native extension" do
```
joshuaspence: This test didn't seem relevant anymore.

lib/bundler/installer/standalone.rb
```diff
@@ -15,24 +15,31 @@ def generate
...
-        next if spec.name == "bundler"
-        Array(spec.require_paths).map do |path|
-          gem_path(path, spec).sub(version_dir, '#{ruby_engine}/#{ruby_version}')
           # This is a static string intentionally. It's interpolated at a later time.
+          replacement = '#{ruby_engine}/#{ruby_version}'
+
+          gemspec = gem_path(spec.loaded_from, spec).sub(version_dir, replacement)
+          full_gem_path = gem_path(spec.full_gem_path, spec).sub(version_dir, replacement)
+
+          file.puts "Gem::Specification.load('\#{path}/#{gemspec}').instance_eval do |spec|)"
```
colby-swandale: don't need `)` at the end

joshuaspence:
Urgh, it worked when I tested it locally but I broke a few things when tidying it up for review. Tests should be green now.

spec/install/gems/standalone_spec.rb
```diff
@@ -67,21 +83,6 @@
...
-    before do
-      install_gemfile <<-G, :standalone => true
-        source "file://#{gem_repo1}"
-        gem "very_simple_binary"
-      G
-    end
-
-    it "generates a bundle/bundler/setup.rb with the proper paths", :rubygems => "2.4" do
```
segiddins: Instead of testing the path, can we change this test to make sure that after running `setup.rb` we can still require the extension?

spec/install/gems/standalone_spec.rb
```diff
@@ -75,10 +91,19 @@
...
+        require "bundler/setup"
+
+        require 'very_simple_binary'
+        puts (Gem.loaded_specs['very_simple_binary'] || Gem::Specification.new).name
+      RUBY
+      Dir.chdir(bundled_app) do
+        ruby testrb, :no_lib => true
+      end
+
+      expect(out).to eq('very_simple_binary')
```
segiddins: double quotes

spec/install/gems/standalone_spec.rb
```diff
@@ -75,10 +91,19 @@
...
-      extension_line = File.read(bundled_app("bundle/bundler/setup.rb")).each_line.find {|line| line.include? "/extensions/" }.strip
-      expect(extension_line).to start_with '$:.unshift "#{path}/../#{ruby_engine}/#{ruby_version}/extensions/'
-      expect(extension_line).to end_with '/very_simple_binary-1.0"'
+    it "loads gems correctly", :rubygems => "2.4" do
+      testrb = String.new <<-RUBY
+        $:.unshift File.expand_path("bundle")
+        require "bundler/setup"
+
+        require 'very_simple_binary'
+        puts (Gem.loaded_specs['very_simple_binary'] || Gem::Specification.new).name
```
segiddins: please `puts VerySimpleBinaryInC` and verify the output, which will make sure that the correct file has been loaded

spec/install/gems/standalone_spec.rb
```diff
@@ -29,6 +29,22 @@
...
+    it "loads gems correctly" do
+      testrb = String.new <<-RUBY
+        $:.unshift File.expand_path("bundle")
+        require "bundler/setup"
+
+      RUBY
+      expected_gems.each do |k, _|
+        testrb << "\nputs (Gem.loaded_specs['#{k}'] || Gem::Specification.new).name"
+      end
+      Dir.chdir(bundled_app) do
```
segiddins: no need to chdir

lib/bundler/installer/standalone.rb
```diff
@@ -15,24 +15,31 @@ def generate
...
-        next if spec.name == "bundler"
-        Array(spec.require_paths).map do |path|
-          gem_path(path, spec).sub(version_dir, '#{ruby_engine}/#{ruby_version}')
           # This is a static string intentionally. It's interpolated at a later time.
+          replacement = '#{ruby_engine}/#{ruby_version}'
+
+          gemspec = gem_path(spec.loaded_from, spec).sub(version_dir, replacement)
+          full_gem_path = gem_path(spec.full_gem_path, spec).sub(version_dir, replacement)
+
+          file.puts %(Gem::Specification.load("\#{path}/#{gemspec}").instance_eval do |spec|)
```
segiddins: could this be `tap` instead of `instance_eval`?

spec/install/gems/standalone_spec.rb
joshuaspence: I tried that originally, but it fails even on `master`

segiddins: in that case, I'm worried this isn't working anyways?

joshuaspence: Possibly. Maybe that should be investigated and fixed separately?

segiddins:
```diff
diff --git a/spec/install/gems/standalone_spec.rb b/spec/install/gems/standalone_spec.rb
index eb9d6e1c7..1babf7d38 100644
--- a/spec/install/gems/standalone_spec.rb
+++ b/spec/install/gems/standalone_spec.rb
@@ -92,16 +92,19 @@ RSpec.shared_examples "bundle install --standalone" do
     end
 
     it "loads gems correctly", :rubygems => "2.4" do
-      testrb = String.new <<-RUBY
-        $:.unshift File.expand_path("bundle")
+      testrb = <<-RUBY
+        $:.unshift "#{bundled_app("bundle")}"
         require "bundler/setup"
 
         require 'very_simple_binary'
         puts (Gem.loaded_specs['very_simple_binary'] || Gem::Specification.new).name
+
+        require 'very_simple_binary_c'
+        puts VerySimpleBinaryInC
       RUBY
       ruby testrb, :no_lib => true
 
-      expect(out).to eq("very_simple_binary")
+      expect(out).to eq("very_simple_binary\nVerySimpleBinaryInC")
     end
   end
 
```

joshuaspence:
Thanks, that seemed to work

lib/bundler/installer/standalone.rb
```diff
@@ -15,24 +15,30 @@ def generate
...
-        Array(spec.require_paths).map do |path|
-          gem_path(path, spec).sub(version_dir, '#{ruby_engine}/#{ruby_version}')
           # This is a static string intentionally. It's interpolated at a later time.
+          replacement = '#{ruby_engine}/#{ruby_version}'
+
+          gemspec = gem_path(spec.loaded_from, spec).sub(version_dir, replacement)
+          full_gem_path = gem_path(spec.full_gem_path, spec).sub(version_dir, replacement)
+
+          file.puts %(Gem::Specification.load("\#{path}/#{gemspec}").tap do |spec|)
+          file.puts %(  spec.full_gem_path = "\#{path}/#{full_gem_path}")
```
segiddins: old versions of rubygems don't seem to have this setter

joshuaspence: Thoughts on how to handle this?

segiddins:
`Gem::Specification#activate` also doesn't always seem to be present :O 

https://travis-ci.org/bundler/bundler/builds/224607732

indirect:
Hey, so I'm not 100% sure about making this change in general. The point of standalone mode is to set up the load path and allow you to require your gems without even loading RubyGems or Bundler at all. Activating the gems as gems kind of defeats the point of standalone mode... at that point, why not use `bundle exec`?

Can you tell us more about what you're trying to do, and why you feel like this change is helpful for you?

joshuaspence:
@indirect sure, let me elaborate (although, I thought that you had already stated that you would accept a pull request similar to this in https://github.com/bundler/bundler/issues/2720#issuecomment-31313879).

So we are currently using Puppet 3 and we install it using Bundler rather than using a system package. The main reason that we do this is because we want to use a bunch of gems in some custom Puppet code that we have written, and it makes our bootstrapping process much simpler if we simply install all of our dependencies using Bundler.

To run Puppet, we use standalone binstubs (i.e. `bundle install --standalone --binstubs && ./bin/puppet apply ...`). In _most_ cases, we can simply use `bundle exec puppet apply` instead of using the standalone binstubs, although this is not always the case. For example, we are using the [sbadia/gitlab](https://forge.puppet.com/sbadia/gitlab) module for installing Gitlab. Since Gitlab itself is a Ruby application, it also uses Bundler:

```puppet
  exec { 'install gitlab':
    command => "bundle install${gitlab_bundler_jobs_flag} --without development aws test ${gitlab_without_gems} ${gitlab_bundler_flags}",
    cwd     => "${git_home}/gitlab",
    unless  => 'bundle check',
    timeout => 0,
    require => [
      Gitlab::Config::Database['gitlab'],
      Gitlab::Config::Unicorn['gitlab'],
      File["${git_home}/gitlab/config/gitlab.yml"],
      Gitlab::Config::Resque['gitlab'],
    ],
    notify  => Exec['run migrations'],
  }
```
This part of the Puppet manifests fails when run through `bundle exec puppet apply` due to the presence of the `BUNDLE_*` environment variables, which presumably alter the behavior of `bundle install` when run from `/home/git/gitlab`.

Our solution here was to use standalone binstubs, but we have run into an issue when trying to upgrade to Puppet 4. Specifically, the standalone binstub for Puppet 4 no longer seems to execute correctly:

```
/etc/puppet/vendor/bundle/ruby/2.1.0/gems/puppet-4.10.0/lib/puppet.rb:76:in `singleton class': undefined method `version' for nil:NilClass (NoMethodError)
    from /etc/puppet/vendor/bundle/ruby/2.1.0/gems/puppet-4.10.0/lib/puppet.rb:55:in `<module:Puppet>'
    from /etc/puppet/vendor/bundle/ruby/2.1.0/gems/puppet-4.10.0/lib/puppet.rb:49:in `<top (required)>'
    from /usr/local/rbenv/versions/2.1.5/lib/ruby/site_ruby/2.1.0/rubygems/core_ext/kernel_require.rb:55:in `require'
    from /usr/local/rbenv/versions/2.1.5/lib/ruby/site_ruby/2.1.0/rubygems/core_ext/kernel_require.rb:55:in `require'
    from /etc/puppet/vendor/bundle/ruby/2.1.0/gems/puppet-4.10.0/lib/puppet/util/command_line.rb:12:in `<top (required)>'
    from /usr/local/rbenv/versions/2.1.5/lib/ruby/site_ruby/2.1.0/rubygems/core_ext/kernel_require.rb:55:in `require'
    from /usr/local/rbenv/versions/2.1.5/lib/ruby/site_ruby/2.1.0/rubygems/core_ext/kernel_require.rb:55:in `require'
    from /etc/puppet/vendor/bundle/ruby/2.1.0/gems/puppet-4.10.0/bin/puppet:4:in `<top (required)>'
    from /etc/puppet/bin/puppet:14:in `load'
    from /etc/puppet/bin/puppet:14:in `<main>'
```

From inspection of the Puppet source code, they are simply checking `Gem.loaded_specs`.

joshuaspence:
So my understanding is that the difference between `bundle install --binstubs` and `bundle install --standalone --binstubs` is that the latter avoids dependencies on both Bundler _and_ RubyGems? I guess that, in my case, really I just want to avoid the dependency on Bundler.

indirect:
@joshuaspence as I stated in the PR that you linked to:

> Standalone mode was never intended to duplicate the regular setup.rb, and in fact exists specifically for times when people don't want to use rubygems or Bundler inside their app, just call "require" and have their load path already set up.

Standalone mode is specifically designed for users who want to use only $LOAD_PATH + `require` at runtime, without RubyGems or Bundler being loaded or used in any way.

Bundler already contains a way to use `bundle exec` and still successfully run other Bundled applications: ``bundle exec ruby -e 'Bundler.with_clean_env { `bundle exec gitlab` }'`` should do exactly what you want. Please see the docs for `with_clean_env`: http://bundler.io/v1.3/man/bundle-exec.1.html#Shelling-out

joshuaspence:
So what did you mean by this?

> That said, now that Rubygems ships inside Ruby, it may not be reasonable to expect people to avoid using Rubygems. Let's do the patch that makes it work.

indirect:
@joshuaspence four years ago, I said that it _might_ make sense to change standalone mode to ensure that the `gem` method does not raise an error, either by configuring Gem::Specification.dir or by defining the `gem` method to be a no-op.

I appreciate your willingness to investigate your problem and come up with a solution. Thank you for sending the PR. The particular problem you are describing (nested bundle invocations) already has a solution that does not require making any changes to Bundler, so I recommend that you use that solution because it both already exists and has been present in every version of Bundler for at least the last five years.

At the meta level, your tone is coming across as pretty combative and entitled right now. Expecting that I will merge this PR because of a comment I made four years ago on a feature request that didn't include any code is extremely shitty and demanding. Please tone it down before you start violating the code of conduct. Thanks.

joshuaspence:
I apologize if my tone came across as combative, that was not intended. I was simply trying to clarify whether there was a different solution that you were referring to in your previous comment or if your previous comment was simply no longer relevant in the current state.

Of course, I don't expect a PR to be merged unless the maintainers feel that the feature is worthwhile, given that they are the ones who would be required to support the code after it has been merged.

joshuaspence:
Is there a way to use `Bundler.with_clean_env` directly from the command line? Something like `bundle exec --with-clean-env puppet apply`.

indirect:
@joshuaspence unfortunately, that won't get you what you want. You need to run Puppet inside the context of your bundle, and then you need to run Gitlab in the context of its own Bundle. Inside Puppet, you'll need to either execute Ruby code directly (maybe subclassing the resource you're using to install Gitlab?) or you'll need to tell Puppet to execute a Ruby script whose contents are `Bundler.with_clean_env { system(*ARGV) }`, or something along those lines.

tl;dr, Bundler can't clean the env for you from the command line because the cleaning needs to be invoked from inside your running Ruby process as it shells out (in this case Puppet running a command).

joshuaspence:
OK, that makes sense. Thanks for your help and feedback. I think that I can come up with a solution on my end to workaround this.

indirect:
👍

PR #5613, opened 2017-04-25
Vendor fileutils since it is now a gem

segiddins:


segiddins:
If there's still interest in this change, @indirect, it should be green now

segiddins:
This is failing on older rubies because they don't have a capacity argument initializer for strings. @indirect maybe we make the vendored fileutils 2.4+ ?

indirect:
only loading the vendored one on 2.3 or 2.4+ seems reasonable to me. maybe add a todo to make it always load after we drop 1.x supported rubies.

segiddins:
No, only 2.4 has that string initializer (it's the only build that's passing on CI)

indirect:
2.4 it is then

segiddins:
Changed to be 2.4 + only

segiddins:
🍏

indirect:
@bundlerbot r+

eregon:
I am curious, why is it necessary to vendor fileutils?
It is still part of the standard library, it's just additionally a default gem.

indirect:
If you think through how Bundler works, I think you'll probably see why this is a problem. Now that `fileutils` is a gem, any user can create a Gemfile that contains `gem "fileutils", "2.0.0"`.

1. What if Bundler only works with `fileutils 1.0.0`? Bundler can't function with `fileutils 2.0.0`, and must exit with an error instead.
1. Bundler has to `require "fileutils"` to work correctly. What happens if the default gem gets loaded, but the Gemfile requires a different version? It's too late. The Gemfile cannot be satisfied, and Bundler must exit with an error.

The workaround for these problems is that we vendor our own copy of FileUtils, with a known version, and a different Ruby namespace.

eregon:
@indirect  Thank you for the detailed explanation.
Indeed, I see the issue now.
It's probably feasible to force requiring the stdlib version of `fileutils`, but that would still conflict (in namespace/methods) if another version of `fileutils` must be loaded in the same Ruby process.

eregon:
One idea how to improve on this long-term: maybe we could have a variant of Kernel#load where we can pass a top-level wrapper module, so then when loading the stdlib `fileutils.rb`, all modules/classes/constants would get defined under the provided wrapper module.
Of course, `fileutils.rb` might do thing like `Object.const_set` or so to break out of the wrapping namespace, but this could nevertheless probably work for a number of standard libraries which behave well (or can be changed to).

PR #5614, opened 2017-04-25
Test that default gem substitutes are properly activated

segiddins:
This is currently failing on ruby trunk


indirect:
Nice. @bundlerbot r+

PR #5619, opened 2017-04-26
[Documentation] Pristine documentation

denniss:


man/bundle-pristine.ronn
```diff
@@ -0,0 +1,13 @@
+bundle-pristine(1) -- Restores installed gems to its pristine condition
```
colby-swandale: i would prefer `to their pristine condition`

man/bundle-pristine.ronn
```diff
@@ -0,0 +1,13 @@
...
+===========================================================================
+
+## SYNOPSIS
+
+`bundle pristine`
+
+## DESCRIPTION
+
+Restores installed gems to its pristine condition.
```
colby-swandale: same as above regarding `their`

colby-swandale: It feels weird reading 4 statements like this out loud. We can blend some these sentences together?  How does something like:

```
`pristine` restores the installed gems in the bundle to their pristine condition using the local gem
cache from RubyGems. For git gems - a forced checkout will be performed.

Note the Bundler gem cannot be restored to its original state with `pristine`. 

```

sound?

denniss: Much better.

colby-swandale:
thank you! @bundlerbot r+

PR #5624, opened 2017-04-28
[Git] Allow specifying non-branch symbolic refs

segiddins:
> The intention here is to allow installing a gem from git using a
> ref of the form "refs/zuul/" (which are used by the zuul
> continuous-integration system). There are likely other use cases for
> arbitrary refs.

Closes #4845


indirect:
👍 @bundlerbot r+

segiddins:
The reason this PR is mergeable where #4845 was not is that it avoids fetching _all_ symbolic refs, and the only non-tag/head refs fetched are those specifically requested by a gemfile

PR #5628, opened 2017-04-30
Permit absolute paths in cache_path setting

### LINKED ISSUE #5627: Remove relative path enforcement on --cache-path
Would it be possible to remove the check done on `--cache-path` [here](https://github.com/bundler/bundler/blob/28e67dc7e4b9e3a6c67524597f7cd64e12ddbf6d/lib/bundler/settings.rb#L218)?

Used in conjunction with `--no-prune` it would then be possible to more easily maintain a central cache of `gem` files useable by multiple local apps. In the case where apps have many common libraries (that are often, if not always, in sync) this is quite useful.

mal:
Removes check on leading `/` from `cache_path` and updates tests.

Fixes #5627

### DIFF
--- a/lib/bundler/settings.rb
+++ b/lib/bundler/settings.rb
@@ -213,11 +213,7 @@ module Bundler
     end
 
     def app_cache_path
-      @app_cache_path ||= begin
-        path = self[:cache_path] || "vendor/cache"
-        raise InvalidOption, "Cache path must be relative to the bundle path" if path.start_with?("/")
-        path
-      end
+      @app_cache_path ||= self[:cache_path] || "vendor/cache"
     end
 
   private
--- a/spec/cache/cache_path_spec.rb
+++ b/spec/cache/cache_path_spec.rb
@@ -23,12 +23,4 @@ RSpec.describe "bundle package" do
       expect(bundled_app("vendor/cache-foo/rack-1.0.0.gem")).to exist
     end
   end
-
-  context "when given an absolute path" do
-    it "exits with non-zero status" do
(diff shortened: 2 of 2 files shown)


spec/cache/cache_path_spec.rb
```diff
@@ -23,12 +23,4 @@
       expect(bundled_app("vendor/cache-foo/rack-1.0.0.gem")).to exist
     end
   end
-
-  context "when given an absolute path" do
-    it "exits with non-zero status" do
```
segiddins: can we change this to a test that it packages correctly?

lib/bundler/settings.rb
```diff
@@ -214,9 +214,7 @@ def ignore_config?
 
     def app_cache_path
       @app_cache_path ||= begin
-        path = self[:cache_path] || "vendor/cache"
-        raise InvalidOption, "Cache path must be relative to the bundle path" if path.start_with?("/")
-        path
+        self[:cache_path] || "vendor/cache"
```
colby-swandale: The `begin` block is no longer necessary here
```
DEFAULT_VENDOR_CACHE_PATH = "vendor/cache"
...

def app_cache_path
  @app_cache_path ||= self[:cache_path] || DEFAULT_VENDOR_CACHE_PATH
end
```

mal: I've held off extracting `"vendor/cache"` into a constant for now because it didn't appear to fit with the rest of the file. It makes sense to do but perhaps as part of a separate wider scoped refactor.

segiddins:
👍🏻

segiddins:
@bundlerbot r+

PR #5634, opened 2017-05-03
Specify `--require spec_helper` in .rspec

koic:
Specifying `--require spec_helper` in .rspec will automatically require spec_helper in *_spec.rb.
It isn't necessary to specify `require "spec_helper"` in individual *_spec.rb. I think that it's a [DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself) way.

Refer: https://github.com/rspec/rspec/wiki#rspec

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,3 +1,4 @@
 --format documentation
 --color
 --warnings
+--require spec_helper
--- a/spec/bundler/bundler_spec.rb
+++ b/spec/bundler/bundler_spec.rb
@@ -1,6 +1,5 @@
 # encoding: utf-8
 # frozen_string_literal: true
-require "spec_helper"
 require "bundler"
 
 RSpec.describe Bundler do
--- a/spec/bundler/cli_spec.rb
+++ b/spec/bundler/cli_spec.rb
@@ -1,5 +1,4 @@
 # frozen_string_literal: true
-require "spec_helper"
 require "bundler/cli"
 
 RSpec.describe "bundle executable" do
--- a/spec/bundler/compact_index_client/updater_spec.rb
+++ b/spec/bundler/compact_index_client/updater_spec.rb
@@ -1,5 +1,4 @@
 # frozen_string_literal: true
-require "spec_helper"
 require "net/http"
 require "bundler/compact_index_client"
 require "bundler/compact_index_client/updater"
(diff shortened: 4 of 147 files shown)


colby-swandale:
Firstly, thanks for helping making Bundler better. Can you explain in more detail why we should merge this? The reason why we have have `require "spec_helper` at the start of every file is so that we load the rspec helpers when we run an individual test ie: `rspec spec/bundler_spec.rb`. I'm not particularly confident that we want to depend on the `.rspec` file for rspec to work correctly.

koic:
Thank you for your reply and maintenance.

As far as I know,  the following behaviors are the same.

- Write `require "spec_helper"` at the start of every spec file
- Write `--require spec_helper` in the .rspec file

In other words, spec_helper is automatically required also in the execution of individual spec file (e.g. `rspec spec/bundler/bundler_spec.rb`) .

```sh
#
# 1. Specify --require spec_helper in .rspec
#
% cat .rspec
--format documentation
--color
--warnings
--require spec_helper
# 
# 2. Do not specify `require "spec_helper"` in a spec file
#
% head spec/bundler/bundler_spec.rb
# encoding: utf-8
# frozen_string_literal: true
require "bundler"

RSpec.describe Bundler do
  describe "#load_gemspec_uncached" do
    let(:app_gemspec_path) { tmp("test.gemspec") }
    subject { Bundler.load_gemspec_uncached(app_gemspec_path) }

    context "with incorrect YAML file" do
#
# 3. Automatically require spec_helper and execute it
#
% rspec spec/bundler/bundler_spec.rb
(snip)
12 examples, 0 failures
```

I think that depend on `.rspec` when using the RSpec framework is not bad.

Of course, this is concluded because all spec files depend on `spec_helper`.

If all spec files do not depend on `spec_helper`, we can not depend on `--require spec_helper` in `.rspec`. For example, a situation where part of spec files depend on `custom_helper` instead of `spec_helper`. In that case, we'd specify `spec_helper` or `custom_helper` at the start of every file.

For this time, spec files depend on `spec_helper` only, so I think that it is better to specify `--require spec_helper` to `.rspec`.

segiddins:
@indirect r?

indirect:
I also prefer this style, thank you @koic 👍 @bundlerbot r+

koic:
Thank you, too!

PR #5635, opened 2017-05-03
Tweak a newgem template for RSpec

koic:
`--require spec_helper` is contained in .rspec by default when running `rspec --init` on RSpec 3.

```sh
% rspec --version
3.5.4
% rspec --init
  create   .rspec
  create   spec/spec_helper.rb
% cat .rspec
--color
--require spec_helper
```

It seems that the code of template premise the RSpec 3, so I think that it was reasonable to adjust to it.

Related PR #5634


indirect:
Good catch, thank you! @bundlerbot r+

PR #5638, opened 2017-05-03
Warn when making an outdated TLS connection to rubygems.org

segiddins:
Right now, output looks like

```
Your Ruby version does not support TLSv1.1 or newer, which will be required to connect to https://index.rubygems.org by June 2018.
Resolving dependencies...
Using rake 11.3.0
Using automatiek 0.1.2 from source at `.`
Using rspec-support 3.5.0
Using diff-lcs 1.3
Using bundler 1.15.0.pre.3
Using rspec-core 3.5.4
Using rspec-expectations 3.5.0
Using rspec-mocks 3.5.0
Using rspec 3.5.0
Bundle updated!
```

Obviously this requires tests, and also backing out forcing the TLS version (which I did to make testing easier). The message could also use improvement, but I figured I'd ask for input from @indirect and @dwradcliffe

### DIFF
--- a/lib/bundler/vendored_persistent.rb
+++ b/lib/bundler/vendored_persistent.rb
@@ -37,10 +37,13 @@ module Bundler
       case ssl_version
       when /TLSv([\d\.]+)/
         version = Gem::Version.new($1)
-        if version < Gem::Version.new("1.1")
-          Bundler.ui.warn "Your Ruby version does not support TLSv1.1 or newer" \
-            ", which will be required to connect to https://#{uri.hostname}" \
-            " by January 2018."
+        if version < Gem::Version.new("1.2")
+          Bundler.ui.warn \
+            "Warning: Your Ruby version is compiled against a copy of OpenSSL that is very old. " \
+            "Starting in January 2018, RubyGems.org will refuse connection requests from these " \
+            "very old versions of OpenSSL. If you will need to continue installing gems after " \
+            "January 2018, please follow this guide to upgrade: http://ruby.to/tls-outdated.",
+            :wrap => true
         end
       end
     end
--- a/lib/bundler/vendored_persistent.rb
+++ b/lib/bundler/vendored_persistent.rb
@@ -27,7 +27,7 @@ module Bundler
 
     def warn_old_tls_version_rubygems_connection(uri, connection)
       return unless connection.use_ssl?
-      return unless (uri.hostname || "").end_with?("rubygems.org")
+      return unless (uri.host || "").end_with?("rubygems.org")
 
       socket = connection.instance_variable_get(:@socket)
       socket_io = socket.io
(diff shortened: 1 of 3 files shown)


colby-swandale:
I would like to see us to maybe write a rubygems.org guide, or bundler.io documentation that we can link to users with information about what is happening, steps to take and anything else. Then we attach it to the end of this warning.

lib/bundler/vendored_persistent.rb
```diff
@@ -15,3 +15,31 @@ module HTTP
...
+      socket = connection.instance_variable_get(:@socket)
+      ssl_version = socket.io.ssl_version
+
+      case ssl_version
+      when /TLSv([\d\.]+)/
+        version = Gem::Version.new($1)
+        if version < Gem::Version.new("1.1")
+          Bundler.ui.warn "Your Ruby version does not support TLSv1.1 or newer" \
+            ", which will be required to connect to https://#{uri.hostname}" \
+            " by June 2018."
```
dwradcliffe: I'd like to say January 2018. That gives us a buffer and we can see who still can't upgrade.

indirect:
This is great, thanks for the proactivity here. I think this PR has two blockers right now: first, I'm really worried about shipping without a test to make sure that this warning prints when it is supposed to and does not print when it shouldn't. We had an OpenSSL warning in the past that was similarly hard to test that wound up silently breaking and we didn't find out for months. :/

Second, and this is part of the general UX philosophy of Bundler, warnings should come with actionable instructions to resolve them. I think in this case we'll need to write a guide to upgrading OpenSSL for Ruby. [Mislav's blog post and tools](https://mislav.net/2013/07/ruby-openssl/) are the most comprehensive existing documentation that I know of for SSL issues, so we'll probably want to lean heavily on that as we write a guide for users who need to understand 1) what TLS 1.1 is, 2) why they can't use it, and 3) how to upgrade OpenSSL and/or recompile Ruby so that they can.

@rubymorillo, I think that guide would make a really good bite-sized thing to start with, maybe aiming to have it done sometime in the next 1-2 weeks?

segiddins:
Can you think of any way to test https connections using sinatra, and forcing the SSL version thereof?

indirect:
I'd be okay with a unit test that monkeypatches net/http, if it's doable in a way that can be un-monkey-patched after we test it.

segiddins:
Specs added, can update the message with whatever you want

segiddins:
Ping on this? I think all I was waiting on was an André approved error message? I'd like us to get this out in 1.16

indirect:
How about:

> Warning: Your Ruby version is compiled against a copy of OpenSSL that is very old. Starting in January 2018, RubyGems.org will refuse connection requests from these very old versions of OpenSSL. If you will need to continue installing gems after January 2018, please follow this guide to upgrade: https://bundler.io/guides/rubygems_ssl_troubleshooting.html

(The guide doesn't exist yet, but I'm planning to take care of that in the next 1-2 days)

indirect:
@segiddins updated with final link to guide, which now exists (albeit in an alpha version for now).

segiddins:
@indirect updated! r?

indirect:
Excellent 👍🏻
@bundlebot r+

indirect:
ugh, I can't type. @bundlerbot r+

PR #5639, opened 2017-05-03
Unify the notation of RubyGems

koic:
This PR is mainly replacing Rubygems with RubyGems. 

`Rubygems` is a Ruby module name (e.g. `Bundler::Source::Rubygems`) , whereas RubyGems is the proper notation as below.

https://github.com/rubygems/rubygems/blob/master/README.md

In this PR, it changed the notation mainly used in console output and documentation  (and some source code comments) .

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -122,14 +122,14 @@ begin
       system "sudo rm -rf #{File.expand_path("../tmp/sudo_gem_home", __FILE__)}"
     end
 
-    # Rubygems specs by version
+    # RubyGems specs by version
     namespace :rubygems do
       rubyopt = ENV["RUBYOPT"]
       # When editing this list, also edit .travis.yml!
       branches = %w(master)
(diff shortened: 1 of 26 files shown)


lib/bundler/cli/pristine.rb
```diff
@@ -5,7 +5,7 @@ module Bundler
   class CLI::Pristine
     def run
       Bundler.load.specs.each do |spec|
-        next if spec.name == "bundler" # Source::Rubygems doesn't install bundler
+        next if spec.name == "bundler" # Source::RubyGems doesn't install bundler
```
segiddins: this shouldn't be changed

koic: Thank you for review. I fixed it.

segiddins:
@bundlerbot r+

koic:
@segiddins Thanks always.

segiddins:
Thank you for these great contributions!

PR #5647, opened 2017-05-09
Add specs for #find_by_name_and_platform and #merge in SpecSet

colby-swandale:
As the title says, this PR just adds a few specs that i noticed were missing in SpecSet.


segiddins:
@bundlerbot r+

PR #5653, opened 2017-05-10
tell_err is private and the last caller was already removed

jrafanie:
tell_err was added in e120f40f72fe53f55e05db9204f98d28f1065e81
It's last caller was removed in d88627a8191644fb970cce2140f3c320f3101c32 as part of a
removal of old deprecation code.

I don't know if this was intentional, but it seems like a good idea to remove this private method.

@segiddins please review.  I was going to use this method since I wanted a way to write to stderr instead of stdout but realized I couldn't use it since it was private.

### DIFF
--- a/lib/bundler/ui/shell.rb
+++ b/lib/bundler/ui/shell.rb
@@ -102,14 +102,6 @@ module Bundler
         end
       end
 
-      def tell_err(message, color = nil, newline = nil)
-        buffer = @shell.send(:prepare_message, message, *color)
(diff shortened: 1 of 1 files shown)


segiddins:
I believe it's added back and used in the 2-0-dev branch

jrafanie:
ok @segiddins, I don't see `tell_err`, only `tell_stderr` on that branch.  I can close this if it's not useful.

segiddins:
I appreciate the PR, but I think we'll wait until we get 2-0-dev onto master to clean this up. Thanks for taking the time, though!

PR #5658, opened 2017-05-12
Allow to add username and password to a remote during a deployment

adrian-gomez:
This is my first PR to bundler, so if something is out of place let me know and I'll do my best to correct it.

The issue I'm trying to fix is allowing to change a remote on a frozen bundler for the same remote adding username and password.
In our dev boxes we have configured bundle to store the username and password for a given domain, but on our qa boxes we don't want to do that for various reasons.

As far as I can tell this should not have any negative side effects as we are keeping the same remote but only adding some auth to it

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -407,8 +407,8 @@ module Bundler
 
       # Check if it is possible that the source is only changed thing
       if (new_deps.empty? && deleted_deps.empty?) && (!new_sources.empty? && !deleted_sources.empty?)
-        new_sources.reject! {|source| source.is_a_path? && source.path.exist? }
-        deleted_sources.reject! {|source| source.is_a_path? && source.path.exist? }
+        new_sources.reject! {|source| (source.path? && source.path.exist?) || equivalent_rubygems_remotes?(source) }
+        deleted_sources.reject! {|source| (source.path? && source.path.exist?) || equivalent_rubygems_remotes?(source) }
       end
 
       if @locked_sources != gemfile_sources
@@ -636,7 +636,7 @@ module Bundler
       if !locked_gem_sources.empty? && !actual_remotes.empty?
         locked_gem_sources.each do |locked_gem|
           # Merge the remotes from the Gemfile into the Gemfile.lock
-          changes |= locked_gem.replace_remotes(actual_remotes)
+          changes |= locked_gem.replace_remotes(actual_remotes, Bundler.settings[:allow_deployment_source_credential_changes])
         end
       end
 
@@ -963,6 +963,12 @@ module Bundler
         requirements[name] = DepProxy.new(dep, locked_spec.platform)
         requirements
       end.values
+    end
+
+    def equivalent_rubygems_remotes?(source)
+      return false unless source.is_a?(Source::Rubygems)
+
+      Bundler.settings[:allow_deployment_source_credential_changes] && source.equivalent_remotes?(sources.rubygems_remotes)
     end
   end
 end
--- a/lib/bundler/settings.rb
+++ b/lib/bundler/settings.rb
@@ -10,6 +10,7 @@ module Bundler
 
     BOOL_KEYS = %w[
       allow_bundler_dependency_conflicts
+      allow_deployment_source_credential_changes
       allow_offline_install
       auto_install
       cache_all
--- a/man/bundle-config.ronn
+++ b/man/bundle-config.ronn
@@ -121,6 +121,9 @@ learn more about their operation in [bundle install(1)][bundle-install].
 * `allow_bundler_dependency_conflicts` (`BUNDLE_ALLOW_BUNDLER_DEPENDENCY_CONFLICTS`):
    Allow resolving to specifications that have dependencies on `bundler` that
    are incompatible with the running Bundler version.
+* `allow_deployment_source_credential_changes` (`BUNDLE_ALLOW_DEPLOYMENT_SOURCE_CREDENTIAL_CHANGES`):
+   When in deployment mode, allow changing the credentials to a gem's source.
+   Ex: `https://some.host.com/gems/path/` -> `https://user_name:password@some.host.com/gems/path`
 * `allow_offline_install` (`BUNDLE_ALLOW_OFFLINE_INSTALL`):
    Allow Bundler to use cached data when installing without network access.
 * `auto_install` (`BUNDLE_AUTO_INSTALL`):
--- a/spec/install/deploy_spec.rb
+++ b/spec/install/deploy_spec.rb
@@ -295,6 +295,82 @@ RSpec.describe "install with --deployment or --frozen" do
       expect(out).not_to include("You have deleted from the Gemfile")
     end
 
+    context "when replacing a host with the same host with credentials" do
+      let(:success_message) do
+        if Bundler::VERSION.split(".", 2).first == "1"
+          "Could not reach host localgemserver.test"
+        else
+          "Bundle complete!"
+        end
+      end
+
+      before do
+        install_gemfile <<-G
+        source "http://user_name:password@localgemserver.test/"
+        gem "rack"
+        G
+
+        lockfile <<-G
+        GEM
+          remote: http://localgemserver.test/
+          specs:
+            rack (1.0.0)
+
+        PLATFORMS
+          #{local}
+
+        DEPENDENCIES
+          rack
+        G
+      end
+
+      it "prevents the replace by default" do
+        bundle :install, forgotten_command_line_options(:deployment => true)
+
+        expect(out).to match(/The list of sources changed/)
+      end
+
+      context "when allow_deployment_source_credential_changes is true" do
+        before { bundle! "config allow_deployment_source_credential_changes true" }
+
+        it "allows the replace" do
+          bundle :install, forgotten_command_line_options(:deployment => true)
+
+          expect(out).to match(/#{success_message}/)
+        end
+      end
+
+      context "when allow_deployment_source_credential_changes is false" do
+        before { bundle! "config allow_deployment_source_credential_changes false" }
+
+        it "prevents the replace" do
+          bundle :install, forgotten_command_line_options(:deployment => true)
+
+          expect(out).to match(/The list of sources changed/)
+        end
+      end
+
+      context "when BUNDLE_ALLOW_DEPLOYMENT_SOURCE_CREDENTIAL_CHANGES env var is true" do
+        before { ENV["BUNDLE_ALLOW_DEPLOYMENT_SOURCE_CREDENTIAL_CHANGES"] = "true" }
+
+        it "allows the replace" do
+          bundle :install, forgotten_command_line_options(:deployment => true)
+
+          expect(out).to match(/#{success_message}/)
+        end
+      end
+
+      context "when BUNDLE_ALLOW_DEPLOYMENT_SOURCE_CREDENTIAL_CHANGES env var is false" do
+        before { ENV["BUNDLE_ALLOW_DEPLOYMENT_SOURCE_CREDENTIAL_CHANGES"] = "false" }
+
+        it "prevents the replace" do
+          bundle :install, forgotten_command_line_options(:deployment => true)
+
+          expect(out).to match(/The list of sources changed/)
+        end
+      end
+    end
+
     it "remembers that the bundle is frozen at runtime" do
       bundle! :lock
 
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -100,6 +100,9 @@ Style/TrailingCommaInArguments:
 Performance/FlatMap:
   Enabled: false
 
+Security/YAMLLoad:
+  Enabled: false
+
 # Metrics
 
 # We've chosen to use Rubocop only for style, and not for complexity or quality checks.
--- a/lib/bundler/source.rb
+++ b/lib/bundler/source.rb
@@ -46,6 +46,10 @@ module Bundler
       "#<#{self.class}:0x#{object_id} #{self}>"
     end
 
+    def path?
+      instance_of?(Bundler::Source::Path)
+    end
+
   private
 
     def version_color(spec_version, locked_spec_version)
--- a/lib/bundler/source/path.rb
+++ b/lib/bundler/source/path.rb
@@ -116,10 +116,6 @@ module Bundler
         Bundler.root
       end
 
-      def is_a_path?
-        instance_of?(Path)
-      end
-
       def expanded_original_path
         @expanded_original_path ||= expand(original_path)
       end
--- a/lib/bundler/source/rubygems.rb
+++ b/lib/bundler/source/rubygems.rb
@@ -219,13 +219,21 @@ module Bundler
         @remotes.unshift(uri) unless @remotes.include?(uri)
       end
 
-      def replace_remotes(other_remotes)
+      def equivalent_remotes?(other_remotes)
+        other_remotes.map(&method(:remove_auth)) == @remotes.map(&method(:remove_auth))
+      end
+
+      def replace_remotes(other_remotes, allow_equivalent = false)
         return false if other_remotes == @remotes
+
+        equivalent = allow_equivalent && equivalent_remotes?(other_remotes)
 
         @remotes = []
         other_remotes.reverse_each do |r|
           add_remote r.to_s
         end
+
+        !equivalent
       end
 
       def unmet_deps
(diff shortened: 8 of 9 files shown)


spec/bundler/definition_spec.rb
```diff
@@ -132,6 +132,32 @@
            #{Bundler::VERSION}
       G
     end
+
+    context "when replacing a host with the same host with credentials" do
```
segiddins: this file is used for unit tests, please move this into `deployment_spec.rb`?

lib/bundler/definition.rb
```diff
@@ -644,7 +644,8 @@ def converge_sources
       if !locked_gem_sources.empty? && !actual_remotes.empty?
         locked_gem_sources.each do |locked_gem|
           # Merge the remotes from the Gemfile into the Gemfile.lock
-          changes |= locked_gem.replace_remotes(actual_remotes)
+          changes |= !locked_gem.equivalent_remotes?(actual_remotes)
```
segiddins: any reason we can't fold this logic into the return value in `replace_remotes`?

adrian-gomez: My first approach was to do it there, but I'm not sure if it would have any other implications, for one it would mean that that the hosts are not actually replaced, so the host with the username and password would not be used. Maybe thats not important but it seemed pretty invasive.
If you think its ok to do that there I'm happy to move it

spec/bundler/definition_spec.rb
adrian-gomez: Will do

adrian-gomez: Would it be better on `commands/install_spec.rb` on a `deployment` context?
It seems its separated by commands and `deployment` seems like a modifier to `install` not a command on its own

lib/bundler/definition.rb
segiddins: I was thinking to replace the hosts but still return false if none of them have changed (besides the auth info)

spec/bundler/definition_spec.rb
segiddins: `spec/install/deploy_spec.rb`

adrian-gomez: go it  and done

lib/bundler/definition.rb
adrian-gomez: done

segiddins:
@indirect what do you think about this change in functionality?

indirect:
Unfortunately, this will break systems like GemFury that expose completely different gems depending on the credentials passed to the gem source. I understand not wanting to store credentials in your QA environment, but if the gem source works without credentials why do you need to store credentials in your other environments?

adrian-gomez:
@indirect This should not break anything as the previous behavior is kept as it was, this is just adding the option to change it. What broke stuff for us was adding this control, we have been stuck with `1.14.4` because of it.
The source does not work without credentials on our dev pc where are doing:
`bundle config <SOURCE> <USERNAME>:<API_KEY>`
thats what we don't want to do on qa or higher envs.

indirect:
@adrian-gomez how is this adding an option? this patch will mean that changing a source from one GemFury user account to another account, with completely different gems, will no longer register as a change in deployment mode. That's not okay. The only way this could work, I think, is if it were an additional config option to ignore the changed credentials.

adrian-gomez:
I was saying that this change wont break anything, if this was merged today and released no application will break by just updating bundler. Until 3 months ago something more forgiving than that was the default behaviour and things seemed to work ok for everybody.
I open, and willing to do the work, to make this configurable.

adrian-gomez:
@indirect @segiddins any more opinions on this? should I put it behind a configuration option?

indirect:
@adrian-gomez I think I'm okay with this if it's a configuration option to ignore changes to source credentials. @segiddins?

segiddins:
I really don't have a strong opinion either way, but I think a setting (`bundle config allow_deployment_source_credential_changes true`) would be reasonable?

segiddins:
I don't have a particularly strong opinion either way, but I think something along the lines of `bundle config allow_deployment_source_credential_changes true` would be fine?

segiddins:
I don't have a particularly strong opinion either way, but I think something along the lines of `bundle config allow_deployment_source_credential_changes true` would be fine?

adrian-gomez:
@segiddins @indirect I'm down for that and will work on it, thanks!

adrian-gomez:
@segiddins @indirect Hi guys sorry it took so long but I finally got time to finish the work on this PR, I added the config flag as suggested.
Let me know if I need to make more changes or if this is good as it is.

Thanks

adrian-gomez:
There is also a conflict between rubocop and the specs.
Rubocop is complaining about ` config = YAML.load(bundled_app(".bundle/config").read)` it wants me to use `YAML.safe_load` but when I tried the tests failed on some versions of ruby as it seems `#safe_load` was not available on those versions.

Not sure what to do about that

indirect:
@adrian-gomez you can adjust the rubocop configuration to disable that rule for now 👍

adrian-gomez:
@indirect done

indirect:
This seems reasonable to me. @segiddins r?

lib/bundler/cli.rb
```diff
@@ -197,6 +197,8 @@ def check
       "Exclude gems that are part of the specified named group."
     deprecated_option "with", :type => :array, :banner =>
       "Include gems that are part of the specified named group."
+    method_option "allow-deployment-source-credential-changes", :type => :boolean, :banner =>
+      "Allow to add credentials for gem sources during deployment."
```
segiddins: I think this should be a config option, and not a CLI option

lib/bundler/settings.rb
```diff
@@ -49,6 +49,7 @@ class Settings
       suppress_install_using_messages
       unlock_source_unlocks_spec
       update_requires_all_flag
+      allow_deployment_source_credential_changes
```
segiddins: please keep the array sorted

lib/bundler/cli.rb
adrian-gomez: done, left only as a config option

lib/bundler/settings.rb
adrian-gomez: done

adrian-gomez:
@indirect I'm having some issues with the specs on travis, I tried to reproduce locally but was unable to. Do you have any hints that I can try? Or is this expected and should just ignore it?

indirect:
@adrian-gomez the only failing spec on travis is 2.0-mode... I think @segiddins can give you a better idea of how to run that locally and how to tweak the tests to make it pass. 👍

spec/install/deploy_spec.rb
```diff
@@ -295,6 +295,106 @@
...
+
+        PLATFORMS
+          #{local}
+
+        DEPENDENCIES
+          rack
+      G
+
+      bundle! "config allow_deployment_source_credential_changes true"
+      bundle "install --deployment"
```
segiddins: this needs to use `forgotten_command_line_option(:deployment => true)` rather than passing the flag directly -- see the surrounding specs for an example

adrian-gomez: thanks! I missed that change after the rebase with master

spec/install/deploy_spec.rb
```diff
@@ -295,6 +295,106 @@
...
+      G
+
+      ENV["BUNDLE_ALLOW_DEPLOYMENT_SOURCE_CREDENTIAL_CHANGES"] = "1"
+
+      bundle :install, forgotten_command_line_options(:deployment => true)
+
+      expect(out).to match(/Could not reach host localgemserver.test/)
+    end
+
+    it "prevents to replace a host with the same host with credentials when allow_deployment_source_credential_changes is false" do
```
colby-swandale: `when allow_deployment_source_credential_changes is false` this should be in a context block

spec/install/deploy_spec.rb
```diff
@@ -295,6 +295,106 @@
...
+          rack
+      G
+
+      bundle! "config allow_deployment_source_credential_changes true"
+      bundle :install, forgotten_command_line_options(:deployment => true)
+
+      expect(out).to match(/Could not reach host localgemserver.test/)
+    end
+
+    it "allows to replace a host with the same host with credentials when BUNDLE_ALLOW_DEPLOYMENT_SOURCE_CREDENTIAL_CHANGES env variable" do
```
colby-swandale: `when BUNDLE_ALLOW_DEPLOYMENT_SOURCE_CREDENTIAL_CHANGES env variable` this should be in a context block

spec/install/deploy_spec.rb
```diff
@@ -295,6 +295,106 @@
       expect(out).not_to include("You have deleted from the Gemfile")
     end
 
+    it "allows to replace a host with the same host with credentials when allow_deployment_source_credential_changes is true" do
```
colby-swandale: `when allow_deployment_source_credential_changes is true` this should be in a context block

spec/install/deploy_spec.rb
```diff
@@ -295,6 +295,106 @@
...
+          rack
+      G
+
+      bundle! "config allow_deployment_source_credential_changes false"
+      bundle :install, forgotten_command_line_options(:deployment => true)
+
+      expect(out).to match(/The list of sources changed/)
+    end
+
+    it "prevents to replace a host with the same host with credentials by default" do
```
colby-swandale: `with the same host with credentials by default` should be wrapped in a context block

spec/install/deploy_spec.rb
```diff
@@ -295,6 +295,106 @@
       expect(out).not_to include("You have deleted from the Gemfile")
     end
 
+    it "allows to replace a host with the same host with credentials when allow_deployment_source_credential_changes is true" do
+      install_gemfile(<<-G, :artifice => "endpoint_strict_basic_authentication", :quiet => true)
```
colby-swandale: are the `install_gemfile` and `lockfile` the same for all these tests? If so we should wrap these tests in a `describe` block and use a `before` block so we're not repeating ourselves.

man/bundle-config.ronn
```diff
@@ -121,6 +121,10 @@ learn more about their operation in [bundle install(1)][bundle-install].
 * `allow_bundler_dependency_conflicts` (`BUNDLE_ALLOW_BUNDLER_DEPENDENCY_CONFLICTS`):
    Allow resolving to specifications that have dependencies on `bundler` that
    are incompatible with the running Bundler version.
+* `allow_deployment_source_credential_changes` (`BUNDLE_ALLOW_DEPLOYMENT_SOURCE_CREDENTIAL_CHANGES`):
+   When in deployment mode allows to make changes to the gems source if
+   those changes are only to the credentials used for the source.
```
colby-swandale: This needs some cleanup. Hows: 
```
When in deployment mode, allow changing the credentials to a gem's source.
Ex: `https://some.host.com/gems/path/` -> `https://user_name:password@some.host.com/gems/path`
```

sound?

spec/install/deploy_spec.rb
adrian-gomez: I was just following how its done for the rest of the file, corrected it now

spec/install/deploy_spec.rb
adrian-gomez: I was just following how its done for the rest of the file, corrected it now

spec/install/deploy_spec.rb
adrian-gomez: I was just following how its done for the rest of the file, corrected it now

spec/install/deploy_spec.rb
adrian-gomez: I was just following how its done for the rest of the file, corrected it now

man/bundle-config.ronn
adrian-gomez: done

adrian-gomez:
@indirect @colby-swandale is there a way I can run the failing tests locally? I tried using the same version of ruby and setting the same env vars that are shown on travis but the specs still work for me.
From what I could see on the travis out put it seems to be trying to call methods from `Bundler::Source::Path` on a `Bundler::Source::Source`, specifically `#is_a_path?`

adrian-gomez:
@segiddins ? I can run the tests locally even the whole suite.... and it all works its just one of the settings on travis that fails, I want to know how to replicate the setup that fails locally

segiddins:
Given that travis hasn't run yet, I have no way of knowing what's failing on travis so I can't give any more specific advice

adrian-gomez:
It has run before and this one is the one that failed

screen shot 2017-08-02 at 12 55 31 pm

`Ruby: 2.4.1` `RGV=v2.6.8 BUNDLER_SPEC_SUB_VERSION=2.0.0`

segiddins:
If you want to reproduce that locally, temporarily change the bundler version to `2.0.0`

adrian-gomez:
@segiddins that did it, thanks!

adrian-gomez:
@segiddins @segiddins @colby-swandale Tests are finally green and also addressed all comments so far. Let me know if there is anything else I need to do.

Thanks for all the support.

segiddins:
@bundlerbot r+

deivid-rodriguez:
Hi! I know this was merged a long time ago, but I'm trying to understand why this was added and I'm not sure I get it. As I understand it, the source is the same, and requires credentials, but in development you configure it with `bundle config <SOURCE> <USERNAME>:<API_KEY>`, but you don't want to do that in production and QA, so you use this option and edit the gemfile during deployment to add the credentials. Is that it?

If this is the only reason to add this, in my opinion it should be reverted. You already have several ways to pass the proper credentials in your production environments through different means than local configuration (and you're probably already using those ways to actually enable this configuration option? :thinking:)

Also, the tests for this feature do something that seems really weird to me, and definitely not recommended, which is editing the lockfile directly... :grimacing:



@adrian-gomez Can you clarif

indirect:
@deivid-rodriguez I think I can try to explain this feature a little better...

Bundler (when you are using --deployment mode) throws an exception if anything has changed in the Gemfile.lock. For example, maybe you added a gem but didn't commit the lock. In that case, it's good to throw an exception while trying to deploy.

There's another case, though, where some people want an exception and some people don't want an exception: usernames and passwords for gem sources. For some users, they want to know that they changed the credentials in the Gemfile but forgot to check in the Gemfile.lock, and the "correct" outcome is an exception. For other users, they need to use a different username and password in production than in development or in CI... for those users the "correct" outcome is not an exception. 

They are providing the username and password via the completely usual means of `bundle config` or ENV var, but the credentials happen to be different on the production machines, because of security requirements, or firewall requirements, or some other logistical reason. In those cases, this flag can be turned on to prevent an exception, since it allows credentials to change in frozen mode without raising an exception.

Does that make sense?

deivid-rodriguez:
But in that case, shouldn't the Gemfile and lockfile have no credentials at all, and the credentials be configured through configuration all the time? Or if they really want to change the credentials on the fly (or manually edit them in a production environment for some reason), can't they also run `bundle lock` and update the lockfile as well, so that `bundle install --deployment` does not complain?

I come here because I'm trying to figure out some issues with the deployment flag, but tests for this feature break and I'm not really sure how this is supposed to work, since they don't edit the Gemfile, but the lockfile directly... :grimacing:.

indirect:
Sorry the tests are giving you so much trouble 😬. I think the goal was to keep the guard against changing gems, while removing the exception if the source creds have changed. As far as I know there’s no other way to get both?

deivid-rodriguez:
No problem, I can change the tests to at least test what was intended.

I still don't see the need for this feature, though. I understand that people use different credentials per environment, but that's what configuration is for, and bundler seems quite flexible for configuring source credentials, so I don't see how this option is better than configuring credentials per environment through `bundle config` :thinking:.

On top of that, it sounds like we plan to no longer write credentials to the lock file, so this wouldn't be an issue anyways after https://github.com/bundler/bundler/pull/6823, I think?

indirect:
I think even if you change the credentials with bundle config it raises and exception because the lock has changed? (or at least used to?)

But yes, maybe we can remove this at the same time as we remove credentials in the lock file. 🙂

deivid-rodriguez:
> I think even if you change the credentials with bundle config it raises and exception because the lock has changed? (or at least used to?)

Right, you would need to re-resolve the lockfile before running `bundle install --deployment`. So this feature would be a means of skipping running `bundle lock` before `bundle install --deployment` in those cases.

> But yes, maybe we can remove this at the same time as we remove credentials in the lock file. :slightly_smiling_face:

I'll rebase and fix that PR and see if that makes this feature unnecessary.

PR #5665, opened 2017-05-16
Print errors to stderr

colby-swandale:
As we discussed, i'm going to start porting over features from the 2-0-dev branch into feature flags.

This PR adds the feature to print bundler errors to `stderr` instead of `stdout`. I had a look to see how feature flags worked and i think i have the idea down but let me know if i missed something.

Thanks!

\cc @segiddins @indirect


man/bundle-config.ronn
```diff
@@ -222,6 +222,8 @@ learn more about their operation in [bundle install(1)][bundle-install].
 * `gem.push_key` (`BUNDLE_GEM__PUSH_KEY`):
    Sets the `--key` parameter for `gem push` when using the `rake release`
    command with a private gemstash server.
+* `stderr` (`BUNDLE_STDERR`)
```
segiddins: This probably should be more verbose, like `error_on_stderr`?

colby-swandale: how's `print_to_stderr`?

segiddins: to me that would imply we print _everything_ to stderr

lib/bundler/ui/shell.rb
```diff
@@ -36,6 +36,7 @@ def warn(msg, newline = nil)
       end
 
       def error(msg, newline = nil)
+        return tell_err(msg, :red, newline) if Bundler.feature_flag.stderr?
```
segiddins: probably should do `return unless level("error")` on the line above this

colby-swandale: Thanks!

spec/bundler/ui/shell_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  describe "#error" do
+    before { subject.level = "error" }
+    it "prints to stdout" do
+      expect { subject.error("error!!!") }.to output("error!!!\n").to_stdout
+    end
+
+    context "when stderr flag is enabled" do
+      before { bundle "config error_on_stderr true" }
+      it "prints to stderr" do
+        expect { subject.error("error!!!") }.to output("\e[31merror!!!\e[0m").to_stderr
```
segiddins: why does this require the ANSI escape sequences while the prior test doesn't?

segiddins:
Awesome! We'll eventually want to clean up the specs to work when the flag is enabled, but I don't think you need to do that now

spec/bundler/ui/shell_spec.rb
segiddins: also, should it have a newline like the prior test?

colby-swandale: i'll have to look at `Shell#tell_err` and see what's going on

colby-swandale: I'm not entirely sure what's happening with the ANSI encoding, if i print to stdout out instead the tests pass. 🤷‍♀️ 

```
diff --git a/lib/bundler/ui/shell.rb b/lib/bundler/ui/shell.rb
index 81743ede4..2ee37d4e6 100644
--- a/lib/bundler/ui/shell.rb
+++ b/lib/bundler/ui/shell.rb
@@ -111,8 +111,8 @@ module Bundler
         buffer = @shell.send(:prepare_message, message, *color)
         buffer << "\n" if newline && !message.to_s.end_with?("\n")
 
-        @shell.send(:stderr).print(buffer)
-        @shell.send(:stderr).flush
+        @shell.send(:stdout).print(buffer)
+        @shell.send(:stdout).flush
       end
 
       def strip_leading_spaces(text)
diff --git a/spec/bundler/ui/shell_spec.rb b/spec/bundler/ui/shell_spec.rb
index 2db2d2864..ce76d67aa 100644
--- a/spec/bundler/ui/shell_spec.rb
+++ b/spec/bundler/ui/shell_spec.rb
@@ -42,7 +42,7 @@ RSpec.describe Bundler::UI::Shell do
     context "when stderr flag is enabled" do
       before { bundle "config error_on_stderr true" }
       it "prints to stderr" do
-        expect { subject.error("error!!!") }.to output("\e[31merror!!!\e[0m\n").to_stderr
+        expect { subject.error("error!!!") }.to output("error!!!\n").to_stdout
       end
     end
   end
```

```
› rspec spec/bundler/ui/shell_spec.rb -f progress
Run options: exclude {:rubygems_master=>true, :git=>"=< 2.13.0", :rubygems=>"=< 2.6.8", :ruby=>"=< 2.4.0", :realworld=>true, :sudo=>true}
......

Retried examples: 0


Finished in 0.85979 seconds (files took 0.27336 seconds to load)
6 examples, 0 failures

```

Any ideas?

colby-swandale: Looking into this more and looked at how `rspec-expectations` is capturing the output of `$stderr` and `$stdout` and there is some weird behaviour happening when i try and follow the same method:


```
b = StringIO.new
$stderr = b
$stderr.print "\e[31merror!!!\e[0m\n"
$stderr.flush
puts " stderr: #{b.string.inspect}"

a = StringIO.new
$stdout = a
$stdout.print "\e[31merror!!!\e[0m\n"
$stdout.flush
puts " stdout: #{a.string.inspect}"
```

```
› ruby test.rb
 stderr: "\e[31merror!!!\e[0m\n"
```

I can't see the output from `$stdout` anywhere 🤔

spec/bundler/ui/shell_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  describe "#error" do
+    before { subject.level = "error" }
+    it "prints to stdout" do
+      expect { subject.error("error!!!") }.to output("error!!!\n").to_stdout
+    end
+
+    context "when stderr flag is enabled" do
+      before { bundle "config error_on_stderr true" }
+      it "prints to stderr" do
+        expect { subject.error("error!!!") }.to output("\e[31merror!!!\e[0m\n").to_stderr
```
segiddins: this still shouldn't include the ANSI escape sequences if the `stdout` version doesn't ... (but good job fixing the newlines!)

spec/bundler/ui/shell_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  describe "#error" do
+    before { subject.level = "error" }
+    it "prints to stdout" do
+      expect { subject.error("error!!!") }.to output("error!!!\n").to_stdout
+    end
+
+    context "when stderr flag is enabled" do
+      before { bundle "config error_on_stderr true" }
```
segiddins: to keep the unit tests from spawning subprocesses, this could be `Bundler.settings. temporary(:error_on_stderr => true)`

spec/bundler/ui/shell_spec.rb
colby-swandale: See my notes above re: the ANSI sequence: https://github.com/bundler/bundler/pull/5665#discussion_r117633484

spec/bundler/ui/shell_spec.rb
segiddins: because `puts` is then calling `$stdout.puts`

colby-swandale: Thanks @segiddins. I solved that last night as well but i still have not figured out why i continue to get the ANSI sequences. Sorry for the long delay. 🙇

segiddins:
```diff
diff --git a/lib/bundler/ui/shell.rb b/lib/bundler/ui/shell.rb
index 81743ede4..40e277331 100644
--- a/lib/bundler/ui/shell.rb
+++ b/lib/bundler/ui/shell.rb
@@ -9,7 +9,7 @@ module Bundler
       attr_writer :shell
 
       def initialize(options = {})
-        if options["no-color"] || !STDOUT.tty?
+        if options["no-color"] || !$stdout.tty?
           Thor::Base.shell = Thor::Shell::Basic
         end
         @shell = Thor::Base.shell.new
@@ -108,6 +108,11 @@ module Bundler
         newline = message.to_s !~ /( |\t)\Z/ unless newline
         message = word_wrap(message) if newline.is_a?(Hash) && newline[:wrap]
 
+        if color && !$stderr.tty?
+          color = nil
+        end
+
+
         buffer = @shell.send(:prepare_message, message, *color)
         buffer << "\n" if newline && !message.to_s.end_with?("\n")
 
diff --git a/spec/bundler/ui/shell_spec.rb b/spec/bundler/ui/shell_spec.rb
index 2db2d2864..050fb9c1f 100644
--- a/spec/bundler/ui/shell_spec.rb
+++ b/spec/bundler/ui/shell_spec.rb
@@ -42,7 +42,7 @@ RSpec.describe Bundler::UI::Shell do
     context "when stderr flag is enabled" do
       before { bundle "config error_on_stderr true" }
       it "prints to stderr" do
-        expect { subject.error("error!!!") }.to output("\e[31merror!!!\e[0m\n").to_stderr
+        expect { subject.error("error!!!") }.to output("error!!!\n").to_stderr
       end
     end
   end
```

colby-swandale:
thank you!! 🙇

segiddins:
👍 
@indirect r?

indirect:
Awesome. This is way better than maintaining a branch. 😁 @bundlerbot r+

PR #5667, opened 2017-05-16
Avoid Rack 1.6.7 in specs for now

segiddins:
See https://github.com/rack/rack/issues/1168
The above issue broke rack on 1.9.3

\c @colby-swandale


colby-swandale:
thank you! @bundlerbot r+

PR #5670, opened 2017-05-19
[DSL] Allow forcing a particular version of a gem

### LINKED ISSUE #4552: Provide an option to force a gem version, even it it conflicts with a gem's declared compatibility.
_From @xaviershay on August 22, 2013 3:22_

From https://github.com/bundler/bundler/issues/2412

Still debate about how this would be done or whether it's a good idea. See linked ticket for context.

_Copied from original issue: bundler/bundler-features#20_

segiddins:
Closes https://github.com/bundler/bundler/issues/4552.

Even if it conflicts. This will hopefully add a way out of situations where a gem author has added overly-strict requirements, without adding a huge amount of overhead in bundler or conceptually for a user. By forcing the Gemfile to explicitly say what version is being forced, there's no opportunity for something unexpected to happen.

This is rough & unpolished, and I'd appreciate feedback on the UX, but I believe the basic implementation works (as I imagined this feature working :P).

### DIFF
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -15,7 +15,7 @@ module Bundler
     VALID_PLATFORMS = Bundler::Dependency::PLATFORM_MAP.keys.freeze
 
     VALID_KEYS = %w[group groups git path glob name branch ref tag require submodules
-                    platform platforms type source install_if].freeze
+                    platform platforms type source install_if force_version].freeze
 
     attr_reader :gemspecs
     attr_accessor :dependencies
@@ -374,6 +374,10 @@ repo_name ||= user_name
       if opts.key?("source")
         source = normalize_source(opts["source"])
         opts["source"] = @sources.add_rubygems_source("remotes" => source)
+      end
+
+      if opts.key?("force_version") && (r = Gem::Requirement.new(version)) && !r.exact?
+        raise GemfileError, "Cannot use force_version for inexact version requirement `#{r}`"
       end
 
       git_name = (git_names & opts.keys).last
--- a/lib/bundler/rubygems_ext.rb
+++ b/lib/bundler/rubygems_ext.rb
@@ -166,6 +166,10 @@ module Gem
 
       requirement.satisfied_by?(spec.version)
     end unless allocate.respond_to?(:matches_spec?)
+
+    def force_version?
+      false
+    end
   end
 
   class Requirement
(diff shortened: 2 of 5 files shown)


lib/bundler/dsl.rb
```diff
@@ -351,6 +351,10 @@ def normalize_options(name, version, opts)
         opts["source"] = @sources.add_rubygems_source("remotes" => source)
       end
 
+      if opts.key?("force_version") && (r = Gem::Requirement.new(version)) && !r.exact?
+        raise GemfileError, "Cannot use force_version for inexact version requirement `#{r}`"
```
indirect: I think the error message would be a lot more helpful if it said "force_version can only be used with exact version numbers", does that seem reasonable?

lib/bundler/rubygems_ext.rb
```diff
@@ -166,6 +166,10 @@ def matches_spec?(spec)
 
       requirement.satisfied_by?(spec.version)
     end unless allocate.respond_to?(:matches_spec?)
+
+    def force_version?
+      false
+    end
```
indirect: can we avoid monkeypatching by doing a responds_to check instead?

deivid-rodriguez:
Seems handy, I just hope it's not abused...

lib/bundler/rubygems_ext.rb
segiddins: we could, I'm just worried about the performance implications of that. Note that we already monkey-patch in `groups`

lib/bundler/dsl.rb
segiddins: yup

indirect:
@segiddins I'm trying to figure out if we can test this prior to a full release. A config-based feature flag seems bad, because then `bundle install` would only work for users who manually set the flag. Maybe we can ship it enabled only in 1.99+? Open to ideas.

segiddins:
It's technically already opt-in, as it's based upon an entirely new dependency option

indirect:
Yeah, that's fair. I'm mostly trying to avoid merging and shipping and then regretting it for reasons we will only discover later. 😬

segiddins:
@indirect we can do it in 2.0, that way you have the whole beta period to mull it over? We can also add a warning that the option is still unsupported if you like, and reserve the right to nuke it for the time being?

indirect:
@segiddins putting it into 2.0 seems like a safe way to do it... that way we can roll back during the beta if everything goes terribly wrong 😬

segiddins:
Rebasing

kinduff:
Was this included in new releases? Or are we waiting for a merge

colby-swandale:
@kinduff This PR has not has been merged.

deivid-rodriguez:
This one was ported over to the new repo at https://github.com/rubygems/rubygems/pull/4178, so closing!

Issue #5467, opened 2017-02-22
Add `bundle plugin list` command


kerrizor:
This command would list all the commands for installed plugins. A potential format for the output might be similar to `bundle plugin help`

```
$ bundle plugin list
Installed Plugins
  foo   # Find all gems whose name contains the word 'foo'
  bar   # Print the word 'bar' 16 times
  baz   # Make modem noises and draw an ANSI picture of a cactus
```

This would require CLI command, of course, as well as extending the plugin API to require that a "description" string be set (with a maximum length?) This could be enforced via `lib/bundler/plugin.rb#validate_plugin!`

Shekharrajak:
I am interested in working on this issue. But I am not able to understand plugin system. I don't find documentation and examples. If there is any related links available let me know.

kerrizor:
There are 2 kinds of plugins. One, the "old" way, is to simply have an executable in your $PATH that is named `bundler-<command>`. The "new" way is to use the Bundler::Plugin::API, which requires users to "register" the plugin by using the `bundle plugin install <command>` technique. An example of a working "new" style plugin can be found at http://github.com/kerrizor/bundler-dependency_graph

Bundler stores information about installed plugins in `~/.bundle/plugins` and I was thinking we could "easily" extend the Plugin::API DSL to allow plugins to declare their description (similar to how we declare the command here: https://github.com/kerrizor/bundler-dependency_graph/blob/master/lib/bundler/dependency_graph/graph_builder.rb#L4

Shekharrajak:
Thanks @kerrizor .

Shekharrajak:
Output must be list of commands (not the plugin name) ,right ?

kerrizor:
My instinct says yes, a list of commands is more useful. 3 options:

- `bundle plugin list` shows plugins, `bundle plugin commands` shows the commands
- show a list of commands as sublists of plugins, something like:
```
$ bundle plugin list
Installed Plugins
  foo
  ---
    foo   # Find all gems whose name contains the word 'foo'

  bar
  ---
    bar   # Print the word 'bar' 16 times
    baz   # Make modem noises and draw an ANSI picture of a cactus
```

- list plugins "inline" so more like this
```
$ bundle plugin list
Installed Plugins
  foo   (foo)   # Find all gems whose name contains the word 'foo'
  bar   (bar)   # Print the word 'bar' 16 times
  baz   (bar)   # Make modem noises and draw an ANSI picture of a cactus
```

What do you think?

[...]

PR #5672, opened 2017-05-23
add `bundle plugin list` command

fotanus:
Hi guys, can someone do an early review in this PR? It is a work in progress to implement the `bundle plugin list` command. I'll commend some parts of the code where I specifically need help. Thanks!

### DIFF
--- a/lib/bundler/plugin.rb
+++ b/lib/bundler/plugin.rb
@@ -45,6 +45,26 @@ module Bundler
       Bundler.ui.error "Failed to install plugin #{name}: #{e.message}\n  #{e.backtrace.join("\n ")}"
     end
 
+    # List installed plugins and commands
+    #
+    def list
+      installed_plugins = index.installed_plugins
+      if installed_plugins.any?
+        output = String.new
+        installed_plugins.each do |plugin|
+          output << "#{plugin}\n"
+          output << "-----\n"
+          index.plugin_commands(plugin).each do |command|
+            output << "  #{command}\n"
+          end
+          output << "\n"
+        end
+      else
+        output = "No plugins installed"
+      end
+      Bundler.ui.info output
+    end
+
     # Evaluates the Gemfile with a limited DSL and installs the plugins
     # specified by plugin method
     #
--- a/lib/bundler/plugin/index.rb
+++ b/lib/bundler/plugin/index.rb
@@ -95,6 +95,14 @@ module Bundler
         @plugin_paths[name]
       end
 
+      def installed_plugins
+        @plugin_paths.keys
+      end
+
+      def plugin_commands(plugin)
+        @commands.find_all {|_, n| n == plugins }.map(&:first)
+      end
+
       def source?(source)
         @sources.key? source
       end
--- a/spec/bundler/plugin_spec.rb
+++ b/spec/bundler/plugin_spec.rb
@@ -32,6 +32,31 @@ RSpec.describe Bundler::Plugin do
     allow(index).to receive(:register_plugin)
   end
 
+  describe "list command" do
+    let(:opts) { { "version" => "~> 1.0", "source" => "foo" } }
+
+    context "when no plugins are installed" do
+      before { allow(index).to receive(:installed_plugins) { [] } }
+      it "outputs no plugins installed" do
+        expect(Bundler.ui).to receive(:info).with("No plugins installed")
+        subject.list
+      end
+    end
+
+    context "with installed plugins" do
+      before do
+        allow(index).to receive(:installed_plugins) { %w[plug1 plug2] }
+        allow(index).to receive(:plugin_commands).with("plug1") { %w[c11 c12] }
+        allow(index).to receive(:plugin_commands).with("plug2") { %w[c21 c22] }
+      end
+      it "list plugins followed by commands" do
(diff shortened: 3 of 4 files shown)


lib/bundler/plugin.rb
```diff
@@ -45,6 +45,22 @@ def install(names, options)
       Bundler.ui.error "Failed to install plugin #{name}: #{e.message}\n  #{e.backtrace.join("\n ")}"
     end
 
+    # List installed plugins and commands
+    #
+    def list
+      output = ""
```
fotanus: Where is the best place to format the string that will be printed as result of the command? And where this should it actually be printed?

lib/bundler/plugin/index.rb
```diff
@@ -95,6 +95,14 @@ def installed?(name)
         @plugin_paths[name]
       end
 
+      def installed_plugins
+        @plugin_paths.keys
+      end
+
+      def plugin_commands(plugin)
+        @commands.find_all{|c,n| n == plugins}.map(&:first)
```
fotanus: Wouldn't be great to have a class for plugins, which lists the commands it have instead of using hashes to do this work? I assume hashes are faster, but classes more readable? We can, of course, add a reverse hash for this specific method.

spec/bundler/plugin_spec.rb
```diff
@@ -33,6 +33,18 @@
     allow(index).to receive(:register_plugin)
   end
 
+  describe "list command" do
+    let(:opts) { { "version" => "~> 1.0", "source" => "foo" } }
+    before do
+      allow(index).to receive(:installed_plugins) { %W(plug1 plug2) }
+      allow(index).to receive(:plugin_commands).with("plug1") { %W(c1 c2) }
+      allow(index).to receive(:plugin_commands).with("plug2") { %W(c1 c2) }
```
fotanus: I don't feel like I'm really testing anything here with index being a double :-(
Is that is how it is supposed to be?

lib/bundler/plugin.rb
fotanus: I just droped Bundler.ui.info in the end with the output information. Not sure if is the best.

fotanus:
This PR is for issue #5467. In the issue I see the idea of adding the command description, but I don't think our plugin index support command descriptions?

lib/bundler/plugin.rb
```diff
@@ -45,6 +45,26 @@ def install(names, options)
       Bundler.ui.error "Failed to install plugin #{name}: #{e.message}\n  #{e.backtrace.join("\n ")}"
     end
 
+    # List installed plugins and commands
+    #
+    def list
+      installed_plugins = index.installed_plugins
```
colby-swandale: i would add add `output = String.new` and then append to it to help make this more readable.

fotanus: Thanks for the review! By append you mean `<<`?

colby-swandale: yes

indirect:
Awesome, thanks! @bundlerbot r+

colby-swandale:
@fotanus can you please fix the rubocop feedback. Thanks! 👍

fotanus:
Fixed! 
on travis:

```
spec/bundler/plugin_spec.rb:48:55: C: Style/PercentLiteralDelimiters: %w-literals should be delimited by [ and ].
        allow(index).to receive(:installed_plugins) { %w(plug1 plug2) }
```

and but running `rubocop` on my machine I got the reverse:

```
spec/bundler/plugin_spec.rb:49:55: C: Style/PercentLiteralDelimiters: %w-literals should be delimited by ( and ).
        allow(index).to receive(:installed_plugins) { %w[plug1 plug2] }
                                                      ^^^^^^^^^^^^^^^
```

Any tips on how to make my rubocop behaves the same than travis?

segiddins:
rebase against master

fotanus:
@segiddins of course, now I get the right errors, ty

segiddins:
@colby-swandale r?

lib/bundler/plugin/index.rb
```diff
@@ -95,6 +95,14 @@ def installed?(name)
         @plugin_paths[name]
       end
 
+      def installed_plugins
+        @plugin_paths.keys
+      end
+
+      def plugin_commands(plugin)
```
colby-swandale: Is this suppose to be `plugins` instead of `plugin`?

colby-swandale: ping @fotanus

fotanus: Sorry for the delay @colby-swandale, I wrote this PR 5 months ago, and I'm a bit out of what is happening on bundler. I think that on line 102 it should be `plugin`, and in line 103 also `plugin`, since the method appears to list the commands for one plugin. I'm actually unsure how this pass on the test? Right now it seems like a typo to me

colby-swandale: Are you still working on this PR?

fotanus: Feel free to take over or reimplement copying any ideas you want :-) I can't deal with this right now.

segiddins:
Woah, how did this get merged?

colby-swandale:
@segiddins This is a GitHub quirk. #6120 was merged which contains the commits of this PR so GitHub just marks it as merged as well. 🤷‍♀️

segiddins:
Ah ok, cool

PR #5673, opened 2017-05-23
[GitProxy] Force deinitializing submodules

segiddins:
Fixes  the following test under newer git versions

```
bundle update
  git sources
    with submodules
      unlocks the source when submodules are removed from git source
```


indirect:
nice 👍 @bundlerbot r+

PR #5679, opened 2017-05-25
[Env] Print well-formed markdown when there are no settings

### LINKED ISSUE #5677: bundle env has bad markdown syntax
When I run `bundle env`, it outputs in Markdown syntax. However, there's an error in the syntax. For example, here's the beginning of the output:

## Environment

```
Bundler   1.15.0
Rubygems  2.6.12
Ruby      2.0.0p648 (2015-12-16 revision 53162) [x86_64-darwin16.6.0]
GEM_HOME  /Users/bhawley/.rvm/gems/ruby-2.0.0-p648@oceans
GEM_PATH  /Users/bhawley/.rvm/gems/ruby-2.0.0-p648@oceans:/Users/bhawley/.rvm/gems/ruby-2.0.0-p648@global
RVM       1.29.1 (master)
Git       2.13.0
Platform  x86_64-darwin-16
rubygems-bundler (1.4.4)
```
```

## Gemfile

### Gemfile

```ruby
source 'https://rubygems.org'
```

It continues from there. The problem is the extra "```" after the Environment block (there are two of them), which messes up the syntax of the Gemfile following it.

segiddins:
Closes #5677

### DIFF
--- a/lib/bundler/env.rb
+++ b/lib/bundler/env.rb
@@ -29,14 +29,16 @@ module Bundler
 
       out << "```\n"
 
-      out << "\n## Bundler settings\n\n```\n" unless Bundler.settings.all.empty?
-      Bundler.settings.all.each do |setting|
-        out << setting << "\n"
-        Bundler.settings.pretty_values_for(setting).each do |line|
-          out << "  " << line << "\n"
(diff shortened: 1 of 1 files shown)


indirect:
@bundlerbot r+

PR #5688, opened 2017-05-31
Fix markdown headers in doc/development/readme

segiddins:
GH now requires a space between `#` and the header


segiddins:
@bundlerbot r+

PR #5690, opened 2017-05-31
Fix `bundle init --gemspec` breaking when gemspec is invalid

colby-swandale:
Also added some specs to init what wasn't being tested.

### DIFF
--- a/lib/bundler/cli/init.rb
+++ b/lib/bundler/cli/init.rb
@@ -19,11 +19,7 @@ module Bundler
           exit 1
         end
 
-        spec = Gem::Specification.load(gemspec)
-        unless spec
-          Bundler.ui.error "Gem specification #{gemspec} is invalid"
-          exit 1
-        end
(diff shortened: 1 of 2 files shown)


colby-swandale:
Seems like travis is having issues with the current  condition of Github. Will wait until morning and restart it if it still doesn't kick in then.

lib/bundler/cli/init.rb
```diff
@@ -18,7 +18,13 @@ def run
           Bundler.ui.error "Gem specification #{gemspec} doesn't exist"
           exit 1
         end
+
         spec = Gem::Specification.load(gemspec)
```
segiddins: Let's use the Bundler module method to load the gemspec?

segiddins: `Bundler.load_gemspec_uncached`

colby-swandale: Yes! 🎉

segiddins:
@bundlerbot r+

PR #5702, opened 2017-06-03
[Installer] Load plugin files from path gems

### LINKED ISSUE #5429: RubyGems post-install hook not triggering
I see a lot of talk about Bundler install hooks on different issues/PRs that (like this thread) mentions that Bundler respects RubyGems post install hooks, but I'm testing it out and that doesn't seem to be the case… if I install via `gem install` the hook runs, but if I add the gem to a Gemfile and install it doesn't seem to be triggering the hook. Is that unexpected?

segiddins:
Closes #5429 .

Because RubyGems doesn't know about path gems before we install, we need to manually load the plugin files for the path gems we're installing. This is basically copying the logic RG uses, but scoped only to those gems that we're entirely responsible for.

### DIFF
--- a/lib/bundler/installer.rb
+++ b/lib/bundler/installer.rb
@@ -162,11 +162,21 @@ module Bundler
     # that said, it's a rare situation (other than rake), and parallel
     # installation is SO MUCH FASTER. so we let people opt in.
     def install(options)
-      Bundler.rubygems.load_plugins
+      load_plugins
       force = options["force"]
       jobs = 1
       jobs = [Bundler.settings[:jobs].to_i - 1, 1].max if can_install_in_parallel?
       install_in_parallel jobs, options[:standalone], force
+    end
+
+    def load_plugins
+      Bundler.rubygems.load_plugins
+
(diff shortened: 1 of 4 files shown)


lib/bundler/installer.rb
```diff
@@ -162,13 +162,23 @@ def generate_standalone_bundler_executable_stubs(spec)
...
       jobs = [Bundler.settings[:jobs].to_i - 1, 1].max if can_install_in_parallel?
       install_in_parallel jobs, options[:standalone], force
     end
 
+    def load_plugins
+      Bundler.rubygems.load_plugins
+
+      requested_path_gems = @definition.requested_specs.select {|s| s.source.is_a?(Source::Path) }
+      path_plugin_files = requested_path_gems.map do |spec|
+        spec.matches_for_glob("rubygems_plugin#{Bundler.rubygems.suffix_pattern}")
```
segiddins: failing on RG < 1.8

segiddins:
This ought to pass now

segiddins:
Fixed

indirect:
@bundlerbot r+

PR #5703, opened 2017-06-03
[Env] Print shebang when it doesnt match Gem.ruby

### LINKED ISSUE #5616: `bundler env` should show bundler's shebang if it differs from the first found ruby
I was starting to tear my hair out today in a crazy debugging situation. I ran `gem install bundler` on OSX, and then trying to install various gems. However, I got ruby version incompatibilities: our rails app specifies ruby 2.3, but bundler reported ruby 2.0.

It turns out that rubygem's had written the bundler binstub with OSX's system ruby for it's shebang, and not the ruby installed with Homebrew. It would have been faster to debug if I had realized the discrepancy earlier.

segiddins:
Closes #5616.

Prints out `Gem.ruby` and the `bundle` binstub shebang when they don't match, which should help us pinpoint when ruby installation shenanigans are the root cause of an issue

### DIFF
--- a/lib/bundler/env.rb
+++ b/lib/bundler/env.rb
@@ -26,6 +26,14 @@ module Bundler
         specs = Bundler.rubygems.find_name(name)
         out << "#{name} (#{specs.map(&:version).join(",")})\n" unless specs.empty?
       end
+      if (exe = caller.last.split(":").first) && exe =~ %r{(exe|bin)/bundler?\z}
+        shebang = File.read(exe).lines.first
+        shebang.sub!(/^#!\s*/, "")
+        unless shebang.start_with?(Gem.ruby, "/usr/bin/env ruby")
+          out << "Gem.ruby  #{Gem.ruby}\n"
+          out << "bundle #! #{shebang}\n"
+        end
+      end
 
       out << "```\n"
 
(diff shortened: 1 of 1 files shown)


lib/bundler/env.rb
```diff
@@ -13,20 +13,11 @@ def self.report(options = {})
...
-      out << "GEM_PATH  #{ENV["GEM_PATH"]}\n" unless ENV["GEM_PATH"] == ENV["GEM_HOME"]
-      out << "RVM       #{ENV["rvm_version"]}\n" if ENV["rvm_version"]
-      out << "Git       #{git_version}\n"
-      out << "Platform  #{Gem::Platform.local}\n"
-      out << "OpenSSL   #{OpenSSL::OPENSSL_VERSION}\n" if defined?(OpenSSL::OPENSSL_VERSION)
-      %w[rubygems-bundler open_gem].each do |name|
-        specs = Bundler.rubygems.find_name(name)
-        out << "#{name} (#{specs.map(&:version).join(",")})\n" unless specs.empty?
+      env = environment
+      environment_ljust = env.map {|(k, _v)| k.to_s.length }.max
```
segiddins: `_v` is giving an unused variable warning on < 2

indirect:
oh my god what a good idea yes please @bundlerbot r+

PR #5705, opened 2017-06-04
[DSL] Deprecate the github source for 2.0

### LINKED ISSUE #5698: Remove `github.https` config option
It isn't needed anymore in 2.0 anyway, so let's get rid of it. Relatedly, it was a bad idea to have this option in the first place, because machines with this option set will generate different lockfiles than machines without this option set. 😬

### LINKED ISSUE #5699: Remove `github` shortcut, with warning in 1.99 that points to docs
Rails apps already generate with a `git_source` directive because of the github HTTPS thing, so this isn't even a big change at this point. On top of that, I buy the argument that we can let users handle their own shortcuts, putting github/gitlab/bitbucket/etc on an equal footing. Let's get rid of our own hardcoded github shortcut and let users set up whatever make sense for them.

I'm also toying around with the idea of adding `git_source(:github) { |name| "https://github.com/#{name}.git" }` to the Gemfile generated by `init`, with a comment explaining that you should change the name and the URL to match whatever git host you use.

segiddins:
Closes #5699.
Closes #5698.

@indirect is this what you wanted?

### DIFF
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -252,6 +252,9 @@ module Bundler
 
     def add_git_sources
       git_source(:github) do |repo_name|
+        warn_deprecated_git_source(:github, <<-'RUBY'.strip, 'Change any "reponame" :github sources to "username/reponame".')
+"https://github.com/#{repo_name}.git"
+        RUBY
         # It would be better to use https instead of the git protocol, but this
         # can break deployment of existing locked bundles when switching between
         # different versions of Bundler. The change will be made in 2.0, which
@@ -268,23 +271,29 @@ module Bundler
         repo_name = "#{repo_name}/#{repo_name}" unless repo_name.include?("/")
         # TODO: 2.0 upgrade this setting to the default
         if Bundler.settings["github.https"]
+          Bundler::SharedHelpers.major_deprecation "The `github.https` setting will be removed"
           "https://github.com/#{repo_name}.git"
         else
-          warn_github_source_change(repo_name)
           "git://github.com/#{repo_name}.git"
         end
       end
 
       # TODO: 2.0 remove this deprecated git source
       git_source(:gist) do |repo_name|
-        warn_deprecated_git_source(:gist, 'https://gist.github.com/#{repo_name}.git')
+        warn_deprecated_git_source(:gist, '"https://gist.github.com/#{repo_name}.git"')
+
         "https://gist.github.com/#{repo_name}.git"
       end
 
       # TODO: 2.0 remove this deprecated git source
       git_source(:bitbucket) do |repo_name|
-        user_name, repo_name = repo_name.split "/"
-        warn_deprecated_git_source(:bitbucket, 'https://#{user_name}@bitbucket.org/#{user_name}/#{repo_name}.git')
+        warn_deprecated_git_source(:bitbucket, <<-'RUBY'.strip)
+user_name, repo_name = repo_name.split("/")
+repo_name ||= user_name
+"https://#{user_name}@bitbucket.org/#{user_name}/#{repo_name}.git"
+        RUBY
+
+        user_name, repo_name = repo_name.split("/")
         repo_name ||= user_name
         "https://#{user_name}@bitbucket.org/#{user_name}/#{repo_name}.git"
       end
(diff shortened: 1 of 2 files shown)


indirect:
@segiddins I think so! The config deprecation needs an explanation of how to build a github option with https to use instead. I think I'd also like to switch the git sources to expect to always have username/reponame, because that makes the code in the gemfiles so much simpler. What do you think?

segiddins:
> The config deprecation needs an explanation of how to build a github option with https to use instead

It will be printed right after we show how to re-implement the `github` shortcut, so I think saying how to do it twice in a row would be redundant?

>  I think I'd also like to switch the git sources to expect to always have username/reponame, because that makes the code in the gemfiles so much simpler. What do you think?

Sure, but then the suggested shortcut wouldn't be a drop-in replacement. Your call.

indirect:
Since we're breaking compat anyway, let's do it. The warning can add an extra line that says "You will also need to give the repo path as username/reponame" if the argument doesn't have a slash in it. 👍

segiddins:
@indirect changes made, tests fixed

indirect:
lol I'm sorry, this is exactly what I said but it turns out not what I meant. If the github source already provides a `user/repo` string, I want the warning to look like this:

```
The :github git source is deprecated, and will be removed in Bundler 2.0. Add this code to your Gemfile to ensure it continues to work:
    git_source(:github) { |repo| "https://github.com/\#{repo}.git" }
```

On the other hand, if the github source provides a `repo` string with no username, I want the warning to include an additional sentence of instructions, like this:

```
The :github git source is deprecated, and will be removed in Bundler 2.0. Add this code to your Gemfile and provide the repo in the form "username/reponame" to ensure it continues to work:
    git_source(:github) { |repo| "https://github.com/\#{repo}.git" }
```

segiddins:
> On the other hand, if the github source provides a repo string with no username, I want the warning to include an additional sentence of instructions, like this:

The issue then is that you could have the deprecation printed twice, once with the warning and once without

indirect:
I'm okay with writing code to ensure it only prints once, or we could have a single unified warning like this:

```
The :github git source is deprecated, and will be removed in Bundler 2.0. Change any "reponame" 
:github sources to "username/reponame", and then add this code to your Gemfile to ensure
it continues to work:
    git_source(:github) { |repo| "https://github.com/\#{repo}.git" }
```

segiddins:
Done! r?

lib/bundler/dsl.rb
```diff
@@ -252,6 +252,10 @@ def method_missing(name, *args)
 
     def add_git_sources
       git_source(:github) do |repo_name|
+        warn_deprecated_git_source(:github, <<-'RUBY'.strip, 'Change any "reponame" :github sources to "username/reponame".')
+# You will also need to give the repo path as username/reponame if the argument doesn't have a slash in it.
```
indirect: comment is now superfluous

spec/other/major_deprecation_spec.rb
```diff
@@ -178,17 +178,33 @@
...
-        msg = "The :github option uses the git: protocol, which is not secure. " \
-        "Bundler 2.0 will use the https: protocol, which is secure. Enable this change now by " \
-        "running `bundle config github.https true`."
+        msg = <<-EOS
+The :github git source is deprecated, and will be removed in Bundler 2.0. Change any "reponame" :github sources to "username/reponame". Add this code to your Gemfile to ensure it continues to work:
+
+    git_source(:github) do |repo_name|
+      # You will also need to give the repo path as username/reponame if the argument doesn't have a slash in it.
+      "https://github.com/\#{repo_name}.git"
+    end
```
indirect: I'd much rather have this be a oneliner for ease of insertion at the top of Gemfiles, as in my comment.

spec/other/major_deprecation_spec.rb
```diff
@@ -178,17 +178,33 @@
 
     context "with github gems" do
       it "warns about the https change" do
-        msg = "The :github option uses the git: protocol, which is not secure. " \
-        "Bundler 2.0 will use the https: protocol, which is secure. Enable this change now by " \
-        "running `bundle config github.https true`."
+        msg = <<-EOS
+The :github git source is deprecated, and will be removed in Bundler 2.0. Change any "reponame" :github sources to "username/reponame". Add this code to your Gemfile to ensure it continues to work:
```
indirect: I think we need "Add this code at the beginning of your Gemfile", because the git source has to be declared before it is used.

segiddins:
Dome

indirect:
@bundlerbot r+

segiddins:
@bundlerbot r=indirect

segiddins:
@bundlerbot retry

PR #5716, opened 2017-06-12
Print build metadata when running `bundle version`

### LINKED ISSUE #5049: Print git commit when running Bundler from a git checkout
When running bundler directly from a git checkout (for example via the `dbundle` alias suggested by DEVELOPMENT.md), it would be SUPER helpful to print the git commit out when running `dbundle --version`, and also when running `bundle env`.

segiddins:
Closes #5049.

Will get all the build metadata into `bundle env` once https://github.com/bundler/bundler/pull/5703 lands, since I don't want conflicts and want to use that code for generating the "tables"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@ man/*
 
 # rspec failure tracking
 .rspec_status
+
+# files generated during packaging
+/lib/bundler/generated/
--- a/Rakefile
+++ b/Rakefile
@@ -355,9 +355,11 @@ task :update_certs => "spec:rubygems:clone_rubygems_master" do
 end
 
 require "bundler/gem_tasks"
-task :build => ["man:build"]
+task :build => ["man:build", "generate_files"]
 task :release => ["man:require", "man:build"]
 
 task :default => :spec
 
 Dir["task/*.{rb,rake}"].each(&method(:load))
+
+task :generate_files => Rake::Task.tasks.select {|t| t.name.start_with?("lib/bundler/generated") }
--- a/Rakefile
+++ b/Rakefile
@@ -354,10 +354,6 @@ task :update_certs => "spec:rubygems:clone_rubygems_master" do
   Bundler::SSLCerts::CertificateManager.update_from!(RUBYGEMS_REPO)
 end
 
-require "bundler/gem_tasks"
-task :build => ["man:build", "generate_files"]
-task :release => ["man:require", "man:build"]
-
 task :default => :spec
 
 Dir["task/*.{rb,rake}"].each(&method(:load))
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -13,7 +13,7 @@ require "bundler/rubygems_integration"
 require "bundler/version"
 require "bundler/constants"
 require "bundler/current_ruby"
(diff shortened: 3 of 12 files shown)


indirect:
Does this also work with `dbundle version`? Maybe I'm missing something there...

segiddins:
> Does this also work with dbundle version? Maybe I'm missing something there...

Only if `rake generate_metadata` has been run, unfortunately. I couldn't think of a way to make it work for both `dbundle` and `rake install` and released version all at the same time quite yet

indirect:
@segiddins maybe the blank metadata file could be dynamic, and use the current date and git sha? Since `rake` always generates a static file, it would always be correct (as long as the rake task puts the dynamic file back when it finishes running).

segiddins:
> (as long as the rake task puts the dynamic file back when it finishes running).

wait, what?!?

indirect:
The blank file is checked in! The rake task writes a static file that gets packaged into the .gem, and it can `git checkout HEAD -- lib/bundler/generated/metadata.rb` as a final step after building the .gem. Maybe.

segiddins:
Try it now! I think this will make it work with `dbundle` as well, unless you ever run `rake generate_files` which shouldn't be needed

segiddins:
Ooooh, super evil idea....
What if generating that file also creates a post-commit hook to run the rake task?

indirect:
The commit that makes the defaults "dynamic" checks in the dynamic values... :joy:

indirect:
Hmm... maybe. I'm a little bit wary of post-commit hooks because I'm hoping that I can use this on any checkout where I run `dbundle`, not only checkouts where I've run the rake task.

segiddins:
But where you haven't run the rake task, it'll pull them dynamically

indirect:
Is there any time a dev would want to run `rake generate_files`? Or is it solely something that happens as a dependency of `rake install` or `rake release`?

indirect:
Does this mean that `dbundle version` stops working after you've run `rake install` or `rake release`?

segiddins:
Nope, since either of those rake tasks would install the git hooks

indirect:
Ahhh, I see. So the file is fully dynamic unless/until a release happens, and then the dynamic file is overwritten with a static file that will be silently updated in a post-commit hook. I'm a little skeptical, mainly because git commit hooks seem like a complex solution to the dynamic file getting overwritten once. I think I would prefer that the file be static only inside the .gem, and dynamic the rest of the time. In theory, that could even work without Rake having to call git at all, by saving off a copy of the dynamic file, writing a static file, building the .gem, and then putting the dynamic file back. What are your thoughts?

segiddins:
But that'll make every `dbundle` call slower 🙃

indirect:
True. Hmm... how about switching from hash values to method bodies, so the git call and the date are lazily evaluated only when they're actually needed to print the version number?

segiddins:
> True. Hmm... how about switching from hash values to method bodies, so the git call and the date are lazily evaluated only when they're actually needed to print the version number?

Done!

indirect:
woooo @bundlerbot r+

PR #5701, opened 2017-06-03
Ensure Bundler.original_env preserves _all_ keys bundler sets

segiddins:
Closes #5700.

Attempts to make sure we properly preserve missing env variables such as `RUBYOPT` across `bundle exec` invocations so that `Bundler.original_env` is accurate

PR #5718, opened 2017-06-13
Eval Gemfiles one fewer time when running `bundle install`

segiddins:
This is just a straight port of https://github.com/bundler/bundler/pull/4952 to master from `2-0-dev`.
Unfortunately, the Gemfile is still eval'ed twice when plugins are enabled.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -194,10 +194,19 @@ module Bundler
       missing
     end
 
-    def missing_dependencies
+    def missing_dependencies?
       missing = []
       resolve.materialize(current_dependencies, missing)
-      missing
+      return false if missing.empty?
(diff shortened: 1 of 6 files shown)


indirect:
@bundlerbot r+

segiddins:
@jules2689 this might also help perf, but I'm not sure

jules2689:
Hrm, I can't actually run master right now due to https://github.com/bundler/bundler/pull/5701#issuecomment-308926849

jules2689:
Including a fix for the issue I noted above, I got 375-389ms on master and 369-381ms using Shopify's core application for `require 'bundler/setup'`, so there is definitely about 6-10ms of savings here. I'd just like to also note that this number used to be basically a full second - this is great :D

segiddins:
@bundlerbot retry

PR #5720, opened 2017-06-13
[2.0] Add a compatibility guard that prints friendly errors on bundler 2+

segiddins:
This way, if someone manages to get bundler 2 installed on a system we don't support, they'll get a very obvious message saying so

### DIFF
--- a/lib/bundler/compatibility_guard.rb
+++ b/lib/bundler/compatibility_guard.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+require "rubygems"
+require "bundler/version"
+
+if Bundler::VERSION.split(".").first.to_i >= 2
+  if Gem::Version.new(Object::RUBY_VERSION) < Gem::Version.new("2.0.0")
+    abort "Bundler 2 requires Ruby 2+. Either install bundler 1 or update to a supported Ruby version."
+  end
+
+  if Gem::Version.new(Gem::VERSION) < Gem::Version.new("2.0.0")
+    abort "Bundler 2 requires RubyGems 2+. Either install bundler 1 or update to a supported RubyGems version."
+  end
+end
--- a/lib/bundler/compatibility_guard.rb
+++ b/lib/bundler/compatibility_guard.rb
@@ -1,14 +1,14 @@
-# frozen_string_literal: true
+# frozen_string_literal: false
 
 require "rubygems"
 require "bundler/version"
 
 if Bundler::VERSION.split(".").first.to_i >= 2
-  if Gem::Version.new(Object::RUBY_VERSION) < Gem::Version.new("2.0.0")
-    abort "Bundler 2 requires Ruby 2+. Either install bundler 1 or update to a supported Ruby version."
+  if Gem::Version.new(Object::RUBY_VERSION) < Gem::Version.new("2.3")
+    abort "Bundler 2 requires Ruby 2.3 or later. Either install bundler 1 or update to a supported Ruby version."
   end
 
-  if Gem::Version.new(Gem::VERSION) < Gem::Version.new("2.0.0")
-    abort "Bundler 2 requires RubyGems 2+. Either install bundler 1 or update to a supported RubyGems version."
+  if Gem::Version.new(Gem::VERSION.dup) < Gem::Version.new("2.5")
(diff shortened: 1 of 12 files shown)


segiddins:
Should be fixed now

segiddins:
Rebased, ought to pass now (only failure was in trampolining, which is now gone)

segiddins:
Rebased again

segiddins:
@indirect r?

lib/bundler/compatibility_guard.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: false
+
+require "rubygems"
+require "bundler/version"
+
+if Bundler::VERSION.split(".").first.to_i >= 2
+  if Gem::Version.new(Object::RUBY_VERSION.dup) < Gem::Version.new("2.3")
+    abort "Bundler 2 requires Ruby 2.3 or later. Either install bundler 1 or update to a supported Ruby version."
```
hmistry: Suggest a less terse statement and flip the action order to show our first choice is upgrade ruby and the fallback plan is to go with v1.x.
`Bundler 2.x requires Ruby 2.3 or later. Please update your Ruby version to 2.3 or higher, or use the latest Bundler 1.x.`

lib/bundler/compatibility_guard.rb
```diff
@@ -0,0 +1,14 @@
...
+  end
+
+  if Gem::Version.new(Gem::VERSION.dup) < Gem::Version.new("2.5")
+    abort "Bundler 2 requires RubyGems 2.5 or later. Either install bundler 1 or update to a supported RubyGems version."
```
hmistry: `Bundler 2.x requires RubyGems 2.5 or later. Please update your RubyGems version to 2.5 or higher, or use the latest Bundler 1.x.`

spec/other/compatibility_guard_spec.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+RSpec.describe "bundler compatibility guard" do
+  context "when the bundler version is 2+" do
+    before { simulate_bundler_version "2.0.a" }
```
hmistry: Do you foresee the need of testing against various version schemes like Rails started off with SemVer and now it's 5.0.1.1 or 5.1.0.rc1 or 5.1.0.alpha and whatever other extensions they started adding on because they didn't feel like incrementing the version.

indirect:
@bundlerbot r+

lib/bundler/compatibility_guard.rb
hmistry: I think it would be good to have a note on Bundler 1.x to suggest it's on a limited maintenance path to encourage the investment to upgrade.

spec/support/helpers.rb
```diff
@@ -481,6 +481,14 @@ def simulate_bundler_version(version)
       ENV["BUNDLER_SPEC_VERSION"] = old if block_given?
     end
 
+    def simulate_rubygems_version(version)
+      old = ENV["BUNDLER_SPEC_RUBYGEMS_VERSION"]
+      ENV["BUNDLER_SPEC_RUBYGEMS_VERSION"] = version.to_s
+      yield if block_given?
+    ensure
+      ENV["BUNDLER_SPEC_RUBYGEMS_VERSION"] = old if block_given?
```
hmistry: If no block is given, will this still return the ENV variable to its previous version?

lib/bundler/compatibility_guard.rb
```diff
@@ -1,14 +1,14 @@
-# frozen_string_literal: true
+# frozen_string_literal: false
 
 require "rubygems"
 require "bundler/version"
 
 if Bundler::VERSION.split(".").first.to_i >= 2
-  if Gem::Version.new(Object::RUBY_VERSION) < Gem::Version.new("2.0.0")
-    abort "Bundler 2 requires Ruby 2+. Either install bundler 1 or update to a supported Ruby version."
+  if Gem::Version.new(Object::RUBY_VERSION) < Gem::Version.new("2.3")
```
nobu: `Object::` seems useless.
You might intend `::RUBY_VERSION`?

PR #5721, opened 2017-06-13
Add --shebang option to binstubs command

segiddins:
Ports https://github.com/bundler/bundler/pull/4070 to master.

bundle install `--binstubs` option is to be removed from Bundler 2.0 but
currently supports setting the shebang using the `--shebang` option.

See #1467 introducing the `--shebang` option, which with this commit
is ported to the binstubs command.


indirect:
@bundlerbot r+

PR #5722, opened 2017-06-14
[2.0] Add `—all` flag to `bundle update`

segiddins:

The problem was that some users would think they needed to run `bundle update` instead of `bundle install`, and we want such a (potentially) destructive action to require more explicit user approval.

My diagnosis was that `bundle update` shouldn't implicitly update _everything_ without an acknowledgement that it will do so.

My fix, based on https://github.com/bundler/bundler/pull/2646, it to require `bundle update --all` for the current `bundle update` behavior.

I chose this fix because it was already on 2-0-dev, courtesy of https://github.com/bundler/bundler-features/issues/18.


indirect:
@bundlerbot r+

PR #5724, opened 2017-06-14
[2.0] Add a feature flag for `bundle install` not `—cache`ing by default

segiddins:

The problem was that `bundle install` would default to always updating the cache in `vendor/bundle` (if it existed).

My diagnosis was that the default for the `--cache` flag should be `false` instead of `true`.

My fix is to invert the default on 2.0, using a feature flag. Taken from https://github.com/bundler/bundler/pull/3555.

I chose this fix because it allowed a feature flag for a breaking change in 2.0 and was minimally invasive.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -163,8 +163,8 @@ module Bundler
       "Specify the number of jobs to run in parallel"
     method_option "local", :type => :boolean, :banner =>
       "Do not attempt to fetch gems remotely and use the gem cache instead"
-    method_option "no-cache", :type => :boolean, :banner =>
-      "Don't update the existing gem cache."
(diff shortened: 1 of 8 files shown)


indirect:
r+

segiddins:
Rebased. Only 1.0 failure was due to trampolining, which is now gone.

segiddins:
This causes a few changes of behavior that we need to decide if we want to fix or change:

https://travis-ci.org/bundler/bundler/jobs/249329317#L407

segiddins:
Rebased.

segiddins:
@indirect should I close this since we got rid of the `--no-cache` option on 2.0 already?

segiddins:
@indirect ping?

indirect:
Now that the `--cache/--no-cache` flag is gone, let's drop this. Thanks!

PR #5725, opened 2017-06-14
Default to printing a help message when `bundle` is run without arguments on 2.0

segiddins:

The problem was that users unfamiliar with Bundler would run `bundle` and either install gems (if in a directory with a gemfile), or else get the error `Could not locate Gemfile`. You'd need to know to run `bundle help` or `bundle -h` to get the man page, and even that could be overwhelming (not to mention turfing you into a pager).

My diagnosis was that we ought to print some form of help when the bare `bundle` command is run, rather than defaulting to running `bundle install`.

This work was prompted by https://trello.com/c/OpuOdTZl/112-print-help-when-bundle-is-run-without-arguments, and is an improvement (rather than a straight port) on https://github.com/bundler/bundler/pull/3831.

My fix was to print the following help output when the CLI is invoked without arguments, contingent upon a feature flag.

```
Bundler version 1.15.1

Commands:
  bundle add GEM VERSION            # Add gem to Gemfile and run bundle install
  bundle binstubs GEM [OPTIONS]     # Install the binstubs of the listed gem
  bundle check [OPTIONS]            # Checks if the dependencies listed in Gemfile are satisfied by currently installed gems
  bundle clean [OPTIONS]            # Cleans up unused gems in your bundler directory
  bundle config NAME [VALUE]        # retrieve or set a configuration value
  bundle console [GROUP]            # Opens an IRB session with the bundle pre-loaded
  bundle doctor [OPTIONS]           # Checks the bundle for common problems
  bundle env                        # Print information about the environment Bundler is running under
  bundle exec [OPTIONS]             # Run the command in context of the bundle
  bundle gem GEM [OPTIONS]          # Creates a skeleton for creating a rubygem
  bundle help [COMMAND]             # Describe available commands or one specific command
  bundle help [COMMAND]             # Describe subcommands or one specific subcommand
  bundle info GEM [OPTIONS]         # Show information for the given gem
  bundle init [OPTIONS]             # Generates a Gemfile into the current working directory
  bundle inject GEM VERSION         # Add the named gem, with version requirements, to the resolved Gemfile
  bundle install PLUGINS            # Install the plugin from the source
  bundle install [OPTIONS]          # Install the current environment to the system
  bundle issue                      # Learn how to report an issue in Bundler
  bundle licenses                   # Prints the license of all gems in the bundle
  bundle lock                       # Creates a lockfile without installing
  bundle open GEM                   # Opens the source directory of the given bundled gem
  bundle outdated GEM [OPTIONS]     # list installed gems with newer versions available
  bundle package [OPTIONS]          # Locks and then caches all of the gems into vendor/cache
  bundle platform [OPTIONS]         # Displays platform compatibility information
  bundle plugin SUBCOMMAND ...ARGS  # manage the bundler plugins
  bundle pristine                   # Restores installed gems to pristine condition from files located in the gem cache. Gem installed from a git repository will be issued `git checkout --force`.
  bundle show GEM [OPTIONS]         # Shows all gems that are part of the bundle, or the path to a given gem
  bundle update [OPTIONS]           # update the current environment
  bundle version                    # Prints the bundler's version information
  bundle viz [OPTIONS]              # Generates a visual dependency graph

Options:
      [--no-color]                 # Disable colorization in output
  -r, [--retry=NUM]                # Specify the number of times you wish to attempt network commands
  -V, [--verbose], [--no-verbose]  # Enable verbose output mode
```

I chose this fix because it aligns the Bundler CLI with many other package managers that display a help message of some sort when invoked without arguments:

- CocoaPods
- npm
- RubyGems
- Homebrew
- pip
- Swift Package Manager
- Carthage
- Cargo

And unlike the following, which _do_ install by default:

- yarn

As to the particular format of the help printed, I thought something concise would be less disruptive than the full man page.

Improvements to the output would be welcome, this is just using the Thor default right now but we can easily implement our own. This will require test coverage, but I thought I'd hold off until the general change got some buy-in.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -55,7 +55,14 @@ module Bundler
     check_unknown_options!(:except => [:config, :exec])
     stop_on_unknown_option! :exec
 
-    default_task :install
+    desc "cli_help", "Prints a summary of bundler commands", :hide => true
+    def cli_help
+      version
+      Bundler.ui.info "\n"
+      self.class.help(shell)
+    end
+    default_task(Bundler.feature_flag.default_cli_command)
+
     class_option "no-color", :type => :boolean, :desc => "Disable colorization in output"
     class_option "retry",    :type => :numeric, :aliases => "-r", :banner => "NUM",
                              :desc => "Specify the number of times you wish to attempt network commands"
--- a/lib/bundler/feature_flag.rb
+++ b/lib/bundler/feature_flag.rb
@@ -5,12 +5,24 @@ module Bundler
       unless Bundler::Settings::BOOL_KEYS.include?(flag.to_s)
         raise "Cannot use `#{flag}` as a settings feature flag since it isn't a bool key"
       end
-      define_method("#{flag}?") do
-        value = Bundler.settings[flag]
+
+      settings_method("#{flag}?", flag, &default)
+    end
+    private_class_method :settings_flag
+
+    def self.settings_option(key, &default)
+      settings_method(key, key, &default)
+    end
+    private_class_method :settings_option
+
+    def self.settings_method(name, key, &default)
+      define_method(name) do
+        value = Bundler.settings[key]
         value = instance_eval(&default) if value.nil? && !default.nil?
         value
       end
     end
+    private_class_method :settings_method
 
     (1..10).each {|v| define_method("bundler_#{v}_mode?") { major_version >= v } }
 
@@ -20,6 +32,8 @@ module Bundler
     settings_flag(:error_on_stderr) { bundler_2_mode? }
     settings_flag(:init_gems_rb) { bundler_2_mode? }
 
+    settings_option(:default_cli_command) { bundler_2_mode? ? :cli_help : :install }
+
     def initialize(bundler_version)
       @bundler_version = Gem::Version.create(bundler_version)
     end
@@ -28,7 +42,5 @@ module Bundler
       @bundler_version.segments.first
     end
     private :major_version
-
-    class << self; private :settings_flag; end
   end
 end
--- a/spec/quality_spec.rb
+++ b/spec/quality_spec.rb
@@ -169,6 +169,7 @@ RSpec.describe "The library itself" do
 
   it "documents all used settings" do
     exemptions = %w[
+      default_cli_command
       gem.coc
       gem.mit
       warned_version


indirect:
I like this. As further evidence on the "should display help" side, most unix utilities (grep, etc) also print their help when they are run without arguments.

chrismo:
my 2 cents would be to leave this on `install`. Personally it'd be a bother to me to have to alias or just get used to typing bundle install each time.

indirect:
@chrismo I have gone back and forth on this for that reason... fwiw, `bundle i` runs install, and will continue to on 2.0. Maybe you can start adjusting your muscle memory ahead of time. 😅 Alternately, you could stay on Bundler 1.x forever. 😆

My personal shell aliases are already `b=bundle exec` and `bi=bundle install`, but I may run those commands more than the average person. 😝

hmistry:
I'm with @chrismo on this. "We've" trained people to use `bundle` as a shortcut to `bundle install` and to now tell them sorry that's `bundle help` will be cause a lot of unhappiness. I'm for change as long as you can convince me that this is going to be better than the existing system. 

In this case, I ask myself why is this change better, and my answer is to help newbies who don't know how to use it for the first few days. This is not the best reason in my opinion. Some learning curves are acceptable and in this case having to learn that `bundle` is shortcut to `bundle install` is acceptable to me because it's not overly confusing or convoluted and once you're proficient it's quite  beneficial to you.

As far as being consistent with other package managers, well then I'd ask are all the other commands consistent with other package managers? If yes, then ok. If not, then it weakens your selling point as to why this change is better (to me anyway).

So other than the above reasons, I ask why is this change really beneficial to bundler users? Is there a huge issue with newbies getting confused that outweighs the benefit to those who are proficient? Is it so important to be consistent with other package managers because all the other commands of bundler are consistent? Is there a future benefit where this is a small step to that direction that will make all this change so much more worthwhile in hindsight?

These are all my thoughts and opinions on seeing this and having strong compelling reasons to the above questions help me decide on features otherwise I'd be catering to everyone. @indirect and @segiddins, I'm sure you've already thought through all of this and whichever way you decide, it'll be ok. Best of luck. 😃

indirect:
@hmistry in the end, I think both OSS in general (and Bundler in this particular case) make things easier for experienced devs at the cost of new or interested devs. I believe that status quo is actively bad: experienced devs can either run `bundle i` or create a shell alias. Inexperienced devs who are confused simply give up, and we lose everything they could have contributed. That, combined with the standards set by both other CLI utilities and other package managers, is why we're going to change this for 2.0.

indirect:
@bundlerbot r+

PR #5727, opened 2017-06-15
Add a feature flag for `bundle update —source NAME` not unlocking a gem with that name

segiddins:

The problem is that `bundle update --source NAME` will unlock a _gem_ with that name, in addition to the source.

My diagnosis was that we unlocked based on `spec.name` instead of `spec.source.name`.

My fix is to put this backwards-compatibility hack behind a feature flag that will turn off on 2.0.

I chose this fix because it allows the current behavior to continue for those who depend upon it, but the hack will be disabled by default on 2.0, or when `unlock_source_unlocks_spec` is set to true.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -748,6 +748,8 @@ module Bundler
         end
       end
 
+      unlock_source_unlocks_spec = Bundler.feature_flag.unlock_source_unlocks_spec?
+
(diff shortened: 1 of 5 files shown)


indirect:
@bundlerbot r+

segiddins:
@bundlerbot r=indirect

PR #5728, opened 2017-06-15
Create Gemfiles with an HTTPS github source defined

segiddins:

The problem was that people are creating new Gemfiles that use the built-in `github` git source, which is being removed in 2.0. Additionally, it does _not_ use an encrypted connection to GitHub.

My diagnosis was that we can't change the default because of backwards compatibility, but we can encourage _new_ Gemfiles to "do the right thing".

My fix is to add our new, recommended definition of the shortcut to all bundler-generated gemfiles.

I chose this fix because it will only affect new Gemfiles.

### DIFF
--- a/lib/bundler/templates/Gemfile
+++ b/lib/bundler/templates/Gemfile
@@ -1,4 +1,6 @@
 # frozen_string_literal: true
 source "https://rubygems.org"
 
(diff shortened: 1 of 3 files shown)


indirect:
@bundlerbot r+

PR #5782, opened 2017-06-17
[2.0] Add a global cache for downloaded .gem files

segiddins:

The problem was that bundler would need to download `foo-1.0.gem` files from a RubyGems server for each different ruby version installed on a user's machine. It also meant that people installing into a per-app path would need to re-download every gem for that bundle an additional time. This adds up, and makes `bundle install` slower than it needs to be.

My diagnosis was that Bundler could keep a (per-source) cache of these `.gem` files, and pull from that cache instead of hitting the network whenever possible.

My fix implements said cache, in a very similar way to the compact index cache (same cache slug per remote strategy, etc). This largely comes from https://github.com/bundler/bundler/pull/3983.

I chose this fix because it is safe when used from multi-source gemfiles, it is easy to clear (`rm -rf bundle cache`), and it minimally interferes with the existing installation process.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -216,8 +216,8 @@ module Bundler
     end
 
     def app_config_path
-      if ENV["BUNDLE_APP_CONFIG"]
-        Pathname.new(ENV["BUNDLE_APP_CONFIG"]).expand_path(root)
+      if app_config = ENV["BUNDLE_APP_CONFIG"]
+        Pathname.new(app_config).expand_path(root)
       else
         root.join(".bundle")
       end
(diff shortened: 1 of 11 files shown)


indirect:
This is looking great to me. Very excited about using `--path` but not redownloading all the gems every time. 👍

segiddins:
Green. Is there anything else I missed in https://github.com/bundler/bundler/pull/3983 that's relevant? (I think there were some commits related to other changes that made that diff needlessly large)

indirect:
@bundlerbot r+

indirect:
(gonna export the env for this and run from master on my machine until this is released, now)

segiddins:
@indirect `bundle config --global global_gem_cache true` should also do the trick

segiddins:
Ugh has this timed out on CI?

segiddins:
@bundlerbot retry

segiddins:
@bundlerbot retry

segiddins:

PR #5783, opened 2017-06-17
[EnvironmentPreserver] Allow preserving MANPATH

### LINKED ISSUE #5730: Bundler throws an exception if you have MANPATH set in your environment.
Bundler master fails when you have MANPATH set in your environment:

e.g.
```shell
~/src/github.com/Shopify/shopify(master*) ➜ echo $MANPATH
/Users/juliannadeau/.nvm/versions/node/v6.10.0/share/man:
```

Stack Trace:
```shell
/Users/juliannadeau/src/github.com/bundler/bundler/lib/bundler/shared_helpers.rb:213:in `set_env': new key MANPATH (ArgumentError)
	from /Users/juliannadeau/src/github.com/bundler/bundler/lib/bundler/runtime.rb:272:in `setup_manpath'
	from /Users/juliannadeau/src/github.com/bundler/bundler/lib/bundler/runtime.rb:47:in `setup'
	from /Users/juliannadeau/src/github.com/bundler/bundler/lib/bundler.rb:101:in `setup'
	from /Users/juliannadeau/src/github.com/bundler/bundler/lib/bundler/setup.rb:10:in `<top (required)>'
	from /opt/rubies/2.3.3/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
	from /opt/rubies/2.3.3/lib/ruby/site_ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55:in `require'
	from /Users/juliannadeau/src/github.com/Shopify/shopify/config/boot.rb:32:in `block in <top (required)>'
	from /Users/juliannadeau/src/github.com/Shopify/shopify/lib/boot_lib/timing_helpers.rb:12:in `_t'
	from /Users/juliannadeau/src/github.com/Shopify/shopify/config/boot.rb:31:in `<top (required)>'
	from bin/rake:2:in `require_relative'
	from bin/rake:2:in `<main>'
```

[Adding MANPATH to the environment_preserver.rb's BUNDLER_KEYS](https://github.com/bundler/bundler/blob/master/lib/bundler/environment_preserver.rb#L5) seems to solve it. I'm not familiar with this functionality to know if this is the right call - but looking at the code, it looks like MANPATH was forgotten in the list.

This does raise a question for me - will bundler crash if I clash with another env var now? If I set another environment variable with the same name as Bundler's will it throw the same exception?

cc @segiddins

segiddins:

The problem was Bundler could throw an exception during setup when a gem contained manpages we were adding to `$MANPATH`.

Fixes #5730.

My diagnosis was that `$MANPATH` wasn't whitelisted as a key we preserved / were 'allowed' to override via `SharedHelpers.set_env`.

My fix is to add `$MANPATH` to that list of keys, as @jules2689 suggested. I also added some test coverage around us setting the proper man path.

I chose this fix because it treats that env var in the same way as others bundler sets.

### DIFF
--- a/lib/bundler/environment_preserver.rb
+++ b/lib/bundler/environment_preserver.rb
@@ -9,6 +9,7 @@ module Bundler
       BUNDLER_VERSION
       GEM_HOME
       GEM_PATH
+      MANPATH
       PATH
       RUBYLIB
       RUBYOPT
--- a/spec/runtime/setup_spec.rb
+++ b/spec/runtime/setup_spec.rb
@@ -740,6 +740,44 @@ end
     expect(err).to lack_errors
   end
 
+  describe "$MANPATH" do
+    before do
+      build_repo4 do
+        build_gem "with_man" do |s|
+          s.write("man/man1/page.1", "MANPAGE")
+        end
+      end
+    end
+
+    context "when the user has one set" do
+      before { ENV["MANPATH"] = "/foo:" }
+
+      it "adds the gem's man dir to the MANPATH" do
+        install_gemfile! <<-G
+          source "file:#{gem_repo4}"
+          gem "with_man"
+        G
+
+        run! "puts ENV['MANPATH']"
+        expect(out).to eq("#{system_gem_path("gems/with_man-1.0/man")}:/foo")
+      end
+    end
+
+    context "when the user does not have one set" do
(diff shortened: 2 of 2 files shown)


segiddins:
@bundlerbot r+

PR #5785, opened 2017-06-18
fix a missing space in the group conflict error message

colby-swandale:

There is a typo in the error message that is printed to users when a user specifies a set of conflicting groups when running bundle install. There is a missing space after the first sentence. 
```
› bundle install --with foo --without foo
You can't list a group in both, --with and --without.The offending groups are: foo.
```

execute `bundle install --with foo --without foo`


segiddins:
Thanks! @bundlerbot r+

Issue #4333, opened 2016-03-04
Respect XDG basedir spec for global configuration file


jasonkarns:
To conform to the XDG basedir spec, bundler's config file ought to be stored at `${XDG_CONFIG_HOME:-$HOME/.config}/bundler/config`

I'm not completely sure what other files are stored under `~/.bundle`, but there are other XDG-defined directories for cache and runtime files, if so needed.

https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

segiddins:
a) that's a breaking change
b) dumb question but what does XDG stand for, and what makes it the authoritative spec for how command line apps store things on disk?

jasonkarns:
doesn't need to be a breaking change. most tools that have migrated do so by first checking if `~/.my-app` exists and using it if so. If it doesn't exist, then the XDG settings are honored. This way people can opt-in by moving the files themselves (or removing the old ones). Some tools go a step further and migrate the files automatically.

XDG is a group that works on interoperability between *nix systems. The basedir spec is just one aspect of the larger focus of the group. The primary benefit to users is: standardized configurability by users/sysadmins across apps. (And it falls in lines with Apple guidelines on not polluting userspace)

git is the largest project that I'm aware of that conforms. keybase does as well. I've seen many others but can't recall them offhand.

gioele:
The problem is not limited to the configuration file, but also cache files.

Cache files should be stored at

    ${$XDG_CACHE_HOME:-$HOME/.cache}/bundler

instead of `~/.bundler/cache`.

If we need to appeal to authority, we can add Firefox, Thunderbird, Chromium and Inkscape to the list of apps that follow the XDG Basedir spec and do not clutter our lovely $HOMEs. :)

LynnCo:
Doesn't seem like anyone on the team is too excited about implementing this, @jasonkarns or @gioele would you want to work on a PR? (we aren't guaranteed to merge it, but it would help make your case)

valeth:
Is anybody already working on this?

segiddins:
@valeth not at the moment!

jasonkarns:
I can give it a shot soon.

colby-swandale:
This has been added in #6024

PR #5787, opened 2017-06-18
Use XDG directory for cache

valeth:

Bundler was storing user cache in the user's home directory.
Issue: #4333 

Store cache in XDG_CACHE_HOME if ~/.bundle/cache does not exist.
Otherwise continue to use the legacy directory.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -181,6 +181,12 @@ module Bundler
       raise e.exception("#{warning}\nBundler also failed to create a temporary home directory at `#{path}':\n#{e}")
     end
 
+    def xdg_home(type, alt)
+      xdg_path = Pathname.new(ENV.fetch("XDG_#{type}_HOME", alt)).join("bundler")
+      xdg_path.mkpath unless xdg_path.exist?
+      xdg_path
+    end
+
     def user_bundle_path
       Pathname.new(user_home).join(".bundle")
     end
@@ -202,7 +208,12 @@ module Bundler
     end
 
     def user_cache
-      user_bundle_path.join("cache")
+      legacy_path = user_bundle_path.join("cache")
+      if legacy_path.exist?
+        legacy_path
+      else
+        xdg_home("CACHE", user_home.join(".cache"))
+      end
     end
 
     def root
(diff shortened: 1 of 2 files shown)


colby-swandale:
Thanks for opening a PR and helping make Bundler better 🎉. This PR is missing tests, we will need some before we considering merging this.

valeth:
Added tests for `xdg_home` and `user_cache` methods.

Still not that experienced with rspec, so any improvements are welcome.

indirect:
@valeth can you provide your reasoning for why this PR is an improvement?

Today, all possible state that Bundler could write fall either into an application at `.bundle` or the user's home directory at `.bundle`. The config files are stored completely symmetrically, at `.bundle/config`. Cache created by Bundler is easy to find and easy to delete if needed, thanks to everything being inside the same directory.

Why would using XDG directories to scatter Bundler-created files across many directories be an improvement?

colby-swandale:
ping @valeth

valeth:
@indirect 
I think it is an improvement because it keeps the user's cache files in one place. So if the use wants to delete all cached data for his applications he just needs to delete one single directory, instead of digging through many directories for every program he wants to clear the cache for.

It would also make backing up configurations easier (if this would be expanded to also handle the configuration file), because you only need to backup one directory to save all your configuration.

Git uses a similar approach, they use the file at `$XDG_CONFIG_HOME/git/config` if it exists or look for `~/.gitconfig` otherwise.

indirect:
@valeth I can understand the argument, but I don't think I'm convinced by it. Very few popular open-source utilities support XDG for either cache or config. Very few users set `$XDG_CONFIG_HOME`, and very few users ask for XDG support. In the 7 years since Bundler 1.0 was released, I only remember 2-3 requests for XDG support, which makes me think this feature will add complexity for everyone while benefiting almost no one.

I'm open to the possibility of allowing the `user_bundle_path` to be configured via an environment variable, but I don't think XDG-specific support is a good use of our extremely limited maintenance time.

valeth:
Making the bundle path configurable would be good enough for me, as long as I can stash the files somewhere that isn't my home directory I'm fine with it.

PR #5788, opened 2017-06-19
[2.0] Allow conflicts between bundler dependencies and the current bundler version

segiddins:

The problem will occur once 2.0 is released -- there are many gems that declare a `s.dependency "bundler", "~> 1.x"` dependency on bundler. This would essentially make them inoperable with Bundler 2. We want to make adopting 2.0 as pain-free as possible, so hard conflicts are a no-go.

My diagnosis was that we need a way to allow conflicts on the bundler dependency once 2.0 comes out, so people can use 2.0.

My fix introduces a feature flag for allowing these conflicts (enabled on 2.0), which when enabled will ignore bundler dependencies during resolution, and warn in the installer when a conflict would've occurred.

I chose this fix because it has the minimal performance penalty with the feature flag enabled. This is mostly a port of https://github.com/bundler/bundler/pull/3871 to master.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -567,10 +567,7 @@ module Bundler
     end
 
     def pretty_dep(dep, source = false)
-      msg = String.new(dep.name)
-      msg << " (#{dep.requirement})" unless dep.requirement == Gem::Requirement.default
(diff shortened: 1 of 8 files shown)


segiddins:
@indirect r?
This is now a blocker for me

indirect:
@bundlerbot r+

PR #5790, opened 2017-06-19
[2.0] Suppress `Using …` messages during installation when a version has not changed

segiddins:

The problem was that `bundle install` output can get very verbose, even when Bundler is not doing anything.

See https://github.com/bundler/bundler-features/issues/33.

My diagnosis was that bundler was printing a bunch of `Using name (version)` messages, even when we were already using that gem at the same version.

My fix is to introduce a feature flag (enabled by default on 2.0), that will only print those extra `Using` messages when `--verbose` is passed, and will continue to print them when there was an old version we can tell users about. Note that we still print a message when installing a gem for the first time.

I chose this fix because it was essentially what had been done in https://github.com/bundler/bundler/pull/3872, and allows for easy feature-flagging.


segiddins:
Rebased

segiddins:
2.0 specs updated, please review

indirect:
I really like using Bundler more after this change, :100: @bundlerbot r+

PR #5791, opened 2017-06-20
[CLI] Dont print defaults in the command printed with --verbose

segiddins:

The problem was that running `bundle lock --verbose`  would print 

    Running `bundle lock    --add-platform  --remove-platform  --verbose` with bundler 1.15.1

even though the platform flags were never specified. This was surfaced by https://github.com/bundler/bundler/pull/5724, which will be fixed by this PR.

My diagnosis was that default values of options were being printed.

My fix is to filter out the default values before getting Thor 

I chose this fix because there is no way, within the `CLI` instance, to tell which flags the user has explicitly given, so I felt that filtering out those options that had default values was the next-best thing.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -621,10 +621,16 @@ module Bundler
     def print_command
       return unless ENV["BUNDLE_POSTIT_TRAMPOLINING_VERSION"] || Bundler.ui.debug?
       _, _, config = @_initializer
-      current_command = config[:current_command].name
-      return if %w[exec version check platform show help].include?(current_command)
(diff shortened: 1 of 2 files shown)


colby-swandale:
@bundlerbot r+

segiddins:
I guess we need to sort the options so the output is consistent on 1.8.7

colby-swandale:
Yea 👍 , sorry i approved this prematurely.

segiddins:
No worries! It's fine to approve stuff even if it doesn't end up passing, that's what CI is for

segiddins:
@bundlerbot r=colby-swandale

PR #4930, opened 2016-08-29
[RubyGems] Stop fetching all specs 😁

segiddins:
aka boolean logic is hard

PR #5792, opened 2017-06-20
[2.0] Remove RubyGems Aggregate & support transitive source pinning

### LINKED ISSUE #3671: Bundler issues warning when a gem is on both the global source and a one-off source

segiddins:

The problem was that the resolver could resolve specs from _any_ of the sources specified in the Gemfile, even if that source had nothing to do with the spec in question. This was such a large security vulnerability that, when discovered, it warranted a CVE and its own minor release of Bundler.

Closes #3671.
Closes #3696.
Closes #4059.

My diagnosis was that we needed to get rid of the notion of a `rubygems aggregate` and enforce that specs could only come either from the source they were declared to come from (the top-level source if declared at the top-level of the Gemfile, else a scoped source), or a source that it transitively "inherited" from the gems that required it.

My fix is to disable multiple top-level sources in the Gemfile, remove the RubyGems aggregate, and filter the sources gems could come from as described above.

I chose this fix because it allows doing the filtering in a reasonably performant manner, and refactors the way we handle sources to abstract some of the grossness in such a way that the machinations to make sure that all of the necessary gem info is downloaded is encapsulated into a single method, driven from the definition, rather than being specific to rubygems sources.

See https://github.com/bundler/bundler/pull/4714 and https://github.com/bundler/bundler/pull/4930 for the prior implementation.

### DIFF
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -415,6 +415,12 @@ module Bundler
         next if name == "bundler"
         next unless search_for(requirement).empty?
 
+        cache_message = begin
+                            " or in gems cached in #{Bundler.settings.app_cache_path}" if Bundler.app_cache.exist?
+                          rescue GemfileNotFound
+                            nil
+                          end
+
         if (base = @base[name]) && !base.empty?
           version = base.first.version
           message = "You have requested:\n" \
@@ -426,18 +432,13 @@ module Bundler
         elsif source = @source_requirements[name]
           specs = source.specs[name]
           versions_with_platforms = specs.map {|s| [s.version, s.platform] }
-          message = String.new("Could not find gem '#{requirement}' in #{source}.\n")
+          message = String.new("Could not find gem '#{requirement}' in #{source}#{cache_message}.\n")
           message << if versions_with_platforms.any?
                        "The source contains '#{name}' at: #{formatted_versions_with_platforms(versions_with_platforms)}"
                      else
                        "The source does not contain any versions of '#{requirement}'"
                      end
         else
-          cache_message = begin
-                            " or in gems cached in #{Bundler.settings.app_cache_path}" if Bundler.app_cache.exist?
-                          rescue GemfileNotFound
-                            nil
-                          end
           message = "Could not find gem '#{requirement}' in any of the gem sources " \
             "listed in your Gemfile#{cache_message}."
         end
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -59,12 +59,31 @@ module Bundler
             o << %(the gems in your Gemfile, which may resolve the conflict.\n)
           elsif !conflict.existing
             o << "\n"
-            if conflict.requirement_trees.first.size > 1
-              o << "Could not find gem '#{conflict.requirement}', which is required by "
-              o << "gem '#{conflict.requirement_trees.first[-2]}', in any of the sources."
+
+            relevant_sources = if conflict.requirement.source
+              [conflict.requirement.source]
+            elsif conflict.requirement.all_sources
+              conflict.requirement.all_sources
+            elsif Bundler.feature_flag.lockfile_uses_separate_rubygems_sources?
+              # every conflict should have an explicit group of sources when we
+              # enforce strict pinning
+              raise "no source set for #{conflict}"
             else
-              o << "Could not find gem '#{conflict.requirement}' in any of the sources\n"
+              []
+            end.compact.map(&:to_s).uniq.sort
+
+            if conflict.requirement_trees.first.size > 1
+              o << "Could not find gem '#{printable_dep(conflict.requirement)}', which is required by "
+              o << "gem '#{printable_dep(conflict.requirement_trees.first[-2])}', "
+            else
+              o << "Could not find gem '#{printable_dep(conflict.requirement)}' "
             end
+
+            o << if relevant_sources.empty?
+                   "in any of the sources.\n"
+                 else
+                   "in any of the relevant sources:\n  #{relevant_sources * "\n  "}\n"
+                 end
           end
           o
         end.strip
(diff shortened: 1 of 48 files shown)


attritionorg:
@segiddins "it warranted a CVE and its own minor release of Bundler" I know the CVE, but don't think I ever saw the versions impacted and the first fixing version. Can you provide that info? Thanks!

segiddins:
> but don't think I ever saw the versions impacted and the first fixing version

https://bundler.io/v1.7/whats_new.html

indirect:
@bundlerbot r+

segiddins:
@bundlerbot r-
I'd like to get the specs running on 2.0 mode first, if that's ok with you? Since I have a hunch this change in particular might break tests

indirect:
👍

segiddins:
There's an off-chance this will pass now, so please review :)

lib/bundler/resolver.rb
```diff
@@ -59,12 +59,31 @@ def message
...
-              o << "Could not find gem '#{conflict.requirement}', which is required by "
-              o << "gem '#{conflict.requirement_trees.first[-2]}', in any of the sources."
+              o << "Could not find gem '#{printable_dep(conflict.requirement)}', which is required by "
+              o << "gem '#{printable_dep(conflict.requirement_trees.first[-2])}', "
             else
-              o << "Could not find gem '#{conflict.requirement}' in any of the sources\n"
+              o << "Could not find gem '#{printable_dep(conflict.requirement)}' "
             end
+
+            o << if relevant_sources.empty?
```
indirect: please append inside the branches rather than indenting the whole conditional 👍🏻

segiddins:
Passing for Bundler 1, down to 5 failures for Bundler 2 (I don't know how I keep on missing them locally, I'm sorry)

colby-swandale:
nothing to be sorry about 😄

segiddins:
@indirect r?

segiddins:
Have two 👍, so going to merge this.

@bundlerbot r+

Benjamin-Dobell:
How do we now specify the preferred source for dependencies of a Gem in our `Gemfile`?

e.g.

We have a dependency on `some-gem`, but `some-gem` presently has several dependencies, one of which is `private-gem`. How do I ensure that Bundler correctly gets `private-gem` from my private Gem repository and not Rubygems?

I tried:

```
source 'https://rubygems.org'

source 'https://ourprivaterepository' do
  gem 'some-gem', git: 'https://ourforkongithub'
end
```

However, `private-gem` is still coming down from Rubygem.

My dependencies are now effectively being poisoned by a `private-gem` version appearing in Rubygems, where as this was not possible in past when I simply specified Rubygem and our private repository as top-level sources - our private repository always had priority.

Is there syntax to rectify this?

I definitely don't want an explicit dependency on `private-gem`, it's not used directly and may disappear as a dependency of `some-gem` in the future.

PR #4034, opened 2015-10-04
Remove console command

Zorbash:
Resolves: #4025

Issue #4025, opened 2015-10-01
Deprecate bundle console command


Zorbash:
In 748d8ec it is stated that:
> Bundler 2.0 will deprecate
the `bundle console` command, and instead add a script to each generated
gem that loads `bundler/setup` and then loads the gem.

Currently there is no relevant issue in the [`2.0`](https://github.com/bundler/bundler/issues?utf8=%E2%9C%93&q=milestone%3A2.0) milestone, so i'm suggesting that we add one if it is still part of the [`2.0`](https://github.com/bundler/bundler/issues?utf8=%E2%9C%93&q=milestone%3A2.0) roadmap.

indirect:
Yeah, good catch! The console command has gotten many, many pull requests for different kinds of customization and so we eventually realized that an editable bin/console is the only way to provide that level of customization. The gem command already creates bin/console, so all that's left is to remove the console command. For bonus points, you could add a deprecation warning to the console command in the 1-99 branch. :)

Zorbash:
Great! I'm preparing a PR for this.

agis-:
This is done (`bundle-console` is deprecated and removed in bundler 2.0).

PR #5793, opened 2017-06-20
[2.0] Remove the console command

segiddins:

> The console command has gotten many, many pull requests for different kinds of customization and so we eventually realized that an editable bin/console is the only way to provide that level of customization. The gem command already creates bin/console, so all that's left is to remove the console command.

From https://github.com/bundler/bundler/issues/4025.

My diagnosis was `bundle console` needed to go.

My fix removes the console command behind a feature flag, enabled by default on 2.0.

See https://github.com/bundler/bundler/pull/4034.

I chose this fix because it didn't require touching any of the console logic and would give the proper missing command error on 2.0.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -395,11 +395,12 @@ module Bundler
       Open.new(options, name).run
     end
 
-    desc "console [GROUP]", "Opens an IRB session with the bundle pre-loaded"
-    def console(group = nil)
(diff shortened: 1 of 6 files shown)


indirect:
@bundlerbot r+

segiddins:

colby-swandale:
@bundlerbot retry

PR #5795, opened 2017-06-20
Avoid warnings in the specs on Ruby 1.8.7

segiddins:

The problem was the specs had method redefinition warnings on 1.8.7.

My diagnosis was the `mkpath` hack was causing some, as well as us requiring support files by their absolute paths.

My fix requires support files by relative paths and updates rspec to 3.6, where the `mkpath` hack is unnecessary.

I chose this fix because it works on 1.8.7 without any adverse effects on modern rubies.


segiddins:
@bundlerbot r+

segiddins:
@bundlerbot retry

segiddins:
@bundlerbot retry

PR #4008, opened 2015-09-21
`bundle-package` is the new `bundle-cache`

agis-:
This PR removes the obsolete `bundle cache(1)` and renames `bundle package(1)` to `bundle cache(1)`.

PR #5798, opened 2017-06-20
Use the Package command for `bundle cache` on Bundler 2

segiddins:

The problem was the current `bundle cache` is obsolete.

My diagnosis was we want `bundle package` to be `bundle cache`, but can't change the `cache` behavior due to backwards compatibility.

My fix is to redirect `bundle cache` to `bundle package` on Bundler 2.

I chose this fix because it implements https://github.com/bundler/bundler/pull/4008 without breaking the current `bundle cache` command on master.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -324,16 +324,20 @@ module Bundler
       Outdated.new(options, gems).run
     end
 
-    desc "cache [OPTIONS]", "Cache all the gems to vendor/cache", :hide => true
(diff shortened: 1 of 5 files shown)


segiddins:
Rebased, specs should now reliably pass rather than being indeterminate on Ruby < 2

segiddins:
Rebased

indirect:
Ha. The implementation made me laugh, I wouldn't have thought of that. 😅 @bundlerbot r+

segiddins:
@bundlerbot retry

PR #5800, opened 2017-06-20
[2.0] Update the specs to pass under Bundler 2

segiddins:

The problem was we have all these _amazing_ Bundler 2.0 features hidden behind feature flags. But we weren't testing all of bundler in that 2.0 mode.

My diagnosis was we needed to get the bundler 2 specs running, and passing!

My fix is to add a travis build entry to change `version.rb` to a 2.0 version and run the tests!

I chose this fix because it will completely imitate what happens once we change the version on `master`, and by keeping the test suite passing on both 1.0 and 2.0 modes, we'll be in a position to release a 1.16 to which we'll be able to (relatively easily) backport fixes that land after master switches to completely target 2.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: ruby
 script: rake spec:travis
 before_script:
   - travis_retry rake spec:travis:deps
+  - travis_retry rake spec:travis:sub_version
   - travis_retry rake man:build
   - travis_retry rake spec:rubygems:clone_rubygems_$RGV
 
@@ -43,6 +44,8 @@ env:
 
 matrix:
   include:
+    - rvm: 2.4.1
+      env: RGV=v2.6.8 BUNDLER_SPEC_SUB_VERSION=2.0.0
     # Ruby 2.4, Rubygems 2.6.8 and up
     # Ruby 2.3, Rubygems 2.5.1 and up
     - rvm: 2.2.6
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -54,7 +54,8 @@ module Bundler
     def initialize(lockfile, dependencies, sources, unlock, ruby_version = nil, optional_groups = [])
       @unlocking = unlock == true || !unlock.empty?
 
-      @dependencies    = dependencies
+      @dependencies = dependencies
+      @dependencies_by_name = dependencies.group_by(&:name)
       @sources         = sources
       @unlock          = unlock
       @optional_groups = optional_groups
@@ -953,8 +954,10 @@ module Bundler
     def additional_base_requirements_for_resolve
       return [] unless @locked_gems && Bundler.feature_flag.only_update_to_newer_versions?
       @locked_gems.specs.reduce({}) do |requirements, locked_spec|
-        dep = Gem::Dependency.new(locked_spec.name, ">= #{locked_spec.version}")
-        requirements[locked_spec.name] = DepProxy.new(dep, locked_spec.platform)
+        name = locked_spec.name
+        next requirements if @locked_deps[name] != @dependencies_by_name[name]
+        dep = Gem::Dependency.new(name, ">= #{locked_spec.version}")
+        requirements[name] = DepProxy.new(dep, locked_spec.platform)
         requirements
       end.values
     end
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -54,8 +54,7 @@ module Bundler
     def initialize(lockfile, dependencies, sources, unlock, ruby_version = nil, optional_groups = [])
       @unlocking = unlock == true || !unlock.empty?
 
-      @dependencies = dependencies
-      @dependencies_by_name = dependencies.group_by(&:name)
+      @dependencies    = dependencies
       @sources         = sources
       @unlock          = unlock
       @optional_groups = optional_groups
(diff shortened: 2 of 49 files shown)


segiddins:
If this now passes, I'll restore all the other build matrix entries. Please review?

segiddins:
Only thing failing on 2.0 are the real world specs 🎆

segiddins:
Blocked by https://github.com/bundler/bundler/pull/5788

.travis.yml
```diff
@@ -26,97 +27,15 @@ addons:
 
 rvm:
   - 2.4.1
-  - 2.3.4
-  - 2.2.7
-  - 2.1.10
-  - 2.0.0
```
jules2689: This is the version that ships with Mac, would it make sense to keep this right now?

segiddins: Please see the PR description

> Don't worry, I'll rebase out the deletion of all the build entries once this is ready to merge.

lib/bundler/definition.rb
```diff
@@ -54,7 +54,8 @@ def self.build(gemfile, lockfile, unlock)
     def initialize(lockfile, dependencies, sources, unlock, ruby_version = nil, optional_groups = [])
       @unlocking = unlock == true || !unlock.empty?
 
-      @dependencies    = dependencies
+      @dependencies = dependencies
+      @dependencies_by_name = dependencies.group_by(&:name)
```
jules2689: I'm a little concerned if this would introduce a boot time regression. I can take a look in a bit and test it out

lib/bundler/definition.rb
```diff
@@ -953,8 +954,10 @@ def compute_requires
     def additional_base_requirements_for_resolve
       return [] unless @locked_gems && Bundler.feature_flag.only_update_to_newer_versions?
       @locked_gems.specs.reduce({}) do |requirements, locked_spec|
-        dep = Gem::Dependency.new(locked_spec.name, ">= #{locked_spec.version}")
-        requirements[locked_spec.name] = DepProxy.new(dep, locked_spec.platform)
+        name = locked_spec.name
+        next requirements if @locked_deps[name] != @dependencies_by_name[name]
```
jules2689: `next` statements are my favourite statements, only superseded by `break` since these mean less work ;P

segiddins: but `break` here would have entirely different semantics, as this is a `reduce`

lib/bundler/definition.rb
segiddins: Agree it's not ideal, but I can't think of how else to fix the `only_update_to_newer_versions` bug

segiddins: I guess I could move this to `additional_base_requirements_for_resolve` so that we don't hit it in `Bundler.setup` ?

lib/bundler/definition.rb
```diff
@@ -54,7 +54,7 @@ def self.build(gemfile, lockfile, unlock)
     def initialize(lockfile, dependencies, sources, unlock, ruby_version = nil, optional_groups = [])
       @unlocking = unlock == true || !unlock.empty?
 
-      @dependencies    = dependencies
+      @dependencies = dependencies
```
andremedeiros: Can you fix the alignment here?

lib/bundler/source/git.rb
```diff
@@ -304,9 +304,9 @@ def git_proxy
 
       def fetch
         git_proxy.checkout
-      rescue GitError
+      rescue GitError => e
         raise unless Bundler.feature_flag.allow_offline_install?
-        Bundler.ui.warn "Using cached git data because of network errors"
+        Bundler.ui.warn "Using cached git data because of network errors\n#{e}"
```
andremedeiros: Maybe add a colon after `errors` to suggest the exception error contains details.

spec/support/helpers.rb
```diff
@@ -173,24 +189,20 @@ def gem_command(command, args = "", options = {})
...
+
       Open3.popen3(cmd.to_s) do |stdin, stdout, stderr, wait_thr|
         yield stdin, stdout, wait_thr if block_given?
         stdin.close
 
-        @exitstatus = wait_thr && wait_thr.value.exitstatus
-        @out = Thread.new { stdout.read }.value.strip
-        @err = Thread.new { stderr.read }.value.strip
+        command_execution.exitstatus = wait_thr && wait_thr.value.exitstatus
+        command_execution.stdout = Thread.new { stdout.read }.value.strip
```
andremedeiros: Don't we want to `.join` the threads here?

andremedeiros:
@bundlerbot r+

spec/support/helpers.rb
segiddins: `value` joins the thread

lib/bundler/definition.rb
segiddins: ugh yeah, rubocop should've caught that :/ (came from reverting another change)

spec/support/helpers.rb
andremedeiros: TIL

lib/bundler/source/git.rb
segiddins: 👍

segiddins:
@andremedeiros I haven't addressed your comments yet... (It's also failing on 1.8.7 because it doesn't seem to support `\h`)

andremedeiros:
@segiddins my bad. Sorry 😢

segiddins:
Rebased, added a PR description.

segiddins:
No worries, just paranoid about pegging our travis VMs (earlier last week stuff was taking half a day to run)

segiddins:
Its. 🍏. 🎆 

@bundlerbot r+

lib/bundler/definition.rb
jules2689: @segiddins I was just making a statement that `next` statements are one of my favourite because they mean less work == often more boot time perf. Was a stupid joke :)

.travis.yml
jules2689: _woops_

Issue #5781, opened 2017-06-16
Bundler not working with Artifactory


jakauppila:
We are looking to proxy RubyGems.org via Artifactory Pro and as such have configured a remote repository.

When defining the Artifactory endpoint as a source with 'gem source -a https://<USERNAME>:<PASSWORD>@artifactorytest.contoso.com/artifactory/api/gems/RubyGemsTest/' this works perfectly fine and we retrieve the gem via Artifactory.

When I configure a Gemfile with that same source and execute `bundle install –verbose` I get the following output: 

```
C:\Temp_Space\test>bundle install --verbose
  HTTP GET https://USERNAME:APIKEY@artifactorytest.contoso.com/artifactory/api/gems/RubyGemsTest/versions
  HTTP 503 Service Unavailable
  Bundler::HTTPError: Net::HTTPServiceUnavailable: <HTML><HEAD>
  <TITLE>Network Error</TITLE>
  </HEAD>
  <BODY>
  <FONT face="Helvetica">
  <big><strong></strong></big><BR>
  </FONT>
  <blockquote>
  <TABLE border=0 cellPadding=1 width="80%">
  <TR><TD>
  <FONT face="Helvetica">
  <big>Network Error (tcp_error)</big>
  <BR>
  <BR>
  </FONT>
  </TD></TR>
  <TR><TD>
  <FONT face="Helvetica">
  A communication error occurred: "Operation timed out"
  </FONT>
  </TD></TR>
  <TR><TD>
  <FONT face="Helvetica">
  The Web Server may be down, too busy, or experiencing other problems preventing it from responding to requests. You may wish to try again at a later time.
  </FONT>
  </TD></TR>
  <TR><TD>
  <FONT face="Helvetica" SIZE=2>
  <BR>
  For assistance, contact your network support team.
  </FONT>
  </TD></TR>
  </TABLE>
  </blockquote>
  </FONT>
  </BODY></HTML>

  C:/Ruby22-x64/lib/ruby/gems/2.2.0/gems/bundler-1.12.5/lib/bundler/fetcher/downloader.rb:36:in `fetch'
    C:/Ruby22-x64/lib/ruby/gems/2.2.0/gems/bundler-1.12.5/lib/bundler/fetcher/compact_index.rb:81:in `block in compact_index_client'
[...]

PR #5801, opened 2017-06-21
Add explicit nil proxy arguments - Fixes no_proxy support

jakauppila:

When the `no_proxy` environment variable is configured it is not being honored and requests for hosts that are present in the list will be sent through the proxy defined via `http_proxy` and `https_proxy`.

Ultimately this means that a user utilizing the `http_proxy`, `https_proxy`, and `no_proxy` environment variables will unable to access an internal RubyGems source.

The underlying `net-http-persistent` gem currently does not explicitly pass in nil arguments when a host matches a value within `no_proxy`, so when `Net::HTTP.new()` is called, it will fall back an utilize the values defined in `http_proxy` and `https_proxy`.

https://github.com/drbrain/net-http-persistent/pull/88 has been submitted to add the explicit nil argument to be defined.

Adding explicit nil arguments for the proxy definition for `Net::HTTP.new()` as defined by the documentation.

https://ruby-doc.org/stdlib-2.4.1/libdoc/net/http/rdoc/Net/HTTP.html#method-c-new

Per the discussion in https://github.com/bundler/bundler/issues/5781, as Bundler is currently stuck to v2.9.4 of `net-http-persistent` for now, it would be best addressed by a change to the vendored code.

A PR has been submitted to `net-http-persistent` which can be pulled into Bundler when v3.x will be integrated.

### DIFF
--- a/lib/bundler/vendor/net-http-persistent/lib/net/http/persistent.rb
+++ b/lib/bundler/vendor/net-http-persistent/lib/net/http/persistent.rb
@@ -616,6 +616,8 @@ class Bundler::Persistent::Net::HTTP::Persistent
     if @proxy_uri and not proxy_bypass? uri.host, uri.port then
       connection_id << @proxy_connection_id
       net_http_args.concat @proxy_args
+    else
+      net_http_args.concat [nil, nil, nil, nil]
     end
 
     connection = connections[connection_id]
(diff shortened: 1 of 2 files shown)


segiddins:
Can you please add test coverage for this change? Thanks!

jakauppila:
Added a few tests. It's not a direct test for the underlying library, but the best I can think of at this level.

I'm not terribly versed with Ruby and RSpec, can I get deeper into the library to verify what is getting passed to `Net::HTTP.new()` ?

segiddins:
I think you're testing at the right level of abstraction already

jakauppila:
Pushed a few more changes and squashed them.

Had to pull out my Macbook to run the tests locally since apparently the repo can't be cloned on Windows due to question marks in a few paths.

indirect:
this looks good to me. @segiddins r?

segiddins:
@bundlerbot r+

PR #5802, opened 2017-06-21
Remove `require "spec_helper"` from spec files

koic:
Follow up of #5634.

Since `--require spec_helper` is specified in the .rspec file, This PR will remove redundancy.


colby-swandale:
Thank you 🎉 

@bundlerbot r+

koic:
Thanks too!

PR #5803, opened 2017-06-21
Make `install --path` relative to the pwd

### LINKED ISSUE #2048: --standalone generates file in incorrect PWD
Hello,
I ran the following

```
bundle install --gemfile=src/main/webapp/WEB-INF/Gemfile --path ./target/bundler/ --standalone
```

and it generated the setup file in the following location `target/bundler/bundler/setup.rb` while it installed the gems in `src/main/webapp/WEB-INF/target/bundler/`. So it assumed that the --path was relative to the Gemfile instead of the PWD. It also created the .bundle/config in the WEB-INF folder.

Personally I think the --path option should be relative to the PWD but it seems to be relative to the Gemfile. So I think we should just make the bundler/setup.rb also relative to the Gemfile's path.

igorbozato:

> I ran the following
> 
> bundle install --gemfile=src/main/webapp/WEB-INF/Gemfile --path ./target/bundler/ --standalone
> and it generated the setup file in the following location target/bundler/bundler/setup.rb while it installed the gems in src/main/webapp/WEB-INF/target/bundler/. So it assumed that the --path was relative to the Gemfile instead of the PWD. It also created the .bundle/config in the WEB-INF folder.

Closes #2048

As discussed on the issue, the path is currently being relative to the Gemfile instead of the cwd.

Making the path relative to the cwd if the new feature flag `path_relative_to_cwd` is set t true.

This work was started by @agis (https://github.com/agis/bundler/commit/1da8a7021bdd9bbe76398dddec8bc499655666dd).

### DIFF
--- a/lib/bundler/cli/common.rb
+++ b/lib/bundler/cli/common.rb
@@ -98,25 +98,5 @@ module Bundler
       clean &&= !Bundler.use_system_gems?
       clean
     end
-
-    def self.set_path(path, if_given = false)
-      if path.nil?
-        Bundler.settings.set_command_option :path, nil unless if_given
-        return
-      end
-
-      path = Pathname.new(path)
-      if path.relative? && Bundler.feature_flag.path_relative_to_cwd?
-        path = path.expand_path
-        root = Bundler.root
-        path = begin
-                 path.relative_path_from(root)
-               rescue ArgumentError
-                 path
-               end
-      end
-
-      Bundler.settings.set_command_option :path, path.to_s
-    end
   end
 end
--- a/lib/bundler/cli/common.rb
+++ b/lib/bundler/cli/common.rb
@@ -98,5 +98,22 @@ module Bundler
       clean &&= !Bundler.use_system_gems?
       clean
     end
+
+    def self.set_path(path, if_given = false)
+      return if if_given && path.nil?
+
+      path = Pathname.new(path)
+      if path.relative? && Bundler.feature_flag.path_relative_to_cwd?
+        path = path.expand_path
+        root = Bundler.root
+        path = begin
+                 path.relative_path_from(root)
+               rescue ArgumentError
+                 path
+               end
+      end
+
+      Bundler.settings.set_command_option :path, path.to_s
+    end
   end
 end
--- a/lib/bundler/cli/common.rb
+++ b/lib/bundler/cli/common.rb
@@ -100,7 +100,10 @@ module Bundler
     end
 
     def self.set_path(path, if_given = false)
-      return if if_given && path.nil?
+      if path.nil?
+        Bundler.settings.set_command_option :path, nil unless if_given
+        return
+      end
 
       path = Pathname.new(path)
       if path.relative? && Bundler.feature_flag.path_relative_to_cwd?
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -71,8 +71,7 @@ module Bundler
       if Bundler.use_system_gems?
         Bundler.ui.confirm "Use `bundle info [gemname]` to see where a bundled gem is installed."
       else
-        absolute_path = File.expand_path(Bundler.configured_bundle_path.base_path)
-        relative_path = absolute_path.sub(File.expand_path(".") + File::SEPARATOR, "." + File::SEPARATOR)
+        relative_path = Bundler.configured_bundle_path.base_path_relative_to_pwd
         Bundler.ui.confirm "Bundled gems are installed into `#{relative_path}`"
       end
 
@@ -168,12 +167,10 @@ module Bundler
     end
 
     def normalize_settings
-      path = Bundler.feature_flag.path_relative_to_cwd? ? File.expand_path(options["path"]) : options["path"] if options["path"]
-
-      Bundler.settings.set_command_option :path, nil if options[:system]
-      Bundler.settings.set_command_option :path, "vendor/bundle" if options[:deployment]
-      Bundler.settings.set_command_option_if_given :path, path
-      Bundler.settings.set_command_option :path, "bundle" if options["standalone"] && Bundler.settings[:path].nil?
+      CLI::Common.set_path nil if options[:system]
+      CLI::Common.set_path "vendor/bundle" if options[:deployment]
+      CLI::Common.set_path options["path"], :if_given
+      CLI::Common.set_path "bundle" if options["standalone"] && Bundler.settings[:path].nil?
 
       bin_option = options["binstubs"]
       bin_option = nil if bin_option && bin_option.empty?
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -168,9 +168,11 @@ module Bundler
     end
 
     def normalize_settings
+      path = Bundler.feature_flag.path_relative_to_cwd? ? File.expand_path(options["path"]) : options["path"] if options["path"]
+
       Bundler.settings.set_command_option :path, nil if options[:system]
       Bundler.settings.set_command_option :path, "vendor/bundle" if options[:deployment]
-      Bundler.settings.set_command_option :path, Bundler.feature_flag.path_relative_to_cwd? ? File.expand_path(options["path"]) : options["path"] if options["path"]
+      Bundler.settings.set_command_option_if_given :path, path
       Bundler.settings.set_command_option :path, "bundle" if options["standalone"] && Bundler.settings[:path].nil?
 
       bin_option = options["binstubs"]
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -170,7 +170,7 @@ module Bundler
     def normalize_settings
       Bundler.settings.set_command_option :path, nil if options[:system]
       Bundler.settings.set_command_option :path, "vendor/bundle" if options[:deployment]
-      Bundler.settings.set_command_option_if_given :path, options["path"]
+      Bundler.settings.set_command_option :path, Bundler.feature_flag.path_relative_to_cwd? ? File.expand_path(options["path"]) : options["path"] if options["path"]
       Bundler.settings.set_command_option :path, "bundle" if options["standalone"] && Bundler.settings[:path].nil?
 
       bin_option = options["binstubs"]
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -167,10 +167,14 @@ module Bundler
     end
 
     def normalize_settings
-      CLI::Common.set_path nil if options[:system]
-      CLI::Common.set_path "vendor/bundle" if options[:deployment]
-      CLI::Common.set_path options["path"], :if_given
-      CLI::Common.set_path "bundle" if options["standalone"] && Bundler.settings[:path].nil?
+      Bundler.settings.set_command_option :path, nil if options[:system]
+      Bundler.settings.temporary(:path_relative_to_cwd => false) do
+        Bundler.settings.set_command_option :path, "vendor/bundle" if options[:deployment]
+      end
+      Bundler.settings.set_command_option_if_given :path, options[:path]
+      Bundler.settings.temporary(:path_relative_to_cwd => false) do
+        Bundler.settings.set_command_option :path, "bundle" if options["standalone"] && Bundler.settings[:path].nil?
+      end
 
       bin_option = options["binstubs"]
       bin_option = nil if bin_option && bin_option.empty?
--- a/lib/bundler/cli/package.rb
+++ b/lib/bundler/cli/package.rb
@@ -10,7 +10,7 @@ module Bundler
 
     def run
       Bundler.ui.level = "error" if options[:quiet]
-      CLI::Common.set_path options[:path], :if_given
+      Bundler.settings.set_command_option_if_given :path, options[:path]
       Bundler.settings.set_command_option_if_given :cache_all_platforms, options["all-platforms"]
       Bundler.settings.set_command_option_if_given :cache_path, options["cache-path"]
 
--- a/lib/bundler/cli/package.rb
+++ b/lib/bundler/cli/package.rb
@@ -10,7 +10,7 @@ module Bundler
 
     def run
       Bundler.ui.level = "error" if options[:quiet]
-      Bundler.settings.set_command_option_if_given :path, options[:path]
+      CLI::Common.set_path options[:path], :if_given
       Bundler.settings.set_command_option_if_given :cache_all_platforms, options["all-platforms"]
       Bundler.settings.set_command_option_if_given :cache_path, options["cache-path"]
 
--- a/lib/bundler/feature_flag.rb
+++ b/lib/bundler/feature_flag.rb
@@ -44,6 +44,7 @@ module Bundler
     settings_flag(:list_command) { bundler_2_mode? }
     settings_flag(:lockfile_uses_separate_rubygems_sources) { bundler_2_mode? }
     settings_flag(:only_update_to_newer_versions) { bundler_2_mode? }
+    settings_flag(:path_relative_to_cwd) { bundler_2_mode? }
     settings_flag(:plugins) { @bundler_version >= Gem::Version.new("1.14") }
     settings_flag(:prefer_gems_rb) { bundler_2_mode? }
     settings_flag(:print_only_version_number) { bundler_2_mode? }
--- a/lib/bundler/settings.rb
+++ b/lib/bundler/settings.rb
@@ -251,9 +251,14 @@ module Bundler
       end
 
       def base_path_relative_to_pwd
-        expanded_base_path = Pathname.new(base_path).expand_path(Bundler.root)
+        base_path = Pathname.new(self.base_path)
+        expanded_base_path = base_path.expand_path(Bundler.root)
         relative_path = expanded_base_path.relative_path_from(Pathname.pwd)
-        relative_path = Pathname.new(File.join(".", relative_path)) unless relative_path.to_s.start_with?("..")
+        if relative_path.to_s.start_with?("..")
+          relative_path = base_path if base_path.absolute?
+        else
+          relative_path = Pathname.new(File.join(".", relative_path))
+        end
         relative_path
       rescue ArgumentError
         expanded_base_path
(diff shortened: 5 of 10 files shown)


lib/bundler/settings.rb
```diff
@@ -32,6 +32,7 @@ class Settings
       silence_root_warning
       unlock_source_unlocks_spec
       update_requires_all_flag
+      path_relative_to_cwd
```
segiddins: please keep the list ordered

lib/bundler/feature_flag.rb
```diff
@@ -22,6 +22,7 @@ def self.settings_flag(flag, &default)
     settings_flag(:prefer_gems_rb) { bundler_2_mode? }
     settings_flag(:unlock_source_unlocks_spec) { !bundler_2_mode? }
     settings_flag(:update_requires_all_flag) { bundler_2_mode? }
+    settings_flag(:path_relative_to_cwd) { bundler_2_mode? }
```
segiddins: please keep the list ordered

spec/install/path_spec.rb
```diff
@@ -49,6 +49,18 @@
...
     end
+
+    context "with path_relative_to_cwd set to true" do
+      before { bundle! "config path_relative_to_cwd true" }
+
+      it "installs the bundle relatively to current working directory" do
+        Dir.chdir(bundled_app.parent) do
+          bundle "install --gemfile='#{bundled_app}/Gemfile' --path vendor/bundle"
+          expect(out).to include("installed into ./vendor/bundle")
+          expect(Dir.exist?("vendor/bundle")).to be true
```
segiddins: `expect(bundled_app("../vendor/bundle")).to be_directory`

spec/install/path_spec.rb
```diff
@@ -49,6 +49,18 @@
       expect(vendored_gems("gems/rack-1.0.0")).to be_directory
       expect(the_bundle).to include_gems "rack 1.0.0"
     end
+
+    context "with path_relative_to_cwd set to true" do
+      before { bundle! "config path_relative_to_cwd true" }
+
+      it "installs the bundle relatively to current working directory" do
+        Dir.chdir(bundled_app.parent) do
+          bundle "install --gemfile='#{bundled_app}/Gemfile' --path vendor/bundle"
```
segiddins: `bundle!`

lib/bundler/settings.rb
segiddins: (you'll also need to document the setting in `bundle-config.ronn`, as this will cause `spec/quality_spec` to fail)

lib/bundler/cli/install.rb
```diff
@@ -205,7 +207,7 @@ def normalize_settings
       Bundler.settings[:no_prune]            = true if options["no-prune"]
 
       Bundler.settings[:no_install]          = true if options["no-install"]
-
+
```
segiddins: trailing whitespace

igorbozato:
Thank you for the feedback @segiddins, just pushed the request changes.

segiddins:
Awesome, thanks!
@indirect r?

indirect:
@bundlerbot r+

lib/bundler/cli/install.rb
```diff
@@ -192,7 +192,9 @@ def normalize_groups
     def normalize_settings
       Bundler.settings[:path]                = nil if options[:system]
       Bundler.settings[:path]                = "vendor/bundle" if options[:deployment]
-      Bundler.settings[:path]                = options["path"] if options["path"]
+
+      Bundler.settings[:path]                = Bundler.feature_flag.path_relative_to_cwd? ? File.expand_path(options["path"]) : options["path"] if options["path"]
```
segiddins: this will store the `path` as being absolute instead of a relative path, which is maybe not what we want?

segiddins: ping on this -- I think you'll want to expand this where its used, rather than in the setting

segiddins:
This seems to break stuff when the feature flag is actually enabled: https://travis-ci.org/bundler/bundler/jobs/269540739#L1078

segiddins:
@igorbozato I have some ideas for this -- do you mind if I push them up to your branch?

igorbozato:
@segiddins sure, do I need to give you push access to my fork?

segiddins:
Nope!
Since you opened a PR I can push to the branch the PR is from. How does this look?

```diff
diff --git a/lib/bundler/cli/check.rb b/lib/bundler/cli/check.rb
index e572787dc..607a5351d 100644
--- a/lib/bundler/cli/check.rb
+++ b/lib/bundler/cli/check.rb
@@ -9,7 +9,7 @@ module Bundler
     end
 
     def run
-      Bundler.settings.set_command_option_if_given :path, options[:path]
+      CLI::Common.set_path options[:path], :if_given
 
       begin
         definition = Bundler.definition
diff --git a/lib/bundler/cli/common.rb b/lib/bundler/cli/common.rb
index 9d40ee9df..2edae86d0 100644
--- a/lib/bundler/cli/common.rb
+++ b/lib/bundler/cli/common.rb
@@ -98,5 +98,22 @@ module Bundler
       clean &&= !Bundler.use_system_gems?
       clean
     end
+
+    def self.set_path(path, if_given = false)
+      return if if_given && path.nil?
+
+      path = Pathname.new(path)
+      if path.relative? && Bundler.feature_flag.path_relative_to_cwd?
+        path = path.expand_path
+        root = Bundler.root
+        path = begin
+                 path.relative_path_from(root)
+               rescue ArgumentError
+                 path
+               end
+      end
+
+      Bundler.settings.set_command_option :path, path.to_s
+    end
   end
 end
diff --git a/lib/bundler/cli/install.rb b/lib/bundler/cli/install.rb
index f0b821ed8..10f3a9800 100644
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -71,8 +71,7 @@ module Bundler
       if Bundler.use_system_gems?
         Bundler.ui.confirm "Use `bundle info [gemname]` to see where a bundled gem is installed."
       else
-        absolute_path = File.expand_path(Bundler.configured_bundle_path.base_path)
-        relative_path = absolute_path.sub(File.expand_path(".") + File::SEPARATOR, "." + File::SEPARATOR)
+        relative_path = Bundler.configured_bundle_path.base_path_relative_to_pwd
         Bundler.ui.confirm "Bundled gems are installed into `#{relative_path}`"
       end
 
@@ -168,10 +167,10 @@ module Bundler
     end
 
     def normalize_settings
-      Bundler.settings.set_command_option :path, nil if options[:system]
-      Bundler.settings.set_command_option :path, "vendor/bundle" if options[:deployment]
-      Bundler.settings.set_command_option_if_given :path, options["path"]
-      Bundler.settings.set_command_option :path, "bundle" if options["standalone"] && Bundler.settings[:path].nil?
+      CLI::Common.set_path nil if options[:system]
+      CLI::Common.set_path "vendor/bundle" if options[:deployment]
+      CLI::Common.set_path options["path"], :if_given
+      CLI::Common.set_path "bundle" if options["standalone"] && Bundler.settings[:path].nil?
 
       bin_option = options["binstubs"]
       bin_option = nil if bin_option && bin_option.empty?
diff --git a/lib/bundler/cli/package.rb b/lib/bundler/cli/package.rb
index 2dcd0e1e2..8aef0265e 100644
--- a/lib/bundler/cli/package.rb
+++ b/lib/bundler/cli/package.rb
@@ -10,7 +10,7 @@ module Bundler
 
     def run
       Bundler.ui.level = "error" if options[:quiet]
-      Bundler.settings.set_command_option_if_given :path, options[:path]
+      CLI::Common.set_path options[:path], :if_given
       Bundler.settings.set_command_option_if_given :cache_all_platforms, options["all-platforms"]
       Bundler.settings.set_command_option_if_given :cache_path, options["cache-path"]
 
diff --git a/lib/bundler/feature_flag.rb b/lib/bundler/feature_flag.rb
index 6a1809cd4..c85aa9cbc 100644
--- a/lib/bundler/feature_flag.rb
+++ b/lib/bundler/feature_flag.rb
@@ -43,6 +43,7 @@ module Bundler
     settings_flag(:list_command) { bundler_2_mode? }
     settings_flag(:lockfile_uses_separate_rubygems_sources) { bundler_2_mode? }
     settings_flag(:only_update_to_newer_versions) { bundler_2_mode? }
+    settings_flag(:path_relative_to_cwd) { bundler_2_mode? }
     settings_flag(:plugins) { @bundler_version >= Gem::Version.new("1.14") }
     settings_flag(:prefer_gems_rb) { bundler_2_mode? }
     settings_flag(:print_only_version_number) { bundler_2_mode? }
diff --git a/lib/bundler/settings.rb b/lib/bundler/settings.rb
index dee7c14ca..546dc639e 100644
--- a/lib/bundler/settings.rb
+++ b/lib/bundler/settings.rb
@@ -42,6 +42,7 @@ module Bundler
       no_install
       no_prune
       only_update_to_newer_versions
+      path_relative_to_cwd
       path.system
       plugins
       prefer_gems_rb
@@ -245,6 +246,15 @@ module Bundler
         path
       end
 
+      def base_path_relative_to_pwd
+        expanded_base_path = Pathname.new(base_path).expand_path(Bundler.root)
+        relative_path = expanded_base_path.relative_path_from(Pathname.pwd)
+        relative_path = Pathname.new(File.join(".", relative_path)) unless relative_path.to_s.start_with?("..")
+        relative_path
+      rescue ArgumentError
+        expanded_base_path
+      end
+
       def validate!
         return unless explicit_path && system_path
         path = Bundler.settings.pretty_values_for(:path)
diff --git a/man/bundle-config.ronn b/man/bundle-config.ronn
index f1f3062ae..3d3063dfb 100644
--- a/man/bundle-config.ronn
+++ b/man/bundle-config.ronn
@@ -218,6 +218,8 @@ learn more about their operation in [bundle install(1)][bundle-install].
    is used, defaults to vendor/bundle.
 * `path.system` (`BUNDLE_PATH__SYSTEM`):
    Whether Bundler will install gems into the default system path (`Gem.dir`).
+* `path_relative_to_cwd` (`PATH_RELATIVE_TO_CWD`)
+   Makes `--path` relative to the CWD instead of the `Gemfile`.
 * `plugins` (`BUNDLE_PLUGINS`):
    Enable Bundler's experimental plugin system.
 * `prefer_gems_rb` (`BUNDLE_PREFER_GEMS_RB`)
diff --git a/spec/install/path_spec.rb b/spec/install/path_spec.rb
index fb356899a..196cbb4ab 100644
--- a/spec/install/path_spec.rb
+++ b/spec/install/path_spec.rb
@@ -24,7 +24,7 @@ RSpec.describe "bundle install" do
 
       Dir.chdir(dir) do
         bundle! :install, forgotten_command_line_options(:path => "vendor/bundle")
-        expect(out).to include("installed into `./vendor/bundle`")
+        expect(out).to include("installed into `../vendor/bundle`")
       end
 
       dir.rmtree
@@ -49,6 +49,37 @@ RSpec.describe "bundle install" do
       expect(vendored_gems("gems/rack-1.0.0")).to be_directory
       expect(the_bundle).to include_gems "rack 1.0.0"
     end
+
+    context "with path_relative_to_cwd set to true" do
+      before { bundle! "config path_relative_to_cwd true" }
+
+      it "installs the bundle relatively to current working directory", :bundler => "< 2" do
+        Dir.chdir(bundled_app.parent) do
+          bundle! "install --gemfile='#{bundled_app}/Gemfile' --path vendor/bundle"
+          expect(out).to include("installed into `./vendor/bundle`")
+          expect(bundled_app("../vendor/bundle")).to be_directory
+        end
+        expect(the_bundle).to include_gems "rack 1.0.0"
+      end
+
+      it "installs the standalone bundle relative to the cwd" do
+        Dir.chdir(bundled_app.parent) do
+          bundle! :install, :gemfile => bundled_app("Gemfile"), :standalone => true
+          expect(out).to include("installed into `./bundle`")
+          expect(bundled_app("../bundle")).to be_directory
+          expect(bundled_app("../bundle/ruby")).to be_directory
+        end
+
+        bundle! "config --delete path"
+
+        Dir.chdir(bundled_app("subdir").tap(&:mkpath)) do
+          bundle! :install, :gemfile => bundled_app("Gemfile"), :standalone => true
+          expect(out).to include("installed into `./bundle`")
+          expect(bundled_app("subdir/bundle")).to be_directory
+          expect(bundled_app("subdir/bundle/ruby")).to be_directory
+        end
+      end
+    end
   end
 
   describe "when BUNDLE_PATH or the global path config is set" do

```

segiddins:
@indirect this should be passing now, please review?

igorbozato:
Looks great, but breaks a few tests here.

segiddins:
r?

spec/install/path_spec.rb
```diff
@@ -24,7 +24,7 @@
 
       Dir.chdir(dir) do
         bundle! :install, forgotten_command_line_options(:path => "vendor/bundle")
-        expect(out).to include("installed into `./vendor/bundle`")
+        expect(out).to include("installed into `../vendor/bundle`")
```
indirect: Why is the second dot added here, if the command running is `install --path vendor/bundle`, presumably it is installing into `$PWD/vendor/bundle`, right? And then the "installed into" would also be `./vendor/bundle`?

lib/bundler/cli/common.rb
```diff
@@ -98,5 +98,25 @@ def self.clean_after_install?
       clean &&= !Bundler.use_system_gems?
       clean
     end
+
+    def self.set_path(path, if_given = false)
+      if path.nil?
+        Bundler.settings.set_command_option :path, nil unless if_given
```
colby-swandale:
```
if path.nil? && !if_given
  Bundler.settings.set_command_option :path, nil
  return
end
```

i think reads better

lib/bundler/cli/common.rb
```diff
@@ -98,5 +98,25 @@ def self.clean_after_install?
...
+        return
```
colby-swandale: does this need to return `nil`?

lib/bundler/cli/common.rb
```diff
@@ -98,5 +98,25 @@ def self.clean_after_install?
...
+      end
+
+      path = Pathname.new(path)
+      if path.relative? && Bundler.feature_flag.path_relative_to_cwd?
+        path = path.expand_path
+        root = Bundler.root
```
colby-swandale: i don't think `root` needs to be in a variable here

lib/bundler/cli/install.rb
```diff
@@ -168,10 +167,10 @@ def normalize_groups
     end
 
     def normalize_settings
-      Bundler.settings.set_command_option :path, nil if options[:system]
-      Bundler.settings.set_command_option :path, "vendor/bundle" if options[:deployment]
-      Bundler.settings.set_command_option_if_given :path, options["path"]
-      Bundler.settings.set_command_option :path, "bundle" if options["standalone"] && Bundler.settings[:path].nil?
+      CLI::Common.set_path nil if options[:system]
+      CLI::Common.set_path "vendor/bundle" if options[:deployment]
+      CLI::Common.set_path options["path"], :if_given
```
colby-swandale: Can this not just be `true`?

lib/bundler/cli/package.rb
```diff
@@ -10,7 +10,7 @@ def initialize(options)
 
     def run
       Bundler.ui.level = "error" if options[:quiet]
-      Bundler.settings.set_command_option_if_given :path, options[:path]
+      CLI::Common.set_path options[:path], :if_given
```
colby-swandale: Same here, can this not just be `true`?

spec/install/path_spec.rb
```diff
@@ -49,6 +49,37 @@
       expect(vendored_gems("gems/rack-1.0.0")).to be_directory
       expect(the_bundle).to include_gems "rack 1.0.0"
     end
+
+    context "with path_relative_to_cwd set to true" do
```
colby-swandale: `with path_relative_to_cwd setting set to true`

spec/install/path_spec.rb
```diff
@@ -49,6 +49,37 @@
...
+      before { bundle! "config path_relative_to_cwd true" }
+
+      it "installs the bundle relatively to current working directory", :bundler => "< 2" do
```
colby-swandale: why are we setting `:bundler => "< 2"` here?

spec/install/path_spec.rb
colby-swandale: this gives a bit of context for anyone looking at this this about we mean

lib/bundler/cli/install.rb
segiddins: `:if_given` tells you what the argument is, since we can't use named parameters yet

lib/bundler/cli/common.rb
segiddins: it just needs the early return

lib/bundler/cli/common.rb
segiddins: this avoids the `rescue` effecting the `Bundler.root` call

spec/install/path_spec.rb
segiddins: because this test expects the gem to be _available_, which it won't be on 2.0

lib/bundler/cli/common.rb
segiddins: but that behaves differently, we need to return even if `if_given` if true

spec/install/path_spec.rb
segiddins: It's because the `relative to cwd` behavior doesn't apply when running `bundle config`, which is what's happening in 2.0 mode, and in 1.0 mode the feature flag is disabled

indirect: Ahh. Okay, I think that means we need to apply `relative to cwd` when running `bundle config path` as well. IMO it is identically confusing to run `bundle config path ./foo` and have that path interpreted relative to the Gemfile dir rather than cwd.

segiddins: Ok, I can look into reimplementing this behavior in the settings validator

segiddins: done !

indirect:
@bundlerbot r+

segiddins:
Did a PR just pass CI in 24 minutes 😱

PR #5804, opened 2017-06-21
Completely remove the postit trampoline

segiddins:

The problem was the bundler trampoline we tried to introduce never completely worked, it worked even less with bundler installed as a default gem (as we intend to do with RubyGems, like, a year ago), and it worked by doing disgusting things that we couldn't be confident would work under all circumstances.

My diagnosis was our best bet was to completely remove the trampoline from bundler itself.
We intend to address the user story of switching bundler versions in RubyGems directly, where the gymnastics required will hopefully be much less obtrusive. Additionally

My fix is to completely delete all references to trampolining / postit from Bundler and admit defeat.

I chose this fix because I'm unwilling to maintain the trampoline any longer, and since it's never been enabled, I feel this is my last chance to pull it from Bundler before I'm stuck maintaining code that doesn't work for all eternity. This will also _finally_ unblock me shipping RubyGems 2.7.


indirect:
Legit 👍 @bundlerbot r+

clupprich:
👋 Just a quick question: Why isn't this removal not mentioned in the `Changelog` for 1.15.2?

segiddins:
Because it was never publicly supported

PR #5806, opened 2017-06-22
Avoid hitting RubyGems remotes when all missing gems are in --without groups

segiddins:

The problem was we would attempt to fetch dependency info from a RubyGems remote even when it was only needed for gems whose groups were excluded.

My diagnosis was that the test we had for this case was insufficient, since it checked only that `stderr` was empty, and since Bundler 1 prints errors to `stdout`... it went undetected for a year.

My fix is to only report that deps are missing when they are either requested or are path/git/gemspec-sourced.

I chose this fix because it keeps the behavior of always ensuring all git repos are checked out.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -593,6 +593,9 @@ module Bundler
 
       # order here matters, since Index#== is checking source.specs.include?(locked_index)
       locked_index != source.specs
+    rescue PathError, GitError => e
+      Bundler.ui.debug "Assuming that #{source} has not changed since fetching its specs errored (#{e})"
+      false
     end
 
     # Get all locals and override their matching sources.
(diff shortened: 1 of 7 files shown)


indirect:
Any chance we can also skip checking out git sources if they only contain gems that are excluded? That would fix the regression I found in 1.8 last year :P

segiddins:
> Any chance we can also skip checking out git sources if they only contain gems that are excluded? That would fix the regression I found in 1.8 last year :P

I don't know how well I can make that work with `rspec ./spec/install/git_spec.rb:39 # bundle install git sources should check out git repos that are missing but not being installed`

indirect:
@segiddins that's when there's no lockfile and Bundler needs to use the gemspec from that git repo to successfully resolve. I'm talking about when we're running in deployment/frozen mode, and there's already a lock, so we don't actually need the gemspec for the without gem.

segiddins:
@indirect I think I did it?

indirect:
:tada:

segiddins:
r?

indirect:
My test for the regression is still failing inside `Definition#converge_paths`, where `specs_changed?` requires the directory to exist. Are we checking every source for changes, whether or not we need any gems from that source, even in frozen mode?

indirect:
(The repro is here if you want to see it for yourself: https://gist.github.com/indirect/1cc86d686f40a40d68b895834ab11cf6)

segiddins:
That looks like a different issue? That gemfile doesn't have any Git sources.

segiddins:
Lol I think I have a fix for it anyways

indirect:
lol 🙌🏻

segiddins:
@indirect take a look :P

indirect:
you did it! 🎉🎉🎉

indirect:
@bundlerbot r+

PR #5809, opened 2017-06-22
[Settings] Print pretty values for settings as their converted values, rather than strings

segiddins:

The problem was `bundle config` would print bool keys as strings (i.e. `true` was printed as `"true"`)

My diagnosis was we needed to convert the values before formatting them

My fix extracts the conversion method, and calls it in `pretty_values_for`


indirect:
@bundlerbot r+

PR #5811, opened 2017-06-22
[2.0] Stop remembering CLI options in Bundler 2

segiddins:

The problem was that on Bundler 1.0, command line options passed to the different commands would be remembered _across command invocations_. This was hella confusing for users, and made Bundler a difficult command line tool to work with -- you'd pass an option once, it'd be set without you realizing it, and Bundler would behave differently.

See https://trello.com/c/yGsPNDpg/48-stop-auto-remembering-any-command-flags for the original discussion, but I've excerpted bits below:

> Silently remembering flags that were passed to a command sometime in the past completely breaks all expectations for how command-line utilities work. It’s convenient for some users some of the time, but at the cost of many, many bugs filed that turn out to be unexpectedly remembered options.

My diagnosis was that, behind a feature flag, we'd change commands behavior to only set settings temporarily (for the life of the current process), and after that disable options that become useless when not remembered.

See https://github.com/bundler/bundler/pull/3955.

My fix is to remove most CLI options in Bundler 2. Those that remain are set only for the duration of the process, rather than persisted to disk.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -52,7 +52,7 @@ module Bundler
       end
     end
 
-    def self.forgotten_option(*args, &blk)
+    def self.deprecated_option(*args, &blk)
       return if Bundler.feature_flag.forget_cli_options?
       method_option(*args, &blk)
     end
@@ -152,13 +152,13 @@ module Bundler
 
       If the bundle has already been installed, bundler will tell you so and then exit.
     D
-    forgotten_option "binstubs", :type => :string, :lazy_default => "bin", :banner =>
+    deprecated_option "binstubs", :type => :string, :lazy_default => "bin", :banner =>
       "Generate bin stubs for bundled gems to ./bin"
     method_option "clean", :type => :boolean, :banner =>
       "Run bundle clean automatically after install"
-    forgotten_option "deployment", :type => :boolean, :banner =>
+    deprecated_option "deployment", :type => :boolean, :banner =>
       "Install using defaults tuned for deployment environments"
-    forgotten_option "frozen", :type => :boolean, :banner =>
+    deprecated_option "frozen", :type => :boolean, :banner =>
       "Do not allow the Gemfile.lock to be updated after this install"
     method_option "full-index", :type => :boolean, :banner =>
       "Fall back to using the single-file index of all gems"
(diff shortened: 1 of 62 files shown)


segiddins:
I think this is no longer a WIP, so feedback (particularly on which options are removed/kept in 2.0) would be 🚀

segiddins:
:all-of-the-conflicts:

lib/bundler/cli.rb
```diff
@@ -53,6 +52,11 @@ def initialize(*args)
       end
     end
 
+    def self.forgotten_option(*args, &blk)
```
colby-swandale: Can i suggest `removed_option` or `deprecate_option`  instead. This confused me a bit with the CLI no longer remembering options.

lib/bundler/installer/gem_installer.rb
```diff
@@ -66,6 +66,7 @@ def out_of_space_message
     end
 
     def generate_executable_stubs
+      return if Bundler.feature_flag.forget_cli_options?
       return if Bundler.settings[:inline]
```
colby-swandale: `return if Bundler.settings[:inline] || Bundler.feature_flag.forget_cli_options?`

lib/bundler/cli.rb
segiddins: 👍 I think I'll go with `deprecated_option`

lib/bundler/installer/gem_installer.rb
segiddins: I prefer having the two separate conditions, since they're basically unrelated

segiddins:
Rebased, addressed Colby's conflicts, fixed specs.
@indirect please review

lib/bundler/cli.rb
```diff
@@ -164,28 +168,28 @@ def check
       "Specify the number of jobs to run in parallel"
     method_option "local", :type => :boolean, :banner =>
       "Do not attempt to fetch gems remotely and use the gem cache instead"
-    method_option "no-cache", :type => :boolean, :banner =>
+    deprecated_option "no-cache", :type => :boolean, :banner =>
       "Don't update the existing gem cache."
     method_option "force", :type => :boolean, :banner =>
       "Force downloading every gem."
     method_option "no-prune", :type => :boolean, :banner =>
```
indirect: I think we should deprecate `--no-prune`... afaik it was only added because of `--cache`. Related, does the 2.0 `cache` command effectively run `install and then fill vendor/cache`? It should probably have a `--prune/--no-prune` flag, to offer equivalent functionality that cache-users can migrate to.

lib/bundler/cli.rb
```diff
@@ -148,13 +152,13 @@ def check
 
       If the bundle has already been installed, bundler will tell you so and then exit.
     D
-    method_option "binstubs", :type => :string, :lazy_default => "bin", :banner =>
+    deprecated_option "binstubs", :type => :string, :lazy_default => "bin", :banner =>
       "Generate bin stubs for bundled gems to ./bin"
     method_option "clean", :type => :boolean, :banner =>
       "Run bundle clean automatically after install"
```
indirect: Can we deprecate this as well? I am pretty sure we can say "you can just run `clean`, which will `install` and then `clean`". The concept of cleaning isn't even really coherent without an install happening first, imo.

lib/bundler/cli.rb
```diff
@@ -164,28 +168,28 @@ def check
       "Specify the number of jobs to run in parallel"
     method_option "local", :type => :boolean, :banner =>
       "Do not attempt to fetch gems remotely and use the gem cache instead"
-    method_option "no-cache", :type => :boolean, :banner =>
+    deprecated_option "no-cache", :type => :boolean, :banner =>
       "Don't update the existing gem cache."
     method_option "force", :type => :boolean, :banner =>
       "Force downloading every gem."
```
indirect: I feel like `force` is a very unclear name for this option, and I'd like to deprecate it in favor of something like `--redownload`. Or maybe better yet, a completely different command that nukes the cache, forcing downloads next time. (The latter is out of scope for this PR, though).

segiddins: `--redownload` sounds good to me!

lib/bundler/cli.rb
segiddins: the 2.0 cache command is the 1.0 package command... it basically runs `install && Bundler.load.cache` and it already has a `--no-prune` flag. unrelated, but this reminds me we need to default `cache_all_platforms` to true in 2.0

lib/bundler/cli.rb
segiddins: sure -- does this mean you'd like to default to cleaning or not after every install into a local `path`?

segiddins:
> I'd also like to deprecate init --gemspec, if that's okay with everyone else.

out of curiosity, why?

segiddins:
@indirect think I got all of the ones you requested!

lib/bundler/cli.rb
indirect: great 👍

lib/bundler/cli.rb
indirect: 👍

lib/bundler/cli.rb
indirect: I would like to default to cleaning if and only if the bundle is being installed into the default application-local `.bundle` directory. If the user has provided some other "shared" location, I think cleaning should default to off. I'm worried that's possibly confusing, though. :/

indirect:
This looks good 👍  I'd like to deprecate `init --gemspec` because I think it should be replaced by using `gemspec` in the Gemfile.

segiddins:
Cool! Just fixed the one test failure, so r?

indirect:
@bundlerbot r+

Issue #5712, opened 2017-06-12
Setting BUNDLE_GEMFILE=Gemfile causes Bundler.setup to fail when in deployment mode


mattbrictson:
Starting with Bundler 1.15.0, `Bundler.setup` fails with an `ArgumentError` raised from `Pathname#relative_path_from` when the following conditions are met:

* Bundler is in deployment mode; and
* A Gemfile is explicitly specified via the `BUNDLE_GEMFILE` environment variable; and
* That Gemfile is *not* an absolute path (e.g. `BUNDLE_GEMFILE=Gemfile`)

```
Bundler   1.15.1
Rubygems  2.6.12
Ruby      2.4.1p111 (2017-03-22 revision 58053) [x86_64-darwin15]
Git       2.13.1
Platform  x86_64-darwin-15
```

Steps to reproduce:

```
$ touch Gemfile
$ bundle install
The Gemfile specifies no dependencies
Resolving dependencies...
Bundle complete! 0 Gemfile dependencies, 1 gem now installed.
Use `bundle info [gemname]` to see where a bundled gem is installed.
$ bundle install --deployment
The Gemfile specifies no dependencies
Bundle complete! 0 Gemfile dependencies, 1 gem now installed.
Bundled gems are installed into ./vendor/bundle.
$ BUNDLE_GEMFILE=Gemfile ruby -rbundler -e Bundler.setup
ruby/2.4.0/pathname.rb:520:in `relative_path_from': different prefix: "" and "/Users/mbrictson/Downloads/bundler-defect" (ArgumentError)
  from ruby/gems/2.4.0/gems/bundler-1.15.1/lib/bundler/definition.rb:392:in `ensure_equivalent_gemfile_and_lockfile'
  from ruby/gems/2.4.0/gems/bundler-1.15.1/lib/bundler/runtime.rb:14:in `setup'
  from ruby/gems/2.4.0/gems/bundler-1.15.1/lib/bundler.rb:101:in `setup'
  from -e:1:in `<main>'
```

This bug seems to have been introduced in f145aa11a5b47b438263aace8bd45dd2d752411c.

mattbrictson:
I have an idea for a fix, which is to simply add `expand_path`:

```diff
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -389,7 +389,7 @@ module Bundler
       msg = String.new
       msg << "You are trying to install in deployment mode after changing\n" \
[...]

PR #5815, opened 2017-06-24
Allow BUNDLE_GEMFILE to not be an absolute path

gxespino:
Hello - I was stalking the open issues and had time to implement this quick PR for your review. This is a direct fix to https://github.com/bundler/bundler/issues/5712.

I implemented the fix that @segiddins suggested in the comments, just added a test scenario. Please let me know if anything else is needed!

---

The problem was that starting with Bundler 1.15.0, `Bundler.setup` fails with an `ArgumentError` raised from `Pathname#relative_path_from` when the following conditions are met:

- Bundler is in deployment mode; and
- A Gemfile is explicitly specified via the `BUNDLE_GEMFILE` environment variable; and
- That Gemfile is not an absolute path (e.g. `BUNDLE_GEMFILE=Gemfile`)

My diagnosis was that in `Bundler::SharedHelpers#default_gemfile`, the `Pathname` object being instantiated was not being expanded. So, in the case that `ENV["BUNDLE_GEMFILE"]` was not an absolute path, `Bundle.setup` would error out `Bundler.default_lockfile.relative_path_from(SharedHelpers.pwd)` since `Bundler.default_lockfile` derived itself from `Bundler.default_gemfile`.

My fix was to add `expand_path` to the `Pathname` object being returned in `Bundler::SharedHelpers#default_gemfile`

I chose this fix because Bundler was already deriving `Bundler::Sharedhelpers#root` with an `#expand_path`, so I wanted to follow that pattern.

### DIFF
--- a/spec/bundler/shared_helpers_spec.rb
+++ b/spec/bundler/shared_helpers_spec.rb
@@ -101,7 +101,7 @@ RSpec.describe Bundler::SharedHelpers do
       context "currently in directory with a Gemfile" do
         before { File.new("Gemfile", "w") }
 
-        it "returns path of the bundle gemfile" do
+        it "returns path of the bundle Gemfile" do
           expect(subject.in_bundle?).to eq("#{bundled_app}/Gemfile")
         end
       end
@@ -118,6 +118,14 @@ RSpec.describe Bundler::SharedHelpers do
 
       it "returns ENV['BUNDLE_GEMFILE']" do
         expect(subject.in_bundle?).to eq("/path/Gemfile")
+      end
+    end
+
+    context "ENV['BUNDLE_GEMFILE'] is set without an absolute path" do
+      before { ENV["BUNDLE_GEMFILE"] = "Gemfile" }
+
+      it "returns path of the bundle Gemfile" do
+        expect(subject.in_bundle?).to eq("#{bundled_app}/Gemfile")
       end
     end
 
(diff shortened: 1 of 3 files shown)


spec/bundler/shared_helpers_spec.rb
```diff
@@ -121,6 +121,14 @@
       end
     end
 
+    context "ENV['BUNDLE_GEMFILE'] is set without an absolute path" do
+      before { ENV["BUNDLE_GEMFILE"] = "Gemfile" }
+
+      it "returns path of the bundle Gemfile" do
+        expect(subject.in_bundle?).to eq("#{bundled_app}/Gemfile")
```
segiddins: please test that `subject.find_gemfile` returns this, as only the truthiness of `in_bundle?` is guaranteed.

gxespino:
Hello. I updated the PR with a higher level unit test as well as an integration spec. I also changed the solution to move `expand_path` further up the chain. 

Let me know if this is not desired and/or if any more changes are needed!

segiddins:
:100: this is perfect, thank you so much!

@bundlerbot r+

mattbrictson:
I noticed this didn't make it into 1.15.2. Will it be ported to 1.15.x at some point?

segiddins:
It'll definitely go into 1.16

Issue #5596, opened 2017-04-18
bundle clean extensions


dylanahsmith:
Install any gem with an extension through bundler then try to remove it with `bundle clean --force` and the extension will still be in the file system.

For example, create a Gemfile

```
source "https://rubygems.org"
gem "cityhash"
```

then install and uninstall the gem using bundler

```
$ bundle install
...
Installing cityhash 0.8.1 with native extensions
...
$ find $GEM_HOME/extensions -name '*.bundle'
/Users/dylansmith/.gem/ruby/2.3.3/extensions/x86_64-darwin-15/2.3.0/cityhash-0.8.1/cityhash/cityhash.bundle
$ echo '' > Gemfile
$ bundle clean --force
Resolving dependencies...
Removing cityhash (0.8.1)
$ find $GEM_HOME/extensions -name '*.bundle'
/Users/dylansmith/.gem/ruby/2.3.3/extensions/x86_64-darwin-15/2.3.0/cityhash-0.8.1/cityhash/cityhash.bundle
```

but the extension wasn't cleaned.

A workaround would be to remove the `$GEM_HOME` path and re-installing bundler.

Note that my actual use case was that I want to uninstall a gem installed through bundler from a git source, where `gem uninstall` can't be used.

segiddins:
This seems intentional, according to https://github.com/bundler/bundler/pull/4030

colby-swandale:
@segiddins is there any action for this issue?

segiddins:
I don't know, the behavior reported seems to be intentional

dylanahsmith:
It looks like #4030 was about git gems having their extensions being removed **when they were still used**.

This issue is about gems that have been removed from the Gemfile.  Notice in the PR description that I do this by simply emptying the Gemfile with the following line

> $ echo '' > Gemfile

in my example, `bundle clean --force` does in fact remove the cityhash gem, so I don't see why the extension for the removed gem shouldn't also get removed.

segiddins:
In that case, a PR changing the behavior would be 💯

indirect:
I can confirm that not removing the extension is an accident, leftover from when extensions were moved outside of the directory containing the gem's other files. A PR to fix this would be great. 👍🏻

colby-swandale:
A PR to fix this issue has been merged and will be part of the next Bundler release.

PR #5816, opened 2017-06-26
[WIP] Bundler::Runtime#clean: remove extensions dirs

igorbozato:

The problem was...
See #5596

My diagnosis was...

My fix...

I chose this fix because...

### DIFF
--- a/lib/bundler/runtime.rb
+++ b/lib/bundler/runtime.rb
@@ -164,6 +164,7 @@ module Bundler
       gem_dirs             = Dir["#{Gem.dir}/gems/*"]
       gem_files            = Dir["#{Gem.dir}/cache/*.gem"]
       gemspec_files        = Dir["#{Gem.dir}/specifications/*.gemspec"]
+      extension_dirs       = Dir["#{Gem.dir}/extensions/#{Bundler.local_platform}/*/*"]
       spec_gem_paths       = []
       # need to keep git sources around
       spec_git_paths       = @definition.spec_git_paths
@@ -171,6 +172,7 @@ module Bundler
       spec_gem_executables = []
       spec_cache_paths     = []
       spec_gemspec_paths   = []
+      spec_extension_dirs  = []
       specs.each do |spec|
         spec_gem_paths << spec.full_gem_path
         # need to check here in case gems are nested like for the rails git repo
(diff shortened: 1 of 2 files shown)


colby-swandale:
HI, can you please fill out the questionnaire.

spec/commands/clean_spec.rb
```diff
@@ -699,4 +699,54 @@ def should_not_have_gems(*gems)
...
+
+    gemfile <<-G
+      source "file://#{gem_repo2}"
+
+      gem "c_extension"
+    G
+
+    bundle "install --path vendor/bundle"
+
+    expect(vendored_gems("extensions/x86_64-linux/2.3.0-static/c_extension-1.0")).to exist
```
segiddins: please don't hardcode the platform, use something like ` bundled_app(*["vendor/bundle", Gem.ruby_engine, Gem::ConfigMap[:ruby_version], c_extention-1.0"].compact)` instead

spec/commands/clean_spec.rb
```diff
@@ -699,4 +699,54 @@ def should_not_have_gems(*gems)
...
+
+  it "removes extensions after clean" do
+    build_repo2 do
+      build_gem "c_extension" do |s|
+        s.extensions = ["ext/extconf.rb"]
+        s.write "ext/extconf.rb", <<-E
+          require "mkmf"
+          name = "c_extension_bundle"
+          dir_config(name)
+          raise "OMG" unless with_config("c_extension") == "hello"
```
segiddins: this seems to be raising on CI

lib/bundler/runtime.rb
```diff
@@ -193,6 +196,9 @@ def clean(dry_run = false)
       stale_gem_dirs       = gem_dirs - spec_gem_paths
       stale_gem_files      = gem_files - spec_cache_paths
       stale_gemspec_files  = gemspec_files - spec_gemspec_paths
+      stale_extension_dirs = extension_dirs - spec_extension_dirs
+
+      stale_extension_dirs.collect {|dir| FileUtils.rm_rf(dir) } unless dry_run
```
segiddins: this should assign to a variable that gets added to `output` a couple of lines below and use the `remove_dir` helper, same as 2 lines below

igorbozato: Using `remove_dir` for this would duplicate the output of the removed gem, what should be the output text for this?

spec/commands/clean_spec.rb
```diff
@@ -699,4 +699,54 @@ def should_not_have_gems(*gems)
...
+
+    gemfile <<-G
+      source "file://#{gem_repo2}"
+
+      gem "c_extension"
+    G
+
+    bundle "install --path vendor/bundle"
+
+    expect(vendored_gems.join("extensions", Gem::Platform.local.to_s, Gem::ConfigMap[:ruby_version] + "-static", "c_extension-1.0")).to exist
```
igorbozato: `Gem::ConfigMap[:ruby_version] + "-static"` still hardcoded, not sure where `"2.3.0-static"` comes from.

PR #5818, opened 2017-06-27
[2.0] Enable specific_platform by default on 2.0

segiddins:

The problem was that Bundler has somewhat suspect handling of multi-platform gems. We'd assume that different platform versions of gems were generally interchangeable, so if Bundler resolved to the "ruby" platform gem we'd just blindly try to swap in the gem for the local platform, which could lead to issues (say if the sets of dependencies were different).

My diagnosis was that we needed to stop only working with the notion of "generic" platforms, which mapped everything to (basically) either java, pure ruby, and windows, and instead keep track of the actual platforms a bundle was being used on, and resolve for those specific platforms.

My fix enables the changes made in https://github.com/bundler/bundler/pull/4836 by default on Bundler 2.

I chose this fix because it means Bundler will default to more correct platforms semantics out of the box.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -861,7 +861,7 @@ module Bundler
             "To add those platforms to the bundle, " \
             "run `bundle lock --add-platform #{mapped_platforms.join " "}`."
         end
-        platforms.each do |p|
+        Resolver.sort_platforms(platforms).each do |p|
           deps << DepProxy.new(dep, p) if remote || p == generic_local_platform
         end
       end
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -847,11 +847,12 @@ module Bundler
     end
 
     def expand_dependencies(dependencies, remote = false)
+      sorted_platforms = Resolver.sort_platforms(@platforms)
       deps = []
       dependencies.each do |dep|
         dep = Dependency.new(dep, ">= 0") unless dep.respond_to?(:name)
         next if !remote && !dep.current_platform?
-        platforms = dep.gem_platforms(@platforms)
+        platforms = dep.gem_platforms(sorted_platforms)
         if platforms.empty?
           mapped_platforms = dep.platforms.map {|p| Dependency::PLATFORM_MAP[p] }
           Bundler.ui.warn \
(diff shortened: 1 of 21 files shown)


segiddins:
Will update the specs on 2.0 once the source changes land, as those already fork a lot of the lockfile output specs

segiddins:
Holy heck this was hard to get passing, took a bunch of time

segiddins:
Should be green now. Please review.

lib/bundler/definition.rb
```diff
@@ -904,7 +904,7 @@ def expand_dependencies(dependencies, remote = false)
             "To add those platforms to the bundle, " \
             "run `bundle lock --add-platform #{mapped_platforms.join " "}`."
         end
-        platforms.each do |p|
+        Resolver.sort_platforms(platforms).each do |p|
```
jules2689: I'm a little concerned about sorting in `definition`. Is this on the require path?
We should skip this block entirely if `remote == true` since the entire functionality is not used.

lib/bundler/resolver.rb
```diff
@@ -26,6 +26,9 @@ def message
           end.min_by(&:size)
           trees.reject! {|t| !maximal.include?(t.last) } if maximal
 
+          trees = trees.sort_by {|t| t.flatten.map(&:to_s) }
```
jules2689: We're doing another sort below, maybe we can combine the `sort_by`, `uniq`, and `sort_by` for perf reasons?

lib/bundler/definition.rb
segiddins: It appears to be

lib/bundler/resolver.rb
segiddins: this is in an error message, so I'm not overly concerned with performance

jules2689: Oh true.

lib/bundler/definition.rb
segiddins: I have an idea to speed this up that I'll try out.

segiddins: Should be improved now, it only sorts the platforms once instead of once per dependency

indirect:
LGTM. 👍 @bundlerbot r+

segiddins:
Rebased.

@bundlerbot r=indirect

PR #5820, opened 2017-06-27
[2.0] [DSL] Remove default git sources on 2.0

segiddins:

The problem was the default git source shortcuts have been deprecated, but still existed in 2.0.

My diagnosis was we needed to avoid adding them in 2.0.

My fix is to introduce a feature flag, which when enabled will stop adding the sources to the DSL, and additionally will disable the `github` DSL method.


segiddins:
Rebased

indirect:
@bundlerbot r+

PR #5822, opened 2017-06-27
Stop gitignoring Gemfile.lock in default template

deivid-rodriguez:
This will be followed up with a documentation PR on some best practices to keep gems tested against up to date dependencies. 

The problem was that sometimes open source contributors get discouraged because they can't setup the project they want to contribute to easily. 

My diagnosis was that a lot of projects gitignore `Gemfile.lock` just because it's gitignored by default in the new gem template. This causes `bundle install` to no longer be guaranteed to work. And `bundle install` is usually the first step towards a contribution.

My fix was to remove the `Gemfile.lock` entry from the gitignore template.

I chose this fix because to prevent a file from being gitignored, removing it from the `.gitignore` file usually does the trick :)

### DIFF
--- a/lib/bundler/templates/newgem/gitignore.tt
+++ b/lib/bundler/templates/newgem/gitignore.tt
@@ -1,6 +1,5 @@
 /.bundle/
 /.yardoc
-/Gemfile.lock
 /_yardoc/
 /coverage/
 /doc/


segiddins:
@indirect r?

indirect:
@deivid-rodriguez thanks for the discussion around this and the PR! looks good. 👍 

@bundlerbot r+

deivid-rodriguez:
@indirect Thanks to you! I'll follow up with a documentation PR as soon as I find time.

indirect:
@deivid-rodriguez awesome! I created a tracking issue for that over here: https://github.com/bundler/bundler/issues/5879

PR #5826, opened 2017-06-29
Avoid Range Not Satisfiable errors during normal request flow

### LINKED ISSUE #5373: Falling back to the dependency API on 1.14
Specifying a ruby version in the Gemfile no longer causes bundler to find versions of gems compatible with the specified ruby version.

Gemfile used for testing:
```
source 'https://rubygems.org'

ruby '2.0.0'

gem 'chef', '~> 11.12.4'
gem 'ohai'
gem 'mime-types', '~> 1.25.0'
gem 'nokogiri', '~> 1.5.5'
```

output of `ruby -v`
`ruby 2.0.0p648 (2015-12-16) [x86_64-linux]`

output of `bundle install`:
```
Resolving dependencies...
Rubygems 2.0.14.1 is not threadsafe, so your gems will be installed one at a time. Upgrade to Rubygems 2.1.0 or higher to enable parallel gem installation.
Installing hashie 2.1.2
Installing json 1.8.1 with native extensions
Installing mixlib-log 1.7.1
Installing rack 2.0.1
Installing diff-lcs 1.3
Installing erubis 2.7.0
Installing highline 1.7.8
Installing mime-types 1.25.1
Installing mixlib-cli 1.7.0
Installing mixlib-config 2.2.4
Installing mixlib-shellout 1.6.1
Installing net-ssh 2.9.4
Installing ipaddress 0.8.3
Installing systemu 2.5.2
Installing yajl-ruby 1.3.0 with native extensions
Installing coderay 1.1.1
Installing method_source 0.8.2
Installing slop 3.6.0
Installing nokogiri 1.5.11 with native extensions
Using bundler 1.14.3
Installing mixlib-authentication 1.4.1
Gem::InstallError: rack requires Ruby version >= 2.2.2.
An error occurred while installing rack (2.0.1), and Bundler cannot continue.
Make sure that `gem install rack -v '2.0.1'` succeeds before bundling.
```

Tested with bundler version 1.14.0, 1.14.1, 1.14.2, 1.14.3. Command works on bundler v 1.13.7

greysteil:
Previously, Bundler was requesting partial response ranges for the Rubygems compact index that could be empty. Since Rubygems was [ignoring the `ETag` header](https://github.com/rubygems/rubygems.org/pull/1652) for these requests, empty ranges would occur whenever the versions index (for instance) hadn't been modified since the version Bundler currently had cached. When this happened, Rubygems would respond with a 416 (Range Not Satisfiable). Bundler would treat this as a `Bundler::HTTPError`, and fall back to using `Fetcher::Dependency` for dependency info. Sadly, that meant metadata about what Ruby version each dependency required was no-longer checked, and updates for gems which should be limited by the system Ruby version were failing.

Closes #5373.

See above

This PR updates the range Bundler requests from Rubygems to ensure it's always satisfiable. It does that but requesting all bytes from (and including) the final byte in the Bundler cache, rather than all bytes after (and not including) it.

An alternative fix would be to catch the 416 responses and retry the index lookup in those cases, asking for a full response. That would mean an extra request in all of those cases, though - this method keeps the number of calls to Rubygems down.

### DIFF
--- a/spec/install/gems/compact_index_spec.rb
+++ b/spec/install/gems/compact_index_spec.rb
@@ -768,6 +768,27 @@ The checksum of /versions does not match the checksum provided by the server! So
     end
   end
 
+  it "performs partial update with a non-empty range" do
+    gemfile <<-G
+      source "#{source_uri}"
+      gem 'rack', '0.9.1'
+    G
+
+    # Initial install creates the cached versions file
+    bundle! :install, :artifice => "compact_index"
+
+    # Update the Gemfile so we can check subsequent install was successful
+    gemfile <<-G
+      source "#{source_uri}"
+      gem 'rack', '1.0.0'
+    G
+
+    # Second install should make only a partial request to /versions
+    bundle! :install, :artifice => "compact_index_partial_update"
+
+    expect(the_bundle).to include_gems "rack 1.0.0"
+  end
+
   it "performs partial update while local cache is updated by another process" do
     gemfile <<-G
       source "#{source_uri}"
--- a/lib/bundler/compact_index_client/updater.rb
+++ b/lib/bundler/compact_index_client/updater.rb
@@ -34,7 +34,14 @@ module Bundler
           if retrying.nil? && local_path.file?
             FileUtils.cp local_path, local_temp_path
             headers["If-None-Match"] = etag_for(local_temp_path)
-            headers["Range"] = "bytes=#{local_temp_path.size}-"
+            headers["Range"] =
+              if local_temp_path.size.nonzero?
+                # Subtract a byte to ensure the range won't be empty.
+                # Avoids 416 (Range Not Satisfiable) responses.
+                "bytes=#{local_temp_path.size - 1}-"
+              else
+                "bytes=#{local_temp_path.size}-"
+              end
           else
             # Fastly ignores Range when Accept-Encoding: gzip is set
             headers["Accept-Encoding"] = "gzip"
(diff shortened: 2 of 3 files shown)


segiddins:
Thanks for this, @greysteil! Can you please look at the failures on Travis and look into adding a spec in `spec/install/gems/compact_index`? Thanks!

greysteil:
Sure thing - will do it today.

greysteil:
@segiddins - done and done. Not sure why Travis isn't rebuilding the commit...

segiddins:
I don't know why this isn't building on CI either...

greysteil:
Tried rebasing and force pushing a couple of days ago and it didn't help. Let me know if you want me to try again.

segiddins:
Sure, why not... Might have just been a temporary travis hiccup where it missed the GH web hook?

greysteil:
No joy 😞 .

segiddins:
I believe this will close https://github.com/bundler/bundler/issues/5373?

greysteil:
It will indeed.

spec/install/gems/compact_index_spec.rb
```diff
@@ -768,6 +768,27 @@ def start
...
+    bundle! :install, :artifice => "compact_index"
+
+    # Update the Gemfile so we can check subsequent install was successful
+    gemfile <<-G
+      source "#{source_uri}"
+      gem 'rack', '1.0.0'
+    G
+
+    # Second install should make only a partial request to /versions
+    bundle! :install, :artifice => "compact_index_partial_update"
```
segiddins: this appears to be failing on some CI runs

greysteil: Interesting. I've replicated the random failure locally, and will look into now.

greysteil: Argh, creating a new `CompactIndex::VersionFile` in the etag_response block was creating a file with a new `created_at` line at the top, accurate to the second. Sometimes the second call was within the same second, sometimes it wasn't. Fixed now, and specs are running.

greysteil:
:greenheart:

segiddins:
@indirect r?

indirect:
awesome, thanks! @bundlerbot r+

greysteil:
🍾  Thanks for everything you guys do maintaining Bundler and keeping everyone else's lives easier!

Issue #5828, opened 2017-06-29
Unable to checkout Bundler repo on Windows


jakauppila:
It is currently not possible to checkout the Bundler repo on Windows due to the vcr_cassettes containing question marks in folder names added in #5220.

Would it be possible to convert the serialized request/response to another format?

```
D:\Dropbox\GIT_Workspaces>git clone https://github.com/bundler/bundler
Cloning into 'bundler'...
remote: Counting objects: 70934, done.
remote: Compressing objects: 100% (18/18), done.
remote: Total 70934 (delta 18), reused 14 (delta 10), pack-reused 70906
Receiving objects: 100% (70934/70934), 31.52 MiB | 6.92 MiB/s, done.
Resolving deltas: 100% (45282/45282), done.
fatal: cannot create directory at 'spec/support/artifice/vcr_cassettes/realworld/api.rubygems.org/api/v1/dependencies?gems=bundler': Invalid argument
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry the checkout with 'git checkout -f HEAD'
```

colby-swandale:
\cc @segiddins

segiddins:
Why is `?` in file names an issue? As far as I'm aware, it's not `File::SEPARATOR` on any platform?

jakauppila:
The following are reserved characters on Windows:


```
The following reserved characters:
- < (less than)
- > (greater than)
- : (colon)
- " (double quote)
- / (forward slash)
- \ (backslash)
- | (vertical bar or pipe)
- ? (question mark)
- * (asterisk)
```
https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247

colby-swandale:
fixed in #5829

jakauppila:
Thanks! It is now possible to check out the repo.

PR #5829, opened 2017-06-30
remove forbidden chars in cassettes pathname

colby-swandale:

Users running Micorsoft Windows are currently unable to clone to Bundler project due to a forbidden character in the folder path: 

`spec/support/artifice/vcr_cassettes/realworld/api.rubygems.org/api/v1/dependencies?gems=bundler'`

The `?` being the forbidden character.

See #5828

Replaced the forbidden character in the folder name with a `-` and updated the VCR spec helper to replace any forbidden character with a `-` in the filename function.

This was the most simple approach to fix the issue.


spec/support/artifice/vcr.rb
```diff
@@ -74,7 +74,7 @@ def key
     end
 
     def file_name_for_key(key)
-      key.join("/")
+      key.join("/").gsub(/[\:*?”<>|]/, "-")
```
segiddins: invalid regexp on old rubies

colby-swandale:
should be right to go now, now sure what was happening with the regex 😕

segiddins:
Thanks Colby!
@bundlerbot r+

Issue #5786, opened 2017-06-18
Feature: Check permissions on Bundler home directory in doctor command 


colby-swandale:
We should have a check in `bundle doctor` for the file/folder permissions in the Bundler home directory.

We should print a warning if there are any files/folders that is owned by another user but is readable/writable but prints an error when a file cannot be read or written to.

See: #5691

indirect:
👍🏻

gxespino:
Working on this issue - should be fun!

ajwann:
@gxespino are you still working on this? If not, do you mind if I try my hand at it?

gxespino:
Sure don't! Have fun!

ajwann:
@colby-swandale I'm working on this right now and will hopefully have a WIP PR up in an hour or two.

ajwann:
Just so I understand this correctly, you want the doctor command to check inside the ```Bundler.home``` directory, find every file not owned by the current user (I was going to use ```Process.uid``` to determine current user), and then check to see if those files are writable. If they are writable, you want Bundler to attempt to write to each of those files to see if an exception is thrown.

Does this sound correct?

colby-swandale:
This has been merged in #6129

PR #5831, opened 2017-07-01
[WIP] Start implementation of permissions command

gxespino:
Hello - This is a WIP in response to https://github.com/bundler/bundler/issues/5786. 

I'm envisioning a `bundle permissions` command that will display either a:

1. Permissions satisfied message
2. Permissions warning in the case `bundle install` can proceed, but the `bundle_path` and/or any child directories are not owned by the executing user.
3. Permissions error in the case bundle install cannot proceed

`bundle permissions` will be ran as a part of `bundle doctor`.

Let me know if I'm headed down the right path. Also, any ideas on how to simulate changing the user of a directory in tests? Can't seem to get specification 2 to work.

Lastly, taking recommendations on the specific message verbiage!

Note: Will fill out PR questionnaire upon completion.

### DIFF
--- a/bundler.gemspec
+++ b/bundler.gemspec
@@ -36,6 +36,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency "rdiscount",  "~> 2.2"
   s.add_development_dependency "ronn",       "~> 0.7.3"
   s.add_development_dependency "rspec",      "~> 3.6"
+  s.add_development_dependency "pry"
 
   s.files = `git ls-files -z`.split("\x0").reject {|f| f.match(%r{^(test|spec|features)/}) }
   # we don't check in man pages, but we need to ship them because
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -567,6 +567,12 @@ module Bundler
       Issue.new.run
     end
 
+    desc "permissions", "CHECKS PERMISSIONS"
+    def permissions
+      require "bundler/cli/permissions"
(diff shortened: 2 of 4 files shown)


colby-swandale:
Thanks for tackling this issue.

Why did you decide to make this a command? The use case i had when i wrote the issue was that users could just run `bundle doctor` if they were having issues with file permissions, so the `bundle permissions` command seems a bit redundant.

gxespino:
Definitely doesn't have to be a separate command. I wrote it separately in order to isolate the logic and test cases with the intention to eventually call `Bundler::CLI::Permissions.new({}).run` in `bundle doctor`. 

Also, `bundle check` seemed to establish a pattern for small utility commands in which it was initially separate but then later also included in `bundle doctor` https://github.com/bundler/bundler/pull/4888

Either way - I definitely considered the redundancy that you brought up and would be happy to move all of this code into `bundle doctor` and have a one-size-fits-all utility command to display potential warnings and errors. Would we want also want to remove `bundle check` if that was the case?

segiddins:
`bundle check` exists as a separate command because it is intended to be invoked directly by users to see if their dependencies are all installed -- `bundle doctor` just so happens to call it. Please move these checks to the `bundle doctor` command

gxespino:
Sounds good!

colby-swandale:
ping @gxespino are you still working on this?

colby-swandale:
I'm closing this for now. If you're still working on this PR feel free to re-open it.

PR #5832, opened 2017-07-01
Extract lockfile generation into a new class

segiddins:

The problem was that eventually we might want to add more stuff to the lockfile.

My diagnosis was that lockfile generation should probably exist outside of a single method.

My fix is to create a `LockfileGenerator` class.

### DIFF
--- a/lib/bundler/lockfile_generator.rb
+++ b/lib/bundler/lockfile_generator.rb
@@ -1,0 +1,94 @@
+# frozen_string_literal: true
+module Bundler
+  class LockfileGenerator
+    attr_reader :definition
+    attr_reader :out
+
+    # @private
+    def initialize(definition)
+      @definition = definition
+      @out = String.new
+    end
+
+    def self.generate(definition)
+      new(definition).generate!
+    end
+
+    def generate!
+      add_sources
+      add_platforms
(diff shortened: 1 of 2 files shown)


colby-swandale:
Are there any specs we can extract from `definition_spec` as well? I'm happy for a TODO for the time being.

segiddins:
Nope, nothing that directly tests `to_lock` in `definition_spec`

lib/bundler/lockfile_generator.rb
```diff
@@ -0,0 +1,94 @@
...
+      add_locked_ruby_version
+      add_bundled_with
+
+      out
+    end
+
+  private
+
+    def add_sources
+      definition.send(:sources).lock_sources.each_with_index do |source, idx|
```
jules2689: Is there any need to use `definition.send(:sources)`? Can you access the `sources` directly? If not, can we look at making `sources` not private?

segiddins: its not public now and I didn't want to change access control. Eventually, I think all the definition properties that are accessed should be passed into the `LockfileGenerator` but considered it as a second step

colby-swandale:
@bundlerbot r+

PR #5833, opened 2017-07-02
Add a bin/bundle binstub

segiddins:

The problem was that developing bundler can be difficult, and letting users use the in-development bundler locally should be as easy as possible.

My diagnosis was we could use a binstub that handles setting up the load path for bundler, ensures the gem spec is activated, and loads the local bundler code.

My fix adds such a binstub.


colby-swandale:
🎉 @bundlerbot r+

PR #5834, opened 2017-07-02
Tweak gemspec template when using Ruby 2.x

koic:

There are unnecessary magic comment for currently maintained Ruby 2.2, 2.3, and 2.4.

The default script encoding from Ruby 2.0 is UTF-8.

https://bugs.ruby-lang.org/issues/6679

Ruby 1.9 is EOL, so I think that there is not much Gem to start developed using it. 

This PR removes magic comment when starting Gem development (i.e. `bundle gem`) with Ruby 2.0 or higher version.


segiddins:
@bundlerbot r+

PR #5837, opened 2017-07-05
Improve logging in default gem activation specs

segiddins:

The problem was those specs were failing on CI and I couldn't figure out why.

My diagnosis was we needed the hack loaded before all bundler code.

My fix gets the hack loaded in via an `-r` switch set in `RUBYOPT`

I chose this fix because it will make spec failures more informative.

### DIFF
--- a/spec/runtime/setup_spec.rb
+++ b/spec/runtime/setup_spec.rb
@@ -1235,7 +1235,7 @@ end
         end << "bundler"
       end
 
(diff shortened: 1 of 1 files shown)


segiddins:
@bundlerbot r+

segiddins:
Fixed on Ruby 1.9

@bundlerbot r+

segiddins:
@bundlerbot retry

PR #5838, opened 2017-07-05
Make Bundler.setup not make Kernel#gem public in Bundler 2

segiddins:

The problem was Bundler 1 accidentally made `Kernel#gem` public, even though RubyGems declares it as private. Oops. We tried to make it private in 1.13, it broke stuff, so we added in a hack to keep it public.

My diagnosis was we could delete that hack in 2.0.

My fix implements a feature flag that skips making `Kernel#gem` public, and adds regression tests for `gem` or `require` accidentally being made public.


indirect:
awesome. @bundlerbot r+

PR #5839, opened 2017-07-05
Sort all documented settings in the bundle-config man page

segiddins:

The problem was we kept on adding new settings to the end of the list, making it completely unordered.

My diagnosis was we should order the list.

My fix orders the list and tests that it is indeed ordered.

I chose this fix because I could write a little script to do it.


colby-swandale:
Great work!

@bundlerbot r+

Issue #4619, opened 2016-05-26
[rfc] Remove separate "frozen" and "deployment"


indirect:
In Bundler 1.x, the `--deployment` flag is an alias for `--frozen --path vendor/bundle`. Now that we don't save flags, we should provide a documented config option for frozen/deployment mode, and remove the confusing combination flag.

Since 2.x defaults to `--path .bundle`, I propose that we remove the `--path vendor/bundle` part from `--deployment`, and remove `frozen` completely, replacing it everywhere with `deployment`, eg `bundle config deployment true` if you want future commands to throw an error rather than update the Gemfile.

Thoughts?

segiddins:
Sounds reasonable

RochesterinNYC:
Agreed. I haven't yet seen any use cases for `--frozen` that lack the context of deployment.

PR #5843, opened 2017-07-06
[2.0] Use deployment instead of frozen as a setting on Bundler 2

segiddins:

The problem was bundler had a `--deployment` flag, a `--frozen` flag and a `frozen` setting. All frozen meant was the lockfile couldn't change, whereas `--deployment` meant the full "deployment experience". This was confusing.

Closes https://github.com/bundler/bundler/issues/4619.

My diagnosis was  we want to get rid of the `--frozen` setting and add a new `deployment` setting that will be used by default on Bundler 2.

My fix adds the deployment setting, and updates the myriad places we use `Bundler.settings[:frozen]` to take it into account, along with fixing the behavior when `bundle config deployment true` is used instead of the command line switches.

### DIFF
--- a/man/bundle-config.ronn
+++ b/man/bundle-config.ronn
@@ -139,7 +139,8 @@ learn more about their operation in [bundle install(1)][bundle-install].
 * `console` (`BUNDLE_CONSOLE`):
    The console that `bundle console` starts. Defaults to `irb`.
 * `deployment` (`BUNDLE_DEPLOYMENT`):
-   Disallow changes to the `Gemfile`.
+   Disallow changes to the `Gemfile`. When the `Gemfile` is changed and the
+   lockfile has not been updated, running Bundler commands will be blocked.
 * `disable_checksum_validation` (`BUNDLE_DISABLE_CHECKSUM_VALIDATION`):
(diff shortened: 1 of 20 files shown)


segiddins:
This should be ready now, please review @colby-swandale @indirect. I'd like to land this change before the Great Command Line Switch Deprecation of 2017.

man/bundle-config.ronn
```diff
@@ -138,6 +138,8 @@ learn more about their operation in [bundle install(1)][bundle-install].
    `bundle install`.
 * `console` (`BUNDLE_CONSOLE`):
    The console that `bundle console` starts. Defaults to `irb`.
+* `deployment` (`BUNDLE_DEPLOYMENT`):
+   Disallow changes to the `Gemfile`.
```
indirect: I think this wording is a little unclear... how do you feel about something like "Block commands if the Gemfile has been changed and no longer matches the lock."

segiddins: I took this from the description of the `frozen` setting, should I update both?

segiddins:
@indirect requested change made, r?

indirect:
@bundlerbot r+

PR #5844, opened 2017-07-06
Speed up no-op installs

### LINKED ISSUE #5842: `install` is ~4x slower than `check`
As requested, here's a ticket with an example of how much slower `install` is than `check`, even when both commands are (theoretically) doing the same amount of work. For the record, in the no-op case I think it's fine for the `install` command to print "Bundle installed and ready. ✨" or something along those lines, instead of printing out "Using..." lines for every gem.

```
$ time bundle check
The Gemfile's dependencies are satisfied

real    0m0.614s
user    0m0.555s
sys     0m0.035s
```

```
$ time bundle install --quiet

real    0m2.182s
user    0m2.015s
sys     0m0.105s
```

Repro repo: https://gist.github.com/indirect/6f60f0db1e2f86804ca3ac7c01f2afc4

segiddins:

The problem was that a "no-op `bundle install`", defined as a user running `bundle install` when `bundle check` would have succeeded was taking roughly 4 times as long as `bundle check`.

Closes #5842.

My diagnosis was that there were a few different bottlenecks:

- loading plugins
- using `Bundler::Worker` when the work done in that workers block would finish nearly instantly
- some unoptimized code in the parallel installer

My fix avoids the bottlenecks by checking if the `install` is guaranteed to be a no-op and short-circuiting all but the core `install` method, which in turn will now avoid the `Bundler::Worker` overhead and quickly cache gems when needed.

I chose this fix because it succeeded in speeding up no-op installs by a _huge_ margin.

This _should_ make `install` nearly as fast as a `bundle check`, but at the moment it's still around ~200ms slower. Not an insignificant gap, but a far cry from what it was before the patch. Additionally, since the `install` code path does things such as ensuring everything is cached.

### Time

Run against https://gist.github.com/indirect/6f60f0db1e2f86804ca3ac7c01f2afc4

```
$ time dbundle check
The Gemfile's dependencies are satisfied

BUNDLE_DISABLE_POSTIT=1 ruby  ~/Development/OpenSource/bundler/exe/bundle   0.32s user 0.06s system 94% cpu 0.403 total
```

```
$ time bundle check
The Gemfile's dependencies are satisfied
BUNDLE_DISABLE_POSTIT=1 ruby  ~/Development/OpenSource/bundler/exe/bundle   0.34s user 0.05s system 96% cpu 0.404 total
```

```
$ time dbundle install --quiet
BUNDLE_DISABLE_POSTIT=1 ruby  ~/Development/OpenSource/bundler/exe/bundle    0.48s user 0.07s system 95% cpu 0.582 total
```

```
$ time bundle install --quiet
BUNDLE_DISABLE_POSTIT=1 ruby  ~/Development/OpenSource/bundler/exe/bundle    1.68s user 0.30s system 92% cpu 2.146 total
```

### DIFF
--- a/lib/bundler/installer.rb
+++ b/lib/bundler/installer.rb
@@ -80,9 +80,14 @@ module Bundler
           return
         end
 
-        resolve_if_needed(options)
-        ensure_specs_are_compatible!
-        warn_on_incompatible_bundler_deps
+        if resolve_if_needed(options)
+          ensure_specs_are_compatible!
+          warn_on_incompatible_bundler_deps
+          load_plugins
+          options.delete(:jobs)
+        else
+          options[:jobs] = 1 # to avoid the overhead of Bundler::Worker
+        end
         install(options)
 
         lock unless Bundler.frozen?
@@ -166,10 +171,14 @@ module Bundler
     # that said, it's a rare situation (other than rake), and parallel
     # installation is SO MUCH FASTER. so we let people opt in.
     def install(options)
-      load_plugins
       force = options["force"]
-      jobs = 1
-      jobs = [Bundler.settings[:jobs].to_i - 1, 1].max if can_install_in_parallel?
+      jobs = options.delete(:jobs) do
+        if can_install_in_parallel?
+          [Bundler.settings[:jobs].to_i - 1, 1].max
+        else
+          1
+        end
+      end
       install_in_parallel jobs, options[:standalone], force
     end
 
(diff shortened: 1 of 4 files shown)


segiddins:
This is ready, please review @indirect.

lib/bundler/installer.rb
```diff
@@ -163,10 +168,14 @@ def generate_standalone_bundler_executable_stubs(spec)
     # that said, it's a rare situation (other than rake), and parallel
     # installation is SO MUCH FASTER. so we let people opt in.
     def install(options)
-      load_plugins
       force = options["force"]
-      jobs = 1
-      jobs = [Bundler.settings[:jobs].to_i - 1, 1].max if can_install_in_parallel?
+      jobs = options.delete(:jobs) do
+        if can_install_in_parallel?
+          [Bundler.settings[:jobs].to_i - 1, 1].max
```
colby-swandale: Any particular reason why you're removing methods like this?  This example from `Installer.max_threads`?

segiddins: it was 100% unused

colby-swandale:
@bundlerbot r+ thanks!

colby-swandale:
@bundlerbot retry

segiddins:
Rebased. @bundlerbot r=colby-swandale

segiddins:
@bundlerbot r-

segiddins:
Don't know why the web interface is being so wonky :/

@bundlerbot r=colby-swandale

segiddins:
@bundlerbot r=colby-swandale

PR #5849, opened 2017-07-07
Upgrade thor to 0.20.0

rafaelfranca:
Thor 0.20.0 was released so I'm upgrading bundler to the newest version.

### DIFF
--- a/lib/bundler/vendor/thor/lib/thor/util.rb
+++ b/lib/bundler/vendor/thor/lib/thor/util.rb
@@ -27,7 +27,7 @@ class Bundler::Thor
       end
 
       # Receives a constant and converts it to a Bundler::Thor namespace. Since Bundler::Thor
-      # commands can be added to a sandbox, this method is also responsible for
+      # commands can be added to a sandbox, this method is also responsable for
       # removing the sandbox namespace.
       #
       # This method should not be used in general because it's used to deal with
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -103,9 +103,6 @@ module Bundler
       end
     end
 
-    # Ensure `bundle help --no-color` is valid
-    all_commands["help"].disable_class_options = false
-
     def self.handle_no_command_error(command, has_namespace = $thor_runner)
       if Bundler.feature_flag.plugins? && Bundler::Plugin.command?(command)
         return Bundler::Plugin.exec_command(command, ARGV[1..-1])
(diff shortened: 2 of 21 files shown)


lib/bundler/vendor/thor/lib/thor/util.rb
```diff
@@ -27,7 +27,7 @@ def find_by_namespace(namespace)
       end
 
       # Receives a constant and converts it to a Bundler::Thor namespace. Since Bundler::Thor
-      # commands can be added to a sandbox, this method is also responsible for
```
segiddins: this is a new typo

segiddins:
Yay!

segiddins:
```
An error occurred while loading ./spec/bundler/cli_spec.rb.
Failure/Error: all_commands["help"].disable_class_options = false
NoMethodError:
  undefined method `disable_class_options=' for #<Bundler::Thor::Command:0x00000001cccee0>
# ./spec/bundler/cli_spec.rb:2:in `<top (required)>'
```

segiddins:
https://travis-ci.org/bundler/bundler/jobs/251510586#L672

rafaelfranca:
I'm still not sure how to proceed with that failure. I'd prefer to not require any changes on bundler but the change on thor implements the feature that bundler was already implementing, an sane behavior when we have a conflict generating the files. The difference is that new thor behavior doesn't stop the execution of the command in the first conflict, while the bundler behavior does.

I'm fine with reverting thor's change, but maybe the new thor behavior is better than the current bundler behavior?

segiddins:
@indirect thoughts?

indirect:
Is the new behavior in thor to print `file_conflict` and silently continue? That seems okay to me.

rafaelfranca:
> Is the new behavior in thor to print file_conflict and silently continue? That seems okay to me.

Yes. It will show a `file_conflict` but will continue to the next command.

segiddins:
that one seems fine with an updated test then, but there is another test that's now raising `ENOTDIR`

segiddins:
Updated specs, should pass now

segiddins:
@rafaelfranca all green 🎆

segiddins:
@rafaelfranca anything else you need from us?

rafaelfranca:
Sorry. I'll release it tomorrow

rafaelfranca:
I just released 0.20.0 and updated this PR. It is ready to go

segiddins:
Yay! @bundlerbot r+

Issue #5846, opened 2017-07-06
Difference between 1.14.6 and 1.15.1 - Not finding a compatible dependency (multi_json)


jramnani:
I have a Gemfile that works in Bundler `1.14.6` and not in `1.15.1`, and I'm not sure why.

Here is the error message I get when running `bundle install` using version `1.15.1`:

```
Downloading pygments.rb-1.1.2 revealed dependencies not in the API or the lockfile (multi_json (>= 1.0.0)).
Either installing with `--full-index` or running `bundle update pygments.rb` should fix the problem.
```

Here is my Bundler environment from using the `bundle env` command:

```
The git source `git://github.com/mylesmegyesi/jekyll-s3-upload.git` uses the `git` protocol, which transmits data without encryption. Disable this warning with `bundle config git.allow_insecure true`, or switch to the `https` protocol to keep your data secure.
```

## Environment

```
Bundler   1.15.1
Rubygems  2.2.2
Ruby      2.1.5p273 (2014-11-13 revision 48405) [x86_64-darwin16.0]
GEM_HOME  /usr/local/var/rbenv/versions/2.1.5/gemsets/bundler-test
GEM_PATH  /usr/local/var/rbenv/versions/2.1.5/gemsets/bundler-test:/usr/local/var/rbenv/versions/2.1.5/gemsets/global:/usr/local/var/rbenv/versions/2.1.5/lib/ruby/gems/2.1.0
Git       2.13.0
Platform  x86_64-darwin-16
```

### Gemfile

```ruby
source 'https://rubygems.org'

group :development do
  gem 'activesupport',        '4.2.1'
  gem 'compass',              '1.0.3'
  gem 'jekyll',               '2.5.3'
  gem 'loofah',               '2.0.2'
  gem 'rake',                 '10.4.2'
  gem 'RedCloth',             '4.2.9'
  gem 'rugged',               '0.22.2'
  gem 'uglifier',             '2.7.1'
  gem 'pry'
end

group :jekyll_plugins do
  gem 'jekyll-archives',      '2.0.0'
  gem 'jekyll-assets',        '0.14.0'
  gem 'jekyll-contentblocks', '1.2.0'
  gem 'jekyll-redirect-from', '0.6.2'
  gem 'jekyll-sitemap',       '0.8.1'
  gem 'jekyll-s3-upload',     github: 'mylesmegyesi/jekyll-s3-upload', ref: '23e5457dcf2d1068e18463d399b38a4f9d5c340b'
end
[...]

PR #5850, opened 2017-07-07
[ParallelInstaller] Show require tree when a gem fails to install due to corrupted lockfile

segiddins:

The problem was that when installation failed due to a corrupted lockfile or checksum mis-match, we wouldn't show the reason that gem was being installed in the first place, as we would when other errors happened.

See https://github.com/bundler/bundler/issues/5846

My diagnosis was we needed to show that requirement tree whenever installation fails, regardless of the reason for the failure.

My fix rescues hard installation errors and re-raises with the tree appended.


segiddins:
@bundlerbot r+

segiddins:
@bundlerbot retry

Issue #5552, opened 2017-03-31
Parallel installs using git sporadically fail


mdavidn:
Parallel `bundle install` processes on the same `Gemfile` will frequently (but not always) fail due to contention on a Git lock file. I noticed this because my CI environment starts parallel builds after every update to my project.

It appears that multiple processes running `GitProxy#copy_to` will both conclude `$destination/.git` doesn't exist before running `git reset --hard` in parallel on the same clone.

I expect parallel `bundle install` processes to wait for the first to finish with a shared resource. Alternatively, `GitProxy` could clone into a process-local temporary directory.

Here's a reproduction script and the error message:

https://gist.github.com/mdavidn/66027b9727f536f9f058108560e6b88f

```
Fetching git@github.com:bundler/bundler.git
fatal: Unable to create '/Users/xxxx/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/bundler/gems/bundler-ca5710e41dac/.git/index.lock': File exists.

Another git process seems to be running in this repository, e.g.
an editor opened by 'git commit'. Please make sure all processes
are terminated then try again. If it still fails, a git process
may have crashed in this repository earlier:
remove the file manually to continue.
Git error: command `git reset --hard ca5710e41dacefce8951f4992378d95b470882dc`
in directory
/Users/xxxx/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/bundler/gems/bundler-ca5710e41dac
has failed.
If this error persists you could try removing the cache directory
'/Users/xxxx/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/cache/bundler/git/bundler-f625eabf2e6cea03567de55cfa13168adc7de8ed'
```

colby-swandale:
@indirect @segiddins any thoughts on this?

segiddins:
Haven't had a chance to look into it yet

segiddins:
I'm sure this issue exists, seems like we might want to try using file locks on POSIX systems to serialize git operations? Just a thought, but I think someone else would have to try and implement it, I don't see myself having the time in the near future.

mikegee:
I think this issue can be closed because #5852 was merged.

PR #5852, opened 2017-07-08
Adding process lock for bundle install operations

### LINKED ISSUE #5851: Issues with concurrent bundle installs with shared gem cache
I am attempting to use a shared gem cache with a docker-compose setup, a sample project is available here: https://github.com/stefansedich/bundle-test which easily replicates the issue by simply running `run.sh`.

The error is consistent and one of the two containers will fail to compile the native extensions for in this case the pg gem I assume due to the concurrent installs attempting to create the Makefile.

What is the best way to handle this scenario, my current solution is to wrap my bundle install in a script that will use flock to limit concurrent installs.

stefansedich:

As per #5851 multiple concurrent bundle installs would cause failures when compiling

As described in #5851 a sample project was created demonstrating the issue when running a docker-compose solution of 2 containers using a gem cache volume

The fix is to implement an installation process lock so that only one process can execute an install at one time

I chose this fix because it was discussed in #5851 as a possible solution

### DIFF
--- a/lib/bundler/installer.rb
+++ b/lib/bundler/installer.rb
@@ -68,23 +68,25 @@ module Bundler
     def run(options)
       create_bundle_path
 
-      if Bundler.settings[:frozen]
-        @definition.ensure_equivalent_gemfile_and_lockfile(options[:deployment])
+      ProcessLock.lock do
+        if Bundler.settings[:frozen]
+          @definition.ensure_equivalent_gemfile_and_lockfile(options[:deployment])
+        end
+
+        if @definition.dependencies.empty?
+          Bundler.ui.warn "The Gemfile specifies no dependencies"
+          lock
+          return
+        end
+
+        resolve_if_needed(options)
+        ensure_specs_are_compatible!
+        warn_on_incompatible_bundler_deps
+        install(options)
+
+        lock unless Bundler.settings[:frozen]
+        Standalone.new(options[:standalone], @definition).generate if options[:standalone]
       end
-
-      if @definition.dependencies.empty?
-        Bundler.ui.warn "The Gemfile specifies no dependencies"
-        lock
-        return
-      end
-
-      resolve_if_needed(options)
-      ensure_specs_are_compatible!
-      warn_on_incompatible_bundler_deps
-      install(options)
-
-      lock unless Bundler.settings[:frozen]
-      Standalone.new(options[:standalone], @definition).generate if options[:standalone]
     end
 
     def generate_bundler_executable_stubs(spec, options = {})
--- a/lib/bundler/process_lock.rb
+++ b/lib/bundler/process_lock.rb
@@ -4,18 +4,21 @@ module Bundler
   class ProcessLock
     def self.lock(bundle_path = Bundler.bundle_path)
       lock_file_path = File.join(bundle_path, "bundler.lock")
+      has_lock = false
 
       File.open(lock_file_path, "w") do |f|
         f.flock(File::LOCK_EX)
-
+        has_lock = true
         yield
+        f.flock(File::LOCK_UN)
       end
-    rescue Errno::ENOLCK # NFS
-      raise if Thread.main != Thread.current
-
+    rescue Errno::EACCES, Errno::ENOLCK
+      # In the case the user does not have access to
+      # create the lock file or is using NFS where
+      # locks are not available we skip locking.
       yield
     ensure
-      File.delete(lock_file_path) if File.exist?(lock_file_path)
+      FileUtils.rm_f(lock_file_path) if has_lock
     end
   end
 end
--- a/lib/bundler/process_lock.rb
+++ b/lib/bundler/process_lock.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+module Bundler
+  class ProcessLock
+    def self.lock(bundle_path = Bundler.bundle_path)
+      lock_file_path = File.join(bundle_path, "bundler.lock")
+
+      File.open(lock_file_path, "w") do |f|
+        f.flock(File::LOCK_EX)
+
+        yield
+      end
+    rescue Errno::ENOLCK # NFS
+      raise if Thread.main != Thread.current
+
+      yield
+    ensure
+      File.delete(lock_file_path) if File.exist?(lock_file_path)
+    end
+  end
+end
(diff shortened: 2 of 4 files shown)


segiddins:
Please fill out the PR template, thanks!

lib/bundler/process_lock.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Bundler
+  class ProcessLock
+    def self.lock(bundle_path = Bundler.bundle_path)
+      lock_file_path = File.join(bundle_path, "bundler.lock")
+      File.open(lock_file_path, "w") do |f|
+        f.flock(File::LOCK_EX)
```
segiddins: this needs to handle when `flock` is unsupported

lib/bundler/installer.rb
```diff
@@ -5,6 +5,7 @@
 require "bundler/installer/parallel_installer"
 require "bundler/installer/standalone"
 require "bundler/installer/gem_installer"
+require "bundler/process_lock"
```
segiddins: please move this to an autoload under the bundler namespace in `bundler.rb`, since the file isn't in the `installer` subdirectory

lib/bundler/process_lock.rb
stefansedich: Do you now when it is not supported, in the reading I did it was said that it would work even under Windows.

lib/bundler/installer.rb
stefansedich: Done!

lib/bundler/process_lock.rb
segiddins: Older versions of Windows, I believe

stefansedich: OK I am going to have to dig further on this one, I had assumed it was well supported and could not find anything in the contrary either looking at the ruby source or docs.

indirect:
looks like this was a problem in rubygems a couple of years ago too: https://github.com/rubygems/rubygems/issues/806

stefansedich:
Thanks @indirect will check it out, heading out for a while but will get back to this later on tonight!

stefansedich:
@indirect I used the exact same solution as the mentioned issue, how do we feel about doing it in this way? the other solution is to bring in one of the gems that can handle this but it felt like it would be more trouble than it is worth?

indirect:
I'm okay with the same level of support as RubyGems has. @segiddins?

segiddins:
sounds good to me!

lib/bundler/process_lock.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Bundler
+  class ProcessLock
+    def self.lock(bundle_path = Bundler.bundle_path)
+      lock_file_path = File.join(bundle_path, "bundler.lock")
+      File.open(lock_file_path, "w") do |f|
```
segiddins: should we delete the file after the block has completed?

stefansedich: I don't think it can hurt, updated to delete at end safely only if the file exists.

lib/bundler/process_lock.rb
```diff
@@ -0,0 +1,21 @@
...
+  class ProcessLock
+    def self.lock(bundle_path = Bundler.bundle_path)
+      lock_file_path = File.join(bundle_path, "bundler.lock")
+
+      File.open(lock_file_path, "w") do |f|
+        f.flock(File::LOCK_EX)
+
+        yield
+      end
+    rescue Errno::ENOLCK # NFS
```
colby-swandale: This is not handling the use case when the user does not have permission to create a new file in the `bundle_path`

stefansedich: Good point, something I had not considered, I wonder if we want to actually handle it or if we just ignore using a lock in this scenario.

lib/bundler/process_lock.rb
stefansedich: On second thoughts @segiddins the delete could be an issue in the way I implemented it, you could have opened the file and awaiting a lock while someone else who had the lock has gone and deleted it from under you.

stefansedich: Ah I am wrong, there is still a file handle open so it won't actually be deleted.

lib/bundler/process_lock.rb
```diff
@@ -0,0 +1,24 @@
...
+  class ProcessLock
+    def self.lock(bundle_path = Bundler.bundle_path)
+      lock_file_path = File.join(bundle_path, "bundler.lock")
+      has_lock = false
+
+      File.open(lock_file_path, "w") do |f|
+        f.flock(File::LOCK_EX)
+        has_lock = true
+        yield
+        f.flock(File::LOCK_UN)
```
stefansedich: Looks like Ruby 1.8.7 needs the explicit unlock as closing the file does not seem to unlock the file like happens on other versions.

lib/bundler/process_lock.rb
```diff
@@ -0,0 +1,24 @@
...
+      end
+    rescue Errno::EACCES, Errno::ENOLCK
+      # In the case the user does not have access to
```
stefansedich: For now we simply ignore locking on NFS or if the user has no write permissions to the bundle_path.

stefansedich:
@indirect @segiddins is there usually issues with the 1.8.7 specs? Originally I did have an issue which I fixed with an explicit unlock, running the tests that fail locally under a fresh 1.8.7 install and everything is green. 

Looking at the builds the errors seem unrelated to my changes, and I did not experience any issues with the full suite both locally and in a fresh 1.8.7 container.

segiddins:
The 1.8.7 specs are green on master

segiddins:
The tests are fixed now on master, so when we go to merge this should pass

stefansedich:
Thanks @segiddins I will merge master into mine anyway to kick off a build for sanity's sake!

mtsmfm:
I think this PR also fixes https://github.com/bundler/bundler/issues/5552 👍

stefansedich:
@segiddins is there anything else needed from me to get this one in? would love to drop my hacky scripts :)

segiddins:
I requested a review from @indirect so I'd like him to take a look and approve the PR

indirect:
looks good to me, let's do it.

@bundlerbot r+

PR #5853, opened 2017-07-09
Change deprecation from `bundle list` to `bundle show`

koic:

I understand that `bundle show` will be removed in Bundler 2.0, not `bundle list`. 

https://github.com/bundler/rfcs/pull/6/files#r125875779

In the current message, there is a possibility of misleading the command which will be removed in Bundler 2.0.

I think that it is better to display `bundle show` instead of `bundle list` in the deprecated warning.

Change the deprecated warning from `bundle list` to `bundle show`.

I thought that changing the console message was a simple and effective change to the end user.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -245,11 +245,11 @@ module Bundler
     method_option "outdated", :type => :boolean,
                               :banner => "Show verbose output including whether gems are outdated."
     def show(gem_name = nil)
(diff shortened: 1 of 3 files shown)


segiddins:
@colby-swandale please review?

colby-swandale:
This looks fine but @indirect are we ok to change an existing deprecation message?

indirect:
Yeah, I think this is fine... printing a deprecation message doesn't count as semver imo 👍🏻

@koic thanks!

@bundlerbot r+

koic:
Thank you all!

PR #5855, opened 2017-07-10
Limit specs to rack-test < 0.7.0 for Ruby 1.8.7 compatibility

segiddins:

The problem was the specs were broken by the release of `rack-test 0.7.0`.

My diagnosis was `rack-test` has intentionally dropped 1.8.7 support, though we didn't catch that immediately because they didn't put a `required_ruby_version` in their gemspec.

My fix pins us to older versions of the gem.


segiddins:
@bundlerbot r+

colby-swandale:
@bundlerbot retry

PR #5858, opened 2017-07-11
Add some unit tests for lockfile parsing

segiddins:

The problem was the lockfile parser was missing some robustness unit tests.

My fix was to add test coverage!


segiddins:
@bundlerbot r+

PR #5860, opened 2017-07-12
Introduce the notion of settings validation & default install to ./.bundle

segiddins:

The problem was...

My diagnosis was...

My fix...

I chose this fix because...

### DIFF
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -154,8 +154,8 @@ module Bundler
 
       check_for_group_conflicts_in_cli_options
 
-      Bundler.settings.set_command_option_if_given :with, options[:with]
-      Bundler.settings.set_command_option_if_given :without, options[:without]
+      Bundler.settings.set_command_option :with, nil if options[:with] == []
+      Bundler.settings.set_command_option :without, nil if options[:without] == []
 
       with = options.fetch(:with, [])
       with |= Bundler.settings[:with].map(&:to_s)
@@ -189,8 +189,13 @@ module Bundler
 
       Bundler.settings.set_command_option_if_given :clean, options["clean"]
 
-      Bundler.settings.set_command_option :without, options[:without] unless Bundler.settings[:without] == options[:without]
-      Bundler.settings.set_command_option :with,    options[:with]    unless Bundler.settings[:with]    == options[:with]
+      unless Bundler.settings[:without] == options[:without] && Bundler.settings[:with] == options[:with]
+        # need to nil them out first to get around validation for backwards compatibility
+        Bundler.settings.set_command_option :without, nil
+        Bundler.settings.set_command_option :with,    nil
+        Bundler.settings.set_command_option :without, options[:without] - options[:with]
+        Bundler.settings.set_command_option :with,    options[:with]
+      end
 
       disable_shared_gems = Bundler.settings[:path] ? true : nil
       Bundler.settings.set_command_option :disable_shared_gems, disable_shared_gems unless Bundler.settings[:disable_shared_gems] == disable_shared_gems
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -70,19 +70,19 @@ module Bundler
       Bundler.ui.confirm "Bundle complete! #{dependencies_count_for(definition)}, #{gems_installed_for(definition)}."
       Bundler::CLI::Common.output_without_groups_message
 
-      if path = Bundler.settings[:path]
-        absolute_path = File.expand_path(path)
-        relative_path = absolute_path.sub(File.expand_path(".") + File::SEPARATOR, "." + File::SEPARATOR)
-        Bundler.ui.confirm "Bundled gems are installed into #{relative_path}."
-      else
+      if Bundler.use_system_gems?
         Bundler.ui.confirm "Use `bundle info [gemname]` to see where a bundled gem is installed."
+      else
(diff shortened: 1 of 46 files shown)


lib/bundler/settings/validator.rb
```diff
@@ -0,0 +1,79 @@
...
+      private_class_method :rule
+
+      def self.validate(key, value, settings)
+        rules_to_validate = rules[key]
+        rules_to_validate.each {|rule| rule.validate(key, value, settings) }
+      end
+
+      rule %w[path disable_shared_gems], "path and disable_shared_gems are mutually exclusive" do |key, value, settings|
+        if key == "path" && value
+          set(settings, :disabled_shared_gems, nil)
```
jules2689: Might be nice to output the thing done. e.g. `puts "Setting disabled_shared_gems to nil to resolve conflict`

lib/bundler/settings/validator.rb
```diff
@@ -0,0 +1,79 @@
...
+        attr_reader :description
+
+        def initialize(keys, description, &validate)
+          @keys = keys
+          @description = description
+          @validate = validate
+        end
+
+        def validate(key, value, settings)
+          instance_exec(key, value, settings, &@validate)
```
jules2689: I've always been partial to the syntax `@validate.call(key, value, settings)` using a proc. I find it easier to parse than `exec`s.

lib/bundler/settings/validator.rb
```diff
@@ -0,0 +1,79 @@
...
+      class Rule
+        attr_reader :description
+
+        def initialize(keys, description, &validate)
+          @keys = keys
+          @description = description
+          @validate = validate
+        end
+
+        def validate(key, value, settings)
```
jules2689: `validate` seems to usually do more "dangerous" things like "fail!" and such. Would `validate!` make more sense?

lib/bundler/settings/validator.rb
```diff
@@ -0,0 +1,79 @@
...
+          reasons.unshift @description
+          Bundler.ui.info "Setting `#{key}` to #{value.inspect}, since #{reasons.join(", ")}"
+          if value.nil?
+            settings.delete(hash_key)
+          else
+            settings[hash_key] = value
+          end
+        end
+
+        def k(key)
```
jules2689: if we send in `key_for(key)` in the call, then we dont need to do a private call

lib/bundler/settings.rb
```diff
@@ -279,24 +289,25 @@ def array_to_s(array)
...
-            p.open("w") {|f| f.write(YAMLSerializer.dump(hash)) }
-          end
-        end
-      end
+      return if hash[key] == value
+
+      hash[key] = value
+      hash.delete(key) if value.nil?
+
+      Validator.validate(raw_key, converted_value(value, raw_key), hash)
```
jules2689: should we validate before modifying state?

lib/bundler/cli/install.rb
```diff
@@ -182,8 +182,13 @@ def normalize_settings
 
       Bundler.settings.set_command_option_if_given :clean, options["clean"]
 
-      Bundler.settings.set_command_option :without, options[:without] unless Bundler.settings[:without] == options[:without]
-      Bundler.settings.set_command_option :with,    options[:with]    unless Bundler.settings[:with]    == options[:with]
+      unless Bundler.settings[:without] == options[:without] && Bundler.settings[:with] == options[:with]
```
jules2689: What was the purpose of this again?

lib/bundler/settings/validator.rb
segiddins: The set method defaults to including the rules description 😄

lib/bundler/settings/validator.rb
segiddins: They're not the same -- this changes the binding the block is evaluated in to the of the Rule instance, making its methods available

lib/bundler/settings/validator.rb
segiddins: But we want to display the lowercased key in messages to the user

jules2689: Isn't passing `key_for` in the same as calling `Bundler.settings.send(:key_for, key)`?

segiddins: I mean that we need access to both the raw / user-visible keys (e.g. `path`) _and_ the transformed key used for storage (e.g. `BUNDLE_PATH`)

lib/bundler/settings.rb
segiddins: I think it's easier to do it afterward -- it means rules can validate multiple things at once just by looking at the passed-in hash

lib/bundler/cli/install.rb
segiddins: It keeps the 1.0 tests passing, because 1.0 has super weird behavior around setting groups

lib/bundler/settings/validator.rb
jules2689: I think it's nicer to pass both in, tbh. I'm a little allergic to `.send` due to some issues in the past with it :(

segiddins: I'm going to make the method public in settings 🙃

indirect:
This looks great to me... giving users clear and explicit errors when the settings don't cohere is :100:. r+ once the tests pass

segiddins:
Green!
@bundlerbot r+

PR #5862, opened 2017-07-12
Remove uneeded files packaged into bunder's gem build on rubygems

NickLaMuro:

I noticed while doing a `bundle open bundler` (inception!) that we have included the `bin/` directory in the packaged gem.  While this is not hurting anything, it provides a bunch of scripts that won't function properly in any capacity when installed through ruby gems.

While doing that, I noticed that the dotfiles for CI were also packaged up in the gem, so I made sure to only include specific files from there that seemed to provide value when being shipped as a gem.

Change the reduction of the results from `git ls-files` to be a whitelist instead of a blacklist.  This was to reduce the complexity of the regexp that would have resulted from change what directories to include, not ship CI files in the top level dir, while also still wanting to ship the `CHANGELOG.md`, `LICENSE.md`, and `README.md` files along with the build, since I think they do provide value being included with the packaged gem.

Some of the choices here were a bit subjective, especially what I decided on as "necessary" regarding the `.md` files that I included.  Projects like rails went with these three, so I decided to use that as a base, but some might think that the `CHANGELOG.md` being included is a bit excessive and is a large file that most won't even read in this form, and I don't have a hard opinion one way or the other.

That said, I do think that developer files should not be packaged in the gem, and this precedent has already been set with the original blacklist, it just wasn't very complete.  How I did it could probably be changed based on developer preference, but I tried to stay as true to what was already there before all else.

### DIFF
--- a/bundler.gemspec
+++ b/bundler.gemspec
@@ -37,7 +37,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency "ronn",       "~> 0.7.3"
   s.add_development_dependency "rspec",      "~> 3.6"
 
-  s.files = `git ls-files -z`.split("\x0").reject {|f| f.match(%r{^(test|spec|features)/}) }
+  s.files = `git ls-files -z`.split("\x0").select {|f| f.match(%r{^(lib|exe)/}) }
   # we don't check in man pages, but we need to ship them because
   # we use them to generate the long-form help for each command.
   s.files += Dir.glob("man/**/*")
--- a/bundler.gemspec
+++ b/bundler.gemspec
@@ -41,6 +41,8 @@ Gem::Specification.new do |s|
   # we don't check in man pages, but we need to ship them because
   # we use them to generate the long-form help for each command.
   s.files += Dir.glob("man/**/*")
+  # Include the CHANGELOG.md, LICENSE.md, README.md manually
+  s.files += %w[CHANGELOG.md LICENSE.md README.md]
 
   s.bindir        = "exe"
   s.executables   = %w[bundle bundler]


bundler.gemspec
```diff
@@ -37,10 +37,12 @@ Gem::Specification.new do |s|
   s.add_development_dependency "ronn",       "~> 0.7.3"
   s.add_development_dependency "rspec",      "~> 3.6"
 
-  s.files = `git ls-files -z`.split("\x0").reject {|f| f.match(%r{^(test|spec|features)/}) }
+  s.files = `git ls-files -z`.split("\x0").select {|f| f.match(%r{^(lib|exe)/}) }
   # we don't check in man pages, but we need to ship them because
   # we use them to generate the long-form help for each command.
   s.files += Dir.glob("man/**/*")
+  # Include the CHANGELOG.md, LICENSE.md, README.md manually
+  s.files += %w[CHANGELOG.md LICENSE.md README.md]
```
segiddins: why not include all of the `.md` files?

NickLaMuro: Thought that the `CODE_OF_CONDUCT.md` and `CONTRIBUTING.md` wouldn't be useful with a rubygems installed `bundler`, and more useful to a developer cloning this repo with the intent to contribute.  That said, the `TROUBLESHOOTING.md` doc mentioned in the `README.md` might be useful, so there is that.

Happy to add or remove things as you see fit, just picked a few the seemed the most relevant to start.

segiddins:
@indirect this one's for you

indirect:
@NickLaMuro cool! Can you comment with a diff of what files are in the gem now, and what files will be in the gem if we accept this PR?

NickLaMuro:
Sure, give me a few to get that worked up.

segiddins:
```diff
1,8d0
< .codeclimate.yml
< .github/PULL_REQUEST_TEMPLATE.md
< .github/welcome_message.md
< .gitignore
< .rspec
< .rubocop.yml
< .rubocop_todo.yml
< .travis.yml
10,11d1
< CODE_OF_CONDUCT.md
< CONTRIBUTING.md
14,37d3
< Rakefile
< bin/bundle
< bin/bundle2
< bin/rake
< bin/rspec
< bin/rubocop
< bin/with_rubygems
< bundler.gemspec
< doc/README.md
< doc/TROUBLESHOOTING.md
< doc/contributing/BUG_TRIAGE.md
< doc/contributing/COMMUNITY.md
< doc/contributing/GETTING_HELP.md
< doc/contributing/HOW_YOU_CAN_HELP.md
< doc/contributing/ISSUES.md
< doc/contributing/README.md
< doc/development/NEW_FEATURES.md
< doc/development/PULL_REQUESTS.md
< doc/development/README.md
< doc/development/RELEASING.md
< doc/development/SETUP.md
< doc/documentation/README.md
< doc/documentation/VISION.md
< doc/documentation/WRITING.md
122d87
< lib/bundler/process_lock.rb
134d98
< lib/bundler/settings/validator.rb
156d119
< lib/bundler/templates/Executable.bundler
318,321d280
< task/build_metadata.rake
< task/bundler_2.rake
< task/git_hooks.rake
< task/release.rake
```

segiddins:
Seems legit 👍

NickLaMuro:
Nevermind... someone got me beat ;)

indirect:
This seems good, thanks for doing it! I think that skipping `bin/` is the biggest improvement, but this is a good change overall.

@bundlerbot r+

PR #5864, opened 2017-07-16
[RuboCop] Enable Layout/EmptyLineAfterMagicComment cop

koic:
This PR makes cosmetic changes. Layout/EmptyLineAfterMagicComment cop is enabled by default in RuboCop as well.

https://github.com/bbatsov/rubocop/blob/v0.49.1/config/enabled.yml#L203-L206

This PR executed the following command to the source codes.

```console
% rubocop -a --only Layout/EmptyLineAfterMagicComment
```


.rubocop.yml
```diff
@@ -38,6 +38,9 @@ Layout/SpaceInsideBlockBraces:
 Layout/DotPosition:
   EnforcedStyle: trailing
 
+Layout/EmptyLineAfterMagicComment:
```
colby-swandale: This does not need to be set if the rule is enabled by default

koic: Oops! You're absolutely right. I updated and rebased this PR.

segiddins:
@bundlerbot r+

PR #5867, opened 2017-07-17
[Rakefile] In a patch release, change version after checkout

segiddins:

The problem was changes to the versions file on the `master` branch would cause checkout to fail


segiddins:
@bundlerbot r+

PR #5870, opened 2017-07-18
Fix spec on 2.0 broken with the v1.15.2 merge

segiddins:

The problem was I merged in a change necessary to keep the 1.15.x specs passing accidentally, even though it broke the 2.0 spec. Oops.


segiddins:
@bundlerbot r+

PR #5871, opened 2017-07-18
Return an empty array if no extensions are found

greysteil:

Attempting to compare endpoint specifications that didn't have any extensions was causing a `NoMethodError: undefined method `empty?' for nil:NilClass` error.

The `extensions` method on `EndpointSpecification` was incorrectly returning `nil` when there were no extensions.

~~Return an empty array in that case, instead~~ Fall back to the base class implementation in that case, instead.

~~Alternative would be to call `super`. I presumed there was a reason that wasn't being done already, but it would also work.~~ Alternative would be to explicitly return an empty array, but all other methods call `super`, which feels more robust.

### DIFF
--- a/spec/bundler/endpoint_specification_spec.rb
+++ b/spec/bundler/endpoint_specification_spec.rb
@@ -7,7 +7,7 @@ RSpec.describe Bundler::EndpointSpecification do
   let(:dependencies) { [] }
   let(:metadata)     { nil }
 
-  subject { described_class.new(name, version, platform, dependencies, metadata) }
+  subject(:spec) { described_class.new(name, version, platform, dependencies, metadata) }
 
   describe "#build_dependency" do
     let(:name)           { "foo" }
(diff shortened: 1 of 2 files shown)


colby-swandale:
This is missing a test for the change

greysteil:
@colby-swandale - updated. Also moved the implementation to use `super`, since all the other methods here do.

spec/bundler/endpoint_specification_spec.rb
```diff
@@ -62,4 +62,8 @@
       end
     end
   end
+
+  it "supports equality comparison" do
+    expect(subject.eql?(subject)).to eq(true)
```
segiddins: you can use `expect(subject).to eel(subject)` here, although I think it would be helpful to test `eql` against a different instance?

greysteil:
@segiddins - updated.

segiddins:
Awesome!

@bundlerbot r+

PR #5873, opened 2017-07-18
Ensure deprecations become a hard error in Bundler 2

segiddins:

The problem was some deprecations aren't guaranteed to be behind a feature flag, so we want to guarantee that we won't continue to support those behaviors when they're intended to be removed.

My diagnosis was we needed to ensure that deprecations don't stay deprecations past the point they're supported.

My fix adds a `major_version` argument to the deprecation method, and raises an error if the current version is `>=` that version

I chose this fix because it will ensure that everything deprecated won't continue to work past when it's supposed to, reducing the chance we'll need to accidentally support things we won't want to.


colby-swandale:
Correct me if i'm wrong but i thought the Gemfile/Gemfile.lock would be deprecated in bundler 2 and then be removed in 3?

segiddins:
@colby-swandale I thought we were just switching the precedence order?

colby-swandale:
Pushed fixes for the breaking in bundler 2. Did you want to tackle the last breaking spec or if you want, I'll just finish it off?

segiddins:
Looks like there are still a whole bunch of failures on 2.0: https://travis-ci.org/bundler/bundler/jobs/260183748#L1425

colby-swandale:
I'll tackle these tonight

colby-swandale:
I'm waiting on #5920 to be merged which fixes a lot fo the bundler 2 errors for free, i'll tackle the rest afterwards.

segiddins:
@colby-swandale it's been merged!

colby-swandale:
I think I screwed up the rebase because Travis is not pulling the newer code, I'll have to have a look at it tomorrow. Sorry 🙇

colby-swandale:
@segiddins is there a test that https://github.com/bundler/bundler/pull/5873/commits/95bb71c9b9d71f0cec1c4693b12a75ae9fdaa32c is solving? This is causing the failures for  bundler `< 2` but all the tests before that commit pass?

segiddins:
It fixes a bug that caused the deprecation warning to _always_ be printed

spec/install/binstubs_spec.rb
```diff
@@ -35,12 +35,6 @@ def Gem.bindir; "/usr/bin"; end
       G
     end
 
-    it "prints a deprecation notice" do
```
colby-swandale: I removed this test because the `--binstubs` has been removed into bundler 2

spec/runtime/with_clean_env_spec.rb
```diff
@@ -3,8 +3,9 @@
 RSpec.describe "Bundler.with_env helpers" do
   describe "Bundler.original_env" do
     before do
+      bundle "config path vendor/bundle"
       gemfile ""
-      bundle "install --path vendor/bundle"
+      bundle "install"
     end
 
     it "should return the PATH present before bundle was activated" do
```
colby-swandale: @segiddins so this is breaking on travis because it's ships a version of ruby 1.9 that sits in `/usr/bin`. And the `bundle!` is running on that install of ruby. Do you have a quick thoughts on how to address this?

segiddins: I think I fixed this in another PR but can't remember. https://github.com/bundler/bundler/pull/5720/commits/d3e08aaa44645450af025f176e3c31d58c51e8ec

segiddins:
Rebased, might finally be green?

colby-swandale:
🎉

segiddins:
@bundlerbot r+

Issue #5876, opened 2017-07-19
Special-case `binstubs bundler` for the app's bundler version


indirect:
Once Bundler 2 ships, each application will depend on a 1.x or 2.x version of Bundler. One way to ensure that the application will always get the version of Bundler that it needs is with an application-and-bundler-version specific binstub. Let's make it so that `bundle binstubs bundler` will generate a binstub for the correct version of Bundler, and simply running `bin/bundle` will always provide the correct version of Bundler.

segiddins:
Should we also change the normal binstubs (that call `require "bundler/setup"`) to make sure the correct bundler version is activated?

indirect:
Good question. I'm not sure, but it sounds like probably we will want this, yes. Ideally in a dynamic way, so that the binstubs don't need to be regenerated after the app migrates from Bundler 1 to Bundler 2? But if we can't do that, we should provide a command that iterates over all existing binstubs and upgrades them from Bundler 1 to Bundler 2.

segiddins:
`bundle binstubs ... --force` will always overwrite binstubs with the newest version

PR #5878, opened 2017-07-20
Add a special bundler binstub that ensures the correct version is activated

segiddins:

Closes https://github.com/bundler/bundler/issues/5876.

> Once Bundler 2 ships, each application will depend on a 1.x or 2.x version of Bundler.

We want to make sure that the user always has an _easy_ and _convenient_ way to activate the version of Bundler their app is depending upon.

> One way to ensure that the application will always get the version of Bundler that it needs is with an application-and-bundler-version specific binstub. Let's make it so that bundle binstubs bundler will generate a binstub for the correct version of Bundler, and simply running bin/bundle will always provide the correct version of Bundler.

My fix implements a `bundle` binstub when `bundle binstubs bundler` is run

I chose this fix because it allows us to dynamically select the correct Bundler version (with appropriate overrides, such as `$BUNDLER_VERSION` and `bundle update --bundler`), both when invoking `bin/bundle`, and also when invoking other bundler-generated binstubs.


lib/bundler/templates/Executable.bundler
```diff
@@ -0,0 +1,71 @@
...
+    return gemfile if gemfile && !gemfile.empty?
+
+    File.expand_path("../<%= relative_gemfile_path %>", __FILE__)
+  end
+
+  def lockfile
+    File.expand_path case File.basename(gemfile)
+                      when 'gems.rb' then gemfile.sub(/\.rb$/, gemfile)
+                      else "#{gemfile}.lock"
+                      end
```
indirect: Can you use a variable so the case statement body isn't indented?

segiddins: Sure, this was lifted directly from postit 👍🏻

segiddins:
I actually have something different in mind for RubyGems, which will handle requiring Bundler/setup as well

segiddins:
@indirect I believe this should work now, and I've made sure to run the file through rubocop

segiddins:
Wow, green within a day! That's like a new record for me

segiddins:
@indirect r?

indirect:
Awesome. @bundlerbot r+

PR #5880, opened 2017-07-20
Use ssl client cert and ca cert settings from gem configuration as fallbacks

stan3:

The problem was having to configure ssl client cert and ca cert in bundler and gem.

My fix makes bundler check the gem configuration for these settings.

I chose this fix because it makes sense for the user to only configure these once.

Couldn't find existing coverage for this in the tests so haven't added any...

Also assume a changelog entry is needed but can't see a unreleased changes section so not sure how this is handled.

### DIFF
--- a/lib/bundler/fetcher.rb
+++ b/lib/bundler/fetcher.rb
@@ -249,8 +249,11 @@ module Bundler
           con.cert_store = bundler_cert_store
         end
 
-        if Bundler.settings[:ssl_client_cert]
-          pem = File.read(Bundler.settings[:ssl_client_cert])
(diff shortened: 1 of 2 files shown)


colby-swandale:
Hi! thanks for opening a helping make Bundler better. This is going to need a test.

stan3:
added some tests

segiddins:
This seems to break on old RubyGems, looking at the failures on CI

stan3:
thanks, I'll look at adding a conditional

indirect:
This looks like a useful change, thanks! @bundlerbot r+

colby-swandale:
@bundlerbot retry

Issue #5643, opened 2017-05-05
CGI escaped username and passwords with bundle config


giovdk21:
Please see: https://github.com/bundler/bundler/issues/4288

> Let's specifically call out that usernames and passwords need to already be CGI escaped.
_(comment by indirect)_ 

I've been having the same problem using bundle config bitbucket.org user:pwd (http://bundler.io/man/bundle-config.1.html)

Could the documentation be updated by specifying that also in relation to the "bundle config" command / env. variables?

Thanks

segiddins:
Yup, we definitely should call this out at https://github.com/bundler/bundler/blob/master/man/bundle-config.ronn#L291!

mikegee:
This issue is resolved since https://github.com/bundler/bundler/pull/5883 was merged.

PR #5883, opened 2017-07-20
Add line re: CGI escape

rubymorillo:
Added a sentence in line 291 re: CGI escaping passwords and usernames. Resolves issue: https://github.com/bundler/bundler/issues/5643

The problem was needing to make it clear that users had to CGI escape usernames and passwords before pushing to remote. Referenced here: https://github.com/bundler/bundler/issues/5643

My diagnosis was to add a sentence to the man page as a reminder.

My fix was to add a sentence to the man page :)

I chose this fix because it would make things clearer for users.


colby-swandale:
@bundlerbot r+

PR #5884, opened 2017-07-20
[CLI] Dont print an outdated version warning when running a parseable command

segiddins:

The problem was the outdated bundler version warning could be printed when running a command whose output should be parseable.

@schneems pointed this out on twitter

My diagnosis was parsable commands needed to have such extraneous output suppressed.

My fix was to use the blacklist from printing the running command to bail out on the outdated bundler version warning as well.

I chose this fix because it means we can have a single list of parseable command.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -6,6 +6,9 @@ require "bundler/vendored_thor"
 module Bundler
   class CLI < Thor
     AUTO_INSTALL_CMDS = %w[show binstubs outdated exec open console licenses clean].freeze
+    PARSEABLE_COMMANDS = %w[
+      check config help exec platform show version
+    ].freeze
 
     def self.start(*)
       super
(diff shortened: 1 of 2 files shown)


schneems:
Awesome, thanks!

FWIW I can't repro locally. The issue only happens a developer accidentally has specified 2 ruby buildpacks for example

- heroku/ruby
- https://github.com/heroku/heroku-buildpack-ruby

There may be other corner cases but that is one commonality. Currently `heroku/ruby` is using 1.15.1 and master is on 1.15.2.

We may also set the flag to disable that check all together on the platform. Either way having output from this command be parsable is a concern into the future.

segiddins:
how to repro locally (I'm guessing):

1. `gem install bundler -v 1.15.1`
2. install a gemfile that has `gem "bundler"`
3. `bundle _1.15.1_ platform --ruby`

segiddins:
@bundlerbot r+

Issue #5881, opened 2017-07-20
TypeError: no implicit conversion of Hash into String


s0rin:
--- ERROR REPORT TEMPLATE -------------------------------------------------------
# Error Report

## Questions

Please fill out answers to these questions, it'll help us figure out
why things are going wrong.

- **What did you do?**

I ran the command `/var/www/.rbenv/versions/2.3.0/bin/bundle install`

- **What did you expect to happen?**

  I expected Bundler to complete the installation without errors

- **What happened instead?**

  Instead, Bundler failed with `TypeError: no implicit conversion of Hash into String` error

- **Have you tried any solutions posted on similar issues in our issue tracker, stack overflow, or google?**

  I tried...

- **Have you read our issues document, https://github.com/bundler/bundler/blob/master/doc/contributing/ISSUES.md?**

  ...

## Backtrace

```
TypeError: no implicit conversion of Hash into String
  /var/www/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/bundler-1.15.2/lib/bundler/cli/install.rb:75:in `expand_path'
  /var/www/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/bundler-1.15.2/lib/bundler/cli/install.rb:75:in `run'
  /var/www/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/bundler-1.15.2/lib/bundler/cli.rb:186:in `block in install'
  /var/www/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/bundler-1.15.2/lib/bundler/settings.rb:92:in `temporary'
  /var/www/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/bundler-1.15.2/lib/bundler/cli.rb:185:in `install'
  /var/www/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/bundler-1.15.2/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
  /var/www/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/bundler-1.15.2/lib/bundler/vendor/thor/lib/thor/invocation.rb:126:in `invoke_command'
  /var/www/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/bundler-1.15.2/lib/bundler/vendor/thor/lib/thor.rb:369:in `dispatch'
[...]

PR #5885, opened 2017-07-20
[YAMLSerializer] Handle empty strings properly

segiddins:

The problem was that empty strings in YAML were being read in as empty hashes.

Closes https://github.com/bundler/bundler/issues/5881.

My diagnosis was we needed to fix the YAML Serializer to understand `""` as an empty string.

My fix ensures that a hash isn't created when the parser encounters `""`.


segiddins:
@bundlerbot r+

PR #5888, opened 2017-07-20
Adding graphs explaining binstubs

rubymorillo:
Added a paragraph that explains what binstubs is and how it works, per feedback from @arbonap.

The problem was...without an explanation of what binstubs is and what it does, it's difficult for 
users to understand where to apply the binstubs command. Shout out to @arbonap for calling this out!

My diagnosis was...to take a wonderful explanation given by @indirect about what binstubs are and turn it into a paragraph! 😄 

My fix...is two intro paragraphs in the "DESCRIPTION" section, before going into what the command does.

I chose this fix because...explanations help users 👍

### DIFF
--- a/man/bundle-binstubs.ronn
+++ b/man/bundle-binstubs.ronn
@@ -7,15 +7,15 @@ bundle-binstubs(1) -- Install the binstubs of the listed gems
 
 ## DESCRIPTION
 
-Binstubs are scripts that wrap aroung executables. Bundler creates a 
-small Ruby file (a binstub) that loads Bundler, runs the command, 
+Binstubs are scripts that wrap aroung executables. Bundler creates a
+small Ruby file (a binstub) that loads Bundler, runs the command,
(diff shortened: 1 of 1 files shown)


indirect:
@rubymorillo one tiny tweak:

> with the additional benefit of linking the binstub inside of an application to the exact version of the gem

Using bundle exec or binstubs are two different ways to get the same result of the exact version, so maybe instead "offering a file that can be run directly that will always run the correct version of the gem used by the application", or something like that?

segiddins:
Test failures: https://travis-ci.org/bundler/bundler/jobs/255783590#L719

rubymorillo:
@segiddins @indirect made the recommended changes on the file; I'm a newb at this PR stuff so let me know if there's any lingering issues!

indirect:
Looks great. Thanks, @rubymorillo!

@bundlerbot r+

PR #5890, opened 2017-07-20
[RubyGemsIntegration] Require bundler/gemdeps before using it

segiddins:

The problem was `Bundler.rubygems.use_gemdeps` would raise a missing constant error.

My diagnosis was we needed to require the gemdeps file

My fix requires the gemdeps file.

I chose this fix because watching autocorrect struggle to understand that `gemdeps` is not the same as `genders` is priceless.


colby-swandale:
@bundlerbot r+

PR #5893, opened 2017-07-22
Add spec for prerelease dependency resolution

greysteil:
Bundler-specific version of https://github.com/CocoaPods/Resolver-Integration-Specs/pull/12. We didn't have any specs for resolving pre-release versions before, and I'm nervous about https://github.com/CocoaPods/Molinillo/pull/69 breaking this when introduced.


segiddins:
We also have https://github.com/bundler/bundler/blob/master/spec/install/prereleases_spec.rb

@bundlerbot r+

Issue #5895, opened 2017-07-24
[2.0] Expand user cache to include git repos as well as downloaded gems


indirect:
Now that we have a "global" user-wide cache of gems, it would be great to also have a user-wide place to keep git repos. In 1.x, the default system path provided that user-wide cache, and losing it because of the default 2.0 path will slow down installs with git repos by a lot.

PR #5896, opened 2017-07-24
[Source::Git] Cache repos globally

segiddins:

Closes https://github.com/bundler/bundler/issues/5895.

The problem was:

> Now that we have a "global" user-wide cache of gems, it would be great to also have a user-wide place to keep git repos. In 1.x, the default system path provided that user-wide cache, and losing it because of the default 2.0 path will slow down installs with git repos by a lot.

My diagnosis was we needed to change the git cache location to something global

My fix uses the global "~/.bundle/cache" directory to cache git gems

I chose this fix because it means the cache (which only contains _downloaded_ information, nothing evaluated at runtime) can be shared across every app and every ruby version

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -207,10 +207,6 @@ module Bundler
       bundle_path.join("specifications")
     end
 
-    def cache
-      bundle_path.join("cache/bundler")
-    end
-
     def user_cache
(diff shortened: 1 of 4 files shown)


segiddins:
Rebased

indirect:
nice! @bundlerbot r+

PR #5897, opened 2017-07-25
Add clear error message when using a gem in the Gemfile with an empty name

colby-swandale:

Users are allowed to add a gem to the Gemfile with an empty string as the name but Bundler won't find a gem with an empty and does not make this clear as possible to the end user. Using a Gemfile like:

```
source 'https://rubygems.org'
gem '', git: 'https://github.com/rack/rack'
```

Bundler will print the following:

```
› dbundler install
Fetching https://github.com/rack/rack
Fetching gem metadata from https://rubygems.org/.
Fetching gem metadata from https://rubygems.org/...
Could not find gem '' in https://github.com/rack/rack (at master@8693330).
The source does not contain any versions of ''

```

Check the name of the gem in the DSL. If the name is empty print a clear message to the user that using an empty name is not allowed

```
[!] There was an error parsing `Gemfile`: an empty gem name is not valid. Bundler cannot continue.

 #  from /Users/c/Desktop/Projects/testapp/Gemfile:2
 #  source 'https://rubygems.org'
 >  gem '', git: 'https://github.com/rack/rack'
```

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -268,6 +268,11 @@ Style/IfInsideElse:
     - 'lib/bundler/cli/install.rb'
 
 # Offense count: 1
+Style/IfUnlessModifier:
+  Exclude:
+    - 'lib/bundler/dsl.rb'
+
+# Offense count: 1
 Style/IfUnlessModifierOfIfUnless:
   Exclude:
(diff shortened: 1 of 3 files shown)


segiddins:
Awesome, @bundlerbot r+

Issue #5381, opened 2017-01-30
Bundler 1.14 is no longer correctly producing compatible version output for Ruby 1.8.7


JonRowe:
On Ruby 1.8.7 Bundler 1.14 no longer produces the correct "compatible version" output. Producing this:

```
Bundler could not find compatible versions for gem "ruby":
  In Gemfile:
    ruby
    rails (~> 3.2.17) was resolved to 3.2.22.5, which depends on
      ruby (>= 1.8.7)
    rails (~> 3.2.17) was resolved to 3.2.22.5, which depends on
      ruby (>= 1.8.7)
    rails (~> 3.2.17) was resolved to 3.2.22.5, which depends on
      ruby (>= 1.8.7)
    rails (~> 3.2.17) was resolved to 3.2.22.5, which depends on
      ruby (>= 1.8.7)
    rails (~> 3.2.17) was resolved to 3.2.22.5, which depends on
      ruby (>= 1.8.7)
    rails (~> 3.2.17) was resolved to 3.2.22.5, which depends on
      ruby (>= 1.8.7)
    rails (~> 3.2.17) was resolved to 3.2.22.5, which depends on
      ruby (>= 1.8.7)
    rails (~> 3.2.17) was resolved to 3.2.22.5, which depends on
      ruby (>= 1.8.7)
    rails (~> 3.2.17) was resolved to 3.2.22.5, which depends on
      ruby (>= 1.8.7)
    rails (~> 3.2.17) was resolved to 3.2.22.5, which depends on
      ruby (>= 1.8.7)
    rails (~> 3.2.17) was resolved to 3.2.22.5, which depends on
      ruby (>= 1.8.7)
    rails (~> 3.2.17) was resolved to 3.2.22.5, which depends on
      ruby (>= 1.8.7)
Could not find gem 'ruby (>= 1.9.2)', which is required by gem 'rails (~>
3.2.17)', in any of the sources.
```

When the actual gem at fault was `rest-client` a dependency of `relish` which required 1.9.2, this makes it very hard to debug what dependency is conflicting. Switching back to `1.13.x` solved this and correctly produced:

`rest-client-1.8.0 requires ruby version >= 1.9.2, which is incompatible with the current version, ruby 1.8.7p374`

[...]

PR #5900, opened 2017-07-26
[Env] Print all gemfiles that have been evaled

segiddins:

The problem was we'd get error reports like https://github.com/bundler/bundler/issues/5381 where the env output was insufficient to reproduce the issue because `eval_gemfile` had been used.

My diagnosis was we needed to print out all gem files that had gone through `eval_gemfile`

My fix keeps track of all eval'ed gem files and prints them all out in the env report.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -15,7 +15,9 @@ module Bundler
       :locked_gems,
       :platforms,
       :requires,
-      :ruby_version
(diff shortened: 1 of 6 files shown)


colby-swandale:
Nice! @bundlerbot r+

indirect:
I'm late on this one, but this is a great catch 💯

PR #5901, opened 2017-07-26
Add tasks to make running in 2.0 mode easier

segiddins:

The problem was it was too difficult to run Bundler in 2.0 mode.

My diagnosis was we needed a way to install a dev version of 2.0, and also to run commands as bundler 2.

My fix adds a `bin/bundle2` bin stub that runs bundler as 2.0.0.dev, and a `bundle_2:install` rake task that will install `bundler-2.0.0.dev`

### DIFF
--- a/bin/bundle2
+++ b/bin/bundle2
@@ -1,0 +1,8 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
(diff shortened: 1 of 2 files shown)


indirect:
Awesome. @bundlerbot r+

PR #5902, opened 2017-07-27
Update to Molinillo 0.6.0

segiddins:

The problem was Molinillo 0.5.x was so old :P

My diagnosis was we needed to _fix every single known bug_, while also improving performance.

My fix is to behold the awesomeness of https://github.com/CocoaPods/Molinillo/releases/0.6.0, and introduce it to Bundler.


andremedeiros:
Always love when "speed up" is part of the changelog! :+1:

indirect:
woo :tada:

greysteil:
So psyched for this!

segiddins:
@bundlerbot r+

PR #5903, opened 2017-07-27
Delete redundant aliases from bundle outdated command 💫

arbonap:

The problem was...
- I was looking at the `bundle outdated` command in `cli.rb`, and noticed some flag aliases were the same as the original flags.
My diagnosis was...
- I thought these aliases weren't necessary

My fix...
- I deleted two aliases from `bundle outdated`.
- I ran `rake spec:deps && bin/rspec spec/commands/outdated_spec.rb` locally and all tests passed.

I chose this fix because...
- to DRY up `cli.rb` 👌🏼


segiddins:
@bundlerbot r+

PR #5904, opened 2017-07-27
[Update] Warn when an explicitly updated spec does not get a newer version

segiddins:

The problem was you could run `bundle update foo bar` and it would be possible that either `foo` or `bar` not move to a newer version.

My diagnosis was we should tell users when this happens!

My fix warns when a locked bundle has `bundle update` called with an explicit list of gems and one of those gems could not actually be _updated_.

I chose this fix because it didn't require mucking around with the resolved.

### DIFF
--- a/lib/bundler/cli/update.rb
+++ b/lib/bundler/cli/update.rb
@@ -67,6 +67,21 @@ module Bundler
         Bundler::CLI::Clean.new(options).run
       end
 
+      if locked_gems = Bundler.definition.locked_gems
+        gems.each do |name|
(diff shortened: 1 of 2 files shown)


segiddins:
@indirect @colby-swandale any interest in this feature?

colby-swandale:
This is good! I ran into an issue which this solves the other day,  i `bundle update <gem>` and didn't realise the gem was never updated afterwards.

indirect:
Yeah, that is excellent 👍🏻

@bundlerbot r+

PR #5906, opened 2017-07-28
Add missing flag aliases to docs 🌟

arbonap:

The problem was...
- I noticed in the documentation that a lot of aliases for flags were missing.

My diagnosis was...
- I added missing aliases for flags into the documentation. 

My fix...
- I added the missing flag aliases into the docs, so that users can more easily access this information. 💖
- I also added in a line describing how RSpec is the default testing framework 

I chose this fix because...
- more bundler knowledge for bundler users is helpful 🍒


colby-swandale:
Thank you! 🎉 @bundlerbot r+

PR #5907, opened 2017-07-31
Add new line after newgem gemspec template

roberts1000:

When using `bundle gem` to create a new gem, the created gemspec file starts with 

```
# coding: utf-8
lib = File.expand_path('../lib', __FILE__)
```

Rubocop recommends having a blank line after magic comments (like `# coding: utf-8`).

The rubocop style guide recommends having a blank line after "magic comments" ([see link](https://github.com/bbatsov/ruby-style-guide#separate-magic-comments-from-code)).   The main bundler.gemspec follows this recommendation, but the .gemspec, that is created for new gems, does not.

Update the template file to have a blank line after the magic comment.

It's what rubocop wants.


colby-swandale:
Thanks! @bundlerbot r+

PR #5908, opened 2017-08-01
[2.0] Globally cache built extensions for RG & Git sources

### LINKED ISSUE #5905: [2.0] Cache compiled gem extensions across bundles
Now that every Bundle will default to install into `./.bundle`, recompiling every gem for every separate app is going to be really tedious. It would be awesome if every app could grab a copy of an already-compiled extension and use that instead of having to compile again from scratch.

segiddins:

The problem was with the switch to defaulting the install path to `./.bundle`, you'd need to rebuild extensions for each app you use a gem in.

Closes #5905 

My diagnosis was we could globally cache built extensions in a per-ruby, per-source directory, similar to how 2.0 global `.gem` caching and the git cache works.

My fix implements that cache.

### Remaining questions:

- [x] Tests
- [x] Right now, the directory structure is `cache/gems/remote_name/extensions/ruby_scope/gem_full_name`. I'm wondering whether we should make it `cache/extensions/ruby_scope/source_slug/gem_full_name` instead to make removing the cache for an old ruby version easier?
- [x] Should the cache path include something about the local OS? ie. do we want to force a re-compile when the user updates OS versions?

### DIFF
--- a/spec/install/global_cache_spec.rb
+++ b/spec/install/global_cache_spec.rb
@@ -186,4 +186,50 @@ RSpec.describe "global gem caching" do
       end
     end
   end
+
+  describe "extension caching", :rubygems => "2.2" do
+    it "works" do
+      build_git "very_simple_git_binary", &:add_c_extension
+      build_lib "very_simple_path_binary", &:add_c_extension
+      revision = revision_for(lib_path("very_simple_git_binary-1.0"))[0, 12]
+
+      install_gemfile! <<-G
+        source "file:#{gem_repo1}"
+
+        gem "very_simple_binary"
+        gem "very_simple_git_binary", :git => "#{lib_path("very_simple_git_binary-1.0")}"
+        gem "very_simple_path_binary", :path => "#{lib_path("very_simple_path_binary-1.0")}"
+      G
+
+      gem_binary_cache = home(".bundle", "cache", "extensions", specific_local_platform.to_s, Bundler.ruby_scope,
+        Digest(:MD5).hexdigest("#{gem_repo1}/"), "very_simple_binary-1.0")
+      git_binary_cache = home(".bundle", "cache", "extensions", specific_local_platform.to_s, Bundler.ruby_scope,
+        "very_simple_git_binary-1.0-#{revision}", "very_simple_git_binary-1.0")
+
+      cached_extensions = Pathname.glob(home(".bundle", "cache", "extensions", "*", "*", "*", "*", "*")).sort
+      expect(cached_extensions).to eq [gem_binary_cache, git_binary_cache].sort
+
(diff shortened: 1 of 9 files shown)


segiddins:
With this change, installing a clean rails project (with extensions cached) goes from 123.3 seconds to 10.8 seconds on my machine. Not bad.

indirect:
😱🎉👏🏻

jules2689:
> do we want to force a re-compile when the user updates OS versions?

yes please. At Shopify we experience significant pain when a new OS comes out as we have lots of issues with OS version-specific compiled versions.

In bootsnap, we use a [number of things](https://github.com/Shopify/bootsnap/blob/e78fa8c0c14ab9a409e59c6ac2098280b8f1fe76/ext/bootsnap/bootsnap.c#L48) for a "cache key". Maybe we can do similar

hmistry:
Agree with @jules2689 on re-compile for OS updates.

Regarding scoping to ruby versions - that's easy I think that should be the case. However I'm unclear about `source_slug` and how it would make it easier to remove old ruby versions? (Because the ruby scope would make that easier right?)

I can see `source_slug` be valuable in separating different compiled options by different sources of the same gem but is that a real issue/need?

Liking the performance gains - great work!

jules2689:
I like the `cache/extensions/ruby_scope/source_slug/gem_full_name` aspect. I find a need to remove cache for specific ruby versions. I feel like we would need to give better error messages though to explain "hey this seems to be broken in 2.4.1, maybe you should delete"

spec/install/global_cache_spec.rb
```diff
@@ -186,4 +186,50 @@ def source2_global_cache(*segments)
...
+        source "file:#{gem_repo1}"
+
+        gem "very_simple_binary"
+        gem "very_simple_git_binary", :git => "#{lib_path("very_simple_git_binary-1.0")}"
+        gem "very_simple_path_binary", :path => "#{lib_path("very_simple_path_binary-1.0")}"
+      G
+
+      gem_binary_cache = home(".bundle", "cache", "extensions", specific_local_platform.to_s, Bundler.ruby_scope,
+        Digest(:MD5).hexdigest("#{gem_repo1}/"), "very_simple_binary-1.0")
+      git_binary_cache = home(".bundle", "cache", "extensions", specific_local_platform.to_s, Bundler.ruby_scope,
```
segiddins: @jules2689 this directory structure look good?

segiddins:
Questions all addressed. @indirect r?

indirect:
This looks awesome, @segiddins! r+

segiddins:
rebased, and so excited for this to land!
@bundlerbot r=indirect

PR #5910, opened 2017-08-01
[CLI] Suggest bundle update --bundler when a newer version is installed

segiddins:

The problem is once users start using our new `bin/bundle` binstub or the RubyGems that helps lock to the bundler version in the lockfile, the outdated bundler version warning was outdated.

My diagnosis was we needed to tell people to `bundle update --bundler`, and restrict the `gem install` suggestion to when the latest version indeed wasn't installed.

My fix takes into account the actual installed bundler versions.

@indirect I think this addresses your concern on the RG PR?

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -677,8 +677,17 @@ module Bundler
 
       current = Gem::Version.new(VERSION)
       return if current >= latest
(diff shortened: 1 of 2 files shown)


indirect:
@segiddins yep, I think this is right. 👍🏻
@bundlerbot r+

PR #5911, opened 2017-08-02
Explain RUBYOPT in Debugging section of Setup docs

arbonap:

The problem was...
- I was looking through `SETUP.md` and I didn't understand why I had to set the `RUBYOPT` env variable to `-rpry dbundle`. I was wondering why I couldn't just normally use `require 'pry'; binding.pry` to debug Bundler's codebase.

My diagnosis was...
- To add more explanations to the Setup documentation.

My fix...
- I asked @indirect and @jules2689 all about `RUBYOPT`, and why we have to set `RUBYOPT` to require pry and require this path to Bundler. I used their wonderful explanations and tried synthesizing their knowledge in a way that made sense to those who aren't familiar with `RUBYOPT`. I also included background context as to why Bundler chooses to do this.

I chose this fix because...
- Having background context of how and why Bundler combines `RUBYOPT` with `dbundle` in the Setup docs will help new contributors ramp-up.
- I also think having this knowledge will (hopefully) help new contributors not put in `require 'pry'; binding.pry` like they may be used to in other Ruby projects. ✨

Let me know what y'all think of this first draft 👌🏼. I'm happy to incorporate changes.

### DIFF
--- a/doc/development/SETUP.md
+++ b/doc/development/SETUP.md
@@ -27,3 +27,14 @@ Bundler doesn't use a Gemfile to list development dependencies, because when we 
 ## Debugging with `pry`
 
 To dive into the code with Pry: `RUBYOPT=-rpry dbundle` to require pry and then run commands.
+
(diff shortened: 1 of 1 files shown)


segiddins:
Excellent!
@bundlerbot r+

PR #5912, opened 2017-08-02
[2.0] Don't create gem home before setting it

koic:

(Follow up of #2884)

It seems that [the PR](#2884) written in [Bundler 2 RFC](https://github.com/bundler/rfcs/pull/6/files#diff-0579edc0374571613cd01f0a42c19bdaR30) is not included in master.

It applies to master branch based on the PR implementation written in RFC.

This PR will switch `Bundler.feature_flag` using `bundler_2_mode`.

Since #2884 does not use `Bundler.feature_flag`, then compatibility with Bundler 1 will be lost.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -524,13 +524,6 @@ EOF
     end
 
     def configure_gem_home
-      # TODO: This mkdir_p is only needed for JRuby <= 1.5 and should go away (GH #602)
-      begin
-        FileUtils.mkdir_p bundle_path.to_s
-      rescue
-        nil
-      end
-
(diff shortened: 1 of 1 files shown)


segiddins:
I don't know if this is strictly necessary, and we can just remove the mkdir call down the road?

koic:
I see. It is the issue that was happening when using old JRuby, and there is a possibility that it is not a problem now. I updated this PR to just remove `FileUtils.mkdir`.

segiddins:
@bundlerbot r+

koic:
Thanks!

PR #5913, opened 2017-08-02
Fix default gem activation test when using Ruby 2.5

koic:

ruby-head (Ruby 2.5.0dev) has failed at Travis CI.

https://travis-ci.org/bundler/bundler/jobs/259991764

Standard libraries are cut out to Gem in Ruby 2.5. `etc` and `fileutils` are also among them.

- Gemify etc ... https://bugs.ruby-lang.org/issues/13256
- Gemify fileutils ... https://bugs.ruby-lang.org/issues/13197

This PR passes tests by including them in exemptions just like `openssl`.


segiddins:
That failure is intentional and will be fixed once a new version of Thor is released

koic:
Thanks for your information. I got it. I close this PR.

PR #5916, opened 2017-08-03
Update broken link in bundle-gem man page

arbonap:

The problem was...
- This resolves [Issue#296](https://github.com/bundler/bundler-site/issues/296) on bundler/bundler-site. 

My diagnosis was...
- @olleolleolle came across a broken link in the `bundle gem` docs. Specifically, the `bundle-config` link in the "See Also" section was broken.

My fix...
- I fixed the broken link 🍒

I chose this fix because...
- To make navigating around the bundler docs a pleasant experience.

### DIFF
--- a/man/bundle-gem.ronn
+++ b/man/bundle-gem.ronn
@@ -75,4 +75,4 @@ configuration file using the following names:
 
 ## SEE ALSO
 
-* bundle-config(1)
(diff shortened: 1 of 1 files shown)


colby-swandale:
@bundlerbot r+

PR #5841, opened 2017-07-06
Capitalize the description lines in Bundler 2

koic:
This PR is a small patch.

### What was the end-user problem that led to this PR?

I made the following changes locally to the code of the master branch and tried Bundler 2.0.

```diff
% git diff
diff --git a/lib/bundler/version.rb b/lib/bundler/version.rb
index dad4a88a2..386782023 100644
--- a/lib/bundler/version.rb
+++ b/lib/bundler/version.rb
@@ -7,7 +7,7 @@ module Bundler
   # We're doing this because we might write tests that deal
   # with other versions of bundler and we are unsure how to
   # handle this better.
-  VERSION = "1.15.1" unless defined?(::Bundler::VERSION)
+  VERSION = "2.0.0.dev" unless defined?(::Bundler::VERSION)

   def self.overwrite_loaded_gem_version
     begin
```

Here is the result of running the `bundle` command.

```console
% bundle
Bundler version 2.0.0.dev

Commands:
  bundle add GEM VERSION            # Add gem to Gemfile and run bundle install
  bundle binstubs GEM [OPTIONS]     # Install the binstubs of the listed gem
  bundle cache [OPTIONS]            # Locks and then caches all of the gems into vendor/cache
  bundle check [OPTIONS]            # Checks if the dependencies listed in Gemfile are satisfied by currently installed gems
  bundle clean [OPTIONS]            # Cleans up unused gems in your bundler directory
  bundle config NAME [VALUE]        # retrieve or set a configuration value
  bundle doctor [OPTIONS]           # Checks the bundle for common problems
  bundle env                        # Print information about the environment Bundler is running under
  bundle exec [OPTIONS]             # Run the command in context of the bundle
  bundle gem GEM [OPTIONS]          # Creates a skeleton for creating a rubygem
  bundle help [COMMAND]             # Describe available commands or one specific command
  bundle help [COMMAND]             # Describe subcommands or one specific subcommand
  bundle info GEM [OPTIONS]         # Show information for the given gem
  bundle init [OPTIONS]             # Generates a Gemfile into the current working directory
  bundle inject GEM VERSION         # Add the named gem, with version requirements, to the resolved Gemfile
  bundle install PLUGINS            # Install the plugin from the source
  bundle install [OPTIONS]          # Install the current environment to the system
  bundle issue                      # Learn how to report an issue in Bundler
  bundle licenses                   # Prints the license of all gems in the bundle
  bundle lock                       # Creates a lockfile without installing
[...]

PR #5920, opened 2017-08-04
implement new list command

colby-swandale:

We're removing the `bundle show` command (which has been replaced with `bundle info`) but we should probably still have a command to list the gems in the bundle.

`bundle show` has been marked as deprecated but a replacement for listing gems does not yet exist

Add a new `bundle list` (including `ls` alias) that prints the gems in the bundle, this was taken from #4754

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -261,7 +261,18 @@ module Bundler
       Show.new(options, gem_name).run
     end
     # TODO: 2.0 remove `bundle show`
-    map %w[list] => "show"
+
+    if Bundler.feature_flag.list_command?
+      desc "list", "list all gems in the bundle"
+      def list
+        require "bundler/cli/list"
+        List.new.run
+      end
+
+      map %w[ls] => "list"
+    else
+      map %w[list] => "show"
+    end
 
     desc "info GEM [OPTIONS]", "Show information for the given gem"
     method_option "path", :type => :boolean, :banner => "Print full path to gem"
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -263,10 +263,11 @@ module Bundler
     # TODO: 2.0 remove `bundle show`
 
     if Bundler.feature_flag.list_command?
-      desc "list", "list all gems in the bundle"
+      desc "list", "List all gems in the bundle"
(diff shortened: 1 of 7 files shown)


lib/bundler/cli/list.rb
```diff
@@ -0,0 +1,17 @@
...
+
+module Bundler
+  class CLI::List
+    def run
+      specs = Bundler.load.specs.reject {|s| s.name == "bundler" }
+
+      return Bundler.ui.info "No gems in the Gemfile" if specs.empty?
+      Bundler.ui.info "Gems included by the bundle:"
+      specs.sort_by(&:name).each do |s|
+        Bundler.ui.info "  * #{s.name} (#{s.version}#{s.git_version})"
```
segiddins: Since this is a new command, my temptation would be just to print out the gems, one per line, without anything else?

colby-swandale: so just print out the gem name only?

colby-swandale: @indirect do you have any opinion about this?

segiddins: name the gem name only, and add the version and git version when verbose?

indirect: gem name only with version on verbose was my first thought, but then I checked npm and yarn, and they both always include the version... I guess is there ever a time that it would be useful to have the names without also having the versions? scripting? should we provide `bundle ls --names-only` for that?

colby-swandale: I'm happy to have a `--names-only` option, @segiddins?

colby-swandale:
@segiddins r?

lib/bundler/cli.rb
```diff
@@ -261,7 +261,19 @@ def show(gem_name = nil)
       Show.new(options, gem_name).run
     end
     # TODO: 2.0 remove `bundle show`
-    map %w[list] => "show"
+
+    if Bundler.feature_flag.list_command?
+      desc "list", "list all gems in the bundle"
```
koic: I think that it is good to capitalize the beginning of the description line.

```diff
- desc "list", "list all gems in the bundle"
+ desc "list", "List all gems in the bundle"
```

Related PR #5841.

segiddins:
Awesome!
@bundlerbot r+

PR #5922, opened 2017-08-04
fix documented init gems rb config env var

colby-swandale:

I documented the env var for the init new gem file feature flag incorrectly

    › BUNDLE_NEW_GEMFILE_NAME=1 dbundler init
    Writing new Gemfile to /Users/c/Desktop/Projects/testapp/Gemfile

does not generate the gems.rb as expected

Use the correct env name to generate gems.rb


indirect:
👍 @bundlerbot r+

Issue #4846, opened 2016-08-05
download_gem not being retried


jkeiser:
It appears that retry applies to fetching specs and talking to git, but not to actually retrieving the gems themselves. We have received this error multiple times in the past month, even with `bundle install ... --retry 5`:

`Gem::RemoteFetcher::FetchError: Errno::ECONNRESET: Connection reset by peer - SSL_connect `

Bundler 1.12.5, Rubygems 2.6.5, Ruby 2.1.8p440.

For one example:

```
Fetching gem metadata from https://rubygems.org/
Fetching version metadata from https://rubygems.org/
Fetching dependency metadata from https://rubygems.org/
Fetching git://github.com/chef/omnibus.git
Fetching git://github.com/chef/omnibus-software.git
Fetching https://github.com/ksubrama/pedump.git
Installing awesome_print 1.7.0
Installing jmespath 1.3.1
Installing fuzzyurl 0.8.0
Installing mixlib-config 2.2.1
Installing mixlib-shellout 2.2.6
Installing chef-sugar 3.4.0
Installing cleanroom 1.0.0
Installing ffi 1.9.14 with native extensions
Installing libyajl2 1.2.0 with native extensions
Installing iostruct 0.0.4
Installing ipaddress 0.8.3
Installing mixlib-cli 1.7.0
Installing mixlib-log 1.6.0
Installing mixlib-versioning 1.1.0
Installing multipart-post 1.2.0
Installing plist 3.2.0
Installing systemu 2.6.5
Installing wmi-lite 1.0.0

Gem::RemoteFetcher::FetchError: Errno::ECONNRESET: Connection reset by peer - SSL_connect (https://rubygems.org/gems/ruby-progressbar-1.8.1.gem)
Installing thor 0.19.1
Installing progressbar 0.21.0
Installing zhexdump 0.0.2
Using bundler 1.12.5
Installing aws-sdk-core 2.5.0
Installing chef-config 12.12.15
Installing ffi-yajl 2.3.0 with native extensions
An error occurred while installing ruby-progressbar (1.8.1), and Bundler cannot
continue.
Make sure that `gem install ruby-progressbar -v '1.8.1'` succeeds before
bundling.
```

[...]

PR #5928, opened 2017-08-06
Retry downloading gems consistently across all versions of RubyGems

shayonj:

`Gem::RemoteFetcher::FetchError` failures were not being retried when downloading gems. Retrying downloading of gems for network blips is very ideal and helpful in CI environment.

- The retry logic that existed (using `Bundler::Retry`) for `Bundler::RubygemsIntegration#download_gem` was not getting called, becase `Bundler.rubygems` defaulted to a sub class of `Bundler::RubygemsIntegration::MoreFuture`, which inherited from `Bundler::RubygemsIntegration::Future` and `#download_gem` in there didn't perform download with retry logic.
- Sending `Gem::RemoteFetcher::FetchError` as an argument to `Bundler::Retry` meant, `Bundler::Retry` would fail first thing,[ if exception being raised was the one being supplied](https://github.com/bundler/bundler/blob/5a61b65ad5ec1df1539ecf8bc5d124f2b254ba14/lib/bundler/retry.rb#L46..L49). The intent is to retry when downloading fails.

**Example of how sub classing issue looked like:**

```ruby
class RubygemsIntegration
  def download_gem
    "I retry when downloading"
  end
end

class Future < RubygemsIntegration
  def download_gem
    "I don't retry when downloading"
  end
end

class MoreFuture < Future
  # no download_gem
end

mf = MoreFuture.new #=> Bundler.rubygems where RubyGem version is > 2.1.0
mf.download_gem
 => "I don't retry when downloading" 
```

**Example of Bundler::Retry**

```ruby
Bundler::Retry.new("download gem from", StandardError).attempts do
  puts "I will only print once, whereas I should print 4 times"
  raise StandardError.new("FooBar")
end
```

I added the retry logic for to `Bundler::RubygemsIntegration::Future#download_gem` and removed `Gem::RemoteFetcher::FetchError` as an argument for failed exceptions.

I chose this fix because this fix fulfills the expected behavior(https://github.com/bundler/bundler/issues/4846). Also added specs that makes sure retry logic is intact, which otherwise would have failed without the current logic in place.

### DIFF
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -317,7 +317,7 @@ module Bundler
     def download_gem(spec, uri, path)
       uri = Bundler.settings.mirror_for(uri)
       fetcher = Gem::RemoteFetcher.new(configuration[:http_proxy])
-      Bundler::Retry.new("download gem #{uri}", Gem::RemoteFetcher::FetchError).attempts do
+      Bundler::Retry.new("download gem from #{uri}").attempts do
         fetcher.download(spec, uri, path)
       end
     end
@@ -752,7 +752,9 @@ module Bundler
         uri = Bundler.settings.mirror_for(uri)
         fetcher = gem_remote_fetcher
         fetcher.headers = { "X-Gemfile-Source" => spec.remote.original_uri.to_s } if spec.remote.original_uri
-        fetcher.download(spec, uri, path)
+        Bundler::Retry.new("download gem from #{uri}").attempts do
+          fetcher.download(spec, uri, path)
+        end
       end
 
       def gem_remote_fetcher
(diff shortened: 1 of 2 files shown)


shayonj:
cc @segiddins || @indirect (as you have context on the original issue) 🙂

segiddins:
Awesome, thanks!
@bundlerbot r+

shayonj:
🙌 @segiddins 
Do you think there is a possibility of releasing a PATH/MINOR version involving this change? 🙂

segiddins:
Why not! I can do this in a 1.15.4 tomorrow since I'm sure it'll cherry pick cleanly

shayonj:
❤️

PR #5933, opened 2017-08-07
Use Bundler::Installer for bundle pristine

NickLaMuro:
Fixes bundler pristine to respect the `bundle config` options for particular gems.  In my particular case, on OSX I have a keg version of `pg` and a bundle config with the following entry:

```console
$ bundle config
...

build.pg
Set for the current user (/Users/nicklamuro/.bundle/config): "--with-pg-config=/usr/local/Cellar/postgresql@9.5/9.5.6/bin/pg_config"

```

This would not be respected

See above, but steps to reproduce (requires `bundler` >= 1.15):

1. Install postgres on your machine if it is not already there (OS dependent)
  
2. Add the following bundler build arg for `pg`:
  
  ```console
  $ bundle config build.pg -- --with-pg-config=$(which pg_config)
  ```
  
2. Make sure `pg_config` is not in your `$PATH`:
  
  ```console
  $ export PATH=$(echo "$PATH" | sed "s|:$(dirname $(which pg_conf)))||")
  ```
  
* Create a simple project with `pg` as a single gem in your Gemfile:
  
  ```
  # Gemfile
  gem "pg"
  ```
  
* Bundle:
  
  ```console
  $ bundle install
  ```
  
* Attempt to `bundle pristine` (you should get an error):

  ```console
  $ bundle pristine
  ```

The newly added `bundle pristine` did no use the same code that was used by `bundle install` to execute the re-installation of the gem.

By making use of the `Bundler::Installer` and `Bundler::Installer::GemInstaller` code that is already used with `bundle install`, we can reuse the code that injects the `bundle config` options into each gem that is being installed.

Didn't want to repeat code that was already being used elsewhere.  Caused a few lines of code to be added that weren't there previously, but nothing obscene.

### DIFF
--- a/lib/bundler/cli/pristine.rb
+++ b/lib/bundler/cli/pristine.rb
@@ -9,6 +9,9 @@ module Bundler
 
     def run
       CLI::Common.ensure_all_gems_in_lockfile!(@gems)
+      definition = Bundler.definition
+      definition.validate_runtime!
+      installer = Bundler::Installer.new(Bundler.root, definition)
 
       Bundler.load.specs.each do |spec|
         next if spec.name == "bundler" # Source::Rubygems doesn't install bundler
@@ -26,14 +29,15 @@ module Bundler
           end
 
           FileUtils.rm_rf spec.full_gem_path
-          source.install(spec, :force => true)
         when Source::Git
           source.remote!
           FileUtils.rm_rf spec.full_gem_path
-          source.install(spec, :force => true)
         else
           Bundler.ui.warn("Cannot pristine #{gem_name}. Gem is sourced from local path.")
+          next
         end
+
+        Bundler::GemInstaller.new(spec, installer, false, 0, true).install_from_spec
       end
     end
   end
--- a/spec/commands/pristine_spec.rb
+++ b/spec/commands/pristine_spec.rb
@@ -11,6 +11,7 @@ RSpec.describe "bundle pristine" do
     build_repo2 do
       build_gem "weakling"
       build_gem "baz-dev", "1.0.0"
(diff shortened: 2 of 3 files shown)


segiddins:
Looks good with a test!

NickLaMuro:
@segiddins will need to look into how I could do that 😁 

Will get back to you in a bit with a test (hopefully...)

NickLaMuro:
Test added, though I admit it is probably pretty fragile...

Didn't want to try adding a real c-ext gem to really test this since I figure that would cause platform dependent nightmares, so it is very much dependent on the `Gem::Installer#install` calling `write_build_info_file` when called with build args (regardless if anything is done with them or not).  Hopefully that doesn't change 😅

NickLaMuro:
Will look into the 1.9.3 failures in a bit.  Obviously a change in rubygems that I didn't account for in the tests, and wasn't testing for locally.  Should be fixable, will just need to do some sleuthing.

segiddins:
Now failing on 2.4.8 and 2.5.2, I recommend doing `expect(...).to match(//).or match(//)`

NickLaMuro:
@segiddins Sorry, missed this on Friday, but just loosened the the regexp a bit to make this work.  

That said, still seems a little brittle, so not sure if you were prefer me making these more "unit test like" (effectively call `Bundler::CLI::Pristine#run` directly in the test, and confirm a `expect(some_class).receive` gets a `:build_args` argument properly), or if we are cool with this for now.

segiddins:
Looks perfect!

segiddins:
@bundlerbot r+

indirect:
This is a great catch, thanks for the patch!

PR #5938, opened 2017-08-09
Use universal rubygems method for installing gems

NickLaMuro:

When following the [setup documentation](https://github.com/bundler/bundler/blob/master/doc/development/SETUP.md) for bundler, it suggests to run `bin/rake spec:deps` to install the development dependencies for bundler.

Doing this with ruby 1.9.3 (and not upgrading rubygems) will cause this to fail.

Ruby 1.9.3 ships with a version of rubygems that is less than 2.0, and doesn't include the mechanisms for installing multiple gems (with specific versions) using colon delimited strings (which is the mechanism used to do so in `spec:deps`).

By using `ruby -S gem install [GEM_NAME] -v [GEM_VERSION]` for each gem, it is a universal way of installing targeted versions of gems on various versions of ruby/rubygems that isn't rubygems dependent.

This was the easiest implementation to do that maintains the same documentaion and instructions necessary to setup bundler for development.

The cons with this approach is that is causes additional subshell invocations to be launched for each gem, instead of one, which will be slower.  Since this is a finite number of calls we are doing this for (5 currently), I expect that to be negligible compared to the total time necessary load up the rubygems `gem` command and fetch each gem from rubygems (we can compare the travis builds from master to see this to compare).

An alternative to this approach would be to call out to the `Gem::Installer` directly from within the rake task, and avoid calling out to the `gem` command all together.  This has the downside of requiring more code to do it, but should alleviate the any extra time caused by using `sh` call to install the gems (might try this in a separate commit so it will have another build on travis and we can compare that result to the initial implementation).

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -47,10 +47,17 @@ namespace :spec do
       deps.delete("rdiscount")
     end
 
-    gem_install_command = "install --no-ri --no-rdoc --conservative " + deps.sort_by {|name, _| name }.map do |name, version|
-      "'#{name}:#{version}'"
-    end.join(" ")
-    sh %(#{Gem.ruby} -S gem #{gem_install_command})
+    if Gem::VERSION < "2.0.0"
+      deps.sort_by {|name, _| name }.map do |name, version|
+        gem_install_command = "install --no-ri --no-rdoc --conservative #{name} -v '#{version}'"
+        sh %(#{Gem.ruby} -S gem #{gem_install_command})
+      end
+    else
+      gem_install_command = "install --no-ri --no-rdoc --conservative " + deps.sort_by {|name, _| name }.map do |name, version|
+        "'#{name}:#{version}'"
+      end.join(" ")
+      sh %(#{Gem.ruby} -S gem #{gem_install_command})
+    end
 
     # Download and install gems used inside tests
     $LOAD_PATH.unshift("./spec")


colby-swandale:
This should only be done if the version of RubyGems does not support installing a list of gems.

NickLaMuro:
@colby-swandale While I agree, part of what I said in the PR description was:

> we can compare the travis builds from master to see this to compare

If we filter on rubygems versions from the get go, we can't see any affects on travis because part of what the [travis version](https://github.com/bundler/bundler/blob/master/Rakefile#L74) of this does is install rubygems greater than 2.0, and then invokes `spec:deps` once that is loaded.  It also means that this code would never be exercised on travis, and we would be going on "faith" that it works.

I have edited the PR title to `[WIP]` (work in progress) to reflect that this is not ready for merging yet until we can get more data, but I think making this a fallback from the get go would de-legitimize it from having any data in the first place that this is a significant performance degradation to the build (which I assume is what your argument for having this as a fall back is in the first place).

NickLaMuro:
Using [this build's travis run](https://travis-ci.org/bundler/bundler/builds/262472367), and [one from master](https://travis-ci.org/bundler/bundler/builds/261819739), I have yet to find a case where the `rake spec:travis:deps` differs by any more than a second to complete, and with it taking around 20s on average to complete across multiple ruby versions, I think that is within an acceptable margin of variance.

Here are some excerpts:

|       | master | #5938 
| --- | --- | --- |
| ruby-2.4.1 RGV=master | [22.78s](https://travis-ci.org/bundler/bundler/jobs/261819740) | [21.99s](https://travis-ci.org/bundler/bundler/jobs/262472368) |
| ruby-1.9.3 RGV=v2.2.5 | [23.79s](https://travis-ci.org/bundler/bundler/jobs/261819740) | [22.89s](https://travis-ci.org/bundler/bundler/jobs/262472389) |
| ruby-1.8.7 RGV=v2.2.5 | [20.54s](https://travis-ci.org/bundler/bundler/jobs/261819777) | [20.98s](https://travis-ci.org/bundler/bundler/jobs/262472396) |

I mostly just picked a couple of different versions of ruby for this, and I think it just so happens that most of the "faster" ones are the ones from my PRs build.  I think this would vary between each one, so I don't think there is any noticeable performance difference (from the perspective of time for the build to run) to be gained or lost with this change.

segiddins:
Once the deps are installed though, it goes from 1.5s to 9.9s on my machine

NickLaMuro:
> Once the deps are installed though, it goes from 1.5s to 9.9s on my machine

@segiddins I assume you are referring to rerunning the `spec:deps` after the gems have installed correctly, right?  I know that this isn't a case we won't run into on Travis (unless we consider using the Travis cache for the gem dependencies to slightly speed up the build), but is rerunning `spec:deps` something that is common for you or anyone else?  I personal don't consider myself a huge and frequent contributor to `bundler` and every time I do come back to add something, I just simply "hope everything works right from last time :crossed_fingers:", so I probably don't have a good perspective on this.

That said, this PR was more focused from the beginning on "new bundler developer approachability", and not on making this a super-ly optimal performant task.  If we instead think that to solve making this more approachable would rather be to do one of the following:

* Updating the docs is to add more steps is a better way to approach this problem
* Don't mind increasing complexity in this task to solve that problem
* We will be switching to not supporting older versions rubygems versions in bundler 2 _**AND**_ never expecting that someone would ever have to write and test backported change to older versions of `bundler`
* Developers can figure it out themselves (technically a solution, though one I wouldn't recommend)

That is fine.  Again, I personally am not concerned about the performance for a developer setup task, and am more focused on making these setup steps as accessible as possible.  Any performance metrics here are simply to help put into perspective the pros and cons of the selection of solutions that we are considering.

segiddins:
I'm happy just to do "if Gem::VERSION < 2.5" or something

NickLaMuro:
Was packed up for the weekend when I got this, but changes made.

segiddins:
Wonderful, thanks so much!

segiddins:
@bundlerbot r+

PR #5940, opened 2017-08-09
Added fileutils to exemptions gems. 

hsbt:
It fixes broken tests caused by default gems with Ruby 2.5(also known as ruby-head on travis).

It is confirmed that fileutils will promote default gems on Ruby 2.5.0.

### DIFF
--- a/spec/runtime/setup_spec.rb
+++ b/spec/runtime/setup_spec.rb
@@ -1231,7 +1231,7 @@ end
         if Gem::Version.new(Gem::VERSION) >= Gem::Version.new("2.7") || ENV["RGV"] == "master"
           []
         else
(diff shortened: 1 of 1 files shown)


segiddins:
We don't want to merge this yet, it's intended to fail so we make sure to fix the bug rather than ignoring it. We're just waiting on a Thor release

olleolleolle:
#5849 brought Thor v0.20.0. Was this the release we were waiting for?

segiddins:
Yup!

segiddins:
This should no longer be needed as of Bundler 1.15.4

hsbt:
I understood. Thanks.

PR #5955, opened 2017-08-19
Default to packaging all on 2.0

segiddins:
This is going to be the last set of changes to defaults in Bundler 2.

@colby-swandale @indirect are there other changes I should be making here?

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -348,7 +348,8 @@ module Bundler
       map %w[cache] => :package
     else
       desc "cache [OPTIONS]", "Cache all the gems to vendor/cache", :hide => true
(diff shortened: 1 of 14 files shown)


indirect:
@segiddins IMO this option shouldn't even exist in 2.0—it should always package --all, and you shouldn't be able to turn it off. 👍

indirect:
(Code looks good, I am r+)

segiddins:
Let me remove the option in 2.0 then!

segiddins:
> IMO this option shouldn't even exist in 2.0—it should always package --all, and you shouldn't be able to turn it off. 👍

Done!

segiddins:
@bundlerbot r=indirect

PR #5956, opened 2017-08-19
[Update] Avoid a normal update when running with only --bundler

segiddins:

The problem was `bundle update --bundler` went through the full update/resolution path when really we just want to pin to a different bundler version.

My diagnosis was we should just update the lockfile, without threading the rest of the process like an update.

This implements what @indirect requested in https://github.com/rubygems/rubygems/pull/1977#issuecomment-322852799.


segiddins:
@indirect please review?

indirect:
Yep, looks great. @bundlerbot r+

PR #5957, opened 2017-08-19
When the locked bundler is under 2.0, allow any version < 2

segiddins:

The problem was the bin stub would lock to a specific version even for a 1.x version, which was an incompatible change.

See @indirect's comment in https://github.com/rubygems/rubygems/pull/1977#issuecomment-322852799

My diagnosis was we needed to only pin to a specific version for 2.0+.

My fix transforms any 1.x bundler version to a "< 2" constraint.


indirect:
Perfect. @bundlerbot r+

Issue #5875, opened 2017-07-19
Auto-clean after 2.0 installs when path is not set


indirect:
After install operations when `path` is _not_ set (eg, bundler is installing into `./.bundle`, the default location), let's run the `clean` code as well, to remove unused dependencies. There is precedent for this in other package managers, and I think it's the sweet spot for keeping per-application bundles from growing out of control over time.

Installation into a specific path directory should not automatically trigger cleaning—cleaning shared directories should continue to be done via the `clean` command.

PR #5961, opened 2017-08-20
[2.0] Automatically bundle clean without a path set

segiddins:

The problem was with the switch to defaulting to a local install in 2.0, we're concerned about taking up too much disk space.

My diagnosis was we'd like to run `bundle clean` when possible.

Closes https://github.com/bundler/bundler/issues/5875.

My fix runs `clean` automatically on 2.0 when a `path` has _not_ been set.

I chose this fix because it avoids the possibility of accidentally cleaning a shared location.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -5,6 +5,8 @@ require "bundler/vendored_thor"
 
 module Bundler
   class CLI < Thor
+    require "bundler/cli/common"
+
     AUTO_INSTALL_CMDS = %w[show binstubs outdated exec open console licenses clean].freeze
     PARSEABLE_COMMANDS = %w[
       check config help exec platform show version
--- a/lib/bundler/cli/add.rb
+++ b/lib/bundler/cli/add.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require "bundler/cli/common"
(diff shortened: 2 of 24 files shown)


lib/bundler/version.rb
```diff
@@ -7,7 +7,7 @@ module Bundler
   # We're doing this because we might write tests that deal
   # with other versions of bundler and we are unsure how to
   # handle this better.
-  VERSION = "1.15.4" unless defined?(::Bundler::VERSION)
+  VERSION = "2.15.4" unless defined?(::Bundler::VERSION)
```
colby-swandale: is the version suppose to be bumped?

segiddins: oops, nope

segiddins:
Fixed

indirect:
@segiddins This looks good, r+ once it's green. How do you feel about changing the setting name to `auto_clean_default_path`?

segiddins:
> How do you feel about changing the setting name to auto_clean_default_path?

But it won't auto-clean the system gem path, which is the default on 1.x

segiddins:
Should be fixed on 2.0 now!

indirect:
Nice. @bundlerbot r+

PR #5964, opened 2017-08-22
deprecate the bundle viz command

### LINKED ISSUE #5180: Move `bundle viz` into a plugin
`bundle viz` generates a basic ERD graph using `graphviz`. While useful, it could be improved, which would expand its scope and footprint. More realistically, it is a neat feature that could be split off into its own project and loaded using the new plugin API. To that end, I'd like to do the following:

- [ ] split existing functionality into a new `bundler-viz` project
- [ ] load `bundler-viz` using plugin API

This would remove the `graphviz` dependency from bundler core, allow for forward development to improve bundler visualization tools, and provide a roadmap for future bundler plugin development by interested parties.

cc @lynnco and @indirect

colby-swandale:

The `bundle viz` command is to been removed from bundler 2 and extracted into a plugin

Closes #5180 

Deprecate the `bundle viz` command with an error explaining that a new gem will replace it's functionality.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -482,20 +482,23 @@ module Bundler
       end
     end
 
-    desc "viz [OPTIONS]", "Generates a visual dependency graph", :hide => true
-    long_desc <<-D
-      Viz generates a PNG file of the current Gemfile as a dependency graph.
-      Viz requires the ruby-graphviz gem (and its dependencies).
-      The associated gems must also be installed via 'bundle install'.
-    D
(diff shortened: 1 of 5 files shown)


segiddins:
We also need a feature flag around the definition of the command in CLI so that a plugin can pick it up

segiddins:
`Cannot use `viz_command` as a settings feature flag since it isn't a bool key`

segiddins:
@indirect thoughts?

indirect:
@segiddins This works for me. If @colby-swandale is too busy to finish it, can you rebase it and fix the bool key thing?

segiddins:
Bye-bye `bundle viz`

@bundlerbot r+

PR #5969, opened 2017-08-25
Space between { and | missing

rocLv:

This does not follow ruby style best practice.

There should be space between { and |.

Add a space between { and |.

I chose this fix because most of us follow the ruby guide style.

### DIFF
--- a/lib/bundler/templates/Gemfile
+++ b/lib/bundler/templates/Gemfile
@@ -2,6 +2,6 @@
 
 source "https://rubygems.org"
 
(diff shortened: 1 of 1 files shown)


segiddins:
The Bundler repo itself does not have a space

colby-swandale:
Thanks for opening a PR and helping improve Bundler. We have our own style guide that is a set of rules defined in Rubocop to help keep a consistent style in the code base.

Sadly this change [does break our style guide](https://travis-ci.org/bundler/bundler/jobs/268208496#L644) so we cannot accept this PR.

PR #5972, opened 2017-08-25
Respect RUBYGEMS_HOST env var in release messaging

wadetandy:

When running the `release` task, bundler didn't know anything
about the RUBYGEMS_HOST environment variable that rubygems respects when
doing a gem push.  This resulted in incorrect messaging, as environments
where this was set without an `allowed_push_host` configured would
message that the gem had been pushed to rubygems.org, when in fact it
was pushed elsewhere.

Bundler was hardcoding `rubygems.org` in the event that an `allowed_push_host` setting was not specified, and didn't know anything about the `RUBYGEMS_HOST` env var.

Added a check for that variable and used it before the hardcoded `rubygems.org` if it exists.

I chose this fix because it seemed the most straightforward way to solve the problem.

### DIFF
--- a/lib/bundler/gem_helper.rb
+++ b/lib/bundler/gem_helper.rb
@@ -118,7 +118,11 @@ module Bundler
     end
 
     def gem_push_host
-      allowed_push_host || "rubygems.org"
+      env_rubygems_host = ENV["RUBYGEMS_HOST"]
+      env_rubygems_host = nil if
+        env_rubygems_host && env_rubygems_host.empty?
+
+      allowed_push_host || env_rubygems_host || "rubygems.org"
     end
 
     def perform_git_push(options = "")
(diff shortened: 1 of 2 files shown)


segiddins:
Thanks for the PR! Could you please look into adding some test coverage? A unit test should be sufficient

wadetandy:
@segiddins Specs added.

colby-swandale:
Thanks for the PR, one of the builds failed [on travis](https://travis-ci.org/bundler/bundler/jobs/268503503)  that we need to fix. Thanks!

colby-swandale:
ping @wadetandy

wadetandy:
@colby-swandale Whoops missed your message from the other day.  Specs fixed.

segiddins:
thanks!

@bundlerbot r+

PR #5973, opened 2017-08-25
Update Molinillo to 0.6.2

segiddins:

The problem was Molinillo 0.6.2 fixes a potential problem.

My fix is to update to Molinillo 0.6.2.


segiddins:
@bundlerbot r+

PR #5976, opened 2017-08-26
Fix bundle binstubs stanadlone flag

colby-swandale:

Currently, specifying the `--standalone` flag breaks `bundler binstubs` unless you have the `path` setting set in your bundle config. This is breaking in different ways but mostly for the same reason.

Bundler Master:
```
RuntimeError: Can't standalone without an explicit path set
  /Users/c/Desktop/Projects/bundler/lib/bundler/installer.rb:158:in `generate_standalone_bundler_executable_stubs'
  /Users/c/Desktop/Projects/bundler/lib/bundler/cli/binstubs.rb:36:in `block in run'
  /Users/c/Desktop/Projects/bundler/lib/bundler/cli/binstubs.rb:26:in `each'
```

Bundler 1.15.4:
```
TypeError: no implicit conversion of nil into String
  /Users/c/.rubies/ruby-2.4.1/lib/ruby/2.4.0/pathname.rb:409:in `initialize'
  /Users/c/.rubies/ruby-2.4.1/lib/ruby/2.4.0/pathname.rb:409:in `new'
  /Users/c/.rubies/ruby-2.4.1/lib/ruby/2.4.0/pathname.rb:409:in `join'
  /Users/c/.gem/ruby/2.4.1/gems/bundler-1.15.4/lib/bundler/installer.rb:142:in `generate_standalone_bundler_executable_stubs'
```

This was not caught in our tests because the `--standalone` flag was infact not being tested.

Another issue is that we have command options for the `bundle binstubs` command which are missing documentation in the man pages.

This error occurs when the `path` setting is not set which `generate_standalone_bundler_executable_stubs` expects to be set

Set a temporary path setting in the binstubs command of either any currently set `path` setting or use `Bundler.root` instead.

This was the most simplest solution that i can think of without having to change the standalone binstub generator.

### DIFF
--- a/lib/bundler/cli/binstubs.rb
+++ b/lib/bundler/cli/binstubs.rb
@@ -33,7 +33,9 @@ module Bundler
 
         if options[:standalone]
           next Bundler.ui.warn("Sorry, Bundler can only be run via RubyGems.") if gem_name == "bundler"
-          installer.generate_standalone_bundler_executable_stubs(spec)
+          Bundler.settings.temporary(:path => (Bundler.settings[:path] || Bundler.root)) do
+            installer.generate_standalone_bundler_executable_stubs(spec)
+          end
         else
           installer.generate_bundler_executable_stubs(spec, :force => options[:force], :binstubs_cmd => true)
         end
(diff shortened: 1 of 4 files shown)


segiddins:
Seems reasonable to me!
@bundlerbot r+

Issue #4904, opened 2016-08-21
Add fancy GitHub release notes to `rake release`


indirect:
It would be great to be able to provide a fancy GitHub releases page message along with gem versions when running `rake release`.

segiddins:
We can just slice the appropriate section from the CHANGELOG and use the GH API (as long as we have some way to grab an access token). See https://github.com/CocoaPods/Rainforest for one implementation of this

segiddins:
How is this for v1.15.4?

```
Bugfixes:

  - handle file conflicts gracefully in `bundle gem` (@rafaelfranca, @segiddins)
  - bundler will fail gracefully when the bundle path contains the system path separator (#5485, ajwann)
  - failed gem downloads will be retried consistently across different RubyGems versions (@shayonj)
  - `bundle pristine` will respect build options while re-building native extensions (@NickLaMuro)

## 1.15.3 (2017-07-21)

Bugfixes:

  - ensure that empty strings passed to `bundle config` are serialized & parsed properly (#5881, @segiddins)
  - avoid printing an outdated version warning when running a parseable command (@segiddins)

## 1.15.2 (2017-07-17)

Features:

  - new gemfiles created by bundler will include an explicit `github` git source that uses `https` (@segiddins)

Bugfixes:

  - inline gemfiles work when `BUNDLE_BIN` is set (#5847, @segiddins)
  - avoid using the old dependency API when there are no changes to the compact index files (#5373, @greysteil)
  - fail gracefully when the full index serves gemspecs with invalid dependencies (#5797, @segiddins)
  - support installing gemfiles that use `eval_gemfile`, `:path` gems with relative paths, and `--deployment` simultaneously (@NickLaMuro)
  - `bundle config` will print settings as the type they are interpreted as (@segiddins)
  - respect the `no_proxy` environment variable when making network requests (#5781, @jakauppila)
  - commands invoked with `--verbose` will not have default flags printed (@segiddins)
  - allow `bundle viz` to work when another gem has a requirable `grapviz` file (#5707, @segiddins)
  - ensure bundler puts activated gems on the `$LOAD_PATH` in a consistent order (#5696, @segiddins)

## 1.15.1 (2017-06-02)

Bugfixes:

  - `bundle lock --update GEM` will fail gracefully when the gem is not in the lockfile (#5693, @segiddins)
  - `bundle init --gemspec` will fail gracefully when the gemspec is invalid (@colby-swandale)
[...]

PR #5977, opened 2017-08-28
[Rakefile] Add a task to generate GitHub releases

segiddins:

The problem was we spend all this time writing up an awesome changelog, yet don't have proper GitHub releases.

Closes https://github.com/bundler/bundler/issues/4904.

My diagnosis was we needed to automate creating new GH releases, and tie that into the existing `rake release` process.

I've already taken the liberty of backfilling existing releases, see https://github.com/bundler/bundler/releases/tag/v1.14.5 for an example


indirect:
Ooh yes, this is great. Thanks! @bundlerbot r+

PR #5978, opened 2017-08-28
Support bundle_path used by war files generated with jruby's warbler

### LINKED ISSUE #5975: PathError in jruby warfile
I suspect that a recent change in bundler 1.15.4 to fix issue #5485 has broken bundler when executed inside a war file created with jruby.

The particular error we're seeing is:

```
(PathError) Your bundle path contains a ':', which is the path separator for your system. Bundler cannot function correctly when the Bundle path contains the system's PATH separator. Please change your bundle path to not include ':'.
Your current bundle path is 'uri:classloader:/WEB-INF/gems'.
```

This only fails on Linux (or Unix-type systems), where the path separator is a colon. It works fine on Windows.

Is there a way we can exclude path components matching that uri: pattern, or else somehow detect that we're running from within a war file and turn off that error?

torcido:

Closes #5975. The problem was that recent code to detect a bad bundle_path by looking for the PATH_SEPARATOR character broke when running from a war file generated by jruby's warbler.

My diagnosis was that the path separator used was not appropriate in this circumstance.

My fix implements the suggestion from @segiddins to use the same path splitting logic used by RubyGems under jruby to determine if the bad path is present.

I chose this fix because it provides a way to detect the original problem that is consistent with the handling within RubyGems.

### DIFF
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -221,7 +221,7 @@ module Bundler
   private
 
     def validate_bundle_path
-      path_separator = Gem.respond_to?(:path_separator) ? Gem.path_separator : File::PATH_SEPARATOR
+      path_separator = Bundler.rubygems.path_separator
       return unless Bundler.bundle_path.to_s.split(path_separator).size > 1
       message = "Your bundle path contains text matching #{path_separator.inspect}, " \
                 "which is the path separator for your system. Bundler cannot " \
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -221,12 +221,12 @@ module Bundler
   private
 
     def validate_bundle_path
-      return unless Bundler.bundle_path.to_s.include?(File::PATH_SEPARATOR)
-      message = "Your bundle path contains a '#{File::PATH_SEPARATOR}', " \
+      return unless Bundler.bundle_path.to_s.split(Gem.path_separator).size > 1
+      message = "Your bundle path contains text matching #{Gem.path_separator.inspect}, " \
                 "which is the path separator for your system. Bundler cannot " \
                 "function correctly when the Bundle path contains the " \
                 "system's PATH separator. Please change your " \
-                "bundle path to not include '#{File::PATH_SEPARATOR}'." \
+                "bundle path to not match #{Gem.path_separator.inspect}." \
                 "\nYour current bundle path is '#{Bundler.bundle_path}'."
       raise Bundler::PathError, message
     end
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -221,7 +221,8 @@ module Bundler
   private
 
     def validate_bundle_path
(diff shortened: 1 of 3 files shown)


segiddins:
Looks perfect, thanks!

spec/bundler/shared_helpers_spec.rb
```diff
@@ -261,20 +261,49 @@
...
         "system's PATH separator. Please change your " \
-        "bundle path to not include ':'.\nYour current bundle " \
+        "bundle path to not match \":\".\nYour current bundle " \
         "path is '#{Bundler.bundle_path}'."
       )
     end
 
+    it "does not exit if bundle path is the jruby/warbler standard uri path" do
+      allow(Gem).to receive(:path_separator).and_return(
+        /(?<!jar:file|jar|file|classpath|uri:classloader|uri|http|https):/
```
segiddins: on Ruby 1.8.7:

```
SyntaxError:
  /home/travis/build/bundler/bundler/spec/bundler/shared_helpers_spec.rb:280: undefined (?...) sequence: /(?<!jar:file|jar|file|classpath|uri:classloader|uri|http|https):/
  /home/travis/build/bundler/bundler/spec/bundler/shared_helpers_spec.rb:288: undefined (?...) sequence: /(?<!jar:file|jar|file|classpath|uri:classloader|uri|http|https):/
```

spec/bundler/shared_helpers_spec.rb
```diff
@@ -279,35 +279,37 @@
...
-    end
-
-    it "exits if bundle path contains another directory with the jruby uri path separator", :ruby => "> 1.8.7" do
-      allow(Gem).to receive(:path_separator).and_return(
-        /(?<!jar:file|jar|file|classpath|uri:classloader|uri|http|https):/
-      )
-      allow(Bundler).to receive(:bundle_path) {
-        Pathname.new("uri:classloader:/WEB-INF/gems:other/dir")
-      }
+    if RUBY_VERSION.to_f > 1.8 # In versions of jruby that supported ruby 1.8, the path separator was the standard File::PATH_SEPARATOR
```
segiddins: please use the `:ruby => "1.9"` tag on the specs that require this

lib/bundler/shared_helpers.rb
```diff
@@ -221,12 +221,13 @@ def md5_available?
   private
 
     def validate_bundle_path
-      return unless Bundler.bundle_path.to_s.include?(File::PATH_SEPARATOR)
-      message = "Your bundle path contains a '#{File::PATH_SEPARATOR}', " \
+      path_separator = Gem.respond_to?(:path_separator) ? Gem.path_separator : File::PATH_SEPARATOR
```
segiddins: this conditional belongs in a method in `RubygemsIntegration`

torcido: Done (via polymorphism instead of a conditional)

spec/bundler/shared_helpers_spec.rb
torcido: After many contortions, this is now done.

segiddins:
<3 thanks for putting up with me, this is great

@bundlerbot r+

Issue #5963, opened 2017-08-22
[2.0] clean up output from `bundle`


indirect:
```diff
--- current output of `bundle`
+++ desired output of `bundle`
@@ -1,21 +1,19 @@
 $ bundle
-2.0.0.dev
+Bundler version 2.0.0.dev
 
 Commands:
   bundle add GEM VERSION            # Add gem to Gemfile and run bundle install
   bundle binstubs GEM [OPTIONS]     # Install the binstubs of the listed gem
   bundle cache [OPTIONS]            # Locks and then caches all of the gems into vendor/cache
   bundle check [OPTIONS]            # Checks if the dependencies listed in Gemfile are satisfied by currently installed gems
-  bundle clean [OPTIONS]            # Cleans up unused gems in your bundler directory
   bundle config NAME [VALUE]        # Retrieve or set a configuration value
   bundle doctor [OPTIONS]           # Checks the bundle for common problems
   bundle env                        # Print information about the environment Bundler is running under
   bundle exec [OPTIONS]             # Run the command in context of the bundle
-  bundle gem GEM [OPTIONS]          # Creates a skeleton for creating a rubygem
+  bundle gem NAME [OPTIONS]         # Generates the skeleton of a new RubyGem project
   bundle help [COMMAND]             # Describe available commands or one specific command
   bundle info GEM [OPTIONS]         # Show information for the given gem
   bundle init [OPTIONS]             # Generates a Gemfile into the current working directory
-  bundle inject GEM VERSION         # Add the named gem, with version requirements, to the resolved Gemfile
   bundle install [OPTIONS]          # Install the current environment to the system
   bundle issue                      # Learn how to report an issue in Bundler
   bundle licenses                   # Prints the license of all gems in the bundle
@@ -24,14 +22,10 @@
   bundle open GEM                   # Opens the source directory of the given bundled gem
   bundle outdated GEM [OPTIONS]     # List installed gems with newer versions available
   bundle platform [OPTIONS]         # Displays platform compatibility information
-  bundle plugin SUBCOMMAND ...ARGS  # Manage the bundler plugins
-  bundle plugin help [COMMAND]      # Describe subcommands or one specific subcommand
-  bundle plugin install PLUGINS     # Install the plugin from the source
-  bundle pristine [GEMS...]         # Restores installed gems to pristine condition from files located in the gem cache. Gem installed from a git repository will be issued `git c...
[...]

PR #5979, opened 2017-08-29
[CLI] Prettify the output in cli_help

segiddins:

The problem was the new help output for `bundle` was very verbose and unorganized, making it look a bit like work vomit.

My diagnosis was we should 

Closes https://github.com/bundler/bundler/issues/5963.

My fix prints the full version info, as well as grouping the commands into `primary` and `utility` as done in the man pages.

I chose this fix because it makes the output pleasing to visually parse.

```
$ bundle
Bundler version 1.15.4 (2017-08-25 commit 3203fdd2a)

Bundler commands:

  Primary commands:
    bundle install [OPTIONS]    # Install the current environment to the system
    bundle update [OPTIONS]     # Update the current environment
    bundle package [OPTIONS]    # Locks and then caches all of the gems into vendor/cache
    bundle exec [OPTIONS]       # Run the command in context of the bundle
    bundle config NAME [VALUE]  # Retrieve or set a configuration value
    bundle help [COMMAND]       # Describe available commands or one specific command

  Utilities:
    bundle add GEM VERSION         # Add gem to Gemfile and run bundle install
    bundle binstubs GEM [OPTIONS]  # Install the binstubs of the listed gem
    bundle check [OPTIONS]         # Checks if the dependencies listed in Gemfile are satisfied by currently installed gems
    bundle console [GROUP]         # Opens an IRB session with the bundle pre-loaded
    bundle doctor [OPTIONS]        # Checks the bundle for common problems
    bundle env                     # Print information about the environment Bundler is running under
    bundle gem NAME [OPTIONS]      # Creates a skeleton for creating a rubygem
    bundle info GEM [OPTIONS]      # Show information for the given gem
    bundle init [OPTIONS]          # Generates a Gemfile into the current working directory
    bundle issue                   # Learn how to report an issue in Bundler
    bundle licenses                # Prints the license of all gems in the bundle
    bundle lock                    # Creates a lockfile without installing
    bundle open GEM                # Opens the source directory of the given bundled gem
    bundle outdated GEM [OPTIONS]  # List installed gems with newer versions available
    bundle platform [OPTIONS]      # Displays platform compatibility information
    bundle plugin                  # Manage the bundler plugins
    bundle pristine [GEMS...]      # Restores installed gems to pristine condition from files located in the gem cache. Gem installed from a git reposit...
    bundle show GEM [OPTIONS]      # Shows all gems that are part of the bundle, or the path to a given gem
    bundle version                 # Prints the bundler's version information

Options:
      [--no-color]                 # Disable colorization in output
  -r, [--retry=NUM]                # Specify the number of times you wish to attempt network commands
  -V, [--verbose], [--no-verbose]  # Enable verbose output mode
```

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -730,8 +730,8 @@ module Bundler
 
       installation = "To install the latest version, run `gem install bundler#{" --pre" if latest.prerelease?}`"
       if latest_installed && latest_installed > current
-        suggestion = "To update to the most recent installed version, run `bundle update --bundler`"
-        suggestion = "#{installation}\n#{suggestion}" if latest_installed > latest
+        suggestion = "To update to the most recent installed version (#{latest_installed}), run `bundle update --bundler`"
+        suggestion = "#{installation}\n#{suggestion}" if latest_installed < latest
       else
         suggestion = installation
       end
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -7,6 +7,8 @@ module Bundler
   class CLI < Thor
     require "bundler/cli/common"
 
+    package_name "Bundler"
+
     AUTO_INSTALL_CMDS = %w[show binstubs outdated exec open console licenses clean].freeze
     PARSEABLE_COMMANDS = %w[
       check config help exec platform show version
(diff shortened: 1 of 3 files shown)


lib/bundler/cli.rb
```diff
@@ -69,7 +71,26 @@ def self.deprecated_option(*args, &blk)
...
+
+      shell.say "Bundler commands:\n\n"
+
+      shell.say "  Primary commands:\n"
+      shell.print_table(primary_commands, :indent => 4, :truncate => true)
+      shell.say
+      shell.say "  Utilities:\n"
+      shell.print_table(utilities, :indent => 4, :truncate => true)
+      shell.say
+      self.class.send(:class_options_help, shell)
```
jules2689: would be nice if this wasnt a `send`

jules2689:
Seems fine to me :)

lib/bundler/cli.rb
segiddins: it's private, but since we vendor Thor it should be safe

jules2689: That might make thor difficult to upgrade in the future. What does `class_options_help` do and is there an alternative?

NickLaMuro:
@segiddins looks like you have to fix a test after your changes ;)

lib/bundler/cli.rb
segiddins: It formats and prints all of the class options

indirect:
@segiddens does `package` switch to `cache` when the 2.0 flag is turned on? If it does, r+

segiddins:
Yeah, although `package` becomes an alias. (essentially `cache` and `package` become the same command)

@bundlerbot r=indirect

lib/bundler/cli.rb
jules2689: Ah, hrm. You'd think that would be a public thing

segiddins:
@bundlerbot r=indirect

PR #5980, opened 2017-08-29
Convert pristine command desc to long_desc

NickLaMuro:

When taking a look at the output that was shown when doing `bundle cli_help`, half of the output for describing the `bundle pristine` command is truncated.

My diagnosis was that we had to long of a "basic description" (`desc`) for the `bundle pristine` command.

Convert the existing `desc` to a `long_desc`, and provide a shorter one for `desc`.

I chose this fix because it retained the same info that was there previously, but provided a much more terse description in summary like situations like `bundle cli_help`.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -603,7 +603,12 @@ module Bundler
       Issue.new.run
     end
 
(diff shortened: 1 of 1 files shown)


NickLaMuro:
@segiddins This is just something I noticed when taking a look at your work in https://github.com/bundler/bundler/pull/5979.

segiddins:
🚀 

@bundlerbot r+

PR #5981, opened 2017-08-29
[2.0] Implement config subcommands

### LINKED ISSUE #4600: Add subcommands to `config`

segiddins:

The problem was the current `bundle config` mega-command is hacky and confusing.

My diagnosis was we should add subcommands for `list`, `get`, `set`, and `unset` to make the CLI clearer.

Closes #4600.

My fix implements those subcommands while preserving the current bare `bundle config` command.

I chose this fix because, as opposed to https://github.com/bundler/bundler/pull/5507, we can keep the existing command in for a little while to help ease the transition.

@denniss how do you feel about this compared to your PR?

### DIFF
--- a/spec/commands/config_spec.rb
+++ b/spec/commands/config_spec.rb
@@ -362,6 +362,92 @@ E
       expect(out).to match(long_string_without_special_characters)
     end
   end
+
+  describe "subcommands" do
+    it "list" do
+      bundle! "config list"
+      expect(last_command.stdout).to eq "Settings are listed in order of priority. The top value will be used.\nspec_run\nSet via BUNDLE_SPEC_RUN: \"true\""
+
+      bundle! "config list", :parseable => true
+      expect(last_command.stdout).to eq "spec_run=true"
+    end
+
+    it "get" do
+      ENV["BUNDLE_BAR"] = "bar_val"
+
+      bundle! "config get foo"
+      expect(last_command.stdout).to eq "Settings for `foo` in order of priority. The top value will be used\nYou have not configured a value for `foo`"
+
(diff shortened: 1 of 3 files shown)


colby-swandale:
Is there any tests for this?

segiddins:
Not yet, need to add them if we're happy with the UX

segiddins:
Specs added @colby-swandale !

spec/commands/config_spec.rb
```diff
@@ -362,6 +362,92 @@
...
+      expect(last_command.stdout).to eq "spec_run=true"
+    end
+
+    it "get" do
+      ENV["BUNDLE_BAR"] = "bar_val"
+
+      bundle! "config get foo"
+      expect(last_command.stdout).to eq "Settings for `foo` in order of priority. The top value will be used\nYou have not configured a value for `foo`"
+
+      ENV["BUNDLE_FOO"] = "foo_val"
```
colby-swandale: I would like to encouraging using context blocks here but i'm not super fussed right now.

segiddins: That just makes the specs slower though

segiddins:
@indirect r?

indirect:
This is great, thanks! @bundlerbot r+

PR #5983, opened 2017-08-30
Add documentation for packaging with --no-install

asehra:

I need to package gems for deployment but don't want to install them on the local system. 

The behaviour is a part of bundler but it is not documented anywhere in man pages or the bundler site. To find the answer, one has to look at issues that were closed a few years ago.

I've updated the man page for the `bundle package` command. Hopefully, this will trickle down to bundler site pages for package command whenever CI publishes it. 

Followed the doc [here](https://github.com/bundler/bundler/blob/master/doc/documentation/WRITING.md) to run tests.

It felt appropriate to document a flag available on a CLI in its help pages.

### DIFF
--- a/man/bundle-package.ronn
+++ b/man/bundle-package.ronn
@@ -65,3 +65,8 @@ machine and check in the gems. For instance, you can run
 `bundle package` on an identical staging box during your
 staging process, and check in the `vendor/cache` before
 deploying to production.
+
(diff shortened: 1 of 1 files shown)


segiddins:
Looks good!

segiddins:
@bundlerbot r+

Issue #4629, opened 2016-05-31
Document source priority in 2.0


indirect:
In Bundler 2, all gems have to have an unambiguous source. Since all sources after the first must have blocks with gems in them, this is pretty straightforward:

``` ruby
source "rubygems"
source "private" do
  gem "depends_on_rack"
end
```

If a usable version of `rack` is present in the private source, it will be installed. If there is no usable version of rack in the private source, a usable version of rack from the "rubygems" source will be installed. We should document this in the Gemfile manpage and in the website guide to sources.

segiddins:
Also what if there are two gems that depend on rack in different source blocks, each which have rack?

indirect:
💥 (undefined, and we should figure that out)

segiddins:
IMHO that should be an error, and you should have to specify rack in one of them

RochesterinNYC:
I'm in support of requiring more explicitness in the `Gemfile` in such cases (or in general when it comes to sources).

ntl:
When more than one source hosts a gem of the same name, and that gem is a dependency of a gem listed in the `Gemfile` (but is not itself listed in the `Gemfile`), then I think bundler should prefer the source of the gem that depends upon it always. Once ambiguity is introduced by multiple gems from different sources depending on that gem, an explicit source ought to be mandated by bundler.

Currently, the state of working with multiple sources in ruby is a bit frustrating, because gems aren't canonically referenced by fully qualified URLs. Instead, there is a global namespace ala DNS.  I, for one, am tired of getting stumped to come up with a name every time I publish an open source library in ruby. As it stands, bundler & rubygems works great if you want to throw up a private server with a few gems to augment what's on http://rubygems.org, but I think this change would move more in the direction of being able to support more decentralized setups, which is a big win, in my view.

 So, I think I'm in agreement with @RochesterinNYC . Just wanted to voice my preference as well.

ntl:
Another note: I think the behavior of dependencies-of-dependencies ought to be the same as the behavior of dependencies. So, going along @indirect 's example above, suppose I have:

```ruby
source 'https://source-a.com'

source 'https://source-b.com' do
  gem 'depends_on_something_that_depends_on_rack'
end
```

In that case, I believe that bundler should prefer any gem called `rack` from `source-b` over `source-a`. Obviously, if there's any gem not hosted on `source-b` that depends on `rack`, bundler should fail unless an explicit source for `rack` is provided.

segiddins:
[...]

PR #5985, opened 2017-08-30
[2.0] [Resolver] Error when it is ambigous which transitive source a gem should come from

segiddins:

The problem was the "source priority" in ambiguous source situations was ... ambiguous.

My diagnosis was we should error and require a user explicitly pin the dependency to a source in those situations, rather than leaving the source used up to an implementation detail.

My fix attempts to implement the priority described in the conversation in https://github.com/bundler/bundler/issues/4629.

I chose this fix because it still allows using the default source as a backup, while only taking the "relevant" sources into account, so that the error/warning is not overzealous.


segiddins:
@indirect r?

indirect:
This looks right to me! @bundlerbot r+

PR #5991, opened 2017-08-31
Create quality spec for docs in spanish

arbonap:

The problem was...
- Now that Spanish localization is under-way, there are no specs that test the localized documentation.

My diagnosis was...
- To write a test that filters out gender-specific pronouns and "_Well, actually..._"'s and the like in Spanish

My fix...
- I looked at `quality_spec.rb` and took the format from that spec and localized it to be Spanish-specific.

I chose this fix because...
- testing documentation quality should also exist for other languages

### DIFF
--- a/spec/quality_es_spec.rb
+++ b/spec/quality_es_spec.rb
@@ -1,0 +1,79 @@
+# encoding: utf-8
+# frozen_string_literal: true
+
+if defined?(Encoding) && Encoding.default_external.name != "UTF-8"
+  # An approximation of ruby -E UTF-8, since it works on 1.8.7
+  Encoding.default_external = Encoding.find("UTF-8")
+end
+
+RSpec.describe "La biblioteca si misma" do
+  def check_for_expendable_words(filename)
+    failing_line_message = []
+    useless_words = %w[
+      básicamente
+      claramente
+      sólo
+      solamente
+      obvio
+      obviamente
+      fácil
+      fácilmente
+      sencillamente
+      simplemente
+    ]
+    pattern = /\b#{Regexp.union(useless_words)}\b/i
+
+    File.readlines(filename).each_with_index do |line, number|
+      next unless word_found = pattern.match(line)
+      failing_line_message << "#{filename}:#{number.succ} contiene '#{word_found}'. Esta palabra tiene un significado subjetivo y es mejor obviarla en textos técnicos."
+    end
+
+    failing_line_message unless failing_line_message.empty?
+  end
+
+  def check_for_specific_pronouns(filename)
+    failing_line_message = []
+    specific_pronouns = /\b(él|ella|ellos|ellas)\b/i
+
+    File.readlines(filename).each_with_index do |line, number|
+      next unless word_found = specific_pronouns.match(line)
+      failing_line_message << "#{filename}:#{number.succ} contiene '#{word_found}'. Use pronombres más genéricos en la documentación."
+    end
+
+    failing_line_message unless failing_line_message.empty?
+  end
+
+  RSpec::Matchers.define :be_well_formed do
+    match(&:empty?)
+
+    failure_message do |actual|
+      actual.join("\n")
+    end
+  end
+
+  it "mantiene la calidad de lenguaje de la documentación" do
+    included = /ronn/
+    error_messages = []
+    Dir.chdir(root) do
+      `git ls-files -z -- man`.split("\x0").each do |filename|
+        next unless filename =~ included
+        error_messages << check_for_expendable_words(filename)
+        error_messages << check_for_specific_pronouns(filename)
+      end
+    end
+    expect(error_messages.compact).to be_well_formed
+  end
+
+  it "mantiene la calidad de lenguaje de oraciones usadas en el código fuente" do
+    error_messages = []
+    exempt = /vendor/
+    Dir.chdir(root) do
+      `git ls-files -z -- lib`.split("\x0").each do |filename|
+        next if filename =~ exempt
+        error_messages << check_for_expendable_words(filename)
+        error_messages << check_for_specific_pronouns(filename)
+      end
+    end
+    expect(error_messages.compact).to be_well_formed
+  end
+end
--- a/spec/quality_spec.rb
+++ b/spec/quality_spec.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 if defined?(Encoding) && Encoding.default_external.name != "UTF-8"
-  # Poor man's ruby -E UTF-8, since it works on 1.8.7
+  # An approximation of ruby -E UTF-8, since it works on 1.8.7
   Encoding.default_external = Encoding.find("UTF-8")
 end
 


spec/quality_spec_es.rb
```diff
@@ -0,0 +1,80 @@
+# frozen_string_literal: true
+
+if defined?(Encoding) && Encoding.default_external.name != "UTF-8"
+  # Poor man's ruby -E UTF-8, since it works on 1.8.7
```
segiddins: let's not use gendered language -- something like "An approximation of ..."

spec/quality_spec_es.rb
```diff
@@ -0,0 +1,80 @@
...
+  def check_for_specific_pronouns(filename)
+    failing_line_message = []
+    specific_pronouns = /\b(él|ella|ellos|ellas)\b/i
+
+    File.readlines(filename).each_with_index do |line, number|
+      next unless word_found = specific_pronouns.match(line)
+      failing_line_message << "#{filename}:#{number.succ} tiene '#{word_found}'. Use pronombres más genericos en documentación."
+    end
+
+    failing_line_message unless failing_line_message.empty?
```
segiddins: needs to always return an array for the `be_well_formed` matcher

spec/quality_spec_es.rb
```diff
@@ -0,0 +1,80 @@
...
+      simplemente
+    ]
+    pattern = /\b#{Regexp.union(useless_words)}\b/i
+
+    File.readlines(filename).each_with_index do |line, number|
+      next unless word_found = pattern.match(line)
+      failing_line_message << "#{filename}:#{number.succ} tiene '#{word_found}'. No use estos tipos de modificantes."
+    end
+
+    failing_line_message unless failing_line_message.empty?
```
segiddins: needs to always return an array for the be_well_formed matcher

arbonap: @segiddins I'm not sure if I follow-- is `error_messages` ever not an array?

segiddins: if `failing_line_message` is empty, the method will return `nil`, and the `be_well_formed` matcher will try to call `nil.empty?` which will raise an exception

arbonap: @segiddins So, I commented out the `unless` clause on lines 32 and 44: `failing_line_message #unless failing_line_message.empty?`. 

I thought if I'm always returning `failing_line_message` in the `check_for_expendable_words` and `check_for_specific_pronouns` methods, raising the exception would be avoided since I'd always be returning an array...? 
But the specs break when I comment out that `unless` clause.🤔

segiddins: ah my bad, I was misunderstanding how the method was being used :/ Ignore me

arbonap:
@segiddins I followed your lead and decided to change the same comment in `quality_spec.rb` and make it more gender-neutral 👌🏼

spec/quality_spec_es.rb
```diff
@@ -0,0 +1,79 @@
...
+if defined?(Encoding) && Encoding.default_external.name != "UTF-8"
+  # An approximation of ruby -E UTF-8, since it works on 1.8.7
+  Encoding.default_external = Encoding.find("UTF-8")
+end
+
+RSpec.describe "La biblioteca si misma" do
+  def check_for_expendable_words(filename)
+    failing_line_message = []
+    useless_words = %w[
+      actualmente
```
deivid-rodriguez: This seems translated from English "actually", but it has a different meaning in Spanish. "Actualmente" means "nowadays" in Spanish. I don't think we want that word in this list?

spec/quality_spec_es.rb
```diff
@@ -0,0 +1,79 @@
...
+      fácil
+      fácilmente
+      sencillamente
+      simplemente
+    ]
+    pattern = /\b#{Regexp.union(useless_words)}\b/i
+
+    File.readlines(filename).each_with_index do |line, number|
+      next unless word_found = pattern.match(line)
+      failing_line_message << "#{filename}:#{number.succ} tiene '#{word_found}'. No use estos tipos de modificantes."
```
deivid-rodriguez: I'd say "contiene '#{word_found}'. Esta palabra no suele aportar nada a la frase y es mejor obviarla." to add some context on the reason for forbidding the word.

spec/quality_spec_es.rb
```diff
@@ -0,0 +1,79 @@
...
+    failing_line_message unless failing_line_message.empty?
+  end
+
+  def check_for_specific_pronouns(filename)
+    failing_line_message = []
+    specific_pronouns = /\b(él|ella|ellos|ellas)\b/i
+
+    File.readlines(filename).each_with_index do |line, number|
+      next unless word_found = specific_pronouns.match(line)
+      failing_line_message << "#{filename}:#{number.succ} tiene '#{word_found}'. Use pronombres más genéricos en documentación."
```
deivid-rodriguez: s/tiene/contiene/

In my opinion, it reads better with an article: "en _la_ documentación", or alternatively, "al documentar".

spec/quality_spec_es.rb
```diff
@@ -0,0 +1,79 @@
+# frozen_string_literal: true
```
segiddins: one last minor quibble: this file should be named `quality_es_spec.rb`, since we usually glob for `*_spec.rb`

spec/quality_es_spec.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
```
segiddins: I think we need the `# encoding: utf-8` comment for ruby < 2 support

arbonap: Gotcha. Does that mean ruby < 2 doesn't encode unicode properly without that magical comment?

segiddins: ruby < 2 defaulted to interpreting source as ASCII, ruby 2+ default to utf8. Ruby < 2 handled unicode just fine, it just didn't default to it

arbonap:
👋🏽 Seems like Travis builds are failing on ruby < 2 since I have some Spanish words with accents in `quality_es_spec.rb`. I added the `# encoding: utf-8` comment at the top, but it wasn't enough since Travis builds are failing.
 
In my local shell, I switched to `ruby -v 1.9.3`, did `export RUBYOPT=-Ku`, ran the test suite, and my specs passed locally. The `-Ku` part specifies the code set used to be UTF-8. 

So, I'm thinking I have to also configure Travis to do something similarly.  

In the `Rakefile`, where RubyGems specs are run by version, I then tried adding in `-Ku` to the `Rakefile` in places where `ENV["RUBYOPT"]` is being manipulated. 🤔

indirect:
@arbonap ah! we can add environment variables to `.travis.yml`: https://docs.travis-ci.com/user/environment-variables/

arbonap:
I'm at a loss of how to configure `travis.yml` to get all versions of ruby & rubygems in the build matrix passing so that ruby < 2 defaults to interpreting source as utf-8. Any thoughts?

segiddins:
I'd avoid moving that into an environment variable -- that might mask issues users would experience _without_ the variable set in other parts of Bundler.

@arbonap if you'd like, I'd be happy to take over getting it passing on Ruby < 2

arbonap:
✨ 💫 🙌🏽 ~ the build passed ~💙 💜 💖

segiddins:
@bundlerbot r+

PR #6003, opened 2017-09-05
Handle missing socket when warning about OpenSSL version

greysteil:

Stubbing Rubygems requests with WebMock was causing `undefined method 'io' for nil:NilClass` errors when using Bundler 1.16.0.pre.1

My diagnosis was that the new warning text about old OpenSSL versions didn't consider the case that a connection might not have an `@socket` variable set.

Guard against this by returning early in that case.

I chose this fix because it works, and because `Net::HTTP` itself has some guards in it around `nil` values for `@socket` ([example](https://github.com/ruby/ruby/blob/trunk/lib/net/http.rb#L858-L860)). This isn't my area, though, so it could be that a fix is needed in WebMock, not here...

### DIFF
--- a/lib/bundler/vendored_persistent.rb
+++ b/lib/bundler/vendored_persistent.rb
@@ -30,6 +30,7 @@ module Bundler
       return unless (uri.host || "").end_with?("rubygems.org")
 
       socket = connection.instance_variable_get(:@socket)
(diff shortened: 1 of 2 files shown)


segiddins:
good catch!
@bundlerbot r+

PR #6007, opened 2017-09-06
Use helper methods for relative path references in the specs

hsbt:
Ruby core needs to change `Spec::Path.root` and gemspec, bin, spec directories structure.

1. I changed spec directory from `spec` to `spec/bundler` because ruby core has rubyspec files under the `spec/rubyspec`.
2. I changed gemspec location to `bundler.gemspec` to `lib/bundler.gemspec`.

ref. https://bugs.ruby-lang.org/issues/12733#note-15

This pull request make we can modify root, gemspec path to flexible locations. After merging this pull request, I will add directory structure of ruby core repository to only `spec/support/path.rb`

### DIFF
--- a/spec/commands/exec_spec.rb
+++ b/spec/commands/exec_spec.rb
@@ -61,8 +61,6 @@ RSpec.describe "bundle exec" do
   it "handles --keep-file-descriptors" do
     require "tempfile"
 
-    bundle_bin = File.expand_path(bindir.join("bundle"), __FILE__)
-
     command = Tempfile.new("io-test")
     command.sync = true
     command.write <<-G
@@ -71,7 +69,7 @@ RSpec.describe "bundle exec" do
       else
         require 'tempfile'
         io = Tempfile.new("io-test-fd")
-        args = %W[#{Gem.ruby} -I#{lib} #{bundle_bin} exec --keep-file-descriptors #{Gem.ruby} #{command.path} \#{io.to_i}]
+        args = %W[#{Gem.ruby} -I#{lib} #{bindir.join("bundle")} exec --keep-file-descriptors #{Gem.ruby} #{command.path} \#{io.to_i}]
         args << { io.to_i => io } if RUBY_VERSION >= "2.0"
         exec(*args)
       end
--- a/spec/commands/exec_spec.rb
+++ b/spec/commands/exec_spec.rb
@@ -61,7 +61,7 @@ RSpec.describe "bundle exec" do
   it "handles --keep-file-descriptors" do
     require "tempfile"
 
-    bundle_bin = File.expand_path("#{Spec::Path.bindir}/bundle", __FILE__)
+    bundle_bin = File.expand_path(Spec::Path.bindir.join("bundle"), __FILE__)
 
     command = Tempfile.new("io-test")
     command.sync = true
@@ -474,7 +474,7 @@ RSpec.describe "bundle exec" do
       Bundler.rubygems.extend(Monkey)
       G
       bundle "install --deployment"
-      bundle "exec ruby -e '`#{Spec::Path.bindir}/bundler -v`; puts $?.success?'"
+      bundle "exec ruby -e '`#{Spec::Path.bindir.join("bundler")} -v`; puts $?.success?'"
       expect(out).to match("true")
     end
   end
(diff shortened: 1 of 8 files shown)


spec/support/helpers.rb
```diff
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+require File.expand_path("../path.rb", __FILE__)
```
colby-swandale: needs to be a space before this line

hsbt: Fixed it 😇

spec/support/path.rb
```diff
@@ -8,6 +8,18 @@ def root
       @root ||= Pathname.new(File.expand_path("../../..", __FILE__))
     end
 
+    def gemspec
+      @gemspec ||= Pathname.new(File.expand_path("#{root}/bundler.gemspec", __FILE__))
```
segiddins: `root.join("bundler.gemspec")`

spec/support/path.rb
```diff
@@ -8,6 +8,18 @@ def root
...
+    end
+
+    def bin
```
segiddins: `def bindir` to align with the method on `Gem::Specification`

spec/support/path.rb
```diff
@@ -8,6 +8,18 @@ def root
...
+      @bin ||= Pathname.new(File.expand_path("#{root}/exe", __FILE__))
```
segiddins: `root.join("exe")`

spec/support/path.rb
```diff
@@ -8,6 +8,18 @@ def root
...
+    end
+
+    def spec
```
segiddins: `spec_dir`

spec/support/path.rb
```diff
@@ -8,6 +8,18 @@ def root
...
+      @spec ||= Pathname.new(File.expand_path("#{root}/spec", __FILE__))
```
segiddins: `root.join("spec")`

spec/support/path.rb
```diff
@@ -83,7 +95,7 @@ def lib_path(*args)
     end
 
     def bundler_path
-      Pathname.new(File.expand_path("../../../lib", __FILE__))
+      Pathname.new(File.expand_path("#{root}/lib", __FILE__))
```
segiddins: `root.join("lib")`

spec/support/helpers.rb
```diff
@@ -88,11 +90,11 @@ def load_error_run(ruby, name, *args)
     end
 
     def lib
-      File.expand_path("../../../lib", __FILE__)
+      File.expand_path("#{Spec::Path.root}/lib", __FILE__)
```
segiddins: since `Spec::Path` is included most everywhere `Helpers` is included, you should just be able to use `root`

additionally, please use `Pathname.join` instead of all the string interpolation and the `File.expand_paths`

spec/support/helpers.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require File.expand_path("../path.rb", __FILE__)
```
segiddins: is this require necessary? `spec_helper.rb` requires everything in `spec/support` already

hsbt:
@segiddins I applied your feedback.

1. Use `Pathname#join` instead of string interpolation.
2. Remove `Spec::Path` on rspec examples.
3. Rename some methods.

spec/commands/exec_spec.rb
```diff
@@ -61,7 +61,7 @@
   it "handles --keep-file-descriptors" do
     require "tempfile"
 
-    bundle_bin = File.expand_path("#{Spec::Path.bin}/bundle", __FILE__)
+    bundle_bin = File.expand_path(bindir.join("bundle"), __FILE__)
```
segiddins: no need to use `File.expand_path` when you already have a `Pathname`. Additionally, `bindir` is absolute, so just `bindir.join("bundle")` should do the trick here

segiddins: (same goes for the rest of the specs, really)

hsbt: Make sense. I remove `File.expand_path` from a needless location.

segiddins:
Awesome, thanks!

@bundlerbot r+

Issue #5941, opened 2017-08-09
Check for corrupt file in cache


richlewis14:
Hi

Recently we have had a url filter implemented at work and this seems to have had an effect when installing gems using bundler. When running bundle install we now get 
 
    Bundler::GemspecError: Could not read gem at /path/to/cache/gem It may be corrupted

Upon inspection of that file it has actually saved the html of that page and not the gem itself in its binary format, the file is normally 

    application/x-tar; charset=binary

But we get back a text/html file

Is there something we can do to check the file type and then maybe retry if not x-tar ?

Thanks

arbonap:
hi @richlewis14, could you please post the output of `bundle env`?

colby-swandale:
If we retried to download the file again if it wasn't an `application/x-tar` then wouldn't you just be downloading the same file again? In terms of detecting the file type, we would have to make HTTP request for each gem to check the file type before downloading it. Which would greatly increase the time it takes of a `bundle install` to be complete.

colby-swandale:
Are you talking about redownloading the gem again (when you have fixed your URL filter) when you run `bundle install`? Or just re-try it again when what was downloaded is not a `.gem` file?

richlewis14:
@colby-swandale I don't think the url filter is going anywhere unfortunately, I was hoping there was a way to retry when what was downloaded is not a `.gem` file. Ideally just running bundle would retry x amount of times ? is this possible?

Output of `bundle env` @arbonap 

## Environment

```
Bundler   1.15.3
Rubygems  2.6.12
Ruby      2.3.3p222 (2016-11-21 revision 56859) [x86_64-darwin16]
GEM_HOME  /Users/lewisr66/.rvm/gems/ruby-2.3.3@hendricks-offline
GEM_PATH  /Users/lewisr66/.rvm/gems/ruby-2.3.3@hendricks-offline:/Users/lewisr66/.rvm/gems/ruby-2.3.3@global
RVM       1.29.2 (latest)
Git       2.11.0 (Apple Git-81)
Platform  x86_64-darwin-16
OpenSSL   OpenSSL 1.0.2l  25 May 2017
rubygems-bundler (1.4.4)
```

## Bundler settings

```
retry
  Set for your local app (/Users/lewisr66/Documents/echo_testing/hendricks-js-offline-integration/.bundle/config): 3
jobs
  Set for your local app (/Users/lewisr66/Documents/echo_testing/hendricks-js-offline-integration/.bundle/config): "3"
```

[...]

PR #6010, opened 2017-09-06
[Source::Rubygems] Remove .gem if downloaded package is invalid

segiddins:

The problem was the user could (once) have downloaded a `.gem` file that isn't actually a `.gem`, and that package would poison their cache.

Closes https://github.com/bundler/bundler/issues/5941.

My diagnosis was we should remove the `.gem` right after downloading it if we can't open it.

My fix `rm_rf`'s the `.gem` on failure.

I chose this fix because it won't accidentally nuke existing cache entries for a user, but it should help prevent Bundler propagating an issue.

### DIFF
--- a/lib/bundler/source/rubygems.rb
+++ b/lib/bundler/source/rubygems.rb
@@ -120,8 +120,14 @@ module Bundler
           uris.uniq!
           Installer.ambiguous_gems << [spec.name, *uris] if uris.length > 1
 
-          s = Bundler.rubygems.spec_from_gem(fetch_gem(spec), Bundler.settings["trust-policy"])
-          spec.__swap__(s)
+          path = fetch_gem(spec)
+          begin
+            s = Bundler.rubygems.spec_from_gem(path, Bundler.settings["trust-policy"])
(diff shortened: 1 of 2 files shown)


colby-swandale:
How does this work as end user? Will bundler just try and download the gem again or will this raise an error?

indirect:
If possible, I'd like to try to re-download one time, and then delete the file again and print an error explaining that we tried to download the gem <x> and the result was corrupted even on retry. 👍🏻

segiddins:
> If possible, I'd like to try to re-download one time, and then delete the file again and print an error explaining that we tried to download the gem and the result was corrupted even on retry. 👍🏻

Sure! Should be possible. Fair warning, I might not be able to get around to that for  few weeks, so someone else is more than welcome to take this over!

indirect:
Let's merge this for now, so that the cache will be cleared before the error, and hopefully it will work on the second run. I'll create a ticket for improving the UX.

@bundlerbot r+

olleolleolle:
@bundlerbot retry

PR #6011, opened 2017-09-07
Expand `env` output

indirect:
We've been printing the version of OpenSSL that Ruby was compiled against, but not the version that was dynamically linked at runtime, which is a huge potential issue. While I was there, I added a few more things that seemed helpful.

### DIFF
--- a/lib/bundler/env.rb
+++ b/lib/bundler/env.rb
@@ -88,12 +88,13 @@ module Bundler
 
     def self.version_of(script)
       return "not installed" unless Bundler.which(script)
-      %x"#{script} --version"
+      `#{script} --version`
     end
 
     def self.chruby_version
       return "not installed" unless Bundler.which("chruby-exec")
-      %x"chruby-exec -- chruby --version".lines.grep(/^chruby:/).first
+      `chruby-exec -- chruby --version`.
+        sub(/.*^chruby: (#{Gem::Version::VERSION_PATTERN}).*/m, '\1')
     end
 
     def self.environment
@@ -106,7 +107,7 @@ module Bundler
       out << ["  Config Dir", Gem::ConfigFile::SYSTEM_CONFIG_PATH]
       out << ["RubyGems", Gem::VERSION]
       out << ["  Gem Home", ENV.fetch("GEM_HOME") { Gem.dir }]
-      out << ["  Gem Path", ENV.fetch("GEM_PATH") { Gem.path.join(":") }]
+      out << ["  Gem Path", ENV.fetch("GEM_PATH") { Gem.path.join(File::PATH_SEPARATOR) }]
       out << ["  User Path", Gem.user_dir]
       out << ["  Bin Dir", Gem.bindir]
       out << ["OpenSSL"] if defined?(OpenSSL)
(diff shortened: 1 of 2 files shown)


lib/bundler/env.rb
```diff
@@ -86,18 +86,40 @@ def self.git_version
...
     end
 
+    def self.version_of(script)
+      return "not installed" unless Bundler.which(script)
+      %x"#{script} --version"
+    end
+
+    def self.chruby_version
+      return "not installed" unless Bundler.which("chruby-exec")
+      %x"chruby-exec -- chruby --version".lines.grep(/^chruby:/).first
```
segiddins: just `chruby --version` seems to work on my machine?

lib/bundler/env.rb
```diff
@@ -86,18 +86,40 @@ def self.git_version
       "not installed"
     end
 
+    def self.version_of(script)
+      return "not installed" unless Bundler.which(script)
+      %x"#{script} --version"
```
segiddins: please don't use straight delimiters, it becomes much harder to read imo. either use normal backticks or something like `%x{}`?

lib/bundler/env.rb
```diff
@@ -86,18 +86,40 @@ def self.git_version
...
-      out << ["GEM_HOME", ENV["GEM_HOME"]] unless ENV["GEM_HOME"].nil? || ENV["GEM_HOME"].empty?
-      out << ["GEM_PATH", ENV["GEM_PATH"]] unless ENV["GEM_PATH"].nil? || ENV["GEM_PATH"].empty?
-      out << ["RVM", ENV["rvm_version"]] if ENV["rvm_version"]
-      out << ["Git", git_version]
-      out << ["Platform", Gem::Platform.local]
-      out << ["OpenSSL", OpenSSL::OPENSSL_VERSION] if defined?(OpenSSL::OPENSSL_VERSION)
+      out << ["  Full Path", Gem.ruby]
+      out << ["  Config Dir", Gem::ConfigFile::SYSTEM_CONFIG_PATH]
+      out << ["RubyGems", Gem::VERSION]
+      out << ["  Gem Home", ENV.fetch("GEM_HOME") { Gem.dir }]
```
segiddins: any reason we can use `Gem.paths.home` for this and `Gem.paths.path` for the next line?

lib/bundler/env.rb
```diff
@@ -86,18 +86,40 @@ def self.git_version
...
+      out << ["  Gem Path", ENV.fetch("GEM_PATH") { Gem.path.join(":") }]
```
segiddins: keep in mind that `File::PATH_SEPARATOR` needn't equal `:`

lib/bundler/env.rb
```diff
@@ -86,18 +86,40 @@ def self.git_version
...
+      out << ["  Cert Dir", OpenSSL::X509::DEFAULT_CERT_DIR] if defined?(OpenSSL::X509::DEFAULT_CERT_DIR)
+      out << ["Tools"]
+      out << ["  Git", git_version]
+      out << ["  RVM", ENV.fetch("rvm_version"){ version_of("rvm") }]
+      out << ["  rbenv", version_of("rbenv")]
+      out << ["  chruby", chruby_version]
+
       %w[rubygems-bundler open_gem].each do |name|
         specs = Bundler.rubygems.find_name(name)
         out << [name, "(#{specs.map(&:version).join(",")})"] unless specs.empty?
```
segiddins: maybe pad these names under `Tools` now?

segiddins:
Updated. @indirect r?

lib/bundler/env.rb
indirect: We can use those instead, I was just copying from the `gem env` command 👍

segiddins: we actually can't yet, since they're 1.8+. we'll transition over eventually

indirect:
looks great! @bundlerbot r+

PR #6014, opened 2017-09-07
Add spec that pre-release versions aren't selected when not in the Gemfile

greysteil:

The problem was that a pre-release version was being installed when the user hadn't asked for one, and a non-prerelease install was possible.

My diagnosis was that this was caused by the change to the way pre-releases get selected for resolution when we moved to Molinillo 0.6.0. See the change to `lib/bundler/index.rb` in https://github.com/bundler/bundler/pull/5902.

My fix... isn't present yet. Basically we want to replicate the `wants_prerelease || only_prerelease` behaviour in `Bundler::Resolver#requirement_satisfied_by?`, but it's late and I haven't thought about how to do that yet. Instead, here's a failing spec.

I chose this fix because it's late and I haven't thought about how to fix this yet, but I at least wanted it flagged.

### DIFF
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -103,7 +103,7 @@ module Bundler
         results = index.search(dependency, @base[dependency.name])
 
         inital_requirement = @initial_requirements.find {|req| req.name == dependency.name }
-        if inital_requirement && !inital_requirement.requirement.prerelease?
+        if !inital_requirement || !inital_requirement.requirement.prerelease?
           # Move prereleases to the beginning of the list, so they're considered
           # last during resolution.
           pre, results = results.partition {|spec| spec.version.prerelease? }
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -42,6 +42,7 @@ module Bundler
     end
 
     def start(requirements)
+      @initial_requirements = requirements
       verify_gemfile_dependencies_are_found!(requirements)
       dg = @resolver.resolve(requirements, @base_dg)
       dg.map(&:payload).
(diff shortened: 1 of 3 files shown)


segiddins:
I don't think this is correct-- if a gem has a dependency on a prerelease of another gem, then activating a prerelease should be OK

greysteil:
@segiddins are you sure? It's a behaviour change from 1.15.4, where you'd only get a pre-release if you specifically asked for it, or if it was the only possibility. I think in the spec I've added here Bundler shouldn't be giving a pre-release version - it would make the user have to work hard to stay on a stable release if it did.

I've update this PR with a simple fix [nicked from Rubygems](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/dependency.rb#L329-L331).

indirect:
I think (although I'm not positive) that getting pres should be opt-in, and you should not get one unless the requirement explicitly sets a pre. At least, that's how it's always been in the past. Should we think about changing it?

greysteil:
@indirect @segiddins - updated to make pres explicit opt-in (or last resort) only. I believe the behaviour in this PR now matches 1.15.4.

segiddins:
I think something like the following will be more correct (you can have multiple initial requirements for the same dependency name) and more performant (limit the number of times we loop over that array):

```diff
diff --git a/lib/bundler/resolver.rb b/lib/bundler/resolver.rb
index e323d4b16..ce02f91ee 100644
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -42,7 +42,11 @@ module Bundler
     end
 
     def start(requirements)
-      @initial_requirements = requirements
+      @prerelease_specified = {}
+      requirements.each do |dep|
+        @prerelease_specified[dep.name] ||= dep.prerelease?
+      end
+
       verify_gemfile_dependencies_are_found!(requirements)
       dg = @resolver.resolve(requirements, @base_dg)
       dg.map(&:payload).
@@ -102,8 +106,7 @@ module Bundler
         index = index_for(dependency)
         results = index.search(dependency, @base[dependency.name])
 
-        inital_requirement = @initial_requirements.find {|req| req.name == dependency.name }
-        if !inital_requirement || !inital_requirement.requirement.prerelease?
+        unless @prerelease_specified[dependency.name]
           # Move prereleases to the beginning of the list, so they're considered
           # last during resolution.
           pre, results = results.partition {|spec| spec.version.prerelease? }

```

segiddins:
Can you also add a spec for where it's ok to resolve the the pre-release dependency?

Something along the lines of:

```
a (1.0)
  b (~> 1.1.0.beta.1)
b (1.0)
b (1.1.0.beta.1)
```

greysteil:
@segiddins - done. All yours!

segiddins:
<3

@bundlerbot r+

Issue #6013, opened 2017-09-07
ArgumentError: new key RB_USER_INSTALL


vtamara:
--- ERROR REPORT TEMPLATE -------------------------------------------------------
# Error Report

## Questions

Please fill out answers to these questions, it'll help us figure out
why things are going wrong.

- **What did you do?**

  I ran the command `/usr/local/bin/bundle install`

- **What did you expect to happen?**

 Install gems

- **What happened instead?**

Error

- **Have you tried any solutions posted on similar issues in our issue tracker, stack overflow, or google?**

   I was testing bundler 1.16.0.pre.2, will go back to bundler 1.15

- **Have you read our issues document, https://github.com/bundler/bundler/blob/master/doc/contributing/ISSUES.md?**

  No

## Backtrace

```
ArgumentError: new key RB_USER_INSTALL
  /usr/local/lib/ruby/gems/2.4/gems/bundler-1.16.0.pre.2/lib/bundler/shared_helpers.rb:279:in `set_env'
  /usr/local/lib/ruby/gems/2.4/gems/bundler-1.16.0.pre.2/lib/bundler/cli/install.rb:17:in `run'
  /usr/local/lib/ruby/gems/2.4/gems/bundler-1.16.0.pre.2/lib/bundler/cli.rb:224:in `block in install'
  /usr/local/lib/ruby/gems/2.4/gems/bundler-1.16.0.pre.2/lib/bundler/settings.rb:135:in `temporary'
  /usr/local/lib/ruby/gems/2.4/gems/bundler-1.16.0.pre.2/lib/bundler/cli.rb:223:in `install'
  /usr/local/lib/ruby/gems/2.4/gems/bundler-1.16.0.pre.2/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
  /usr/local/lib/ruby/gems/2.4/gems/bundler-1.16.0.pre.2/lib/bundler/vendor/thor/lib/thor/invocation.rb:126:in `invoke_command'
  /usr/local/lib/ruby/gems/2.4/gems/bundler-1.16.0.pre.2/lib/bundler/vendor/thor/lib/thor.rb:387:in `dispatch'
  /usr/local/lib/ruby/gems/2.4/gems/bundler-1.16.0.pre.2/lib/bundler/cli.rb:27:in `dispatch'
[...]

PR #6015, opened 2017-09-08
Add RB_USER_INSTALL to preserved ENV keys

olleolleolle:
This PR adds the environment variable `RB_USER_INSTALL` to the list of preserved keys. That variable is used on FreeBSD.

  - see #6013 

Without this change, Bundler's usage of it with the shared helpers to set a preserved env would raise this:

> ArgumentError: new key RB_USER_INSTALL

(This seems to be the only env var which is used but not listed in the preserved keys list.)

### DIFF
--- a/lib/bundler/environment_preserver.rb
+++ b/lib/bundler/environment_preserver.rb
@@ -14,6 +14,7 @@ module Bundler
       PATH
       RUBYLIB
       RUBYOPT
+      RB_USER_INSTALL
     ].map(&:freeze).freeze
     BUNDLER_PREFIX = "BUNDLER_ORIG_".freeze
 
--- a/lib/bundler/environment_preserver.rb
+++ b/lib/bundler/environment_preserver.rb
@@ -12,9 +12,9 @@ module Bundler
       GEM_PATH
       MANPATH
       PATH
+      RB_USER_INSTALL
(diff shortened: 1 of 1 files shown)


olleolleolle:
Perhaps we can check for usages:

```
$ egrep -h -o -R 'set_env "([A-Z_])+"' lib  | sort | uniq
set_env "BUNDLER_VERSION"
set_env "BUNDLE_BIN_PATH"
set_env "BUNDLE_GEMFILE"
set_env "GEM_HOME"
set_env "MANPATH"
set_env "PATH"
set_env "RB_USER_INSTALL"
set_env "RUBYLIB"
set_env "RUBYOPT"
```

lib/bundler/environment_preserver.rb
```diff
@@ -14,6 +14,7 @@ class EnvironmentPreserver
       PATH
       RUBYLIB
       RUBYOPT
+      RB_USER_INSTALL
```
segiddins: please keep the array sorted

olleolleolle: Thanks!

segiddins:
Thanks!
@bundlerbot r+

olleolleolle:
@bundlerbot retry

PR #6024, opened 2017-09-11
Allow user to override ~/.bundle with environment variables

gwerbin:

As in #4333, users wanted a way to make Bundler respect the XDG specification.

The directory `~/.bundle` was hard-coded and could not be configured.

Allow users to configure `~/.bundle` and its relevant sub-directories by setting environment variables. The environment variables and their fallbacks are as follows:

| variable | fallback if unset |
|---|---|
| `BUNDLE_USER_HOME` | `$HOME/.bundle` |
| `BUNDLE_USER_CACHE`  | `$BUNDLE_USER_HOME/cache` |
| `BUNDLE_USER_CONFIG` | `$BUNDLE_USER_HOME/config` |
| `BUNDLE_USER_PLUGIN` | `$BUNDLE_USER_HOME/plugin` |

Unlike https://github.com/bundler/bundler/pull/5787, This solution is not specific to the XDG specification. Users have all kinds of setups, and this is a very general system for allowing them to configure their development machines however they need. It tries to keep all files created by Bundler in the same place (as per https://github.com/bundler/bundler/pull/5787#issuecomment-310154273), but allows the user to override that convention _if they really want to and they know what they are doing_.

If they want to use XDG for everything, they can do it by explicitly setting the `BUNDLE_USER_*` variables to the equivalent `XDG_DATA_*`. If they just want to get `.bundle` out of their home directory, they can do it by setting `BUNDLE_USER_HOME` and don't have to mess with the more specific env variables if they don't want to.

To me, this solution strikes the right balance between "fine-grained control for power users" and "simple, sane defaults for everyone else".

Please let me know if my tests can be improved. My only Ruby experience so far has been writing Homebrew formulas and configuring Jekyll, so I'm sure I have a lot to learn.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -193,19 +193,19 @@ module Bundler
       raise e.exception("#{warning}\nBundler also failed to create a temporary home directory at `#{path}':\n#{e}")
     end
 
-    def user_bundle_path(dir="home")
+    def user_bundle_path(dir = "home")
       env_var, fallback = case dir
-                when "home"
-                  ["BUNDLE_USER_HOME", Pathname.new(user_home).join(".bundle")]
-                when "cache"
-                  ["BUNDLE_USER_CACHE", user_bundle_path.join("cache")]
-                when "config"
-                  ["BUNDLE_USER_CONFIG", user_bundle_path.join("config")]
-                when "plugin"
-                  ["BUNDLE_USER_PLUGIN", user_bundle_path.join("plugin")]
-                else
-                  raise BundlerError, "Unknown user path requested: #{dir}"
-                end
+                          when "home"
+                            ["BUNDLE_USER_HOME", Pathname.new(user_home).join(".bundle")]
+                          when "cache"
+                            ["BUNDLE_USER_CACHE", user_bundle_path.join("cache")]
+                          when "config"
+                            ["BUNDLE_USER_CONFIG", user_bundle_path.join("config")]
+                          when "plugin"
+                            ["BUNDLE_USER_PLUGIN", user_bundle_path.join("plugin")]
+                          else
+                            raise BundlerError, "Unknown user path requested: #{dir}"
+                          end
       # `fallback` will already be a Pathname, but Pathname.new() is
       # idempotent so it's OK
       Pathname.new(ENV.fetch(env_var, fallback))
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -193,8 +193,25 @@ module Bundler
       raise e.exception("#{warning}\nBundler also failed to create a temporary home directory at `#{path}':\n#{e}")
     end
 
-    def user_bundle_path
+    def user_bundle_path_default
       Pathname.new(user_home).join(".bundle")
+    end
+
+    def user_bundle_path(dir="home")
+      # if "home", user_bundle_path_default
+      env_var, fallback = case dir
+                when "home"
+                  ["BUNDLE_USER_HOME", user_bundle_path_default]
+                when "cache"
+                  ["BUNDLE_USER_CACHE", user_bundle_path.join("cache")]
+                when "config"
+                  ["BUNDLE_USER_CONFIG", user_bundle_path.join("config")]
+                when "plugin"
+                  ["BUNDLE_USER_PLUGIN", user_bundle_path.join("plugin")]
+                else
+                  nil
+                end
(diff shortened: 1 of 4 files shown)


lib/bundler.rb
```diff
@@ -193,8 +193,26 @@ def tmp_home_path(login, warning)
       raise e.exception("#{warning}\nBundler also failed to create a temporary home directory at `#{path}':\n#{e}")
     end
 
-    def user_bundle_path
-      Pathname.new(user_home).join(".bundle")
+    def user_bundle_path(dir="home")
```
segiddins: space around `=`

colby-swandale:
I have a few issues with this PR, mainly it's making things relating to the bundle path pretty confusing but this can be addressed in a later PR.

@bundlerbot r+

gwerbin:
@colby-swandale what issues do you have in mind? Obviously I'm happy to have my PR merged ("it works for me"), but I don't want to be responsible for any design debt in the future.

soc:
Maybe I missed something, but setting

```
export BUNDLE_USER_CACHE="$XDG_CACHE_HOME/bundler/cache"
export BUNDLE_USER_CONFIG="$XDG_CONFIG_HOME/bundler/config"
export BUNDLE_USER_PLUGIN="$XDG_DATA_HOME/bundler/plugin"
```

has no effect.

Running `bundle install` keeps creating .bundle and .bundles directories in the home directory.

I'm kind of surprised by this approach in general, as the various `BUNDLE_` envs seem to be extremely close to the respective XDG directories, but this approach discards all the benefits of simply using XDG:

- Working by default
- Not forcing users to coerce each individual application into conformance
- Being able to adjust directories for all applications instead of having to search for env directories for individual applications

jasonkarns:
@gwerbin The discussion above (mainly the env var fallbacks table) seemed to imply that the vars all indicate directories.

> Allow users to configure ~/.bundle and its relevant sub-directories by setting environment variables.

The code seems to indicate that `BUNDLE_USER_CONFIG` should point to a file, not a directory.

https://github.com/bundler/bundler/blob/23dfadcd8d561d48e89c7287cbbdd816e0a10ab8/lib/bundler/settings.rb#L392-L402

Is this the case? Can this be documented somewhere?

gwerbin:
@jasonkarns the relevant diff from my patch corresponding to that snippet is here: https://github.com/bundler/bundler/commit/e6cc7a24d5ca9da842aeefd23167be4d850c9aa5#diff-75219bd6bc0defc0e7f4ba322d92f950

As you can see, I didn't change anything that was previously stored under `~/.bundle`. My patch just allows the user to move the contents around.

It would have been more correct if I had written:

> Allow users to relocate `~/.bundle`, or specific contents thereof, by setting environment variables.

jasonkarns:
@gwerbin cool. Since this feature was significantly driven by (but not solely by) the XDG request, I was easily mislead by the directories vs files. Was having an issue with `BUNDLE_USER_CONFIG` not working when set to a directory: `$XDG_CONFIG_HOME/bundler`. I think the added confusion was because `BUNDLE_CONFIG` already existed as an env var. Since that var already existed, it didn't make sense to me why `BUNDLE_USER_CONFIG` would be created with identical use/value. (Therefore I assumed that `BUNDLE_USER_CONFIG` would be a directory which would _contain_ a file. ie `BUNDLE_CONFIG` essentially being the same as: `$BUNDLE_USER_CONFIG/config.yml` or something.

Anyway, thanks for the clarification.

jasonkarns:
Welp, I've found my problem!!! This apparently hasn't been released yet? I figured that since it was merged in September, and both 1.16.0 and 1.16.1 were released in October, December respectively, that this would have been included.

Does anyone have any rough ideas on when this might land?

soc:
@jasonkarns I'm pretty confused about the whole state of this, too. I also checked this a while ago and assumed the change wasn't shipped because it failed to fix the problem people reported in https://github.com/bundler/bundler/issues/4333, and devs wanted to rework this before officially releasing it.

segiddins:
We cut the 1.16 branch before this was merged, and have been focusing on getting the 2.0 train out the door, and tend to not backport new features after a branch cut

soc:
@segiddins Is there a chance to fix #4333 before 2.0 is released?

segiddins:
Since its a feature rather than a bug fix, it's likely it wont ship before 2.0

PR #6025, opened 2017-09-11
Avoid warning for stubbing `nil`

segiddins:
since we now test what happens if `socket` is `nil`, and `rspec-mocks` warns when we do that


segiddins:
@bundlerbot r+

PR #6030, opened 2017-09-13
[CLI] Switch around the --strict option for bundle outdated

segiddins:

The problem was

> outdated would be more consistent to change --strict to --filter-strict and --update-strict to --strict

My diagnosis was

> let's rename --strict to --filter-strict and rename the current --update-strict to --strict. In another ticket, later, we can expand the way outdated works to not need as many flags to show the relevant info

My fix switches around the options, by just making `--strict` an alias that means something different on 2.0 than on 1.0

I chose this fix because it allowed keeping temporary 1.x compatibility, versus wholesale renaming of the option.


indirect:
@bundlerbot r+

olleolleolle:
@bundlerbot retry

chrismo:
+1 - I didn't do a thorough review, but I'm cool with this.

PR #6031, opened 2017-09-13
[CLI] For now, disable the new version warning by default

segiddins:

The problem was people were getting confused by the terseness and bluntness of the current outdated bundler version warning.

My diagnosis was we need to spend time fine-tuning the message, including when it is displayed. But given the lack of time to do so immediately, I thought it best to disable the warning for now, until someone can make a PR that addresses the comments and concerns raised in #6004.

### DIFF
--- a/lib/bundler/settings.rb
+++ b/lib/bundler/settings.rb
@@ -73,6 +73,7 @@ module Bundler
     ].freeze
 
     DEFAULT_CONFIG = {
+      :disable_version_check => true,
       :redirect => 5,
       :retry => 3,
       :timeout => 10,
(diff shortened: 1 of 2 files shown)


stevensona:
If a user were to install a version with this change applied, then he would never again be made aware of a newer version unless attempting to run bundle install with a lock file that was created by a newer version. Probably don't want to make this the last version they'll ever use.

indirect:
@bundlerbot r+

mathieujobin:
should we consider a `1.15.5` release removing the warning? or 1.16.0 is around the corner?

PR #6036, opened 2017-09-17
Call IO.popen instead of backticks

nobu:
IO.popen with the command in an array doesn't need command line
quotes, and is safer.

Back-ticks with command line quoting is dangerous.

Back-ticks is not for complex purpose.

Use `IO.popen` and redirection option instead.

`IO.popen` and options of `spawn` have been introduced for such purpose.

### DIFF
--- a/lib/bundler/gem_helper.rb
+++ b/lib/bundler/gem_helper.rb
@@ -75,7 +75,7 @@ module Bundler
 
     def build_gem
       file_name = nil
-      sh(%W"gem build -V #{spec_path}") do
+      sh(%W[gem build -V #{spec_path}]) do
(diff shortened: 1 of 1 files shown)


olleolleolle:
@nobu Thanks for this change!

The CI build fails on a single style check: seems like `%i`-literals (` %i[child out]`) need to be lists of symbols instead (`[:child, :out]`).

colby-swandale:
I've created #6039 which sets Rubocop to target ruby version 2.3 which runs the rubocop parser correctly for this PR.  If you can please wait for that to be merged.

indirect:
@bundlerbot r+

segiddins:
@bundlerbot retry

Issue #6029, opened 2017-09-13
Git error when running Bundler from git checkout


indirect:
When running Bundler from a git checkout, as described in [SETUP.md](https://github.com/bundler/bundler/blob/master/doc/development/SETUP.md#development-setup), all commands print a git error.

Here's an example:

```
$ dbundle -v
fatal: Not a git repository (or any of the parent directories): .git
2.0.0.dev
```

The git error is coming from the gemspec, where we shell out to git:

https://github.com/bundler/bundler/blob/master/bundler.gemspec#L46

Git assumes that commands should run in the current directory, and so the gemspec's git command is running in the wrong place. To fix this, we should use the git option `-C` to pass the directory containing the gemspec file to the git command.

arbonap:
👋🏽 I'd love to work on this and help out.

colby-swandale:
@arbonap sure, let us know if you need any help solving this issue

olleolleolle:
@arbonap This issue here, #6053 fixes it, does it not?

PR #6053, opened 2017-09-22
Pass gemspec directory path to Git

arbonap:

The problem was...
- This WIP PR tries to solve Issue #6029 
- When running commands like `dbundle -v` as described in [SETUP.md](https://github.com/bundler/bundler/blob/master/doc/development/SETUP.md#development-setup), commands give us a git error.
- Like stated in Issue #6029, the git option `-C` should be used to pass the directory containing the gemspec file to the git command.

My diagnosis was...
- to give git the correct directory path where the gemspec file is

My fix...
- I'm wondering if I'm on the right track with this? 
- So, I ran `bin/rspec /spec/install/gemfile/gemspec_spec` and the specs passed. I then run the entire test suite and the `gemspec_spec` and a few other specs fail throughout the test suite. 
Then, I re-ran the failing specs files individually-- both as a whole, as well as on the failing lines, and yet they're green. I'm assuming I'll similarly get failing specs on Travis. I'm not really familiar with how to troubleshoot test suites with test pollution. 
- Could folks point me in the right direction? Any help would be much appreciated!

I chose this fix because...
- I chose to `require "pathname"`because I found it more intuitive to manipulate file paths with that library.

### DIFF
--- a/bundler.gemspec
+++ b/bundler.gemspec
@@ -4,6 +4,7 @@
 lib = File.expand_path("../lib/", __FILE__)
 $:.unshift lib unless $:.include?(lib)
 require "bundler/version"
+require "shellwords"
 
 Gem::Specification.new do |s|
   s.name        = "bundler"
@@ -43,7 +44,8 @@ Gem::Specification.new do |s|
   s.add_development_dependency "ronn",       "~> 0.7.3"
   s.add_development_dependency "rspec",      "~> 3.6"
 
-  s.files = `git ls-files -z`.split("\x0").select {|f| f.match(%r{^(lib|exe)/}) }
+  s.files = `git -C #{Shellwords.escape File.dirname(__FILE__)} ls-files -z`.split("\x0").select {|f| f.match(%r{^(lib|exe)/}) }
+
   # we don't check in man pages, but we need to ship them because
   # we use them to generate the long-form help for each command.
   s.files += Dir.glob("man/**/*")


bundler.gemspec
```diff
@@ -43,7 +44,9 @@ Gem::Specification.new do |s|
   s.add_development_dependency "ronn",       "~> 0.7.3"
   s.add_development_dependency "rspec",      "~> 3.6"
 
-  s.files = `git ls-files -z`.split("\x0").select {|f| f.match(%r{^(lib|exe)/}) }
+  gemspec_path = Pathname.glob("**/bundler.gemspec").first.dirname.to_path
+  s.files = `git -C #{gemspec_path} ls-files -z`.split("\x0").select {|f| f.match(%r{^(lib|exe)/}) }
```
ixti: There's no need in `Pathname` IMO. What you need is to figure our dirname of current `__FILE__`. So you can either use `File.dirname(__FILE__)` here, or simply using `__dir__`. Also it worth to shell-escape this path otherwise it will break on most of Windows setups (where spaces in paths are very common):

``` ruby
s.files = `git -C #{Shellwords.escape __dir__} ls-files -z`.split("\x0").select {|f| f.match(%r{^(lib|exe)/}) }
```

arbonap:
hi @ixti 👋 I implemented your advice, let me know what you think. I had to use `File.dirname(__FILE__)` instead of `__dirname__` so that Travis could pass on Ruby 1.8.7. Apparently the `__dirname__` method was introduced in Ruby 2.

ixti:
Oh. You're totally right. `__dir__` was [introduced](https://bugs.ruby-lang.org/issues/1961) in Ruby 2.0.0 - in fact by some reason I thought we're not supporting that old rubies :D

colby-swandale:
This looks good, @indirect can you test that this solves your problem?

indirect:
I'm a tiny bit worried that `Shellwords` doesn't work on Windows, but I guess that just means the gemspec won't work if your Windows path to the Bundler checkout has special characters in it. ¯\_(ツ)_/¯

@arbonap thanks!

@bundlerbot r+

segiddins:
@bundlerbot retry

PR #6054, opened 2017-09-23
improve .rubocop_todo.yml

walf443:

No

No

My fix reduce ruboco_todo.yml lines.

- Style/EmptyLinesAroundExceptionHandlingKeywords
- Style/SpaceAroundOperators
- Style/SpaceInsideBlockBraces
- Lint/DuplicateMethods
- Lint/Void
- Style/IfUnlessModifier
- Style/MixinGrouping
- Style/NestedParenthesizedCalls
- Style/OrAssignment
- Style/RedundantParentheses
- Style/TernaryParentheses


lib/bundler/spec_set.rb
```diff
@@ -61,7 +62,6 @@ def []=(key, value)
       @specs << value
       @lookup = nil
       @sorted = nil
-      value
```
walf443: This may break compatibility.

segiddins:
@bundlerbot r+

PR #6073, opened 2017-10-04
Fix uninitialised constant Gem::ConfigFile::SYSTEM_CONFIG_PATH

colby-swandale:

`bundle env` uses the constant `Gem::ConfigFile::SYSTEM_CONFIG_PATH` which is not defined in early versions of RubyGems.

See: https://travis-ci.org/bundler/bundler/builds/276981165

While `SYSTEM_CONFIG_PATH` is not defined for earlier versions of RubyGems, a constant called `Gem::ConfigFile::SYSTEM_WIDE_CONFIG_FILE` does appear to be both in current and early versions of RubyGems which contains the path (plus the ruby config file) we want.

This fix is compatible with all the tested versions of RubyGems in CI and is not overly complex.

### DIFF
--- a/lib/bundler/env.rb
+++ b/lib/bundler/env.rb
@@ -104,7 +104,7 @@ module Bundler
       out << ["  Platforms", Gem.platforms.join(", ")]
       out << ["Ruby", ruby_version]
       out << ["  Full Path", Gem.ruby]
-      out << ["  Config Dir", Gem::ConfigFile::SYSTEM_CONFIG_PATH]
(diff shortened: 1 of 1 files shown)


duckinator:
I'd recommend adding a small comment to that explaining why it's used in favor of `SYSTEM_CONFIG_PATH`, just so the context is there w/o digging around for the specific commit where you fixed it.

Something like `# SYSTEM_CONFIG_PATH is unavailable on older versions of RubyGems, so use SYSTEM_WIDE_CONFIG_FILE instead.`

indirect:
@colby-swandale thanks for figuring this out! 

@bundlerbot r+

PR #6077, opened 2017-10-05
Don't suggest upgrading to prerelease versions

gabebw:

The problem was that people on my team are reluctant to upgrade to a pre-release version of bundler due to fears about bugs, breaking changes, etc.

Teams may be reluctant to upgrade to pre-release versions and printing a warning message about a prerelease version each time `bundle install` is run is a little misleading if the team is on the latest stable version of Bundler.

My diagnosis was that Bundler is including pre-release versions in its check for the current maximum version of Bundler.

My fix is to not include pre-release versions in the version check.

I chose this fix (versus `bundle config disable_version_check true`)  because I believe that this problem probably affects other teams as well.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -724,6 +724,7 @@ module Bundler
                instance_variable_get(:@cache).
                dependencies("bundler").
                map {|d| Gem::Version.new(d.first) }.
+               reject(&:prerelease?).
                max
       return unless latest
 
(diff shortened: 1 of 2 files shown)


gabebw:
Hmm, CI is failing for Ruby 2.3.5 (and only 2.3.5) with an error about 

> RuntimeError: Digest::Base cannot be directly inherited in Ruby

Some Googling suggests that this might be because `Digest` isn't threadsafe; maybe the specs will pass if they're rerun?

colby-swandale:
The latest version notification has been disabled by default until we can decide a better approach.

gabebw:
@colby-swandale Thanks! Looks like that was merged as #6031, which is in v1.16.0.pre.3 (https://github.com/bundler/bundler/commit/e9eeb951027d7f28a6aa881c6401bba2d25c229a).

I'll close this PR.

PR #6081, opened 2017-10-06
Fix bundler update instructions in TROUBLESHOOTING

sunny:

Doing the previous command from the TROUBLESHOOTING guide returns an error:

```sh
$ `gem update bundler`
zsh: command not found: Updating
```

Removing both ``` ` ``` in the documentation.


olleolleolle:
@bundlerbot +r

PR #6080, opened 2017-10-06
Rename `-rubygems` option.

hsbt:
It only needs with Ruby 1.8 and Ruby 2.5 will remove it.

This commit picked r60125 from ruby/ruby: https://github.com/ruby/ruby/commit/9de6c712b66aad77df40661c1fc6d37e9a5c251a

PR #6059, opened 2017-09-24
use rubocop --parallel option

walf443:
### What was the end-user problem that led to this PR?
No

### What is your fix for the problem, implemented in this PR?

My fix enable rubocop's parallel option.
This may improve rubocop's test time.

PR #6084, opened 2017-10-09
Fix flaky tests

colby-swandale:

Currently in travis CI specs are failing intermittantly and makes merging code into master difficult. See: 

* https://github.com/bundler/bundler/pull/6080
* https://github.com/bundler/bundler/pull/6059

### DIFF
--- a/lib/bundler/rubygems_gem_installer.rb
+++ b/lib/bundler/rubygems_gem_installer.rb
@@ -48,7 +48,7 @@ module Bundler
       return true unless source = @package.instance_variable_get(:@gem)
       return true unless source.respond_to?(:with_read_io)
       digest = source.with_read_io do |io|
-        digest = Digest::SHA256.new
+        digest = Digest(:SHA256).new
         digest << io.read(16_384) until io.eof?
         io.rewind
         send(checksum_type(checksum), digest)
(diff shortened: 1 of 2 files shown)


olleolleolle:
@colby-swandale This seemed to clean up the situation immediately. Is this still work-in-progress?

colby-swandale:
yes

lib/bundler/rubygems_gem_installer.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "rubygems/installer"
+require "digest/sha2"
```
segiddins: instead of this, we should use `Digest(:SHA256)` on line 52?

colby-swandale: Yea, that does look better to use and it's theadsafe as well which i think is the the problem we're dealing with.  https://github.com/ruby/ruby/blob/trunk/ext/digest/lib/digest.rb#L96

segiddins: ping on making this change?

colby-swandale: updated, sorry for the delay 🙇

segiddins:
@bundlerbot r+

Issue #6028, opened 2017-09-13
Doubled informative message in verbose mode


indirect:
When running `bundle install --verbose`, the informational message about the state of the Gemfile, lockfile, and expected resolutions is doubled.

Here's an example of how to reproduce the issue:

```
$ dbundle init
fatal: Not a git repository (or any of the parent directories): .git
Writing new gems.rb to /Users/andre/src/bundler/testcases/new/gems.rb

$ dbundle add rake
fatal: Not a git repository (or any of the parent directories): .git
Fetching gem metadata from https://rubygems.org/.
Fetching gem metadata from https://rubygems.org/.........
Fetching gem metadata from https://rubygems.org/..
Resolving dependencies...
Fetching gem metadata from https://rubygems.org/.
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies...

$ dbundle install --verbose
fatal: Not a git repository (or any of the parent directories): .git
Running `bundle install --verbose` with bundler 2.0.0.dev
Found no changes, using resolution from the lockfile
Found no changes, using resolution from the lockfile
Using rake 12.1.0
0:  rake (12.1.0) from /Users/andre/src/bundler/testcases/new/.bundle/ruby/2.4.0/specifications/rake-12.1.0.gemspec
Using bundler 2.0.0.dev
0:  bundler (2.0.0.dev) from /Users/andre/src/bundler/bundler/lib/bundler/source
Bundle complete! 1 Gemfile dependency, 2 gems now installed.
Bundled gems are installed into `./.bundle`
```

The line `Found no changes, using resolution from the lockfile` is printed twice, when it only needs to be printed once.

segiddins:
Fixed by #6098.

PR #6098, opened 2017-10-14
Fix doubled message in verbose mode

akhramov:

The problem was doubled informative message in verbose mode (#6028)

`Bundler::Definition#missing_specs?` clears the `@resolve` ivar due to
the fact that `ensure` clause evaluates regardless whether exception
occurred. That results in repeated resolution, and, therefore, in
repeated debug messages.

This change moves ivars clearing into the `rescue` clause, so
`@resolve` will be cleared only in case when `BundlerError` occurs

I chose this fix because running deps resolution twice is a bad idea anyway and it needs to be prevented.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -205,19 +205,18 @@ module Bundler
     end
 
     def missing_specs?
-      missing = []
-      resolve.materialize(requested_dependencies, missing)
+      missing = missing_specs
       return false if missing.empty?
       Bundler.ui.debug "The definition is missing #{missing.map(&:full_name)}"
       true
     rescue BundlerError => e
-      Bundler.ui.debug "The definition is missing dependencies, failed to resolve & materialize locally (#{e})"
-      true
-    ensure
       @index = nil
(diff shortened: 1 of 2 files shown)


segiddins:
Cool, thanks!
@bundlerbot r+

segiddins:
@colby-swandale looks like the digest issue isn't entirely fixed: https://travis-ci.org/bundler/bundler/jobs/288028476#L775

colby-swandale:
dang, i'll have a look into this.

akhramov:
@colby-swandale, as you may know, the `Digest`'s problem relates not only to `SHA256` (see https://bugs.ruby-lang.org/issues/9494). Would it make sense to apply the fix to the rest of digests, i.e. to replace `Digest::DIGEST_NAME` with `Digest(:DIGEST_NAME)` in all places?

This should fix flaky tests on all rubies from 2.2 and up.

segiddins:
@akhramov yeah, I think we need to do that

colby-swandale:
@bundlerbot retry

PR #6100, opened 2017-10-15
improve development setup docs and remove redundant quotes

colby-swandale:

Some distrobutions of linux do not ship with the `col` command which `ronn` requires. And there are also some code blocks that don't need to quoted.

Update the developer docs to include instruction to install the package that has the `col` command and remove the code block quotes that don't need to be there.

### DIFF
--- a/doc/development/SETUP.md
+++ b/doc/development/SETUP.md
@@ -4,25 +4,29 @@ Bundler doesn't use a Gemfile to list development dependencies, because when we 
 
 1. [Fork the Bundler repo](https://github.com/bundler/bundler), and clone the fork onto your machine. ([Follow this tutorial](https://help.github.com/articles/fork-a-repo/) for instructions on forking a repo.)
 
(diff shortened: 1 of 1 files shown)


doc/development/SETUP.md
```diff
@@ -6,23 +6,27 @@ Bundler doesn't use a Gemfile to list development dependencies, because when we
 
 2. Install `groff-base` and `graphviz` packages using your package manager, e.g for Ubuntu:
 
-      `$ sudo apt-get install graphviz groff-base -y`
+      $ sudo apt-get install graphviz groff-base -y
+
+  *Note:* You may also have to install the `bsdmainutils` package on linux if your distrobution does not include the `col` command.
```
segiddins: `distribution`

colby-swandale:
Ah, sorry @segiddins, I really should have double checked before i pushed 😞

colby-swandale:
@bundlerbot retry

PR #6101, opened 2017-10-15
Load digest subclasses in a thread-safe manner

segiddins:

The problem was `digest` files would sometimes be loaded twice.

My diagnosis was we needed to load the `digest` files using the `::Digest` method, which is explicitly thread-safe.


colby-swandale:
Thanks for doing this. @bundlerbot r+

segiddins:
@bundlerbot retry

PR #6102, opened 2017-10-16
Fix "assigned but unused variable" warning in Ruby 2.5

amatsuda:
We're seeing several "assigned but unused variable" warnings when we run `bundle` in Ruby 2.5 with `RUBYOPT=-w`.
This is because Ruby 2.5's unused variable check has been improved since https://github.com/ruby/ruby/commit/37b04894f20a5d3701309a2055bfa03b9d760090, and [previous double-assignment technique](https://github.com/bundler/bundler/blob/fea177f61df71121c6f9422a74b093e42c504388/lib/bundler/installer.rb#L116) no longer works as an anti-warning workaround.

For example, here's a short reproducible code that doesn't warn in 2.4 but warns in 2.5:
```ruby
`def a() x = x = 1; end`
```

So here are updated workarounds against this warning, which are valid for both Ruby < 2.5 and >= 2.5.

### DIFF
--- a/lib/bundler/installer.rb
+++ b/lib/bundler/installer.rb
@@ -113,9 +113,12 @@ module Bundler
       end
 
       # double-assignment to avoid warnings about variables that will be used by ERB
-      bin_path = bin_path = Bundler.bin_path
-      relative_gemfile_path = relative_gemfile_path = Bundler.default_gemfile.relative_path_from(bin_path)
-      ruby_command = ruby_command = Thor::Util.ruby_command
+      bin_path = Bundler.bin_path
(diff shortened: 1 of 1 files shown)


segiddins:
I had opened a bug about this (https://bugs.ruby-lang.org/issues/13872), and this change breaks RuboCop -- you can run `bin/rake spec:deps && bin/rubocop` to lint your changes locally

amatsuda:
@segiddins Oops, sorry, fixed the rubocop warning. And responded to your "bug" report.

segiddins:
@bundlerbot r+

Issue #6066, opened 2017-09-28
Bundler 1.15+: Installing packaged gems from git repositories requires git executable on path


Gerg:
# Issue:

We package our gems as follows (on OSX dev machine): 
```
  BUNDLE_WITHOUT=development:test bundle package --all --all-platforms --no-install --path ./vendor/cache
```
[source](https://github.com/cloudfoundry/capi-release/blob/27f70b36a3cc8080e92682516a26832f85f6254f/packages/cloud_controller_ng/pre_packaging#L10)

On a different machine (on Linux packaging machine), we install from the packaged gems:
```
bundle install --local --deployment --without development test
```
[source](https://github.com/cloudfoundry/capi-release/blob/27f70b36a3cc8080e92682516a26832f85f6254f/packages/cloud_controller_ng/packaging#L12)

With Bundler 1.14.6, this works fine. Upon upgrading to 1.15.4, we started getting this error on the `bundle install`:

```
Command exited with 11; Stdout: installing your bundle as root will break this application for all non-root
users on this machine.
You need to install git to be able to use gems from git repositories. For help
installing git, please refer to GitHub's tutorial at
https://help.github.com/articles/set-up-git
, Stderr: + cp -a capi-release/src/cloud_controller_ng /var/vcap/packages/cloud_controller_ng
+ cd /var/vcap/packages/cloud_controller_ng/cloud_controller_ng
+ mariadb_dir=/var/vcap/packages/mariadb_connector_c
+ libpq_dir=/var/vcap/packages/libpq
+ bundle_cmd=/var/vcap/packages/ruby-2.4/bin/bundle
+ /var/vcap/packages/ruby-2.4/bin/bundle config build.pg --with-pg-lib=/var/vcap/packages/libpq/lib --with-pg-include=/var/vcap/packages/libpq/include
+ /var/vcap/packages/ruby-2.4/bin/bundle config build.mysql2 --with-mysql-config=/var/vcap/packages/mariadb_connector_c/bin/mariadb_config
+ /var/vcap/packages/ruby-2.4/bin/bundle install --local --deployment --without development test
```

Relevant bit: `Command exited with 11... You need to install git to be able to use gems from git repositories.`

# Expected Behavior

I shouldn't need git on my path to install packaged gems from git sources on bundler 1.15+.

# Bundle Env: 
## Environment

```
Bundler   1.15.4
Rubygems  2.6.13
[...]

PR #6112, opened 2017-10-20
Add a spec for installing git deps after packaging without git

segiddins:

The problem was v1.15 seemed to break installing without git when the git deps had already been packaged (see https://github.com/bundler/bundler/issues/6066).

My diagnosis was we actually seemed to have fixed this for 1.16.

My fix adds a test to ensure we won't regress.


deivid-rodriguez:
I rebased this one and it's passing, so I guess there's no harm on merging it.

deivid-rodriguez:
@bundlerbot r+

PR #6119, opened 2017-10-22
 Avoid activating stdlib digest under Ruby 2.5

segiddins:

The problem was, under ruby 2.5, Bundler would activate digest, which is now a stdlib gem, preventing users from choosing a different version.

With this change, the travis build against trunk should go green, hopefully.

My diagnosis was we needed to lazy-load digest.

My fix removes all `require "digest"` top-level statements from the Bundler codebase, and replaces the top-level `Digest` method with one that will `require "digest"`.

I chose this fix because it would keep the use of digest classes thread-safe.

\c @hsbt

### DIFF
--- a/bundler.gemspec
+++ b/bundler.gemspec
@@ -1,9 +1,7 @@
 # coding: utf-8
 # frozen_string_literal: true
 
-lib = File.expand_path("../lib/", __FILE__)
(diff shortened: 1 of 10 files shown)


colby-swandale:
@bundlerbot r+

PR #6123, opened 2017-10-25
Remove a blank line in generated gemspec for RUBY_VERSION >= "2.0.0"

meganemura:
### What was the end-user problem that led to this PR?

Need to remove a leading blank line of generated gemspec.

### What was your diagnosis of the problem?

```erb
<%- if RUBY_VERSION < "2.0.0" -%>
# coding: utf-8
<%- end -%>

lib = File.expand_path("../lib", __FILE__)
...
```

When executing `bundle gem xxx` with ruby (< 2.0), the leading blank line is appeared.

### What is your fix for the problem, implemented in this PR?

Move a blank line into the `if` block.

### Why did you choose this fix out of the possible options?

😃

PR #6125, opened 2017-10-25
Fix failing spec for `bundle plugin list`

meganemura:

The problem was failure of Travis CI build. (refs: [build for #6123](https://travis-ci.org/bundler/bundler/builds/292436821?utm_source=github_status&utm_medium=notification))

Bundler::Plugin::Index#installed_plugins is keys of Hash,
and Hash is not ordered in prior to Ruby 1.9.
`foo` and `bar` plugins defined in spec are not always listed in specified order.

My fix is to separate tests into for Ruby < 1.9 or not.

I chose this fix because this makes code easy to remove if we stop building Ruby < 1.9 in Travis CI.

### DIFF
--- a/spec/plugins/list_spec.rb
+++ b/spec/plugins/list_spec.rb
@@ -53,8 +53,18 @@ RSpec.describe "bundler plugin list" do
       plugin_should_be_installed("foo", "bar")
       bundle "plugin list"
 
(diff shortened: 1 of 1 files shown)


colby-swandale:
Thanks! @bundlerbot r+

PR #6128, opened 2017-10-26
Prioritise explicitly requested gems in dependency resolution sort order

greysteil:

Calling `bundle update github-pages` on https://github.com/Floppy/cv was causing a downgrade in the `github-pages` version.

The `Bundler::Resolver#sort_dependencies` method didn't have any logic to prioritise explicitly requested gems. Therefore, where two rival resolutions were possible (in this case one with `activesupport` at 5.1.4 and `github-pages` at 24, and one with `github-pages` at 166 and `activesupport` at 4.2.9), Bundler didn't always correct the most desirable resolution.

My fix is to explicitly prioritise requested gems in the dependency sort order during resolution.

I chose this fix because it's intuitive.

### DIFF
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -205,12 +205,14 @@ module Bundler
       dependencies.sort_by do |dependency|
         dependency.all_sources = relevant_sources_for_vertex(activated.vertex_named(dependency.name))
         name = name_for(dependency)
+        vertex = activated.vertex_named(name)
         [
           @base_dg.vertex_named(name) ? 0 : 1,
-          activated.vertex_named(name).payload ? 0 : 1,
+          vertex.payload ? 0 : 1,
+          vertex.root? ? 0 : 1,
           amount_constrained(dependency),
           conflicts[name] ? 0 : 1,
-          activated.vertex_named(name).payload ? 0 : search_for(dependency).count,
+          vertex.payload ? 0 : search_for(dependency).count,
           self.class.platform_sort_key(dependency.__platform),
         ]
       end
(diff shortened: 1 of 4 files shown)


segiddins:
Looks good, but this will delay releasing 1.16.0.
(Can you also send over a PR to Molinillo making the same change for the appropriate spec index class?)

greysteil:
I know 😢 . And yep, will do, will try to get this specced tonight.

lib/bundler/resolver.rb
```diff
@@ -208,6 +208,7 @@ def sort_dependencies(dependencies, activated, conflicts)
         [
           @base_dg.vertex_named(name) ? 0 : 1,
           activated.vertex_named(name).payload ? 0 : 1,
+          activated.vertex_named(name).root? ? 0 : 1,
```
segiddins: let's extract `activated.vertex_named(name)` into a local variable, since it's used multiple times

greysteil:
@segiddins - now with spec. All yours.

segiddins:
@bundlerbot r+

spec/resolver/basic_spec.rb
```diff
@@ -35,6 +35,13 @@
     should_resolve_as %w[berkshelf-2.0.7 chef-10.26 chef_app-1.0.0 json-1.7.7]
   end
 
+  it "prefers expicitly requested dependencies when resolving an index which would otherwise be ambiguous" do
```
isc: There is a missing `l` for `explicitly`.

greysteil: Ha, thanks! https://github.com/bundler/bundler/pull/6135

PR #6133, opened 2017-10-29
Ensure that autoloads are resolved in the real working directory

segiddins:
Fixes an issue on 2.5.0-dev where using local git overrides would cause an exception because the default bundler had been loaded

The problem was Bundler was failing 8 specs against ruby trunk.

My diagnosis was when the autoload for `URICredentialsFilter` fired in a `chdir` block, the default bundler gem was getting activated.

My fix eagerly trips that auto-load in the git proxy.

I chose this fix because it's a one-liner.

### DIFF
--- a/lib/bundler/source/git/git_proxy.rb
+++ b/lib/bundler/source/git/git_proxy.rb
@@ -221,6 +221,7 @@ module Bundler
 
         def in_path(&blk)
           checkout unless path.exist?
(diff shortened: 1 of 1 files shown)


segiddins:
Indeed, the ruby-head build now passes! https://travis-ci.org/bundler/bundler/jobs/294600700

colby-swandale:
Nice work! @bundlerbot r+

PR #6138, opened 2017-10-31
fix the links to the relevent GitHub PRs for the 1.16 stable changelog

colby-swandale:

The links to the GitHub issues for the 1-16 release are not correct, this PR fixes those.


segiddins:
Ugh thanks. I’ll fix this in bundler-changelog and just re-generate.

PR #5660, opened 2017-05-12
Ensure that manpages are always built when running specs

segiddins:
Fixes #5657 by ensuring manpages are always built, much like how `Spec::Rubygems.setup` ensures gem dependencies are always installed

\c @colby-swandale

PR #6143, opened 2017-11-02
Include `index.txt` in the man directory

abunashir:

In the bundler documentation site, most of the markdown reference-style links are broken. Looks like it's because of the missing `index.txt` file in the current man directory but the `ronn` expects that to map the link references to proper URLs.

Broken markdown style links (eg:)
- https://bundler.io/v1.16/man/bundle.1.html#PRIMARY-COMMANDS
- https://bundler.io/v1.16/man/bundle-install.1.html  (All the way to the bottom - summery section)
- And a couple of other pages where we are using markdown link reference.

### Why does it work fine in development or test?

In development or test, we normally invoke the `man:build`, which generates the index file on the fly and that's why it works fine but this does not include that updated file in the version control.

In the bundler site, we only clones the repo and generate the doc using the `ronn man/*.ronn`, which does not contain any reference file (`index.txt`) and that's why markdown link references are not working as expected.

This was working fine on the previous versions as it included the `index.txt` in the man directory and it was changed in PR #5660.

This commit includes the `index.txt` file in the man directory. This also updates the `.gitignore` file and `man: clean` task to not to remove the `index.txt` file when it's cleaning up the man directory. This should not have any side effect as whenever there are more man pages then running `:deps` task will update the `index.txt` file as well. Thanks in advance and please let me know if I am missing something here :)

Ronn: http://rtomayko.github.io/ronn/ronn.1.html#LINK-INDEXES


segiddins:
I think I’d rather just update the site to build the index file?

abunashir:
Yap, that would work great too, thanks @segiddins!

segiddins:
https://github.com/bundler/bundler-site/pull/349 should do the trick

colby-swandale:
Closing.

Issue #6114, opened 2017-10-20
Dependency resolution error message generator consumes ridiculous amounts of memory for large conflict trees


halfbyte:
Debugging a failing dependency update for a [Depfu](https://depfu.com) client, I found out that if the conflict tree gets to a certain size, the code to display the reduced conflict set consumes way too much memory. I've never let it finish but usually stopped the code when reaching the 4GB mark.

I've identified the following change as the culprit:

https://github.com/bundler/bundler/commit/5ebc8c1c6f0954332420475c9bf97a3bb30ce728#diff-a068dff3f4ba60cbf2c11bc4e65eed02R27

The code moved a bit since then and is currently at https://github.com/bundler/bundler/blob/f093e405a8e10bafa0b5d4621069a0b7ad86c37f/lib/bundler/resolver.rb#L311

The issue here is the use of Array#combination. The specific conflict trees array had a length of 26. Here's a simple test case to consume a ton of memory on your machine to exemplify the problem:

``` ruby
([1] * 26).combination(12)
```

(and please note that the bundler code from above would do that for every value from 1 to 26 as input for Array#combination.)

How to fix this?

One thing I noticed is that the unreduced "trees" variable contains a lot of 1:1 dupes. I've removed them with the following (not very elegant) code, inserted before [line 310](https://github.com/bundler/bundler/blob/f093e405a8e10bafa0b5d4621069a0b7ad86c37f/lib/bundler/resolver.rb#L310):

```ruby
trees.uniq! { |tree| [tree.first, tree.last] }
```

As my knowledge of the resolving process is limited, I am not sure if this will somehow ruin the intended output. It does bring down my 26 trees to 12 after which the reduction still takes a good 10 seconds or so on my machine but consumes way less memory. Not optimal, but better.

As I am afraid that the 26 I encountered are not the theoretical maximum of conflict trees, I think it would be a good idea to add a bailout clause, so that if the deduped trees size is > 10 or so, the reduction is skipped, which would change the full lambda to:

```ruby
trees.uniq! {|tree| [tree.first, tree.last] }
if trees.size <= 10
  maximal = 1.upto(trees.size).map do |size|
    trees.map(&:last).flatten(1).combination(size).to_a
  end.flatten(1).select do |deps|
    Bundler::VersionRanges.empty?(*Bundler::VersionRanges.for_many(deps.map(&:requirement)))
  end.min_by(&:size)
  trees.reject! {|t| !maximal.include?(t.last) } if maximal

[...]

PR #6142, opened 2017-11-02
[Travis] Test against RubyGems 2.7.0

segiddins:
Since I just released 2.7.0, we now have to test against it!

PR #6145, opened 2017-11-02
Bail out of reducing depency trees on huge dependency conflict sets

halfbyte:

The problem is described in #6114 - An unusually large set of conflicting dependencies leads to bundler get stuck and lead to very unhealthy amounts of memory and CPU time be consumed. 

This happens because because a too large array is fed into Array#combination in the "reduce_trees" lambda in `version_conflict_message`. 

By simply bailing out when the conflict set is bigger than a certain size,
we'll get a result that is similar to what happened in earlier bundler versions
but skip a ton of CPU cycles and memory allocations.

I've chosen the limit rather unscientifically by playing around with the result
set sizes. 15 seems to be a good compromise but really anything larger than 10
should work (and with work I mean "should not usually not have to be invoked").

Reducing the conflict trees has been a rather new addition to bundler, so defaulting back to the old behavior of not reducing the trees seems like an OK option. It may be possible to also optimize the reduction algorithm, but since this is very much an edge case that only happens when using Bundler slightly out of the normal procedures, I think a simple bail out is sufficient.

### DIFF
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -309,6 +309,8 @@ module Bundler
         :solver_name => "Bundler",
         :possibility_type => "gem",
         :reduce_trees => lambda do |trees|
+          # bail out if tree size is too big for Array#combination to make any sense
+          return trees if trees.size > 15
           maximal = 1.upto(trees.size).map do |size|
             trees.map(&:last).flatten(1).combination(size).to_a
           end.flatten(1).select do |deps|


halfbyte:
@segiddins Thanks for the approval. The test failure kind of baffles me, though. I don't really understand the failure message and also this seems to be a failing test against Ruby 1.8.7 - I thought bundler 2.0 is not supposed to run on anyithing below Ruby 2?

(Mind this could be me just not comprehending your test setup which is somewhat complex, understandably so)

halfbyte:
@segiddins After taking another look and seeing that your own PR #6142 fails with the exact same problem: Could this be something related to the 1.6 release? Anyway, not wanting to fix something that isn't really broken, I'll await your comment. (One thing I did notice is that the "bundle pristine" call does not use the full path to the bundler executable as the other calls prior seem to use, could this be the issue?)

segiddins:
@bundlerbot r+

PR #6153, opened 2017-11-06
Update `config` page link to the recent version

abunashir:

In the `man/bundle-gem.ronn`, we are using a hard-coded version number to link the `bundle-config` doc, which is `v1.14` in this case. Ref: https://bundler.io/v1.16/man/bundle-gem.1.html

<del>This commit updates the link and uses a relative path so it will always point to the recent version.</del>

Edit: This commit updates the link to the current stable version.

### DIFF
--- a/man/bundle-gem.ronn
+++ b/man/bundle-gem.ronn
@@ -75,4 +75,4 @@ configuration file using the following names:
 
 ## SEE ALSO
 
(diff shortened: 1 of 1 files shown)


abunashir:
Sure thing, thanks @colby-swandale! I will update the PR in a bit :)

colby-swandale:
Thank you! @bundlerbot r+

PR #6155, opened 2017-11-07
[FIX] skip bundler gem when warning when an explicitly updated spec d…

oded-zahavi:
…oes not get a newer version (618c09b59d1318958c23b1b0031c68c93186851a)

The problem was when running bundle update --group xxx 

My diagnosis was that bundler was taken into consideration and fails the updating process

My fix is to skip bundler when checking whether to warn when an explicitly updated spec does not get a newer version

I chose this fix because it is a rather new code (added in v1.16.0) and it is rather simple and doesn't require lots of changes in code. Another option would be to revert the lines added between 1.15.4 and 16.0.0 altogether. Up to the maintainers to decide.

### DIFF
--- a/lib/bundler/cli/update.rb
+++ b/lib/bundler/cli/update.rb
@@ -68,6 +68,8 @@ module Bundler
 
       if locked_gems = Bundler.definition.locked_gems
         gems.each do |name|
+          next if name == "bundler"
+
           locked_version = locked_gems.specs.find {|s| s.name == name }.version
           new_version = Bundler.definition.specs[name].first
           new_version &&= new_version.version


colby-swandale:
Thanks for opening a pull request, we will need a test for this change to make sure we don't regress in the future.

segiddins:
Looks good with an added test case!

oded-zahavi:
Thanks @segiddins any plans to merge any time soon?

colby-swandale:
@oded-zahavi This still requires a test and this also needs to rebase with master before we merge.

oded-zahavi:
Imho it shouldn't have passed previous CI because the bug was actually
inserted in 0.16.0. Working like a charm in 0.15.4
Anyways - Should I write tests of should it pass your tests?

segiddins:
Yes, please add a test so we don't regress in the future

colby-swandale:
ping @oded-zahavi

PR #6157, opened 2017-11-07
Fix NoMethodError during `bundle update --group`

mattbrictson:
### What was the end-user problem that led to this PR?

#6156: NoMethodError: undefined method `version' for nil:NilClass 

### What was your diagnosis of the problem?

The `bundler` gem does not participate in the lockfile, but it can still be included in the list of dependencies that are being updated by `bundle update` if `--group` is specified.

For example, if a Gemfile contains `bundler-audit` (which depends on `bundler`) in the `:development` group, then updating with the option `--group=development` will naturally include `bundler` in the list of gems to evaluate for updating.

The trouble is that since `bundler` is excluded from the lockfile, searching the locked gems for a gemspec for bundler will return `nil`. This caused the following error during `bundle update`:

    NoMethodError: undefined method `version' for nil:NilClass

### What is your fix for the problem, implemented in this PR?

This PR solves this bug by skipping over gems (i.e `bundler`) that are not in the lockfile when comparing gem versions at the conclusion of the upgrade command.

Fixes #6156.

### Why did you choose this fix out of the possible options?

I chose this fix because the bug seems to have been introduced by 618c09b59d1318958c23b1b0031c68c93186851a. My fix takes place within the new feature that was added in that commit, so it seems safe and unlikely to have side-effects.

PR #6161, opened 2017-11-08
Stop requesting changelogs in PRs

indirect:
We added this requirement before the changelog was automated. Now that the changelog generated by a script, it's not helpful for PRs to update the changelog anymore, so we shouldn't expect it anymore.

We had to [ask a PR to remove a changelog entry](https://github.com/bundler/bundler/pull/6157#discussion_r149701953), even though the PR was following [the docs](https://github.com/bundler/bundler/blob/374caf6ede2cf4de7d2196aa2e2adef782bea5bd/doc/development/PULL_REQUESTS.md#changelogmd).

It looks like we failed to update the docs when we automated generating the changelog.

Removing the changelog requirement from the docs hopefully means that future PRs will not include changelog entries, and we can generate future changelogs automatically.

I chose this fix because it seemed like the smallest change that would accomplish the goal.

/cc @mattbrictson


mattbrictson:
Nice! Thanks for updating that. 👍

segiddins:
@bundlerbot r+

PR #6170, opened 2017-11-13
Correctly set `loaded_from` when use installed gem

y-yagi:

In bundler 1.16.0, the following script using `bundler/inline` fails.

```ruby
# script.rb
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  gem "rails", github: "rails/rails"
  gem "arel", github: "rails/arel"
end
```

```
$ ruby script.rb
(snip)

.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/bundler-1.16.0/lib/bundler/runtime.rb:28:in `block in setup': rake-12.2.1 is missing. Run `bundle install` to get it. (Bundler::GemNotFound)
	from .rbenv/versions/2.4.1/lib/ruby/2.4.0/forwardable.rb:229:in `each'
	from .rbenv/versions/2.4.1/lib/ruby/2.4.0/forwardable.rb:229:in `each'
	from .rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/bundler-1.16.0/lib/bundler/runtime.rb:26:in `map'
	from .rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/bundler-1.16.0/lib/bundler/runtime.rb:26:in `setup'
	from .rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/bundler-1.16.0/lib/bundler/inline.rb:70:in `gemfile'
	from script.rb:3:in `<main>'
```

My diagnosis was that `loaded_from` was not set when using installed gem.

My fix is set `loaded_from` when `loaded_from` is nil

In my understanding, it seems there are no other options, so I chose this fix.

### DIFF
--- a/lib/bundler/source/rubygems.rb
+++ b/lib/bundler/source/rubygems.rb
@@ -107,6 +107,8 @@ module Bundler
         end
 
         if installed?(spec) && !force
+          spec.loaded_from ||= loaded_from(spec)
+
           print_using_message "Using #{version_message(spec)}"
           return nil # no post-install message
         end


colby-swandale:
Thanks for opening a PR, this will need a test to make sure we don't regress in the future.

y-yagi:
Hmmm, I'm trying to add a test, but it does not go well.
Although the above reproduction script works in the test, is it good to add the script as it is in the test?

lib/bundler/source/rubygems.rb
```diff
@@ -107,6 +107,8 @@ def install(spec, opts = {})
         end
 
         if installed?(spec) && !force
+          spec.loaded_from = loaded_from(spec) unless spec.loaded_from
```
segiddins: You can use `||=` here

y-yagi: Thanks, I fixed!

y-yagi:
Fixed by #6188

Issue #6172, opened 2017-11-14
[2.0] Replacement for `show —paths` still needed


indirect:
We’ve deprecated `show` in favor of `list` and `info`, which is a good change, IMO. Now that we’re planning on deprecating the `show` command, we need a replacement for `show —paths`, to print every directory containing the gems installed in the bundle. This is needed for workflows like `rg SEARCH $(bundle show —paths)`, which does a recursive search for SEARCH through all the source code in the entire bundle.

colby-swandale:
This feature has been merged. Closing.

PR #6177, opened 2017-11-19
add `--paths` option to `bundle list` command.

colby-swandale:
See #6172 

This option mimics the `--paths` options in `bundle show` which is
being removed in Bundler 2.

Example Usage:

```
  $ bundle list --paths
  /Users/c/Desktop/Projects/testapp/.bundle/ruby/2.4.0/gems/actioncable-5.1.4
  /Users/c/Desktop/Projects/testapp/.bundle/ruby/2.4.0/gems/actionmailer-5.1.4
  /Users/c/Desktop/Projects/testapp/.bundle/ruby/2.4.0/gems/actionpack-5.1.4
  /Users/c/Desktop/Projects/testapp/.bundle/ruby/2.4.0/gems/actionview-5.1.4
```

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -285,6 +285,7 @@ module Bundler
     if Bundler.feature_flag.list_command?
       desc "list", "List all gems in the bundle"
       method_option "name-only", :type => :boolean, :banner => "print only the gem names"
+      method_option "paths", :type => :boolean, :banner => "print the path to each gem in the bundle"
       def list
         require "bundler/cli/list"
         List.new(options).run
--- a/spec/commands/list_spec.rb
+++ b/spec/commands/list_spec.rb
@@ -8,6 +8,13 @@ RSpec.describe "bundle list", :bundler => "2" do
     G
   end
 
+  context "with name-only and paths option" do
+    it "raises an error" do
+      bundle "list --name-only --paths"
+      expect(out).to eq "The `--name-only` and `--paths` options cannot be used together"
+    end
+  end
+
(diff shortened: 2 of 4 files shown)


colby-swandale:
\cc @indirect

lib/bundler/cli.rb
```diff
@@ -285,6 +285,7 @@ def show(gem_name = nil)
     if Bundler.feature_flag.list_command?
       desc "list", "List all gems in the bundle"
       method_option "name-only", :type => :boolean, :banner => "print only the gem names"
+      method_option "paths", :type => :boolean, :banner => "print the path to each gem in the bundle"
```
segiddins: what happens if you pass `bundle list  --paths --name-only`?

spec/commands/list_spec.rb
```diff
@@ -15,6 +15,21 @@
     end
   end
 
+  context "with paths option" do
+    before do
+      install_gemfile <<-G
+        source "file://#{gem_repo1}"
```
segiddins: let's also include a `:git` gem and a `gemspec` gem for good measure?

lib/bundler/cli.rb
colby-swandale: Bundler will just print the same output as `bundler list --name-only`

segiddins: we probably should make it an error if both are passed, then

colby-swandale:
ping @indirect

indirect:
Yep, this looks right! Let’s do it.

@bundlerbot r+

PR #6183, opened 2017-11-22
Update rubygems 2.7 branch to 2.7.2

biow0lf:

Update rubygems 2.7 branch to 2.7.2. This should trigger travis build to check that everything works. I think, build should be break.


segiddins:
Why'd you close this? We should merge it

Issue #5879, opened 2017-07-20
Update documentation around checking in the Gemfile.lock


indirect:
When Bundler first shipped, we gitignored the Gemfile.lock inside gems, following the advice given in [this blog post](http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/).

Over time, however, it became clear that this practice forces the pain of broken dependencies onto new contributors, while leaving existing contributors potentially unaware of the problem. As a result, we have revised our guidance for gem authors to now [recommend checking in the lock for gems](http://bundler.io/v1.15/guides/creating_gem.html#testing-our-gem). We have also [removed Gemfile.lock from the gitignore file generated in new gems](https://github.com/bundler/bundler/pull/5822#issuecomment-316530749).

This is a tracking ticket for any further documentation changes needed to support gem developers in this new pattern of checking in `Gemfile.lock` to support new contributors, while continuing to run at least one CI build against the newest version of all dependencies, resolved from scratch at the time of the CI run.

arbonap:
I'd be happy to help out with documentation changes reflecting the new recommendation to check in the lock for gems.

colby-swandale:
This looks like it's been completed. Closing.

mvz:
@colby-swandale I think this part has not yet been addressed:
> while continuing to run at least one CI build against the newest version of all dependencies, resolved from scratch at the time of the CI run.

colby-swandale:
@mvz You're right, thanks for pointing that out.

rubymorillo:
@indirect @colby-swandale I'm happy to take this one on. I need help clarifying this, though: where do we want this:

> while continuing to run at least one CI build against the newest version of all dependencies, resolved from scratch at the time of the CI run.

 to be called out? In the `bundle-install.ronn` man page?

segiddins:
Maybe in https://bundler.io/v1.16/guides/creating_gem.html ?

rubymorillo:
@segiddins possibly. It's a how-to guide, though, and doesn't already make mention of CI. Here are the high-level sections:

- Getting Started
- Testing our gem
- Using other gems
- Testing a CLI
- Crafting a CLI
- Testing a generator
- Writing a generator
- Releasing the gem

At which point in that outline would it make most sense to call out CI builds? cc @indirect

segiddins:
Probably within testing our gem?

ianfixes:
[...]

PR #6184, opened 2017-11-22
Update docs to reflect revised guidance to check in locks for gems

arbonap:

The problem was...
- Bundler stopped gitignoring gem locks. The change was merged in but we need to reflect that change in documentation.

My diagnosis was...
- To add in documentation explaining why Bunder now does _not_ gitignore gem locks.

My fix...
- Update the `gemfile.lock` section of `bundle install` man pages as well as the `gemfile` man page.

I chose this fix because...
- This addresses open issue https://github.com/bundler/bundler/issues/5879

### DIFF
--- a/man/bundle-install.ronn
+++ b/man/bundle-install.ronn
@@ -295,11 +295,19 @@ of a dependency of a gem in your Gemfile(5)) can result in radically
 different gems being needed to satisfy all dependencies.
 
 As a result, you `SHOULD` check your `Gemfile.lock` into version
-control. If you do not, every machine that checks out your
-repository (including your production server) will resolve all
+control, in both applications and gems. If you do not, every machine that
+checks out your repository (including your production server) will resolve all
 dependencies again, which will result in different versions of
 third-party code being used if `any` of the gems in the Gemfile(5)
 or any of their dependencies have been updated.
+
+When Bundler first shipped, the `Gemfile.lock` was gitignored inside gems.
+Over time, however, it became clear that this practice forces the pain of
+broken dependencies onto new contributors, while leaving existing contributors
+potentially unaware of the problem. Since `bundle install` is usually the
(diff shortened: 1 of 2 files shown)


man/bundle-install.ronn
```diff
@@ -295,12 +295,20 @@ of a dependency of a gem in your Gemfile(5)) can result in radically
...
 As a result, you `SHOULD` check your `Gemfile.lock` into version
-control. If you do not, every machine that checks out your
-repository (including your production server) will resolve all
+control, in both applications and gems. If you do not, every machine that
+checks out your repository (including your production server) will resolve all
 dependencies again, which will result in different versions of
 third-party code being used if `any` of the gems in the Gemfile(5)
 or any of their dependencies have been updated.
 
+When Bundler first shipped, the `Gemfile.lock` was gitignored inside gems.
```
segiddins: was `gitignored`

man/gemfile.5.ronn
```diff
@@ -504,3 +504,13 @@ bundler uses the following priority order:
      `rubygems.org`
   3. The sources specified via global `source` lines, searching each source in
      your `Gemfile` from last added to first added.
+
+## Should I check my `Gemfile.lock` into version control?
```
colby-swandale: This feels like this is missing a bit of context. We go from talking about the Gemfile.lock in your application but then jump into the Gemfile.lock inside a gem which is not imminently obvious.

How about we change the title to something like:

> Should I check my `Gemfile.lock` into version control inside my Gem?

?

segiddins: or maybe `for my gem?`

colby-swandale: ping @arbonap

indirect:
@bundlerbot r+

Issue #6181, opened 2017-11-21
Infinite "Resolving dependencies" using 1.16 with fastlane and atlassian-stash


jshier:
Bundler 1.16 spins forever in the "Resolving dependencies" step, outputting periods, with `Gemfile` containing only `fastlane` and `atlassian-stash` gems. 100% reproducible.

## Environment

```
Bundler             1.16.0
  Platforms         ruby, x86_64-darwin-17
Ruby                2.4.2p198 (2017-09-14 revision 59899) [x86_64-darwin17]
  Full Path         /Users/jshier/.rvm/rubies/ruby-2.4.2/bin/ruby
  Config Dir        /Users/jshier/.rvm/rubies/ruby-2.4.2/etc
RubyGems            2.6.14
  Gem Home          /Users/jshier/.rvm/gems/ruby-2.4.2@resolutiontest
  Gem Path          /Users/jshier/.rvm/gems/ruby-2.4.2@resolutiontest:/Users/jshier/.rvm/gems/ruby-2.4.2@global
  User Path         /Users/jshier/.gem/ruby/2.4.0
  Bin Dir           /Users/jshier/.rvm/gems/ruby-2.4.2@resolutiontest/bin
Tools               
  Git               2.13.6 (Apple Git-96)
  RVM               1.29.3 (latest)
  rbenv             not installed
  chruby            not installed
  rubygems-bundler  (1.4.4)
```

## Bundler Build Metadata

```
Built At          2017-10-31
Git SHA           10f20fa33
Released Version  true
```

### Gemfile

```ruby
source "https://rubygems.org"

gem "fastlane"
gem "atlassian-stash"
```

### Gemfile.lock

```
<No /Users/jshier/Desktop/ResolutionTest/Gemfile.lock found>
```

arbonap:
Interesting...I also got the same thing to happen as @jshier when I didn't tie any of the gems to a version with Bundler 1.16. I let Bundler 1.16 try to resolve dependencies for at least a good 10 minutes.

@jshier not sure if this is useful to you, but when I tied the fastlane gem to the latest version, ie like `gem "fastlane", "= 2.66.2"` my dependencies resolved quickly with Bundler 1.16.

colby-swandale:
@arbonap is correct, Bundler is trying to resolve a very wide set of possible solutions. Put a version constraint on both gems in the Gemfile and Bundler should resolve a lot quicker.

jshier:
[...]

PR #6186, opened 2017-11-23
Resolver: treat dependencies with prerelease versions as slightly constrained

greysteil:
Fixes https://github.com/bundler/bundler/issues/6181.

In Bundler 1.15.4, the way we checked for pre-releases meant that dependencies with a lot of them sorted as more constrained than those with few/none. When we moved to our updated resolution strategy in 1.16.0 we accidentally lost that behaviour.

I'm not wild about this PR, but it fixes the resolver regression. It's unsatisfying because it would be nice to believe the resolver will always resolve performantly, regardless of the sort order thrown at it, but clearly that isn't the case.

Conceptually, this can be justified as "when two resolutions are possible, we prefer the one that does not require one dependency to be at a pre-release version".

If others are happy with this I can write a test for it.

### DIFF
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -249,7 +249,8 @@ module Bundler
           if all <= 1
             all - 1_000_000
           else
-            search = search_for(dependency).size
+            search = search_for(dependency)
+            search = @prerelease_specified[dependency.name] ? search.count : search.count {|s| !s.version.prerelease? }
             search - all
           end
         end
--- a/spec/resolver/basic_spec.rb
+++ b/spec/resolver/basic_spec.rb
@@ -42,6 +42,13 @@ RSpec.describe "Resolving" do
     should_resolve_as %w[a-1.0.0 b-2.0.0 c-1.0.0 d-1.0.0]
   end
 
+  it "prefers non-prerelease resolutions in sort order" do
+    @index = optional_prereleases_index
+    dep "a"
+    dep "b"
+    should_resolve_as %w[a-1.0.0 b-1.5.0]
+  end
+
   it "resolves a index with root level conflict on child" do
     @index = a_index_with_root_conflict_on_child
     dep "i18n", "~> 0.4"
--- a/spec/support/indexes.rb
+++ b/spec/support/indexes.rb
@@ -401,5 +401,20 @@ module Spec
         gem("d", %w[1.0.0 2.0.0])
       end
     end
+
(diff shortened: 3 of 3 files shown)


lib/bundler/resolver.rb
```diff
@@ -249,7 +249,10 @@ def amount_constrained(dependency)
           if all <= 1
             all - 1_000_000
           else
-            search = search_for(dependency).size
+            search = search_for(dependency)
+            unless @prerelease_specified[dependency.name]
+              search = search.reject { |s| s.version.prerelease? }.size
```
segiddins: This can use .count

indirect:
Sorting the resolver input has historically been one of the biggest ways that we're able to "cheat" on the whole NP complete dependency graph problem. This is a good catch, and I suspect that we're always going to want to do some amount of this as a way to force resolutions that human ruby developers are more likely to want to the front. Thanks for the patch!

@segiddins do you have any suggestions on how to build a test for this, so that we can get it merged?

segiddins:
Honestly a test in Molinillo using a new index fixture would be fine with me

greysteil:
Test added - apologies for being a little slow on that!

greysteil:
I'll take a look at a Molinillo spec, too, but that doesn't need to block this - the `sort_dependencies` logic is Bundler specific, not generic to Molinillo. (It may also be tricky, too, as from memory we don't have much in the way of pre-release handling in the Molinillo test indexes.)

lib/bundler/resolver.rb
```diff
@@ -249,8 +249,11 @@ def amount_constrained(dependency)
           if all <= 1
             all - 1_000_000
           else
-            search = search_for(dependency).size
-            search - all
+            search = search_for(dependency)
```
segiddins: This should help avoid intermediary array allocations?

```ruby
search = search_for(dependency)
search = @prerelease_specified[dependency.name] ? search.count : search.count { !s.version.prerelease? }
search - all
``

segiddins:
> the sort_dependencies logic is Bundler specific, not generic to Molinillo

correct, but don't forget we https://github.com/CocoaPods/Molinillo/blob/beb24a0789aa67f9adef47991d3f19f72f62c4a8/spec/spec_helper/index.rb#L91-L102

greysteil:
Molinillo spec added - this should be good to go now 🎉

indirect:
@greysteil 🎉

@bundlerbot r+

Issue #6072, opened 2017-10-01
Problem with inline gemfiles in 1.16.0.pre2


deivid-rodriguez:
##### test.rb 

```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "thor"
end
```

##### With bundler 1.15.4

```console
$ ruby test.rb 
Fetching gem metadata from https://rubygems.org/.
Fetching version metadata from https://rubygems.org/
Resolving dependencies...
Using bundler 1.15.4
Using thor 0.20.0 (was 0.19.4)
```

##### With bundler 1.16.0.pre.2

```console
$ ruby test.rb
Fetching gem metadata from https://rubygems.org/.
Fetching gem metadata from https://rubygems.org/.......
Fetching gem metadata from https://rubygems.org/.....
Fetching gem metadata from https://rubygems.org/...
Resolving dependencies...
Using bundler 1.16.0.pre.2
Using thor 0.20.0 (was 0.19.4)
/home/deivid/.gem/ruby/2.4.2/gems/bundler-1.16.0.pre.2/lib/bundler/runtime.rb:30:in `block in setup': thor-0.20.0 is missing. Run `bundle install` to get it. (Bundler::GemNotFound)
	from /home/deivid/.rubies/ruby-2.4.2/lib/ruby/2.4.0/forwardable.rb:229:in `each'
	from /home/deivid/.rubies/ruby-2.4.2/lib/ruby/2.4.0/forwardable.rb:229:in `each'
	from /home/deivid/.gem/ruby/2.4.2/gems/bundler-1.16.0.pre.2/lib/bundler/runtime.rb:28:in `map'
	from /home/deivid/.gem/ruby/2.4.2/gems/bundler-1.16.0.pre.2/lib/bundler/runtime.rb:28:in `setup'
	from /home/deivid/.gem/ruby/2.4.2/gems/bundler-1.16.0.pre.2/lib/bundler/inline.rb:70:in `gemfile'
	from test.rb:3:in `<main>'
```

olleolleolle:
I tried again, with v1.16.0.pre.3:

```
$ ruby test.rb
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies...
Using bundler 1.16.0.pre.3
Using thor 0.20.0
```

Environment:

```
$ ruby -v
ruby 2.4.0p0 (2016-12-24 revision 57164) [x86_64-darwin15]
$ bundle -v
[...]

PR #6188, opened 2017-11-27
Stop overriding specs during double-checks

indirect:
This fixes my test-case reproduction of #6072. I’m pretty sure it’s because
the double-check was overriding the locally installed index specs with
specs fetched from the RubyGems API, causing Bundler to conclude that the
gems aren’t installed, even though they are.

@deivid-rodriguez @y-yagi can you verify that this patch fixes the issues with `inline` that you were seeing?


y-yagi:
@indirect Thank you for fixing! I confirmed that this patch is solved my issue.

spec/runtime/inline_spec.rb
```diff
@@ -267,4 +263,29 @@ def confirm(msg, newline = nil)
     expect(last_command).to be_success
     expect(last_command.stdout).to eq "1.0.0"
   end
+
+  describe "when path gems depend on API server gems" do
```
segiddins: this test doesn't seem to use `inline` at all?

spec/runtime/inline_spec.rb
```diff
@@ -267,4 +263,29 @@ def confirm(msg, newline = nil)
...
+    let(:source_hostname) { "localgemserver.test" }
+    let(:source_uri) { "http://#{source_hostname}" }
+
+    before do
+      build_lib "needs_rake", "1.0.0" do |s|
```
segiddins: `needs_rack`?

spec/runtime/inline_spec.rb
```diff
@@ -267,4 +263,29 @@ def confirm(msg, newline = nil)
...
+        s.add_dependency "rack", "= 1.0.0"
+      end
+
+      gemfile <<-G
+        source "#{source_uri}"
+        gem "needs_rake", path: "#{lib_path("needs_rake-1.0.0")}"
```
segiddins: hash rockets

indirect:
@segiddins addressed your feedback, the test now does what I thought would trigger the bug... but the test passes. Any ideas about how to set up the right kind of double-check that will overwrite the index entry for the installed gem with an index entry for a gem from the API server?

segiddins:
Are you saying the test doesn’t fail without your patch? Also, rubocop.

indirect:
@segiddins fixed, and yes the test passes even without my patch. :(

indirect:
Some progress... managed to reduce a very small repro case that still fails when run directly via `ruby`, over here: https://gist.github.com/indirect/e84b36fc854409916a569e53e2756deb

indirect:
OKAY this is a working test. At least, it's red for me locally before the patch in this PR, and green for me after the patch in this PR. @segiddins r?

segiddins:
This is fine, but I'd really like to get rid of the worldworld spec eventually

indirect:
@segiddins I completely agree, but I think we need a better simulator of a source as complex as production rubygems to be able to do that. And I don’t want that to be a blocker for this bugfix. 👍🏻

indirect:
@bundlerbot r+

hsbt:
Can anyone retry this?

segiddins:
@bundlerbot retry

Issue #6185, opened 2017-11-23
Test suite does not work in path with dash


voxik:
Test suite does not work in patch such as ```rubygem-bundler```, e.g. the path with dash in it. This is apparently due to d967ea961a1c269751344f00c75956892eb71a35. I believe that there is nothing magical about "-" character and I'd be surprised if any commonly used system has issues handling dash in path. So I wonder what are the "special characters" you are trying to protect against?

@hmistry was talking about "white spaces" (and asking you to mention this somewhere, but this does not happened :/ ), but you are matching for "\w", which stands for "word character" (typically matches matches the ASCII characters [A-Za-z0-9_] and probably more unicode characters).

Why don't you match the white space characters, i.e. \s (which matches [ \t\r\n\f]), as was the original request?

segiddins:
Because at the moment the tests don't all pass when those characters are present

voxik:
Could you please be more specific about "those characters"? Does the set include dash? Let me try that ...

segiddins:
I believe so. And since none of the maintainers keep bundler in a path with non-ascii characters, it's hard for us to guarantee things will work, which would make for a bad experience for someone trying to contribute for the first time

voxik:
You are right, there is one test failure when dash is included in path:

~~~

  1) bundle update --bundler updates the bundler version in the lockfile without re-resolving
     Failure/Error: bundle! :update, :bundler => true, :verbose => true
     
     RuntimeError:
       Invoking bundle!(:update, {:bundler=>true, :verbose=>true, "no-color"=>true}) failed:
       Running `bundle update --bundler "> 0.a" --no-color --verbose` with bundler 1.16.0
       Found changes from the lockfile, re-resolving dependencies because the list of sources changed
       Fetching source index from file:/builddir/build/BUILD/rubygem-bundler-1.16.0/usr/share/gems/gems/bundler-1.16.0/tmp/gems/remote4/
       Gem::RemoteFetcher::FetchError: Errno::ENOENT: No such file or directory @ rb_sysopen - /builddir/build/BUILD/rubygem-bundler-1.16.0/usr/share/gems/gems/bundler-1.16.0/tmp/gems/remote4/specs.4.8.gz (file:/builddir/build/BUILD/rubygem-bundler-1.16.0/usr/share/gems/gems/bundler-1.16.0/tmp/gems/remote4/specs.4.8.gz)
[...]

PR #6194, opened 2017-11-29
Be more precise when replacing Bundler version in lock file.

voxik:
Lets say the current Bundler version is 1.16.0. If the test suite is, by a chance, executed in directory which contains such version string, this path is stored in Gemfile.lock file. Later, the test tries to replace
the version of Bundler in the lock file, but instead, it replaces the version in path. Be more careful what should be actually replaced.

This is related to #6185 although it does not resolve the original concern.

### DIFF
--- a/spec/commands/update_spec.rb
+++ b/spec/commands/update_spec.rb
@@ -657,7 +657,7 @@ RSpec.describe "bundle update --bundler" do
       source "file:#{gem_repo4}"
       gem "rack"
     G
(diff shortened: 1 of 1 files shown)


segiddins:
This seems reasonable, but the RuboCop failure needs to be looked at. https://travis-ci.org/bundler/bundler/jobs/309074361#L674

voxik:
Fixed version pushed.

segiddins:
Thanks!
@bundlerbot r+

Issue #6165, opened 2017-11-10
Unable to detect bundler spec (1.16.0)


mleopard:
I keep running into this error when using bundler 1.16.0:
```
LoadError: no such file to load -- /home/dev/.rvm/rubies/jruby-1.7.4/lib/ruby/gems/shared/gems/bundler-1.16.0/exe/bundle
    load at org/jruby/RubyKernel.java:1073
  (root) at /home/dev/.rvm/gems/jruby-1.7.4/bin/bundle:23
```

I have also seen this error on another VM using a newer version of jruby:
```
Unable to detect bundler spec under '/home/dev/.rvm/rubies/jruby-1.7.27/lib/ruby/gems/shared/gems/bundler-1.16.0'' and its sub-dirs
```

And I noticed the directory structure between 1.15.4 and 1.16.0 are a bit different.

Bundler 1.15.4
├── bin
├── bundler.gemspec
├── CHANGELOG.md
├── CODE_OF_CONDUCT.md
├── CONTRIBUTING.md
├── doc
├── exe
├── lib
├── LICENSE.md
├── man
├── Rakefile
├── README.md
└── task

Bundler 1.16.0
├── CHANGELOG.md
├── exe
├── lib
├── LICENSE.md
├── man
└── README.md

If I explicitly set bundler to 1.15.4, my project will build without errors.

Is the different directory structure expected behavior?

colby-swandale:
There was some changes to the files that are included in `bundler.gem` in #5862 but the `bin` directory nor the gemspec itself has been moved. Can you try reinstalling the bundler gem?

mleopard:
Reinstalling the bundler gem didn't make a difference. I'm noticing an issue with how my gems are being installed, so perhaps this might be due to how I set up gem on my machine and not bundler 1.16. I tried reinstalling both 1.15.4 and 1.16 and the structure is still the same where 1.16 is missing a lot of files.

colby-swandale:
ping @mleopard were you able to solve this problem?

dr-itz:
Same problem here, but with warbler on JRuby 9.1.x. 

I manually downloaded the .gem from rubygems.org, manually extracted it to check. `bundler.gemspec` is missing. Also, https://github.com/bundler/bundler/pull/5862#issuecomment-319521002 lists the file as gone...

See also jruby/warbler#421

PR in a bit

colby-swandale:
Thanks for the PR! I'm closing this now.

PR #6195, opened 2017-11-29
Include bundler.gemspec in the gem because warbler breaks w/o it

dr-itz:
f06405 caused a regression with warbler, described in #6165 and jruby/warbler#421

If the fix is ok, it should probably be cherry-picked to 1-16-stable. 

Upgrading from bundler v1.15.x to v1.16.0 causes warbler to break/abort with error.

1. error message from warbler: missing file
2. see why file is missing
3. google to see if it's just me 😄 

Put the missing (from warblers point of view at least) back.

Seems easier than to fix warbler.

### DIFF
--- a/bundler.gemspec
+++ b/bundler.gemspec
@@ -49,6 +49,8 @@ Gem::Specification.new do |s|
   s.files += Dir.glob("man/**/*")
   # Include the CHANGELOG.md, LICENSE.md, README.md manually
   s.files += %w[CHANGELOG.md LICENSE.md README.md]
+  # include the gemspec itself because warbler breaks w/o it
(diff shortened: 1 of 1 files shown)


segiddins:
What is warbler? It shouldn't be necessary to ship the gemspec, as when the gem is installed the real gemspec that RubyGems uses is installed separately.

hsbt:
Warbler is one of central packaging tool of JRuby. It needs to gemspec file for packaging to war file. 

1.15 ships with gemspec file, but 1.16 is not. It seems to breaking changes. I hope to merge this pull request and help JRuby users.

segiddins:
@bundlerbot r+

PR #6202, opened 2017-12-04
Show warning message about binstub outside generation.

### LINKED ISSUE #6149: 1.16.0 is intercepting binstub arguments and throwing 'unknown switches' error
I've just created a new Rails project with bundler 1.16.0, and when trying to invoke the webpack binstub with command line arguments for the underlying executable, Bundler is throwing the message 'unknown switches':

```
$ ./bin/webpack --colors --progress
Unknown switches '--colors, --progress'
```

This did not / does not happen with 1.15.x Bundler. 

I would expect that binstub command line arguments should still be proxied through to the underlying executable? I can find no documentation or issues that would suggest this chnge in behaviour was wanted.

hsbt:
/cc @indirect 

Fixes #6149.

Originated from https://github.com/bundler/bundler/issues/6149#issuecomment-347783823


lib/bundler/templates/Executable
```diff
@@ -8,8 +8,15 @@
 # this file is here to facilitate running it.
 #
 
+def bundle_stub?(path)
```
segiddins: defining the method like this will leak it to the global scope. Can you also add a test for that this is fixing? Thanks!

hsbt:
Added test example.

@segiddins Can you review it?

hsbt:
@bundlerbot r+

PR #6203, opened 2017-12-07
Fixed failing examples with Ruby 2.5.

hsbt:
(It targets 1-16-stable branch. I will create another pull request for master)

It's introduced by `URI::Generic` changes.

see  https://github.com/ruby/ruby/commit/ed48bfa5e8770a345424abd7f24f94ea9bbf5973

I added a hostname to examples that expect `file:` string in result messages.


hsbt:
@bundlerbot r+

PR #6204, opened 2017-12-08
Fix broken status on Travis CI

hsbt:
In Dec 2017, 1-16-stable branch of Travis was broken caused outside condition. I try to fix it.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -101,6 +101,8 @@ RSpec.configure do |config|
 
   original_wd  = Dir.pwd
   original_env = ENV.to_hash.delete_if {|k, _v| k.start_with?(Bundler::EnvironmentPreserver::BUNDLER_PREFIX) }
+  original_default_specs = Dir[File.join(Gem.default_dir, "specifications", "default", "bundler*")]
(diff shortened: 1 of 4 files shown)


spec/spec_helper.rb
```diff
@@ -101,13 +101,19 @@
...
   original_env = ENV.to_hash.delete_if {|k, _v| k.start_with?(Bundler::EnvironmentPreserver::BUNDLER_PREFIX) }
+  original_default_specs = Dir[File.join(Gem.default_dir, "specifications", "default", "bundler*")]
 
   config.expect_with :rspec do |c|
     c.syntax = :expect
   end
 
   config.before :all do
     build_repo1
+    original_default_specs.each {|s| FileUtils.mv(s, s + ".org") }
```
segiddins: why do we need this?

hsbt: some examples that use fake bundle commands are failing status when there is bundler as default gems.

see https://travis-ci.org/bundler/bundler/jobs/313289214

hsbt:
I'm going to `git squash` before merging working commits. These changes is that bundler installation as default gems move and restore during rspec invoking.

@segiddins Do you have another comment? I hope to finish to prepare bundler-1.16.1 releasing until Dec 12.

hsbt:
@bundlerbot r+

hsbt:
@bundlerbot retry

PR #6207, opened 2017-12-09
Let users generate gemfiles inside paths whos parent contains a Gemfile

### LINKED ISSUE #6205: Bundle init failed when ancestor directory has Gemfile
# Steps to reproduce

```
% tree sample
sample
├── Gemfile
└── hoge
% cd sample/hoge
% bundle init
/Users/kyoshida/work/sample/Gemfile already exists at /Users/kyoshida/work/sample/Gemfile
```

# Expected behavior

Gemfile is generated in `hoge` directory.

Because [bundle init document](http://bundler.io/v1.16/man/bundle-init.1.html) says the following.

> bundle-init - Generates a Gemfile into the current working directory

I believe that current woking is the same as `$PWD`'s value.

However, the code is intentionally looking for ancestor directories.
1. https://github.com/bundler/bundler/blob/v1.16.0/lib/bundler/cli/init.rb#L40
1. https://github.com/bundler/bundler/blob/v1.16.0/lib/bundler/shared_helpers.rb#L240

This behavior is in the specifications?

# Version

Bundler version 1.16.0

colby-swandale:

Users are unable to generate a new Gemfile with `bundle init` if there is a Gemfile in any
parent folder.

Closes #6205 

Don't use `SharedHelpers.default_gemfile` as that searches up the paths for a Gemfile. Instead only check the current directory.

### DIFF
--- a/lib/bundler/cli/init.rb
+++ b/lib/bundler/cli/init.rb
@@ -36,11 +36,7 @@ module Bundler
   private
 
     def gemfile
-      @gemfile ||= begin
-        Bundler.default_gemfile
-      rescue GemfileNotFound
-        Bundler.feature_flag.init_gems_rb? ? "gems.rb" : "Gemfile"
-      end
+      @gemfile ||= Bundler.feature_flag.init_gems_rb? ? "gems.rb" : "Gemfile"
     end
   end
 end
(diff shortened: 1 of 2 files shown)


segiddins:
👍🏻 once the 1.x build is fixed

segiddins:
@bundlerbot r+

hsbt:
@bundlerbot retry

PR #6209, opened 2017-12-11
Ignore to generate documentation of templates.

hsbt:

When users invoke `rdoc -o rdoc lib`, it generates documentation contains template files under the `lib/bundler/template` like `rdoc/lib/bundler/templates/newgem/bin/setup_tt.html`

These files are not useful for users.

I added `.document` file to its directory.

### DIFF
--- a/lib/bundler/templates/.document
+++ b/lib/bundler/templates/.document
@@ -1,0 +1,1 @@
+# Ignore all files in this directory


colby-swandale:
@bundlerbot r+

hsbt:
@colby-swandale Thanks!

PR #6211, opened 2017-12-11
Add explicit source when available in the error msg upon failure

papanikge:

My bundling failed and the command that `bundler` output was wrong-ish, in the sense that my gem was suppose to come from a different source.

The error message was not explicit.

Bundler now explicitly print the source when there's just one. I think it's a good case just to print it when there's one. If there are more it's ambiguous, so we fallback to the old way. Maybe we could have a warning there that the command might not be exactly what the user wants.

See previous answer. I didn't find any easier ways of getting the source url.
Let me know if there are better implementations :)

### DIFF
--- a/lib/bundler/installer/gem_installer.rb
+++ b/lib/bundler/installer/gem_installer.rb
@@ -44,7 +44,12 @@ module Bundler
     end
 
     def gem_install_message
-      "Make sure that `gem install #{spec.name} -v '#{spec.version}'` succeeds before bundling."
+      remotes = spec.source.remotes
+      if remotes.size == 1
+        "Make sure that `gem install #{spec.name} -v '#{spec.version}' --source '#{remotes.first}'` succeeds before bundling."
+      else
+        "Make sure that `gem install #{spec.name} -v '#{spec.version}'` succeeds before bundling."
+      end
     end
 
     def spec_settings
--- a/spec/commands/install_spec.rb
+++ b/spec/commands/install_spec.rb
@@ -348,14 +348,14 @@ RSpec.describe "bundle install with gem sources" do
       end
 
       install_gemfile <<-G, :full_index => true
-        source "file://#{gem_repo2}"
+        source "file:\/\/localhost#{gem_repo2}"
 
         gem "ajp-rails", "0.0.0"
       G
 
       expect(last_command.stdboth).not_to match(/Error Report/i)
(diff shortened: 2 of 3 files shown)


colby-swandale:
Thanks for opening a pull request. This will require a test before we consider merging.

colby-swandale:
ping @papanikge

papanikge:
sorry I didn't respond earlier. I will add a test.

papanikge:
Hello, @colby-swandale et al,

I thought about it. I could stub `generate_executable_stubsz` to raise an error and test that the message is what it should be in case of available remote source, but testing a string sounds way too much.
I don't know what else I could test. Any hints/ideas?

segiddins:
See the specs in `spec/install/failure_spec.rb`, and copy like the first one, just adding an additional source to the gemfile?

papanikge:
thanks for the tip @segiddins 
I also fixed a broken spec I didn't see last time, and rebased over master...
take a look, if you will :)

papanikge:
the broken test occurs because of the way those tests are written methinks. I could remove the extra spec and alter the message in the first to include the `--source [...]` part.

segiddins:
> the broken test occurs because of the way those tests are written methinks

how so?
it just looks like maybe the expected strings need to be updated

papanikge:
@segiddins should I keep both then?

segiddins:
If they're testing different things (i.e. have different setup or different assertions), then sure!

papanikge:
It's now only broken for 2.5.0, which as far as I know is still under development.
Should I investigate?

colby-swandale:
Yes, Ruby 2.5.0 was released to the general public nearly a month ago.

papanikge:
ok, I had an old `rbenv` version. sorry about that 😁

colby-swandale:
no worries 👍

papanikge:
fixed the errors, rebased and squashed :)

papanikge:
`spec/commands/exec_spec.rb:759` passes locally on my system :(

colby-swandale:
#6275 should fix this issue.

papanikge:
so I guess I should wait until that is merged, and rebase over it?

segiddins:
@papanikge no need to rebase, we can `:+1` it via our mergebot once that PR lands!

segiddins:
@bundlerbot r+

PR #6213, opened 2017-12-13
Unregister the bundler default spec in the tests

segiddins:

The problem was the specs would fail when the version did not match that installed as a default gem.

My diagnosis was we needed to unregister the default spec when running tests.

I chose this fix because it avoids actually touching any of the installed files, meaning we're not relying upon the tests exiting cleanly & being able to restore them.

### DIFF
--- a/spec/commands/exec_spec.rb
+++ b/spec/commands/exec_spec.rb
@@ -29,7 +29,7 @@ RSpec.describe "bundle exec" do
       gem "rack"
     G
 
-    bundle "exec 'cd #{tmp("gems")} && rackup'"
+    bundle "exec 'cd #{tmp("gems")} && rackup'", :env => { :RUBYOPT => "-r#{spec_dir.join("support/hax")}" }
 
     expect(out).to include("1.0.0")
   end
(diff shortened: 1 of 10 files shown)


segiddins:
Why is this so hard 😭

hsbt:
I could resolve some failing examples with replacing `Gem.ruby` to `Gem.ruby -r#{Spec::Path.spec_dir}/support/hax.rb`. But some examples still failing...

segiddins:
For the 1.8.7 issue, I'm wondering if it has to do with `require` being sensitive to being giving a short path (resolved against `$LOAD_PATH`) versus an absolute path. For the 1.9.3 ones, I think it's because we're calling `bundle exec ruby` and need to put `-rsupport/hax` in `$RUBYOPT` instead of just passed being the command line

segiddins:
Just pushed up another commit, I hope it fixes things :/

hsbt:
I couldn't reproduce this fail with my local machine: https://travis-ci.org/bundler/bundler/jobs/317472750

It seems an issue of travis environment.

hsbt:
I also could not reproduce test fail of `env RGV=v2.6.8 BUNDLER_SPEC_SUB_VERSION=2.0.0 bin/rake spec` with Ruby 2.4. It's very strange...

hsbt:
https://github.com/bundler/bundler/pull/6217 was inherited this branch. and fix its problem.

PR #6217, opened 2017-12-19
Fix test failure with new bundler version and default gems of bundler

hsbt:
This pull-request and branch are only investigation for Travis.

PR #6218, opened 2017-12-19
Workaround for rubygems&bundler release

hsbt:

The current 1-16-stable branch is the complex status for Travis environment. So We hard to fix it. Because it has two rubygems versions(RGV=vx.y.z and rvm installation version) and two bundler versions(bundler-1.16.0 as default gems and test target version).

Above situation caused that Rubygems bundles bundler-1.16.0. So, We can resolve it status with Rubygems 2.7.4 and Bundler-1.16.1 as default gems installed by rubygems installer.

@segiddins How about this?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -110,9 +110,17 @@ matrix:
   allow_failures:
     # Remove 1.8.7-2.1.10 matrix after merging bundler-1.16.1 into rubygems-2.7.x
     - rvm: 2.1.10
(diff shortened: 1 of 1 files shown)


hsbt:
I specified to version of RGV. it seems only RGV 2.1.11 and 2.2.5 problems.

hsbt:
This pull request is unnecessary. See. https://github.com/bundler/bundler/pull/6217

Issue #6150, opened 2017-11-04
weird bundle exec - nohup behavior where SIGHUP gets raised to child process


oneamtu:
At the company I work for we ran into this weird problem when running processes through `nohup` and `bundle exec`. `nohup` is a utility that is supposed to trap SIGHUPs from bubbling to a child process, thus allowing a user to daemonize a command by running it with `nohup` and then closing the parent shell. 

For whatever reason, with more recent versions of bundler past 1.12.5 SIGHUPs bubble up to the child process. 1.12.5 works great. We also found a workaround with more recent versions of bundler where the ordering of the commands changes the behavior - `nohup bundle exec` fails and `bundle exec nohup` works. Here's an example of the test which I uploaded here - https://github.com/oneamtu/bundler_nohup_test:

```
[19:56:02] oneamtu:test $ nohup rake loop_de_loop >> loop.log 2>&1 &
[1] 7553
[19:56:27] oneamtu:test $ kill -hup 7553
[19:56:34] oneamtu:test $ kill -kill 7553
[1]  + 7553 killed     nohup rake loop_de_loop >> loop.log 2>&1
[19:56:44] oneamtu:test $ nohup bundle exec rake loop_de_loop >> loop.log 2>&1 &
[1] 7657
[19:56:52] oneamtu:test $ kill -hup 7657
[1]  + 7657 exit 1     nohup bundle exec rake loop_de_loop >> loop.log 2>&1
[19:56:59] oneamtu:test $ bundle exec nohup rake loop_de_loop >> loop.log 2>&1 &
[1] 7746
[19:57:12] oneamtu:test $ kill -hup 7746
[19:57:17] oneamtu:test $ kill -kill 7746
[1]  + 7746 killed     bundle exec nohup rake loop_de_loop >> loop.log 2>&1

[6:09:53] oneamtu:test $ nohup --version
nohup (GNU coreutils) 8.25
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Jim Meyering.
```

## Environment

```
Bundler       1.16.0
  Platforms   ruby, x86_64-linux
Ruby          2.4.2p198 (2017-09-14 revision 59899) [x86_64-linux]
  Full Path   /home/oneamtu/.asdf/installs/ruby/2.4.2/bin/ruby
[...]

PR #6223, opened 2017-12-23
Only trap INT signal and set to DEFAULT

shayonj:

The problem was commands like `nohup bundler exec {program}` wouldn't work as intended. For  example, if a `HUP` signal were to be sent to the process running the `bundle exec ..`, it should in theory not terminate. Because, `nohup` would `IGNORE` that signal. But, that is not what the case is case is currently.

My diagnosis was, if a process/bundler execution already had a `SIGNAL` set to it, example `HUP`, then performing `bundle exec {program}`, would mean that bundler resets any prior `SIGNAL`s on that process and sets them to `DEFAULT`.

My fix to the problem is to only trap `SIGNAL`s that we think should be set to `DEFAULT`, in this case, `INT`. 

I chose this fix because its lot less aggressive than setting every signal to `DEFAULT`, and gives us the work with a smaller set of `SIGNAL`s. It also felt cleaner than having to trap a signal first and then restore to its predecessor value.

----

This is a dump that shows the before and after signals, when `nohup bundle exec {program }` gets run.

```
SIGEXIT: Before Handler: (), Current Handler: (DEFAULT)
SIGHUP: Before Handler: (IGNORE), Current Handler: (DEFAULT)
SIGINT: Before Handler: (#<Proc:0x00007f8e100534f8@/Users/<>/.rvm/gems/ruby-2.4.2/gems/bundler-1.16.0/exe/bundle:5>), Current Handler: (DEFAULT)
SIGQUIT: Before Handler: (DEFAULT), Current Handler: (DEFAULT)
SIGTRAP: Before Handler: (SYSTEM_DEFAULT), Current Handler: (DEFAULT)
SIGABRT: Before Handler: (SYSTEM_DEFAULT), Current Handler: (DEFAULT)
SIGIOT: Before Handler: (SYSTEM_DEFAULT), Current Handler: (DEFAULT)
SIGEMT: Before Handler: (SYSTEM_DEFAULT), Current Handler: (DEFAULT)
SIGSYS: Before Handler: (), Current Handler: (DEFAULT)
SIGPIPE: Before Handler: (), Current Handler: (DEFAULT)
SIGALRM: Before Handler: (DEFAULT), Current Handler: (DEFAULT)
SIGTERM: Before Handler: (DEFAULT), Current Handler: (DEFAULT)
SIGURG: Before Handler: (SYSTEM_DEFAULT), Current Handler: (DEFAULT)
SIGTSTP: Before Handler: (SYSTEM_DEFAULT), Current Handler: (DEFAULT)
SIGCONT: Before Handler: (SYSTEM_DEFAULT), Current Handler: (DEFAULT)
SIGCHLD: Before Handler: (SYSTEM_DEFAULT), Current Handler: (DEFAULT)
SIGCLD: Before Handler: (SYSTEM_DEFAULT), Current Handler: (DEFAULT)
SIGTTIN: Before Handler: (SYSTEM_DEFAULT), Current Handler: (DEFAULT)
SIGTTOU: Before Handler: (SYSTEM_DEFAULT), Current Handler: (DEFAULT)
SIGIO: Before Handler: (SYSTEM_DEFAULT), Current Handler: (DEFAULT)
SIGXCPU: Before Handler: (SYSTEM_DEFAULT), Current Handler: (DEFAULT)
SIGXFSZ: Before Handler: (SYSTEM_DEFAULT), Current Handler: (DEFAULT)
SIGPROF: Before Handler: (SYSTEM_DEFAULT), Current Handler: (DEFAULT)
SIGWINCH: Before Handler: (SYSTEM_DEFAULT), Current Handler: (DEFAULT)
SIGUSR1: Before Handler: (DEFAULT), Current Handler: (DEFAULT)
SIGUSR2: Before Handler: (DEFAULT), Current Handler: (DEFAULT)
SIGINFO: Before Handler: (SYSTEM_DEFAULT), Current Handler: (DEFAULT)
```

From this, we can see only `INT` is being trapped by bundler
```
SIGINT: Before Handler: (#<Proc:0x00007f8e100534f8@/Users/<>/.rvm/gems/ruby-2.4.2/gems/bundler-1.16.0/exe/bundle:5>), Current Handler: (DEFAULT)
```

hence, the only one being restored back to `DEFAULT`

----

Issue: https://github.com/bundler/bundler/issues/6150

### DIFF
--- a/lib/bundler/cli/exec.rb
+++ b/lib/bundler/cli/exec.rb
@@ -6,7 +6,7 @@ module Bundler
   class CLI::Exec
     attr_reader :options, :args, :cmd
 
-    RESERVED_SIGNALS = %w[SEGV BUS ILL FPE VTALRM KILL STOP].freeze
+    TRAPPED_SIGNALS = %w[INT].freeze
 
     def initialize(options, args)
       @options = options
@@ -70,8 +70,7 @@ module Bundler
       ui = Bundler.ui
       Bundler.ui = nil
       require "bundler/setup"
-      signals = Signal.list.keys - RESERVED_SIGNALS
-      signals.each {|s| trap(s, "DEFAULT") }
+      TRAPPED_SIGNALS.each {|s| trap(s, "DEFAULT") }
       Kernel.load(file)
     rescue SystemExit, SignalException
       raise
--- a/spec/support/helpers.rb
+++ b/spec/support/helpers.rb
@@ -217,6 +217,14 @@ module Spec
     end
     bang :gem_command
 
+    def test_signals
+      open3_reserved_signals = %w[CHLD CLD PIPE]
+      reserved_signals = %w[SEGV BUS ILL FPE VTALRM KILL STOP EXIT]
+      bundler_signals = %w[INT]
+
+      Signal.list.keys - (bundler_signals + reserved_signals + open3_reserved_signals)
+    end
+
     def sys_exec(cmd)
       command_execution = CommandExecution.new(cmd.to_s, Dir.pwd)
 
--- a/spec/support/helpers.rb
+++ b/spec/support/helpers.rb
@@ -217,14 +217,6 @@ module Spec
     end
     bang :gem_command
 
-    def test_signals
-      open3_reserved_signals = %w[CHLD CLD PIPE]
-      reserved_signals = %w[SEGV BUS ILL FPE VTALRM KILL STOP EXIT]
-      bundler_signals = %w[INT]
-
-      Signal.list.keys - (bundler_signals + reserved_signals + open3_reserved_signals)
-    end
-
     def sys_exec(cmd)
       command_execution = CommandExecution.new(cmd.to_s, Dir.pwd)
 
--- a/spec/commands/exec_spec.rb
+++ b/spec/commands/exec_spec.rb
@@ -726,6 +726,28 @@ __FILE__: #{path.to_s.inspect}
         expect(out).to eq("foo")
       end
     end
+
+    context "signals not being trapped by bunder" do
+      let(:executable) { strip_whitespace <<-RUBY }
+        #{shebang}
+
+        signals = "#{test_signals.join(", ")}".split(", ") # ruby 1.8.7 doesn't preserve array in string interpolation, hence the join/split.
+        result = signals.map do |sig|
+          Signal.trap(sig, "IGNORE")
+        end
+        puts result.select { |ret| ret == "IGNORE" }.count
+      RUBY
+
+      it "makes sure no unexpected signals are restored to DEFAULT" do
+        test_signals.each do |n|
+          Signal.trap(n, "IGNORE")
+        end
+
+        bundle("exec #{path}")
+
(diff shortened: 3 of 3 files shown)


lib/bundler/cli/exec.rb
```diff
@@ -6,7 +6,7 @@ module Bundler
   class CLI::Exec
     attr_reader :options, :args, :cmd
 
-    RESERVED_SIGNALS = %w[SEGV BUS ILL FPE VTALRM KILL STOP].freeze
+    RESTORE_SIGNALS_TO_DEFAULT = %w[INT].freeze
```
segiddins: `TRAPPED_SIGNALS`

segiddins:
Looks good, can you please add a test for this change? Thanks!

shayonj:
@segiddins There is a spec specifically for `INT`

https://github.com/bundler/bundler/blob/d5856a036642ae1f255c3bd301cd6937d1ed87bd/spec/commands/exec_spec.rb#L715

Which otherwise (without trapping the signal) fails. Do you think we should still introduce a new spec?

segiddins:
Yes, we should add a spec that other signals aren't changed back to the default if they're set to ignore (i.e. testing the `nohup` case that led to this issue)

shayonj:
Sounds good! I added spec for all signals that we could test for.

olleolleolle:
@segiddins Has your feedback been met with that test?

oneamtu:
As the end-user who signaled :vertical_traffic_light: this problem, thanks for your patch!

spec/support/helpers.rb
```diff
@@ -217,10 +217,21 @@ def gem_command(command, args = "", options = {})
...
+      bundler_signals = %w[INT]
+
+      Signal.list.keys - (bundler_signals + reserved_signals + open3_reserved_signals)
+    end
+
     def sys_exec(cmd)
       command_execution = CommandExecution.new(cmd.to_s, Dir.pwd)
 
       Open3.popen3(cmd.to_s) do |stdin, stdout, stderr, wait_thr|
+        test_signals.each do |n|
```
segiddins: this probably shouldn't be executed on every call to `sys_exec`?

shayonj: I didn't see any concerns, but fair, moved to `bundler`, as its more clear now.

segiddins: No, I mean, shouldn't this just be done for the single test case?

shayonj: um! yeah we could, though I figured this would help have a consistent implementation of bundle exec in specs, when it comes to Signals. Do you anticipate concerns?

shayonj: I updated so that signal trapping is now specific to the spec itself.

segiddins:
Perfect, thanks! I'm going to make a couple of tiny changes, but this is basically good to go 👍

olleolleolle:
@segiddins Thanks for making those test changes!

Is this PR good to go now?

segiddins:
@bundlerbot r+

PR #6237, opened 2017-12-29
Fix warning in rubygems testing re __FILE__

MSP-Greg:

Testing with trunk & rubygems, many warnings of:
`rubygems/lib/rubygems.rb:940: warning: __FILE__ in eval may not return location in binding; use Binding#source_location instead`

Updates for warnings & 'security' in trunk are showing issues that should be addressed.

It's one line...

Unaware of other options...

Not sure if this will work with all ruby versions.  Did not test my fork...

### DIFF
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -570,8 +570,10 @@ module Bundler
       @replaced_methods.each do |(sym, klass), method|
         redefine_method(klass, sym, method)
       end
-      post_reset_hooks.reject! do |proc|
-        proc.binding.eval("__FILE__") == __FILE__
+      if Binding.public_method_defined?(:source_location)
+        post_reset_hooks.reject! {|proc| proc.binding.source_location == __FILE__ }
+      else
+        post_reset_hooks.reject! {|proc| proc.binding.eval("__FILE__") == __FILE__ }
       end
       @replaced_methods.clear
     end


MSP-Greg:
Interesting all the failures.  On appveyor, I added the patch.

[Log before](https://ci.appveyor.com/project/MSP-Greg/rubygems/build/58/job/ndpovfpfwkqm2gen)

[Log after](https://ci.appveyor.com/project/MSP-Greg/rubygems/build/job/gnswc0cae5y3rq64)

lib/bundler/rubygems_integration.rb
```diff
@@ -570,8 +570,10 @@ def undo_replacements
       @replaced_methods.each do |(sym, klass), method|
         redefine_method(klass, sym, method)
       end
-      post_reset_hooks.reject! do |proc|
-        proc.binding.eval("__FILE__") == __FILE__
+      if RUBY_VERSION < "2.2"
```
segiddins: can we do this with `respond_to?` instead of checking the version?

olleolleolle:
@segiddins The new diff looks like you describe it.

I think this looks good, now!

MSP-Greg:
Thanks.  The only reason I used `Binding.public_method_defined?`, is that it doesn't require creating a binding object...

segiddins:
This is perfect, thanks!

@bundlerbot r+

PR #6238, opened 2017-12-30
[CLI] Skip the outdated bundler check when MD5 is not available

### LINKED ISSUE #6032: Outdated version check causes bundle install to fail on FIPS enabled systems
I'm currently trying to get Rails apps installed on a FIPS enabled server, and appreciate the efforts that you've made recently to avoid using MD5 hashes (#4989 and follow-up #5433, along with recent refactor 6e1506ff8cc7393111267c3c5a0a29f8ecb4ecaa).

However the outdated version check that you introduced in 1.15.0 (1e05d24050b4177e9a8001b7c1083d713b474faf) is directly using the CompactIndex, and this causes bundle install to fail when FIPS is enabled.

Once I'd figured this out it was easy to fix using the disable_version_check setting (and it makes sense to turn this off anyway in my case).

I wonder though if you'd consider saving others this small pain by skipping the check regardless of the setting on FIPS enabled systems?

Many thanks.

segiddins:

The problem was Bundler would try to do the outdated version check on FIPS systems, leading to an exception.

Closes #6032.

My diagnosis was we needed to skip the check when MD5 is unavailable.

I chose this fix because using `#available?` would make network requests, which is not acceptable.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -718,6 +718,8 @@ module Bundler
       command_name = current_command.name
       return if PARSEABLE_COMMANDS.include?(command_name)
 
+      return unless SharedHelpers.md5_available?
+
       latest = Fetcher::CompactIndex.
                new(nil, Source::Rubygems::Remote.new(URI("https://rubygems.org")), nil).
                send(:compact_index_client).


olleolleolle:
@bundlerbot r+

olleolleolle:
@bundlerbot retry

segiddins:
@bundlerbot retry

Issue #6233, opened 2017-12-28
Can't specify platforms if force_ruby_platform is true


ahorek:
I want to specify a gem only for specific platforms.
```
gem "nokogiri", platforms: [:mri, :mingw, :x64_mingw]
```
this works fine. My problem is that for example nokogiri doesn't have a binary version for Ruby 2.5 yet, but it can be compiled for it.
```
Fetching nokogiri 1.8.1 (x64-mingw32)
Installing nokogiri 1.8.1 (x64-mingw32)
Gem::RuntimeRequirementNotMetError: nokogiri requires Ruby version < 2.5, >=
2.2. The current ruby version is 2.5.0.
An error occurred while installing nokogiri (1.8.1), and Bundler cannot
continue.
```

so I did it
```
gem install nokogiri --platform=ruby
```

however bundle update still wants x64-mingw32 platform

I tried
```
bundle lock --add-platform ruby
bundle config force_ruby_platform true

source 'https://rubygems.org'

gem "nokogiri"
```
now bundler recognises ruby platform.
```
Resolving dependencies...
Using bundler 1.16.1
Using mini_portile2 2.3.0
Using nokogiri 1.8.1
Bundle updated!
```

but if I specify the ruby platform in my Gemfile
```
gem "nokogiri", platforms: [:ruby]
```

```
Resolving dependencies...
Using bundler 1.16.1
Bundle updated!
```
nokogiri is gone.

**desired behaviour**
In my opinion the best way would be to use a compiled version automatically if a binary version isn't available, but it was rejected.

according to https://github.com/bundler/bundler/issues/5337 the platform can be forced only for a whole Gemfile, but there's a problem
```
bundle lock --add-platform ruby
bundle config force_ruby_platform true

Gemfile
gem 'nokogiri'
gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby, :ruby]

=> both gems should be installed
```

**actual behaviour**
```
'tzinfo-data' is ignored unless I remove all platforms specifications.
```
I want to have a single platform independent Gemfile that works for all 3 cases:
Linux
Windows with binary gems
Windows with force_ruby_platform

Am I missing something? Any suggestions?

## Environment

```
Bundler       1.16.1
  Platforms   ruby, x64-mingw32
Ruby          2.5.0p0 (2017-12-25 revision 61468) [x64-mingw32]
  Full Path   C:/Ruby25-x64/bin/ruby.exe
[...]

Issue #5337, opened 2017-01-19
Bundler doesn't respect other than the native platform while resolving the gems to install


larskanis:
Original issue is https://github.com/larskanis/rubyinstaller2/issues/7 : Nokogiri installs successfully on [RubyInstaller2](https://github.com/larskanis/rubyinstaller2) when `--platform=ruby` is used. There is also a fat binary gem for x86-mingw32, but it is limited to Ruby versions 2.0 to 2.3. It can not be installed on Ruby-2.4.0 due to this ruby version constraint, which is expected behavior. However bundler doesn't use the `platform=ruby` gem version as an alternative, regardless whether it is already installed or not.

Installed versions (with nokogiri installed as platform=ruby):
```sh
C:\Users\lars\railstest>ruby --version
ruby 2.4.0p0 (2016-12-24 revision 57164) [i386-mingw32]

C:\Users\lars\railstest>bundle --version
Bundler version 1.14.0.pre.2

C:\Users\lars\railstest>gem list nokogiri
*** LOCAL GEMS ***
nokogiri (1.7.0.1)
```

Using a fresh Rails-5.0.1 project (generated per `rails new railstest`) with [this Gemfile](https://gist.github.com/larskanis/f030e70dc823a882fadeac67fa8621a1) .

When platform `ruby` is added per `bundle lock`, I would expect, that the installed version will be used. But instead an older nokogiri version (1.6.6.4) for `x86-mingw32` is installed. This version doesn't work on Ruby-2.4, since it doesn't contain the Ruby-2.4 binary extension, but it wasn't labeled with proper ruby version constraints to that point in time:

```sh
C:\Users\lars\railstest>bundle lock --add-platform=ruby
Fetching gem metadata from https://rubygems.org/..........
Fetching version metadata from https://rubygems.org/..
Fetching dependency metadata from https://rubygems.org/.
Resolving dependencies........................................................................................................................................................................................................................................
Writing lockfile to C:/Users/lars/railstest/Gemfile.lock

C:\Users\lars\railstest>bundle
Fetching gem metadata from https://rubygems.org/.........
Fetching version metadata from https://rubygems.org/..
Fetching dependency metadata from https://rubygems.org/.
Using rake 12.0.0
[...]
Installing nokogiri 1.6.6.4 (x86-mingw32)
[...]
Bundle complete! 12 Gemfile dependencies, 56 gems now installed.
```

[...]

Issue #598, opened 2010-08-21
Patch for not expanding path with `gemspec`


mitchellh:
I confirmed this bug earlier today with Carl. This is in regards to issue 597 which was closed but is indeed a bug. The linked patch fixes this and tests continue to pass. The reason path is modified again instead of doing it all in the `Dir` is so that if there is an error, then the path given in the error message is still correct.

https://gist.github.com/17eea379a14ec8206b52

PR #6247, opened 2018-01-06
Add two specs for choosing platform specific gems

larskanis:
This is a follow-up to #5337. It adds two specs for choosing platform specific gems. The first spec succeeds, but the second spec fails currently.

Bundler doesn't properly resolve platform gems with ruby version constraints.

rake-compiler [tags all fat binary gems](https://github.com/rake-compiler/rake-compiler/blob/b00d1ce46e316126c0ee8890888d1788faa2cee1/lib/rake/extensiontask.rb#L261-L264) with appropriate `required_ruby_version` constraints. The intention behind this tag is to give bundler a chance to select the most suitable gem for the given platform and ruby version. Unfortunately it doesn't work.

This leads to issues for all Windows users whenever a new Ruby version comes out. The intended effect (to make the install of gems easy) turns to the opposite, because the platform specific gem effectively blocks the installation of the "ruby" platform gem.

These issues can be seen at many popular gems with C extensions: [ffi #598](https://github.com/ffi/ffi/issues/598), [nokogiri #1706](https://github.com/sparklemotion/nokogiri/issues/1706), [pg](https://github.com/ged/ruby-pg)

The first test case doesn't offer a x64-mingw32 gem in the latest version. Bundler therefore selects the latest "ruby" platform gem so that the installation succeeds. This is OK and expected.

The second test is very similar, but offers a x64-mingw32 gem in addition. This additional gem however is limited to ruby versions not matching the running ruby version. Bundler should exclude this gem therefore and should select the "ruby" platform gem as in the first test. However it doesn't resolve to the "ruby" platform gem, but to an older x64-mingw32 gem, which wasn't tagged by a `required_ruby_version` constraint to that point in time (and will not work with the latest ruby version in practice). That is the current situation with ffi and nokogiri.

The current workaround is to set `force_ruby_platform` to true, [but this has several negative consequences](https://github.com/bundler/bundler/issues/6233) since it's a global and not a per-gem setting. Also it has to be enabled manuelly by each (Windows) user. Insofar it's not a general solution.

I didn't dig deeper into bundlers internals, but my hope is that someone with more knowledge of bundler can solve this issue that way or at least give me a pointer where I could start to fix this.

### DIFF
--- a/spec/resolver/platform_spec.rb
+++ b/spec/resolver/platform_spec.rb
@@ -40,6 +40,20 @@ RSpec.describe "Resolving platform craziness" do
     should_resolve_as %w[foo-1.1.0]
   end
 
+  it "takes the ruby version if the platform version is incompatible" do
+    @index = build_index do
+      gem "bar", "1.0.0"
+      gem "foo", "1.0.0"
+      gem "foo", "1.0.0", "x64-mingw32" do
+        dep "bar", "< 1"
+      end
+    end
+    dep "foo"
+    platforms "x64-mingw32"
+
+    should_resolve_as %w[foo-1.0.0]
+  end
+
   it "takes the latest ruby gem if the platform specific gem doesn't match the required_ruby_version" do
     @index = build_index do
       gem "foo", "1.0.0"
--- a/spec/resolver/platform_spec.rb
+++ b/spec/resolver/platform_spec.rb
@@ -28,6 +28,35 @@ RSpec.describe "Resolving platform craziness" do
     end
   end
 
+  it "takes the latest ruby gem, even if an older platform specific version is available" do
+    @index = build_index do
+      gem "foo", "1.0.0"
+      gem "foo", "1.0.0", "x64-mingw32"
+      gem "foo", "1.1.0"
+    end
+    dep "foo"
+    platforms "x64-mingw32"
+
+    should_resolve_as %w[foo-1.1.0]
+  end
+
+  it "takes the latest ruby gem if the platform specific gem doesn't match the required_ruby_version" do
+    @index = build_index do
+      gem "foo", "1.0.0"
+      gem "foo", "1.0.0", "x64-mingw32"
+      gem "foo", "1.1.0"
+      gem "foo", "1.1.0", "x64-mingw32" do |s|
+        s.required_ruby_version = [">= 2.0", "< 2.4"]
+      end
+      gem "ruby\0", "2.5.1"
+    end
+    dep "foo"
+    dep "ruby\0", "2.5.1"
+    platforms "x64-mingw32"
+
+    should_resolve_as %w[foo-1.1.0]
+  end
+
   describe "with mingw32" do
     before :each do
       @index = build_index do


indirect:
According to the scenario you just described, it sounds like Bundler is correctly choosing a gem that declares itself to work in the current situation.

If a gem itself is broken, Bundler resolving to that gem is not a bug. Can you explain more about why you think this is a bug in Bundler?

larskanis:
See the following table. "X" marks which of the given 4 gems are available on the index. "=>" marks, which gem is chosen in which situation.

<table><th ><td  ><p>1) Ruby-2.3 or 2.5</p></td><td  ><p>2) Ruby-2.5</p></td><td  ><p>3) Ruby-2.5</p></td><td  ><p>4) Ruby-2.3</p></td><td  ><p>5) Ruby-2.3 or 2.5</p></td></th><tr ><td  ><p>foo-1.0.0</p></td><td  ><p>X</p></td><td  ><p>X</p></td><td  ><p>X</p></td><td  ><p>X</p></td><td  ><p>X</p></td></tr><tr ><td  ><p>foo-1.0.0-x64-mingw32</p></td><td  ><p>X</p></td><td  ><p>X &lt;= current</p></td><td  > </td><td  ><p>X</p></td><td  > </td></tr><tr ><td  ><p>foo-1.1.0</p></td><td  ><p>X &lt;= OK</p></td><td  ><p>X &lt;= expected</p></td><td  ><p>X &lt;= expected</p></td><td  ><p>X</p></td><td  ><p>X &lt;= OK</p></td></tr><tr ><td  ><p>foo-1.1.0-x64-mingw32</p><p>only Ruby &lt; 2.5</p></td><td  > </td><td  ><p>X</p></td><td  ><p>X</p></td><td  ><p>X &lt;= OK</p></td><td  > </td></tr><tr ><td  > </td><td  > </td><td  > </td><td  ><p>=&gt; resolver fails currently</p></td><td  > </td><td  > </td></tr></table>

Case 1) corresponds to the first and 2) to the second spec in the attached PR. Cases 1), 4) and 5) succeed and the result is as expected.

In case 2) bunder finds a valid solution, but the solution is bad for two reasons: 1. it installs a (possibly very) old gem, which didn't receive any security or feature updates. And 2. it is an inconsistent solution, because the result of the resolver is different to 1) although the list of suitable gems is exactly the same as in 1).

To avoid situation 2) it would be an option to yank all the old x64-mingw32 gems, which are not tagged with `require_ruby_version`, but this leads directly to case 3): bundler completely fails to resolve.

In case 3) bundler doesn't find a solution due to the availability of a x64-mingw32 gem in the latest version 1.1.0. This is although the list of suitable gems is exactly the same as in case 5). It should therefore pick "foo-1.1.0" instead of giving up.

For some reasons bundler doesn't always resolve like in case 2), but also like in case 3). I guess this is due to some optimization while downloading the gem metadata. Therefore the linked issues describe case 3) - bundler fails to resolve: [ffi #598](https://github.com/ffi/ffi/issues/598), [nokogiri #1706](https://github.com/sparklemotion/nokogiri/issues/1706)

indirect:
@larskanis thank you for that extremely detailed explanation! 

It sounds like you are describing two completely different (although possibly related) problems:

Problem 1: When specific-platform gems are disallowed by `required_ruby_version`, Bundler fails the resolution when you would instead expect it to fall back to a pure-ruby version of that gem.

Problem 2: Rather than choose the valid option of a newer pure-ruby gem, Bundler chooses the equally valid option of an older platform-specific gem.

If I am understanding correctly, problem 1 sounds like a bug that we need to investigate and fix. Once that bug is fixed, if problem 2 is still present it will likely need to be fixed by changes to users' Gemfiles. Getting a different valid resolution than the one you want means you need to inform Bundler about what it is that you want, so that Bundler can deliver it.

Thank you again for your detailed report and explanation!

segiddins:
I think (2) is reasonable -- imho, a closer version should be preferred to a closer platform match.

larskanis:
@segiddins Thank you for looking at this issue! Your spec shows another interesting failure.

However I imploringly hope this issue can be fixed! RubyInstaller-2.5.0 was downloaded almost 50000 times within the last two weeks, although I [explicitly discourage user to use 2.5](https://rubyinstaller.org/downloads/). Now almost every user stumbles upon this issue. And I get daily "same here" reports about these failures.

The only way to work around it, is to use `force_ruby_platform`, or go without bundler and just use rubygems with `--platform ruby` option. Although `force_ruby_platform` is more usable since RubyInstaller-2.4, thanks to the MSYS2 foundation, it's still far from "just works".

btir:
Any updates here? The problem remains in place after 2 years!
@larskanis @segiddins @indirect

kou:
I've implemented these added specs in #7522.
#7522 includes specs added in this pull request.

deivid-rodriguez:
Closing in favor of #7522 then. Thanks @kou and everybody helping out here!

Issue #6220, opened 2017-12-22
invalid byte sequence in US-ASCII (ArgumentError)


biow0lf:
Hello,

```
$ bundle
Fetching https://github.com/plataformatec/devise.git
Traceback (most recent call last):
	10: from /Users/biow0lf/.rbenv/versions/2.5.0-rc1/bin/bundle:23:in `<main>'
	 9: from /Users/biow0lf/.rbenv/versions/2.5.0-rc1/bin/bundle:23:in `load'
	 8: from /Users/biow0lf/.rbenv/versions/2.5.0-rc1/lib/ruby/gems/2.5.0/gems/bundler-1.16.1/exe/bundle:22:in `<top (required)>'
	 7: from /Users/biow0lf/.rbenv/versions/2.5.0-rc1/lib/ruby/gems/2.5.0/gems/bundler-1.16.1/lib/bundler/friendly_errors.rb:121:in `with_friendly_errors'
	 6: from /Users/biow0lf/.rbenv/versions/2.5.0-rc1/lib/ruby/gems/2.5.0/gems/bundler-1.16.1/lib/bundler/friendly_errors.rb:126:in `rescue in with_friendly_errors'
	 5: from /Users/biow0lf/.rbenv/versions/2.5.0-rc1/lib/ruby/gems/2.5.0/gems/bundler-1.16.1/lib/bundler/friendly_errors.rb:17:in `log_error'
	 4: from /Users/biow0lf/.rbenv/versions/2.5.0-rc1/lib/ruby/gems/2.5.0/gems/bundler-1.16.1/lib/bundler/ui/shell.rb:45:in `error'
	 3: from /Users/biow0lf/.rbenv/versions/2.5.0-rc1/lib/ruby/gems/2.5.0/gems/bundler-1.16.1/lib/bundler/ui/shell.rb:105:in `tell_me'
	 2: from /Users/biow0lf/.rbenv/versions/2.5.0-rc1/lib/ruby/gems/2.5.0/gems/bundler-1.16.1/lib/bundler/vendor/thor/lib/thor/shell/basic.rb:88:in `say'
	 1: from /Users/biow0lf/.rbenv/versions/2.5.0-rc1/lib/ruby/gems/2.5.0/gems/bundler-1.16.1/lib/bundler/vendor/thor/lib/thor/shell/basic.rb:88:in `!~'
[...]

PR #6254, opened 2018-01-13
[FriendlyErrors] Fallback to the original error if the friendly message raises

segiddins:

This would have made diagnosing https://github.com/bundler/bundler/issues/6220 much easier

My diagnosis was we do a lot in our error handler, so it needs a fallback

My fix falls back to just re-raising the original error


indirect:
oh yes, this is an extremely good idea. thank you. r+

segiddins:
@bundlerbot r+

segiddins:
@bundlerbot retry

PR #6257, opened 2018-01-15
Rename Bundler.frozen? to avoid Object method conflict

### LINKED ISSUE #6252: Bundler.frozen? overriding Object.frozen? causing false positive frozen checks
https://ruby-doc.org/core-2.5.0/Object.html#method-i-frozen-3F

- `Bundler.frozen?` => gemfile frozen
- `Object.frozen?` => object is frozen

Because of the override one cannot actually check if the Bundler module itself is frozen. Which is a problem as, for example, the mocking framework mocha (https://github.com/freerange/mocha) will check `Klass.frozen?` and raise an error when trying to mock a method on a frozen object.

example code
```
require 'bundler'
require 'mocha'
require 'mocha/object_methods'
Object.send(:include, Mocha::ObjectMethods) # test framework hooks do this

Bundler.stubs(:foo)
```

with

```
echo "---\nBUNDLE_FROZEN: \"true\"" >> .bundle/config
```

results in 

```
➜ ruby test
Traceback (most recent call last):
test:6:in `<main>': can't stub method on frozen object: Bundler (Mocha::StubbingError)
```

It would be good if frozen? was renamed so it doesn't change the meaning inherited from Object.frozen? and one can again check if the module is frozen.

segiddins:

The problem was `Bundler` overrode the `Object#frozen?` method to mean something else, breaking anything that checks that method to determine if an object is frozen.

Closes #6252.

My diagnosis was the method name had to be changed.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -136,7 +136,7 @@ module Bundler
       end
     end
 
-    def frozen?
+    def frozen_bundle?
       frozen = settings[:deployment]
       frozen ||= settings[:frozen] unless feature_flag.deployment_means_frozen?
       frozen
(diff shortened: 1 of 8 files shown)


segiddins:
@bundlerbot r+

PR #6258, opened 2018-01-15
[Source::RubyGems] Ensure the bindir exists before installing

segiddins:

The problem was occasional failures on CI w/ ruby 2.5

My diagnosis was that there was a race condition in the rubygems installer, see https://github.com/rubygems/rubygems/pull/2148 for the fix there.

My fix will ensure that directory is created before we get into that rubygems code path


colby-swandale:
Thank you for looking into this issue! 👍  @bundlerbot r+

PR #6262, opened 2018-01-18
add the new team members of the bundler core team to the author list in the Bundler gemspec

colby-swandale:
Just adding myself and Hiroshi-san to the list of authors in the Bundler gemspec

\cc @indirect


hsbt:
👍

Issue #6261, opened 2018-01-16
Zlib::GzipFile::Error: unexpected end of file


chrishickman:
# Error Report

## Questions

Please fill out answers to these questions, it'll help us figure out
why things are going wrong.

- **What did you do?**

  I ran the command `/usr/local/rbenv/versions/2.3.3/bin/bundle install`

- **What did you expect to happen?**

  I expected Bundler to install successfully

- **What happened instead?**

  Instead, what happened was, I received the error below after bundler failed.

- **Have you tried any solutions posted on similar issues in our issue tracker, stack overflow, or google?**

  I haven't found any posted solutions but after a series of trial and error, I noticed that it is happening when Rails attempts to install a new em via the Gemfile.  We are in a classroom with 20 students and are receiving this error with 5/15 on what should be identical environments.  On 4 of the 5 machines having issues, if we list gems previously used in other projects bundle install works fine.  Anytime we add a new gem we receive the error listed below with Gzip.  The same 4 machines can install gems directly using "gem install bootstrap..." etc. and then when listed in the Gemfile it works fine.  1 of the 5 machines will not run bundle install no matter what we do.  Brand new project with nothing added to the Gemfile other than the default.... still receiving the same error.   

- **Have you read our issues document, https://github.com/bundler/bundler/blob/master/doc/contributing/ISSUES.md?**

  ...

## Backtrace

```
Zlib::GzipFile::Error: unexpected end of file
  /usr/local/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/bundler-1.16.0/lib/bundler/compact_index_client/updater.rb:56:in `read'
  /usr/local/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/bundler-1.16.0/lib/bundler/compact_index_client/updater.rb:56:in `block in update'
  /usr/local/rbenv/versions/2.3.3/lib/ruby/2.3.0/tmpdir.rb:89:in `mktmpdir'
  /usr/local/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/bundler-1.16.0/lib/bundler/compact_index_client/updater.rb:31:in `update'
  /usr/local/rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/bundler-1.16.0/lib/bundler/compact_index_client.rb:82:in `update'
[...]

PR #6264, opened 2018-01-19
handle gzip corruption errors in the compact index client

colby-swandale:

Bundler will raise an exception and crash if the Gzip that the Compact Client Index downloaded was corrupt.

See #6261 

Handle the exception and allow Bundler to continue without crashing


indirect:
thanks for catching this! @bundlerbot r+

Issue #6219, opened 2017-12-21
Arch Linux 3.10.18 ARM V7 --  command Bundle init crash


aperfeito:
[aperfeito@alarm /]$ ./home/aperfeito/.gem/bin/bundle init
--- ERROR REPORT TEMPLATE -------------------------------------------------------
# Error Report

## Questions

Please fill out answers to these questions, it'll help us figure out
why things are going wrong.

- **What did you do?**

  I ran the command `./home/aperfeito/.gem/bin/bundle init`

- **What did you expect to happen?**

  I expected Bundler to...

- **What happened instead?**

  Instead, what happened was...

- **Have you tried any solutions posted on similar issues in our issue tracker, stack overflow, or google?**

  I tried...

- **Have you read our issues document, https://github.com/bundler/bundler/blob/master/doc/contributing/ISSUES.md?**

  ...

## Backtrace

```
Errno::EACCES: Permission denied @ rb_sysopen - Gemfile
  /home/aperfeito/.gem/gems/bundler-1.16.1/lib/bundler/vendor/fileutils/lib/fileutils.rb:1290:in `initialize'
  /home/aperfeito/.gem/gems/bundler-1.16.1/lib/bundler/vendor/fileutils/lib/fileutils.rb:1290:in `open'
  /home/aperfeito/.gem/gems/bundler-1.16.1/lib/bundler/vendor/fileutils/lib/fileutils.rb:1290:in `block in copy_file'
  /home/aperfeito/.gem/gems/bundler-1.16.1/lib/bundler/vendor/fileutils/lib/fileutils.rb:1289:in `open'
  /home/aperfeito/.gem/gems/bundler-1.16.1/lib/bundler/vendor/fileutils/lib/fileutils.rb:1289:in `copy_file'
  /home/aperfeito/.gem/gems/bundler-1.16.1/lib/bundler/vendor/fileutils/lib/fileutils.rb:430:in `copy_file'
  /home/aperfeito/.gem/gems/bundler-1.16.1/lib/bundler/vendor/fileutils/lib/fileutils.rb:357:in `block in cp'
  /home/aperfeito/.gem/gems/bundler-1.16.1/lib/bundler/vendor/fileutils/lib/fileutils.rb:1461:in `block in fu_each_src_dest'
  /home/aperfeito/.gem/gems/bundler-1.16.1/lib/bundler/vendor/fileutils/lib/fileutils.rb:1477:in `fu_each_src_dest0'
[...]

PR #6266, opened 2018-01-19
[Init] Check if the current directory is writeable

nilsding:

The problem was that when running `bundle init` inside a directory which is not writable by the current user (e.g. `/` as demonstrated in #6219) Bundler prints out an `EACCES` error with a huge backtrace.  In the mentioned PR @segiddins suggested to print out a better error message.  This PR addresses that.

See [this comment on said PR](https://github.com/bundler/bundler/issues/6219#issuecomment-359002919).

My fix is simple: adding a check whether the current directory is writeable before trying to create `gems.rb`/`Gemfile`.  If that's not the case, print out an error and exit.

I chose this fix because... it was really simple to implement.

### DIFF
--- a/spec/commands/init_spec.rb
+++ b/spec/commands/init_spec.rb
@@ -64,6 +64,22 @@ RSpec.describe "bundle init" do
     end
   end
 
+  context "when the dir is not writable by the current user" do
+    let(:subdir) { "child_dir" }
+
+    it "notifies the user that it can not write to it" do
+      FileUtils.mkdir bundled_app(subdir)
+      FileUtils.chmod "a-w", bundled_app(subdir)
+
+      Dir.chdir bundled_app(subdir) do
+        bundle :init
+      end
+
+      expect(out).to include("directory is not writable")
+      expect(bundled_app(subdir)).to be_empty
+    end
+  end
+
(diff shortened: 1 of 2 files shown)


spec/commands/init_spec.rb
```diff
@@ -64,6 +64,22 @@
     end
   end
 
+  context "when the dir is not writable by the current user", :bundler => ">= 2" do
```
colby-swandale: This shouldn't be tagged with `:bundler => ">= 2"` because this is not testing any specific Bundler 2 functionality.

nilsding: Is there any other 1.x release planned?

colby-swandale: Yes but that isn't why i'm saying don't use `:bundler => ">= 2"`. We use that tag when we have to test something specifically for Bundler 2 such as UI changes like new options to commands. The error you're asserting for in your spec won't change between Bundler 1 and Bundler 2.

colby-swandale: yes

nilsding: oh, alright.  I’ll remove that tag there then and try to make the test pass with Ruby 1.8

colby-swandale:
@bundlerbot r+

PR #6267, opened 2018-01-20
Add base error class to new gems.

### LINKED ISSUE #6260: Consider creating error base class in `bundler gem`
Ruby 2 updated the docs for `Exception` [from](https://ruby-doc.org/core-1.9.3/Exception.html):

> Programs may subclass Exception, or more typically StandardError to provide custom classes and add additional information.

[to instead read](https://ruby-doc.org/core-2.0.0/Exception.html):

> It is recommended that a library should have one subclass of StandardError or RuntimeError and have specific exception types inherit from it. This allows the user to rescue a generic exception type to catch all exceptions the library may raise even if future versions of the library add new exception subclasses.
>
> For example:
> 
> ```ruby
> class MyLibrary
>   class Error < RuntimeError
>   end
> 
>   class WidgetError < Error
>   end
> 
>   class FrobError < Error
>   end
> 
> end
> ```
>
> To handle both WidgetError and FrobError the library user can rescue MyLibrary::Error.

This is [awesome, and important](https://twitter.com/mperham/status/952969180856205312).

------

I'd like to propose adding a default error base class to the `bundle gem` scaffolding to encourage this good pattern. This could be as simple as `bundle gem foo` producing:

```diff
  # foo/lib/foo.rb
  require "foo/version"

  module Foo
+   class Error < StandardError; end
    # Your code goes here...
  end
```

It is nearly as effortless to delete both that line and the comment line as the comment line itself; but plants the seed for a very important and often-overlooked best practice early in a gem's development.

------

If well received I think we could discuss whether or not to add more context/comments to it than just the one-liner class; as well as minutiae like if the `; end` is good form. Initially just gonna pitch the general concept, though, I don't want to get too caught up in those other more focused discussions immediately.

christhekeele:
Closes #6260.

Room for discussion:

- Which error class to use (`StandardError` makes sense to me)
- What formatting to use (the three-lines-with-comment seemed nicest to me)
- Whether or not using the flag to provide a different error base class is useful, and if it should validate the user's choice or not (I threw it in because it seemed harmless; is it? a boolean flag would work fine too)

---

Libraries don't always follow best practice from discussion in linked issue.

Bundler could encourage best practice by adding it to the gem scaffold.

I added a base error class to the templates, and provided a flag to change/disable this behaviour.

Like any best-practice-by-default, this could ruin someones workflow/go against someone's preferences so I made it as configurable as possible.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -515,8 +515,6 @@ module Bundler
     method_option :mit, :type => :boolean, :desc => "Generate an MIT license file. Set a default with `bundle config gem.mit true`."
     method_option :test, :type => :string, :lazy_default => "rspec", :aliases => "-t", :banner => "rspec",
                          :desc => "Generate a test directory for your library, either rspec or minitest. Set a default with `bundle config gem.test rspec`."
-    method_option :error_class, :type => :string, :default => "StandardError", :aliases => "-e",
-                                :desc => "Generate a base error class for your library."
     def gem(name)
     end
 
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -515,6 +515,8 @@ module Bundler
     method_option :mit, :type => :boolean, :desc => "Generate an MIT license file. Set a default with `bundle config gem.mit true`."
     method_option :test, :type => :string, :lazy_default => "rspec", :aliases => "-t", :banner => "rspec",
                          :desc => "Generate a test directory for your library, either rspec or minitest. Set a default with `bundle config gem.test rspec`."
+    method_option :error_class, :type => :string, :default => "StandardError", :aliases => "-e",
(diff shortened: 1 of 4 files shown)


colby-swandale:
Thanks for opening a PR and helping make Bundler better. I apologize for saying this after you've done all this work but i imagined this PR would've just add the one line from #6267. Adding in a CLI option and adding in validation just doesn't seem necessary here. It's a simple change for the user if they want to remove and use a different error as the base class.

christhekeele:
Hey, no hassle—done!

olleolleolle:
I think this looks good to go!

colby-swandale:
@bundlerbot r+

colby-swandale:
Thanks @christhekeele for the PR!

PR #6269, opened 2018-01-23
Keep generating a `Gemfile` in `bundle init` for Bundler 2 and add an option to generate `gems.rb`

colby-swandale:

The `bundle init` command in Bundler 2 currently generates a `gems.rb` file by default. The issue with this is that i believe we're pushing `gems.rb` as a "successor" to the Gemfile that everyone should be using even though there is no real incentive for nearly any ruby project to do so.

All the documentation on the internet about the the `gemfile` is referred to as `Gemfile` which I can see will confuse new users to Ruby. If they're looking to add a gem to their project and see the project's README there is a very good chance it will say to "add <gemname> to the Gemfile and run bundle install" [Example 1](https://github.com/sinatra/sinatra#with-bundler), [Example 2](https://github.com/puma/puma#rails) people will get confused from this. Or if they're learning Bundler for the first time and come across a blog post that talks about the `gemfile` which is actually `gems.rb` in their project will also cause confusion for the user.

The last issue i see is that we'll be creating a situation where a lot of projects will be using `Gemfile` and others will be using `gems.rb` and that this transition may never be complete because the incentive is just not there.

I believe these issues outweigh the benefit of forcing `gems.rb` onto users in the UI and I want to propose that we instead have `bundle init` keep generating a `Gemfile` as it has been but introduce an option to generate a `gems.rb` if the user needs it.

(Note: the `bundle gem` command is un-effected as we didn't get around to update the command to use `gems.rb` in Bundler 2)

Make `bundle init` continue to generate a `Gemfile` by default and add an option to allow generating a `gems.rb` file if needed.

cc @indirect @segiddins

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -143,6 +143,7 @@ module Bundler
       dependency listed in the gemspec file to the newly created Gemfile.
     D
     deprecated_option "gemspec", :type => :string, :banner => "Use the specified .gemspec to create the Gemfile"
+    method_option "gemsrb", :type => :boolean, :default => false, :banner => "Generate a gems.rb instead of a Gemfile"
     def init
       require "bundler/cli/init"
       Init.new(options.dup).run
--- a/lib/bundler/cli/init.rb
+++ b/lib/bundler/cli/init.rb
@@ -36,7 +36,7 @@ module Bundler
   private
 
     def gemfile
-      @gemfile ||= Bundler.feature_flag.init_gems_rb? ? "gems.rb" : "Gemfile"
+      @gemfile ||= @options[:gemsrb] ? "gems.rb" : "Gemfile"
     end
   end
 end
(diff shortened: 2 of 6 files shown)


segiddins:
I think I agree with this line of reasoning?

colby-swandale:
@segiddins did you have any questions?

indirect:
I think I'm okay with this. In the interests of reducing surprise, shipping 2.x without making `gems.rb` the default seems reasonable. Then we can have the discussion about possibly changing the default again when we get to 3.0. 🙂

indirect:
After thinking about this for a few more minutes, I personally would like to be able to set a config flag with the same effect as `--gemsrb`. Can we keep the code that makes it a setting, but remove the default value change for Bundler 2?

deivid-rodriguez:
I agree with this PR in combination with @indirect's suggestion of keeping the feature flag! :+1:

indirect:
@colby-swandale still interested in re-enabling the feature flag and merging this?

colby-swandale:
i'll put the setting back.

deivid-rodriguez:
This PR was superseded by #7113, which implements the same line of reasoning, so closing!

PR #6272, opened 2018-01-24
Update contributing documentation to use new tags

colby-swandale:
Github has defined a couple of tags to use for classifying issues that are suitable for new contributors to fix/contribute towards. We have started using these tags in Github so we should start linking to them in our documentation as well.


colby-swandale:
@bundlerbot r+

PR #6275, opened 2018-01-26
Fix failing spec in the bundle exec system gem spec

colby-swandale:

There is an issue where a spec was failing because it was activating the version of Bundler inside of RubyGems and not the development version that we're working on.

This would cause the Lockfile to raise an error because we were generating a Bundler 2 lockfile in our tests but Bundler 1 was being activated making it complain.

Add the Bundler development dir in the Ruby include argument.

This error is raising in CI and is preventing PRs from being merged.

See https://travis-ci.org/bundler/bundler/jobs/333434747


segiddins:
@bundlerbot r+

PR #6284, opened 2018-02-01
Remove vestigial CONTRIBUTING doc.

indirect:
The primary contributing doc is now located at docs/README.md, and we
link to that file from the main project README. I think this file causes
more confusion than it helps with, and anyone interested in contributing
can get better info from the "Contributing" section in the readme.


olleolleolle:
@indirect Reducing confusion in new contributors is _so good_.

@bundlerbot r+

PR #6291, opened 2018-02-09
To use dynamic expectation with same versioned bundler installed as default gems

hsbt:
I fixed test fails with bundler-1.16.1 as default gems.

https://travis-ci.org/rubygems/rubygems/jobs/339325369

When the system has a same version of bundler as default gems(Currently it's 1.16.1), It conflicts on command invocation in bundler's examples. I added two workarounds.

1. Bump virtual version of bundler in `binstubs_spec.rb`
2. In other examples, I modified expected variables when it was installed bundler as default gems.

### DIFF
--- a/spec/commands/binstubs_spec.rb
+++ b/spec/commands/binstubs_spec.rb
@@ -100,6 +100,9 @@ RSpec.describe "bundle binstubs <gem>" do
         bundle! "binstubs bundler rack prints_loaded_gems"
       end
 
+      # When environment has a same version of bundler as default gems.
+      # `system_gems "bundler-x.y.z"` will detect system binstub.
+      # We need to avoid it by virtual version of bundler.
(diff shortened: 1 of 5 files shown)


spec/commands/binstubs_spec.rb
```diff
@@ -100,14 +100,22 @@
         bundle! "binstubs bundler rack prints_loaded_gems"
       end
 
-      let(:system_bundler_version) { Bundler::VERSION }
+      let(:system_bundler_version) { Gem::Version.new(Bundler::VERSION).bump.to_s }
```
segiddins: Could you add a comment here why the bump is necessary?

hsbt: It makes sense. I added its detail at https://github.com/bundler/bundler/pull/6291/commits/cf5dd0e6a35ddfc554f705099dfba207c3b54583

segiddins:
@bundlerbot r+

Issue #6193, opened 2017-11-28
Make comment with timestamp optional when using bundle add


cpgo:
This is not a big deal, but I dont think the timestamp comment adds a lot of value on the Gemfile.
At this point its almost certain that the developer will be using some version control system.
`bundle add mygem --timestamp` could be an option

arbonap:
I'd love to take a hand at this 👌🏼

indirect:
This is definitely a leftover from the old `inject` command... we definitely want to change this. Ideally, we will eventually come up with some way for the `add` command to insert new gems into sorted lists of gems, if sorted lists exist, but this is a great first change. :)

colby-swandale:
@arbonap are you still working on this? If not, let us know incase someone else want's to work on this issue.

315tky:
Hello, I'm looking to start contributing to a ruby project, and have chosen bundler. I'm going to try this "issue" for starters.  

After a quick look, seems the "add" option ( lib/cli/add.rb ) is calling lib/injector.rb and I can see the line in there that creates the timestamp in Gemfile....

colby-swandale:
Probably not, i think we're just going to remove that piece of functionality entirely.

colby-swandale:
This was merged in #6300. Closing.

PR #6300, opened 2018-02-20
Remove comment with timestamp on `bundle add`

cpgo:
As discussed on #6193 this PR simply removes the comment timestamp when adding a gem via command line.

### DIFF
--- a/lib/bundler/injector.rb
+++ b/lib/bundler/injector.rb
@@ -84,9 +84,6 @@ module Bundler
     def append_to(gemfile_path, new_gem_lines)
       gemfile_path.open("a") do |f|
         f.puts
-        if @options["timestamp"] || @options["timestamp"].nil?
-          f.puts "# Added at #{Time.now} by #{`whoami`.chomp}:"
-        end
         f.puts new_gem_lines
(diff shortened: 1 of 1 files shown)


lib/bundler/injector.rb
```diff
@@ -84,9 +84,6 @@ def build_gem_lines(conservative_versioning)
     def append_to(gemfile_path, new_gem_lines)
       gemfile_path.open("a") do |f|
         f.puts
-        if @options["timestamp"] || @options["timestamp"].nil?
```
segiddins: we should probably give a deprecation warning is the `timestamp` option is given via the command line though?

indirect:
The timestamp is honestly purely a leftover implementation detail from `bundle inject`, and I never intended for it to be included in `add`. I'm happy to remove it completely from `add`.

colby-swandale:
Thanks!

@bundlerbot r+

rmm5t:
Is this something that can be backported to v1.16, or is it only slated for v2?

colby-swandale:
This will be part of the next bundler patch release.

PR #6302, opened 2018-02-21
Document `# frozen_string_literal` in `bundle init` Gemfile

315tky:

An enhancement request to bundle-init docu as per - issue 6140 

n/a

Updated the man/bundle-init.ronn file to include brief description of frozen string literal, and added a SEE ALSO section with link to Gemfile(5) webpage

File edited to try and reflect the previous discussion regarding what was required.

### DIFF
--- a/man/bundle-init.ronn
+++ b/man/bundle-init.ronn
@@ -12,16 +12,18 @@ adding a [`Gemfile(5)`][Gemfile(5)] to a gem with a gemspec, the `--gemspec` opt
 automatically add each dependency listed in the gemspec file to the newly
 created [`Gemfile(5)`][Gemfile(5)].
 
-One of the lines included in the default [`Gemfile(5)`][Gemfile(5)]
-generated is `# frozen_string_literal: true`. This is a magic comment
-supported for the first time in Ruby 2.3. The presence of this line
-results in all string literals in the file being implicitly frozen.
-Frozen string literals will be the default in Ruby 3.
-
 ## OPTIONS
 
 * `--gemspec`:
   Use the specified .gemspec to create the [`Gemfile(5)`][Gemfile(5)]
+
+## FILES
+
+Included in the default [`Gemfile(5)`][Gemfile(5)]
+generated is the line `# frozen_string_literal: true`. This is a magic comment
+supported for the first time in Ruby 2.3. The presence of this line
+results in all string literals in the file being implicitly frozen.
+Frozen string literals will be the default in Ruby 3.
 
 ## SEE ALSO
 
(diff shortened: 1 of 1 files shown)


man/bundle-init.ronn
```diff
@@ -12,7 +12,13 @@ adding a [`Gemfile(5)`][Gemfile(5)] to a gem with a gemspec, the `--gemspec` opt
 automatically add each dependency listed in the gemspec file to the newly
 created [`Gemfile(5)`][Gemfile(5)].
 
+One of the lines included in the default [`Gemfile(5)`][Gemfile(5)] generated is `# frozen_string_literal: true`. This is a magic comment supported for the first time in Ruby 2.3. The presence of this line results in all string literals in the file being implicitly frozen. Frozen string literals will be the default in Ruby 3.
```
colby-swandale: You need to split this line up to match the max width, which is about 80 characters. Checkout the other man pages to see what i'm talking about.

315tky:
Thanks, have corrected

man/bundle-init.ronn
315tky: Thanks, have corrected

rubymorillo:
@colby-swandale Hey Colby, do you think we can get this one merged? Thanks!

man/bundle-init.ronn
```diff
@@ -12,7 +12,17 @@ adding a [`Gemfile(5)`][Gemfile(5)] to a gem with a gemspec, the `--gemspec` opt
 automatically add each dependency listed in the gemspec file to the newly
 created [`Gemfile(5)`][Gemfile(5)].
 
+One of the lines included in the default [`Gemfile(5)`][Gemfile(5)]
```
colby-swandale: This should be in it's own section as we're no talking longer about the command itself.

colby-swandale: > in the default

This implies there's a way to have a Gemfile without the magic comment, which there isn't. How about `included in the generated Gemfile is...`?

man/bundle-init.ronn
```diff
@@ -16,3 +16,15 @@ created [`Gemfile(5)`][Gemfile(5)].
...
 * `--gemspec`:
   Use the specified .gemspec to create the [`Gemfile(5)`][Gemfile(5)]
+
+## FILES
+
+Included in the default [`Gemfile(5)`][Gemfile(5)]
+generated is the line `# frozen_string_literal: true`. This is a magic comment
+supported for the first time in Ruby 2.3. The presence of this line
+results in all string literals in the file being implicitly frozen.
+Frozen string literals will be the default in Ruby 3.
```
koic: AFAIK, Frozen string literal will be default in Ruby 3.0 or not yet determinable. Here `may be` seems more appropriate than `will be`.

indirect:
@bundlerbot r+

PR #6303, opened 2018-02-22
Correctly re-install extensions when running `pristine` for a git source

### LINKED ISSUE #6294: Running `bundle pristine` on a git source gem doesn't rebuild the native extension
I believe this can be fixed by adding the following before https://github.com/bundler/bundler/blob/v1.16.1/lib/bundler/cli/pristine.rb#L33:

```rb
FileUtils.rm_rf spec.extension_dir if spec.have_extensions?
```

I'm happy to send a PR if this sounds like the correct thing to do.

wagenet:
Unfortunately, I was unable to figure out how to prevent it from reloading
the built extension from the global cache in 2.0. That means this is 1.16
only for now. I will investigate more.

Resolves #6294

### DIFF
--- a/lib/bundler/cli/pristine.rb
+++ b/lib/bundler/cli/pristine.rb
@@ -30,6 +30,7 @@ module Bundler
           FileUtils.rm_rf spec.full_gem_path
         when Source::Git
           source.remote!
+          FileUtils.rm_rf spec.extension_dir if spec.have_extensions?
           FileUtils.rm_rf spec.full_gem_path
         else
           Bundler.ui.warn("Cannot pristine #{gem_name}. Gem is sourced from local path.")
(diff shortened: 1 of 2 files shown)


colby-swandale:
This should be getting merged into master, not `1-16-stable`

wagenet:
@colby-swandale I agree, but as you can see from my comment, I was unable to get it to work on master due to my inability to clear the entry from the global cache.

wagenet:
@colby-swandale the issue I'm seeing is that for git gems, instead or actually rebuilding, we're just loading from `bundler_extension_cache_path`. 

https://github.com/bundler/bundler/blob/master/lib/bundler/rubygems_gem_installer.rb#L22

I believe the solution would either be to have the `force` option be true or to delete the extension_cache entry. Do you have a suggestion on which is preferable?

wagenet:
Reopened in #6305 with a potential solution.

PR #6304, opened 2018-02-22
Fix `bundle gem` generating unparsable ruby

deivid-rodriguez:

The problem was that the main file in a gem generated by `bundle gem` can't be inspected via `rubocop`.

My diagnosis was `bundler` was generating unparsable ruby in the generated gem.

My fix was to change the offending template to generate valid ruby code.

I chose this fix because it's the only one, really.


colby-swandale:
Nice catch! Thanks! @bundlerbot r+

deivid-rodriguez:
Thanks @colby-swandale!

PR #6305, opened 2018-02-22
Correctly re-install extensions when running `pristine` for a git source

### LINKED ISSUE #6294: Running `bundle pristine` on a git source gem doesn't rebuild the native extension

wagenet:

I have a gem with a native extension that is installed via git. I had to recompile it due to some needed build arguments.

### The problem was...

Running `bundle pristine` would not recompile it as expected.

### My diagnosis was...

After digging into the source, I discovered that the built extension lived in a different location than the cloned git repo. `bundle pristine` was only removing the git repo so the built extension was not getting rebuilt.

### My fix...

Update `bundle pristine` to also remove the built extension. For 2.0, this also required removing the built extension cache. Without doing that, the built extension directory would just be recreated from the cache.

### I chose this fix because...

As far as I know, it's the only solution.

Resolves #6294

### DIFF
--- a/lib/bundler/cli/pristine.rb
+++ b/lib/bundler/cli/pristine.rb
@@ -30,6 +30,10 @@ module Bundler
           FileUtils.rm_rf spec.full_gem_path
         when Source::Git
           source.remote!
+          if extension_cache_path = source.extension_cache_path(spec)
+            FileUtils.rm_rf extension_cache_path
+          end
+          FileUtils.rm_rf spec.extension_dir
           FileUtils.rm_rf spec.full_gem_path
         else
           Bundler.ui.warn("Cannot pristine #{gem_name}. Gem is sourced from local path.")
--- a/lib/bundler/source.rb
+++ b/lib/bundler/source.rb
@@ -54,6 +54,15 @@ module Bundler
       instance_of?(Bundler::Source::Path)
     end
 
+    def extension_cache_path(spec)
+      return unless Bundler.feature_flag.global_gem_cache?
+      return unless source_slug = extension_cache_slug(spec)
+      Bundler.user_cache.join(
+        "extensions", Gem::Platform.local.to_s, Bundler.ruby_scope,
+        source_slug, spec.full_name
+      )
+    end
+
   private
 
     def version_color(spec_version, locked_spec_version)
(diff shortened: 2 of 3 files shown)


colby-swandale:
Can you please follow the [PR template](https://github.com/bundler/bundler/blob/master/.github/PULL_REQUEST_TEMPLATE.md) and put in the details that this PR is fixing. It's important because the PR description is used for the merge commit that we then use for things like releases.

wagenet:
@colby-swandale sure! I'll do that tomorrow.

wagenet:
@colby-swandale updated. Let me know if there's anything else I'm missing.

lib/bundler/cli/pristine.rb
```diff
@@ -30,6 +30,10 @@ def run
           FileUtils.rm_rf spec.full_gem_path
         when Source::Git
           source.remote!
+          if extension_cache_path = source.extension_cache_path(spec)
```
segiddins: should we be doing this for rubygems sources as well?

wagenet: @segiddins from what I could tell, the extension isn't being cached when installed via RubyGems. I may be mistaken, however, as most of my initial investigation was done on the 1.6 which doesn't have the cache.

indirect: I believe extension-caching is RubyGems only, so this seems right to me. 👍

indirect:
@bundlerbot r+

lib/bundler/cli/pristine.rb
indirect: Ignore me, that comment doesn't actually make sense.

indirect:
@bundlerbot r-

After looking at this some more, I think pristine will fail to recompile .gem extensions as well. We should at least be sure one way or the other before merging. 👍

wagenet:
@indirect can you elaborate on the uncovered scenario. The other scenario I could see is covered here: https://github.com/wagenet/bundler/blob/a936ef3761c1971055cc438180b43af4315f08f1/spec/commands/pristine_spec.rb#L162, but I'm guessing I'm misunderstanding.

colby-swandale:
ping @indirect

indirect:
@bundlerbot r+

indirect:
I think it looks right, sorry for the hassle, and thanks for the patch!

PR #6306, opened 2018-02-23
Don't GPG-sign test setup commits

alyssais:
Running `bin/rake spec` in the Bundler repository causes a bunch of failures if `commit.gpgsign` is set in the global Git configuration.

Bundler's tests were making Git commits and trying to GPG sign them.

Set `commit.gpgsign` to `false` when initializing a Git repository in the tests.

I chose this fix because the alternative would be to pass `--no-gpg-sign` every time a commit is made in the tests, which would have a bigger maintenance impact.


indirect:
Great catch, thank you!

colby-swandale:
Thanks! @bundlerbot r+

PR #6307, opened 2018-02-23
Improve yanked gem error message

alyssais:
The problem was that the error message for a yanked Gem told the user to:

> update your bundle to a different version of foo (1.0)

This is slightly misleading because in this example, foo 1.0 is the version that has been yanked. They should use any verion of foo **other** than 1.0.

My diagnosis was that the error message interpolated the gem specification (which converts to a string as "NAME (VERSION)") rather than just the name of the gem in this instance.

My fix was to change that occurrence of the gem specification to just the gem name.

I chose this fix because it was the only way that occurred to me to do it.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -175,7 +175,7 @@ module Bundler
           raise GemNotFound, "Your bundle is locked to #{locked_gem}, but that version could not " \
                              "be found in any of the sources listed in your Gemfile. If you haven't changed sources, " \
                              "that means the author of #{locked_gem} has removed it. You'll need to update your bundle " \
-                             "to a different version of #{locked_gem} that hasn't been removed in order to install."
+                             "to a different version of #{gem_name} that hasn't been removed in order to install."
         end
         unless specs["bundler"].any?
           bundler = sources.metadata_source.specs.search(Gem::Dependency.new("bundler", VERSION)).last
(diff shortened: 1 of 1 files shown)


lib/bundler/definition.rb
```diff
@@ -175,7 +175,7 @@ def specs
           raise GemNotFound, "Your bundle is locked to #{locked_gem}, but that version could not " \
                              "be found in any of the sources listed in your Gemfile. If you haven't changed sources, " \
                              "that means the author of #{locked_gem} has removed it. You'll need to update your bundle " \
-                             "to a different version of #{locked_gem} that hasn't been removed in order to install."
+                             "to a different version of #{gem_name} that hasn't been removed in order to install."
```
hmistry: Thank you for making this PR to correct the message. I like how you phrased in PR comment and prefer we change it to that. I prefer the gem version is specified in the call to action to ensure the user knows not to install that version especially if they're quickly skimming the message. So something like:
```
You'll need to update your bundle to a version other than #{locked_gem} that hasn't been removed in order to install.
```
What do you think?

colby-swandale: ping @alyssais

alyssais: That sounds good to me. I'll update the PR.

segiddins:
I'ms surprised this change doesnt break any tests...

indirect:
@bundlerbot r+

PR #6308, opened 2018-02-23
Better error message for `bundle install` git failure

deivid-rodriguez:

The problem was that I wouldn't manage to run `bundle install` and would keep getting the following error.

```
Fetching gem metadata from https://rubygems.org/.......
Fetching https://github.com/decidim/decidim
fatal: No se pudo analizar el objeto 'ff36d24dbe1cf7aa99e5eeea1ff38e6aeeb2af92'.
Git error: command `git reset --hard ff36d24dbe1cf7aa99e5eeea1ff38e6aeeb2af92` in directory
/home/deivid/.rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/bundler/gems/decidim-ff36d24dbe1c has failed.
'/home/deivid/.rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/cache/bundler/git/decidim-9495d3039168996748af12c0fdb04debdea1039
``` 

None of the advice in there would work. I left some `puts` around in `git_proxy.rb` but I wasn't able to figure out the problem. At some point, I tried to delete the lock file and rebundle and it worked!
 
You can see the log of the full session here:  https://gist.github.com/deivid-rodriguez/effda7c001b97522ab53fc18d7b6bf53.

My diagnosis was that the advice in the error message could be improved since the solution that worked for me is included.

Change the error message to include removing the `Gemfile.lock` file.

Because it was the one that worked for me.

### DIFF
--- a/lib/bundler/source/git/git_proxy.rb
+++ b/lib/bundler/source/git/git_proxy.rb
@@ -29,7 +29,9 @@ module Bundler
           msg = String.new
           msg << "Git error: command `git #{command}` in directory #{SharedHelpers.pwd} has failed."
           msg << "\n#{extra_info}" if extra_info
-          msg << "\nIf this error persists you could try removing the cache directory '#{path}'" if path && path.exist?
+          msg << "\nIf this error persists you could try the following:"
+          msg << "\n  * Removing the cache directory '#{path}'" if path && path.exist?
+          msg << "\n  * Removing your 'Gemfile.lock' (or 'gems.rb.locked') file"
           super msg
         end
       end


colby-swandale:
I'm unsure if removing the lockfile is advice that we want to be giving out to our users. @segiddins do you have any opinions on this?

deivid-rodriguez:
Are you concerned about the wording (without expliciting the rerun of `bundle install` again to regenerate the lock file, and with the chance of introducing local changes), or about the advice itself?

For what it's worth, in my case the problem was not a corrupted lockfile or anything like that. Removing the lock file and rebundling did the trick but didn't cause any changes in the lock file.

When I get time, I'd like to invest some time on reproducing the problem, it's related to the usage of `bundle config local.my_gem /my/gem`, and outdated sha1 references.

segiddins:
I would rather address the underlying failure than suggest removing the lockfile, if at all possible. In this particular case, I wonder if `bundle update gem_name` would suffice

deivid-rodriguez:
Next time I run into this I'll try that!

deivid-rodriguez:
Hei!

So I run into it again, and I have more clue about what's going on now. First thing, `bundle update gem_name` doesn't work either.

The problem happens when you reference a branch in your `Gemfile` and have your lock file locked to a specific revision in that branch. If history is modified in that branch so that it no longer includes that sha reference, you will get that error. The problem (and a solution similar to mine) is referenced here: https://semaphoreci.com/docs/fail-could-not-parse-object.html.

I tried to reproduce this with a test, but in my test the sha reference I use does not exist at all, so the error message is somewhat more clear:

```diff
@@ -1019,6 +1045,25 @@ RSpec.describe "bundle install with git sources" do
       bundle "install"
       expect(out).to include("Revision deadbeef does not exist in the repository")
     end
+
+    it "gives a helpful error when the locked ref no longer belongs to the branch" do
+      build_git "foo"
+      revision = revision_for(lib_path("foo-1.0"))
+
+      install_gemfile <<-G
+        gem "foo", :git => "file://#{lib_path("foo-1.0")}", :ref => "#{revision}"
+      G
+      bundle "install"
+      expect(out).to_not match(/Revision.*does not exist/)
+
+      lockfile = File.read(bundled_app("Gemfile.lock"))
+      File.open(bundled_app("Gemfile.lock"), "w") do |file|
+        file.puts lockfile.gsub(/revision: #{revision}/, "revision: deadbeef")
+      end
+
+      bundle "install"
+      expect(out).to include("Revision deadbeef does not exist in the repository")
+    end
   end
 
   describe "bundle install --deployment with git sources" do
```

The test fails but with a easier to figure out error:

```console
Fetching file:///home/deivid/Code/bundler/tmp/libs/foo-1.0
fatal: Not a valid object name deadbeef
fatal: ambiguous argument 'deadbeef': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
Git error: command `git reset --hard deadbeef` in directory /home/deivid/Code/bundler/tmp/bundled_app/.bundle/ruby/2.5.0/bundler/gems/foo-1.0-deadbeef has failed.
If this error persists you could try removing the cache directory '/home/deivid/Code/bundler/tmp/home/.bundle/cache/git/foo-1.0-418786a06b87262951e20e512049b395a833843c'
```

I also tried to reproduce the exact problem by force-pushing to a branch, but I couldn't figure it out. This is my attempt (that bundles just fine :angry:)

```diff
@@ -809,6 +809,27 @@ RSpec.describe "bundle install with git sources" do
     expect(the_bundle).to include_gems "forced 1.0"
   end
 
+  it "handles branches that have been force-pushed" do
+    build_git "forced", "1.0"
+
+    update_git "forced", branch: "my_branch" do |s|
+      s.write "lib/forced.rb", "FORCED = '1.1'"
+    end
+
+    install_gemfile <<-G
+      gem "forced", :git => "#{lib_path("forced-1.0")}", :branch => "my_branch"
+    G
+
+    bundle "install"
+
+    Dir.chdir(lib_path("forced-1.0")) do
+      sys_exec "git commit --amend -m 'Changed history' --allow-empty"
+    end
+
+    bundle "update install"
+    expect(out).to include("does not exist in the repository")
+  end
+
   it "ignores submodules if :submodule is not passed" do
     build_git "submodule", "1.0"
     build_git "has_submodule", "1.0" do |s|
```

I think one can reproduce this for real by cloning [decidim-module-votings](https://github.com/podemos-info/decidim-module-votings) at its [current master branch](https://github.com/podemos-info/decidim-module-votings/commit/ef92898f56053d631fccaeb4f80bfe7aa4e7359f).

Hopefully we can figure this out :)

deivid-rodriguez:
The last test was not working because I was not force-pushing to the remote at all :laughing:. I'm now learning more about the test suite so hopefully I'll be able to reproduce this with a test and eventually get it fixed :)

indirect:
@deivid-rodriguez this seems like a real bug! maybe you can open a new issue with a repro script, if you can't figure out how to manage a failing test?

indirect:
oh! I missed your last message, oops. opening a new PR with a working test sounds even better. 😅

segiddins:
Great investigative work!

deivid-rodriguez:
Thanks! I reticketed this anyways in #6324.

PR #6309, opened 2018-02-25
Add config variable and check for platform warnings

### LINKED ISSUE #6124: The ability to silence platform warnings
https://github.com/bundler/bundler/blob/b9c12ee67661581f8a4fb5f47a375b5c7da2ed1e/lib/bundler/definition.rb#L880-L884

There really isn't any reason to ship that warning to me, and it would be nice if I could silence it, as I am well aware of the platform I am on, and I always inspect Gemfiles before I install anything, regardless of whether it's my own project or not.

agrim123:

The user needed a way to turn off platform warnings.

Creating a config variable to solve the above problem.

Added a key `disable_platform_warnings` in settings and placed check at the relevant place to disable warnings.

We will by default show warnings but the user might want to disable them, so using a config variable looked a good option.

Fixes #6124

### DIFF
--- a/man/bundle-config.ronn
+++ b/man/bundle-config.ronn
@@ -167,7 +167,7 @@ learn more about their operation in [bundle install(1)][bundle-install(1)].
    instead of warnings.
    Use `bundle config --delete disable_multisource` to unset.
 * `disable_platform_warnings` (`BUNDLE_DISABLE_PLATFORM_WARNINGS`):
-   Whether Bundler should show platform warnings.
+   Disable warnings during bundle install when a dependency is unused on the current platform.
 * `disable_shared_gems` (`BUNDLE_DISABLE_SHARED_GEMS`):
    Stop Bundler from accessing gems installed to RubyGems' normal location.
 * `disable_version_check` (`BUNDLE_DISABLE_VERSION_CHECK`):
--- a/man/bundle-config.ronn
+++ b/man/bundle-config.ronn
@@ -166,6 +166,8 @@ learn more about their operation in [bundle install(1)][bundle-install(1)].
    When set, Gemfiles containing multiple sources will produce errors
    instead of warnings.
    Use `bundle config --delete disable_multisource` to unset.
+* `disable_platform_warnings` (`BUNDLE_DISABLE_PLATFORM_WARNINGS`):
+   Whether Bundler should show platform warnings.
 * `disable_shared_gems` (`BUNDLE_DISABLE_SHARED_GEMS`):
    Stop Bundler from accessing gems installed to RubyGems' normal location.
 * `disable_version_check` (`BUNDLE_DISABLE_VERSION_CHECK`):
--- a/spec/install/gemfile/platform_spec.rb
+++ b/spec/install/gemfile/platform_spec.rb
@@ -230,21 +230,46 @@ RSpec.describe "bundle install with platform conditionals" do
     expect(out).not_to match(/Could not find gem 'some_gem/)
   end
 
-  it "prints a helpful warning when a dependency is unused on any platform" do
-    simulate_platform "ruby"
-    simulate_ruby_engine "ruby"
+  context "when disable_platform_warnings is false (by default)" do
+    before { bundle! "config disable_platform_warnings false" }
 
-    gemfile <<-G
-      source "file://#{gem_repo1}"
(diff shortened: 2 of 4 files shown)


colby-swandale:
Thanks for helping make Bundler better! A couple of things that need to be done before we can merge this. This is going to need a test to make sure this is behaving as expected and so that we don't regress in the future, you will want to look at [this test file](https://github.com/bundler/bundler/blob/master/spec/install/gemfile/platform_spec.rb#L233). You will also need to document the new configuration option to the [bundle config man page](https://github.com/bundler/bundler/blob/master/man/bundle-config.ronn#L116)

spec/install/gemfile/platform_spec.rb
```diff
@@ -230,21 +230,46 @@
     expect(out).not_to match(/Could not find gem 'some_gem/)
   end
 
-  it "prints a helpful warning when a dependency is unused on any platform" do
-    simulate_platform "ruby"
-    simulate_ruby_engine "ruby"
+  context "when disable_platform_warnings is false (by default)" do
```
colby-swandale: This test doesn't need a `context` blocked wrapping around it as the setting is set in Bundler automatically.

spec/install/gemfile/platform_spec.rb
```diff
@@ -230,21 +230,46 @@
...
+      expect(out).to include <<-O.strip
+  The dependency #{Gem::Dependency.new("rack", ">= 0")} will be unused by any of the platforms Bundler is installing for. Bundler is installing for ruby but the dependency is only for x86-mingw32, x86-mswin32, x64-mingw32, java. To add those platforms to the bundle, run `bundle lock --add-platform x86-mingw32 x86-mswin32 x64-mingw32 java`.
+      O
+    end
+  end
+
+  context "when disable_platform_warnings is true" do
+    before { bundle! "config disable_platform_warnings true" }
+
+    it "doesnot print the warning when a dependency is unused on any platform" do
```
colby-swandale: `does not`

spec/install/gemfile/platform_spec.rb
```diff
@@ -230,21 +230,46 @@
...
+      gemfile <<-G
+        source "file://#{gem_repo1}"
+
+        gem "rack", :platform => [:mingw, :mswin, :x64_mingw, :jruby]
+      G
+
+      bundle! "install"
+
+      expect(out).not_to include <<-O.strip
+  The dependency #{Gem::Dependency.new("rack", ">= 0")} will be unused by any of the platforms Bundler is installing for. Bundler is installing for ruby but the dependency is only for x86-mingw32, x86-mswin32, x64-mingw32, java. To add those platforms to the bundle, run `bundle lock --add-platform x86-mingw32 x86-mswin32 x64-mingw32 java`.
```
colby-swandale: This is a lot to match exactly against in the output which could easily be missed if we change the test or behavior. I would match with:

    expect(out).to_not include(/The dependency (.*) will be unused/)

spec/install/gemfile/platform_spec.rb
agrim123: My bad!

spec/install/gemfile/platform_spec.rb
agrim123: Cool. Will change it.

colby-swandale: Yep, that was a typo on my part.

agrim123: Using 
```bash
expect(out).not_to match(/The dependency (.*) will be unused/)
```
does the trick. Is it good?

man/bundle-config.ronn
```diff
@@ -166,6 +166,8 @@ learn more about their operation in [bundle install(1)][bundle-install(1)].
    When set, Gemfiles containing multiple sources will produce errors
    instead of warnings.
    Use `bundle config --delete disable_multisource` to unset.
+* `disable_platform_warnings` (`BUNDLE_DISABLE_PLATFORM_WARNINGS`):
+   Whether Bundler should show platform warnings.
```
colby-swandale: I would reword this to be a bit more explicit, something like:

```
Disable warnings during bundle install when a dependency is unused on the current platform
```

agrim123: Done :smile:

colby-swandale:
Thanks! @bundlerbot r+

PR #6310, opened 2018-02-26
Fix some rescue calls that do not specifiy error type.

utilum:

The problem I noticed was in style, several instances of `rescue` clauses that do not specify an exception type. This is noted in in the Ruby Style Guide as [Avoid rescuing the Exception class](https://github.com/bbatsov/ruby-style-guide#no-blind-rescues).

My diagnosis was that at least some of those are leftover style. They are noted in `.rubocop_todo.yml`. To make sure, I asked on Slack.

My fix is to make as many of them more specific, specifying at least `StandardError`, and trying to be more specific.

No other ways of addressing this came to mind. I'd be happy to consider.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -745,7 +745,7 @@ module Bundler
       end
 
       Bundler.ui.warn "The latest bundler is #{latest}, but you are currently running #{current}.\n#{suggestion}"
-    rescue
+    rescue RuntimeError
       nil
     end
   end
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -189,7 +189,7 @@ module Bundler
         end
         tmp_home_path.join(login).tap(&:mkpath)
       end
-    rescue => e
+    rescue RuntimeError => e
       raise e.exception("#{warning}\nBundler also failed to create a temporary home directory at `#{path}':\n#{e}")
     end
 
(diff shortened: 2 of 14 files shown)


colby-swandale:
Hi, please fill out the [PR template](https://github.com/bundler/bundler/blob/master/.github/PULL_REQUEST_TEMPLATE.md).

lib/bundler/cli.rb
```diff
@@ -745,7 +745,7 @@ def warn_on_outdated_bundler
       end
 
       Bundler.ui.warn "The latest bundler is #{latest}, but you are currently running #{current}.\n#{suggestion}"
-    rescue
+    rescue RuntimeArgument
```
segiddins: What is “RuntimeArgument”?

utilum: Naught but a tyop. Fixed.

utilum:
> How did you decide between Runtime and Standard error?
I preferred `RuntimeError` as it is more specific than `StandardError, won't rescue a `SystemCallError` for example.
When using `RuntimeError` caused a failure of the tests and I was not certain which narrower `Exception` may need rescuing, I resorted to `StandardError`. Perhaps those cases are redundant? Would you have them removed?

indirect:
This looks good to me once the tests are green. 👍

colby-swandale:
ping @utilum

utilum:
Sorry for the delay.
I don't have Ruby 1.8.7 locally, and that's where the test failures occur. Just looking at the test log I wonder if the fix to this would not actually be in updating the tests... Anyone?

utilum:
Anyone?

segiddins:
@bundlerbot r+

PR #6314, opened 2018-03-01
Use Gem::Util.inflate instead of Gem.inflate

hsbt:

When  I released RubyGems 3.0, users can not use bundler with it.

bundler still uses deprecated methods with https://github.com/rubygems/rubygems/pull/2214

Replace `Gem.inflate` to `Gem::Util.inflate`


hsbt:
@bundlerbot r+

colby-swandale:
Was this reviewed was someone?

hsbt:
It's not required. This is explictly changes: https://github.com/rubygems/rubygems/pull/2214/files#diff-e2bd4eb9bbb8b25862825c636e55bd6dR640

PR #6315, opened 2018-03-01
Removed compatibility hack for old rubies.

hsbt:

Nothing.

Bundler will never use them in version 2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,6 +77,7 @@ jobs:
     # 1.x mode (we want to keep stuff passing in 1.x mode for now)
     - rvm: 2.6.0
       env: RGV=v2.7.7 BUNDLER_SPEC_SUB_VERSION=1.98
+      stage: test
 
   allow_failures:
     - rvm: ruby-head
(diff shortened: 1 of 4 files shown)


hsbt:
I wonder why we still test with Ruby 1.8

https://travis-ci.org/bundler/bundler/jobs/347596545

deivid-rodriguez:
I think @segiddins left it there to make backporting to 1.x branches easier for the time being. Not sure though.

segiddins:
Yes, it is to make it possible to be confident back porting to 1.x will not break things

indirect:
Since the 1.x series has been branched off of master for a while now, I think it makes sense to stop testing master against old rubies, and make that compatibility part of the backport work if/when backports are made. 👍

segiddins:
> make that compatibility part of the backport work if/when backports are made

At the very least, I'd like to keep them until 2.0 final is released (and we dont even have a prerelease out yet). Since we still need to support 1.16.x as the latest stable version

hsbt:
This is no rush. I wait a release day of 2.0.

deivid-rodriguez:
I don't really see the point of keeping 1.8 compatibility on master... I can see backports needing a bit of extra work if it's dropped but the alternative is "extra work by default".

hsbt:
@bundlerbot r+

PR #6318, opened 2018-03-01
Check correct paths are writable in requires_sudo?

jhawthorn:

Bundler was attempting to use `sudo`, despite having permissions to create `bundle_path` (which didn't yet exist).

For some reason this became an issue in recent versions of bundler, where it wasn't one before. I'm not sure what change caused this problem to be exposed since `requires_sudo?` has not changed in a while.

`requires_sudo?` checks that bundle_path, and directories that bundler may need to write within it are writable. If bundle_path itself does not exist, we instead find the nearest existent parent directory, and check that that exists.

Unfortunately, when these two rules were applied together, we got the wrong result. If `bundle_path` did not exist, bundler would check that the nearest parent directory **and everything within that parent directory** were writable. This could lead to false positives for `requires_sudo?` when `bundle_path` did not yet exist.

This commit fixes the issue by always checking the writability of `$bundle_path/*` instead of `$parent_path/*`.

If we are able to create `bundle_path`, we know that we can write to anything within it. Changing the test to use `Dir["$bundle_path/*"]` is a succinct way to implement this since it will return `[]` if `bundle_path` does not yet exist.

### DIFF
--- a/spec/bundler/bundler_spec.rb
+++ b/spec/bundler/bundler_spec.rb
@@ -2,6 +2,7 @@
 # frozen_string_literal: true
 
 require "bundler"
+require "tmpdir"
 
 RSpec.describe Bundler do
   describe "#load_gemspec_uncached" do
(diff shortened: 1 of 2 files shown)


segiddins:
Thanks for the PR! Could you please look into adding a test for this fix?

colby-swandale:
ping @jhawthorn

jhawthorn:
@segiddins @colby-swandale I finally got around to this.

There are now specs for the basic behaviour of `requires_sudo?` (there were none previously) and a regression spec exposing the issue this PR fixes.

spec/bundler/bundler_spec.rb
```diff
@@ -228,6 +228,74 @@
...
+      Bundler.send(:remove_instance_variable, :@requires_sudo_ran)
+      Bundler.send(:remove_instance_variable, :@requires_sudo)
+    end
+
+    before do
+      clear_cached_requires_sudo
+      allow(Bundler).to receive(:which).with("sudo").and_return("/usr/bin/sudo")
+
+      require "tmpdir"
+      @tmpdir = Dir.mktmpdir
```
segiddins: please use `let` instead of ivars

jhawthorn: done

segiddins:
Thanks!

@bundlerbot r+

PR #6319, opened 2018-03-02
Link out directly to contributor guidelines

rubymorillo:
Hey team,

The "Contributing" section confusingly sends contributors over to the "Documentation" README, and from there, contributors have to click on "Overview" to actually get to the guidelines. Since it's easier to cut out those two steps, I've rephrased a sentence and provide the direct link to the contributor guidelines here.

The problem was...
...that it wasn't clear why contributors had to complete two steps to get to the contributor guidelines. And because it isn't clear once they get to "Documentation" which link is the actual guideline, it's much simpler to link out to it directly from the Bundler README.

My diagnosis was...
....to cut out the middle man, so to speak.

My fix...

1. rephrase a sentence
2. link out directly to the contributor guidelines README

I chose this fix because...
...it's easy to implement and makes the most sense from the user's POV.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,8 @@ To get in touch with the Bundler core team and other Bundler users, please see [
 
 ### Contributing
 
-If you'd like to contribute to Bundler, that's awesome, and we <3 you. There's a guide to contributing to Bundler (both code and general help) over in [our documentation section](doc/README.md).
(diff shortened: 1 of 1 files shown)


hmistry:
@rubymorillo Good catch. There's some overlap with the content on doc/readme.md and contributing.md as both are a form of TOC to guide users to appropriate documentation. I haven't looked at other pages so maybe not... perhaps they can all be unified with a single hierarchy structure and funnel users accordingly.

colby-swandale:
@bundlerbot r+

PR #6326, opened 2018-03-06
Raise an error when attempting to set an expectation on nil

segiddins:

The problem was we could accidentally set an expectation on `nil`, leading to an RSpec warning

My diagnosis was we should just error when we do so


colby-swandale:
@bundlerbot r+

Issue #6312, opened 2018-02-27
Bundler attempts to load gems in "without" groups


folgerc:
Hi,

First some context: I'm using Rails and running into an issue where I'm trying to remove test dependencies from my production environment. The problem is that these dependencies are defined in their own 'test' group in my Gemfile [1], but they are still being loaded on startup. I know this because I get 'could not find gem X' when viewing the Rails site. I tried setting the 'without' setting but this seems to have no effect.

Looking further that appears to be because in this initialize block, the 'without' settings are not taken into account due to the direct call to expand_dependencies [2], where the code path that removes the 'without' groups is different [3].

Is this expected behavior? I'm fairly new to Ruby/Rails so would love to know this is my fault.

[1] gemfile contents
```ruby
group :test do
  # ... bunch of stuff
  # this is the test dependency I want to keep out
  gem 'teaspoon-jasmine'
end
```

[2] https://github.com/bundler/bundler/blob/master/lib/bundler/definition.rb#L123
[3] seems to start here https://github.com/bundler/bundler/blob/master/lib/bundler/definition.rb#L223

segiddins:
Please share the output of `bundle env`, thanks!

folgerc:
Here you go,

``` 
 report is Bundler 1.3.5
 Ruby 2.3.1 (2016-04-26 patchlevel 112) [x86_64-linux]
 Rubygems 2.5.1
 GEM_HOME /placeholder/ruby2.3.x/lib/ruby/gems/2.3.0
 GEM_PATH /placeholder/ruby2.3.x/lib/ruby/gems/2.3.0:/.gem/ruby/2.3.0

 Bundler settings
   without
     Set for your local app (/placeholder/rails-root/.bundle/config): "test"
   bin_path
     Set via BUNDLE_BIN_PATH: "/placeholder/ruby2.3.x/lib/ruby/gems/2.3.0/gems/bundler-1.3.5/bin/bundle"
   gemfile
     Set via BUNDLE_GEMFILE: "/placeholder/rails-root/Gemfile"

 Gemfile
 gem 'rails'
 gem 'haml'

 group :test do
   gem 'haml-rails'
   gem 'rspec-rails'
   gem 'simplecov'
   gem 'simplecov-cobertura'
   gem 'teaspoon-jasmine'
 end


 Gemfile.lock
 <No Gemfile.lock found>
```

segiddins:
[...]

PR #6328, opened 2018-03-07
Add a spec for bundle lock not downloading gems

segiddins:

The problem was we were unsure (in https://github.com/bundler/bundler/issues/6312) whether lock had this behavior.

My diagnosis was it did, and I came to that conclusion via this test!


colby-swandale:
@bundlerbot r+

PR #6329, opened 2018-03-07
Stop printing “failed to update” message when changing sources

### LINKED ISSUE #6325: `bundle update` shows incorrect message when updating git gems
If I have

```ruby
gem "decidim", git: "https://github.com/decidim/decidim"
```

And I run

```
bundle update decidim
```

And that results in `bundler` updating the locked sha1 reference in my lock file, I wouldn't expect `bundler` to display the "Bundler attempted to update decidim but its version stayed the same" message.

Is this a reasonable expectation?

segiddins:

The problem was `bundle update` shows incorrect message when updating git gems

Closes #6325 

My diagnosis was we need to check if the source has changed

My fix only prints the message when the source has not changed

Note that is does not yet fix the "updated the git sha" case, since the locked spec and the new spec actually share the same source object!

### DIFF
--- a/spec/commands/update_spec.rb
+++ b/spec/commands/update_spec.rb
@@ -473,6 +473,22 @@ RSpec.describe "bundle update in more complicated situations" do
     expect(the_bundle).to include_gems "thin 2.0", "rack 10.0", "rack-obama 1.0"
   end
 
+  it "will not warn when an explicitly updated git gem changes sha but not version" do
+    build_git "foo"
+
+    install_gemfile! <<-G
+      gem "foo", :git => '#{lib_path("foo-1.0")}'
+    G
+
+    update_git "foo" do |s|
+      s.write "lib/foo2.rb", "puts :foo2"
+    end
+
+    bundle! "update foo"
+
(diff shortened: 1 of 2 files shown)


spec/commands/update_spec.rb
```diff
@@ -421,6 +421,39 @@
     expect(the_bundle).to include_gems "thin 2.0", "rack 10.0", "rack-obama 1.0"
   end
 
+  it "will not warn when an explicitly updated git gem changes sha but not version" do
```
segiddins: this is the test that's failing, and I'm not 100% how to refactor to fix it

deivid-rodriguez: Fixed the test by saving the string representation of each source _before_ it's actually updated, so the comparison later works as expected.

deivid-rodriguez:
@segiddins What do you think about this approach?

indirect:
I'm happy with this :+1:

segiddins:
Heh, cheeky ;) I like it! Thanks so much for taking this up

@bundlerbot r+

Issue #6327, opened 2018-03-07
allow to lock ruby on major version


mathieujobin:
Hello

instead of doing `ruby '~> 2.3.4'` which `rvm` does not support without setuping custom aliases.
it would be nice and simpler to just set `ruby '2.3'`

but bundler complains the version don't match

`Your Ruby version is 2.3.6, but your Gemfile specified 2.3`

in my opinion, 2.3.x should match 2.3

thanks

segiddins:
Doing `~> 2.3.0` is what will allow any patch version of `2.3` to satisfy the requirement

mathieujobin:
@segiddins the creates an annoyance for people using `rvm` though. 
it will complain `~> 2.3.0` is not a known version

mathieujobin:
of course, the other solution would be to rvm to support it 

https://github.com/rvm/rvm/issues/3717

PR #6330, opened 2018-03-07
allow requirement 2.3 to match versions 2.3.x

mathieujobin:

suggested solution for #6327 (allow to lock ruby major version)
 
> instead of doing `ruby '~> 2.3.4'` which `rvm` does not support without setuping custom aliases.
> it would be nice and simpler to just set `ruby '2.3'`
> 
> but bundler complains the version don't match
>
> `Your Ruby version is 2.3.6, but your Gemfile specified 2.3`
>
> in my opinion, 2.3.x should match 2.3

version with patchlevel should match major version requirement

if the required version front match the other, than its only a patch-level difference
this should also support versions 2.3.6.1 to match 2.3.6

I ran the code through the debugger as deep as 
```
[242, 251] in /home/mathieu/.rvm/rubies/ruby-2.3.6/lib/ruby/site_ruby/2.3.0/rubygems/requirement.rb
   242: 
   243:   def satisfied_by? version
   244:     raise ArgumentError, "Need a Gem::Version: #{version.inspect}" unless
   245:       Gem::Version === version
   246:     # #28965: syck has a bug with unquoted '=' YAML.loading as YAML::DefaultKey
=> 247:     requirements.all? { |op, rv| (OPS[op] || OPS["="]).call version, rv }
   248:   end
```
and this appeared to be the right place to handle it, not break other version match than the ruby version and yep

### DIFF
--- a/lib/bundler/ruby_version.rb
+++ b/lib/bundler/ruby_version.rb
@@ -88,7 +88,7 @@ module Bundler
       raise ArgumentError, "Can only diff with a RubyVersion, not a #{other.class}" unless other.is_a?(RubyVersion)
       if engine != other.engine && @input_engine
         [:engine, engine, other.engine]
-      elsif versions.empty? || !matches?(versions, other.gem_version) || versions_string(other.versions).match(/^#{versions_string(versions)}/).nil?
+      elsif versions.empty? || (!matches?(versions, other.gem_version) && versions_string(other.versions).match(/^#{versions_string(versions)}/).nil?)
         [:version, versions_string(versions), versions_string(other.versions)]
       elsif @input_engine && !matches?(engine_versions, other.engine_gem_version)
         [:engine_version, versions_string(engine_versions), versions_string(other.engine_versions)]
--- a/lib/bundler/ruby_version.rb
+++ b/lib/bundler/ruby_version.rb
@@ -88,7 +88,7 @@ module Bundler
       raise ArgumentError, "Can only diff with a RubyVersion, not a #{other.class}" unless other.is_a?(RubyVersion)
       if engine != other.engine && @input_engine
         [:engine, engine, other.engine]
-      elsif versions.empty? || !matches?(versions, other.gem_version)
+      elsif versions.empty? || !matches?(versions, other.gem_version) || versions_string(other.versions).match(/^#{versions_string(versions)}/).nil?
         [:version, versions_string(versions), versions_string(other.versions)]
       elsif @input_engine && !matches?(engine_versions, other.engine_gem_version)
         [:engine_version, versions_string(engine_versions), versions_string(other.engine_versions)]


colby-swandale:
This needs a test to verify the expected behavior is working as expected.

colby-swandale:
Thanks for opening a pull request to help make Bundler better! There is a fundamental difference between specifying version requirements between RVM and Bundler. In Bundler, specifying `2.4` as a version requirement means exactly `2.4` and not `>= 2.4.0 && < 2.5.0`. Unfortunately we can't accept the PR as this would introduce behavior into Bundler that does not match our documentation and would cause confusion for our users.

Let us know if you have any questions.

mathieujobin:
except, that there are no such thing as "= ruby 2.4"

usmonster:
Hello! @colby-swandale, can you please consider reopening this, or welcoming the equivalent feature in a new PR?

> In Bundler, specifying `2.4` as a version requirement means exactly `2.4` and not `>= 2.4.0 && < 2.5.0`.

In a semver world, there is absolutely no reason that an unspecified patch version should _not_ mean exactly this. I actually cannot think of a valid use case where this would be unexpected or unwanted behavior, but please describe one if I'm missing something obvious.

> [...] this would introduce behavior into Bundler that does not match our documentation and would cause confusion for our users.

I strongly believe that the best solution would be simply to update the documentation and maybe advertise the change, as I suspect that the vast majority of bundler users would fully understand and actually welcome the improvement.

Finally, as the patch author noted, this improved version matching will only apply to ruby versions, and thus won't break any existing Gemfile.

In summary: This is a _safe change_, and it will address a _rather common use case_, so please reconsider merging it in! Thanks for your time and consideration. :)

indirect:
Hi @usmonster, thanks for taking the time to write up your thoughts. Unfortunately, making this change would _not_ be safe, because it would suddenly cause Gemfiles that previously required to Ruby 2.4.0 to now require to Ruby 2.4.x. As you can probably imagine, there are a lot of production users depending on Bundler to keep their Ruby versions set to exactly the same version they asked for. Suddenly allowing additional versions of Ruby where we didn't allow them before is enough to break some applications.

We do want to improve this, but because Bundler is used by literally millions of computers around the world, we have to make changes like this slowly and carefully, and only in major version bumps, so that users will be aware that their previously working application might potentially break.

The `master` branch is just now transitioning to target Bundler 3.x, which would be a good time to potentially ship a change to how Ruby version specifiers are handled. To make the transition as smooth as possible, I imagine the PR would need to do something like this:

1. Verify that Bundler supports version requirements for `ruby`, like "~> 2.4" or "~> 2.3.0".
2. Verify that the Bundler documentation covers those version requirements.
3. Add code to print a warning on Ruby versions that are pure `"X.Y"` asking users to update their version to either "~> X.Y" or "~> X.Y.0", depending on what they actually want.
4. Continue to treat pure "X.Y" as "= 2.4.0" for one major version, while printing the warning.
5. In the next major version, treat "X.Y" as "= X.Y".

I hope that helps!

usmonster:
Hi @indirect, thanks for your swift and thoughtful response! Indeed, it makes sense to roll out this change in a major version, especially if it's potentially breaking.

> [...] it would suddenly cause Gemfiles that previously required to Ruby 2.4.0 to now require to Ruby 2.4.x.

I guess you could say that the basis of my argument is that it's rather uncommon for a ruby version requirement to be given in the form "X.Y", as that is technically & semantically not a valid version, and thus would reasonably result in undefined behavior unless otherwise documented. I imagine most folks specify a real, valid version if it matters to them, though I could also imagine that some folks might rely on _observed_/undocumented behavior, and programmers are notoriously lazy... 😉Also, I have no real numbers on how many Gemfiles out there require a ruby "X.Y" instead of "X.Y.0" (though curious to know if you have any data?), so I trust your assessment, understand your concerns, and agree that your approach is reasonable.

For a PR targeting Bundler 3.x, you suggest something slightly different and more explicit than this one, i.e. fully supporting version operators and leaving no room for ambiguity. I like the principle of the idea, but I'd like some clarification. Given the currently-supported parameters, we likely won't be able to support conjunctions/multiple version specifiers (e.g. `">= 2.0.0 && < 2.1.6"`), and it would be essentially incompatible with `:patchlevel`.

To clarify your proposal, I can think of a few different possibilities for step 1:
1. Add support only for the _pessimistic_ version operator in the version specifier, e.g., `ruby "~> 2.5.1", ...`
2. Add support for _any_ version operator in the version specifier, e.g., `ruby "=> 2.5.1", ...`.
3. Add support for any version operator in _any number of_ version specifiers, e.g., `ruby ">= 2.5.1 && <= 2.5.3", ...` _or_ `ruby [">= 2.5.1", "<= 2.5.3"], ...`.

_Note: In all of these, there would be no way to exclude or set a version range limit based on `patchlevel`._

While option 1 is the simplest and addresses what is arguably most common use case, it feels misleading and incomplete—I'd actually still prefer the original PR's logic over it, i.e. treat `"X.Y"` as `"~> X.Y.0"`, and nothing more.

Option 2 seems rather half-baked and not sufficient for the more realistic & "healthy" use cases where one would want to specify both a lower and an upper bound.

Option 3 is the most powerful, but also the most complex and potentially disruptive.

Maybe these aren't the only options, but I'd love to hear your thoughts. Thanks again!

mathieujobin:
I don't think this change introduce any breaking changes for anyone using bundler today.

current `~> 2.x.y` remains unchanged, 

and current version of bundler does not understand what is `ruby '2.x'`

it just error out as such
`Your Ruby version is 2.5.3, but your Gemfile specified 2.5`

I think this only allow people to use it, existing Gemfile do not require to be changed, thus non-breaking

indirect:
@mathieujobin Today, we are using the `RubyGem::Requirement` class for both `ruby` and for `gem`. This change would make `ruby` a special and different case from those... can you explain why you think it is an improvement to add something new and undocumented and completely different to `ruby` that is not true for `gem`?

indirect:
After reading back through your PR and the linked previous issue, it sounds like the entire reason for this proposed change is because you want your Gemfile to contain a Ruby version number that RVM will accept. Bundler has allowed Ruby versions this way for a long time, and RVM started trying to read the Ruby version out of Gemfiles without asking or telling us about it.

I am not very convinced that Bundler should do something completely non-standard just because RVM decided to read the Gemfile Ruby version without actually supporting version specifiers the way Bundler does.

indirect:
@mathieujobin one last thing... it did hit me that if RVM wants to load a ruby version, it should be reading the ruby version from the Gemfile.lock, and not from the Gemfile. That way RVM has an exact version to work with, and your application gets run under the exact Ruby version it is locked to. That’s what I would recommend RVM do instead 👍🏻

usmonster:
> did you know that `ruby` already accepts version specifiers?

😶

No, I did not! I guess I should've tried. :) That said, I'd actually prefer using `~> X.Y.0` over `X.Y`, since it's more explicit. I'd also welcome dropping `:patchlevel`, and I'd support RVM implementing a change on their side to check Gemfile.lock before checking Gemfile.

mathieujobin:
Thanks, after all the time that has passed, I now see this only as a "nice to have" not really critical.
for deployed environment, its easy enough to select the right ruby via script, that you use rvm, rbenv or OS packages. 

the auto select that is more loose was more to eliminate a small annoyance for developers on their local setup.

I have been using `rvm alias` to overcome this personally

i think we can forget about this idea, at least on the bundler side. 

thank you for the time you spent on this

Issue #6270, opened 2018-01-24
`BUNDLE_GEMFILE` does not override `.bundle/config` path to Gemfile


joelvh:
It's possible to specify a development version of the Gemfile in `.bundle/config` (such as `Gemfile.dev`). However, when running something like `BUNDLE_GEMFILE=Gemfile bundle exec rake` will still use `Gemfile.dev`.

Shouldn't the environment variable override what's specified in `.bundle/config`?

joelvh:
Would love to submit a PR for this, but spent a lot of time trying to track down where the Gemfile path is determined and haven't been able to identify the proper fix yet....

colby-swandale:
This seems like a bug, thanks for the report! A PR would be awesome! I suggest looking at [SharedHelpers](https://github.com/bundler/bundler/blob/master/lib/bundler/shared_helpers.rb#L240) as you'll find most of the logic around finding the Gemfile there.

joelvh:
My finding at the moment is that running `BUNDLE_GEMFILE=Gemfile JOEL=yes bundle exec rails c` brings up the Rails console, and `ENV` has the proper value for `JOEL` but `BUNDLE_GEMFILE` contains the value of what's specified in `.bundle/config`. (In this case, it's reset to "Gemfile.dev" instead of "Gemfile" specified in the command.)

@colby-swandale do you know if/why `BUNDLE_GEMFILE` is overwritten with what's in `.bundle/config`?

joelvh:
@colby-swandale I see there's an "environment preserver" here, but still digging into things.

https://github.com/bundler/bundler/blob/2c059142d17df414ab6f06842cfbfd7321a3e33e/lib/bundler.rb#L21

joelvh:
@colby-swandale If I don't use `bundle exec` it works as expected. Maybe there's a conflict between `bundle exec` and Rails loading bundler internally?

joelvh:
Alright, so I've tested `bundle exec irb` and I see in `ENV` that the value I want to override what's specified in `.bundle/config` is overwritten and the value I want to use is in `BUNDLER_ORIG_BUNDLE_GEMFILE`

joelvh:
The real issue I have right now is `BUNDLE_GEMFILE=Gemfile bundle update` uses `Gemfile.dev` from `.bundle/config`. I'm not sure why the existing environment variables get backed up and replaced @colby-swandale ?

joelvh:
[...]

PR #6331, opened 2018-03-07
Add check for BUNDLE_GEMFILE in cli

agrim123:

The problem was `BUNDLE_GEMFILE` is not respected when `.bundle/config` specifies an alternate Gemfile. 

My diagnosis was bundler was copying the config variables and aways using them.

My fix was to check if the user has provided `BUNDLE_GEMFILE` option, then give priority to this env variable.

I chose this fix because checking and setting the gemfile variable earlier in the process is the best way to give precedence to env variable over config variable.

Addresses #6270


colby-swandale:
This needs a test to verify this behavior works as expected.

agrim123:
Sure. Where does the test reside for this?

lib/bundler/cli.rb
```diff
@@ -34,8 +34,14 @@ def initialize(*args)
       super
 
       custom_gemfile = options[:gemfile] || Bundler.settings[:gemfile]
+
+      if ENV["BUNDLE_GEMFILE"] && !ENV["BUNDLE_GEMFILE"].empty?
```
colby-swandale: We don't need to set `custom_gemfile` with `ENV["BUNDLE_GEMFILE"]`  if we're going to skip using it using `unless ENV["BUNDLE_GEMFILE"]`

spec/bundler/cli_spec.rb
```diff
@@ -51,6 +51,22 @@
     end
   end
 
+  context "when ENV['BUNDLE_GEMFILE'] is specified" do
+    before { create_file "Gemfile.dev" }
```
colby-swandale: You don't need to `create_file`, `gemfile` will create it for us.

colby-swandale:
The test in this PR will continue to pass without the fix. We need to make sure we have a failing test firstly.

lib/bundler/cli.rb
agrim123: Then we will have to move `Bundle.reset_paths!` outside if condition to run it in any case?

agrim123:
@colby-swandale 
The current function is
```bash
$ bundle install  # Uses gemfile specified in the config file
$ BUNDLE_GEMFILE=Gemfile bundle install  # Uses this specified file
$ BUNDLE_GEMFILE=  bundle install  # Searches for default gemfiles (Gemfile, gems.rb)
```
Is this our desired opertaion?

spec/bundler/cli_spec.rb
agrim123: Done

agrim123:
I tried something like this 
```ruby
before { Bundler.settings.set_local :gemfile, "Gemfile.dev" }
```
before the context block but still it is looking for `Gemfile` not `Gemfile.dev` in the first test. :cry:

agrim123:
@colby-swandale 
Observed something weird!

The test that I wrote first was 
```ruby
it "uses the value specified in config file" do
  bundle :install

  expect(the_bundle).to include_gems "rack 1.0.0"
do
```
was failing because when `Bundle.setup()` is called then it does not respect config variables, which resulted in test failures. But on modifying it to 
```ruby
it "uses the value specified in config file" do
  bundle :install
  bundle :list

  expect(out).to include "rack (1.0.0)"
end
```
works. This happens because `bundle :list` calls the [cli](https://github.com/bundler/bundler/blob/3f10a37c527ef7664ef08df441c704dd88111cc3/spec/support/helpers.rb#L150) which picks up the [config variable](https://github.com/bundler/bundler/blob/3f10a37c527ef7664ef08df441c704dd88111cc3/lib/bundler/cli.rb#L36).

While trying to fix the problem, I found this [find_gemfile](https://github.com/bundler/bundler/blob/3f10a37c527ef7664ef08df441c704dd88111cc3/lib/bundler/shared_helpers.rb#L240-L246) function, which does not look for `Bundler.settings`. 

So, I tried something like this (a way to give respect to config var)

```ruby
def find_gemfile(order_matters = false)
  given = ENV["BUNDLE_GEMFILE"]
  return given if given && !given.empty?

  # Check for gemfile config variable
  from_config = Bundler.settings[:gemfile]
  return from_config if from_config && !from_config.empty?
  names = gemfile_names
  names.reverse! if order_matters && Bundler.feature_flag.prefer_gems_rb?
  find_file(*names)
end
```

But, this resulted in `Stack level too deep (SystemStackError)` error.

```bash
An error occurred while loading ./spec/bundler/cli_spec.rb.
Failure/Error: gemfile = find_gemfile

SystemStackError:
  stack level too deep
# ./lib/bundler.rb:232:in `root'
# ./lib/bundler.rb:244:in `app_config_path'
# ./lib/bundler.rb:271:in `settings'
# ./lib/bundler.rb:232:in `root'
# ./lib/bundler.rb:244:in `app_config_path'
# ./lib/bundler.rb:271:in `settings'
# ./lib/bundler.rb:232:in `root'
# ./lib/bundler.rb:244:in `app_config_path'
# ./lib/bundler.rb:271:in `settings'
# ./lib/bundler.rb:232:in `root'
# ./lib/bundler.rb:244:in `app_config_path'
# ./lib/bundler.rb:271:in `settings'
# ./lib/bundler.rb:232:in `root'
# ./lib/bundler.rb:244:in `app_config_path'
# ./lib/bundler.rb:271:in `settings'
# ./lib/bundler.rb:232:in `root'
# ./lib/bundler.rb:244:in `app_config_path'
# ./lib/bundler.rb:271:in `settings'
# ./lib/bundler.rb:232:in `root'
# ./lib/bundler.rb:244:in `app_config_path'
# ./lib/bundler.rb:271:in `settings'
# ./lib/bundler.rb:232:in `root'
# ./lib/bundler.rb:244:in `app_config_path'
# ./lib/bundler.rb:271:in `settings'
# ./lib/bundler.rb:232:in `root'
# ./lib/bundler.rb:244:in `app_config_path'
# ./lib/bundler.rb:271:in `settings'
# ./lib/bundler.rb:232:in `root'
# ./lib/bundler.rb:244:in `app_config_path'
...
```

[`app_config_path`](https://github.com/bundler/bundler/blob/3f10a37c527ef7664ef08df441c704dd88111cc3/lib/bundler.rb#L240) seems to be calling [`root`](https://github.com/bundler/bundler/blob/3f10a37c527ef7664ef08df441c704dd88111cc3/lib/bundler.rb#L230) which calls [`root`](https://github.com/bundler/bundler/blob/3f10a37c527ef7664ef08df441c704dd88111cc3/lib/bundler/shared_helpers.rb#L26) in `shared_helpers` that has [call to that same function](https://github.com/bundler/bundler/blob/3f10a37c527ef7664ef08df441c704dd88111cc3/lib/bundler/shared_helpers.rb#L27) `find_gemile`. Thus resulting in infinite recursion.

This seems to a bug in bundler!

Any suggestions?

PR #6335, opened 2018-03-10
Use Bundler.rubygems.inflate instead of the Gem::Util method directly

segiddins:

The problem was Bundler was using RubyGems methods directly that may change their interface over time.

My diagnosis was  we should use `Bundler.rubygems` to encapsulate those method calls.


segiddins:
@bundlerbot r+

Issue #6289, opened 2018-02-05
Errno::ENOSPC: No space left on device - sendfile


Yvens16:
--- ERROR REPORT TEMPLATE -------------------------------------------------------
# Error Report

## Questions

Please fill out answers to these questions, it'll help us figure out
why things are going wrong.

- **What did you do?**

  I ran the command `/home/cabox/.rvm/gems/ruby-2.3.4@global/bin/bundle install`

- **What did you expect to happen?**

  I expected Bundler to...

- **What happened instead?**

  Instead, what happened was...

- **Have you tried any solutions posted on similar issues in our issue tracker, stack overflow, or google?**

  I tried...

- **Have you read our issues document, https://github.com/bundler/bundler/blob/master/doc/contributing/ISSUES.md?**

  ...

## Backtrace

```
Errno::ENOSPC: No space left on device - sendfile
  /home/cabox/.rvm/rubies/ruby-2.3.4/lib/ruby/2.3.0/fileutils.rb:1395:in `copy_stream'
  /home/cabox/.rvm/rubies/ruby-2.3.4/lib/ruby/2.3.0/fileutils.rb:1395:in `block (2 levels) in copy_file'
  /home/cabox/.rvm/rubies/ruby-2.3.4/lib/ruby/2.3.0/fileutils.rb:1394:in `open'
  /home/cabox/.rvm/rubies/ruby-2.3.4/lib/ruby/2.3.0/fileutils.rb:1394:in `block in copy_file'
  /home/cabox/.rvm/rubies/ruby-2.3.4/lib/ruby/2.3.0/fileutils.rb:1393:in `open'
  /home/cabox/.rvm/rubies/ruby-2.3.4/lib/ruby/2.3.0/fileutils.rb:1393:in `copy_file'
  /home/cabox/.rvm/rubies/ruby-2.3.4/lib/ruby/2.3.0/fileutils.rb:487:in `copy_file'
  /home/cabox/.rvm/rubies/ruby-2.3.4/lib/ruby/2.3.0/fileutils.rb:404:in `block in cp'
  /home/cabox/.rvm/rubies/ruby-2.3.4/lib/ruby/2.3.0/fileutils.rb:1572:in `block in fu_each_src_dest'
  /home/cabox/.rvm/rubies/ruby-2.3.4/lib/ruby/2.3.0/fileutils.rb:1588:in `fu_each_src_dest0'
[...]

PR #6336, opened 2018-03-10
[CompactIndexClient::Updater] Use filesystem_access when copying files

segiddins:

The problem was users could see the error template when they do not have write permissions to their temporary directory.

Closes https://github.com/bundler/bundler/issues/6289

My diagnosis was we need to use `SharedHelpers.filesystem_access` when writing files.

My fix wraps usage of `cp`


colby-swandale:
Thanks! @bundlerbot r+

Issue #6324, opened 2018-03-05
Problems resolving dead git references on a branch in lock file


deivid-rodriguez:
Sometimes when I have an outdated git reference on a branch in my lock file (because of force-pushing on the branch leaving the sha reference invalid), I start getting the following kind of errors:

```console
$ LC_ALL=C bundle
Fetching gem metadata from https://rubygems.org/.......
Fetching https://github.com/decidim/decidim
fatal: Could not parse object 'cb70aded58b9215a495f0509e49daef930eec478'.
Git error: command `git reset --hard cb70aded58b9215a495f0509e49daef930eec478` in directory
/home/deivid/.rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/bundler/gems/decidim-cb70aded58b9 has failed.
If this error persists you could try removing the cache directory
'/home/deivid/.rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/cache/bundler/git/decidim-9495d3039168996748af12c0fdb04debdea10392'
```

I can reliably reproduce this at the moment, but I'm not sure if it's specific to my environment or a consistent reproduction. The steps would be:

```console
git clone https://github.com/podemos-info/decidim-module-votings decidim-votings
git checkout ef92898f56053d631fccaeb4f80bfe7aa4e7359f
LC_ALL=C bundle
```

Could someone confirm that this is specific to my envirnoment?

deivid-rodriguez:
There's a related discussion in https://github.com/bundler/bundler/pull/6308, but I'm reticketing this to keep the discussion on its own issue!

segiddins:
I can make this give a slightly better error message, but you can (in the meantime) run `bundle update decidim --source decidim` or `bundle update decidim decidim-dev` or `bundle update --source decidim`

segiddins:
https://github.com/bundler/bundler/pull/6337 should help give a better error message!

deivid-rodriguez:
Thanks so much @segiddins! :muscle

colby-swandale:
This seems like this has been resolved. Closing.

PR #6337, opened 2018-03-11
[GitProxy] Fail gracefully when resetting to the given revision fails

segiddins:

See https://github.com/bundler/bundler/issues/6324 for context.

The problem was when a git gem referenced a branch, and you tried to install on a machine without the revision in the lockfile cached, Bundler would print the following error:

```
Git error: command `git reset --hard cb70aded58b9215a495f0509e49daef930eec478` in directory
/home/deivid/.rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/bundler/gems/decidim-cb70aded58b9 has failed.
'/home/deivid/.rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/cache/bundler/git/decidim-9495d3039168996748af12c0fdb04debdea10392'
```

The problem was that removing the cache directory doesn't help.

My diagnosis was we needed to print a `RevisionNotFound` error when that `git reset --hard` failed.

My fix rescues git command failures resulting from that `git reset --hard` and re-raises a more appropriate error message:

```
Revision cb70aded58b9215a495f0509e49daef930eec478 does not exist in the repository https://github.com/decidim/decidim. Maybe you misspelled it?
```

I chose this fix because it didn't involve bubbling information up from the git proxy to the definition. Ideally we'd re-rescue this error somewhere and suggest a `bundle update --source`, but that can be done in a separate PR.

### DIFF
--- a/lib/bundler/source/git/git_proxy.rb
+++ b/lib/bundler/source/git/git_proxy.rb
@@ -131,7 +131,12 @@ module Bundler
           # method 2
           SharedHelpers.chdir(destination) do
             git_retry %(fetch --force --quiet --tags "#{path}")
-            git "reset --hard #{@revision}"
+
+            begin
+              git "reset --hard #{@revision}"
+            rescue GitCommandError
+              raise MissingGitRevisionError.new(@revision, URICredentialsFilter.credential_filtered_uri(uri))
+            end
 
             if submodules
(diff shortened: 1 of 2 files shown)


colby-swandale:
Thanks! @bundlerbot r+

PR #6338, opened 2018-03-11
Add local git repository source option (`--local_git`) to plugin installation

### LINKED ISSUE #5445: Add option to `plugin install` for local plugins
I ran into this when working on [bundler-dependency_graph](https://github.com/kerrizor/bundler-dependency_graph) `bundle plugin` currently supports `--git` and `--source` (rubygems) options, which both expect a URL. I had mixed success providing it with a `file://` URL, but a full-fledged `--file` option that accepted a filepath would've sped things along.

saveriomiroddi:
Convenience option for installing plugins from a local repository, using the local path:

    bundle plugin install --file /path/to/bundler-dependency_graph bundler-dependency_graph

Closes #5445.

### DIFF
--- a/lib/bundler/cli/plugin.rb
+++ b/lib/bundler/cli/plugin.rb
@@ -5,7 +5,7 @@ module Bundler
   class CLI::Plugin < Thor
     desc "install PLUGINS", "Install the plugin from the source"
     long_desc <<-D
-      Install plugins either from the rubygems source provided (with --source option) or from a remote/git source provided (with respectively --git and --local-git options). If no sources are provided, it uses Gem.sources
+      Install plugins either from the rubygems source provided (with --source option) or from a remote/git source provided (with respectively --git and --local_git options). If no sources are provided, it uses Gem.sources
    D
     method_option "source", :type => :string, :default => nil, :banner =>
       "URL of the RubyGems source to fetch the plugin from"
@@ -13,7 +13,7 @@ module Bundler
       "The version of the plugin to fetch"
     method_option "git", :type => :string, :default => nil, :banner =>
       "URL of the git repo to fetch from"
-    method_option "local-git", :type => :string, :default => nil, :banner =>
+    method_option "local_git", :type => :string, :default => nil, :banner =>
       "Path of the local git repo to fetch from"
     method_option "branch", :type => :string, :default => nil, :banner =>
       "The git branch to checkout"
--- a/lib/bundler/cli/plugin.rb
+++ b/lib/bundler/cli/plugin.rb
@@ -5,7 +5,7 @@ module Bundler
   class CLI::Plugin < Thor
     desc "install PLUGINS", "Install the plugin from the source"
     long_desc <<-D
-      Install plugins either from the rubygems source provided (with --source option) or from a remote/git source provided (with respectively --git and --local_git options). If no sources are provided, it uses Gem.sources
+      Install plugins either from the rubygems source provided (with --source option) or from a git source provided with --git (for remote repos) or --local_git (for local repos). If no sources are provided, it uses Gem.sources
    D
     method_option "source", :type => :string, :default => nil, :banner =>
       "URL of the RubyGems source to fetch the plugin from"
--- a/lib/bundler/cli/plugin.rb
+++ b/lib/bundler/cli/plugin.rb
@@ -5,7 +5,13 @@ module Bundler
   class CLI::Plugin < Thor
     desc "install PLUGINS", "Install the plugin from the source"
     long_desc <<-D
-      Install plugins either from the rubygems source provided (with --source option) or from a git source provided with (--git option). If no sources are provided, it uses Gem.sources
+      Install plugins from any of the sources provided:
+
+      - from rubygems (`--source` option)
+      - from a remote git source (`--git` option)
+      - from a local git repo (`--file /path/to/repo` option)
+
+      If no sources are provided, it uses Gem.sources.
    D
     method_option "source", :type => :string, :default => nil, :banner =>
       "URL of the RubyGems source to fetch the plugin from"
@@ -13,6 +19,8 @@ module Bundler
       "The version of the plugin to fetch"
     method_option "git", :type => :string, :default => nil, :banner =>
       "URL of the git repo to fetch from"
+    method_option "file", :type => :string, :default => nil, :banner =>
+      "Path of the local git repo to fetch from"
     method_option "branch", :type => :string, :default => nil, :banner =>
       "The git branch to checkout"
     method_option "ref", :type => :string, :default => nil, :banner =>
--- a/lib/bundler/cli/plugin.rb
+++ b/lib/bundler/cli/plugin.rb
@@ -5,13 +5,7 @@ module Bundler
   class CLI::Plugin < Thor
     desc "install PLUGINS", "Install the plugin from the source"
     long_desc <<-D
-      Install plugins from any of the sources provided:
-
-      - from rubygems (`--source` option)
-      - from a remote git source (`--git` option)
-      - from a local git repo (`--local-git /path/to/repo` option)
-
-      If no sources are provided, it uses Gem.sources.
+      Install plugins either from the rubygems source provided (with --source option) or from a remote/git source provided (with respectively --git and --local-git options). If no sources are provided, it uses Gem.sources
    D
     method_option "source", :type => :string, :default => nil, :banner =>
       "URL of the RubyGems source to fetch the plugin from"
--- a/lib/bundler/cli/plugin.rb
+++ b/lib/bundler/cli/plugin.rb
@@ -9,7 +9,7 @@ module Bundler
 
       - from rubygems (`--source` option)
       - from a remote git source (`--git` option)
-      - from a local git repo (`--file /path/to/repo` option)
+      - from a local git repo (`--local-git /path/to/repo` option)
 
       If no sources are provided, it uses Gem.sources.
    D
@@ -19,7 +19,7 @@ module Bundler
       "The version of the plugin to fetch"
     method_option "git", :type => :string, :default => nil, :banner =>
       "URL of the git repo to fetch from"
-    method_option "file", :type => :string, :default => nil, :banner =>
+    method_option "local-git", :type => :string, :default => nil, :banner =>
       "Path of the local git repo to fetch from"
     method_option "branch", :type => :string, :default => nil, :banner =>
       "The git branch to checkout"
--- a/lib/bundler/plugin/installer.rb
+++ b/lib/bundler/plugin/installer.rb
@@ -43,11 +43,16 @@ module Bundler
 
     private
 
+      # Rubocop misunderstands the semantics of this method, assuming an `else` code block
+      # that doesn't actually exist. See https://github.com/bbatsov/rubocop/issues/5702.
+      #
+      # rubocop:disable Style/GuardClause
       def check_sources_consistency!(options)
         if options[:git] && options[:'local-git']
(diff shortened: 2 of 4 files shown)


saveriomiroddi:
Hello! Assuming there I've understood correctly when implementing the feature, one thing to define is if all the combination of parameters are accepted or not (ie. `--file` along with `--git`, which technically works, but doesn't make too much sense).

lib/bundler/plugin/installer.rb
```diff
@@ -38,40 +36,42 @@ def definition.lock(*); end
...
 
-      def install_git(names, version, options)
-        uri = options.delete(:git)
-        options["uri"] = uri
+      def extract_sources_from_options!(options)
+        sources = {}
 
+        sources[:git] = options.delete(:git) if options[:git]
+        sources[:file] = options.delete(:file) if options[:file]
+        sources[:source] = options.delete(:source) if options[:source]
```
kerrizor: Do we actually need to remove these from `options`?

lib/bundler/plugin/installer.rb
```diff
@@ -14,12 +14,10 @@ class Installer
       def install(names, options)
         version = options[:version] || [">= 0"]
         Bundler.settings.temporary(:lockfile_uses_separate_rubygems_sources => false, :disable_multisource => false) do
-          if options[:git]
-            install_git(names, version, options)
-          else
-            sources = options[:source] || Bundler.rubygems.sources
-            install_rubygems(names, version, sources)
-          end
+          sources = extract_sources_from_options!(options)
```
kerrizor: Existing code prioritizes `--git` over `--source` - are we preserving this? I'm not sure if anyone was relying on respecting `--git` over `--source` where they were using both, but it is a behavioral change to add multiple sources to the `SourceList`.

lib/bundler/plugin/installer.rb
kerrizor: ...looking at how `prepare_source_list` interacts with the `sources` hash you're building up here, I think you're ok populating it with `nil` so ` sources[:git] = options[:git]` would be legit, or even

```
def extract_sources_from_options!(options)
  sources = {}

  [:file, :git, :sources].each do |source|
    sources[source] = options[source]
  end

  sources
end
```

saveriomiroddi: > Do we actually need to remove these from options?

So! The rationale behind the removal is to pass whatever is needed downstream, to make it clearer when reading the calling class file (installer.rb) rather than checking the called one (git.rb). In particular, when I was reading the code, this structure would have made it a bit easier.

However I'm fairly neutral with both philosophies (it's still easy to go downstream and see what's used); I'll apply this change :+1:.

lib/bundler/plugin/installer.rb
saveriomiroddi: So! The rationale behind having a separate refactoring commit (e496b30b75) was in fact to make it clear what was happening. Such refactoring commit did not change the behavior.

This is what was happening before:

```ruby
def install_git(names, version, options)
  # [some stuff]

  source_list = SourceList.new
  source_list.add_git_source(options)

  # To support both sources
  if options[:source]
    source_list.add_rubygems_source("remotes" => options[:source])
  end

  # [some stuff]
end
```

the post-refactoring code is clearer (I hope :laughing:):

saveriomiroddi: (note: I've edited the comment due to a misplaced Ctrl+Enter :smile:)

saveriomiroddi:
@kerrizor I've laid down the plan details (see updated PR description); once point 1. is confirmed (and consequently point 2.), I'll apply the rest.

lib/bundler/cli/plugin.rb
```diff
@@ -5,14 +5,22 @@ module Bundler
...
+      If no sources are provided, it uses Gem.sources.
    D
     method_option "source", :type => :string, :default => nil, :banner =>
       "URL of the RubyGems source to fetch the plugin from"
     method_option "version", :type => :string, :default => nil, :banner =>
       "The version of the plugin to fetch"
     method_option "git", :type => :string, :default => nil, :banner =>
       "URL of the git repo to fetch from"
+    method_option "file", :type => :string, :default => nil, :banner =>
+      "Path of the local git repo to fetch from"
```
segiddins: I think it's confusing we use `--file` to point to a local git repo instead of just a path on disk, maybe call it something like `--local-git` ?

saveriomiroddi:
@kerrizor  @segiddins  so, I've completed this iteration of the PR.

I've updated the refactor strategy, for two reasons:

- `:extract_sources_from_options!` is not really needed after kerry's review note
- the (mocked) units tests are quite invasive and stiff (they stub private methods, and they also cause obscure errors on failure), so I've opted for making the refactoring compatible with the logic

I've also made `--git` and `--local-git` mutually exclusive - I didn't receive any explicite directive on this, so I'm assuming it's a sensible choice (easy to revert; it's self-contained in a single commit).

I would have liked to have the options help long_description in multiple lines, but based on what I see from the code/documentation/source, this is not possible (except for text separated by multiple blank lines, which is not the case here).

I've encountered what I think is a Rubocop bug, which I had to workaround.

lib/bundler/plugin/installer.rb
```diff
@@ -38,22 +43,29 @@ def definition.lock(*); end
 
     private
 
+      # Rubocop misunderstands the semantics of this method, assuming an `else` code block
+      # that doesn't exist. See https://github.com/bbatsov/rubocop/issues/5702.
+      #
+      # rubocop:disable Style/GuardClause
```
kerrizor: Quality comment 👍

lib/bundler/cli/plugin.rb
```diff
@@ -5,14 +5,16 @@ module Bundler
...
-      Install plugins either from the rubygems source provided (with --source option) or from a git source provided with (--git option). If no sources are provided, it uses Gem.sources
+      Install plugins either from the rubygems source provided (with --source option) or from a remote/git source provided (with respectively --git and --local-git options). If no sources are provided, it uses Gem.sources
    D
     method_option "source", :type => :string, :default => nil, :banner =>
       "URL of the RubyGems source to fetch the plugin from"
     method_option "version", :type => :string, :default => nil, :banner =>
       "The version of the plugin to fetch"
     method_option "git", :type => :string, :default => nil, :banner =>
       "URL of the git repo to fetch from"
+    method_option "local-git", :type => :string, :default => nil, :banner =>
```
kerrizor: One thought: if we made this instead `local_git` we can avoid the quotes in `options[:'local-hash']`

lib/bundler/plugin/installer.rb
```diff
@@ -38,22 +43,29 @@ def definition.lock(*); end
...
+      def check_sources_consistency!(options)
+        if options[:git] && options[:'local-git']
```
kerrizor: I believe it would be slightly safer to use `options.has_key? :git && options.has_key? :local_git` here.. someone **shouldn't** do something like `bundle install --git=false` (which would be a string, regardless..) but Just In Case we get a boolean inserted into these further up the chain..

lib/bundler/plugin/installer.rb
```diff
@@ -38,22 +43,29 @@ def definition.lock(*); end
...
-        if options[:source]
-          source_list.add_rubygems_source("remotes" => options[:source])
-        end
+        install_all_sources(names, version, options, options[:source])
+      end
 
-        deps = names.map {|name| Dependency.new name, version }
+      def install_local_git(names, version, options)
+        uri = options.delete(:'local-git')
+        options["uri"] = uri
```
kerrizor: I like to not mix the "case" of my hashes - less surprise to have them all the same. This change is for the downstream code on l87, correct?

kerrizor: (I was thinking "oh just use `stringify_keys`" but that's a Rails thing.. 🙄 )

saveriomiroddi: I think this can't be changed in isolation :scream:; following the code path, it leads to https://github.com/bundler/bundler/blob/77dbd12d0759ca635e3cb5cb0e90840cdac7f0d0/lib/bundler/source/git.rb#L23

lib/bundler/plugin/installer.rb
saveriomiroddi: Addressed. Note that `:options` key checking was previously done via direct access, so this is a effectively a minor refactoring/improvement, so depending on the project guidelines on refactoring, I could also apply the same to all the other checks in the file - let me know!

Also, the project Rubocop configuration doesn't accept `has_key?` - it requires `key?`, which I've used.

lib/bundler/cli/plugin.rb
saveriomiroddi: The rationale for using `local-git` rather than `local_git` was that, in absence of Thor cosmetic automations (I'm not aware of any way to automatically convert hyphen to underscore, aside aliasing), between presenting a non-standard user-facing style (Unix parameters typically use hyphen) and a non-standard programmtic style, the priority was for the user.

Note that at a brief search (`method_option.*_`), I can't find any other method using an underscore style (eg. see `lib/bundler/cli.rb`).

I've pushed the change, anyway :smile:.

saveriomiroddi:
Addressed the last batch of review notes.

saveriomiroddi:
@kerrizor hello! is there anything else I need to do? :smiley:

lib/bundler/plugin/installer.rb
kerrizor: Ah, ok.. well, shoot, that's a pain. It might behoove us to do some work to clean up inconsistencies in the codebase as a whole.. but for now, 👍

lib/bundler/cli/plugin.rb
```diff
@@ -5,14 +5,16 @@ module Bundler
   class CLI::Plugin < Thor
     desc "install PLUGINS", "Install the plugin from the source"
     long_desc <<-D
-      Install plugins either from the rubygems source provided (with --source option) or from a git source provided with (--git option). If no sources are provided, it uses Gem.sources
+      Install plugins either from the rubygems source provided (with --source option) or from a remote/git source provided (with respectively --git and --local_git options). If no sources are provided, it uses Gem.sources
```
kerrizor: Just rereading this and I wonder if we can make it clearer. What do you think of:

```
Install plugins either from the rubygems source provided (with --source option) or from a git source provided with --git (for remote repos) or --local_git (for local repos.) If no sources are provided, it uses Gem.sources
```

Too long?

lib/bundler/plugin/installer.rb
```diff
@@ -12,10 +12,15 @@ class Installer
...
       def install(names, options)
+        check_sources_consistency!(options)
+
         version = options[:version] || [">= 0"]
+
         Bundler.settings.temporary(:lockfile_uses_separate_rubygems_sources => false, :disable_multisource => false) do
           if options[:git]
             install_git(names, version, options)
+          elsif options[:local_git]
+            install_local_git(names, version, options)
```
kerrizor: For my own future reference, the appearance of an `elsif` is the start of a smell for me, to replace with a `case/when` but this looks ok for now.

lib/bundler/cli/plugin.rb
saveriomiroddi: :+1: for me.

saveriomiroddi:
@kerrizor ok, all the review notes has been applied, and the build is green.

kerrizor:
@saveriomiroddi looking good.

@segiddins any other feedback?

saveriomiroddi:
Closing for inactivity. Whoever wants to follow up with the work, feel free to rip the code :smile:

PR #6346, opened 2018-03-15
Explain spec files and output a link to bundler.io guides

### LINKED ISSUE #6246: Gemspec template should explain how `spec.files` works
I think it would be useful if the gemspec created by `bundle gem <gemname>` had some comments explaining how the `spec.files` configuration works.

I recently created my first gem, and the default files config based on shelling out to `git ls-files -z` was opaque to me... until I spent quite a while trying to figure out why a new file I added to the gem wasn't being copied over when I ran `rake install`. (The answer was that I had not yet added the file to git... which I hadn't done because the gem wasn't working as intended... which was because I hadn't added it to git!)

Once I realized that the installed gem will include all the files in an array assigned to `spec.files` in the gemspec, it was relatively easy to figure out what the default git-based configuration was doing, and why it made sense as one option for defining the gem's files.

Adding some comments explaining what `git ls-files -z` is doing, and how `spec.files` works to define the actual code included in the installed gem, would probably be helpful for other newbie gem builders. It looks like this would just require adding some comments to `newgem.gemspec.tt`. I'd gladly draft a PR if this seems like a good idea.

nesaulov:
Resolves #6246 

1. There is no helpful comment about `spec.files` method in a new gem's gemspec.
2. `$ bundle gem gem_name` is missing link to further reading about developing, testing and releasing the newly created gem.

1. Add a comment in the `newgem.gemspec.tt`
2. Add a message after new gem generation with link to [bundler.io guides](https://bundler.io/guides/creating_gem.html)

Example: 
```bash
๛  ~/code/oss/tmp $  ../bundler/bin/bundle gem new_gem
Creating gem 'new_gem'...
      create  new_gem/Gemfile
      create  new_gem/lib/new_gem.rb
      create  new_gem/lib/new_gem/version.rb
      create  new_gem/new_gem.gemspec
      create  new_gem/Rakefile
      create  new_gem/README.md
      create  new_gem/bin/console
      create  new_gem/bin/setup
      create  new_gem/.gitignore
Initializing git repo in /Users/nesaulov/code/oss/tmp/new_gem
Gem 'new_gem' was successfully created. For detailed information on further steps please visit https://bundler.io/guides/creating_gem.html
๛  ~/code/oss/tmp $
```

### DIFF
--- a/lib/bundler/cli/gem.rb
+++ b/lib/bundler/cli/gem.rb
@@ -160,8 +160,7 @@ module Bundler
       open_editor(options["edit"], target.join("#{name}.gemspec")) if options[:edit]
 
       Bundler.ui.info "Gem '#{name}' was successfully created. " \
-        "For detailed information on further steps please visit https://bundler.io/guides/creating_gem.html"
-
+        "For detailed information on further steps pleaseg visit https://bundler.io/guides/creating_gem.html"
     rescue Errno::EEXIST => e
       raise GenericSystemCallError.new(e, "There was a conflict while creating the new gem.")
     end
--- a/lib/bundler/cli/gem.rb
+++ b/lib/bundler/cli/gem.rb
@@ -160,7 +160,7 @@ module Bundler
       open_editor(options["edit"], target.join("#{name}.gemspec")) if options[:edit]
 
       Bundler.ui.info "Gem '#{name}' was successfully created. " \
-        "For detailed information on further steps please visit https://bundler.io/guides/creating_gem.html"
+        "For more information on making a RubyGem visit https://bundler.io/guides/creating_gem.html"
     rescue Errno::EEXIST => e
       raise GenericSystemCallError.new(e, "There was a conflict while creating the new gem.")
     end
--- a/lib/bundler/cli/gem.rb
+++ b/lib/bundler/cli/gem.rb
@@ -158,6 +158,10 @@ module Bundler
 
       # Open gemspec in editor
       open_editor(options["edit"], target.join("#{name}.gemspec")) if options[:edit]
+
+      Bundler.ui.info "Gem '#{name}' was successfully created. " \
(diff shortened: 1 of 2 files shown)


ragesoss:
I think it would be help to also add a comment explaining the default for spec.files. Here's what I had:

```
# The `git ls-files -z` default will base this on all the files checked in to git. 
```

nesaulov:
@ragesoss yeah, you are right, I'll add this.

lib/bundler/cli/gem.rb
```diff
@@ -158,6 +158,9 @@ def run
 
       # Open gemspec in editor
       open_editor(options["edit"], target.join("#{name}.gemspec")) if options[:edit]
+
+      Bundler.ui.info "Gem '#{name}' was successfully created. " \
+        "For detailed information on further steps please visit https://bundler.io/guides/creating_gem.html"
```
colby-swandale: > For detailed information on further steps please visit https://bundler.io/guides/creating_gem.html`

I would rewrite this as: 
```
For more information on making a RubyGem visit https://bundler.io/guides/creating_gem.html
```

lib/bundler/templates/newgem/newgem.gemspec.tt
```diff
@@ -28,6 +28,8 @@ Gem::Specification.new do |spec|
       "public gem pushes."
   end
 
+  # Specify which files should be added to the gem when it is released.
+  # The `git ls-files -z` default will base this on all the files checked in to git.
```
colby-swandale: > The `git ls-files -z` default will base this on all the files checked in to git.

This doesn't read that well when i say it out loud. How about we change this to something like:

```
The `git ls-files -z` loads the files in the RubyGem that have been added into git
```

?

nesaulov: sure

segiddins:
Thank you!
@bundlerbot r+

PR #6348, opened 2018-03-16
Update brew install for groff

kunruh9:
[dupes](https://github.com/Homebrew/homebrew-dupes) tap is deprecated, `groff` is now in [hombrew-core](https://github.com/Homebrew/homebrew-core/blob/master/Formula/groff.rb).

Also fixed a missing single quote while I was at it :)


hsbt:
Thanks!

hsbt:
@bundlerbot r+

Issue #6190, opened 2017-11-27
eyaml failure when STDERR closed


gepandz:
# Error Report

## Questions

Please fill out answers to these questions, it'll help us figure out
why things are going wrong.

- **What did you do?**

  I ran the command `( echo $ENCRSTRING | /home/myid/.rbenv/versions/2.3.1/bin/eyaml --stdin --output string --pkcs7-public-key=./public_key.pkcs7.pem --quiet ) 2>&-`

- **What did you expect to happen?**

  I expected Bundler to print the encrypted string without the "[hiera-eyaml-core] Loaded config from ..." line. I would appreciate if `--quiet` actually suppressed this, as that information is not useful 99% of the time.

- **What happened instead?**

  Instead, what happened was a Ruby error: "Errno::EBADF: Bad file descriptor @ io_write - <STDERR>" and this template (thank you, for this, btw), which tells me that Ruby, or this gem, doesn't react well if you close STDERR because you're uninterested in the error output. This is common behavior on Linux and Unix systems, so this error is inappropriate.

- **Have you tried any solutions posted on similar issues in our issue tracker, stack overflow, or google?**

  I tried Google, piping into `grep -v "[hiera-eyaml-core] Loaded config from"`, and a couple of other things, none of which were ... elegant.

- **Have you read our issues document, https://github.com/bundler/bundler/blob/master/doc/contributing/ISSUES.md?**

  yes

## Backtrace

```
Errno::EBADF: Bad file descriptor @ io_write - <STDERR>
  /home/myid/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/bundler-1.15.3/lib/bundler/cli/exec.rb:81:in `write'
  /home/myid/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/bundler-1.15.3/lib/bundler/cli/exec.rb:81:in `abort'
  /home/myid/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/bundler-1.15.3/lib/bundler/cli/exec.rb:81:in `rescue in kernel_load'
  /home/myid/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/bundler-1.15.3/lib/bundler/cli/exec.rb:65:in `kernel_load'
  /home/myid/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/bundler-1.15.3/lib/bundler/cli/exec.rb:27:in `run'
[...]

PR #6349, opened 2018-03-16
Fix to check if stderr was closed before writing to it

shime:

See https://github.com/bundler/bundler/issues/6190

The error is happening because we're not checking if stderr is closed prior to writing to it.

My fix is to add a check to determine if stderr was closed.

I chose to fix it here for a start and discuss with others if more checks were needed. Maybe we need to check if stderr is closed for all references of `$stderr.puts` and replace `abort` with `$stderr.puts` and `exit(1)`?

### DIFF
--- a/spec/bundler/ui/shell_spec.rb
+++ b/spec/bundler/ui/shell_spec.rb
@@ -59,6 +59,15 @@ RSpec.describe Bundler::UI::Shell do
       it "prints to stderr" do
         expect { subject.error("error!!!") }.to output("error!!!\n").to_stderr
       end
+
+      context "when stderr is closed" do
+        it "doesn't report anything" do
+          output = capture(:stderr, :closed => true) do
+            subject.error("Something went wrong")
+          end
+          expect(output).to_not eq("Something went wrong\n")
+        end
+      end
     end
   end
 end
--- a/lib/bundler/ui/shell.rb
+++ b/lib/bundler/ui/shell.rb
@@ -109,6 +109,8 @@ module Bundler
       end
 
       def tell_err(message, color = nil, newline = nil)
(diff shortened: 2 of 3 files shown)


colby-swandale:
It would be better to have the tests in `spec/other/shell_error_spec.rb` instead be in `spec/bundler/ui/shell_spec.rb`, which is where we have our tests for testing output to the user.

spec/bundler/ui/shell_spec.rb
```diff
@@ -61,4 +61,15 @@
       end
     end
   end
+
+  describe "#tell_err" do
+    context "when stderr is closed" do
+      it "doesn't report anything" do
+        output = capture(:stderr, closed: true) do
+          subject.send(:tell_err, "Something went wrong")
```
colby-swandale: I wouldn't be testing a private method, instead test against the public API, so the `#error` method in this case.

(make sure to enable `Bundler.settings.temporary(:error_on_stderr => true)`) because `Bundler::UI::Shell` does not print to stderr.

colby-swandale:
This PR doesn't fix #6190 but it will be good to have it in anyway for bundler 2.

agrim123:
I guess changing [this](https://github.com/bundler/bundler/blob/7f4c1a81b7b92e234fce0e3592d27c396d2451ac/lib/bundler/cli/exec.rb#L81) line to 
```ruby
abort "#{e.class}: #{e.message}\n  #{backtrace.join("\n  ")}" unless STDERR.closed?
```
may be the fix.

segiddins:
This looks excellent (even though it doesn't fix the specific situation in #6190), thank you!

@bundlerbot r+

Issue #6350, opened 2018-03-17
Bundle update <gem> raises a RuntimeError for non-local platform gems


greysteil:
Given a Gemfile and Gemfile.lock for a JRuby project, like the below, running `bundle update json` will raise a `RuntimeError` saying `Unable to find a spec...`.
```
# Gemfile
source 'https://rubygems.org'

gem "json"
```
```
# Gemfile.lock
GEM
  remote: https://rubygems.org/
  specs:
    json (1.4.3-java)

PLATFORMS
  java

DEPENDENCIES
  json

BUNDLED WITH
   1.16.1
```

The cause of the above is [these lines](https://github.com/bundler/bundler/blob/master/lib/bundler/definition.rb#L122-L124) which call `expand_dependencies` on the list passed gem names without allowing them to be expanded to non-local platforms, and then pass the returned (ruby-only) list `SpecSet#for` with a `java` platform dependency.

I *think* the fix is to change those lines to:

```
unless @unlock[:lock_shared_dependencies]
    eager_unlock = expand_dependencies(@unlock[:gems], true)
    @unlock[:gems] = @locked_specs.for(eager_unlock, [], false, false, false).map(&:name)
end
```

(I.e., pass `remote=true` to the dependency expansion, so we don't exclude any platforms, and pass `raise_on_missing=false` to `for`, so that the missing `ruby` platform isn't an issue.)

However, I know very little about different Ruby platforms, Bundler's handling of them, or where this should be specced. I'd love someone to take it over.

segiddins:
Tests:

```diff
diff --git a/spec/commands/update_spec.rb b/spec/commands/update_spec.rb
index 799aac1f5..b5938b909 100644
--- a/spec/commands/update_spec.rb
+++ b/spec/commands/update_spec.rb
@@ -437,6 +437,47 @@ RSpec.describe "bundle update in more complicated situations" do
     bundle "update"
     expect(the_bundle).to include_gems "thin 1.0"
   end
+
+  context "when the lockfile is for a different platform" do
+    before do
+      build_repo4 do
+        build_gem("a", "0.9")
+        build_gem("a", "0.9") {|s| s.platform = "java" }
+        build_gem("a", "1.1")
+        build_gem("a", "1.1") {|s| s.platform = "java" }
+      end
+
+      gemfile <<-G
+        source "file://#{gem_repo4}"
+        gem "a"
+      G
+
+      lockfile <<-L
+        GEM
+          remote: file://#{gem_repo4}
[...]

PR #6444, opened 2018-03-18
Handle updating a specific gem for a non-local platform

greysteil:
Fixes https://github.com/bundler/bundler/issues/6350.

Kudos to @segiddins for the test.


segiddins:
This is the exact change I was testing last night 😅 something something great minds, right?

@bundlerbot r+

PR #6446, opened 2018-03-19
[SpecGroup] Remove unused ivar

segiddins:

The problem was this ivar was unused.

My diagnosis was we should remove it. For performance, you know.


colby-swandale:
Thanks! @bundlerbot r+

PR #6447, opened 2018-03-19
Update error message on bundle add in case of duplication

### LINKED ISSUE #6341: Inappropriate error message on running `bundle add [gem]` in case of duplication
I observed `bundle add rake` throws an error if I have already specified gem before

```bash
[!] There was an error parsing `injected gems`: You cannot specify the same gem twice with different version requirements.
You specified: rake (~> 12.3) and rake (>= 0). Bundler cannot continue.

 #  from injected gems:1
 >  gem "rake", ">= 0"
```
I think an appropriate error would be to tell the user that gem is installed and can be updated by running `bundle update rake`.

agrim123:

If the user tries to add a gem giving version requirement then bundler throws an error if the gem is already present.

The error displayed is very descriptive but if the user is specifying a gem version maybe he wants to update the gem.

Added an instructive message to inform the user that gem can also be updated.

This seemed to be the best informative message.

Closes #6341

### DIFF
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -109,10 +109,16 @@ module Bundler
             return if dep.type == :development
 
             update_prompt = ""
+            change_version = ". You may also need to change the version requirement specified in the Gemfile if it's too restrictive."
 
             if File.basename(@gemfile) == "injected gems"
-              update_prompt = ". If you want to update the gem version, run `bundle update #{current.name}`. " \
-                              "You may also need to change the version requirement specified in the Gemfile if it's too restrictive"
+              if dep.requirements_list.include?(">= 0") && !current.requirements_list.include?(">= 0")
+                update_prompt = ". Gem already added"
+              else
+                update_prompt = ". If you want to update the gem version, run `bundle update #{current.name}`"
+
+                update_prompt += change_version unless current.requirements_list.include?(">= 0")
+              end
             end
 
             raise GemfileError, "You cannot specify the same gem twice with different version requirements.\n" \
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -103,13 +103,19 @@ module Bundler
       # if there's already a dependency with this name we try to prefer one
       if current = @dependencies.find {|d| d.name == dep.name }
         deleted_dep = @dependencies.delete(current) if current.type == :development
-
         if current.requirement != dep.requirement
           unless deleted_dep
             return if dep.type == :development
-            raise GemfileError, "You cannot specify the same gem twice with different version requirements.\n" \
-                            "You specified: #{current.name} (#{current.requirement}) and #{dep.name} (#{dep.requirement}). " \
-                            "If you want to update the gem version, run `bundle update #{current.name}`. You may need to change the version requirement specified in the Gemfile if it's too restrictive"
+
+            # If no version is specified on adding gem
+            if "#{dep.requirement}" == ">= 0"
+              raise GemfileError, "Gem `#{current.name}` is already added.\n" \
+                              "If you want to update the gem version, run `bundle update #{current.name}`."
+            else
+              raise GemfileError, "You cannot specify the same gem twice with different version requirements.\n" \
+                              "You specified: #{current.name} (#{current.requirement}) and #{dep.name} (#{dep.requirement}). " \
+                              "If you want to update the gem version, run `bundle update #{current.name}`. You may need to change the version requirement specified in the Gemfile if it's too restrictive"
+            end
           end
 
         else
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -108,9 +108,9 @@ module Bundler
             return if dep.type == :development
 
             # If no version is specified on adding gem
-            if "#{dep.requirement}" == ">= 0"
+            if dep.requirement.to_s == ">= 0"
               raise GemfileError, "Gem `#{current.name}` is already added.\n" \
-                              "If you want to update the gem version, run `bundle update #{current.name}`."
+                              "If you want to update the gem version, run `bundle update #{current.name}`"
             else
               raise GemfileError, "You cannot specify the same gem twice with different version requirements.\n" \
                               "You specified: #{current.name} (#{current.requirement}) and #{dep.name} (#{dep.requirement}). " \
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -109,15 +109,14 @@ module Bundler
             return if dep.type == :development
 
             update_prompt = ""
-            change_version = ". You may also need to change the version requirement specified in the Gemfile if it's too restrictive."
 
-            if File.basename(@gemfile) == "injected gems"
+            if File.basename(@gemfile) == Injector::INJECTED_GEMS
               if dep.requirements_list.include?(">= 0") && !current.requirements_list.include?(">= 0")
                 update_prompt = ". Gem already added"
               else
                 update_prompt = ". If you want to update the gem version, run `bundle update #{current.name}`"
 
-                update_prompt += change_version unless current.requirements_list.include?(">= 0")
+                update_prompt += ". You may also need to change the version requirement specified in the Gemfile if it's too restrictive." unless current.requirements_list.include?(">= 0")
               end
             end
 
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -108,7 +108,8 @@ module Bundler
           unless deleted_dep
             return if dep.type == :development
             raise GemfileError, "You cannot specify the same gem twice with different version requirements.\n" \
-                            "You specified: #{current.name} (#{current.requirement}) and #{dep.name} (#{dep.requirement})"
+                            "You specified: #{current.name} (#{current.requirement}) and #{dep.name} (#{dep.requirement}). " \
+                            "You can also update the gem by running `bundle update #{current.name}`"
           end
 
         else
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -109,7 +109,7 @@ module Bundler
             return if dep.type == :development
             raise GemfileError, "You cannot specify the same gem twice with different version requirements.\n" \
                             "You specified: #{current.name} (#{current.requirement}) and #{dep.name} (#{dep.requirement}). " \
-                            "You can also update the gem by running `bundle update #{current.name}`"
+                            "If you want to update the gem version, run `bundle update #{current.name}`. You may need to change the version requirement specified in the Gemfile if it's too restrictive"
           end
 
         else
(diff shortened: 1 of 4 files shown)


lib/bundler/dsl.rb
```diff
@@ -107,7 +107,8 @@ def gem(name, *args)
           unless deleted_dep
             return if dep.type == :development
             raise GemfileError, "You cannot specify the same gem twice with different version requirements.\n" \
-                            "You specified: #{current.name} (#{current.requirement}) and #{dep.name} (#{dep.requirement})"
+                            "You specified: #{current.name} (#{current.requirement}) and #{dep.name} (#{dep.requirement}). " \
+                            "You can also update the gem by running `bundle update #{current.name}`"
```
hmistry: Let's change the wording to be more clearer... saying "You can also update..." leaves questions as to what are the other solutions to "also".
I see a couple of error cases - the users intent could be to update the gem version or might have forgotten it was already specified. I think saying something like this would be better.

```
"You cannot specify the same gem twice with different version requirements.\n" \
"You specified: #{current.name} (#{current.requirement}) and #{dep.name} (#{dep.requirement}). " \
"If you want to update the gem version, run `bundle update #{current.name}`. You may need to change the version requirement specified in the Gemfile if it's too restrictive."
```

hmistry: It would be great if you can add tests for this. Also check if this error is triggered in other use cases like if a gem was specified twice in a Gemfile and a user runs `bundle install`. The error should still makes sense for those use cases.

lib/bundler/dsl.rb
```diff
@@ -107,7 +107,8 @@ def gem(name, *args)
           unless deleted_dep
             return if dep.type == :development
             raise GemfileError, "You cannot specify the same gem twice with different version requirements.\n" \
-                            "You specified: #{current.name} (#{current.requirement}) and #{dep.name} (#{dep.requirement})"
+                            "You specified: #{current.name} (#{current.requirement}) and #{dep.name} (#{dep.requirement}). " \
+                            "If you want to update the gem version, run `bundle update #{current.name}`. You may need to change the version requirement specified in the Gemfile if it's too restrictive"
```
hmistry: Missing a period at the end 🙂

spec/commands/add_spec.rb
```diff
@@ -105,4 +105,16 @@
...
+  it "throws an error if a gem is added which is already specified in Gemfile" do
+    install_gemfile <<-G
+      source "file://#{gem_repo2}"
+      gem "rack", "1.0"
+    G
+
+    bundle "add 'rack'"
+
+    expect(out).to include "You cannot specify the same gem twice with different version requirements"
+    expect(out).to include "If you want to update the gem version, run `bundle update rack`. You may need to change the version requirement specified in the Gemfile if it's too restrictive"
```
hmistry: Interesting - this test example highlights that this is a poor error message for this particular case. As a user, if I did ran this (`bundle add rake`), I'd expect to see an error that says "gem already added. Skip adding the gem" and not the above message. I can see this being a potential for confusion in future.

I'd only expect the above error (i.e. cannot specify same gem twice with different versions) when I tried to run `bundle add rack --version "1.4"`.

Would it be simple for you to make this distinction in error messages between these 2 cases?

lib/bundler/dsl.rb
agrim123: A period is prepended and appended to the error message [here](https://github.com/bundler/bundler/blob/master/lib/bundler/dsl.rb#L554)

spec/commands/add_spec.rb
agrim123: Done. Please have a look :smile:

agrim123:
@colby-swandale How to place a check for `install` and `add` as [this](https://github.com/bundler/bundler/blob/master/lib/bundler/dsl.rb#L94) function does not know what command is calling it?

spec/commands/add_spec.rb
```diff
@@ -105,4 +105,30 @@
...
+    it "without version requirements" do
+      install_gemfile <<-G
+        source "file://#{gem_repo2}"
+        gem "rack", "1.0"
+      G
+
+      bundle "add 'rack'"
+
+      expect(out).to include("Gem `rack` is already added.")
+      expect(out).to include("If you want to update the gem version, run `bundle update rack`.")
```
hmistry: Remove the "If you want to update..." for this scenario because it's enough to say the gem is already added. The `add` command doesn't signify any intent to update in this situation.

spec/commands/install_spec.rb
```diff
@@ -321,6 +321,28 @@
...
 
+    it "throws an error if a gem is added twice in Gemfile without version requirements" do
+      install_gemfile <<-G
+        source "file://#{gem_repo2}"
+        gem "rack"
+        gem "rack"
+      G
+
+      expect(out).to include("Your Gemfile lists the gem rack (>= 0) more than once.")
+      expect(out).to include("You should probably keep only one of them")
```
hmistry: Clarify the message and be more direct:
`Remove any duplicate entries and specify the gem only once (per group).`

spec/commands/install_spec.rb
```diff
@@ -321,6 +321,28 @@
...
+
+    it "throws an error if a gem is added twice in Gemfile with version requirements" do
+      install_gemfile <<-G
+        source "file://#{gem_repo2}"
+        gem "rack", "1.0"
+        gem "rack", "1.1"
+      G
+
+      expect(out).to include("You cannot specify the same gem twice with different version requirements")
+      expect(out).to include("If you want to update the gem version, run `bundle update rack`. You may need to change the version requirement specified in the Gemfile if it's too restrictive")
```
hmistry: This should have the same message as the one above. No mention of bundle update.

hmistry:
@agrim123 When you think from a user perspective, the only scenario a user might intend to update a gem using `bundle add` that we should concern ourselves with is when they specify a version constraint in the command and it conflicts with the Gemfile. Keep in mind this is a small small chance because the `add` command doesn't imply `update`. All the other scenarios in the tests like specifying gems twice in the Gemfile or adding one that's already in the Gemfile, are more mistakes than a user intending to update a gem. Therefore it's much more important to address them as such by responding with gems specified twice or already added than the one slim chance they might be trying to update it. This distinction is key so think about it like that when solving this.

spec/commands/install_spec.rb
agrim123: I am still unable to figure out how to show different messages based on commands `add` and `install`

agrim123:
@colby-swandale I am unable to figure out how to differentiate error based on commands? Haven't been able to find a way to check for command in `gem` function.

hmistry:
@agrim123 One thought I had for differentiating based on commands is to look at the calling method or class ancestry. The commands are separated into classes like `Add`, `Install`, etc. The other thing is bundler uses Thor so the commands are method names in the base `Cli` class - Thor might be saving command called in a ivar. Maybe you can use that to distinguish appropriate messages.

agrim123:
@hmistry I tried using `caller` earlier, it returns an array and extracting the class out of array does not seem to be a good idea, because the class appears at different places in the array. 
So, I used `injected gems` [check](https://github.com/bundler/bundler/pull/6447/files#diff-ea5f81b03d4209ef02c02845cd95fcbfR112), if gems are [injected](https://github.com/bundler/bundler/blob/master/lib/bundler/injector.rb#L32), that is only in case of `add` then we have our check in place.

agrim123:
@hmistry I did not find anything from `Thor`. However, I got it working using `injected gems`. Can you please review?

spec/commands/add_spec.rb
```diff
@@ -106,4 +106,44 @@
...
+    it "without version requirements" do
+      install_gemfile <<-G
+        source "file://#{gem_repo2}"
+        gem "rack", "1.0"
+      G
+
+      bundle "add 'rack'"
+
+      expect(out).to include("You cannot specify the same gem twice with different version requirements")
+      expect(out).to include("If you want to update the gem version, run `bundle update rack`. You may also need to change the version requirement specified in the Gemfile if it's too restrictive")
```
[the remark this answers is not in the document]
agrim123: Added `Gem already added` statement.

spec/commands/add_spec.rb
```diff
@@ -106,4 +106,44 @@
...
+    it "with different version requirements" do
+      install_gemfile <<-G
+        source "file://#{gem_repo2}"
+        gem "rack"
+      G
+
+      bundle "add 'rack' --version=1.1"
+
+      expect(out).to include("You cannot specify the same gem twice with different version requirements")
+      expect(out).to include("If you want to update the gem version, run `bundle update rack`. You may also need to change the version requirement specified in the Gemfile if it's too restrictive")
```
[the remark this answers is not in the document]
agrim123: Done.

agrim123:
@hmistry Let me know if something more needs to be done here. 😄

spec/commands/add_spec.rb
```diff
@@ -106,4 +106,46 @@
     bundle "add 'baz' --source='file://does/not/exist'"
     expect(out).to include("Could not fetch specs from file://does/not/exist/")
   end
+
+  context "throws an error if a gem is added which is already specified in Gemfile with version" do
+    it "with different version requirement" do
```
colby-swandale: Your test description sounds more like a context for the tested behaviour rather than what the behaviour should be. This also applies to the other ‘it’ blocks in this PR

hmistry:
@segiddins This looks good to me.
@agrim123 Good job on this. Thanks for listening and incorporating the feedback.

lib/bundler/dsl.rb
```diff
@@ -106,13 +106,29 @@ def gem(name, *args)
         if current.requirement != dep.requirement
           unless deleted_dep
             return if dep.type == :development
+
+            update_prompt = ""
+            change_version = ". You may also need to change the version requirement specified in the Gemfile if it's too restrictive."
```
segiddins: this is only used once, so it can be inlined into that location, instead of sitting at the outer scope

lib/bundler/dsl.rb
```diff
@@ -106,13 +106,29 @@ def gem(name, *args)
...
+
+            if File.basename(@gemfile) == "injected gems"
```
segiddins: nit: extract `"injected gems" into a constant in the `Injector` class

lib/bundler/dsl.rb
agrim123: When [this](https://github.com/agrim123/bundler/blob/9a7328ba1275eb60b36212b91bee8e8700f07ec7/lib/bundler/dsl.rb#L113) condition is false then [this](https://github.com/agrim123/bundler/blob/9a7328ba1275eb60b36212b91bee8e8700f07ec7/lib/bundler/dsl.rb#L123-L125) would throw error.

segiddins: that exception message doesn't include `change_version` ?

hmistry: @agrim123 Let's move line 111 into line 119 where it's used. That way ruby doesn't allocate memory for that string when it's not needed. This is what @segiddins is getting at.

agrim123: Oh! Sorry. Fixing it

segiddins:
@bundlerbot r+

PR #6448, opened 2018-03-19
[WIP] Lockfile checksum

agrim123:

The main aim is to reduce Gemfile resolve time if checksum remains same.

Add a section `CHECKSUMS` to lockfile that will store the checksum and will be used to match for future.  
I continued the work done by @segiddins. 

The main approach was to store a checksum of current gemfile under `CHECKSUMS` sections in `Gemfile.lock`

In continuation of #5584.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -14,6 +14,7 @@ module Bundler
       :platforms,
       :requires,
       :ruby_version,
+      :optional_groups,
       :lockfile,
       :gemfiles
     )
@@ -30,6 +31,13 @@ module Bundler
       gemfile = Pathname.new(gemfile).expand_path
 
       raise GemfileNotFound, "#{gemfile} not found" unless gemfile.file?
+
+      if lockfile && File.exist?(lockfile)
+        # parsed_lockfile = LockfileParser.new(Bundler.read_file(lockfile))
+        # use_lockfile = parsed_lockfile.gemfile_checksums.reduce do |file, checksum|
+        #   p file, checksum
+        # end
+      end
 
       Dsl.evaluate(gemfile, lockfile, unlock)
     end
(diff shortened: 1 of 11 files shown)


agrim123:
@segiddins I am not able to test static gemfile option, do not know how to call it from Gemfile. Can you please help me here? And how shall I proceed with this PR?

segiddins:
`_static_gemfile!` in the DSL

simi:
Hello! I really like this feature.

I went thru code and I have 2 initial questions. Sorry if they were answered already, but I can't find any previous discussion related to those questions.

1. Why checksum can't be stored in file next to `lock` files?
2. I see `md5` is used to digest lock. Can't something different (ex. SHA1) be used since `md5` is blocked on `FIPS` enabled systems? There were some issues with `md5` and FIPS [before](https://github.com/bundler/bundler/issues?q=is%3Aissue+fips+is%3Aclosed). I know there's `md5_available?` helper used to avoid problems. But do we really need to introduce new features with `md5` and disable them for `FIPS` enabled systems?

segiddins:
MD5 is quicker

agrim123:
This test block is failing
```ruby
context "with a current version" do
  let(:base_version) { Gem::Version.create(Bundler::VERSION) }

  it "returns an empty array" do
    expect(subject).to eq([])
  end
end
```
We have added sections in `2.0` [version](https://github.com/bundler/bundler/pull/6448/files#diff-3c625d3cd7d7604b3e2e3c5487a5ede6R51). And when we are testing with `BUNDLER_SPEC_SUB_VERSION=1.98` this obviously fails. How do I counter this?

segiddins:
The lockfile parser hash of known sections would need an update

agrim123:
I didn't quite understand how to proceed. Should we not change `BUNDLER_SPEC_SUB_VERSION` for test?

PR #6450, opened 2018-03-21
[Binstubs] Add --all options

segiddins:

The problem was we have removed the `--binstubs` argument to `bundle install`, so saying "oh well, lets install all the bin stubs" became more tedious, requiring multiple calls to `bundle binstubs` and probably some swearing since you forgot what gem contains what binstub.

My diagnosis was we need a one-liner to generate bin stubs for all gems.

My fix adds such an option to `bundle binstubs`!

I chose this fix because it keeps binstub generation in its own command, rather than continuing to piggy-back off of `install`

### DIFF
--- a/spec/commands/binstubs_spec.rb
+++ b/spec/commands/binstubs_spec.rb
@@ -39,6 +39,18 @@ RSpec.describe "bundle binstubs <gem>" do
       expect(bundled_app("bin/rails")).to exist
     end
 
+    it "allows installing all binstubs" do
+      install_gemfile! <<-G
+        source "file://#{gem_repo1}"
+        gem "rails"
+      G
+
+      bundle! :binstubs, :all => true
+
+      expect(bundled_app("bin/rails")).to exist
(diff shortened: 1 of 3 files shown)


spec/commands/binstubs_spec.rb
```diff
@@ -50,6 +62,17 @@
...
 
+    it "displays an error when used with --all and gems" do
+      install_gemfile <<-G
+        source "file://#{gem_repo1}"
+        gem "rack"
+      G
+
+      bundle "binstubs rack", :all => true
+      expect(last_command).to be_failure
+      expect(last_command.stderr).to include("Cannot specify --all with specific gems")
```
segiddins: `bundler_err`

colby-swandale:
Just needs to adjust the failing spec to use `bundle_err`

segiddins:
Should be fixed now, sorry

colby-swandale:
Thanks! @bundlerbot r+

Issue #6058, opened 2017-09-23
Should the `path:` option add gems to LOAD_PATH if they're not included in dependencies?


deivid-rodriguez:
I'm not sure if this is expected behaviour or not. We're might be doing unsupported stuff, but I thought it might be worth asking.

* We have a project structure like the following:

```shell
.
├── decidim.gemspec
├── decidim-assemblies
│   └── decidim-assemblies.gemspec
├── decidim-admin
│   └── decidim-admin.gemspec
├── decidim-participatory_processes
│   └── decidim-participatory_processes.gemspec
... # More similar gems
└── test_app
    ├── Gemfile
    └── Gemfile.lock

3 directories, 5 files
```

* Every `decidim-*` subgem is a runtime dependency of the `decidim` root gem, _except for_ `decidim-assemblies`.

* In the `test_app`'s `Gemfile`, we specifiy `gem "decidim" , path: ".."` so that it uses the local version of our library.

With this setup, I would expect `decidim-assemblies` to be ignored from the test app dependencies. However, this is only partly true since even though `decidim-assemblies` is not included in the test app `Gemfile.lock` file, it _is_ added to the LOAD_PATH.

This is because `decidim-assemblies.gemspec` adds itself to the LOAD_PATH as it's normally done in gemspecs, but I wonder if this gemspec should not be read at all in this situation?

I wrote a test case to better illustrate the situation

```diff
diff --git a/spec/install/gemfile/path_spec.rb b/spec/install/gemfile/path_spec.rb
index f7789e7ea..f0fd38f55 100644
--- a/spec/install/gemfile/path_spec.rb
+++ b/spec/install/gemfile/path_spec.rb
@@ -114,6 +114,51 @@ RSpec.describe "bundle install with explicit source paths" do
     expect(out).to eq("WIN OVERRIDE")
   end
 
+  it "does not add local gems to load path when they're not dependencies" do
+    build_lib "foo", :path => lib_path("nested/foo")
+
+    gemspec = lib_path("nested/foo").join("foo.gemspec").to_s
+    File.open(gemspec, "w") do |f|
+      f.write <<-G
+        # -*- encoding: utf-8 -*-
+        # stub: foo 1.0 ruby lib
+
+        lib = File.expand_path("../lib", __FILE__)
+        $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
+
+        Gem::Specification.new do |s|
+          s.name = "foo".freeze
+          s.version = "1.0"
[...]

PR #6451, opened 2018-03-21
[CLI::Gem] Store the VERSION in a separate file

segiddins:

The problem was the `bundle gem` template generates a `gemspec` file that manipulates `$LOAD_PATH`, which can lead to unexpected behavior.

My diagnosis was we should continue storing the version in a single place, but one that is not a ruby file.

My fix puts the version in a `VERSION` file in the top level of the generated repo. That file is then read by both the gemspec and `version.rb`.

I chose this fix because the alternatives seemed like string parsing `version.rb`, and string parsing is best avoided.

This will also make tooling around version management easier, as it's easier to manipulate a plain text file that contains only the version than an entire ruby file.


colby-swandale:
Do you have any examples of this unexpected behavior?

indirect:
@colby-swandale in my experience, it has been any code that evaluates gemspecs (especially from path gems) and then does not actually activate that gem after examining the gemspec. At that point, it's too late, and the gem's `lib` is already on the load path because of having evaluated the gemspec.

hsbt:
What's advantage to use `require_relative` instead of this?

deivid-rodriguez:
I was beaten by `LOAD_PATH` causing unexpected behaviour in https://github.com/bundler/bundler/issues/6058 (although my surprise was not the `LOAD_PATH` being modified, but the gemspec being read at all), so I'm happy to try these solutions!

segiddins:
@indirect r?

indirect:
No more load path seems great, but won't `require_relative` give us the same advantage? Is the separate file to make it easier to programatically increment, or something?

segiddins:
> Is the separate file to make it easier to programatically increment, or something?

Exactly that. No having to write regular expressions to update the gem version

segiddins:
Any feedback on this?

deivid-rodriguez:
I would personally fix the problem by using `require_relative` and separately discuss the best way for new gems to specify their version going forward.

Maybe a separate file aligns us with the standard ".ruby-version" file version managers use, so we could call it `.gem-version` or something like that. But I feel this is part of a separate discussion from the "avoid $LOAD_PATH manipulation in gemspecs" issue.

PR #6455, opened 2018-03-25
[CLI::Gem] Add a --rubocop option

segiddins:
Based upon #6451.

The problem was I always have to remember how to add RuboCop to my Rakefile when I set up a new gem.

My diagnosis was that RuboCop has become enough of a community standard that it makes sense to offer it in `bundle gem`.

My fix adds an option to `bundle gem` to add in RuboCop, in the same way options like `:coc`and `:mit` are handled.

I chose this fix because it does not require bundler have an opinion on _how_ rubocop is configured.

### DIFF
--- a/lib/bundler/cli/gem.rb
+++ b/lib/bundler/cli/gem.rb
@@ -100,7 +100,7 @@ module Bundler
         end
       end
 
-      config[:test_task] = config[:test] == "minitest" ? "test" : "spec"
+      config[:test_task] = config[:test] == "minitest" ? :test : :spec
 
       if ask_and_set(:mit, "Do you want to license your code permissively under the MIT license?",
         "This means that any other developer or company will be legally allowed to use your code " \
@@ -122,6 +122,15 @@ module Bundler
         config[:coc] = true
         Bundler.ui.info "Code of conduct enabled in config"
         templates.merge!("CODE_OF_CONDUCT.md.tt" => "CODE_OF_CONDUCT.md")
+      end
+
+      if ask_and_set(:rubocop, "Do you want to add rubocop as a dependency for gems you generate?",
+        "RuboCop is a static code analyzer that has out-of-the-box rules for many " \
+        "of the guidelines in the community style guide. " \
+        "For more information, see the RuboCop docs (https://docs.rubocop.org/en/stable/) " \
+        "and the Ruby Style Guides (https://github.com/rubocop-hq/ruby-style-guide).")
+        config[:rubocop] = true
+        Bundler.ui.info "RuboCop enabled in config"
       end
 
       templates.merge!("exe/newgem.tt" => "exe/#{name}") if config[:exe]
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -542,6 +542,7 @@ module Bundler
     method_option :ext, :type => :boolean, :default => false, :desc => "Generate the boilerplate for C extension code"
     method_option :git, :type => :boolean, :default => true, :desc => "Initialize a git repo inside your library."
     method_option :mit, :type => :boolean, :desc => "Generate an MIT license file. Set a default with `bundle config set gem.mit true`."
+    method_option :rubocop, :type => :boolean, :desc => "Add rubocop to the generated Rakefile and gemspec. Set a default with `bundle config set gem.rubocop true`."
     method_option :test, :type => :string, :lazy_default => "rspec", :aliases => "-t", :banner => "rspec",
                          :desc => "Generate a test directory for your library, either rspec or minitest. Set a default with `bundle config set gem.test rspec`."
     def gem(name)
--- a/lib/bundler/templates/newgem/Gemfile.tt
+++ b/lib/bundler/templates/newgem/Gemfile.tt
@@ -10,3 +10,6 @@ gem "rake-compiler"
 <%- if config[:test] -%>
 gem "<%= config[:test] %>", "~> <%= config[:test_framework_version] %>"
 <%- end -%>
+<%- if config[:rubocop] -%>
+gem "rubocop"
+<%- end -%>
--- a/lib/bundler/templates/newgem/Rakefile.tt
+++ b/lib/bundler/templates/newgem/Rakefile.tt
@@ -1,4 +1,5 @@
 require "bundler/gem_tasks"
+<% default_task_names = [config[:test_task]] -%>
 <% if config[:test] == "minitest" -%>
 require "rake/testtask"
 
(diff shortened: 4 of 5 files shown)


colby-swandale:
Great!! @bundlerbot r+

colby-swandale:
Oh, i didn't notice the failing specs

colby-swandale:
This looks like it's rebased with some other commits it shouldn't have.

colby-swandale:
Oh i didn't notice that this wasn't going into master.

segiddins:
Rebased `segiddins/bundle-gem-version-file`

lib/bundler/templates/newgem/VERSION
```diff
@@ -0,0 +1 @@
+0.1.0
```
koic: I thought it would be better to add a newline for EOF.

segiddins:
@indirect r?

indirect:
looks good, @bundlerbot r+

segiddins:
@bundlerbot retry

colby-swandale:
@bundlerbot retry

segiddins:
Will rebase this once https://github.com/bundler/bundler/pull/6451 gets merged

segiddins:
I based these changes off of #6451, since I wanted to be able to use both when creating my own new gems

lib/bundler/cli/gem.rb
```diff
@@ -125,6 +125,15 @@ def run
         templates.merge!("CODE_OF_CONDUCT.md.tt" => "CODE_OF_CONDUCT.md")
       end
 
+      if ask_and_set(:rubocop, "Do you want to add rubocop as a dependency for gems you generate?",
+        "RuboCop is a static code analyzer that has out-of-the-box rules for many" \
+        "of the guidelines in the community style guide." \
+        "For more information, see the RuboCop docs (https://rubocop.readthedocs.io/en/latest/)" \
```
koic: FYI, the document URL of RuboCop is now below.
http://www.rubocop.org/en/latest/

Refer https://github.com/rubocop-hq/rubocop/blob/v0.58.2/rubocop.gemspec#L33

lib/bundler/cli/gem.rb
```diff
@@ -125,6 +125,15 @@ def run
...
+        "and the Ruby Style Guides (https://github.com/bbatsov/ruby-style-guide).")
```
koic: The URL of Ruby Style Guide is now below.
https://github.com/rubocop-hq/ruby-style-guide

deivid-rodriguez:
@segiddins Ah, ok, I thought there was some kind of dependency between both PRs and that was why tests were failing.

deivid-rodriguez:
@segiddins I just noticed that this PR is not targetted to master but to another branch. Since both PRs seem independent, and #6451 is still under discussion, what do you think about changing the target branch of this PR to master, so it can be merged?

segiddins:
@deivid-rodriguez feel free to go for it 👍

deivid-rodriguez:
I can do that, :+1:. What do you think about https://github.com/bundler/bundler/pull/6455#pullrequestreview-143380582 though? I think we should either add a rubocop config file to the generator, or switch the skeleton to use rubocop defaults.

segiddins:
First thing I’m going to do is run autocorrect anyways, so I don’t think it’s terribly necessary

deivid-rodriguez:
I'm not sure I understood what you mean.

My idea is that it'd be a good thing that the gem generated with the `--rubocop` option has a green rubocop configuration, so that no further intervention is needed from the user to get started. I don't think it's good that the user needs to run `rubocop --auto-correct` on the freshly generated skeleton in order to get started, if that's what you meant.

colby-swandale:
I agree with @deivid-rodriguez, i believe new gems should follow the community style guidelines (which i guess is set by rubocop), rather then following Bundler's style guidelines.

deivid-rodriguez:
I think I would actually prefer to keep bundler style, so I would add an initial rubocop configuration to this PR to enforce it. But if we prefer to stick to bundler's defaults, then yeah I would edit the templates to follow them.

deivid-rodriguez:
@segiddins Do you mind if I wrap up this PR? You said it's not terribly necessary but I guess you don't oppose to my proposal?

segiddins:
Go for it

deivid-rodriguez:
So I had a look at this and I think it's premature to do what I wanted to do, since the gem templates still support 1.8.7 and rubocop was not even a thing at the time. So, I now think @segiddins' version is good enough for the initial version of this feature.

lib/bundler/cli/gem.rb
deivid-rodriguez: Should I add the URL in parentheses after "in the community style guide"?

deivid-rodriguez:
Only changes I made is pointing this PR to master, and adding some missing spaces to the CLI message.

lib/bundler/templates/newgem/newgem.gemspec.tt
```diff
@@ -54,4 +54,7 @@ Gem::Specification.new do |spec|
 <%- if config[:test] -%>
   spec.add_development_dependency "<%= config[:test] %>", "~> <%= config[:test_framework_version] %>"
 <%- end -%>
+<%- if config[:rubocop] -%>
+  spec.add_development_dependency "rubocop"
+<%- end -%>
```
ixti: Worth also to add `rubocop-rspec` if chosen test framework is rspec IMO

colby-swandale:
This is going to have to wait for bundler 2 to be released before we can merge it. For some reason the ruby parser in 1.8.7 is locking up in `Bundler::UI:Shell#yes?`

colby-swandale:
This should be ok to merge now. I can work on getting this over the line.

lib/bundler/cli/gem.rb
```diff
@@ -124,6 +124,15 @@ def run
         templates.merge!("CODE_OF_CONDUCT.md.tt" => "CODE_OF_CONDUCT.md")
       end
 
+      if ask_and_set(:rubocop, "Do you want to add rubocop as a dependency for gems you generate?",
+        "RuboCop is a static code analyzer that has out-of-the-box rules for many " \
+        "of the guidelines in the community style guide. " \
+        "For more information, see the RuboCop docs (http://www.rubocop.org/en/latest/) " \
```
koic: The RuboCop docs URL has been updated to https://docs.rubocop.org.

cf. https://github.com/rubocop-hq/rubocop/commit/86b4a41e59fdba6fbd70b26a0fc0f1a2c916c56b#diff-458bc38d82061cdac508dfb1be2b918b

lib/bundler/cli/gem.rb
```diff
@@ -124,6 +124,15 @@ def run
...
+        "and the Ruby Style Guides (https://github.com/rubocop-hq/ruby-style-guide).")
```
koic: The Ruby Style Guide URL has been updated to https://rubystyle.guide.

cf. https://github.com/rubocop-hq/rubocop/issues/7120

deivid-rodriguez:
@colby-swandale, are you still planning to wrap up this PR? I can follow up if you are very busy :+1:.

deivid-rodriguez:
@colby-swandale Forgot to mention that I updated this PR and it should be finally ready, in case you want to have a look at it before merging.

deivid-rodriguez:
@bundlerbot merge

PR #6462, opened 2018-03-27
Made a few, nitpicky copyedits and removed some redundant phrases

rubymorillo:

The problem was...
...nothing big, just a review of the file to make sure the doc made sense. My only updates were to fix some punctuation, capitalization, and remove redundant phrasing.

My fix...
...capitalize "Bundler", fix some colons, and remove "in the context of"

I chose this fix because...
...it's good to review docs to make sure they still make sense! I removed "in the context of" because the phrase could be confusing to readers, making the sentences more complicated than needed.


rubymorillo:
Opening a new PR from the latest master

PR #6464, opened 2018-03-28
Document the order Bundler loads settings

### LINKED ISSUE #6334: Document the order that Bundler loads settings
Bundler settings can be configured in a number of different ways but the [bundler-config](https://github.com/bundler/bundler/blob/master/man/bundle-config.ronn) docs do not talk about in which [order Bundler loads this settings](https://github.com/bundler/bundler/blob/master/lib/bundler/settings.rb#L93). Which seems to be not as people expect (local > ENV). This has caused confusion and people have open issues (#6270) about this. This should be something that is documented that makes this clear to the user.

rubymorillo:
Parsed out the information in the first graph in order to help users understand the order of priority Bundler follows when loading config settings.

The problem was...
...[confusion from end users](https://github.com/bundler/bundler/issues/6270) about the order that Bundler loads settings in. 

Before:
```
This command allows you to interact with bundler's configuration system.
Bundler retrieves its configuration from the local application (`app/.bundle/config`),
environment variables, and the user's home directory (`~/.bundle/config`),
in that order of priority.
```

My diagnosis was...
...[Colby pointed out how the bundle-config doc](https://github.com/bundler/bundler/issues/6334) needed to be updated to reflect this. After reviewing the `bundle-config` doc, it appears there was some information about how Bundler loads config settings, but it wasn't complete and as apparent to users. So I decided to separate the graph out and document the settings as a list so users could spot that easier.

My fix...
...was to verify the order of priority of settings and write them out as a list.

After:
```
This command allows you to interact with Bundler's configuration system.

Bundler loads configuration settings in this order:

1. Local config (`app/.bundle/config`)
2. Environmental variables (`ENV`)
3. Global config (`~/.bundle/config`)
4. Bundler default config
```

I chose this fix because...
it was clear that the document had to be updated, and lists are easier for readability.

fixes #6334

### DIFF
--- a/man/bundle-config.ronn
+++ b/man/bundle-config.ronn
@@ -7,10 +7,14 @@ bundle-config(1) -- Set bundler configuration options
 
 ## DESCRIPTION
 
-This command allows you to interact with bundler's configuration system.
-Bundler retrieves its configuration from the local application (`app/.bundle/config`),
-environment variables, and the user's home directory (`~/.bundle/config`),
-in that order of priority.
+This command allows you to interact with Bundler's configuration system.
+
+Bundler loads configuration settings in this order:
+
+1. Local config (`app/.bundle/config`)
+2. Environmental variables (`ENV`)
+3. Global config (`~/.bundle/config`)
+4. Bundler default config
 
 Executing `bundle config` with no parameters will print a list of all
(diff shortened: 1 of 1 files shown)


colby-swandale:
Great! @bundlerbot r+

PR #6466, opened 2018-03-29
Bump the bundle_binstub check-length to 300 characters

tduffield:

I could not start a binstub binary.

Some packaging systems require modifications to the hash-bang
interpreter path, making them longer than normal. One such example of
this is the habitat-sh/habitat project. The result is that the 150
character limit does not find the regex it is looking for, and prevents
an otherwise valid binary from starting.

This change doubles the length of the check from 150 characters to 300
characters. This change has been validated in an impacted environment.

It was the most straightforward.


segiddins:
@bundlerbot r+

segiddins:
@bundlerbot retry

Issue #6162, opened 2017-11-08
Bundler 1.16 incorrectly captures/stores `BUNDLE_GEMFILE` information into binstubs, which breaks in scenarios where multiple Gemfiles have been `bundle installed`.


jchesterpivotal:
# Issue

Bundler 1.16 incorrectly captures/stores `BUNDLE_GEMFILE` information into binstubs, which breaks in scenarios where multiple Gemfiles have been `bundle installed`.

We think we did not observe this sooner because of the time it took for bundler 1.16 to propagate to our CI pipeline: first from Bundler to the standard `ruby:2.4.2` Dockerfile, then from there to our CI-image-building job, and then a delay until a job using our CI image was triggered. But we believe the issue existed from the time that bundler 1.16 was released.

# Investigation

Yesterday our CI builds began to fail inexplicably:

```
bundle exec rake
rake aborted!
LoadError: cannot load such file -- active_record/railtie
```

We investigated the commits leading up to the build, but none of them were related to Bundler, Docker, Rake, any Gemfiles etc.

We noticed that our logs had changed in an interesting way, however:

Seen in broken build:
```
Bundle complete! 12 Gemfile dependencies, 48 gems now installed.
Bundled gems are installed into `/usr/local/bundle`
```

Seen in working build immediately prior to broken build:
```
Bundle complete! 86 Gemfile dependencies, 220 gems now installed.
Bundled gems are installed into /usr/local/bundle.
```

At first we suspected we'd messed up our Gemfile, but it had not changed. We then realised that the "12 Gemfile dependencies" result was from a _different_ Gemfile from the one in the directory.

## Our CI / Docker process

How could this happen? Well, to begin with, we run all our builds inside Docker containers using [Concourse](http://concourse.ci/). As an optimisation, we pre-build images that contain dependencies used by multiple codebases in our system. Then all CI builds use the common image. `bundle exec` is used to keep dependencies straight and previously worked fine.

Here is a Dockerfile based on our internal one:

```
FROM ruby:2.4.2

WORKDIR /usr/src/app

COPY package.json \
     first.gemfile \
     first.gemfile.lock \
     second.gemfile \
     second.gemfile.lock \
     third.gemfile \
     third.gemfile.lock \
     /usr/src/app/

RUN bundle config --global jobs 16 && \
    BUNDLE_GEMFILE=/usr/src/app/first.gemfile bundle install && \
    BUNDLE_GEMFILE=/usr/src/app/second.gemfile bundle install && \
[...]

Issue #6154, opened 2017-11-06
Upgrading to Bundler 1.16 causes bundler to look for Gemfile in root directory


Gerg:
# Summary

@jenspinney and @deniseyu discovered this issue when using bundler in a docker image we build containing bundler 1.16: https://github.com/cloudfoundry/capi-dockerfiles/blob/26ac90206a234ad29b2775500d471a267d8fced4/capi-ruby-units/Dockerfile#L4

The error looks like this: 
```
[!] There was an error parsing `Gemfile`: No such file or directory @ rb_sysopen - /Gemfile. Bundler cannot continue.
```
Our workaround is to explicitly sets `BUNDLE_GEMFILE` to point to the correct Gemfile.

# Details
After upgrading from Bundler 1.15 to 1.16, `bundler config` is reporting that the `BUNDLE_GEMFILE` env var is set to `/Gemfile`, despite no such environment variable exisiting. 

Looking in `/usr/local/bundle`, we see a bunch of files in the `/usr/local/bundle/bin` directory that all follow this pattern:

```root@5e3bc6a7-b2f5-4dac-4fd7-a3ebb5140aa6:/usr/local/bundle# cat bin/clockwork
#!/usr/bin/env ruby
# frozen_string_literal: true

#
# This file was generated by Bundler.
#
# The application 'clockwork' is installed as part of a gem, and
# this file is here to facilitate running it.
#

bundle_binstub = File.expand_path("../bundle", __FILE__)
load(bundle_binstub) if File.file?(bundle_binstub)

require "pathname"
ENV["BUNDLE_GEMFILE"] ||= File.expand_path("../../../../../Gemfile",
  Pathname.new(__FILE__).realpath)

require "rubygems"
require "bundler/setup"

load Gem.bin_path("clockwork", "clockwork")
```

That `../../../../../Gemfile` filepath looks pretty suspicious, and we don't know why Bundler is generating it.

rheaton:
I wonder if the behavior I'm seeing is related. 
We upgraded to 1.16 from 1.15 and discovered the adding a `--system --gemfile=path/to/Gemfile` set the BUNDLE_GEMFILE environment variable outside of the execution of that command. 

So in 1.15, the following worked:

```
bundle install --system --gemfile=/path/1/Gemfile && bundle install --system --gemfile=/path/2/Gemfile
cd /path/1 && bundle exec ruby -e "require 'gem_only_in_1_gemfile'"
```

[...]

PR #6469, opened 2018-03-31
Replace hardcoded gemfile path with methods to discover Gemfile for bundle binstub

eanlain:

Issues #6154 and #6162, in which Bundler v1.16.1 appeared to be looking for the wrong Gemfile when using Docker. Further investigation showed that this issue was not just limited to the use of Docker (see [non-Docker repro of 6154](https://gist.github.com/eanlain/83f97f919701c864fb68dfdcf813cf5a)).

My diagnosis was that Bundler (v1.16.1) creates a binstub of itself every time `bundle install` is run and is storing the filepath of the last gemfile that `bundle install` was run with. 

In the event where no `ENV["BUNDLE_GEMFILE"]` is provided, the binstub, when executed (e.g. `bundle exec`), will try and use the gemfile of the last successful `bundle install`. For applications that use multiple gemfiles this has resulted in gems not being installed/loaded/run, the wrong gemfile being written to the Bundler binstub, and a persistent `bundle config gemfile` entry.

My fix adds methods to the `Executable.bundler` template that enable the Bundler binstub to discover gemfiles based on its current directory. Some code was adapted from `lib/bundler/shared_helpers.rb`, which provides methods like `default_gemfile` to discover gemfiles.

I chose this fix because it decouples the Bundler binstub's reliance on a specific Gemfile and instead enables dynamic discovery of a gemfile, which is more inline with the behavior in prior versions of Bundler.

### DIFF
--- a/lib/bundler/templates/Executable.bundler
+++ b/lib/bundler/templates/Executable.bundler
@@ -8,7 +8,6 @@
 # this file is here to facilitate running it.
 #
 
-require "pathname"
 require "rubygems"
 
 m = Module.new do
@@ -41,7 +40,7 @@ m = Module.new do
   def gemfile
     gemfile = find_gemfile
     return unless gemfile
-    Pathname.new(gemfile).untaint.expand_path.to_s
+    File.expand_path(gemfile)
   end
 
   def find_gemfile
@@ -62,7 +61,7 @@ m = Module.new do
 
   def search_up(*names)
     previous = nil
-    current  = File.expand_path(Pathname.pwd).untaint
+    current  = File.expand_path(Dir.pwd)
 
     until !File.directory?(current) || current == previous
       names.each do |name|
--- a/lib/bundler/templates/Executable.bundler
+++ b/lib/bundler/templates/Executable.bundler
@@ -8,6 +8,7 @@
 # this file is here to facilitate running it.
 #
 
+require "pathname"
 require "rubygems"
 
 m = Module.new do
@@ -38,10 +39,39 @@ m = Module.new do
   end
 
   def gemfile
-    gemfile = ENV["BUNDLE_GEMFILE"]
-    return gemfile if gemfile && !gemfile.empty?
+    gemfile = find_gemfile
+    return unless gemfile
+    Pathname.new(gemfile).untaint.expand_path.to_s
+  end
 
-    File.expand_path("../<%= relative_gemfile_path %>", __FILE__)
+  def find_gemfile
+    given = ENV["BUNDLE_GEMFILE"]
+    return given if given && !given.empty?
+    find_file(*gemfile_names)
+  end
+
+  def gemfile_names
+    ["Gemfile", "gems.rb"]
+  end
+
+  def find_file(*names)
+    search_up(*names) do |filename|
+      return filename if File.file?(filename)
+    end
+  end
+
+  def search_up(*names)
+    previous = nil
+    current  = File.expand_path(Pathname.pwd).untaint
+
+    until !File.directory?(current) || current == previous
+      names.each do |name|
+        filename = File.join(current, name)
+        yield filename
+      end
+      previous = current
+      current = File.expand_path("..", current)
+    end
   end
 
   def lockfile
--- a/spec/commands/binstubs_spec.rb
+++ b/spec/commands/binstubs_spec.rb
@@ -188,51 +188,64 @@ RSpec.describe "bundle binstubs <gem>" do
         end
       end
 
-      context "when using multiple gemfiles" do
+      context "when working with multiple gemfiles" do
+        let(:bundle_binstub) { bundled_app("bin/bundle") }
+        let(:load_bundle_exe) { "load Gem.bin_path(\"bundler\", \"bundle\")" }
+        let(:output_bundle_gemfile) { "puts ENV[\"BUNDLE_GEMFILE\"]" }
+
         before do
           build_repo2 do
             build_gem "print_hello", "1.0" do |s|
               s.executables = "print_hello"
               s.bindir = "exe"
               s.write "exe/print_hello", <<-R
-                puts "Hello"
+                puts "Hello!"
+              R
+            end
+            build_gem "print_goodbye", "1.0" do |s|
+              s.executables = "print_goodbye"
+              s.bindir = "exe"
+              s.write "exe/print_goodbye", <<-R
+                puts "Goodbye."
               R
             end
           end
 
-          bundle! "config --global path #{bundled_app("bundle")}"
-          bundle! "config --global bin #{bundled_app("bundle/bin")}"
+          bundle! "config --global path #{bundled_app}"
+          bundle! "config --global bin #{bundled_app("bin")}"
+
           install_gemfile! <<-G
             source "file://#{gem_repo2}"
           G
           bundle! "binstubs bundler"
         end
 
-        it "does not contain references to a specific gemfile path" do
+        it "uses the correct Gemfile" do
           create_file "one/Gemfile", <<-G
-            source "file://#{gem_repo2}"
-            gem "rack"
-          G
-          create_file "two/Gemfile", <<-G
             source "file://#{gem_repo2}"
             gem "print_hello"
           G
+          create_file "two/Gemfile", <<-G
+            source "file://#{gem_repo2}"
+            gem "print_goodbye"
+          G
+          Dir.chdir("one") { bundle! "install" }
+          Dir.chdir("two") { bundle! "install" }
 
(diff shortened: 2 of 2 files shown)


lib/bundler/templates/Executable.bundler
```diff
@@ -8,6 +8,7 @@
 # this file is here to facilitate running it.
 #
 
+require "pathname"
```
segiddins: is requiring pathname here necessary?

spec/commands/binstubs_spec.rb
```diff
@@ -187,6 +187,54 @@
...
+          create_file "one/Gemfile", <<-G
+            source "file://#{gem_repo2}"
+            gem "rack"
+          G
+          create_file "two/Gemfile", <<-G
+            source "file://#{gem_repo2}"
+            gem "print_hello"
+          G
+
+          Dir.chdir("one")
```
segiddins: please use blocks for `chdir`

spec/commands/binstubs_spec.rb
```diff
@@ -187,6 +187,54 @@
...
+          bundle! "install"
+          expect(out).to include("rack")
+
+          bundler_binstub = sys_exec "cat #{bundled_app("bundle/bin")}/bundle"
```
segiddins: running `cat` here is unnecessary, can just use `File.read`

actually, we should be executing the binstubs

lib/bundler/templates/Executable.bundler
eanlain: Nope, not necessary, will remove.

eanlain:
Alrighty, I removed the use of `pathname` from the bundler executable template and fleshed out the spec a bit more to make the expected behavior a bit clearer 👍

spec/commands/binstubs_spec.rb
```diff
@@ -187,6 +187,67 @@
...
+          create_file "two/Gemfile", <<-G
+            source "file://#{gem_repo2}"
+            gem "print_goodbye"
+          G
+          Dir.chdir("one") { bundle! "install" }
+          Dir.chdir("two") { bundle! "install" }
+
+          binstub = File.read(bundled_app("bin/bundle"))
+          # Replace load statement with BUNDLE_GEMFILE puts
+          binstub.gsub!(load_bundle_exe, output_bundle_gemfile)
```
segiddins: can we put these in the binstub for the gem, so that way we won't have to mess with the file we're supposed to be testing?

lib/bundler/templates/Executable.bundler
```diff
@@ -38,10 +38,39 @@ m = Module.new do
...
+
+  def find_file(*names)
+    search_up(*names) do |filename|
+      return filename if File.file?(filename)
+    end
+  end
+
+  def search_up(*names)
+    previous = nil
+    current  = File.expand_path(Dir.pwd)
```
segiddins: should this use the binstub directory instead of the current working directory?

eanlain: It should be the current working directory. If a Gemfile is not found in that current directory it should continue searching upwards to find a Gemfile, similar to [`SharedHelpers.find_gemfile`](https://github.com/bundler/bundler/blob/cecdfdb5b2a76133b0a83093ff6d80d1ffd97b46/lib/bundler/shared_helpers.rb#L240-L246).

The problem that this PR addresses is that when a bundler binstub is created and then used to install other gem executables, the bundler binstub currently stores the path of the Gemfile used in the last successful `bundle install`. Trying to run `bundle exec` when bundling against multiple gemfiles could result in a `Gem::Exception` or `LoadError` because bundler may not be using the correct Gemfile path.

From #6162:

> ...`bundle exec foo` should _always_ find the Gemfile of the current directory (when BUNDLE_GEMFILE is not set) and then run the `foo` for that Gemfile.

spec/commands/binstubs_spec.rb
eanlain: 🤔 I don't believe so... I tried a few different ways, but `ENV['BUNDLE_GEMFILE']` is `nil` outside of the bundler binstub. It is [temporary assigned](https://github.com/bundler/bundler/blob/cecdfdb5b2a76133b0a83093ff6d80d1ffd97b46/lib/bundler/templates/Executable.bundler#L71) within the scope of the binstub execution.

lib/bundler/templates/Executable.bundler
segiddins: but this isn't `bundle exec foo`, this is `bin/foo` ?

eanlain: I'm not quite sure I follow... 

This particular template, `lib/bundler/templates/Executable.bundler`, sets up a special bundler binstub that ensures the correct version of bundler is activated ([commit](https://github.com/bundler/bundler/commit/01fe509b27b317dd4d640beb9140ab31331ea349)). When `bundle exec foo` is run, this binstub is executed and [loads a version of bundler](https://github.com/bundler/bundler/blob/ba49ed283fa20d313f95cdefcd32f8f82a786c9f/lib/bundler/templates/Executable.bundler#L104).

The bundler binstub sets `ENV['BUNDLE_GEMFILE']`, which is then passed along to bundler. At this point, if `ENV['BUNDLE_GEMFILE']` is pointing to a gemfile that doesn't include `foo`, a `Gem::Exception` or `LoadError` may be raised even if `bin/foo` exists.

Not sure if this helps, but it may be easier to follow through this [Bundler repro script for bundler issue 6154](https://gist.github.com/eanlain/83f97f919701c864fb68dfdcf813cf5a).

segiddins: correct, but you quoted me documentation for `bundle exec`, and how _it_ should find a gemfile. Shouldn't the `bundle binstubs` binstubs _always_ refer to the Gemfile they were generated with, even if executed from another directory?

eanlain: I see what you're saying... yes, I agree, except in the case of bundler itself (`bin/bundle`). The binstub for bundle shouldn't refer to the Gemfile that it was generated with as it runs into the previously mentioned issues.

segiddins: the only thing is that the other binstubs all use the same logic as the `bundle` binstubs when it comes to this, iirc

indirect:
Let me see if I understand the problem, and then see what kind of options we have.

The change in 1.16 was that we started generating a `bin/bundle` and calling it from the other binstubs. The idea at the time was that it would be beneficial to be able to share some functionality across binstubs, including activating the exact version of bundler used by the given application. It also seemed like an improvement because it guaranteed you could run any binstub from any directory, and load the correct Gemfile needed for that application. 

As far as I can tell, this caused two new problems:
1. Docker (and other users) used to be able to treat bundler binstubs as if they were rubygems binstubs, not tied to any application, and usable on any ruby where that gem was installed. This change made that no longer true, and tied each binstub a to a specific application and Gemfile.
2. Previously, `bundle exec CMD` would find the Gemfile based on the CWD, and even if Bundler exec’ed a binstub from the PATH that was tied to a different application, that was fine, because binstubs were fully generic. It sounds like the `bin/bundle` binstub no longer does that, and instead when you run `bin/bundle exec rspec`, you get `bin/rspec`, which is tied to a specific application and Gemfile.

Behavior 1 (Bundler binstubs being usable from any ruby or any bundle) was never intentional, but seems to have become unexpectedly popular. For users who need that behavior, the current workaround is to use the binstubs in $BUNDLE\_PATH/gems/bin instead. Those binstubs remain fully generic, and can be used in any application without locking to a specific Gemfile.

In the long run, we might need to provide a command or flag that allows generating application-independent binstubs for users who need those.

Behavior 2 seems like an unintentional regression—it seems to me like `bundle exec rspec` should always search for a Gemfile based on the CWD, and should not simply run an application-locked binstub.

Does that seem accurate? Did I miss anything?

If that’s an accurate summary, I suggest we resolve these issues like this:

1. Change the `binstubs` command back to generating generic binstubs by default, with a flag to generate application-locked binstubs.
2. Then, change the default value of the flag in Bundler 2.0, so that application-locked binstubs become the default and the flag can be used to generate the old rubygems-style binstubs.
3. Update the docs to make it clear that application-locked binstubs (including `bin/bundle`) should not be added to your PATH unless you explicitly want to be locked in to that application’s bundle.

Does that seem like it would work? I’d love to hear any feedback or suggestions. 

/cc @eanlain @segiddins

eanlain:
@indirect your synopsis of the issues is correct. In regards to resolving these issues, I think the steps that you outlined sound like a solid plan 👍 

I'm more than happy to make those changes (as another PR).

segiddins:
That sounds good to me!

PR #6474, opened 2018-04-05
Fix bundle lock when default gemfile is present

### LINKED ISSUE #6460: If `Gemfile.lock` exists, `bundle lock --lockfile=<path>` does not do what you expect
On a project with an existing `Gemfile.lock`, when I run

```bash
bundle lock --lockfile=AlternativeGemfile.lock
```

I get the output

```
Writing lockfile to /home/deivid/Code/nipanipa/AlternativeGemfile.lock
```

However, no `AlternativeGemfile.lock` file is actually written.

If I remove the `Gemfile.lock` file, then things work as I expect.

agrim123:

The problem was on running `bundle lock --lockfile=AlternativeGemfile.lock` if a default lockfile already exists then `AlternativeGemfile.lock` is not created.

My diagnosis was that the [lock](https://github.com/bundler/bundler/blob/master/lib/bundler/definition.rb#L340) function does not check the file but for contents, so a new file is not created in case of an existing lockfile.

My fix was to check for the file existence.

Closes #6460

### DIFF
--- a/spec/commands/lock_spec.rb
+++ b/spec/commands/lock_spec.rb
@@ -97,6 +97,14 @@ RSpec.describe "bundle lock" do
     expect { read_lockfile }.to raise_error(Errno::ENOENT)
   end
 
+  it "writes to custom location using --lockfile when a default lockfile is present" do
+    bundle "install"
+    bundle "lock --lockfile=lock"
+
+    expect(out).to match(/Writing lockfile to.+lock/)
+    expect(read_lockfile("lock")).to eq(@lockfile)
+  end
+
   it "update specific gems using --update" do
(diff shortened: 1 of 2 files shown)


spec/commands/lock_spec.rb
```diff
@@ -97,6 +97,14 @@ def read_lockfile(file = "Gemfile.lock")
     expect { read_lockfile }.to raise_error(Errno::ENOENT)
   end
 
+  it "writes to custom location using --lockfile when a default lockfile is present", :focus do
```
segiddins: please remove the `:focus`

agrim123: Done.

segiddins:
@bundlerbot r+

PR #6476, opened 2018-04-06
Document parameters and return value of Injector#inject

tobias-grasse:
Summary: Use File.open instead of Kernel.open to avoid private method call when using Bundler::Injector in Ruby script.

My app should allow users to install extensions at runtime. When using `Bundler::Injector`'s `inject` in a Ruby script, the internal method `append_to` fails with the error below.

Calling system('bundle inject GEM VERSION') is not feasible in my use case, since I want to inject the gem in a secondary Gemfile.

The error that occurs when trying to inject a dependency through a call to `Bundler::Injector.inject`:

`bundler-1.16.1/lib/bundler/injector.rb:85:in 'append_to': private method `open' called for "Gemfile.local":String (NoMethodError)`

My fix replaces Kernel.open with File.open to open the Gemfile.

I chose this fix because it does not break current `bundle inject GEM VERSION` CLI behavior and does not introduce additional dependencies.

### DIFF
--- a/lib/bundler/injector.rb
+++ b/lib/bundler/injector.rb
@@ -12,6 +12,9 @@ module Bundler
       @options = options
     end
 
+    # @param [Pathname] gemfile_path The Gemfile in which to inject the new dependency.
+    # @param [Pathname] lockfile_path The lockfile in which to inject the new dependency.
+    # @return [Array]
     def inject(gemfile_path, lockfile_path)
       if Bundler.frozen_bundle?
         # ensure the lock and Gemfile are synced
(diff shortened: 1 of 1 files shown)


segiddins:
That method is expecting a pathname, which does have a public open method

tobias-grasse:
@segiddins Whoops, I guess I just outed myself as a Ruby newbie here 🙈 
Didn't know about Ruby's `Pathname` class, I assumed that passing `File.absolute_path` (which produces a String) would be sufficient. I've updated the PR to add RDoc parameter info to make the Pathname type explicit.

olleolleolle:
@tobias-grasse Cool change! (Suggested PR title: "[docs] Injector#inject parameter list")

colby-swandale:
This looks good to merge, just need to remove the first commit as that's no longer being used.

tobias-grasse:
Just coming back to this – thanks for merging and the title update @colby-swandale !

PR #6477, opened 2018-04-06
Support setting up a bundle without loading most of Bundler

metcalf:
While working on our internal developer CLIs at Stripe I noticed that a simple `require 'bundler'` was adding hundreds of milliseconds to our commands. I wrote something that copies the behavior of `Bundler.setup` using information cached in a file and refreshed when the Gemfile changes. This has made our CLIs much more responsive but I had to resort to awful monkeypatching to keep this working when RubygemsIntegration added a dependency on `Bundler.feature_flags`.

This PR makes the portions of RubygemsIntegration used in `Bundler.setup` runnable without loading the majority of the Bundler codebase. It also moves portions of the `setup` logic that don't interact with the rest of Bundler into RubygemsIntegration. This makes it easier to use RubygemsIntegration without the time penalty of requiring all of Bundler.

Making RubygemsIntegration.setup runnable without the rest of Bundler is a step towards making something like Bundler's standalone feature load much faster. I think it'd be great to get to a place where most of the Bundler codebase does not need to be loaded for a simple `bundle exec`.

### DIFF
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -48,6 +48,23 @@ module Bundler
       Gem.loaded_specs[name]
     end
 
+    def setup(specs, opts = {})
+      replace_entrypoints(specs, opts[:make_kernel_gem_public])
+      load_paths = specs.map do |spec|
+        mark_loaded(spec)
+        spec.load_paths.reject {|path| $LOAD_PATH.include?(path) }
+      end.reverse.flatten
+
+      # See Gem::Specification#add_self_to_load_path (since RubyGems 1.8)
+      if insert_index = load_path_insert_index
+        # Gem directories must come after -I and ENV['RUBYLIB']
+        $LOAD_PATH.insert(insert_index, *load_paths)
+      else
+        # We are probably testing in core, -I and RUBYLIB don't apply
+        $LOAD_PATH.unshift(*load_paths)
+      end
+    end
+
     def mark_loaded(spec)
       if spec.respond_to?(:activated=)
         current = Gem.loaded_specs[spec.name]
@@ -363,7 +380,7 @@ module Bundler
       false
     end
 
-    def replace_gem(specs, specs_by_name)
+    def replace_gem(specs, specs_by_name, make_kernel_gem_public)
       reverse_rubygems_kernel_mixin
 
       executables = nil
@@ -405,7 +422,7 @@ module Bundler
         end
 
         # backwards compatibility shim, see https://github.com/bundler/bundler/issues/5102
-        kernel_class.send(:public, :gem) if Bundler.feature_flag.setup_makes_kernel_gem_public?
+        kernel_class.send(:public, :gem) if make_kernel_gem_public
       end
     end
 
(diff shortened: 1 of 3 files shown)


metcalf:
The CI failures are consistent with several recent PRs so I think it's an issue with master.

indirect:
I definitely like the idea of being able to run setup without loading the rest of Bundler! @segiddins, do you have any thoughts on this specific approach?

lib/bundler/rubygems_integration.rb
```diff
@@ -48,6 +48,23 @@ def loaded_specs(name)
       Gem.loaded_specs[name]
     end
 
+    def setup(specs, opts = {})
```
segiddins: imho this shouldn't be a method on `RubygemIntegration` since it doesn't map to a method in rubygems

segiddins (commented):
I'm sorry, I really don't understand the purpose of this. `-rbundler/setup` always uses the Runtime class to set things up... this looks like it's trying to introduce a new public API into Bundler that bundler itself doesn't need? And for that I think we should have an RFC

lib/bundler/rubygems_integration.rb
metcalf: Sorry, I didn't realize that was a design goal here. There are already quite a few methods on here that don't seem to map 1-1 to a method on Gem, including most of the methods that this calls (e.g. `mark_loaded`, `replace_*`). Is the eventual goal to split those out of here to make this a more pure wrapper around the Gem APIs?

metcalf:
I agree that the long term goal here would require a new public API and corresponding RFC. I talked with @indirect about this a bit at RubyConf and he suggested that he was interested in being able to run `setup` without loading the rest of Bundler but that he'd prefer to address that larger question after rolling out 2.0. My hope was that by enforcing the ability to run these parts of RubygemsIntegration without requiring the rest of Bundler, it would become easier to pursue a new public API in the future.

If it helps the discussion, I could break this into two separate PRs. The first would handle taking the feature flag as a parameter and enforcing the independence of RubygemsIntegration from loading the rest of Bundler. The second would create the `setup` method and move over those pieces of logic from `Runtime`. I believe those are entirely independent changes and one could land but not the other.

lib/bundler/runtime.rb
```diff
@@ -20,29 +20,20 @@ def setup(*groups)
...
-      Bundler.rubygems.replace_entrypoints(specs)
 
-      # Activate the specs
-      load_paths = specs.map do |spec|
+      specs.each do |spec|
         unless spec.loaded_from
           raise GemNotFound, "#{spec.full_name} is missing. Run `bundle install` to get it."
         end
 
         check_for_activated_spec!(spec)
```
deivid-rodriguez: This method seems very dependent on rubygems, why not moving everything to the new `setup` method?

deivid-rodriguez (commented):
I like this patch, it makes this part of bundler more easily loadable in isolation.

I also agree with the idea that we don't yet need a new public API of RFC for this right now. I consider this just an internal refactoring, which happens to help people like @metcalf load their scripts faster, which is good :+1:.

PR #6478, opened 2018-04-07
Fix specs against RubyGems master

segiddins:

The problem was the specs on master are broken running against RubyGems master

My diagnosis was This was broken by RubyGems changing how requirements are sorted.

My fix uses `Gem::Requirement` to get the same sorting rubygems has

I chose this fix because the sorting is different between rubygems versions


segiddins:
Merging to unblock others

@bundlerbot r+

Issue #6461, opened 2018-03-27
bundler blows up when running on readonly device


grosser:
(I'm using the Bundler::CLI module here to make it easier to reproduce, otherwise would need a Gemfile)

```
docker run --read-only ruby ruby -r bundler/cli -e "puts Bundler::VERSION; Bundler::CLI.start(['exec', 'ruby', '-v'], :debug => true)"
/usr/local/lib/ruby/site_ruby/2.5.0/bundler.rb:193:in `rescue in tmp_home_path': `/root` is not writable. (Bundler::GenericSystemCallError)
Bundler also failed to create a temporary home directory at `/tmp/bundler/home':
There was an error accessing `/tmp/bundler/home`.
The underlying system error is Errno::EROFS: Read-only file system @ dir_s_mkdir - /tmp/bundler
	from /usr/local/lib/ruby/site_ruby/2.5.0/bundler.rb:181:in `tmp_home_path'
	from /usr/local/lib/ruby/site_ruby/2.5.0/bundler.rb:172:in `user_home'
	from /usr/local/lib/ruby/site_ruby/2.5.0/bundler.rb:197:in `user_bundle_path'
	from /usr/local/lib/ruby/site_ruby/2.5.0/bundler/plugin.rb:99:in `global_root'
	from /usr/local/lib/ruby/site_ruby/2.5.0/bundler/plugin/index.rb:73:in `global_index_file'
	from /usr/local/lib/ruby/site_ruby/2.5.0/bundler/plugin/index.rb:32:in `initialize'
	from /usr/local/lib/ruby/site_ruby/2.5.0/bundler/plugin.rb:78:in `new'
	from /usr/local/lib/ruby/site_ruby/2.5.0/bundler/plugin.rb:78:in `index'
	from /usr/local/lib/ruby/site_ruby/2.5.0/bundler/plugin.rb:114:in `command?'
	from /usr/local/lib/ruby/site_ruby/2.5.0/bundler/cli.rb:130:in `handle_no_command_error'
	from /usr/local/lib/ruby/site_ruby/2.5.0/bundler/vendor/thor/lib/thor/command.rb:31:in `run'
	from /usr/local/lib/ruby/site_ruby/2.5.0/bundler/vendor/thor/lib/thor/command.rb:128:in `run'
	from /usr/local/lib/ruby/site_ruby/2.5.0/bundler/vendor/thor/lib/thor/invocation.rb:126:in `invoke_command'
[...]

PR #6482, opened 2018-04-11
Do not blow up when installing on a readonly filesystem without a hom…

grosser:
…e directory

fixes https://github.com/bundler/bundler/issues/6461

```
docker run --read-only ruby ruby -r bundler/cli -e "puts Bundler::VERSION; Bundler::CLI.start(['exec', 'ruby', '-v'], :debug => true)"
/usr/local/lib/ruby/site_ruby/2.5.0/bundler.rb:193:in `rescue in tmp_home_path': `/root` is not writable. (Bundler::GenericSystemCallError)
Bundler also failed to create a temporary home directory at `/tmp/bundler/home':
There was an error accessing `/tmp/bundler/home`.
The underlying system error is Errno::EROFS: Read-only file system @ dir_s_mkdir - /tmp/bundler
  from /usr/local/lib/ruby/site_ruby/2.5.0/bundler.rb:181:in `tmp_home_path'
  from /usr/local/lib/ruby/site_ruby/2.5.0/bundler.rb:172:in `user_home'
  from /usr/local/lib/ruby/site_ruby/2.5.0/bundler.rb:197:in `user_bundle_path'
  from /usr/local/lib/ruby/site_ruby/2.5.0/bundler/plugin.rb:99:in `global_root'
  from /usr/local/lib/ruby/site_ruby/2.5.0/bundler/plugin/index.rb:73:in `global_index_file'
  from /usr/local/lib/ruby/site_ruby/2.5.0/bundler/plugin/index.rb:32:in `initialize'
  from /usr/local/lib/ruby/site_ruby/2.5.0/bundler/plugin.rb:78:in `new'
```

@segiddins

### DIFF
--- a/lib/bundler/plugin/index.rb
+++ b/lib/bundler/plugin/index.rb
@@ -29,7 +29,12 @@ module Bundler
         @hooks = {}
         @load_paths = {}
 
-        load_index(global_index_file, true)
+        begin
+          load_index(global_index_file, true)
+        rescue GenericSystemCallError
+          # no need to fail when on a read-only FS, for example
+          nil
+        end
         load_index(local_index_file) if SharedHelpers.in_bundle?
       end
 
--- a/spec/bundler/plugin/index_spec.rb
+++ b/spec/bundler/plugin/index_spec.rb
@@ -175,4 +175,12 @@ RSpec.describe Bundler::Plugin::Index do
       include_examples "it cleans up"
     end
   end
+
+  describe "readonly disk without home" do
+    it "ignores being unable to create temp home dir" do
+      expect_any_instance_of(Bundler::Plugin::Index).to receive(:global_index_file).
+        and_raise(Bundler::GenericSystemCallError.new("foo", "bar"))
+      Bundler::Plugin::Index.new
+    end
+  end
 end


lib/bundler/plugin/index.rb
```diff
@@ -29,7 +29,12 @@ def initialize
         @hooks = {}
         @load_paths = {}
 
-        load_index(global_index_file, true)
+        begin
+          load_index(global_index_file, true)
+        rescue GenericSystemCallError
+          # no need to fail when on a read-only FS, for example
+          nil
```
colby-swandale: This should just rescue in the method, no need to have a `begin .. end` block.

lib/bundler/plugin/index.rb
```diff
@@ -29,7 +29,12 @@ def initialize
         @hooks = {}
         @load_paths = {}
 
-        load_index(global_index_file, true)
+        begin
+          load_index(global_index_file, true)
+        rescue GenericSystemCallError
```
colby-swandale: We probably just want to catch `Errno::EROFS` instead of `GenericSystemCallError` because there are errors we do want to catch, such as when the user not have the correct permissions, disk if full etc.

lib/bundler/plugin/index.rb
grosser: I'd prefer it being precise, this way nothing is caught by accident and it's more obvious why/what is being guarded, if the code ever needs moving nobody needs to go through the history and figure out what is being rescued

lib/bundler/plugin/index.rb
grosser: the error being surfaced is Bundler::GenericSystemCallError, rescuing anything else won't work ?

colby-swandale: Oh, i got confused between `GenericSystemCallError` and `SystemCallError`, sorry.

lib/bundler/plugin/index.rb
colby-swandale: fair enough

colby-swandale:
Thanks! @bundlerbot r+

PR #6488, opened 2018-04-14
Fix failing exec spec that is blocking 1.16.2

colby-swandale:
### Overview

There is a test that is [failing in travis](https://travis-ci.org/bundler/bundler/builds/365371973) which is preventing v1.16.2 from being released. The spec in question is checking the `BUNDLER_SPEC_SUB_VERSION` env var and changes the assertion of the test depending on the stubbed version. The issue is that the 1-16-stable branch does not really use `BUNDLER_SPEC_SUB_VERSION` and the spec is assuming it's testing Bundler 2.

Checking for `BUNDLER_SPEC_SUB_VERSION` is not really necessary in the first place because the test is already scoped for Bundler 1 & 2 (using rspec tags). This PR removes the checking for `BUNDLER_SPEC_SUB_VERSION﻿`.


segiddins:
@bundlerbot r+

PR #6490, opened 2018-04-17
Filter git uri credentials in source description

segiddins:

The problem was HTTP basic auth credentials were leaking into Bundler's output when used in git sources

My diagnosis was we needed to filter credentials in `Git#to_s`

I chose this fix because it doesn't require updating every place that uses `Source#to_s`, and is symmetric with what the rubygems source does to filter creds


colby-swandale:
Great! thanks! @bundlerbot r+

PR #6493, opened 2018-04-18
Add documentation for --all flag of bundle update

agrim123:

`bundle update` throws an error if run without any flags. `--all` flag was introduced to update all gems instead of no flags. This was not mentioned in docs.


colby-swandale:
Thanks! @bundlerbot r+

Issue #6491, opened 2018-04-17
Strange lockfile behaviour with bundle update/install and platform-specific gems


nilsding:
Hey everyone,
I recently stumbled upon some weird behaviour with `bundle update` and `bundle install` and some platform-specific gem dependencies in a few of our projects.  I successfully managed to track it down to the following:

- The `pry` gem, which has at least two gem versions: one for any platform (I'll call that one `ruby` from now on) and one for `java`
- That `java` platform version has an additional dependency on `spoon`, which has a dependency on `ffi`
- Any bundle command which modifies the Gemfile which is not `bundle install` (e.g. `bundle update` or `bundle lock`) adds the `ffi` dependency for the `ruby` platform.
- `bundle install` will correctly remove the `ffi` gem for the `ruby` platform from the `Gemfile.lock`, as it determined it's apparently not needed for that platform.

I have created a git repository with each step in its own commit to reproduce this behaviour: https://github.com/nilsding/bundler-lockfile-issue/commits/master.

----

Output of `bundle env`, if it's of any help:

## Environment

```
Bundler             1.16.1
  Platforms         ruby, universal-java-1.8
Ruby                2.3.3p0 (2018-02-21 revision 56859) [java]
  Full Path         /Users/nilsding/.rvm/rubies/jruby-9.1.16.0/bin/jruby
  Config Dir        /Users/nilsding/.rvm/rubies/jruby-9.1.16.0/etc
RubyGems            2.6.14.1
  Gem Home          /Users/nilsding/.rvm/gems/jruby-9.1.16.0
  Gem Path          /Users/nilsding/.rvm/gems/jruby-9.1.16.0:/Users/nilsding/.rvm/gems/jruby-9.1.16.0@global
  User Path         /Users/nilsding/.gem/jruby/2.3.0
  Bin Dir           /Users/nilsding/.rvm/gems/jruby-9.1.16.0/bin
Tools
  Git               2.16.2
  RVM               1.29.3 (master)
  rbenv             not installed
  chruby            not installed
  rubygems-bundler  (1.4.4)
```

## Bundler Build Metadata

```
Built At          2017-12-21
Git SHA           0034ef341
Released Version  true
```

## Bundler settings

```
jobs
  Set for the current user (/Users/nilsding/.bundle/config): "8"
gem.test
  Set for the current user (/Users/nilsding/.bundle/config): "rspec"
gem.mit
  Set for the current user (/Users/nilsding/.bundle/config): false
gem.coc
[...]

PR #6495, opened 2018-04-19
[Definition] Filter out unneeded gem platforms after resolving

segiddins:

The problem was the lockfile would contain platform-specific gems that it didn't need

Closes https://github.com/bundler/bundler/issues/6491

My diagnosis was the resolver sometimes activates platforms it doesn't need, since it can't know it won't need them

My fix is to use `SpecSet#for` to filter out gems that won't ever be used

I chose this fix because it isn't re-inventing the wheel!

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -245,17 +245,22 @@ module Bundler
     def resolve
       @resolve ||= begin
         last_resolve = converge_locked_specs
-        if Bundler.frozen_bundle?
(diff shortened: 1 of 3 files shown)


ThomasKoppensteiner:
Any updates on this? Is this PR merge-ready?

segiddins:
It’s just awaiting review

segiddins:
@bundlerbot r+

PR #6498, opened 2018-04-20
Fix calling undefined `Gem::Util.inflate` in RubyGemsIntegration for old RubyGems versions

colby-swandale:

RubyGems integration is calling `Gem::Util.inflate` which is not available in all supported versions of RubyGems.

See https://travis-ci.org/bundler/bundler/jobs/368919903

Check if `Gem::Util` is defined, and if not, fallback to using the old way calling `inflate` in RubyGems


colby-swandale:
@bundlerbot r+

Issue #6465, opened 2018-03-29
`require 'bundler/setup'` fails on 2.5.1 if bundler is located in symlinked directory


ojab:
I don't really know if the root cause is symlinked directory, but it's my best shot (and I think [this change](https://bugs.ruby-lang.org/issues/14557) can be somehow related).
ruby-2.5.1 installed via rvm, rvm directory with all gems located in directory under symlink (i. e. `/symlinked_directory/ -> /real_directory/`, rvm is placed in `/real_directory/.rvm`, but `$LOAD_PATH` contains `/symlinked_directory/.rvm/…`)

`require 'bundler/setup'` in any project with `Gemfile.lock` fails with various `LoadError`s.

In empty project:
```
$ mkdir testapp && cd testapp && bundle init && irb
Writing new Gemfile to /tmp/testapp/Gemfile
2.5.1 :001 > require 'bundler/setup'
Traceback (most recent call last):
       16: from (irb):1
       15: from /real_directory/.rvm/rubies/ruby-2.5.1/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:39:in `require'
       14: from /real_directory/.rvm/rubies/ruby-2.5.1/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:135:in `rescue in require'
       13: from /real_directory/.rvm/rubies/ruby-2.5.1/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:135:in `require'
       12: from /real_directory/.rvm/gems/ruby-2.5.1/gems/bundler-1.16.1/lib/bundler/setup.rb:10:in `<top (required)>'
       11: from /real_directory/.rvm/gems/ruby-2.5.1/gems/bundler-1.16.1/lib/bundler.rb:107:in `setup'
       10: from /real_directory/.rvm/gems/ruby-2.5.1/gems/bundler-1.16.1/lib/bundler/runtime.rb:20:in `setup'
        9: from /real_directory/.rvm/gems/ruby-2.5.1/gems/bundler-1.16.1/lib/bundler/runtime.rb:108:in `block in definition_method'
        8: from /real_directory/.rvm/gems/ruby-2.5.1/gems/bundler-1.16.1/lib/bundler/definition.rb:226:in `requested_specs'
        7: from /real_directory/.rvm/gems/ruby-2.5.1/gems/bundler-1.16.1/lib/bundler/definition.rb:237:in `specs_for'
[...]

PR #6502, opened 2018-04-23
Use realpath in clean_load_path

ojab:
see #6465, basically we're not rejecting `bundler_lib` because we have symlink there and real path in `$LOAD_PATH`
```
$ irb
2.5.1 :001 > require 'bundler/setup'
# what happens next will shock you
From: /real_path/.rvm/gems/ruby-2.5.1/gems/bundler-1.16.2/lib/bundler/shared_helpers.rb @ line 339 Bundler::SharedHelpers#clean_load_path:

    330: def clean_load_path
    331:   # handle 1.9 where system gems are always on the load path
    332:   return unless defined?(::Gem)
    333:
    334:   bundler_lib = bundler_ruby_lib
    335:
    336:   loaded_gem_paths = Bundler.rubygems.loaded_gem_paths
    337:
    338:   binding.pry
 => 339:   $LOAD_PATH.reject! do |p|
    340:     path = File.expand_path(p)
    341:     path = File.realpath(path) if File.exist?(path)
    342:     next if path.start_with?(bundler_lib)
    343:     loaded_gem_paths.delete(p)
    344:   end
    345:   $LOAD_PATH.uniq!
    346: end

[1] pry(#<Bundler::Runtime>)> bundler_lib
=> "/real_path/.rvm/gems/ruby-2.5.1/gems/bundler-1.16.2/lib"
[2] pry(#<Bundler::Runtime>)> loaded_gem_paths
=> ["/symlinked_path/.rvm/gems/ruby-2.5.1@global/gems/did_you_mean-1.2.0/lib",
 "/symlinked_path/.rvm/gems/ruby-2.5.1/gems/bundler-1.16.2/lib",
 "/symlinked_path/.rvm/gems/ruby-2.5.1/gems/byebug-10.0.2/lib",
 "/symlinked_path/.rvm/gems/ruby-2.5.1/extensions/x86_64-linux/2.5.0/byebug-10.0.2",
 "/symlinked_path/.rvm/gems/ruby-2.5.1/gems/coderay-1.1.2/lib",
 "/symlinked_path/.rvm/gems/ruby-2.5.1/gems/method_source-0.9.0/lib",
 "/symlinked_path/.rvm/gems/ruby-2.5.1/gems/pry-0.11.3/lib",
 "/symlinked_path/.rvm/gems/ruby-2.5.1/gems/pry-byebug-3.6.0/lib",
 "/real_path/.rvm/rubies/ruby-2.5.1/lib/ruby/gems/2.5.0/gems/fileutils-1.0.2/lib",
 "/real_path/.rvm/rubies/ruby-2.5.1/lib/ruby/gems/2.5.0/gems/psych-3.0.2/lib",
 "/real_path/.rvm/rubies/ruby-2.5.1/lib/ruby/gems/2.5.0/extensions/x86_64-linux/2.5.0/psych-3.0.2"]
```

### DIFF
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -344,9 +344,10 @@ module Bundler
 
     def resolve_path(path)
       expanded = File.expand_path(path)
+      return expanded unless File.respond_to?(:realpath)
 
-      if File.respond_to?(:realpath)
-        expanded = File.realpath(expanded) until expanded == File.realpath(expanded)
+      while File.exist?(expanded) && File.realpath(expanded) != expanded
+        expanded = File.realpath(expanded)
       end
 
       expanded
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -336,7 +336,11 @@ module Bundler
       loaded_gem_paths = Bundler.rubygems.loaded_gem_paths
 
       $LOAD_PATH.reject! do |p|
-        next if File.expand_path(p).start_with?(bundler_lib)
+        expanded_p = File.expand_path(p)
+        next if expanded_p.start_with?(bundler_lib)
+        if File.exist?(expanded_p) && File.respond_to?(:realpath)
+          next if File.realpath(expanded_p).start_with?(bundler_lib)
+        end
         loaded_gem_paths.delete(p)
       end
       $LOAD_PATH.uniq!
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -344,7 +344,11 @@ module Bundler
 
     def resolve_path(path)
       expanded = File.expand_path(path)
-      expanded = File.realpath(expanded) until expanded == File.realpath(expanded)
+
+      if File.respond_to?(:realpath)
+        expanded = File.realpath(expanded) until expanded == File.realpath(expanded)
+      end
+
       expanded
     end
 
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -324,7 +324,7 @@ module Bundler
     end
 
     def bundler_ruby_lib
-      File.expand_path("../..", __FILE__)
+      resolve_path File.expand_path("../..", __FILE__)
     end
 
     def clean_load_path
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -336,14 +336,16 @@ module Bundler
       loaded_gem_paths = Bundler.rubygems.loaded_gem_paths
 
       $LOAD_PATH.reject! do |p|
-        expanded_p = File.expand_path(p)
-        next if expanded_p.start_with?(bundler_lib)
-        if File.exist?(expanded_p) && File.respond_to?(:realpath)
-          next if File.realpath(expanded_p).start_with?(bundler_lib)
-        end
+        next if resolve_path(p).start_with?(bundler_lib)
         loaded_gem_paths.delete(p)
       end
       $LOAD_PATH.uniq!
+    end
+
+    def resolve_path(path)
+      expanded = File.expand_path(path)
+      expanded = File.realpath(expanded) until expanded == File.realpath(expanded)
+      expanded
     end
 
     def prints_major_deprecations?
--- a/spec/runtime/setup_spec.rb
+++ b/spec/runtime/setup_spec.rb
@@ -857,6 +857,50 @@ end
     expect(out).to eq("true\ntrue")
   end
 
+  context "with bundler is located in symlinked GEM_HOME" do
+    let(:gem_home) { Dir.mktmpdir }
+    let(:symlinked_gem_home) { Tempfile.new("gem_home") }
+    let(:bundler_dir) { File.expand_path("../../..", __FILE__) }
+    let(:bundler_lib) { File.join(bundler_dir, "lib") }
+
+    before do
+      FileUtils.ln_sf(gem_home, symlinked_gem_home.path)
+      gems_dir = File.join(gem_home, "gems")
+      specifications_dir = File.join(gem_home, "specifications")
+      Dir.mkdir(gems_dir)
+      Dir.mkdir(specifications_dir)
+
+      FileUtils.ln_s(bundler_dir, File.join(gems_dir, "bundler-#{Bundler::VERSION}"))
+
+      gemspec = File.read("#{bundler_dir}/bundler.gemspec").
+                sub("Bundler::VERSION", %("#{Bundler::VERSION}"))
+      gemspec = gemspec.lines.reject {|line| line =~ %r{lib/bundler/version} }.join
+
+      File.open(File.join(specifications_dir, "bundler.gemspec"), "wb") do |f|
+        f.write(gemspec)
+      end
+    end
+
+    it "should succesfully require 'bundler/setup'" do
+      install_gemfile ""
+
+      ENV["GEM_PATH"] = symlinked_gem_home.path
+
+      ruby <<-R
+        if $LOAD_PATH.include?("#{bundler_lib}")
+          # We should use bundler from GEM_PATH for this test, so we should
+          # remove path to the bundler source tree
+          $LOAD_PATH.delete("#{bundler_lib}")
+        else
+          raise "We don't have #{bundler_lib} in $LOAD_PATH"
+        end
+        puts (require 'bundler/setup')
+      R
+
+      expect(out).to eql("true")
+    end
+  end
+
   it "stubs out Gem.refresh so it does not reveal system gems" do
     system_gems "rack-1.0.0"
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ addons:
       secure: "TrzIv116JLGUxm6PAUskCYrv8KTDguncKROVwbnjVPKTGDAgoDderd8JUdDEXrKoZ9qGLD2TPYKExt9/QDl71E+qHdWnVqWv4HKCUk2P9z/VLKzHuggOUBkCXiJUhjywUieCJhI3N92bfq2EjSBbu2/OFHqWOjLQ+QCooTEBjv8="
 
 rvm:
-  - 2.5.0
+  - 2.5.1
   - 2.4.3
   - 2.3.6
 
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -31,8 +31,9 @@ require "digest"
 # Delete any copies of Bundler that have been dumped into site_ruby without
 # a gemspec. RubyGems cannot manage that Bundler, and so our tricks to make
 # sure that the correct version of Bundler loads will stop working.
-Dir.glob(File.join(RbConfig::CONFIG["sitelibdir"], "bundler*")).each do |f|
-  FileUtils.rm_rf f
+require "fileutils"
+Dir.glob(File.join(RbConfig::CONFIG["sitelibdir"], "bundler*")).each do |file|
+  FileUtils.rm_rf(file)
 end
 
 if File.expand_path(__FILE__) =~ %r{([^\w/\.-])}
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -12,13 +12,6 @@ module Gem
   end
 end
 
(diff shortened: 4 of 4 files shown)


colby-swandale:
This needs a test

colby-swandale:
This will need a test

alexdean:
@ojab i took a pass at writing some tests for this, but i'm not sure i'm on the right track. (not clear on what should be in `$LOAD_PATH` when this method finishes.)

ojab:
Sorry for delay.
+ rebased to master
+ added spec that fails on ruby-2.5.1 + bundler master and passes on ruby-2.5.1 and bundler from my branch

So the question is: what exactly we should test? Right now I'm setting up symlinked RUBY_HOME w/ bundler and requiring 'bundler/setup' (and it looks quite ugly tbh, suggestions are welcome). 

1. Do we need to test that `clean_load_path` actually removes `bundler_lib` from `$LOAD_PATH`?
2. What should we do with `Bundler.rubygems.loaded_gem_paths`? In my test setup `$LOAD_PATH` contains both symlinked and real path for some gems, while `Bundler.rubygems.loaded_gem_paths` contains real paths only for system gems (I don't know if "system" is proper term here), such as `fiddle`/`fileutils`/`json`/etc, all other have symlinked paths. So should we convert it all to the real paths during `clean_load_path` or we're already ok here? Everything works as is AFAICS, but I don't know if we have can of worms here.
3. Should I add some missing directories to `$LOAD_PATH` that I'm testing? If so — is it ok to add it to the same `it` block or better add another one (I don't know preferred style in bundler)?
4. Right now travis testing 2.5.0, should I bump it to 2.5.1 (and the same for 2.3/2.4 rubies)? If so — do we need to retain 2.5.0, considering behaviour change and unknown reason for that change? (we know when this failure is happening, but don't know why and when it actually started happening)

ojab:
5. Do we need to check that `p.start_with?(bundler_lib)` or this case is impossible?

lib/bundler/shared_helpers.rb
```diff
@@ -336,7 +336,9 @@ def clean_load_path
       loaded_gem_paths = Bundler.rubygems.loaded_gem_paths
 
       $LOAD_PATH.reject! do |p|
-        next if File.expand_path(p).start_with?(bundler_lib)
+        path = File.expand_path(p)
+        path = File.realpath(path) if File.exist?(path)
```
segiddins: this needs to also check `if File.respond_to?(:realpath)`

spec/runtime/setup_spec.rb
```diff
@@ -857,6 +857,42 @@ def clean_load_path(lp)
...
+        f.write(gemspec)
+      end
+    end
+
+    it "should succesfully require 'bundler/setup'" do
+      install_gemfile ""
+
+      ENV["GEM_PATH"] = symlinked_gem_home.to_path
+      ruby <<-R
+        $LOAD_PATH.delete_at(0)
```
segiddins: please update this to delete a particular path, with a comment why it's being deleted

spec/runtime/setup_spec.rb
```diff
@@ -857,6 +857,42 @@ def clean_load_path(lp)
...
+        puts (require 'bundler/setup')
```
segiddins: imho we should add something like `raise "missing bundler lib path in #{$LOAD_PATH}" unless $LOAD_PATH.find ...`

lib/bundler/shared_helpers.rb
```diff
@@ -336,7 +336,9 @@ def clean_load_path
...
+        next if path.start_with?(bundler_lib)
```
segiddins: `next if path.start_with?(bundler_lib) || p.start_with?(bundler_lib)` ?

ojab: `|| File.expand_path(p).start_with?(bundler_lib)`?

ojab:
Done, but ruby-2.5.0 tests fail with `You must use Bundler 2 or greater with this lockfile. (Bundler::LockfileError)` and I have no idea how to properly get rid of that.

ojab:
After some debugging I've found that default rvm installation has bundled `/home/ojab/.rvm/rubies/ruby-2.5.0/lib/ruby/site_ruby/2.5.0/bundler.rb` which is required instead of our `bundler` from source tree and without fiddling with rvm installation I can't really test it on 2.5.0. And the issue is reproducible on 2.5.1 and not on 2.5.0 exactly because of that.

Replaced ruby-2.5.0 with ruby-2.5.1 on travis, specs are passing now. 
Hope that's ok, suggestions are welcome.

ojab:
ping?

ojab:
Anyone? @indirect @colby-swandale @segiddins

colby-swandale:
We will merge this when we have time. Please be patient.

chewi:
I feel like this should be getting more urgency. I've run into it on a couple of servers and now again on my own development box. I'm certainly not alone. The patch fixed all cases for me.

alexdean:
we have delayed upgrading past ruby 2.5.0 until this is resolved.

ojab:
Same here.

colby-swandale:
@bundlerbot r+

ojab:
…and looks like the bundled bundler is back on travis. 
I have no idea what is the proper way to fix it. 
`rm -rf ~/.rvm/rubies/*/lib/ruby/site_ruby/*/bundler/`?

segiddins:
@indirect r?

indirect:
@segiddins thanks for the ping, this seems reasonable

@bundlerbot r+

indirect:
@bundlerbot retry

segiddins:
@bundlerbot retry

colby-swandale:
This breaking spec needs to be fixed before we can merge it

colby-swandale:
Just wanted to post an update saying that I've been looking into getting this test passing. I've stumbled across an interesting problem. We need to be able to tell rubygems to load a specific version of Bundler (the development version). But at the same time we can't specify a path to Bundler like we normally do with `ruby -I `, because that won't get the specific scenario setup needed to test this code.

I'll keep looking into it.

deivid-rodriguez:
@colby-swandale I got it passing with this patch on top of the current:

```diff
diff --git a/spec/runtime/setup_spec.rb b/spec/runtime/setup_spec.rb
index 4f8e70f8e..25abe81ae 100644
--- a/spec/runtime/setup_spec.rb
+++ b/spec/runtime/setup_spec.rb
@@ -885,6 +885,7 @@ end
       install_gemfile ""
 
       ENV["GEM_PATH"] = symlinked_gem_home.path
+      ENV["RUBYLIB"] = "#{symlinked_gem_home.path}/gems/bundler-#{Bundler::VERSION}/lib"
 
       ruby <<-R
         if $LOAD_PATH.include?("#{bundler_lib}")
```

deivid-rodriguez:
But then it doesn't fail without the code changes... :disappointed:

indirect:
@chewi since this is so urgent for you, can you provide a working test that we can use to verify this issue will stay fixed?

indirect:
@colby-swandale this is maybe a ridiculous idea, but can we move the RVM-bundled file `bundler.rb`, just for this test, and then move it back?

ojab:
RVM-bundled `bundler.rb` potentially breaks other tests too (i. e. it can be required instead of bundler from source tree), why it should be moved back?
The other possibility is dropping travis' rvm completely: default rvm build doesn't contain bundler, but travis' one contains bundler as default gem.

deivid-rodriguez:
In case it might be helpful:

* Rvm 1.29.4 was released yesterday after almost a year without releases: https://github.com/rvm/rvm/releases. Big release. But it'll take a while before travis upgrades, I guess, so it doesn't really affects here.

* This issue also reproduces with `rbenv`. I guess it reproduces as long as the system ruby includes `bundler` as a default gem.

indirect:
@ojab it should be moved back because we want our tests to test against real world scenarios. Today, if you run `gem install --system`, you will get Bundler as a default gem. The reason Travis includes Bundler as a default gem is that they update rubygems via `gem install --system` after installing Ruby with RVM. I expect that's also what rbenv does, and we need to make sure that Bundler works under those circumstances.

chewi:
@indirect, I have patched around this where necessary. I spoke up because I didn't want to run into it again and, more importantly, it will affect just about every RVM (and rbenv?) user out there. I would take a look but I'm not familiar with the Travis setup and I'm also away from home and work this week.

lib/bundler/shared_helpers.rb
```diff
@@ -336,7 +336,11 @@ def clean_load_path
       loaded_gem_paths = Bundler.rubygems.loaded_gem_paths
 
       $LOAD_PATH.reject! do |p|
-        next if File.expand_path(p).start_with?(bundler_lib)
+        expanded_p = File.expand_path(p)
+        next if expanded_p.start_with?(bundler_lib)
+        if File.exist?(expanded_p) && File.respond_to?(:realpath)
+          next if File.realpath(expanded_p).start_with?(bundler_lib)
```
eregon: `bundler_lib` should also be resolved with `File.realpath` (probably in `def bundler_ruby_lib` itself).
Specifically, the value of `__FILE__` is not a realpath and can contain symlinks (see https://github.com/oracle/truffleruby/issues/1383#issuecomment-403162930 for details).

eregon:
I also got this issue in TruffleRuby when installing with RVM (https://github.com/oracle/truffleruby/issues/1383#issuecomment-403162930)
Some ideas to fix this:
* First it seems a bad idea to use symlinks in GEM_PATH, and nothing resolving them to realpaths. Maybe one way to fix is to canonicalize paths from `ENV['GEM_PATH']` when reading them (probably in RubyGems). I'll file an issue for RVM so it doesn't even give symlinks in `GEM_PATH`.
* Recently, RVM is [installing bundler automatically again](https://github.com/rvm/rvm/commit/9a6bcd441ab95a626b67754dd313ce901d4ef80d). When this change lands in TravisCI it might help. (But, IIRC, TravisCI already modifies the list of `default.gems` to install bundler).
* Maybe there is a way to prevent RVM to upgrade RubyGems? This would make everything simpler. Or maybe we should just `rm -rf /home/travis/.rvm/rubies/ruby-2.5.1/lib/ruby/site_ruby/2.5.0/bundler`, since testing bundler with another version installed is going to be brittle.
Or the other way around, maybe we should install the current bundler in `site_ruby/2.5.0/bundler` for testing?
* Finally, the test could be adapted to see what's loaded. If anything is loaded from `site_ruby/2.5.0/bundler` then that's wrong as that's using another version of bundler.

eregon:
Last-resort option: skip this test on Ruby >= 2.5 until there is a good way to avoid the site_ruby bundler to get in the way. The fact it works on all other versions is already a good proof this fix helps and fixes the bug on those versions.

eregon:
@indirect Could you share your opinion on which way to go seems best?
This issue affects every RVM user for JRuby & TruffleRuby, as well as anyone having Bundler installing in a Gem directory referenced as a symlink in `GEM_PATH`.

eregon:
@indirect Thank you. I agree removing that copy on site_ruby should make everything saner for testing.

I think this copy of bundler in `site_ruby` is created by running `gem update --system` with MRI 2.5.1.
At least I can confirm `gem update --system` adds `lib/ruby/site_ruby/2.5.0/bundler` when I run it locally.
TravisCI's MRI 2.5.1 (https://rubies.travis-ci.org/ubuntu/14.04/x86_64/ruby-2.5.1.tar.bz2) already contains the `site_ruby` bundler.
I'd guess this might be due to https://github.com/travis-ci/travis-rubies/commit/e62622db40364cd293dbdc5f6aecb7d7811bd02d or RVM updating RubyGems on `rvm install ruby-2.5.1`.

lib/bundler/shared_helpers.rb
```diff
@@ -336,12 +336,23 @@ def clean_load_path
...
       end
       $LOAD_PATH.uniq!
     end
 
+    def resolve_path(path)
+      expanded = File.expand_path(path)
+      return expanded unless File.respond_to?(:realpath)
+
+      while File.exist?(expanded) && File.realpath(expanded) != expanded
+        expanded = File.realpath(expanded)
```
eregon: AFAIK File.realpath() resolves all symlinks it can, so a single call should be enough, unlike `readlink(1)`.

eregon: Thanks @ojab for making a PR to fix this :)

PR #6503, opened 2018-04-23
Use `__dir__` instead of `__FILE__` in newgem.gemspec template

koic:

Since Ruby 2.0 we've had `__dir__` as well as `__FILE__`.

The initial gem codes written with `bundle gem` using Ruby 2.0 or higher is an old description using `__FILE__`.

Ruby 1.9 is EOL, so I think that there is not much Gem to start developed using it.

This PR uses `__dir__` when starting Gem development (i.e. `bundle gem`) using Ruby 2.0 or higher version.


hsbt:
It's reasonable.

hsbt:
@bundlerbot r+

koic:
Thanks!

PR #6504, opened 2018-04-23
rubygem_push again even after a failure

kbrock:
Hi Bundle team,

I ran `rake release` but it failed while running the (sub) task `release:rubygem_push`.

I tried to run `rake release:rubygem_push` but it kept failing with a cryptic error.
I did not want to `rake release` again since that does other tasks like tagging git.

The issue seemed to come from an instance variable `built_gem_path` [[here]](https://github.com/bundler/bundler/blob/948771063c05a56af8888acf5c01171293371214/lib/bundler/gem_helper.rb#L38) that is shared across rake tasks.

The variable is set in task `build`, which is done as part of the `release` task.
Since I did not run `release` (or `build` for that matter), the gem path was not set and the command sent to a sub bash shell was invalid.

All the methods defined in the `install` method have a fallback when`build_gem_path` is not set. Well, all methods except for `rubygem_push`. I added this fallback for this method as well.

This was the least number of lines to change and seemed to mimic the style of the other methods. I consider the code around pushing a gem to be sensitive. So I wanted to introduce the least amount of risk.

An alternative solution is to remove the instance variable. It seemed the cleanest but a little over kill for the issue at hand. `build` would be the only one to overwrite the `gem` file, and the others would just use the assumed name

Thanks so much all for keeping bundler running smoothly.

### DIFF
--- a/lib/bundler/gem_helper.rb
+++ b/lib/bundler/gem_helper.rb
@@ -66,7 +66,7 @@ module Bundler
         tag_version { git_push(args[:remote]) } unless already_tagged?
       end
 
-      task "release:rubygem_push" do
+      task "release:rubygem_push" => "build" do
         rubygem_push(built_gem_path) if gem_push?
       end
 
@@ -98,6 +98,7 @@ module Bundler
   protected
 
     def rubygem_push(path)
+      path ||= build_gem
       gem_command = %W[gem push #{path}]
       gem_command << "--key" << gem_key if gem_key
       gem_command << "--host" << allowed_push_host if allowed_push_host


kbrock:
It looks like the tests do not check if the subshell command is valid (tests can't check everything)
When I change the code to create a valid `gem_command`, it is blowing up on me.

I'll have to revisit the tests to get them passing

```ruby
def rubygem_push(path) # path = nil
  gem_command = %W[gem push #{path}] # command is invalid
  gem_command << "--key" << gem_key if gem_key
  gem_command << "--host" << allowed_push_host if allowed_push_host
  unless allowed_push_host || Bundler.user_home.join(".gem/credentials").file?
    raise "Your rubygems.org credentials aren't set. Run `gem push` to set them."
  end
  sh(gem_command) # blows up (why I'm creating the pr)
  Bundler.ui.confirm "Pushed #{name} #{version} to #{gem_push_host}"
end
```

segiddins:
> I did not want to rake release again since that does other tasks like tagging git.

I believe we delete the tag if pushing to rubygems failed?

kbrock:
@segiddins thanks - I'll check

I can't get this to work. and it has only happened to me once in years so... seems good enough

PR #6513, opened 2018-04-27
Add `bundle remove`

### LINKED ISSUE #6506: Add `bundle remove [gems]`
In reference to this rfc [command](https://github.com/shime/rfcs/blob/bundle-add-improvements/text/0002-bundle-add-and-remove.md#5-removing-gems).

Basic usage:
```bash
bundle remove rack rails
```

agrim123:
Features of the command implemented:
- Multiple gems support
```bash
$ bundle remove rack rails
```
- Remove any empty block that might occur after removing the gem or otherwise present

Things yet to implement:
- Add `rm` alias. _Optional_
- [x] Add `--install` flag to remove gems from `.bundle`.
- [x] Handling multiple gems on the same line.
- [x] Handle gem spec
- [x] Handle eval_gemfile cases ([one](https://github.com/bundler/bundler/pull/6513#discussion_r195632603) case left)

Closes #6506

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -166,9 +166,9 @@ module Bundler
       Check.new(options).run
     end
 
-    desc "remove [GEMS] [OPTIONS]", "Removes the gem from the Gemfile"
+    desc "remove [GEM [GEM ...]]", "Removes gems from the Gemfile"
     long_desc <<-D
-      Removes the specified gems from the Gemfile. If the gem is not found, Bundler prints a error message and if gem could not be removed due to any reason Bundler will display a warning.
+      Removes the given gems from the Gemfile while ensuring that the resulting Gemfile is still valid. If the gem is not found, Bundler prints a error message and if gem could not be removed due to any reason Bundler will display a warning.
     D
     method_option "install", :type => :boolean, :banner =>
       "Runs 'bundle install' after removing the gems from the Gemfile"
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -166,12 +166,12 @@ module Bundler
       Check.new(options).run
     end
 
-    desc "remove [GEMS]", "Removes the gem from gemfile"
+    desc "remove [GEMS] [OPTIONS]", "Removes the gem from gemfile"
     long_desc <<-D
-      Removes gems from the Gemfile. If a gem is not found, Bundler prints a error message and if gem could not be removed due to any reason Bundler warns the user.
+      Removes gems from the Gemfile. If a gem is not found, Bundler prints a error message and if gem could not be removed due to any reason Bundler displays warning.
     D
     method_option "install", :type => :boolean, :banner =>
-      "Removes gems from bundle"
+      "Runs 'bundle install' after removing the gems from the gemfile"
     def remove(*gems)
       require "bundler/cli/remove"
       Remove.new(gems, options).run
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -168,10 +168,10 @@ module Bundler
 
     desc "remove [GEMS] [OPTIONS]", "Removes the gem from the Gemfile"
     long_desc <<-D
-      Removes the specified gems from the Gemfile in the current directory. If a gem is not found, Bundler prints a error message and if gem could not be removed due to any reason Bundler displays warning.
+      Removes the specified gems from the Gemfile. If the gem is not found, Bundler prints a error message and if gem could not be removed due to any reason Bundler will display a warning.
     D
     method_option "install", :type => :boolean, :banner =>
-      "Runs 'bundle install' after removing the gems from the gemfile"
+      "Runs 'bundle install' after removing the gems from the Gemfile"
     def remove(*gems)
       require "bundler/cli/remove"
       Remove.new(gems, options).run
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -166,13 +166,13 @@ module Bundler
       Check.new(options).run
     end
 
-    desc "remove [GEM_NAME]", "Removes the gem from gemfile"
+    desc "remove [GEMS]", "Removes the gem from gemfile"
     long_desc <<-D
-      Scans the gemfile for the gem and removes it. If not found, Bundler prints a error message.
+      Scans the gemfile for the gems and removes them. If not found, Bundler prints a error message.
     D
-    def remove(gem_name)
+    def remove(*gems)
       require "bundler/cli/remove"
-      Remove.new(gem_name).run
+      Remove.new(gems).run
     end
 
     desc "install [OPTIONS]", "Install the current environment to the system"
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -170,9 +170,11 @@ module Bundler
     long_desc <<-D
       Scans the gemfile for the gems and removes them. If not found, Bundler prints a error message.
     D
+    method_option "install", :type => :boolean, :banner =>
+      "Remove gem from bundle"
     def remove(*gems)
       require "bundler/cli/remove"
-      Remove.new(gems).run
+      Remove.new(gems, options).run
     end
 
     desc "install [OPTIONS]", "Install the current environment to the system"
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -166,9 +166,9 @@ module Bundler
       Check.new(options).run
     end
 
-    desc "remove [GEMS] [OPTIONS]", "Removes the gem from gemfile"
+    desc "remove [GEMS] [OPTIONS]", "Removes the gem from the Gemfile"
     long_desc <<-D
-      Removes gems from the Gemfile. If a gem is not found, Bundler prints a error message and if gem could not be removed due to any reason Bundler displays warning.
+      Removes the specified gems from the Gemfile in the current directory. If a gem is not found, Bundler prints a error message and if gem could not be removed due to any reason Bundler displays warning.
     D
     method_option "install", :type => :boolean, :banner =>
       "Runs 'bundle install' after removing the gems from the gemfile"
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -166,6 +166,15 @@ module Bundler
       Check.new(options).run
     end
 
+    desc "remove [GEM_NAME]", "Removes the gem from gemfile"
+    long_desc <<-D
+      Scans the gemfile for the gem and removes it. If not found, Bundler prints a error message.
+    D
+    def remove(gem_name)
+      require "bundler/cli/remove"
+      Remove.new(gem_name).run
+    end
+
     desc "install [OPTIONS]", "Install the current environment to the system"
     long_desc <<-D
       Install will install all of the gems in the current bundle, making them available
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -168,10 +168,10 @@ module Bundler
 
     desc "remove [GEMS]", "Removes the gem from gemfile"
     long_desc <<-D
-      Scans the gemfile for the gems and removes them. If not found, Bundler prints a error message.
+      Removes gems from the Gemfile. If a gem is not found, Bundler prints a error message and if gem could not be removed due to any reason Bundler warns the user.
     D
     method_option "install", :type => :boolean, :banner =>
-      "Remove gem from bundle"
+      "Removes gems from bundle"
     def remove(*gems)
       require "bundler/cli/remove"
       Remove.new(gems, options).run
--- a/lib/bundler/cli/remove.rb
+++ b/lib/bundler/cli/remove.rb
@@ -11,6 +11,7 @@ module Bundler
       # Raise error if no gems are specified
       raise InvalidOption, "Please specify gems to remove." if @gems.empty?
 
+      # remove requested gems
       removed_deps = Injector.remove(@gems, {})
 
       # Remove gems from .bundle if install flag is specified
--- a/lib/bundler/cli/remove.rb
+++ b/lib/bundler/cli/remove.rb
@@ -47,20 +47,33 @@ module Bundler
       re = Regexp.union(patterns)
 
       lines = ""
+      group = false
+      whole_group = ""
+      inside_group = ""
       IO.readlines(Bundler.default_gemfile).map do |line|
-        lines += line unless line.match(re)
-      end
+        group = true if line =~ /group /
 
-      # TODO: remove any empty blocks
-      lines = remove_empty_blocks(lines)
+        lines += line if !line.match(re) && !group
+
+        next unless group
+        whole_group += line unless line.match(re)
+
+        if line =~ /end/
+          group = false
+        elsif line !~ /group /
+          inside_group += line unless line.match(re)
+        end
+
+        if inside_group =~ /gem / && !group
+          lines += whole_group
+          inside_group = ""
+          whole_group = ""
+        end
+      end
 
       File.open(Bundler.default_gemfile, "w") do |file|
         file.puts lines
       end
-    end
-
-    def remove_empty_blocks(lines)
-      lines
     end
 
     def display_removed_gems
--- a/lib/bundler/cli/remove.rb
+++ b/lib/bundler/cli/remove.rb
@@ -8,13 +8,11 @@ module Bundler
     end
 
     def run
-      # Raise error if no gems are specified
       raise InvalidOption, "Please specify gems to remove." if @gems.empty?
 
       # remove requested gems
       Injector.remove(@gems, {})
 
-      # Remove gems from .bundle if install flag is specified
       Installer.install(Bundler.root, Bundler.definition) if @options["install"]
     end
   end
--- a/lib/bundler/cli/remove.rb
+++ b/lib/bundler/cli/remove.rb
@@ -16,6 +16,7 @@ module Bundler
       # Remove gems from .bundle if install flag is specified
       Installer.install(Bundler.root, Bundler.definition) if @options["install"]
 
+      # print success for removed gems
       removed_deps.each {|dep| Bundler.ui.confirm "#{SharedHelpers.pretty_dependency(dep)} was removed." }
     end
   end
--- a/lib/bundler/cli/remove.rb
+++ b/lib/bundler/cli/remove.rb
@@ -16,7 +16,7 @@ module Bundler
       # Remove gems from .bundle if install flag is specified
       Installer.install(Bundler.root, Bundler.definition) if @options["install"]
 
-      removed_deps.each {|dep| Bundler.ui.confirm "#{dep.name}(#{dep.requirement}) was removed." }
+      removed_deps.each {|dep| Bundler.ui.confirm "#{SharedHelpers.pretty_dependency(dep)} was removed." }
     end
   end
 end
--- a/lib/bundler/cli/remove.rb
+++ b/lib/bundler/cli/remove.rb
@@ -14,18 +14,12 @@ module Bundler
       # store removed gems to display on successfull removal
       @removed_deps = builder.remove_gems(@gems)
 
-      # resolve to see if the after removing dep broke anything
+      # remove gems from lockfile
       @definition = builder.to_definition(Bundler.default_lockfile, {})
-      @definition.resolve_remotely!
-
-      # since nothing broke, we can remove those gems from the gemfile
-      remove_gems_from_gemfile
-
-      # since we resolved successfully, write out the lockfile
       @definition.lock(Bundler.default_lockfile)
 
-      # invalidate the cached Bundler.definition
-      Bundler.reset_paths!
+      # remove those gems from the gemfile
+      remove_gems_from_gemfile
 
       # TODO: Discuss about using this,
       # currently gems are not removed from .bundle
@@ -64,11 +58,10 @@ module Bundler
           inside_group += line unless line.match(re)
         end
 
-        if inside_group =~ /gem / && !group
-          lines += whole_group
-          inside_group = ""
-          whole_group = ""
-        end
+        next unless inside_group =~ /gem / && !group
+        lines += whole_group
+        inside_group = ""
+        whole_group = ""
       end
 
       File.open(Bundler.default_gemfile, "w") do |file|
--- a/lib/bundler/cli/remove.rb
+++ b/lib/bundler/cli/remove.rb
@@ -9,7 +9,7 @@ module Bundler
 
     def run
       # Raise error if no gems are specified
-      raise InvalidOption, "Please specify gems to remove" if @gems.empty?
+      raise InvalidOption, "Please specify gems to remove." if @gems.empty?
 
       removed_deps = Injector.remove(@gems, {})
 
--- a/lib/bundler/cli/remove.rb
+++ b/lib/bundler/cli/remove.rb
@@ -12,13 +12,10 @@ module Bundler
       raise InvalidOption, "Please specify gems to remove." if @gems.empty?
 
       # remove requested gems
-      removed_deps = Injector.remove(@gems, {})
+      Injector.remove(@gems, {})
 
       # Remove gems from .bundle if install flag is specified
       Installer.install(Bundler.root, Bundler.definition) if @options["install"]
-
-      # print success for removed gems
-      removed_deps.each {|dep| Bundler.ui.confirm "#{SharedHelpers.pretty_dependency(dep)} was removed." }
     end
   end
 end
--- a/lib/bundler/cli/remove.rb
+++ b/lib/bundler/cli/remove.rb
@@ -2,77 +2,21 @@
 
 module Bundler
   class CLI::Remove
-    def initialize(gems)
+    def initialize(gems, options)
       @gems = gems
-      @removed_deps = []
+      @options = options
     end
 
     def run
-      builder = Dsl.new
-      builder.eval_gemfile(Bundler.default_gemfile)
+      # Raise error if no gems are specified
+      raise InvalidOption, "Please specify gems to remove" if @gems.empty?
 
-      # store removed gems to display on successfull removal
-      @removed_deps = builder.remove_gems(@gems)
+      removed_deps = Injector.remove(@gems, {})
 
-      # remove gems from lockfile
-      @definition = builder.to_definition(Bundler.default_lockfile, {})
-      @definition.lock(Bundler.default_lockfile)
+      # Remove gems from .bundle if install flag is specified
+      Installer.install(Bundler.root, Bundler.definition) if @options["install"]
 
-      # remove those gems from the gemfile
-      remove_gems_from_gemfile
-
-      # TODO: Discuss about using this,
-      # currently gems are not removed from .bundle
-      # Installer.install(Bundler.root, Bundler.definition)
-
-      display_removed_gems
-    end
-
-  private
-
-    def remove_gems_from_gemfile
-      # store patterns of all gems to be removed
-      patterns = []
-      @gems.each do |g|
-        patterns << /gem "#{g}"/
-      end
-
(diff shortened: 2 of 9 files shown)


lib/bundler/dsl.rb
```diff
@@ -91,6 +91,25 @@ def gemspec(opts = nil)
       end
     end
 
+    def remove_gems(gems)
```
segiddins: does this need to go in the DSL, or could this be implemented in the Injector class that already handles adding gems?

lib/bundler/cli/remove.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    def run
+      builder = Dsl.new
+      builder.eval_gemfile(Bundler.default_gemfile)
+
+      # store removed gems to display on successfull removal
+      @removed_deps = builder.remove_gems(@gems)
+
+      # remove gems from lockfile
+      @definition = builder.to_definition(Bundler.default_lockfile, {})
```
segiddins: looks like this doesnt need to be an instance variable

lib/bundler/cli/remove.rb
```diff
@@ -0,0 +1,78 @@
...
+      # TODO: Discuss about using this,
+      # currently gems are not removed from .bundle
+      # Installer.install(Bundler.root, Bundler.definition)
+
+      display_removed_gems
+    end
+
+  private
+
+    def remove_gems_from_gemfile
```
segiddins: does this actually assert that all requested gems _can_ be removed by the regular expressions?

spec/commands/remove_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+    it "when gem is present in gemfile" do
+      gemfile <<-G
+        source "file://#{gem_repo1}"
+
+        gem "rack"
+      G
+
+      bundle "remove rack"
+
+      expect(gemfile).to_not match(/gem "rack"/)
```
segiddins: can you please change these tests to check for equality? That'll make reviewing much easier

spec/commands/remove_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+RSpec.describe "bundle remove" do
+  context "remove a single gem from gemfile" do
+    it "when gem is present in gemfile" do
+      gemfile <<-G
+        source "file://#{gem_repo1}"
+
+        gem "rack"
+      G
+
+      bundle "remove rack"
```
segiddins: when the commands are expected to succeed, please use `bundle!`

lib/bundler/cli/remove.rb
```diff
@@ -0,0 +1,78 @@
...
+      # store patterns of all gems to be removed
+      patterns = []
+      @gems.each do |g|
+        patterns << /gem "#{g}"/
```
segiddins: what if the Gemfile uses single quotes?

lib/bundler/dsl.rb
agrim123: In add command, the injector class appends to a gemfile, whereas in this case, we need to rewrite the whole gemfile. So we will need to put both the logic in `Injector` class.  
We can then create a new method with the same name and move all remove logic to that method(including builder)?

spec/commands/remove_spec.rb
agrim123: I was thinking about writing a test something like this
```ruby
expected_gemfile = "source \"file://#{gem_repo1}\"\n"
expect(gemfile).to eq(expected_gemfile)
```

for some test, directly matching the gemfile? Is it fine to match this way?

spec/commands/remove_spec.rb
agrim123: Sure. Just discovered the `bang` method.

lib/bundler/cli/remove.rb
agrim123: Removing it.

lib/bundler/cli/remove.rb
agrim123: Not yet. Just the basic ones. It does not handle the nested groups well. Working on that for now.

lib/bundler/cli/remove.rb
agrim123: Thinking of something better to remove them as per your points.  
Maybe write a method that would remove gems from the array of dependencies and then write to gemfile.

agrim123:
@segiddins In accordance with your points:
0. Trying something for this
1. Will we have to create separate functions for all these cases? And can they be nested also? And if they become empty then remove empty blocks also?
2. Apart from the case of single quotes, if regex matches whole line is removed.
3. In current scenario, all instances of gem from gemfile would be removed. And I suppose that we cannot specify same gem more than once?
4. Have not encountered this case? Can you point me to a spec for this (for any other method) or an example?
5. Does this means the case of various gemfiles (custom and default)? In that case we would need to ask user which gemfile to operate on (`BUNDLE_GEMFILE`) and by default use default configuration of user.
6. I do not know how this is implemented or used?

Rest all the points

segiddins:
> Will we have to create separate functions for all these cases? And can they be nested also? And if they become empty then remove empty blocks also?

yes to all of them. though it shouldn't be necessary to write separate functions, we can handle all of them at once and just keep a stack of the blocks we're in

> Apart from the case of single quotes, if regex matches whole line is removed.

we'll need to make this more strict -- to start out with, it should probably only remove if the only other thing on the line is whitespace or a comment -- it would be bad for us to remove more than we're supposed to

> In current scenario, all instances of gem from gemfile would be removed

that's fine, just need to add test for that

> Have not encountered this case? Can you point me to a spec for this (for any other method) or an example?

```ruby
gem 'a'; gem "b"

[gem "a", "gem "b"]

puts(gem("a"), gem("b"))
```

are all simple examples -- not saying they are frequent, but since this is just ruby, that's a valid use of the DSL

> Does this means the case of various gemfiles (custom and default)? In that case we would need to ask user which gemfile to operate on (BUNDLE_GEMFILE) and by default use default configuration of user.


See the `Definition#eval_gemfile` method, which users can call to (conditional) include additional `gem` statements in their gemfile, we have tests for it somewhere

> I do not know how this is implemented or used? Can you brief me on this?

See `DSL#gemspec`, or the tests inside `spec/install/gemfile/gemspec_spec.rb`

spec/commands/remove_spec.rb
```diff
@@ -229,4 +229,110 @@
...
+
+  describe "arbitrary gemfile" do
+    context "when mutiple gems are present in same line" do
+      it "does not removes the gem" do
+        install_gemfile <<-G
+          source "file://#{gem_repo1}"
+          gem "rack"; gem "rails"
+        G
+
+        expected_gemfile = "source \"file://#{gem_repo1}\"\n\n" \
```
segiddins: please use heredoc strings for these, it makes reviewing much easier

lib/bundler/dsl.rb
segiddins: this still needs to be moved out of the DSL

lib/bundler/injector.rb
```diff
@@ -90,5 +115,57 @@ def append_to(gemfile_path, new_gem_lines)
         f.puts new_gem_lines
       end
     end
+
+    def remove_gems_from_gemfile(removed_deps, gemfile_path)
+      # store patterns of all gems to be removed
+      patterns = []
```
segiddins: `pattern = /gem\s+['"]#{Regexp.union(removed_deps)}\1|gem\(['"]#{Regexp.union(removed_deps)}\2\)/` ?
This will match everything in a single regular expression

lib/bundler/injector.rb
```diff
@@ -90,5 +115,57 @@ def append_to(gemfile_path, new_gem_lines)
...
+      removed_deps.each do |dep|
+        patterns << /gem "#{dep.name}"/
+        patterns << /gem '#{dep.name}'/
+      end
+
+      # create a union of patterns to match any of them
+      re = Regexp.union(patterns)
+
+      new_gemfile = []
+      IO.readlines(gemfile_path).each {|line| new_gemfile << line unless line.match(re) }
```
segiddins: `new_gemfile = IO.readlines.select ` ?

lib/bundler/injector.rb
```diff
@@ -90,5 +115,57 @@ def append_to(gemfile_path, new_gem_lines)
...
+
+      # remove lone \n and append them with other strings
+      new_gemfile.each_with_index do |_line, index|
+        if new_gemfile[index + 1] == "\n"
+          new_gemfile[index] += new_gemfile[index + 1]
+          new_gemfile.delete_at(index + 1)
+        end
+      end
+
+      blocks = ["group ", "source ", "env ", "install_if"]
```
segiddins: why do some of these have a trailing space but not `install_if` ?

lib/bundler/injector.rb
```diff
@@ -90,5 +115,57 @@ def append_to(gemfile_path, new_gem_lines)
...
+      blocks.each {|block| remove_nested_blocks(new_gemfile, block) }
+
+      File.open(gemfile_path, "w") {|file| file.puts new_gemfile.join.chomp }
```
segiddins: this should use `SharedHelpers.filesystem_access`

lib/bundler/injector.rb
```diff
@@ -90,5 +115,57 @@ def append_to(gemfile_path, new_gem_lines)
...
+    end
+
+    def remove_nested_blocks(gemfile, block_name)
+      nested_blocks = 0
+
+      # count number of nested groups
+      gemfile.each_with_index do |line, index|
+        nested_blocks += 1 if gemfile[index + 1] =~ /#{block_name}/ && line =~ /#{block_name}/
```
segiddins: `include?(block_name)` ?

lib/bundler/injector.rb
```diff
@@ -90,5 +115,57 @@ def append_to(gemfile_path, new_gem_lines)
...
+      end
+
+      while nested_blocks >= 0
+        nested_blocks -= 1
+
+        gemfile.each_with_index do |line, index|
+          next if line !~ /#{block_name}/
+          if gemfile[index + 1] =~ /end/
```
segiddins: this should probably only match `/^\s*end\s*$` ? otherwise we could accidentally do something like remove a line with a comment that contains the work `end`

lib/bundler/cli/remove.rb
```diff
@@ -0,0 +1,22 @@
...
+    def run
+      # Raise error if no gems are specified
+      raise InvalidOption, "Please specify gems to remove" if @gems.empty?
+
+      removed_deps = Injector.remove(@gems, {})
+
+      # Remove gems from .bundle if install flag is specified
+      Installer.install(Bundler.root, Bundler.definition) if @options["install"]
+
+      removed_deps.each {|dep| Bundler.ui.confirm "#{dep.name}(#{dep.requirement}) was removed." }
```
segiddins: `"#{SharedHelpers.pretty_dependency(dep)} was removed` ?

lib/bundler/dsl.rb
agrim123: Moving this method to `Injector` under same name and takes `builder` as input and is a private method?

lib/bundler/injector.rb
agrim123: Done

lib/bundler/injector.rb
agrim123: Fixed

lib/bundler/injector.rb
agrim123: Done

lib/bundler/injector.rb
agrim123: Needed an edge case for the last element to use `include`. Done

lib/bundler/injector.rb
agrim123: This throws an error 
```bash
TypeError: no implicit conversion of Bundler::Dependency into String
```
Looking into it.

agrim123:
@segiddins I revisited an old test case. Suppose user wants to remove mutiple gems from gemfile but some of them are not present, should we abort the operation altogether or remove those which are present? 
Something like this
```ruby
it "when a gem is not present in gemfile" do
  gemfile <<-G
    source "file://#{gem_repo1}"

    gem "rails"
    gem "minitest"
    gem "rspec"
  G

  bundle! "remove rails rack minitest"
  expect(out).to include("rails (>= 0) was removed.")
  expect(out).to include("minitest (>= 0) was removed.")
  expect(out).to include("`rack` is not specified in Gemfile so not removed.")
  gemfile_should_be <<-G
   source "file://#{gem_repo1}"

   gem "rspec"
  G
end
```

segiddins:
> Suppose user wants to remove mutiple gems from gemfile but some of them are not present, should we abort the operation altogether (and not remove any gems) or remove those which are present?

I think we should abort the operation, that way the command can be fixed & re-run

lib/bundler/injector.rb
```diff
@@ -90,5 +112,144 @@ def append_to(gemfile_path, new_gem_lines)
...
+    # @return [Array] removed_deps Array of removed dependencies.
+    def remove_gems_from_dependencies(builder, gems)
+      removed_deps = []
+
+      gems.each do |gem_name|
+        deleted_dep = builder.dependencies.find {|d| d.name == gem_name }
+
+        if deleted_dep.nil?
+          Bundler.ui.warn "`#{gem_name}` is not specified in Gemfile so not removed."
+          return []
```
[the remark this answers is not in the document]
agrim123: Used `next` and displayed a [warning](https://github.com/bundler/bundler/pull/6513/files#diff-822473873989fb564577766f37b75adaR163).
Wanted to clarify one thing though. when we evaluate the main gemfile we get all the dependencies that are present? So we have a valid check and we can safely abort the operation on the first iteration when the gem is not found. So, maybe aborting operation altogether is a good choice. Reverting back.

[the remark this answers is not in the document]
agrim123: However there is one case that is confusing though [here](https://github.com/bundler/bundler/pull/6513/files#diff-500766ca67a2c0e99af9a217fb36d9c8R419)

lib/bundler/cli.rb
```diff
@@ -166,6 +166,17 @@ def check
       Check.new(options).run
     end
 
+    desc "remove [GEMS]", "Removes the gem from gemfile"
```
segiddins: `from the Gemfile`

lib/bundler/cli.rb
```diff
@@ -166,6 +166,17 @@ def check
...
+    long_desc <<-D
+      Removes gems from the Gemfile. If a gem is not found, Bundler prints a error message and if gem could not be removed due to any reason Bundler warns the user.
```
segiddins:
```
Removes the specified gems from the Gemfile in the current directory.
```

lib/bundler/cli/remove.rb
```diff
@@ -0,0 +1,21 @@
...
+
+module Bundler
+  class CLI::Remove
+    def initialize(gems, options)
+      @gems = gems
+      @options = options
+    end
+
+    def run
+      # Raise error if no gems are specified
```
segiddins: comment is necessary -- it's just repeating what the line of code below is saying.

in general, comments should explain the _why_ or _how_ rather that the _what_, since that's what the code does :)

lib/bundler/injector.rb
```diff
@@ -48,7 +53,27 @@ def inject(gemfile_path, lockfile_path)
...
+
+    # @param [Pathname] gemfile_path The Gemfile from which to remove dependencies.
+    # @param [Pathname] lockfile_path The lockfile from which to remove dependencies.
+    # @return [Array]
+    def remove(gemfile_path, lockfile_path)
+      # evaluate the main Gemfile
+      builder = Dsl.new
+      builder.eval_gemfile(Bundler.default_gemfile)
+
+      definition = builder.to_definition(lockfile_path, {})
```
segiddins: all this can probably now be replaced with `Bundler.definition.gemfiles.each`

lib/bundler/injector.rb
```diff
@@ -90,5 +115,147 @@ def append_to(gemfile_path, new_gem_lines)
...
+      # inform user of the the gemfile currently
+      # being evaluated
+      Bundler.ui.info "Removing gems from #{gemfile_path}"
+
+      # evaluate the Gemfile we have
+      builder = Dsl.new
+      builder.eval_gemfile(gemfile_path)
+
+      # remove gems from dependencies
+      removed_deps = remove_gems_from_dependencies(builder, @deps)
```
segiddins: this causes the following output, which is not quite right:

```ruby
# Gemfile
eval_gemfile 'Gemfile2'
gem 'ab'
# Gemfile2
gem("ab")
```

=>

```
Removing gems from Gemfile
ab could not be removed.
No gems were removed from the gemfile.
Removing gems from Gemfile2
ab (>= 0) was removed.
```

lib/bundler/injector.rb
```diff
@@ -90,5 +115,147 @@ def append_to(gemfile_path, new_gem_lines)
...
+    # @param [Array]  gems Array of names of gems to be removed.
+    # @return [Array] removed_deps Array of removed dependencies.
+    def remove_gems_from_dependencies(builder, gems)
+      removed_deps = []
+
+      gems.each do |gem_name|
+        deleted_dep = builder.dependencies.find {|d| d.name == gem_name }
+
+        if deleted_dep.nil?
+          raise GemfileError, "`#{gem_name}` is not specified in Gemfile so could not be removed."
```
segiddins: should use the actual path in the error message, and `so it could not be removed`

lib/bundler/injector.rb
```diff
@@ -90,5 +115,147 @@ def append_to(gemfile_path, new_gem_lines)
...
+      nested_blocks = 0
+
+      # count number of nested blocks
+      gemfile.each_with_index {|line, index| nested_blocks += 1 if !gemfile[index + 1].nil? && gemfile[index + 1].include?(block_name) && line.include?(block_name) }
+
+      while nested_blocks >= 0
+        nested_blocks -= 1
+
+        gemfile.each_with_index do |line, index|
+          next if line !~ /#{block_name}/
```
segiddins: `next unless line.include?(block_name)`

lib/bundler/injector.rb
```diff
@@ -90,5 +115,147 @@ def append_to(gemfile_path, new_gem_lines)
...
+          if gemfile[index + 1] =~ /^\s*end\s*$/
+            gemfile[index] = nil
+            gemfile[index + 1] = nil
+          end
+        end
+
+        # remove nil elements
+        gemfile.reject!(&:nil?)
```
segiddins: `gemfile.compact!`

lib/bundler/injector.rb
```diff
@@ -90,5 +115,147 @@ def append_to(gemfile_path, new_gem_lines)
...
+
+      # record gems which were removed but not requested
+      extra_removed_gems = original_deps - builder.dependencies
+
+      # if some extra gems were removed then raise error
+      # and revert Gemfile to original
+      unless extra_removed_gems.empty?
+        write_to_gemfile(gemfile_path, initial_gemfile.join)
+
+        raise InvalidOption, "Gems could not be removed."
```
segiddins: this should be a bit more detailed -- `Gems could not be removed -- accidentally removed gems x, y, z` or something like that?

lib/bundler/shared_helpers.rb
```diff
@@ -194,13 +194,15 @@ def ensure_same_dependencies(spec, old_deps, new_deps)
         "\nEither installing with `--full-index` or running `bundle update #{spec.name}` should fix the problem."
     end
 
-    def pretty_dependency(dep, print_source = false)
+    def pretty_dependency(dep, print_source = false, print_requirements = false)
```
segiddins: why the third parameter? I dont think we _ever want to print `(>= 0)`

lib/bundler/injector.rb
agrim123: We also need an additional check `!line.nil?`

lib/bundler/injector.rb
agrim123: How about `Gems could not be removed. #{extra_removed_gems.join(", ")} will also be removed. Aborting."`

lib/bundler/shared_helpers.rb
agrim123: Yeah. removed it 😄

lib/bundler/injector.rb
agrim123: Looking into this.

lib/bundler/injector.rb
segiddins: `would also have been removed` 👍

lib/bundler/injector.rb
agrim123: How about if we add a method `gemfile` to `Dependency` that will hold the gemfile to which the gem belongs? In that way, we can easily filter out correct gems.

lib/bundler/cli.rb
```diff
@@ -166,6 +166,17 @@ def check
       Check.new(options).run
     end
 
+    desc "remove [GEMS] [OPTIONS]", "Removes the gem from the Gemfile"
+    long_desc <<-D
+      Removes the specified gems from the Gemfile in the current directory. If a gem is not found, Bundler prints a error message and if gem could not be removed due to any reason Bundler displays warning.
```
colby-swandale: > in the current directory

This is not necessarily true, it should remove the gems from whichever Gemfile is configured as well.

colby-swandale: > and if gem

`the gem`

colby-swandale: > any reason Bundler displays warning

`any reason, Bundler will display a warning`

lib/bundler/cli.rb
```diff
@@ -166,6 +166,17 @@ def check
...
+    D
+    method_option "install", :type => :boolean, :banner =>
+      "Runs 'bundle install' after removing the gems from the gemfile"
```
colby-swandale: > gemfile

`Gemfile`

lib/bundler/injector.rb
```diff
@@ -101,5 +119,147 @@ def append_to(gemfile_path, new_gem_lines)
         f.puts new_gem_lines
       end
     end
+
+    # evalutes a gemfile to remove the specified gem
+    # from it.
+    def remove_deps(gemfile_path)
+      # get initial snap shot of the gemfile
```
colby-swandale: I find that comments like these don't really offer any useful information as they're only describing what the code is doing, which we can see already from the code. There is also the risk that they could fall out of date and end up being wrong the in the future.

I would prefer we remove these

lib/bundler/injector.rb
```diff
@@ -101,5 +119,147 @@ def append_to(gemfile_path, new_gem_lines)
...
+      errored_deps = builder.dependencies.select {|d| d.gemfile == gemfile_path } & removed_deps.select {|d| d.gemfile == gemfile_path }
+
+      # warn user regarding those gems
+      show_warning "#{errored_deps.map(&:name).join(", ")} could not be removed." unless errored_deps.empty?
+
+      # return actual removed dependencies
+      removed_deps - errored_deps
+    end
+
+    def show_warning(message)
```
colby-swandale: [Bundler::UI::Shell](https://github.com/bundler/bundler/blob/master/lib/bundler/ui/shell.rb#L33) has a `#warn` method the will print text in yellow already.

man/bundle-remove.ronn
```diff
@@ -0,0 +1,25 @@
+bundle-remove(1) -- Removes gem specified in Gemfile
+===========================================================================
+
+## SYNOPSIS
+
+`bundle remove [GEMS] [--install]`
+
+## DESCRIPTION
+
+Removes a set of specified gems from the Gemfile or if needed from the bundle itself.
```
colby-swandale: > if needed from the bundle itself

This won't make a huge amount of sense to a lot of people. I would instead add a paragraph that explains the `--install` option.

man/bundle-remove.ronn
```diff
@@ -0,0 +1,25 @@
...
+If any gem specified to be removed does not exist the in Gemfile, Bundler prints error message.
+If gem could not be removed due to any reason Bundler warns the user.
```
colby-swandale: > If gem

`the Gem`

colby-swandale: > reason Bundler warns the user

`reason, Bundler warns the user`

colby-swandale:
Does this make feature make the distinction between a gem not existing in the Gemfile and not being able to find the gem in the Gemfile? Seeing the Gemfile can be written in pretty much any way using Ruby. There will be a cases where i have a gem in the Gemfile but is not specified using `gem 'mygem'`

lib/bundler/injector.rb
agrim123: It deduplicates the message. Sometimes we need to show the same message for mutiple gemfiles.

agrim123:
We are only considering the case of a standard gemfile. We can't know apart from standard how things are in Gemfile so we remove based on only standard procedure.

colby-swandale:
Apologies, i wasn't clear on this. It's probably good in terms of error handling to print an error to the user when they want to remove a gem in their Gemfile, but we couldn't find & remove the declaration in the Gemfile.

lib/bundler/injector.rb
agrim123: Added the `gemfile` [here](https://github.com/bundler/bundler/pull/6513/commits/d16ad19f60ba42adf880da7fabafe8308a40b1a9)

segiddins:
@colby-swandale @indirect r?

man/bundle-remove.ronn
```diff
@@ -0,0 +1,25 @@
+bundle-remove(1) -- Removes gem specified in Gemfile
```
indirect: Can we update this to: `bundle-remove(1) -- Removes gems from the Gemfile`?

man/bundle-remove.ronn
```diff
@@ -0,0 +1,25 @@
...
+===========================================================================
+
+## SYNOPSIS
+
+`bundle remove [GEMS] [--install]`
```
indirect: Can we update this to:
\`bundle remove [GEM [GEM ...]] [--install]\`

man/bundle-remove.ronn
```diff
@@ -0,0 +1,25 @@
...
+
+## DESCRIPTION
+
+Removes a set of specified gems from the Gemfile or if needed runs `bundle install` and removes them from bundle as well.
+If any gem specified to be removed does not exist the in Gemfile, Bundler prints error message.
+If the gem could not be removed due to any reason, Bundler warns the user.
```
indirect: Can we update this to:
> Removes the given gems from the Gemfile while ensuring that the resulting Gemfile is still valid. If a gem cannot be removed, a warning is printed. If a gem is already absent from the Gemfile, and error is raised.

man/bundle-remove.ronn
```diff
@@ -0,0 +1,25 @@
...
+
+## OPTIONS
+
+* `--install`:
+  Runs 'bundle install' after removing the gems from the gemfile. This option sync the lockfile.
```
indirect: Can we change this to:
> Runs `bundle install` after the given gems have been removed from the Gemfile, which ensures that both the lockfile and the installed gems on disk are also updated to remove the given gem(s).

lib/bundler/cli.rb
```diff
@@ -166,6 +166,17 @@ def check
       Check.new(options).run
     end
 
+    desc "remove [GEMS] [OPTIONS]", "Removes the gem from the Gemfile"
```
indirect: Can we update this to:
> desc "remove [GEM [GEM ...]]", "Removes gems from the Gemfile"

lib/bundler/cli.rb
```diff
@@ -166,6 +166,17 @@ def check
...
+    long_desc <<-D
+      Removes the specified gems from the Gemfile. If the gem is not found, Bundler prints a error message and if gem could not be removed due to any reason Bundler will display a warning.
```
indirect: Can we update this to:
> Removes the given gems from the Gemfile while ensuring that the resulting Gemfile is still valid.

PR #6524, opened 2018-05-08
Add flag to generate generic binstubs

eanlain:

Issues #6154 and #6162, in which Bundler v1.16.1 appeared to be looking for the wrong Gemfile when using Docker. Further investigation showed that this issue was not just limited to the use of Docker (see non-Docker repro of 6154).

The diagnosis was that as of 1.16 Bundler started generating a `bin/bundle` and calling it from other binstubs. This change led to two problems (taken from conversation in #6469):

> 1. Docker (and other users) used to be able to treat bundler binstubs as if they were rubygems binstubs, not tied to any application, and usable on any ruby where that gem was installed. This change made that no longer true, and tied each binstub a to a specific application and Gemfile.
> 2. Previously, `bundle exec CMD` would find the Gemfile based on the CWD, and even if Bundler exec’ed a binstub from the PATH that was tied to a different application, that was fine, because binstubs were fully generic. It sounds like the `bin/bundle` binstub no longer does that, and instead when you run `bin/bundle exec rspec`, you get `bin/rspec`, which is tied to a specific application and Gemfile.

My fix was to add a flag to the `binstubs` command to generate application-locked binstubs and to make that flag the default for Bundler >= 2.

I chose this fix because of the conversation from #6469.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -317,6 +317,8 @@ module Bundler
       "Binstub destination directory (default bin)"
     method_option "shebang", :type => :string, :banner =>
       "Specify a different shebang executable name than the default (usually 'ruby')"
+    method_option "locked", :type => :boolean, :default => Bundler.bundler_major_version >= 2, :banner =>
+      "Make binstubs that are locked to a specific application"
     method_option "standalone", :type => :boolean, :banner =>
       "Make binstubs that can work without the Bundler runtime"
     method_option "all", :type => :boolean, :banner =>
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -317,7 +317,7 @@ module Bundler
       "Binstub destination directory (default bin)"
     method_option "shebang", :type => :string, :banner =>
       "Specify a different shebang executable name than the default (usually 'ruby')"
-    method_option "locked", :type => :boolean, :default => Bundler.bundler_major_version >= 2, :banner =>
+    method_option "locked", :type => :boolean, :default => Bundler.feature_flag.application_locked?, :banner =>
       "Make binstubs that are locked to a specific application"
     method_option "standalone", :type => :boolean, :banner =>
       "Make binstubs that can work without the Bundler runtime"
--- a/lib/bundler/feature_flag.rb
+++ b/lib/bundler/feature_flag.rb
@@ -29,6 +29,7 @@ module Bundler
 
     settings_flag(:allow_bundler_dependency_conflicts) { bundler_2_mode? }
     settings_flag(:allow_offline_install) { bundler_2_mode? }
+    settings_flag(:application_locked) { bundler_2_mode? }
     settings_flag(:auto_clean_without_path) { bundler_2_mode? }
     settings_flag(:auto_config_jobs) { bundler_2_mode? }
     settings_flag(:cache_all) { bundler_2_mode? }
--- a/lib/bundler/installer.rb
+++ b/lib/bundler/installer.rb
@@ -119,6 +119,8 @@ module Bundler
       relative_gemfile_path = relative_gemfile_path
       ruby_command = Thor::Util.ruby_command
       ruby_command = ruby_command
+      application_locked = options[:locked] || Bundler.settings[:application_locked]
+      application_locked = application_locked
       template_path = File.expand_path("../templates/Executable", __FILE__)
       if spec.name == "bundler"
         template_path += ".bundler"
@@ -128,6 +130,8 @@ module Bundler
 
       exists = []
       spec.executables.each do |executable|
+        next if executable == "bundle" && application_locked == false
+
         binstub_path = "#{bin_path}/#{executable}"
         if File.exist?(binstub_path) && !options[:force]
           exists << executable
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,6 +5,7 @@ AllCops:
   TargetRubyVersion: 1.9
   Exclude:
     - tmp/**/*
+    - lib/bundler/templates/Executable
     - lib/bundler/vendor/**/*
   DisplayCopNames: true
 
@@ -20,10 +21,6 @@ Lint/EndAlignment:
 
 Lint/UnusedMethodArgument:
   Enabled: false
-
-Lint/ScriptPermission:
-  Exclude:
-    - 'lib/bundler/templates/Executable'
 
 # Layout
 
--- a/lib/bundler/cli/binstubs.rb
+++ b/lib/bundler/cli/binstubs.rb
@@ -16,7 +16,7 @@ module Bundler
       Bundler.settings.set_command_option_if_given :shebang, options["shebang"]
       installer = Installer.new(Bundler.root, Bundler.definition)
 
-      installer_opts = { :force => options[:force], :binstubs_cmd => true }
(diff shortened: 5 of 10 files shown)


eanlain:
Added some additional tests.

eanlain:
@segiddins any other changes needed?

lib/bundler/cli.rb
```diff
@@ -317,6 +317,8 @@ def info(gem_name)
       "Binstub destination directory (default bin)"
     method_option "shebang", :type => :string, :banner =>
       "Specify a different shebang executable name than the default (usually 'ruby')"
+    method_option "locked", :type => :boolean, :default => Bundler.bundler_major_version >= 2, :banner =>
```
segiddins: can this use a feature flag instead of directly checking the version?

eanlain: 👍 I added a commit to use a feature flag instead of directly checking the version

lib/bundler/installer.rb
```diff
@@ -119,6 +119,8 @@ def generate_bundler_executable_stubs(spec, options = {})
       relative_gemfile_path = relative_gemfile_path
       ruby_command = Thor::Util.ruby_command
       ruby_command = ruby_command
+      application_locked = options[:locked] || Bundler.settings[:application_locked]
```
segiddins: `Bundler.feature_flag.application_locked?`

eanlain: Amended the commit with this change.

eanlain: Actually, the use of `Bundler.settings[:application_locked]` is just returning whether the `:application_locked` setting for this run of the installer as opposed to returning the state of the feature. For example, if using Bundler >= 2 and using `--locked=false` with the `binstub` command.

lib/bundler/feature_flag.rb
```diff
@@ -29,6 +29,7 @@ def self.settings_method(name, key, &default)
 
     settings_flag(:allow_bundler_dependency_conflicts) { bundler_2_mode? }
     settings_flag(:allow_offline_install) { bundler_2_mode? }
+    settings_flag(:application_locked) { bundler_2_mode? }
```
segiddins: nit -- `application_locked_binstubs` ?

indirect:
From a conversation with @eanlain:

In Ye Olden Times™, anytime you installed gems (with rubygems) you would get rubygems binstubs
when bundler layered on top of that, you still got rubygems binstubs, but then you had to start using `bundle exec` if you wanted to run the command using the bundled gems. so `rake` was the rubygems binstub, and `bundle exec rake` was the rubygems binstub but with env vars that forced the bundled version of rake to load.

then we had the idea to add bundle-specific binstubs, so you could run `bin/rake`, and it would NOT be the rubygems binstub, and would instead be a binstub that loaded the bundled version of rake and then ran rake.

it seems like at some point, the docker folks realized that `bin/rake` could be used to load any rake, and they started treating bundler binstubs as if they were rubygems binstubs… but no one on the bundler team even realized that was possible, because we had all come from this situation where you used the rubygems binstubs if you needed any version of rake, and you used the bundler binstubs if you needed just the exact version of rake that was specified in the bundle.

then we started getting bug reports about binstubs, where sometimes they would NOT load the version of rake specified in the application bundle, because they were being called from a CWD that contained a different gemfile. for example, `cd app2; ../app1/bin/rake` would give you the rake from app2, which was bad… because you were calling the app1 binstub specifically to get the app1 version of rake.

that’s when we “fixed the bug” by adding a hardcoded gemfile path to `bin/bundle` and requiring `bin/bundle` from all the other binstubs—and that’s when docker broke… because it was depending on behavior that the bundler team considered a bug (and also didn’t even really know was possible, tbh).

@eanlain:
> just looking at my company’s usecase with docker, we have dockerfiles where we set the bundle `path` and `bin` (via `bundle config --global path "$GEM_HOME"` and `bundle config --global bin "$GEM_HOME/bin"`), and when we `bundle install` the bundle binstubs are generated in the bundle `bin` path

wow, okay... I now see what you are saying. I am kind of astonished and impressed and horrified. So that configuration will actually cause bundler to create the rubygems binstubs and then _overwrite them_ with bundler-application-specific binstubs.

if you want your binstubs to only work with one application, it’s kind of genius. Someone clearly came up with a clever plan for docker images that only have one application in them, to stop needing bundle exec. Of course, if you do that, you have made it completely impossible for yourself to run any command _without_ bundle exec, which is why bundler doesn’t do that :slightly_smiling_face:.

So if you want binstubs that you can use _outside_ your bundled application, you should unset both of those configs and run `bundle install --system` instead. Bundler will install gems to GEM_HOME, and install rubygems binstubs to GEM_HOME/bin. Then you can add `$GEM_HOME/bin` to your PATH, and you will be set. :+1::skin-tone-2: (And, as usual, you will have to use application binstubs or run `bundle exec NAME` to run commands inside your application bundle.)

indirect:
After the discussion, we have tentatively decided to close this PR and resolve the issues discussed via Bundler and Docker configuration instead. Hopefully @eanlain can follow up with what he eventually settles on in his Dockerfile, and we can use that as the basis for a future website page about using Bundler with Docker.

eanlain:
The Dockerfiles that I ran into issues with were explicitly setting `GEM_HOME`, `BUNDLE_BIN`, `BUNDLE_PATH`, and then adding the `BUNDLE_BIN` to the `PATH`, similar to the the following snippet:

```Dockerfile
ENV GEM_HOME="/usr/local/bundle"
ENV BUNDLE_PATH="$GEM_HOME"
ENV BUNDLE_BIN="$GEM_HOME/bin"
ENV PATH="$BUNDLE_BIN:$PATH"
```

The Dockerfiles would then run `bundle install` in one or more applications. This behavior resulted in Bundler creating rubygem binstubs at `$BUNDLE_PATH/bin` then overwriting those generic binstubs with the new, as of 1.16.0, application-locked bundler binstubs (because `BUNDLE_BIN` was set).

The application-locked bundler binstubs were "locked" to the gemfile that was present when running `bundle install`. In the instances where multiple applications (multiple gemfiles) were used within the same Docker container this causes a number of issues including:

* Parsing Gemfile issues:

  ```
  [!] There was an error parsing `Gemfile`: No such file or directory @ rb_sysopen - /Gemfile. Bundler cannot continue.
  ```

* LoadErrors:

  ```
  LoadError: cannot load such file -- active_record/railtie
  ```

* Issues loading commands:

  ```
  bundler: failed to load command: rspec (/usr/local/bundle/bin/rspec)
  ```

The fix was to unset `BUNDLE_PATH`, `BUNDLE_BIN`, and change the `PATH` to use `$GEM_HOME/bin` which resulted in the original generic rubygems binstubs being created in `$GEM_HOME/bin`. This still allows for the direct usage of gem commands like `rake` _IF_ there's only one version of the gem installed. If more than one version of a gem is installed or if you want to use the application/gemfile specified version of a gem, `bundle exec <name>` accomplishes the task.

```Dockerfile
ENV GEM_HOME="/usr/local/bundle"
ENV PATH $GEM_HOME/bin:$GEM_HOME/gems/bin:$PATH
```

It looks like similar changes were made in the [docker-library/ruby](https://github.com/docker-library/ruby) dockerfiles, see:

* https://github.com/docker-library/ruby/pull/208
* https://github.com/docker-library/ruby/pull/209

indirect:
Awesome, thanks for the writeup!

PR #6531, opened 2018-05-12
Add option to filter gem-dependencies from output of 'bundle outdated'

### LINKED ISSUE #5366: outdated: option to check only direct/explicit dependencies?
We're running `bundle outdated` automatically against our code base to find and report any outdated dependency, let's say, as a part of the CI build.

To notice that there is a new version that's falling out of our version constraints specified, we also use `--no-strict` (it's unnecessary right now, ofc, but may be necessary later).

Some pretty useful gems, like `pry` or even `actioncable`, are depending on not-the-latest-version of some gems. For instance, at the time of the posting, running `bundle outdated` with `pry 0.10.5` and `actioncable 5.0.1` in project produces:
```
  * nio4r (newest 2.0.0, installed 1.2.1)
  * slop (newest 4.4.1, installed 3.6.0)
```

It would be nice to have a flag (like with `--strict/--no-strict`) to check only top-level/direct/explicit dependencies. This way the bundle check we use can be set up in two separate checks: 0) for direct dependencies, not strictly and 1) for all dependencies, strictly. 

Is there anyone else also interested in this feature?

peret:
Resolves #5366 by adding a new option '--filter-dependencies' to `bundle outdated`. When present, `outdated` will only check the `gemfile_specs` and skip the `dependency_specs`.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -368,6 +368,8 @@ module Bundler
     method_option "filter-patch", :type => :boolean, :banner => "Only list patch newer versions"
     method_option "parseable", :aliases => "--porcelain", :type => :boolean, :banner =>
       "Use minimal formatting for more parseable output"
+    method_option "filter-dependencies", :type => :boolean, :banner =>
+      "Only list gems specified in your Gemfile, not their dependencies"
     def outdated(*gems)
       require "bundler/cli/outdated"
       Outdated.new(options, gems).run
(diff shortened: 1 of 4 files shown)


lib/bundler/cli.rb
```diff
@@ -368,6 +368,8 @@ def add(gem_name)
     method_option "filter-patch", :type => :boolean, :banner => "Only list patch newer versions"
     method_option "parseable", :aliases => "--porcelain", :type => :boolean, :banner =>
       "Use minimal formatting for more parseable output"
+    method_option "filter-dependencies", :type => :boolean, :banner =>
```
segiddins: maybe this should be something like `--only-explicit` ? When I see `filter`, I think I'm going to be able to pass in a filtering function myself

peret: Yeah, I wasn't sure about the best name... I tried following the "convention" of the --filter-[major|minor|patch] flags here. But if you want, I can change it to something else.
Whatever gets it merged ;)

peret: I just updated the PR with the new name for this option.

segiddins:
Thanks for this!

@bundlerbot r+

PR #6536, opened 2018-05-17
Removed self-hosted YAML Serializer and PsychedYAML

hsbt:

 It is not compatible with YAML spec of the standard library of Ruby.

I did revert a feature in last year: https://github.com/ruby/psych/pull/336

`YAMLSerializer` was introduced from https://github.com/bundler/bundler/commit/bfd4be3756fa6bb43c263e303df2995ab0e7ace9

I'm not sure why that was required.

Removed `YAMLSerializer`. and use `YAML` provided by `pyched_yaml`.


segiddins:
It is required since YAML is in a gem, and we load the list of plugins & settings before Bundler.setup, so using the Psych gem will mean no Bundler users can choose what version of pysch they use

hsbt:
I know the problem that psych activation problem. But Bundler already uses psych directly. Users couldn't specify its version.

deivid-rodriguez:
I thinking, what @segiddins said makes sense. But what @hsbt says makes sense too. Even though we have our own YAML implementation in order to not interfere with `yaml` / `psych` loading, we were actually activating `psych`. Shouldn't we use our custom implementation everywhere instead of using the default gem?

hsbt:
FYI: https://bugs.ruby-lang.org/issues/13847

PR #6539, opened 2018-05-18
Fix man page links

BanzaiMan:
With a markup

    [`bundle platform(1)`][bundle-platform(1)]

ronn creates this HTML fragment

    <a href="bundle-platform.html"><code>bundle platform(1)</code></a>

At the same time, it generates HTML file `bundle-platform.1.html` based on the
man page section, and this results in certain inter-man-page links 404.

We resolve this inconsistency by spelling out the href attributes.

-----

404's on https://bundler.io.

For example, https://bundler.io/v1.16/man/bundle.1.html has the link to `bundle platform(1)`:

bundler__bundle

This points to https://bundler.io/v1.16/man/bundle-platform.html, but it is 404. The correct link is https://bundler.io/v1.16/man/bundle-platform.1.html.

My diagnosis was... 

that the document source was wrong.

My fix...

is this PR.

I chose this fix because...

`ronn -5 man/*.ronn` appears to generate `<a>` tags with the correct `href` attributes.

I tried serving the generated files with `bundler-site`, but I could not convince it to do so; `bundler-site`'s Rake tasks assumes the document source *always* comes from the remote `bundler/bundler`, and coming up with a way to test it seemed rather involved.

### DIFF
--- a/man/bundle-binstubs.ronn
+++ b/man/bundle-binstubs.ronn
@@ -40,4 +40,4 @@ Calling binstubs with [GEM [GEM]] will create binstubs for all given gems.
 ## BUNDLE INSTALL --BINSTUBS
 
 To create binstubs for all the gems in the bundle you can use the `--binstubs`
-flag in [bundle install(1)][bundle-install(1)].
+flag in [bundle install(1)](bundle-install.1.html).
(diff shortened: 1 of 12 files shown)


BanzaiMan:
One can argue that this should be something that ronn should fix; however, that gem hasn't had a release in 8 years, so I reckon we should not count on it.

segiddins:
Thanks!

@bundlerbot r+

PR #6542, opened 2018-05-20
add option to Bundler#mkdir_p to force Bundler to not use sudo

### LINKED ISSUE #6535: Errno::EPERM: Operation not permitted @ chown_internal in v1.16.2
Errno::EPERM: Operation not permitted @ chown_internal in v1.16.2

## Expected Behavior

`bundle install` should run normally without error

## Current Behavior

Here are the error message:
```
Fetching rake 12.3.0
Installing rake 12.3.0
Errno::EPERM: Operation not permitted @ chown_internal -
/tmp/bundler20180517-11343-fttco1rake-12.3.0/bin
An error occurred while installing rake (12.3.0), and Bundler cannot continue.
Make sure that `gem install rake -v '12.3.0' --source 'https://rubygems.org/'`
succeeds before bundling.
```

## Possible Solution

I don't know, right now we only revert back to 1.16.1

## Steps to Reproduce

1. Install chruby 0.3.9
2. Install ruby 2.4.1
3. Install bundler 1.16.2
4. `bundle install` for Rails project

colby-swandale:

There is a bug for functionality that was added in #6258.  In certain scenarios, Bundler will create a folder for a temporary gem install with `root:root` permissions. This is happening because [Bundler#mkdir_p](https://github.com/bundler/bundler/blob/master/lib/bundler.rb#L377) checks for `requires_sudo?` which is creating a folder owned by `root:root` when it should be creating the folder with the current user.

See #6535

I can see that Bundler is creating the `bin` folder with `root:root` permissions

```
[vagrant@localhost ~]$ ls -la /tmp/bundler20180519-24861-1y67io7rake-12.3.1/
total 4
drwx------. 3 vagrant vagrant   17 May 19 07:36 .
drwxrwxrwt. 9 root    root    4096 May 19 07:36 ..
drwxr-xr-x. 2 root    root       6 May 19 07:36 bin
```

Add an option for `Bundler#mkdir_p` to prevent it from using `sudo` even though `requires_sudo?` is true.

Fixes #6535

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -373,8 +373,8 @@ EOF
       @requires_sudo = settings.allow_sudo? && sudo_present && sudo_needed
     end
 
-    def mkdir_p(path)
-      if requires_sudo?
+    def mkdir_p(path, options = {})
+      if requires_sudo? && !options[:no_sudo]
         sudo "mkdir -p '#{path}'" unless File.exist?(path)
       else
         SharedHelpers.filesystem_access(path, :write) do |p|
--- a/lib/bundler/source/rubygems.rb
+++ b/lib/bundler/source/rubygems.rb
@@ -144,7 +144,7 @@ module Bundler
             bin_path     = Bundler.system_bindir
           end
 
-          Bundler.mkdir_p bin_path unless spec.executables.empty? || Bundler.rubygems.provides?(">= 2.7.5")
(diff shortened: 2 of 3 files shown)


colby-swandale:
@bundlerbot r+

colby-swandale:
@bundlerbot r+

colby-swandale:
@bundlerbot retry

PR #6547, opened 2018-05-22
Add mutiple gems by names

### LINKED ISSUE #6543: Add support for mutiple gems to `bundle add`
Sample Usage
```bash
bundle add rack rails
```
For now, only support names and add support for the version in the future version.

agrim123:
Add support for mutiple gems by names only

```bash
bundle add rack rails
```
Concerns:
- All the options/switches used would be applied to the all gems specified and a warning is displayed for the time being.

Closes #6543

### DIFF
--- a/lib/bundler/cli/add.rb
+++ b/lib/bundler/cli/add.rb
@@ -11,6 +11,9 @@ module Bundler
     def run
       raise InvalidOption, "You can not specify `--strict` and `--optimistic` at the same time." if @options[:strict] && @options[:optimistic]
 
+      # raise error when no gems are specified
+      raise InvalidOption, "Please specify gems to add." if @gems.empty?
+
       version = @options[:version].nil? ? nil : @options[:version].split(",").map(&:strip)
 
       unless version.nil?
--- a/lib/bundler/cli/add.rb
+++ b/lib/bundler/cli/add.rb
@@ -2,8 +2,8 @@
 
 module Bundler
   class CLI::Add
-    def initialize(options, gem_name)
-      @gem_name = gem_name
+    def initialize(options, gems)
+      @gems = gems
       @options = options
       @options[:group] = @options[:group].split(",").map(&:strip) if !@options[:group].nil? && !@options[:group].empty?
     end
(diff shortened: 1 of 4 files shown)


lib/bundler/cli/add.rb
```diff
@@ -2,23 +2,38 @@
...
       end
-      dependency = Bundler::Dependency.new(@gem_name, version, @options)
 
-      Injector.inject([dependency], :conservative_versioning => @options[:version].nil?) # Perform conservative versioning only when version is not specified
+      # to store dependencies formed from gem names
+      dependencies = []
+
+      # create a dependency from gem name and
+      # push to dependencies
+      @gems.each {|g| dependencies << Bundler::Dependency.new(g, version, @options) }
```
jules2689: Might be nicer as
```ruby
dependencies = @gems.each_with_object([]) {|g,acc| acc << Bundler::Dependency.new(g, version, @options) }
```

The comment seems a but redundant as the line isn't super complex as well.
Same with the `inject the dependencies` comment since the line is actually `Inject.inject(dependencies...)`

agrim123: Sure. Will make the changes.

lib/bundler/cli/add.rb
```diff
@@ -2,23 +2,32 @@
...
 
       unless version.nil?
         version.each do |v|
           raise InvalidOption, "Invalid gem requirement pattern '#{v}'" unless Gem::Requirement::PATTERN =~ v.to_s
         end
       end
-      dependency = Bundler::Dependency.new(@gem_name, version, @options)
 
-      Injector.inject([dependency], :conservative_versioning => @options[:version].nil?) # Perform conservative versioning only when version is not specified
+      dependencies = @gems.each_with_object([]) {|g, d| d << Bundler::Dependency.new(g, version, @options) }
```
jules2689: ugh, this was added in Ruby 1.8.7
```ruby
dependencies = @gems.inject {|g, d| d << Bundler::Dependency.new(g, version, @options); d }
```
works according [to this](https://stackoverflow.com/questions/5481009/why-is-enumerableeach-with-object-deprecated) for 1.8.7

Though, I thought we had dropped support for old Rubies. 🤔

agrim123: I should have cross checked it 😓. My bad. 
Will fix it though.

agrim123: Won't be using a simple map be fine?
```ruby
dependencies = @gems.map {|g| Bundler::Dependency.new(g, version, @options) }
```

jules2689: that would work actually, yea!

segiddins:
> All the options/switches used would be applied to the all gems specified and a warning is displayed for the time being.

That's ok, there's no need for a warning

lib/bundler/cli/add.rb
```diff
@@ -2,23 +2,32 @@
...
       @options[:group] = @options[:group].split(",").map(&:strip) if !@options[:group].nil? && !@options[:group].empty?
     end
 
     def run
+      # raise error when no gems are specified
+      raise InvalidOption, "Please specify gems to add." if @gems.empty?
+
+      # show warning when options are used with mutiple gems
+      # for time being they will be applied to the all gems
+      Bundler.ui.warn "All the options will be applied to all gems`."
```
segiddins: there's no need for this, discussed offline, it's how `gem install` works

agrim123: Yup removing it.

spec/commands/add_spec.rb
```diff
@@ -17,6 +17,14 @@
     G
   end
 
+  context "when no gems are specified" do
+    it "shows error" do
+      bundle "add"
+
+      expect(out).to include("Please specify gems to add")
```
segiddins: `expect(last_command.bundler_err).to`

agrim123: Done

segiddins:
@bundlerbot r+

PR #6549, opened 2018-05-23
Add registered plugin events for before-all, before, and after gem install

jules2689:
Depends on #6548

We only had one plugin hook, which limited the plugin's capabilities a lot.
This adds more

We need more plugin hooks for gem stuff

Add more hooks in after-all (including dependencies) and before/after (including install spec)

These seemed like the obvious spots to put the hooks, but I could be wrong. The passed objects also seem to include all the info we need to action on the installations (errors, etc)

### DIFF
--- a/lib/bundler/plugin/events.rb
+++ b/lib/bundler/plugin/events.rb
@@ -31,9 +31,31 @@ module Bundler
       end
 
       # @!parse
-      #   # A hook called before any gems install
+      #   A hook called before each individual gem is installed
+      #   Includes a Bundler::ParallelInstaller::SpecInstallation.
+      #   No state, error, post_install_message will be present as nothing has installed yet
+      #   GEM_BEFORE_INSTALL = "before-install"
+      define :GEM_BEFORE_INSTALL, "before-install"
+
+      # @!parse
+      #   A hook called after each individual gem is installed
+      #   Includes a Bundler::ParallelInstaller::SpecInstallation.
+      #     - If state is failed, an error will be present.
+      #     - If state is success, a post_install_message may be present.
+      #   GEM_AFTER_INSTALL = "after-install"
+      define :GEM_AFTER_INSTALL,  "after-install"
+
+      # @!parse
+      #   A hook called before any gems install
+      #   Includes an Array of Bundler::Dependency objects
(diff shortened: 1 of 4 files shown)


lib/bundler/plugin/events.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module Bundler
+  module Plugin
+    module Events
+      def self.define(const, event)
```
segiddins: consider adding a yard macro to this, so we get docs for all of the constants?

segiddins:
Should we wait to merge this until those hooks actually get called?

lib/bundler/plugin/events.rb
```diff
@@ -0,0 +1,37 @@
...
+        const_set(const.to_sym, event) unless const_defined?(const.to_sym)
```
segiddins: should set the const to a frozen string.

jules2689:
> Should we wait to merge this until those hooks actually get called?

What do you mean?

segiddins:
Sorry, ignore me 🤐
Got confused between the two PRs. 👍🏻

lib/bundler/plugin/events.rb
jules2689: Opted to use `@parse` syntax below. I couldnt figure out macros >.<

jules2689:
This one is ready now @colby-swandale

segiddins:
Looks good! A quick spec for each of the events in `spec/plugins/hook_spec.rb` would be 💯

jules2689:
Done @segiddins 💪

lib/bundler/plugin/events.rb
```diff
@@ -31,9 +31,31 @@ def self.defined_event?(event)
...
+      #   GEM_AFTER_INSTALL = "after-install"
+      define :GEM_AFTER_INSTALL,  "after-install"
+
+      # @!parse
+      #   A hook called before any gems install
+      #   Includes an Array of Bundler::Dependency objects
       #   GEM_BEFORE_INSTALL_ALL = "before-install-all"
       define :GEM_BEFORE_INSTALL_ALL, "before-install-all"
+
+      # @!parse
```
colby-swandale: What's the `@!parse` for?

Issue #6546, opened 2018-05-22
Alternative to Fallback to a temp dir when the home directory is not usable


brodock:
The changes in https://github.com/bundler/bundler/pull/4951 introduced many issues for people bundling ruby into omnibus packages, 

There are valid use-cases to not have the home directory writable. Alternatively to falling back to a tempdir, I suggest we try to find the `.bundle` and `. gem` folders first and assert they are writable. If that fails, than we fallback to tempdir.

So the heuristic should be like:

```ruby
home = Bundler.rubygems.user_home 
dot_bundle = File.join(home, '.bundle')
dot_gem = File.join(home, '.gem')

if home.nil?
  warning += "\n * It is not set at all"
elsif !File.directory?(home)
  warning += "\n * `#{home}` is not a directory"
elsif !File.writable?(home) && [dot_bundle, dot_gem].any? { |dir| !File.directory?(dir) || !File.writable?(dir) }
  warning += "\n * `#{home}` is not writable"
else
  return @user_home = Pathname.new(home)
end
```

That allows package software to provide required `.gem` and `. bundle` folders with right permission without messing with the rest of the environment.

svoop:
This approach should fix deployment issues to application servers with tight security as well.

PR #6550, opened 2018-05-23
Don't fallback to tempdir when required directories exist.

brodock:

When running Omnibus packaged software with updated bundler, a warning is displayed because the home folder is not owned by the user:

```
`/var/opt/gitlab` is not writable.
Bundler will use `/tmp/bundler/home/root' as your home directory temporarily.
```

There are valid reasons why this is desired, and I don't have control over it. What I can do is create the required folders used by bundler and provide them with the right permissions.

See https://github.com/bundler/bundler/issues/6546

In practice instead of asking for permission on a higher level, if required folders are present and they have the right permissions, we shouldn't fallback to warning + temp directory, we should just use what is provided.

When home directory is not writable, but the required .gem and .bundle
are, we should use them instead of falling back to use tempdirs.

This creates a workaround for more restrictive setups using Omnibus
Docker or any hardened setup, to overcome the annoyances introduced by #4951.

This allows for distributions, package maintainers, etc to provide an alternative while keeping their hardenings requirements.

When provided the required folders with the required ownership/permission, we should not bother by not having any write permissions on the `$HOME` directory.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -159,12 +159,13 @@ module Bundler
     def user_home
       @user_home ||= begin
         home = Bundler.rubygems.user_home
+        bundle_home = home ? File.join(home, ".bundle") : nil
 
         warning = if home.nil?
           "Your home directory is not set."
         elsif !File.directory?(home)
           "`#{home}` is not a directory."
-        elsif !File.writable?(home)
+        elsif !File.writable?(home) && (!File.directory?(bundle_home) || !File.writable?(bundle_home))
           "`#{home}` is not writable."
         end
 
--- a/spec/bundler/bundler_spec.rb
+++ b/spec/bundler/bundler_spec.rb
@@ -199,6 +199,57 @@ EOF
         allow(File).to receive(:writable?).with(path).and_return true
         expect(Bundler.user_home).to eq(Pathname(path))
       end
+
+      context "is not a directory" do
+        it "should issue a warning and return a temporary user home" do
+          path = "/home/oggy"
+          allow(Bundler.rubygems).to receive(:user_home).and_return(path)
+          allow(File).to receive(:directory?).with(path).and_return false
+          allow(Etc).to receive(:getlogin).and_return("USER")
+          allow(Dir).to receive(:tmpdir).and_return("/TMP")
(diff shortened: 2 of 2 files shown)


lib/bundler.rb
```diff
@@ -159,12 +159,13 @@ def ruby_scope
     def user_home
       @user_home ||= begin
         home = Bundler.rubygems.user_home
+        user_dirs = home ? %w[.bundle .gem].map {|path| File.join(home, path) } : []
```
segiddins: why is `.gem` called out here? I could set my `GEM_HOME` to something other than `~/.gem`

brodock: Good question, I didn't think too much on the required folders. I think now we only need the `.bundle`.

brodock:
@segiddins code now checks only `.bundle` directory

colby-swandale:
Is the ‘.bundle’ folder not defined anywhere in SharedHelpers or in Bundler? I’m concerned about having a hard coded string.

brodock:
@colby-swandale There are other 3 relevant hard-coded `'.bundle'` strings in the same `lib/bundler.rb`:

https://github.com/brodock/bundler/blob/facf62525b61f675e2ec6f54099aa36102c254a9/lib/bundler.rb#L200

https://github.com/brodock/bundler/blob/facf62525b61f675e2ec6f54099aa36102c254a9/lib/bundler.rb#L245

https://github.com/brodock/bundler/blob/facf62525b61f675e2ec6f54099aa36102c254a9/lib/bundler.rb#L274

colby-swandale:
Thanks for finding that out, we will have to refactor that out in the future.

lib/bundler.rb
```diff
@@ -159,12 +159,13 @@ def ruby_scope
     def user_home
       @user_home ||= begin
         home = Bundler.rubygems.user_home
+        dotbundle = home ? File.join(home, ".bundle") : nil
```
colby-swandale: This is super nitpicky but can we rename this to `bundle_home`. `dotbundle` doesn't really mean anything. :pray:

brodock:
@colby-swandale done :)

colby-swandale:
Thanks! @bundlerbot r+

Issue #6541, opened 2018-05-20
Error Report


Reetok3:
--- ERROR REPORT TEMPLATE -------------------------------------------------------
# Error Report

## Questions

Please fill out answers to these questions, it'll help us figure out
why things are going wrong.

- **What did you do?**

  I ran the command `/usr/local/bin/bundle init`

- **What did you expect to happen?**

  I expected Bundler to...

- **What happened instead?**

  Instead, what happened was...

- **Have you tried any solutions posted on similar issues in our issue tracker, stack overflow, or google?**

  I tried...

- **Have you read our issues document, https://github.com/bundler/bundler/blob/master/doc/contributing/ISSUES.md?**

  ...

## Backtrace

```
Errno::EACCES: Permission denied @ rb_sysopen - Gemfile
  /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/fileutils.rb:1393:in `initialize'
  /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/fileutils.rb:1393:in `open'
  /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/fileutils.rb:1393:in `block in copy_file'
  /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/fileutils.rb:1392:in `open'
  /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/fileutils.rb:1392:in `copy_file'
  /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/fileutils.rb:486:in `copy_file'
  /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/fileutils.rb:403:in `block in cp'
  /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/fileutils.rb:1571:in `block in fu_each_src_dest'
  /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/fileutils.rb:1587:in `fu_each_src_dest0'
  /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/fileutils.rb:1569:in `fu_each_src_dest'
[...]

PR #6554, opened 2018-05-25
Use filesystem_access when reading a Gemfile

segiddins:

The problem was users would see an issue template when their `Gemfile` did not have read permissions

Fixes https://github.com/bundler/bundler/issues/6541

My fix is to use `SharedHelpers.filesystem_access` when reading the Gemfile


segiddins:
@Bundlerbot r+

Issue #6553, opened 2018-05-24
Provide a way to update dependency version constraints in the Gemfile


thisismydesign:
I see people torn between [pessimistic and optimistic](https://guides.rubygems.org/patterns/#declaring-dependencies) declaration of dependencies with semantic versioned gems (i.e. `~> 1.1` vs `>= 1.1`) because the former does not allow tools to automatically update dependencies to their newest versions (e.g. a tool that uses bundler to automatically open PRs with dependency updates).

It would be great to have a feature that can update dependency version constraints in the Gemfile so people can take advantage of both.

segiddins:
cc @agrim123

agrim123:
We need to add a few flags to the `bundle install` command, I suppose? So that users have flexibility over version of gems that can affect their code.
Can we also add these flags to `bundle add` maybe? Currently, we are using the pessimistic way when adding gems.
Should we also look to implement something to update the declaration? say from optimistic to pessimistic?

agrim123:
Basic specs can be 
```ruby
context "version specifications" do
  it "adds optimistic version" do
    bundle! "add 'foo' --optimistic"
    expect(bundled_app("Gemfile").read).to include %(gem "foo", ">= 2.0")
  end

  it "adds strict version" do
    bundle! "add 'foo' --strict"
    expect(bundled_app("Gemfile").read).to include %(gem "foo", "= 2.0")
  end

  it "adds pessimistic version" do
    bundle! "add 'foo'"
    expect(bundled_app("Gemfile").read).to include %(gem "foo", "~> 2.0")
  end
end
```

segiddins:
In my mind, it makes more sense to make another command, something like `bundle edit-gemfile`, but I’m not sure

agrim123:
Do you mean a command that would update gems version inidividually?

agrim123:
I am adding these two options to `bundle add` for now. Will open a PR regarding this by today evening. Then we can discuss more.

agrim123:
I did some experimenting [here](https://github.com/agrim123/bundler/commit/323fd2a567209d2203c102095c17a39c9d636acf)

thisismydesign:
@agrim123 @segiddins Thanks for picking this up so quickly.

Would clarify my use case with an example.

```
$ cat Gemfile | grep rails
=> gem "rails", "~> 5.1.4"

$ bundle new-feature-update-constraints

$ cat Gemfile | grep rails
=> gem "rails", "~> 5.2"
```

[...]

PR #6556, opened 2018-05-28
[bundle add] Add version prefix flexibility

agrim123:

By default, on `bundle add` we use "pessimistic" way (~>) of versions. According to this [comment](https://github.com/bundler/bundler/issues/6553#issue-326023952) some users face problems.

Adding flags to provide flexibility to change this declaration namely `optimistic` and `strict`

Adding flags to opt for other declarations.

Currently, its an experiment and I have added it to only `bundle add` but still the version locked in lockfile are "pessimistic". Need suggestions on this and on how to proceed.

Addresses #6553

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -335,6 +335,8 @@ module Bundler
     method_option "source", :aliases => "-s", :type => :string
     method_option "skip-install", :type => :boolean, :banner =>
       "Adds gem to the Gemfile but does not install it"
+    method_option "optimistic", :type => :boolean
+    method_option "strict", :type => :boolean
     def add(gem_name)
       require "bundler/cli/add"
       Add.new(options.dup, gem_name).run
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -335,8 +335,8 @@ module Bundler
     method_option "source", :aliases => "-s", :type => :string
     method_option "skip-install", :type => :boolean, :banner =>
       "Adds gem to the Gemfile but does not install it"
-    method_option "optimistic", :type => :boolean
-    method_option "strict", :type => :boolean
+    method_option "optimistic", :type => :boolean, :banner => "Adds optimistic declaration of version to gem"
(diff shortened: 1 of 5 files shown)


lib/bundler/cli.rb
```diff
@@ -335,6 +335,8 @@ def binstubs(*gems)
     method_option "source", :aliases => "-s", :type => :string
     method_option "skip-install", :type => :boolean, :banner =>
       "Adds gem to the Gemfile but does not install it"
+    method_option "optimistic", :type => :boolean
```
segiddins: these need `banner`s

spec/commands/add_spec.rb
```diff
@@ -116,4 +116,28 @@
...
+
+  describe "with --optimistic" do
+    it "adds optimistic version" do
+      bundle! "add 'foo' --optimistic"
+      expect(bundled_app("Gemfile").read).to include %(gem "foo", ">= 2.0")
+      expect(the_bundle).to include_gems "foo 2.0"
+    end
+  end
+
+  describe "with --strict option" do
```
segiddins: should we error when passing both` --optimistic` and `--strict` ?

agrim123: Yup. adding it.

agrim123:
If I specify version then a strict declaration is added and these options have no affect. Should we disable the use of these options when version is specified? or inco-oporate these flags with the version also?

segiddins:
I think it’s fine to allow the specific requirement given to override the flags?

agrim123:
But if we provide flexibility then I think we should allow 🤔 , any side effects for this?
And we need to figure out a new command may be for this purpose?

segiddins:
> But if we provide flexibility then I think we should allow 🤔 , any side effects for this?
> And we need to figure out a separate command may be for this purpose?

For what ?

agrim123:
I meant something like this
```bash
bundle add rack --version=1.1 --optimistic
```
This would do something like this
```ruby
gem "rack", ">= 1.1"
```

colby-swandale:
Great! This looks amazing! @bundlerbot r+

PR #6558, opened 2018-05-31
No need to setup BUNDLE_BIN in using DynamicCommand with --path

ta1kt0me:

When we use a bundler's custom sub command provided by gem with `--path` option, we must use BUNDLE_BIN or bundle config bin to add it to PATH.

We can use BUNDLE_BIN as workaround, but I think it's easy for user to use the custom sub command with no option like as normal command installed by gem just after `bundle install`.

I've changed to use a new `CustomSubCommand` class rather than Thor's `DynamicCommand` class directly in custom sub command.

It is the responsibility of Bundler to setup for its environment.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -30,6 +30,17 @@ module Bundler
       end
     end
 
+    def self.dynamic_command_class
+      CustomSubCommand
+    end
+
+    class CustomSubCommand < DynamicCommand
+      def run(*)
+        SharedHelpers.set_bundle_environment
+        super
+      end
+    end
+
     def initialize(*args)
       super
 
--- a/spec/bundler/cli_spec.rb
+++ b/spec/bundler/cli_spec.rb
@@ -14,6 +14,7 @@ RSpec.describe "bundle executable" do
   end
 
   it "looks for a binary and executes it if it's named bundler-<task>" do
+    Dir.mkdir ".bundle"
     File.open(tmp("bundler-testtasks"), "w", 0o755) do |f|
       f.puts "#!/usr/bin/env ruby\nputs 'Hello, world'\n"
     end
@@ -88,6 +89,7 @@ RSpec.describe "bundle executable" do
   describe "printing the outdated warning" do
     shared_examples_for "no warning" do
       it "prints no warning" do
+        Dir.mkdir ".bundle"
         bundle "fail"
         expect(last_command.stdboth).to eq("Could not find command \"fail\".")
       end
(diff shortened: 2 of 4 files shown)


segiddins:
Could you please provide a bit more clarity into what issue you're trying to address?

ta1kt0me:
Im my case, I use [bundler_diffgems](https://github.com/sinsoku/bundler_diffgems) gem which provides `diffgems` sub command like,

```dockerfile
FROM ruby:2.5.1

RUN bundle init
ENV BUNDLE_BIN=/usr/local/bundle/bin
RUN echo "gem 'bundler_diffgems'" >> Gemfile
RUN bundle install --path vendor/bundle
CMD ["/bin/bash"]
# Then execute `$ docker run IMAGE_ID bundle diffgems -h`
```

We must set `BUNDLE_BIN` or add `vendor/bundle/ruby/2.5.0/bin` to `PATH`. If not, `Could not find command "diffgems"` error occurrs. This change will decrease the settings like,

```dockerfile
FROM ruby:2.5.1

RUN bundle init
RUN echo "gem 'bundler_diffgems'" >> Gemfile
RUN bundle install --path vendor/bundle
CMD ["/bin/bash"]
# Then execute `$ docker run IMAGE_ID bundle diffgems -h`
```

colby-swandale:
I don't agree with this change in principle. We shouldn't be adding support for code to modify Bundler when we have a Plugin API.

ta1kt0me:
> What's your bundle env ?

https://gist.github.com/ta1kt0me/db0fcc503030cd6949b761983f8a8e6c

> I don't agree with this change in principle. We shouldn't be adding support for code to modify Bundler when we have a Plugin API.

As a different idea, I thought about below, how is it?

```diff
diff --git a/lib/bundler/cli.rb b/lib/bundler/cli.rb
index f3340f7cf..2e8139533 100644
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -131,6 +131,8 @@ module Bundler
         return Bundler::Plugin.exec_command(command, ARGV[1..-1])
       end

+      Bundler::SharedHelpers.set_bundle_environment
+
       return super unless command_path = Bundler.which("bundler-#{command}")

       Kernel.exec(command_path, *ARGV[1..-1])
```

segiddins:
> As a different idea, I thought about below, how is it?

I'm still not sure why that's necessary -- can you maybe write a test that demonstrates the situation you're trying to address?

colby-swandale:
Closing due to no reply.

PR #6565, opened 2018-06-05
Enable bundler caching for generated travis config

ianks:

Whenever I create a gem using `bundle gem`, I always have to manually specify that I would like travis to cache bundler.

Specify `cache: bundler` in the generated .travis.yml.


colby-swandale:
I like this idea, thanks! @bundlerbot r+

PR #6570, opened 2018-06-07
select a pre-release if it's specified in the Gemfile

### LINKED ISSUE #6449: Error when resolving dependencies
This is a simple script that reproduces the issue.

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  # Activate the gem you are reporting the issue against.
  gem "activesupport", "5.2.0.rc1"
  gem "rspec-rails", "~> 3.7.2"
end
```

When you execute it you will get:

```
Bundler could not find compatible versions for gem "actionpack": (Bundler::VersionConflict)
  In Gemfile:
    rspec-rails (~> 3.7.2) was resolved to 3.7.2, which depends on
      actionpack (>= 3.0)

Could not find gem 'actionpack (>= 3.0)', which is required by gem 'rspec-rails (~> 3.7.2)', in any of the sources.

Bundler could not find compatible versions for gem "activesupport":
  In Gemfile:
    activesupport (= 5.2.0.rc1)

    rspec-rails (~> 3.7.2) was resolved to 3.7.2, which depends on
      actionpack (>= 3.0) was resolved to 3.0.0, which depends on
        activesupport (= 3.0.0)
```

It may be related with pre releases since if I change "5.2.0.rc1" with "5.1.0" it works.

Rubygem vesion: 2.7.6
Bundler version: 1.16.1
Ruby version: 2.5.0

akihiro17:

The problem was #6449

Closes #6449 

My diagnosis was that Bundler didn't select a pre-release even if one of the dependencies of a gem was specified with a pre-release version in the Gemfile.

My fix is to change `Bundler::Resolver#requirement_satisfied_by?` so that it does not reject a pre-release if at least one of the dependencies is pre-released.

### DIFF
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -183,10 +183,6 @@ module Bundler
 
     def requirement_satisfied_by?(requirement, activated, spec)
       return false unless requirement.matches_spec?(spec) || spec.source.is_a?(Source::Gemspec)
-      if spec.version.prerelease? && !requirement.prerelease? && search_for(requirement).any? {|sg| !sg.version.prerelease? }
-        vertex = activated.vertex_named(spec.name)
-        return false if vertex.requirements.none?(&:prerelease?)
-      end
       spec.activate_platform!(requirement.__platform) if !@platforms || @platforms.include?(requirement.__platform)
       true
     end
--- a/spec/resolver/basic_spec.rb
+++ b/spec/resolver/basic_spec.rb
@@ -80,6 +80,13 @@ RSpec.describe "Resolving" do
     should_resolve_as %w[need-pre-1.0.0 activesupport-3.0.0.beta1]
   end
 
+  it "selects a pre-release if it's specified in the Gemfile" do
+    dep "activesupport", "= 3.0.0.beta"
+    dep "actionpack"
+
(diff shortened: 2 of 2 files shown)


segiddins:
Thanks! 

@greysteil any objections to me merging this?

greysteil:
@akihiro17 - nice work on this! It's a thorny part of the resolution process to be digging into. I've confirmed locally that this fixes the issue and the test is 👌.

As I understand it, the logic here is "allow a pre-release version for a non-pre-release requirement if any of the sub-dependency requirements of that pre-releases version specify pre-releases".

I don't think that logic makes sense, and think this might instead be working because actually we can remove line 186-189 entirely. It was added before #6014 was, and I *think* #6014 should do everything required to ensure pre-releases aren't selected when another option is possible.

What do you reckon? Am I missing something in the logic here?

akihiro17:
Thank you for your review.

I think we can remove pre-release logic(line 186-19) from `requirement_satisfied_by?` because #6014 seems to avoid pre-release when a non-pre-release version is possible.

greysteil:
Awesome - do you want to update this PR with that change? I'm then 👍. @segiddins - sound good to you?

akihiro17:
I'll update the PR soon.

Thanks.

segiddins:
This seems fine? But with this sort of changes, I've learned it's hard to be 100% sure ahead of time

@bundlerbot r+

Issue #6563, opened 2018-06-05
NoMethodError: undefined method `remotes' for #<Bundler::Source::Gemspec:0x47101377289620 source at `.`> Did you mean?  remote!


513ry:
I am experiencing this problem when trying to run `bundler update` or `bundler install`.
My Bundler version is 1.16.2.

## Backtrace
```
Did you mean?  remote!
  /usr/local/lib/ruby/gems/2.5.0/gems/bundler-1.16.2/lib/bundler/installer/parallel_installer.rb:164:in `rescue in do_install'
  /usr/local/lib/ruby/gems/2.5.0/gems/bundler-1.16.2/lib/bundler/installer/parallel_installer.rb:161:in `do_install'
  /usr/local/lib/ruby/gems/2.5.0/gems/bundler-1.16.2/lib/bundler/installer/parallel_installer.rb:147:in `install_serially'
  /usr/local/lib/ruby/gems/2.5.0/gems/bundler-1.16.2/lib/bundler/installer/parallel_installer.rb:102:in `call'
  /usr/local/lib/ruby/gems/2.5.0/gems/bundler-1.16.2/lib/bundler/installer/parallel_installer.rb:78:in `call'
  /usr/local/lib/ruby/gems/2.5.0/gems/bundler-1.16.2/lib/bundler/installer.rb:266:in `install_in_parallel'
  /usr/local/lib/ruby/gems/2.5.0/gems/bundler-1.16.2/lib/bundler/installer.rb:202:in `install'
  /usr/local/lib/ruby/gems/2.5.0/gems/bundler-1.16.2/lib/bundler/installer.rb:91:in `block in run'
  /usr/local/lib/ruby/gems/2.5.0/gems/bundler-1.16.2/lib/bundler/process_lock.rb:19:in `rescue in lock'
  /usr/local/lib/ruby/gems/2.5.0/gems/bundler-1.16.2/lib/bundler/process_lock.rb:6:in `lock'
  /usr/local/lib/ruby/gems/2.5.0/gems/bundler-1.16.2/lib/bundler/installer.rb:72:in `run'
  /usr/local/lib/ruby/gems/2.5.0/gems/bundler-1.16.2/lib/bundler/installer.rb:25:in `install'
  /usr/local/lib/ruby/gems/2.5.0/gems/bundler-1.16.2/lib/bundler/cli/install.rb:65:in `run'
[...]

PR #6571, opened 2018-06-08
Check if source responds to `#remotes` before printing gem install error message

colby-swandale:

There is a bug that was introduced in #6211 that prints an error message during `bundle install` if a gem failed to be installed. The error message attempts to call the `#remotes` method on `Bundler::Source::Git` which it does not implement.

See #6563

Check if the `source` responds to `#remotes` before and return early

This seems to be the most simplest fix without having to refactor a lot of code.


indirect:
good catch! @bundlerbot r+

Issue #6564, opened 2018-06-05
[Feature request] Specifying groups for `bundle list`


ilikebits:
For example, I would like to be able to `bundle list --without test` or `bundle list --without development` (or perhaps `bundle list --only default`?) so that I can list only my production dependencies. My use case is that I'd like to do dependency analysis (in this case, checking for license compliance) on only the gems that are used in production.

colby-swandale:
I like this idea but i'm a bit weary of adding the same options from `bundle install`

agrim123:
I would like to take up this task if we are planning on implementing these options 😄.

agrim123:
@ilikebits I did some experimenting [here](https://github.com/agrim123/bundler/tree/agr-bundle-list-options). Let me know if this was what you were requesting.
@colby-swandale Should I go ahead with PR?

colby-swandale:
no need to ask, more then welcome to make a PR

agrim123:
@ilikebits Should we close this issue? 😄

PR #6572, opened 2018-06-08
[bundle list] add `--without-group` and `--only-group`

agrim123:
Listing gems according to groups, either excluding a particular or from a specific group.

Usage:
```bash
bundle list --without-group dev
```

```bash
bundle list --only-group dev
```

Addresses #6564

### DIFF
--- a/lib/bundler/cli/list.rb
+++ b/lib/bundler/cli/list.rb
@@ -11,9 +11,11 @@ module Bundler
 
       raise InvalidOption, "The `--name-only` and `--paths` options cannot be used together" if @options["name-only"] && @options[:paths]
 
-      specs = Bundler.load.specs.reject {|s| s.name == "bundler" }.sort_by(&:name)
-
-      specs = filtered_specs_by_groups(specs) if @options[:only] || @options[:without]
+      specs = if @options[:only] || @options[:without]
+        filtered_specs_by_groups
+      else
+        Bundler.load.specs
+      end.reject {|s| s.name == "bundler" }.sort_by(&:name)
 
       return Bundler.ui.info "No gems in the Gemfile" if specs.empty?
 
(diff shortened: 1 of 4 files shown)


spec/commands/list_spec.rb
```diff
@@ -4,21 +4,70 @@
...
+      end
+    end
+  end
+
+  describe "with only option" do
+    context "when group is present" do
+      it "prints the gems in the specified group" do
+        bundle! "list --only default"
+
+        expect(out).to include("  * rack (1.0.0)")
```
ilikebits: You might also want to ensure that the output does _not_ include gems in other groups.

colby-swandale:
Thank you for looking into adding the feature but I don't like the idea of bringing the same `--with` and `--without` options from `bundle install`. My concern is that these options are for performing an operation on a different unit (ie: this command lists gems, not groups) and those options really don't convey the intent very well without having to lookup the documentation. Maybe instead just have `--filter-by-group`? i'm not really sure what would be better in this context
 
My other concern is how complex this PR is now making the list command. Having 2 different ways to filter the output seems over complicated and unnecessary.

lib/bundler/cli/list.rb
```diff
@@ -7,19 +7,54 @@ def initialize(options)
...
+  private
+
+    def verify_group_exists
+      raise InvalidOption, "`#{@options[:without]}` group could not be found." if @options[:without] && !@definition.groups.include?(@options[:without].to_sym)
+
+      raise InvalidOption, "`#{@options[:only]}` group could not be found." if @options[:only] && !@definition.groups.include?(@options[:only].to_sym)
+    end
+
+    def filtered_specs_by_groups(specs)
+      builder = Dsl.new
```
colby-swandale: Why are you parsing the Gemfile again?

agrim123: To filter by groups. Does spec set contains gems groups?

agrim123:
> Maybe instead just have --filter-by-group? i'm not really sure what would be better in this context

How about `--without-group` and `--only-group` maybe?

lib/bundler/cli/list.rb
segiddins: the definition has the ability to restrict to a group (and its transitive dependencies). See `Definition#specs_for`

agrim123: Also added conditional for groups case

segiddins:
@colby-swandale r?

colby-swandale:
@bundler r+

hsbt:
@bundlerbot r+

Issue #6566, opened 2018-06-06
Error install rails on NFS


brabo-hi:
I tried to install rails on mac on NFS, and got that error

**rails new blog**

`## Backtrace

```
Errno::ENOTSUP: Operation not supported @ rb_file_flock - /nfs/2017/b/brabo-hi/.rvm/gems/ruby-2.3.1/bundler.lock
  /nfs/2017/b/brabo-hi/.rvm/gems/ruby-2.3.1/gems/bundler-1.16.2/lib/bundler/process_lock.rb:10:in `flock'
  /nfs/2017/b/brabo-hi/.rvm/gems/ruby-2.3.1/gems/bundler-1.16.2/lib/bundler/process_lock.rb:10:in `block in lock'
  /nfs/2017/b/brabo-hi/.rvm/gems/ruby-2.3.1/gems/bundler-1.16.2/lib/bundler/process_lock.rb:9:in `open'
  /nfs/2017/b/brabo-hi/.rvm/gems/ruby-2.3.1/gems/bundler-1.16.2/lib/bundler/process_lock.rb:9:in `lock'
  /nfs/2017/b/brabo-hi/.rvm/gems/ruby-2.3.1/gems/bundler-1.16.2/lib/bundler/installer.rb:72:in `run'
  /nfs/2017/b/brabo-hi/.rvm/gems/ruby-2.3.1/gems/bundler-1.16.2/lib/bundler/installer.rb:25:in `install'
  /nfs/2017/b/brabo-hi/.rvm/gems/ruby-2.3.1/gems/bundler-1.16.2/lib/bundler/cli/install.rb:65:in `run'
  /nfs/2017/b/brabo-hi/.rvm/gems/ruby-2.3.1/gems/bundler-1.16.2/lib/bundler/cli.rb:224:in `block in install'
  /nfs/2017/b/brabo-hi/.rvm/gems/ruby-2.3.1/gems/bundler-1.16.2/lib/bundler/settings.rb:136:in `temporary'
  /nfs/2017/b/brabo-hi/.rvm/gems/ruby-2.3.1/gems/bundler-1.16.2/lib/bundler/cli.rb:223:in `install'
  /nfs/2017/b/brabo-hi/.rvm/gems/ruby-2.3.1/gems/bundler-1.16.2/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
[...]

PR #6573, opened 2018-06-08
handle Errno::ENOTSUP in the Bundler process lock

colby-swandale:

Bundler is trying to perform an operation on an NFS that is raising an exception (Errno::ENOTSUP)

See #6566 

Catch the exception and allow Bundler to continue

This conforms with the existing pattern in regards to handling these types of errors in the Bundler process lock logic.


segiddins:
Looks good!

@bundlerbot r+

PR #6583, opened 2018-06-14
Support URI::File of Ruby 2.6

hsbt:

When users use file protocol with bundler, URI of Ruby 2.6 will raise auth info assignment and normalize host parameter given localhost variable with its case.

URI behavior of Ruby 2.6 is reasonable for me. Because I will fix this problem on bundler side.

### DIFF
--- a/spec/commands/exec_spec.rb
+++ b/spec/commands/exec_spec.rb
@@ -789,6 +789,8 @@ __FILE__: #{path.to_s.inspect}
       end
 
       it "overrides disable_shared_gems so bundler can be found" do
+        skip "bundler 1.16.x is not support with Ruby 2.6 on Travis CI" if RUBY_VERSION >= "2.6"
+
         file = bundled_app("file_that_bundle_execs.rb")
         create_file(file, <<-RB)
           #!#{Gem.ruby}
--- a/spec/support/matchers.rb
+++ b/spec/support/matchers.rb
@@ -235,8 +235,9 @@ module Spec
       end
     end
 
-    def lockfile_should_be(expected)
-      expect(bundled_app("Gemfile.lock")).to read_as(strip_whitespace(expected))
+    def lockfile_should_be(expected, actual = bundled_app("Gemfile.lock"))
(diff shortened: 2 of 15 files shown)


hsbt:
I found the normalization of `localhost` params with Ruby 2.6. I try to fix it for the old ruby versions.

spec/support/matchers.rb
```diff
@@ -236,6 +236,7 @@ def plugin_should_not_be_installed(*names)
     end
 
     def lockfile_should_be(expected)
+      expected = expected.dup.gsub!(%r{file:\/\/localhost}, "file://") if defined?(URI::File)
```
segiddins: why is this change necessary?

segiddins:
Looks like there are still a bunch of failures on 2.6, not sure if that's expected? https://travis-ci.org/bundler/bundler/jobs/392208030

spec/support/matchers.rb
hsbt: `URI::File` normalize `localhost` params with https://github.com/ruby/ruby/blob/trunk/lib/uri/file.rb#L57

segiddins: Is that new in 2.6? If so, mind adding a comment ?

hsbt:
@segiddins I update latest commits. For now, All of the versions of Ruby works fine. Can you review again?

spec/commands/exec_spec.rb
```diff
@@ -789,6 +789,8 @@ def bin_path(a,b,c)
       end
 
       it "overrides disable_shared_gems so bundler can be found" do
+        skip "bundler 1.16.x is not support with Ruby 2.6 on Travis CI" if RUBY_VERSION >= "2.6"
```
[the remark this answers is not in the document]
hsbt: System bundler is not contained this pull request. So, It always raises `URI::File` with Ruby 2.6. After v1.16.3 contained this fix was released, We can remove this.

segiddins:
@bundlerbot r+

Issue #6582, opened 2018-06-14
Bundler 2 Release Checklist


indirect:
It’s finally here! This is the checklist and countdown for releasing Bundler 2. The items listed here are probably each an entire issue or pull request by themselves, and this issue serves as a place to track those items. This issue is also a good place to discuss adding or removing items in this list.

### Deprecations
- [ ] Modify deprecation checker to print a warning on the deprecated version, and remove the feature entirely on the major after the deprecated version. #6586
- Verify that `with_clean_env` still works but prints a warning, and stops working in Bundler 3
- Move the implementation of `with_clean_env` to `without_bundle_env`, add docs for `without_bundle_env`.
- Update `with_clean_env` warning to say something like "It turns out `with_clean_env` was confusing because it tried to do two different things. Please check the docs and use either `with_original_env` or `without_bundle_env` depending on what you need."

### Compatibility
- Install Bundler 1.x and 2.x, have an existing Rails 5.2 app continue to work without issues
- Install _only_ Bundler 2.x, have an existing Rails 5.2 app continue to work without issues on Bundler 1.x
- Upgrade an existing Rails 5.2 app from Bundler 1 to 2, continue to run without issues
- Run `bundle init; echo 'gem "rails"' >> Gemfile; bundle install; bundle exec rails new .; bundle update --bundler` to install Rails 5.2, generate a new app, bump to Bundler 2, and use Rails without issues
- Deploy a Rails 5.2 app running Bundler 2 to Heroku without issues

### Documentation
- Review and update man pages for 2.x
- Review and update bundler.io pages for 2.x
- Review and update bundler.io guides for 2.x
- Review and update the `doc/` directory for 2.x
- Add a guide to bundler.io on how to upgrade from 1.x to 2.x
- Add a guide to bundler.io for how to use Bundler 2 on Heroku
- Give bundler.io a huge header-level toggle between 1.x and 2.x docs

### Announce the release schedule
- Prereleases for 4 weeks
- Release RC, fix reported issues, release new RC
- Final release is the first RC with no reported blockers for 2 weeks

### Publish blog posts
- One blog post covering a breaking change each week
- In every post:
	- Re-explain the release schedule
	- Link to the 1.x to 2.x upgrade guide
	- Ask for testers to report issues

colby-swandale:
The Bundler 2/3 releases has changed around a bit and the TODO list for a major release is being reviewed from the feedback when we released Bundler 2. I'm going to close this as there will be a PR that will add to the repo docs.

PR #6586, opened 2018-06-15
[SharedHelpers] Only hard-error on major deprecations in the following version

segiddins:

The problem was, actually I'm not sure. @indirect ?

Addresses the first bullet point in https://github.com/bundler/bundler/issues/6582.

My diagnosis was we should only hard error on deprecations in the major version _after_ we stop supporting them.

### DIFF
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -140,12 +140,13 @@ module Bundler
     end
 
     def major_deprecation(major_version, message)
-      if Bundler.bundler_major_version >= major_version
+      bundler_major_version = Bundler.bundler_major_version
+      if bundler_major_version > major_version
         require "bundler/errors"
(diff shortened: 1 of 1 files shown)


indirect:
The problem was we removed e.g. `with_clean_env` without a deprecation period, instantly breaking any application that upgraded from Bundler 1 to 2. My recollection of the steps leading up to this is:

1. We planned to ship version 1.99 and version 2.0 at the same time, with warnings in 1.99 and breaking in 2.0.
1. We realized that most people run `gem install bundler`, and would therefore miss out on the warnings from 1.99, so we decided to drop the 1.99 release.
1. At that point, we had 2.0 with breaking changes, but no version with warnings for those breaking changes.
1. User feedback convinced us to guarantee a period of warning-without-breakage, and the easiest way to make that happen is by printing warnings in this version, and having the change break in the next version.

indirect:
@bundlerbot r+

PR #6597, opened 2018-06-22
[Feature] bundle change

### LINKED ISSUE #6581: Add `bundle change`
**Summary**
Change group/version/source of a gem

**Usage**
```bash
bundle change [gem] [--group=GROUP] [--version=VERSION] [--source=SOURCE]
```

agrim123:
Currently, this command uses `remove` and `add` internally (reason for so many commits, though only last five, are related to this command).

Closes #6581

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -356,9 +356,9 @@ module Bundler
       Add.new(options.dup, gems).run
     end
 
-    desc "change GEM [OPTIONS]", "Change gem properties"
+    desc "change GEM [OPTIONS]", "Changes properties of a gem"
     long_desc <<-D
-      Change
+      Provide flexibilty of editing gemfile from command line by providing option to change gem properties.
     D
     method_option "version", :aliases => "-v", :type => :string
     method_option "group", :aliases => "-g", :type => :string
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -350,6 +350,7 @@ module Bundler
       "Adds gem to the Gemfile but does not install it"
     method_option "optimistic", :type => :boolean, :banner => "Adds optimistic declaration of version to gem"
     method_option "strict", :type => :boolean, :banner => "Adds strict declaration of version to gem"
+    method_option "pessimistic", :type => :boolean, :banner => "Adds pessimistic declaration of version to gem"
     def add(*gems)
       require "bundler/cli/add"
       Add.new(options.dup, gems).run
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -355,6 +355,18 @@ module Bundler
       Add.new(options.dup, gems).run
     end
 
+    desc "change GEM [OPTIONS]", "Change gem properties"
+    long_desc <<-D
+      Change
+    D
+    method_option "version", :aliases => "-v", :type => :string
+    method_option "group", :aliases => "-g", :type => :string
+    method_option "source", :aliases => "-s", :type => :string
+    def change(gem_name)
+      require "bundler/cli/change"
+      Change.new(options.dup, gem_name).run
+    end
+
     desc "outdated GEM [OPTIONS]", "List installed gems with newer versions available"
     long_desc <<-D
       Outdated lists the names and versions of gems that have a newer version available
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -360,9 +360,9 @@ module Bundler
     long_desc <<-D
       Provide flexibilty of editing gemfile from command line by providing option to change gem properties.
     D
-    method_option "version", :aliases => "-v", :type => :string
-    method_option "group", :aliases => "-g", :type => :string
-    method_option "source", :aliases => "-s", :type => :string
+    method_option "version", :type => :string
+    method_option "group", :type => :string
+    method_option "source", :type => :string
     def change(gem_name)
       require "bundler/cli/change"
       Change.new(options.dup, gem_name).run
--- a/lib/bundler/cli/change.rb
+++ b/lib/bundler/cli/change.rb
@@ -18,17 +18,24 @@ module Bundler
 
       @pass_options = {}
 
+      initial_gemfile = IO.readlines(Bundler.default_gemfile)
+
       set_group_options
 
       set_version_options
 
       @pass_options[:source] = @options[:source] if @options[:source]
 
-      require "bundler/cli/remove"
-      CLI::Remove.new([@gem_name], {}).run
+      begin
+        require "bundler/cli/remove"
+        CLI::Remove.new([@gem_name], {}).run
 
-      require "bundler/cli/add"
-      CLI::Add.new(@pass_options, [@gem_name]).run
+        require "bundler/cli/add"
+        CLI::Add.new(@pass_options, [@gem_name]).run
+      rescue StandardError => e
+        Bundler.ui.error e
+        SharedHelpers.write_to_gemfile(Bundler.default_gemfile, initial_gemfile)
+      end
     end
 
   private
--- a/lib/bundler/cli/change.rb
+++ b/lib/bundler/cli/change.rb
@@ -8,12 +8,22 @@ module Bundler
     end
 
     def run
-      builder = Dsl.new
-      builder.eval_gemfile(Bundler.default_gemfile)
-
-      @definition = builder.to_definition(Bundler.default_lockfile, {})
+      @definition = Bundler.definition
 
       raise InvalidOption, "`#{@gem_name}` could not be found in the Gemfile." unless @definition.dependencies.find {|d| d.name == @gem_name }
+
+      pass_options = {
+        "group" => [],
+        "version" => nil,
+        "source" => nil
+      }
+      pass_options["group"] << @options[:group] if @options[:group]
+
+      require "bundler/cli/remove"
+      CLI::Remove.new([@gem_name], {}).run
+
+      require "bundler/cli/add"
+      CLI::Add.new(pass_options, [@gem_name]).run
     end
   end
 end
--- a/lib/bundler/cli/change.rb
+++ b/lib/bundler/cli/change.rb
@@ -63,9 +63,12 @@ module Bundler
     def set_group_options
       groups = @dep.groups
       if @options[:group]
-        Bundler.ui.warn "`#{@gem_name}` is already in `#{@options[:group]}`." if groups.include?(@options[:group].to_sym)
+        uniq_groups = @options[:group].split(",").uniq
+        common_groups = uniq_groups & groups.map(&:to_s)
 
-        @pass_options["group"] = @options[:group].to_s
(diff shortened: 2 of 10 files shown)


segiddins:
Can you change the target branch of this to `agr-bundler-remove` so it's possible to leave comments on just these changes?

agrim123:
@segiddins Now this is reviewable 😄

lib/bundler/cli/change.rb
```diff
@@ -0,0 +1,77 @@
...
+      @pass_options["source"] = @options[:source] if @options[:source]
+
+      begin
+        require "bundler/cli/remove"
+        CLI::Remove.new([@gem_name], {}).run
+
+        require "bundler/cli/add"
+        CLI::Add.new(@pass_options, [@gem_name]).run
+      rescue StandardError => e
+        Bundler.ui.error e
```
segiddins: should the error be re-raised instead, after restoring the initial gemfile?

lib/bundler/cli/change.rb
```diff
@@ -0,0 +1,77 @@
...
+        SharedHelpers.write_to_gemfile(Bundler.default_gemfile, initial_gemfile)
+      end
+    end
+
+  private
+
+    def set_version_options
```
segiddins: can you add a comment explaining what this method is trying to do ?

lib/bundler/shared_helpers.rb
```diff
@@ -226,6 +226,8 @@ def digest(name)
       Digest(name)
     end
 
+    # @param [Pathname] gemfile_path  The Gemfile from which to remove dependencies.
+    # @param [String]  contents       Content to written to Gemfile.
     def write_to_gemfile(gemfile_path, contents)
```
segiddins: this should probably just be `write_file`

spec/commands/change_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+
+      expect(out).to include("`rake` could not be found in the Gemfile.")
+    end
+  end
+
+  context "without options" do
+    it "throws error" do
+      bundle "change rack"
+
+      expect(out).to include("Please supply atleast one option to change.")
```
segiddins: `at least`

lib/bundler/cli/change.rb
```diff
@@ -0,0 +1,77 @@
...
+
+      @pass_options = {}
+
+      initial_gemfile = IO.readlines(Bundler.default_gemfile)
+
+      set_group_options
+
+      set_version_options
+
+      @pass_options["source"] = @options[:source] if @options[:source]
```
segiddins: should this also handle path/git ?

lib/bundler/cli/change.rb
agrim123: It is shown as error do we need to raise an exception?

segiddins: yes, we should let the `FriendlyErrors` module handle showing the error to the user, and ensure we exit with the proper exit code

lib/bundler/cli/change.rb
agrim123: Yeah, it should. Adding them.

agrim123: Should they be editable?

agrim123: Also @segiddins , [these](https://github.com/bundler/bundler/blob/master/lib/bundler/dsl.rb#L18-L19) are valid keys we also need to persist them! I think we may have to change our approach?

segiddins: yeah we definitely need to persist all keys

lib/bundler/cli.rb
```diff
@@ -350,11 +350,24 @@ def binstubs(*gems)
       "Adds gem to the Gemfile but does not install it"
     method_option "optimistic", :type => :boolean, :banner => "Adds optimistic declaration of version to gem"
     method_option "strict", :type => :boolean, :banner => "Adds strict declaration of version to gem"
+    method_option "pessimistic", :type => :boolean, :banner => "Adds pessimistic declaration of version to gem"
```
segiddins: this option needs to be tested

lib/bundler/cli.rb
```diff
@@ -350,11 +350,24 @@ def binstubs(*gems)
...
       require "bundler/cli/add"
       Add.new(options.dup, gems).run
     end
 
+    desc "change GEM [OPTIONS]", "Changes properties of a gem"
+    long_desc <<-D
+      Provide flexibilty of editing gemfile from command line by providing option to change gem properties.
+    D
+    method_option "version", :aliases => "-v", :type => :string
+    method_option "group", :aliases => "-g", :type => :string
```
segiddins: let's not add this aliases for now

lib/bundler/cli/change.rb
```diff
@@ -0,0 +1,81 @@
...
+
+      begin
+        require "bundler/cli/remove"
+        CLI::Remove.new([@gem_name], {}).run
+
+        require "bundler/cli/add"
+        CLI::Add.new(@pass_options, [@gem_name]).run
+      rescue StandardError => e
+        SharedHelpers.write_file(Bundler.default_gemfile, initial_gemfile)
+        FriendlyErrors.log_error(e)
```
segiddins: I think instead of this, we should just `raise`

lib/bundler/cli/change.rb
```diff
@@ -0,0 +1,81 @@
...
+    def run
+      raise InvalidOption, "Please supply at least one option to change." unless @options[:group] || @options[:version] || @options[:source]
+
+      definition = Bundler.definition
+
+      @dep = definition.dependencies.find {|d| d.name == @gem_name }
+
+      raise InvalidOption, "`#{@gem_name}` could not be found in the Gemfile." unless @dep
+
+      @pass_options = {}
```
segiddins: maybe call this `add_options`, and instead of making it an ivar, pass it into `set_group_options` / `set_version_options`

lib/bundler/cli/change.rb
```diff
@@ -0,0 +1,81 @@
...
+      @gem_name = gem_name
+      @options = options
+    end
+
+    def run
+      raise InvalidOption, "Please supply at least one option to change." unless @options[:group] || @options[:version] || @options[:source]
+
+      definition = Bundler.definition
+
+      @dep = definition.dependencies.find {|d| d.name == @gem_name }
```
segiddins: same as below, shouldn't be an ivar

man/bundle-change.ronn
```diff
@@ -0,0 +1,23 @@
...
+================================================================
+
+## SYNOPSIS
+
+`bundle change` <GEM_NAME> [--group=GROUP] [--version=VERSION] [--source=SOURCE]
+
+## DESCRIPTION
+Provide flexibilty of editing gemfile from command line by providing option to change gem properties.
+
+Example:
```
segiddins: this example should show the before & after gemfiles

spec/commands/change_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+          source "file://#{gem_repo1}"
+
+          gem "weakling", ">=  0.0.1"
+
+          group :test do
+            gem "rack-test", "= 1.0"
+            gem "rspec"
+          end
+
+          gem "rack", "~> 1.0", :group => [:dev1]
```
segiddins: this should either be `:group => :dev1` or `:groups => [:dev1]`

spec/commands/change_spec.rb
```diff
@@ -0,0 +1,158 @@
+# frozen_string_literal: true
+
+RSpec.describe "bundle change" do
+  before :each do
+    install_gemfile <<-G
+      source "file://#{gem_repo1}"
+
+      gem "rack", "~> 1.0", :group => [:dev]
+      gem "weakling", ">=  0.0.1"
```
segiddins: please add a test that changes both of these 2 gems into the `test` group

spec/commands/change_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+      end
+    end
+  end
+
+  describe "with --version option" do
+    context "when specified version exists" do
+      it "changes version of the gem" do
+        bundle! "change rack --version 0.9.1"
+
+        expect(bundled_app("Gemfile").read).to include("gem \"rack\", \"~> 0.9.1\", :group => [:dev]")
```
segiddins: use single quotes to avoid having to escape the double quotes

spec/commands/change_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+        gem "rspec"
+      end
+    G
+  end
+
+  describe "when gem is not present" do
+    it "throws error" do
+      bundle "change rake --group dev1"
+
+      expect(out).to include("`rake` could not be found in the Gemfile.")
```
segiddins: please test that failure does not change the gemfile

```ruby
expect { bundle ... }.not_to change { gemfile }
```

lib/bundler/cli/change.rb
agrim123: After rescue then again we should raise error?

spec/commands/change_spec.rb
agrim123: We will have to change the injector https://github.com/bundler/bundler/blob/master/spec/commands/add_spec.rb#L72

spec/commands/change_spec.rb
agrim123: No operation is done on gemfile until this operation.

lib/bundler/cli/change.rb
segiddins: yes -- the rescue here is only to restore the initial gemfile

spec/commands/change_spec.rb
segiddins: right but we should test that that's the case :)

segiddins:
This looks like a great 1st implementation of the `bundle change` feature to me! Well done @agrim123 !

@colby-swandale @deivid-rodriguez review would be appreciated 🙏

lib/bundler/cli/change.rb
```diff
@@ -14,6 +14,8 @@ def run
 
       raise InvalidOption, "`#{@gem_name}` could not be found in the Gemfile." unless dep
 
+      check_for_unsupported_options(dep)
```
agrim123: I am in process of writing for this function. Still to figure out how to skip before hook for one test.

spec/commands/change_spec.rb
agrim123: Handled [here](https://github.com/bundler/bundler/pull/6627)

segiddins (approved):

agrim123:
@segiddins @colby-swandale  The milestone seems to be have passed :sweat_smile: 
Should this be merged now?

fatkodima:
Is it just me, but I don't see any benefits of adding and using this? It is **much** easier to make changes directly in Gemfile and run `bundle install`/`bundle update gemname`, than remembering that there is yet another `change` command and how to use it, what are the arguments.

PR #6599, opened 2018-06-24
Respect encodings when reading gemspecs

### LINKED ISSUE #6598: Bundler 1.16.2 regression: gemspec referencing code with non-ascii characters requires "encoding: UTF-8" magic comment
Hello there and thanks for Bundler!

I'm the author of the [Persistent-💎](https://rubygems.org/gems/persistent-dmnd/) gem, which uses as its namespace a `module Persistent💎`.

The [`gemspec`](https://gitlab.com/ivoanjo/persistent-dmnd/blob/master/persistent-dmnd.gemspec) for the gem thus includes:

```ruby
Gem::Specification.new do |spec|
  spec.name          = 'persistent-dmnd'
  spec.version       = Persistent💎::VERSION
```

Because I support this gem also on Ruby 1.9.3, [I've added a `# encoding: UTF-8` at the top of every file](https://gitlab.com/ivoanjo/persistent-dmnd/commit/29d0977df8429ed1617f07d325b3b6169dc6154e) so Ruby correctly handles the emoji.

But, while debugging a related TruffleRuby issue (see https://github.com/oracle/truffleruby/issues/1376) we discovered that whereas with Bundler 1.16.1 the gem's gemspec file works correctly on Ruby 2.0->2.6 without a `# encoding: UTF-8` comment at the top of the file:

```
$ ruby -v
ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux]
$ bundle -v
Bundler version 1.16.1
$ git diff
diff --git a/persistent-dmnd.gemspec b/persistent-dmnd.gemspec
index 72c50a6..e4b3dad 100644
--- a/persistent-dmnd.gemspec
+++ b/persistent-dmnd.gemspec
@@ -1,4 +1,3 @@
-# encoding: UTF-8
 
 # Persistent-💎: Ruby gem for easily creating immutable data structures
 # Copyright (c) 2017 Ivo Anjo <ivo.anjo@ist.utl.pt>
$ bundle install
Warning: the running version of Bundler (1.16.1) is older than the version that created the lockfile (1.16.2). We suggest you upgrade to the latest version of Bundler by running `gem install bundler`.
[snip]
Bundle complete! 9 Gemfile dependencies, 18 gems now installed.
```

...on bundler 1.16.2 this breaks:

```
$ bundle -v
Bundler version 1.16.2
$ bundle install
Traceback (most recent call last):
        24: .rvm/gems/ruby-2.5.1/bin/ruby_executable_hooks:15:in `<main>'
        23: .rvm/gems/ruby-2.5.1/bin/ruby_executable_hooks:15:in `eval'
        22: .rvm/gems/ruby-2.5.1/bin/bundle:23:in `<main>'
        21: .rvm/gems/ruby-2.5.1/bin/bundle:23:in `load'
        20: .rvm/gems/ruby-2.5.1/gems/bundler-1.16.2/exe/bundle:22:in `<top (required)>'
        19: .rvm/gems/ruby-2.5.1/gems/bundler-1.16.2/lib/bundler/friendly_errors.rb:124:in `with_friendly_errors'
        18: .rvm/gems/ruby-2.5.1/gems/bundler-1.16.2/exe/bundle:30:in `block in <top (required)>'
        17: .rvm/gems/ruby-2.5.1/gems/bundler-1.16.2/lib/bundler/cli.rb:18:in `start'
        16: .rvm/gems/ruby-2.5.1/gems/bundler-1.16.2/lib/bundler/vendor/thor/lib/thor/base.rb:466:in `start'
        15: .rvm/gems/ruby-2.5.1/gems/bundler-1.16.2/lib/bundler/cli.rb:27:in `dispatch'
        14: .rvm/gems/ruby-2.5.1/gems/bundler-1.16.2/lib/bundler/vendor/thor/lib/thor.rb:387:in `dispatch'
        13: .rvm/gems/ruby-2.5.1/gems/bundler-1.16.2/lib/bundler/vendor/thor/lib/thor/invocation.rb:126:in `invoke_command'
        12: .rvm/gems/ruby-2.5.1/gems/bundler-1.16.2/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
        11: .rvm/gems/ruby-2.5.1/gems/bundler-1.16.2/lib/bundler/cli.rb:223:in `install'
        10: .rvm/gems/ruby-2.5.1/gems/bundler-1.16.2/lib/bundler/settings.rb:136:in `temporary'
         9: .rvm/gems/ruby-2.5.1/gems/bundler-1.16.2/lib/bundler/cli.rb:224:in `block in install'
         8: .rvm/gems/ruby-2.5.1/gems/bundler-1.16.2/lib/bundler/cli/install.rb:62:in `run'
         7: .rvm/gems/ruby-2.5.1/gems/bundler-1.16.2/lib/bundler.rb:135:in `definition'
         6: .rvm/gems/ruby-2.5.1/gems/bundler-1.16.2/lib/bundler/definition.rb:35:in `build'
         5: .rvm/gems/ruby-2.5.1/gems/bundler-1.16.2/lib/bundler/dsl.rb:12:in `evaluate'
         4: .rvm/gems/ruby-2.5.1/gems/bundler-1.16.2/lib/bundler/dsl.rb:42:in `eval_gemfile'
         3: .rvm/gems/ruby-2.5.1/gems/bundler-1.16.2/lib/bundler/dsl.rb:51:in `rescue in eval_gemfile'
         2: .rvm/gems/ruby-2.5.1/gems/bundler-1.16.2/lib/bundler/dsl.rb:51:in `message'
         1: .rvm/gems/ruby-2.5.1/gems/bundler-1.16.2/lib/bundler/dsl.rb:51:in `to_s'
.rvm/gems/ruby-2.5.1/gems/bundler-1.16.2/lib/bundler/dsl.rb:575:in `to_s': incompatible character encodings: UTF-8 and ASCII-8BIT (Encoding::CompatibilityError)
```

deivid-rodriguez:
This PR fixes #6598. Thanks @eregon for the help and the explanation that helped me understand the issue :)!

On gems using UTF-8 symbols defined in other files in their gemspecs, `bundle install` would crash.

The problem was that since #6279 gemspecs are read binarily. However, files required from them as read as text. That means that the constants defined on those files and used in the gemspec are interpreted different and thus don't match.

My fix is to go back to reading gemspec as text again. Explictly passing the encoding when reading them still fixes the problem that the PR introducing the regression was meant to fix. 

I chose this fix because it fixes the problem, and keeps the situation that the PR introducing the regression fixed also working.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -444,7 +444,7 @@ EOF
 
     def load_gemspec_uncached(file, validate = false)
       path = Pathname.new(file)
-      contents = path.read(:encoding => Encoding::UTF_8)
+      contents = File.open(file, "r:UTF-8", &:read)
       spec = if contents.start_with?("---") # YAML header
         eval_yaml_gemspec(path, contents)
       else
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -444,7 +444,7 @@ EOF
 
     def load_gemspec_uncached(file, validate = false)
       path = Pathname.new(file)
-      contents = read_file(file)
+      contents = path.read(:encoding => Encoding::UTF_8)
       spec = if contents.start_with?("---") # YAML header
         eval_yaml_gemspec(path, contents)
       else
--- a/spec/install/gemspecs_spec.rb
+++ b/spec/install/gemspecs_spec.rb
@@ -64,7 +64,7 @@ RSpec.describe "bundle install" do
   end
 
   it "reads gemspecs respecting their encoding" do
-    skip "Too cool for 1.8" if RUBY_VERSION < "1.9"
+    skip "unicode constants are most likely not supported on 1.8" if RUBY_VERSION < "1.9"
 
     create_file("version.rb", <<-RUBY)
       module Persistent💎
--- a/spec/install/gemspecs_spec.rb
+++ b/spec/install/gemspecs_spec.rb
@@ -64,6 +64,8 @@ RSpec.describe "bundle install" do
   end
 
   it "reads gemspecs respecting their encoding" do
+    skip "Too cool for 1.8" if RUBY_VERSION < "1.9"
+
     create_file("version.rb", <<-RUBY)
       module Persistent💎
         VERSION = "0.0.1"
--- a/spec/install/gemspecs_spec.rb
+++ b/spec/install/gemspecs_spec.rb
@@ -63,6 +63,31 @@ RSpec.describe "bundle install" do
     expect(out).to include("Bundle complete!")
   end
 
+  it "reads gemspecs respecting their encoding" do
+    create_file("version.rb", <<-RUBY)
+      module Persistent💎
+        VERSION = "0.0.1"
+      end
+    RUBY
+
+    create_file("persistent-dmnd.gemspec", <<-G)
+      require_relative "version"
+
+      Gem::Specification.new do |gem|
+        gem.name = "persistent-dmnd"
+        gem.version = Persistent💎::VERSION
+        gem.author = "Ivo Anjo"
+        gem.summary = "Unscratchable stuff"
+      end
+    G
+
+    install_gemfile <<-G
+      gemspec
+    G
+
+    expect(out).to include("Bundle complete!")
+  end
+
   context "when ruby version is specified in gemspec and gemfile" do
     it "installs when patch level is not specified and the version matches" do
       build_lib("foo", :path => bundled_app) do |s|


spec/install/gemspecs_spec.rb
```diff
@@ -63,6 +63,31 @@
...
+    create_file("persistent-dmnd.gemspec", <<-G)
+      require_relative "version"
+
+      Gem::Specification.new do |gem|
+        gem.name = "persistent-dmnd"
+        gem.version = Persistent💎::VERSION
+        gem.author = "Ivo Anjo"
+        gem.summary = "Unscratchable stuff"
+      end
+    G
```
deivid-rodriguez: Looking at how nice the `version.rb` file in this diff looks, we might want to change the `G` tag to `RUBY` everywhere :)

lib/bundler.rb
```diff
@@ -444,7 +444,7 @@ def load_gemspec(file, validate = false)
 
     def load_gemspec_uncached(file, validate = false)
       path = Pathname.new(file)
-      contents = read_file(file)
+      contents = path.read(encoding: Encoding::UTF_8)
```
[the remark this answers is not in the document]
deivid-rodriguez: Thanks, should be fixed now!

lib/bundler.rb
```diff
@@ -444,7 +444,7 @@ def load_gemspec(file, validate = false)
 
     def load_gemspec_uncached(file, validate = false)
       path = Pathname.new(file)
-      contents = read_file(file)
+      contents = path.read(:encoding => Encoding::UTF_8)
```
eregon: It's surprising Bundler still supports the very old Ruby 1.8.7.
Maybe `File.open(file, "r:UTF-8", &:read)` would work on 1.8.7 too.

eregon: Thanks for the info, I agree in general. I filed https://github.com/bundler/bundler/issues/6600 just before seeing this, let's see.

deivid-rodriguez: It didn't work... :disappointed: Up to the maintainers how to follow up from here.

eregon:
Maybe it's OK to skip this spec on 1.8.7?
IIRC, Ruby 1.8.7 has very limited support for Unicode characters, notably I think it doesn't quite like them in constants.
The rest seems to work fine, only this spec fails.

deivid-rodriguez:
Fine with me! Let me skip it.

spec/install/gemspecs_spec.rb
```diff
@@ -63,6 +63,33 @@
     expect(out).to include("Bundle complete!")
   end
 
+  it "reads gemspecs respecting their encoding" do
+    skip "Too cool for 1.8" if RUBY_VERSION < "1.9"
```
colby-swandale: This is not informative, please use a comment that explains why we need to skip 1.8 in this test.

deivid-rodriguez: What about "unicode was probably not fully supported on 1.8"?

deivid-rodriguez: Basically the reason is that we have decided it's not worth even trying to make this work on 1.8. We're not sure it's actually supported and blind guessing from Travis logs feedback is... tough.

deivid-rodriguez: Pushed a new skip message, hopefully more serious and informative :)

colby-swandale:
@bundlerbot r+

PR #6601, opened 2018-06-26
More flexible stub

deivid-rodriguez:

My problem is that I can't get https://github.com/rubygems/rubygems/pull/2332 CI passing.

My diagnosis was that current bundler test don't support my changes in that PR, because previously the `validate` method in `rubygems` would be called without arguments, and now it's called with two arguments.

My fix is to make the specification stub more flexible so it supports any number of arguments.

I chose this fix because it supports my changes and doesn't break the previous version either.

### DIFF
--- a/spec/support/builders.rb
+++ b/spec/support/builders.rb
@@ -601,7 +601,7 @@ module Spec
           @spec.rubygems_version = options[:rubygems_version]
           def @spec.mark_version; end
 
-          def @spec.validate; end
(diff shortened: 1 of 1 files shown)


segiddins:
@bundlerbot r+

deivid-rodriguez:
May I request this PR added to the 1.16.3 milestone in order to move forward with rubygems/rubygems#2332? Thanks! :)

PR #6602, opened 2018-06-26
Ci against latest versions

deivid-rodriguez:

My local tests don't match Travis environment, since I use ruby 2.5.1 but travis doesn't.

My diagnosis was that we should be testing against the latest patch level releases.

My fix is to update versions.

I chose this fix because I can't think of another fix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ env:
   # We need to know if changes to rubygems will break bundler on release
   - RGV=master
   # Test the latest rubygems release with all of our supported rubies
-  - RGV=v2.7.4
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
A couple of entries failed. Not sure if they're surfacing a problem in the code or the tests, but looks like there's something going on.

deivid-rodriguez:
I removed the commits breaking the build. I'll open issues for them but this one is ready!

hsbt:
@deivid-rodriguez Great! Thanks for your work.

hsbt:
@bundlerbot r+

PR #6612, opened 2018-06-30
Updated 'bundle add' to rspec install guide

hdf1986:

There's a new command `bundle add` which is almost unknown

We are promoting to use a simple append to the Gemfile when there's a `bundle add` command available

My fix is just a little readme change :sweat_smile: 

Because it's a simple fix and provides more exposure to the command


segiddins:
Awesome, thanks!

@bundlerbot r+

PR #6613, opened 2018-06-30
Document that `bundle show [--paths]` sorts results

kemitchell:

The problem was that I could not tell from the manpage or documentation website whether `bundle show` and `bundle show --paths` would list gems and gem paths in the same order.

I am using `bundle show` and `bundle show --paths` to inventory gems that projects depend upon.

My diagnosis was that the implementation of `bundle show` _does_ sort results by name, but that the manpage for the subcommand doesn't mention this.

My fix involved slightly editing the existing manpage for `bundle show`.

I chose this fix because the `.ronn` file text matched the current text I was seeing via the documentation website.

### DIFF
--- a/man/bundle-show.ronn
+++ b/man/bundle-show.ronn
@@ -9,7 +9,7 @@ bundle-show(1) -- Shows all the gems in your bundle, or the path to a gem
 ## DESCRIPTION
 
 Without the [GEM] option, `show` will print a list of the names and versions of
-all gems that are required by your [`Gemfile(5)`][Gemfile(5)].
+all gems that are required by your [`Gemfile(5)`][Gemfile(5)], sorted by name.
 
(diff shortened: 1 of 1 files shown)


kemitchell:
I forgot to link to the code lines that show sorting being done.

Here is the line for `--paths`: https://github.com/bundler/bundler/blob/56c5c119b537ce8fc2227984a9e7a2d7b00b32c1/lib/bundler/cli/show.rb#L34

Here is the line without `--paths`: https://github.com/bundler/bundler/blob/56c5c119b537ce8fc2227984a9e7a2d7b00b32c1/lib/bundler/cli/show.rb#L39

colby-swandale:
Thanks! @bundlerbot r+

kemitchell:
Thanks, Bundler team! The work you do is super important.

PR #6614, opened 2018-07-01
Update vendored molinillo to 0.6.5

segiddins:

The problem was bundler was using an outdated version of Molinillo.

My fix was to bump the version.

I chose this fix because it includes resolver performance improvements when using `recursive_predecessors` / `recursive_successors`


segiddins:
@bundlerbot r+

PR #6624, opened 2018-07-12
Prefer to use `--no-document` option in spec

hsbt:

RubyGems will remove `--no-ri` and `--no-rdoc` options at RubyGems 3.0. But bundler spec is also broken when removing them.

* https://github.com/rubygems/rubygems/pull/2354
* https://travis-ci.org/rubygems/rubygems/jobs/402911887

To use --no-document option after RubyGems 2.0.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -56,7 +56,7 @@ namespace :spec do
         sh %(#{Gem.ruby} -S gem #{gem_install_command})
       end
     else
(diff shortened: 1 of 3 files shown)


segiddins:
👍🏻

colby-swandale:
Thanks! 🙇  @bundlerbot r+

PR #6627, opened 2018-07-15
Fix singular groups on injecting gem

agrim123:

The problem was that on adding a gem to a group via
```bash
bundle add rack --group=dev
```
It gets added as 
```ruby
gem "rack", :group => [:dev]
```
It should rather be 
```ruby
gem "rack", :group => :dev
```

My diagnosis was to not add single groups in the array.

My fix was to pick the element from the array of groups and append using
```ruby
":group => :#{d.groups.first}"
```

I chose this fix because it seemed to most appropriate.

### DIFF
--- a/lib/bundler/injector.rb
+++ b/lib/bundler/injector.rb
@@ -107,7 +107,7 @@ module Bundler
         end
 
         if d.groups != Array(:default)
-          group = d.groups.size == 1 ? ", :group => #{d.groups.inspect}" : ", :groups => #{d.groups.inspect}"
+          group = d.groups.size == 1 ? ", :group => #{d.groups.first.inspect}" : ", :groups => #{d.groups.inspect}"
         end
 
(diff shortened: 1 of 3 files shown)


lib/bundler/injector.rb
```diff
@@ -107,7 +107,7 @@ def build_gem_lines(conservative_versioning)
         end
 
         if d.groups != Array(:default)
-          group = d.groups.size == 1 ? ", :group => #{d.groups.inspect}" : ", :groups => #{d.groups.inspect}"
+          group = d.groups.size == 1 ? ", :group => :#{d.groups.first}" : ", :groups => #{d.groups.inspect}"
```
segiddins: nit: `:group => #{d.groups.first.inspect}`

deivid-rodriguez:
Thanks @agrim123! :heart:

Issue #6619, opened 2018-07-06
bundle path confusing behavior


pegasd:
Environment:

```
macOS High Sierra
bundler 1.16.2
Ruby 2.4.4 (through RVM)
```

Steps to reproduce:

- Simple Gemfile for testing:

```bash
 % cat Gemfile
source 'https://rubygems.org'
gem 'rake'
```

- No local or global bundle configuration files (to begin with)

```bash
 % bundle config
Settings are listed in order of priority. The top value will be used.
```

#### Behavior 1

Attempt to set path using global config

```bash
 % bundle config path ~/.bundle
 % bundle config
Settings are listed in order of priority. The top value will be used.
path
Set for the current user (/Users/pegas/.bundle/config): "/Users/pegas/.bundle"
```

Expected location of installed gems: `~/.bundle`
Actual location of installed gems: `~/.rvm/gems` (this is the "system" gem location)

```bash
 % bundle show rake
/Users/pegas/.rvm/gems/ruby-2.4.4/gems/rake-12.3.1
```

#### Behavior 2

Attempt to disable system path

```bash
 % bundle config path.system false
 % bundle config
Settings are listed in order of priority. The top value will be used.
path
Set for the current user (/Users/pegas/.bundle/config): "/Users/pegas/.bundle"

path.system
Set for the current user (/Users/pegas/.bundle/config): false
```

Expected location of installed gems: `~/.bundle`
Actual location of installed gems: `~/.bundle` (non Ruby-version dependent)

```bash
 % bundle show rake
/Users/pegas/.bundle/gems/rake-12.3.1
```

#### Behavior 3

Attempt to set path locally
```bash
 % bundle config --local path ~/.bundle
 % bundle config
Settings are listed in order of priority. The top value will be used.
path
Set for your local app (/Users/pegas/test/.bundle/config): "/Users/pegas/.bundle"
```

Expected location of installed gems: `~/.bundle`
Actual location of installed gems: `~/.bundle` (Ruby-version dependent)

```bash
 % bundle show rake
/Users/pegas/.bundle/ruby/2.4.0/gems/rake-12.3.1
```

### Questions, thoughts

* What is the proper workflow for this use case? The only logical gem location (from my POV) results from storing local `.bundle/config` files all over my apps (`behavior 3`), which does not seem very user-friendly. How do I make a global configuration that behaves like this?
* Why is `behavior 1` not doing the same as `behavior 3`?
[...]

PR #6628, opened 2018-07-17
[Settings] Append the ruby scope on Bundler 2 with a global path setting

segiddins:

The problem was `bundle config path` behaving differently depending on whether the config is local or global was... confusing.

Closes https://github.com/bundler/bundler/issues/6619#issuecomment-405460114.

My diagnosis was setting the `path` setting should be consistent, regardless of the level of configuration it is set at.

My fix appends the "ruby scope" to the path when set globally on Bundler 2.

I chose this fix because it means users won't have to re-install gems when switching rubies with a global `BUNDLE_PATH` set.


indirect:
@bundlerbot r+

segiddins:
Took a while, but got the specs updated!

colby-swandale:
@bundlerbot r+

PR #6636, opened 2018-07-23
Don't call File.realpath in a loop, single call does what we want

ojab:

We're calling `File.realpath` in a loop

Single call already resolves symlinks recursively

Drop call in a loop


segiddins:
Thanks! @Bundlerbot r+

indirect:
@bundlerbot r+

PR #6645, opened 2018-07-31
require `etc` in Bundler#user_home

### LINKED ISSUE #6640: Read only homedir leads to #<Class:Bundler>::Etc (NameError)
When running in Alpine 3.8, 1.16.2 in Docker, as root, with a read-only /root filesystem:

```
$ bundle exec
/usr/lib/ruby/gems/2.5.0/gems/bundler-1.16.2/lib/bundler.rb:172:in `user_home': uninitialized constant #<Class:Bundler>::Etc (NameError)
    from /usr/lib/ruby/gems/2.5.0/gems/bundler-1.16.2/lib/bundler.rb:197:in `user_bundle_path'
    from /usr/lib/ruby/gems/2.5.0/gems/bundler-1.16.2/lib/bundler/settings.rb:377:in `global_config_file'
    from /usr/lib/ruby/gems/2.5.0/gems/bundler-1.16.2/lib/bundler/settings.rb:80:in `initialize'
    from /usr/lib/ruby/gems/2.5.0/gems/bundler-1.16.2/lib/bundler.rb:257:in `new'
    from /usr/lib/ruby/gems/2.5.0/gems/bundler-1.16.2/lib/bundler.rb:257:in `settings'
    from /usr/lib/ruby/gems/2.5.0/gems/bundler-1.16.2/lib/bundler/feature_flag.rb:21:in `block in settings_method'
    from /usr/lib/ruby/gems/2.5.0/gems/bundler-1.16.2/lib/bundler/cli.rb:97:in `<class:CLI>'
    from /usr/lib/ruby/gems/2.5.0/gems/bundler-1.16.2/lib/bundler/cli.rb:7:in `<module:Bundler>'
    from /usr/lib/ruby/gems/2.5.0/gems/bundler-1.16.2/lib/bundler/cli.rb:6:in `<top (required)>'
    from /usr/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
    from /usr/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
    from /usr/lib/ruby/gems/2.5.0/gems/bundler-1.16.2/exe/bundle:23:in `block in <top (required)>'
    from /usr/lib/ruby/gems/2.5.0/gems/bundler-1.16.2/lib/bundler/friendly_errors.rb:124:in `with_friendly_errors'
    from /usr/lib/ruby/gems/2.5.0/gems/bundler-1.16.2/exe/bundle:22:in `<top (required)>'
    from /usr/bin/bundle:23:in `load'
    from /usr/bin/bundle:23:in `<main>'
```

Performing `gem install etc` seems to work around the issue, replacing the failure with a benign:

```
`/root` is not writable.
Bundler will use `/tmp/bundler/home/unknown' as your home directory temporarily
```

1.16.3 seems to exhibit the same behavior.

colby-swandale:

In some cases, we are sending messages to the`Etc` library which may not be loaded.

See #6640 

Require the `etc` library before we use it

Fixes #6640

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -170,6 +170,7 @@ module Bundler
         end
 
         if warning
+          Kernel.send(:require, "etc")
           user_home = tmp_home_path(Etc.getlogin, warning)
           Bundler.ui.warn "#{warning}\nBundler will use `#{user_home}' as your home directory temporarily.\n"
           user_home


colby-swandale:
Thanks for the review @hsbt ❤️ 

@bundlerbot r+

lib/bundler.rb
```diff
@@ -170,6 +170,7 @@ def user_home
         end
 
         if warning
+          Kernel.send(:require, "etc")
```
smsohan: I'm trying to understand why we need to call `.send` here. Would it work if we just called `Kernel.require 'etc'`

deivid-rodriguez: I was confused by this too. I think `require "etc"` should work just fine (or `Kernel.require 'etc'`) as you suggest.

PR #6649, opened 2018-08-03
Add space inside block braces

AlexVPopov:
With default settings, rubocop detects this as an offense of Layout/SpaceInsideBlockBraces. It is quite common to leave a space between `{` and `|`.

The problem was that with default settings, Rubocop detects this as an offense. I think it is common to leave a space between `{` and `|`.

N/A

Add a space between `{` and `|`.

I chose this fix because I think that disabling this particular cop makes less sense.

### DIFF
--- a/lib/bundler/templates/Gemfile
+++ b/lib/bundler/templates/Gemfile
@@ -2,6 +2,6 @@
 
 source "https://rubygems.org"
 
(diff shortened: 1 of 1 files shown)


segiddins:
Bundler’s style guide does not have a space betwen `{|`

PR #6661, opened 2018-08-11
Use UTF-8 for reading files including Gemfile

### LINKED ISSUE #6660: Bundler uses the binary encoding to read Gemfile
This means any non-US-ASCII byte (>=128) in the file cannot be interpreted.

Instead, I believe UTF-8, the default source encoding for Ruby should be used, unless the user specifies a magic encoding comment like this at the top of the Gemfile:
```ruby
# encoding: iso-8859-1
```

That magic encoding comment is already handled by `eval`/`instance_eval`/etc.

https://github.com/bundler/bundler/pull/6599/files fixed this for `.gemspec` files but not Gemfile and other files read in Bundler.
I'll make a PR to fix this issue.
This will fix https://github.com/oracle/truffleruby/issues/1410.

eregon:

See #6660 and https://github.com/oracle/truffleruby/issues/1410.

The above issue details the problem.

To read the Gemfile and other files in Bundler with the default source encoding of Ruby, UTF-8, instead of the binary encoding which cannot interpret non-US-ASCII characters.

Because it's what Ruby does for other source files.

Fixes #6660.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -424,7 +424,7 @@ EOF
 
     def read_file(file)
       SharedHelpers.filesystem_access(file, :read) do
-        File.open(file, "rb", &:read)
+        File.open(file, "r:UTF-8", &:read)
       end
     end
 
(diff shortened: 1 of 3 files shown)


eregon:
@deivid-rodriguez :)
Is there anything else needed to merge this PR?

deivid-rodriguez:
@bundlerbot r+

deivid-rodriguez:
@eregon Nope! I was just coming back to the PR right now to merge it! :)

eregon:
@deivid-rodriguez Thank you!
Do you know when Bundler 1.16.4 will be released?

eregon:
@deivid-rodriguez Great to hear!

PR #6664, opened 2018-08-15
Avoid printing git error when checking version on badly packaged version

### LINKED ISSUE #6453: `bundle version` v1.16.1 prints strange "fatal: Not a git repository ...' message
I see strange 'Not a git repo' messages when running `bundle version` command. All other commands (e.g. `bundle install`) do not print this message.

Terminal session:
```
$ \bundle version
fatal: Not a git repository (or any of the parent directories): .git
Bundler version 1.16.1 (2018-03-23 commit )
$ \bundle install
Using rake 12.3.1
Using concurrent-ruby 1.0.5
Using i18n 0.9.5
... omitted
$ which bundle
/Users/laise/.rvm/gems/ruby-2.5.0@my-project/bin/bundle
```

RVM: rvm 1.29.3 (master)
Ruby: ruby 2.5.0p0 (2017-12-25 revision 61468) [x86_64-darwin17]

greysteil:

The problem was https://github.com/bundler/bundler/issues/6453.

A use had a version of Bundler packaged with Ruby 2.5.0 in RVM, but this version didn't include Bundler's git repo or any of the ivars which should have been set. As a result, Bundler was shelling out to git even though it wasn't in a git repo.

My fix is to show a nicer result in this case (just say that the SHA is unknown)

I chose this fix because ensuring Bundler is always packaged correctly by others isn't possible (although it seems weird and unlikely that a version without instance variables set would be used). Falling back to a cleaner error message is marginally nicer than what we currently have, and not much work.

Fixes #6453.

### DIFF
--- a/lib/bundler/build_metadata.rb
+++ b/lib/bundler/build_metadata.rb
@@ -23,7 +23,15 @@ module Bundler
 
     # The SHA for the git commit the bundler gem was built from.
     def self.git_commit_sha
-      @git_commit_sha ||= Dir.chdir(File.expand_path("..", __FILE__)) do
+      return @git_commit_sha if @git_commit_sha
+
+      # If Bundler has been installed without its .git directory and without a
+      # commit instance variable then we can't determine its commits SHA.
+      git_dir = File.join(File.expand_path("../../..", __FILE__), ".git")
+      return "unknown" unless File.directory?(git_dir)
+
+      # Otherwise shell out to git.
+      @git_commit_sha = Dir.chdir(File.expand_path("..", __FILE__)) do
         `git rev-parse --short HEAD`.strip.freeze
       end
     end


lib/bundler/build_metadata.rb
```diff
@@ -23,8 +23,15 @@ def self.built_at
 
     # The SHA for the git commit the bundler gem was built from.
     def self.git_commit_sha
-      @git_commit_sha ||= Dir.chdir(File.expand_path("..", __FILE__)) do
-        `git rev-parse --short HEAD`.strip.freeze
+      return @git_commit_sha if @git_commit_sha
+
+      # If Bundler has been installed without its .git directory and without a
+      # commit instance variable then we can't determine its commits SHA.
+      return "unknown" unless File.directory?(".git")
```
segiddins: this isn't quite right -- this will check whether the _current_ working directory has a `.git` subdirectory

lib/bundler/build_metadata.rb
```diff
@@ -23,8 +23,15 @@ def self.built_at
...
+
+      # Otherwise shell out to git.
+      @git_commit_sha = Dir.chdir(File.expand_path("..", __FILE__)) do
+        `git rev-parse --short HEAD`.strip.freeze if
```
segiddins: missing conditional

greysteil: That will teach me not to test again after cleaning up! Fixed.

colby-swandale:
Thank you!

@bundlerbot r+

colby-swandale:
Oops that was me.

colby-swandale:
@bundlerbot retry

PR #6668, opened 2018-08-20
Fix encoding spec from #6661 to also work 1.9.3

eregon:
* String#b is Ruby 2.0+.

### What was the problem that led to this PR?

The CI is failing on the `1-16-stable` branch.

The CI on master doesn't run against MRI 1.9.3, so this failure slipped in in #6661.
Probably testing against MRI 1.9.3 should be done on `master` too?

Make the spec work on 1.9.3.

cc @hsbt

### DIFF
--- a/spec/install/gemfile_spec.rb
+++ b/spec/install/gemfile_spec.rb
@@ -133,7 +133,7 @@ RSpec.describe "bundle install" do
       # NOTE: This works thanks to #eval interpreting the magic encoding comment
       install_gemfile <<-G
         # encoding: iso-8859-1
-        str = "Il #{"\xE9".b}tait une fois ..."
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
> Probably testing against MRI 1.9.3 should be done on master too?

Maybe... although one of the main points of "dropping" old ruby compatibility on master was build time, I think. So I tend to stick with my current opinion: drop 1.8.7 as well on master and delay compatibilty work to backporting PRs, instead of having to do it on every PR (that might not even be backported).

deivid-rodriguez:
:+1:, to the three statements made :)

eregon:
@deivid-rodriguez CI is green, could you trigger the merge? :)

deivid-rodriguez:
@bundlerbot r+

PR #6669, opened 2018-08-22
Fix DepProxy == method

ChrisBr:
After implementing a new hash table strategy in JRuby, bundle is broken for JRuby. The problem is caused that the ``==`` method in bundler does not check the class of the ``other`` object. This causes problems now when calling ``==`` on object other than DepProxy or nil.

 jruby/jruby#5280 travis-ci/travis-ci#9994

The code crashes for anything other than DepProxy class or nil.

Checking now also that other class is the same as self.

### DIFF
--- a/lib/bundler/dep_proxy.rb
+++ b/lib/bundler/dep_proxy.rb
@@ -10,7 +10,7 @@ module Bundler
     end
 
     def hash
-      @hash ||= dep.hash
+      @hash ||= [dep, __platform].hash
     end
 
     def ==(other)
--- a/lib/bundler/dep_proxy.rb
+++ b/lib/bundler/dep_proxy.rb
@@ -14,7 +14,7 @@ module Bundler
     end
 
     def ==(other)
-      return if other.nil?
+      return false if other.class != self.class
       dep == other.dep && __platform == other.__platform
     end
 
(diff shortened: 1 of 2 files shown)


colby-swandale:
Forgive the ignorance but i don't understand how this is a Bundler issue. If JRuby is incorrectly returning objects it shouldn't, patching the application code to go around that bad behavior seems backwards to me. Shouldn't the change in JRuby that is causing this be reverted or patched instead?

ChrisBr:
> Forgive the ignorance but i don't understand how this is a Bundler issue. If JRuby is incorrectly returning objects it shouldn't, patching the application code to go around that bad behavior seems backwards to me. Shouldn't the change in JRuby that is causing this be reverted or patched instead?

Hey @colby-swandale, thanks for your response. JRuby is not incorrectly returning objects, it was just hidden before because of checking hash AND object equality, so it never checked object equality when hash is different. We will change JRuby to the old behavior but I think the code in bundler is just wrong. E.g. this will crash and not return false as it is expected:

```
proxy = Bundler::DepProy.new('foobar', 'java')
proxy == "string"
```

colby-swandale:
Thanks for clearing that up, i don't know where i got the idea of an incorrect object being returned from. 😞

ChrisBr:
> Thanks for clearing that up, i don't know where i got the idea of an incorrect object being returned from.

No worries! Should I add tests for this? Altough I didn't find specs for dep_proxy so far ...

segiddins:
Yes, please do add a test for the bug this is trying to fix

ChrisBr:
@segiddins @colby-swandale added tests, please have a look again!

segiddins:
Thanks for this!

Just a note that the prior definition of `hash` was technically acceptable. `eql?` implies `hash` equality, but its OK (though undesirable) for `!eql?` objects to have the same `hash` value.

@bundlerbot r+

PR #6673, opened 2018-08-23
Add adding a team member to bundlerbot

colby-swandale:

The "adding a team member" was missing adding the person to bundlerbot


segiddins:
@bundlerbot r+

PR #6675, opened 2018-08-24
Handle RangeNotSatisfiable for compact index

MaxLap:

On ruby 2.1, bundler would try to install the latest version of psych, which is not compatible with it, instead of using an older version.

This is the error message i would get:
`Gem::RuntimeRequirementNotMetError: psych requires Ruby version >= 2.2.2. The current ruby version is 2.1.0.`

Looking online, i managed to discover that there are 2 API in use, and if bundler can't use the compact index, it will not be able to detect ruby version requirements. (It might be a good idea to have a message that explain that more clearly somewhere, I found that in a random thread somewhere)

Using `bundle install --verbose`, I found out that a RangeNotSatisfiable was being returned by the `info/psych` call and it stopped the usage of the compact index. I figured out that this was because the local cache file was too long after reading te code.

I also knew that psych did a mistake in one of their release and had to yank versions in the past.

When receiving the `RangeNotSatisfiable`, instead of generating an exception which bubbles up and forces bundler to switch to the dependency API, I return the response, and test for it in th caller. When the response is that one, I trigger a retry, which reloads the whole while by not requesting a specific range.

It seems like the only logical fix to do in the current situation. 

If the API didn't use the HTTP header stuff, it would have been more efficient to change rubygems to return the full file or just the new etag instead of a RangeNotSatisfiable. Doing so would have solved the issue for all previous versions of Bundler. But the use of caching servers and so on means that it's better to just respect HTTP and do a second query.

### DIFF
--- a/lib/bundler/fetcher/downloader.rb
+++ b/lib/bundler/fetcher/downloader.rb
@@ -27,6 +27,11 @@ module Bundler
             new_uri.password = uri.password
           end
           fetch(new_uri, headers, counter + 1)
+        when Net::HTTPRequestedRangeNotSatisfiable
+          new_headers = headers.dup
+          new_headers.delete("Range")
+          new_headers["Accept-Encoding"] = "gzip"
+          fetch(uri, new_headers)
         when Net::HTTPRequestEntityTooLarge
           raise FallbackError, response.body
         when Net::HTTPUnauthorized
--- a/lib/bundler/fetcher/downloader.rb
+++ b/lib/bundler/fetcher/downloader.rb
@@ -11,10 +11,10 @@ module Bundler
         @redirect_limit = redirect_limit
       end
 
-      def fetch(uri, options = {}, counter = 0)
+      def fetch(uri, headers = {}, counter = 0)
         raise HTTPError, "Too many redirects" if counter >= redirect_limit
 
-        response = request(uri, options)
+        response = request(uri, headers)
         Bundler.ui.debug("HTTP #{response.code} #{response.message} #{uri}")
 
         case response
@@ -26,7 +26,7 @@ module Bundler
             new_uri.user = uri.user
             new_uri.password = uri.password
           end
-          fetch(new_uri, options, counter + 1)
+          fetch(new_uri, headers, counter + 1)
         when Net::HTTPRequestEntityTooLarge
           raise FallbackError, response.body
         when Net::HTTPUnauthorized
@@ -38,11 +38,11 @@ module Bundler
         end
       end
 
-      def request(uri, options)
+      def request(uri, headers)
         validate_uri_scheme!(uri)
 
         Bundler.ui.debug "HTTP GET #{uri}"
-        req = Net::HTTP::Get.new uri.request_uri, options
+        req = Net::HTTP::Get.new uri.request_uri, headers
         if uri.user
           user = CGI.unescape(uri.user)
           password = uri.password ? CGI.unescape(uri.password) : nil
--- a/spec/install/gems/compact_index_spec.rb
+++ b/spec/install/gems/compact_index_spec.rb
@@ -821,6 +821,37 @@ The checksum of /versions does not match the checksum provided by the server! So
     expect(the_bundle).to include_gems "rack 1.0.0"
   end
 
+  it "performs full update of compact index info cache if range is not satisfiable" do
+    gemfile <<-G
+      source "#{source_uri}"
+      gem 'rack', '0.9.1'
+    G
+
+    rake_info_path = File.join(Bundler.rubygems.user_home, ".bundle", "cache", "compact_index",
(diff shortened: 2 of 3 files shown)


greysteil:
@MaxLap just to double-check, which version of Bundler are you seeing this on? I only ask because you mention Ruby 2.1, and it's possible the Bundler version being used there is old. I wrote a fix for this that was included in 1.15.2.

MaxLap:
I tried it with the most recent one, and it also happens on Travis CI (which uses 1.16.4), here is the job link: https://travis-ci.org/deep-cover/deep-cover/jobs/418477777

Are you talking about #5826? I saw the code and wondered if I should remove it since this PR is a superset fixing-wise. But it could also be considered a performance thing to avoid double query to the server in those cases.

Just to clarify, this is the info/psych that my colleague had. [bad_psych_info.txt](https://github.com/bundler/bundler/files/2318399/bad_psych_info.txt) It's about 7135 bytes long. Where as the current correct one is 6598 bytes long. So doing minus X won't do the trick here.

lib/bundler/fetcher/downloader.rb
```diff
@@ -11,22 +11,22 @@ def initialize(connection, redirect_limit)
...
+      def fetch(uri, headers = {}, counter = 0)
         raise HTTPError, "Too many redirects" if counter >= redirect_limit
 
-        response = request(uri, options)
+        response = request(uri, headers)
         Bundler.ui.debug("HTTP #{response.code} #{response.message} #{uri}")
 
         case response
-        when Net::HTTPSuccess, Net::HTTPNotModified
+        when Net::HTTPSuccess, Net::HTTPNotModified, Net::HTTPRequestedRangeNotSatisfiable
```
segiddins: this looks a little suspect to me -- why are we treating 416 responses as successful?

greysteil: Agreed.

@MaxLap - I can see what you're doing here, but I think it would be clearer to add retry logic to this method (i.e., set `headers["Accept-Encoding"] = "gzip"` and then call `fetch(..)` again in a new `when Net::HTTPRequestedRangeNotSatisfiable` case.

With that done, `lib/bundler/compact_index_client/updater.rb` should remain unchanged.

greysteil:
👍 from me, after moving the logic.

FYI, this doesn't replace #5826 - that logic is still necessary to avoid 416s in the happy case (saying "give me the last zero bytes" isn't an acceptable request, which is why we ask for the last one byte). What this does is allow us to use the CompactIndex in the unhappy case. That's a great improvement, though - thanks for the PR!

lib/bundler/fetcher/downloader.rb
MaxLap: Well, if you specify a range and that range is wrong, being told by the server that it was wrong is the normal expected answer in a way. 

But I do also agree with what you are saying, and that was my initial fix idea. I just found it clunky to have to alter the headers in a method & class that just receives them and pass them through. 

I'll do the change tonight.

MaxLap:
The requested change is done. I rebased on master in case that was what made travis fail.

greysteil:
@bundlerbot r+

greysteil:
Thanks for digging into this @MaxLap, and for the PR!

MaxLap:
@greysteil I just thought of this but, for the happy case (the files already match), that should be handled by the `If-None-Match` header and not require #5826.

Doing some testing I see If-None-Match appears to be ignored when Range is given, but reading the HTTP specs, I see that it should not be ignored ([last paragraph or Range](https://tools.ietf.org/html/rfc7233#section-3.1))... So it seems like this is a bug with Fastly or Rubygems? Any idea of who I should contact about this?

Here are some curl queries to show the problem, might have to update the If-None-Match with the ETag if it changes:

```
curl -i -H 'If-None-Match: "163daf9bf9777a94c9af1ea7214c8ea8"' https://index.rubygems.org/info/psych
#=> NotModified

curl -i -H 'If-None-Match: "163daf9bf9777a94c9af1ea7214c8ea8"' -H 'Range: bytes=6200-' https://index.rubygems.org/info/psych
#=> PartialContent (Should still be NotModified)

curl -i -H 'If-None-Match: "163daf9bf9777a94c9af1ea7214c8ea8"' -H 'Range: bytes=10000-' https://index.rubygems.org/info/psych
#=> RangeNotSatisfiable (Should still be NotModified)
```

indirect:
@MaxLap I think it’s the Fastly servers returning those responses based on the full file in the cache, so maybe you can use fastly-debug.com to open a ticket with them? I’m happy to join that ticket or file a dupe if that will help.

MaxLap:
Alright, thnx. I did a ticket there, we'll see the feedback.

MaxLap:
Got an answer, Fastly confirmed the bug. It might take some time to fix it as it's somewhere in the way their servers communicate. I'll keep you posted.

Issue #5924, opened 2017-08-04
Adding --gemfile= selection as an argument to exec


lloeki:
I was trying to fix [this](https://github.com/vim-syntastic/syntastic/issues/2051) in a way that doesn't involve gruesome hacks and matryoshka wrappers by going something along [this route](https://github.com/vim-syntastic/syntastic#faqconfig) (4.9). Here's what I found:

- setting `BUNDLE_GEMFILE=where/is/Gemfool` and running `bundle exec rubocop -c where/is/.rubocop.yml where/is/my/file.rb` works but requires setting an env var which is impractical and cannot be done without side effects in some cases (e.g multiple projects open in a single vim instance)
- `cd`'ing into `where/is/my` and running `bundle exec rubocop file.rb` works but requires changing the pwd which cannot be done without side effects in some cases (vim plugins don't like things like `autochdir`)
- `bundle install --gemfile=Gemfool` is a thing when it can use `BUNDLE_GEMFILE` already, so *“there's BUNDLE_GEMFILE, use that”* doesn't seem to carry much weight
- running `env BUNDLE_GEMFILE=where/is/Gemfool bundle exec rubocop -c where/is/.rubocop.yml where/is/my/file.rb` works but may be hard to set as a command without side effects (vim-syntastic) so why is there `--gemfile=` for `install`? Also, maybe unix/shell dependent?
- `bundle exec` can already take an optional argument (`[--keep-file-descriptors]`) so it's not like it can't take more, so not accepting one more doesn't seem to carry much weight either

So what do you think about adding `--gemfile=` to `bundle exec`, and possibly the other commands too for consistency?

pvsukale:
@colby-swandale I am a first-time contributor. Can I start working on this?

colby-swandale:
Sure! let us know if you have any questions.

blackst0ne:
Any updates on this? :slightly_smiling_face:  

@pvsukale have you started to work on this proposal?

pvsukale:
Sorry. Was busy with college work. Will start working on this tonight.
Is this okay?

blackst0ne:
@pvsukale sure. Take your time! 

I just wanted to be sure you're in business. :slightly_smiling_face:

ankitkataria:
@colby-swandale if this issue is still open, can I work on this? I'm a first-time contributor. :smile:

prithvi16:
[...]

PR #6676, opened 2018-08-24
Add gemfile flag to exec command

ankitkataria:

This PR resolves the issue #5924 

- added a `gemfile` flag in `cli.rb`
- if `options[:gemfile]` is set I created a SharedHelper method `custom_gemfile` which used the predefined `find_file` function to find and set the environment variable with the argument provided

### DIFF
--- a/lib/bundler/cli/exec.rb
+++ b/lib/bundler/cli/exec.rb
@@ -23,8 +23,6 @@ module Bundler
     def run
       validate_cmd!
       SharedHelpers.set_bundle_environment
-      SharedHelpers.custom_gemfile(@options[:gemfile]) unless @options[:gemfile].nil?
-
       if bin_path = Bundler.which(cmd)
         if !Bundler.settings[:disable_exec_load] && ruby_shebang?(bin_path)
           return kernel_load(bin_path, *args)
--- a/lib/bundler/cli/exec.rb
+++ b/lib/bundler/cli/exec.rb
@@ -23,6 +23,10 @@ module Bundler
     def run
       validate_cmd!
       SharedHelpers.set_bundle_environment
+
+      # setting the gemfile if mentioned
+      SharedHelpers.custom_gemfile(@options[:gemfile]) unless @options[:gemfile].nil?
+
       if bin_path = Bundler.which(cmd)
         if !Bundler.settings[:disable_exec_load] && ruby_shebang?(bin_path)
           return kernel_load(bin_path, *args)
--- a/lib/bundler/cli/exec.rb
+++ b/lib/bundler/cli/exec.rb
@@ -23,8 +23,6 @@ module Bundler
     def run
       validate_cmd!
       SharedHelpers.set_bundle_environment
-
-      # setting the gemfile if mentioned
       SharedHelpers.custom_gemfile(@options[:gemfile]) unless @options[:gemfile].nil?
 
       if bin_path = Bundler.which(cmd)
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -230,6 +230,11 @@ module Bundler
       filesystem_access(gemfile_path) {|g| File.open(g, "w") {|file| file.puts contents } }
     end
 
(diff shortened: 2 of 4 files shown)


blackgurl1981:
Hi

colby-swandale:
This needs a test to make sure the feature is behaving as expected and to make sure it will continue to work when new changes are made.

ankitkataria:
@colby-swandale I'll start working on the test. Will update the pr asap. :wink:

spec/commands/exec_spec.rb
```diff
@@ -6,6 +6,16 @@
...
   end
 
+  it "works with --gemfile flag", :focus do
+    create_file "CustomGemfile", <<-G
+      gem "rack"
+    G
+
+    bundle "exec --gemfile CustomGemfile test"
+
+    expect(ENV["BUNDLE_GEMFILE"]).to eq("CustomGemfile")
```
segiddins: I dont think this will test what you want it to, because the `bundle` helper actually runs bundler in a subprocess

ankitkataria:
@segiddins Thankyou so much for your comment :smile: I understood what I was doing wrong. I've updated the test now.

lib/bundler/cli/exec.rb
```diff
@@ -23,6 +23,10 @@ def initialize(options, args)
     def run
       validate_cmd!
       SharedHelpers.set_bundle_environment
+
+      # setting the gemfile if mentioned
```
colby-swandale: This comment doesn't seem all that useful, the code already explains this

colby-swandale:
Great! Thank you! 🎉 

@bundlerbot r+

ankitkataria:
@colby-swandale I don't understand why the jobs corresponding to the failing tests were canceled. What needs to be done? :thinking:

spec/commands/exec_spec.rb
```diff
@@ -6,6 +6,22 @@
     system_gems(system_gems_to_install, :path => :bundle_path)
   end
 
+  it "works with --gemfile flag" do
+    create_file "CustomGemfile", <<-G
```
colby-swandale: Thinking more about this test, do you need 2 different Gemfiles?

ankitkataria: I did that to ensure that the environment variable was set correctly both the times and `rackup` produced the version corresponding to the mentioned gemfiles.

lib/bundler/shared_helpers.rb
```diff
@@ -230,6 +230,11 @@ def write_to_gemfile(gemfile_path, contents)
       filesystem_access(gemfile_path) {|g| File.open(g, "w") {|file| file.puts contents } }
     end
 
+    def custom_gemfile(gemfile)
```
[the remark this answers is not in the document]
ankitkataria: It was mentioned in the issue that the flag could be extended to other commands as well for consistency. So I made a shared helper which could be used later. 
I used "custom" in the sense that it was mentioned by the user.  if it is not mentioned then the default gemfiles will be used.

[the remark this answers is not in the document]
ankitkataria: @colby-swandale does this explanation make sense? is there a requirement to change the name of the function? :thinking:

spec/commands/exec_spec.rb
colby-swandale: You're already testing `bundle exec` with the `--gemfile` option. So testing it again is redundant, can you please remove it.

lib/bundler/shared_helpers.rb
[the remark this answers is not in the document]
colby-swandale: You should be able to use the custom Gemfile logic in the [CLI](https://github.com/bundler/bundler/blob/master/lib/bundler/cli.rb#L36) unless i'm missing something?

colby-swandale:
Thanks! @bundlerbot r+

colby-swandale:
@budlerbot retry

colby-swandale:
@bundlerbot retry

Issue #6677, opened 2018-08-24
Bundle install hangs on a specific combination of settings


deivid-rodriguez:
Running `bundle install` hangs like this:

```console
$ time bundle install
Fetching gem metadata from https://rubygems.org/.........
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies..............^C

real	6m43,428s
user	6m42,348s
sys	0m0,304s
```

With the following environment:

## Environment

```
Bundler       1.16.4
  Platforms   ruby, x86_64-linux
Ruby          2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux]
  Full Path   /home/deivid/.rbenv/versions/2.5.1/bin/ruby
  Config Dir  /home/deivid/.rbenv/versions/2.5.1/etc
RubyGems      2.7.7
  Gem Home    /home/deivid/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0
  Gem Path    /home/deivid/.gem/ruby/2.5.0:/home/deivid/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0
  User Path   /home/deivid/.gem/ruby/2.5.0
  Bin Dir     /home/deivid/.rbenv/versions/2.5.1/bin
Tools         
  Git         2.18.0
  RVM         not installed
  rbenv       rbenv 1.1.1

  chruby      not installed
```

## Bundler Build Metadata

```
Built At          2018-08-20
Git SHA           a356d8cf9
Released Version  true
```

## Bundler settings

```
specific_platform
  Set for your local app (/home/deivid/Code/playground/test_gemfile/.bundle/config): true
lockfile_uses_separate_rubygems_sources
  Set for your local app (/home/deivid/Code/playground/test_gemfile/.bundle/config): true
disable_multisource
  Set for your local app (/home/deivid/Code/playground/test_gemfile/.bundle/config): true
```

### Gemfile

```ruby
source "https://rubygems.org"

gem "decidim", "0.13.1"
```

### Gemfile.lock

```
GEM
  remote: https://rubygems.org/
  specs:
    actioncable (5.2.0)
      actionpack (= 5.2.0)
      nio4r (~> 2.0)
      websocket-driver (>= 0.6.1)
    actionmailer (5.2.0)
      actionpack (= 5.2.0)
      actionview (= 5.2.0)
      activejob (= 5.2.0)
      mail (~> 2.5, >= 2.5.4)
      rails-dom-testing (~> 2.0)
[...]

PR #6682, opened 2018-08-30
Remove extra new lines from `bundle env` output

deivid-rodriguez:

When reporting my `bundle env` for https://github.com/bundler/bundler/issues/6677, I noticed an extra blank line between the `rbenv` and the `chruby` lines. Not really a big problem, other than people like me getting distracted by those little inconsistencies.

My diagnosis was that the formatting method was somewhere unintentionally printing an extra new line. After checking it out, it looks like any versions that are grabbed by shelling out to `#{tool} --version` will have this problem, since the new line from the subprocess output is included in the `bundle env` output, together with its own line returns.

My fix is to prune any trailing new lines from the output of those subprocesses. 

I chose this fix because it seemed like the simplest way to fix the issue.

### DIFF
--- a/spec/bundler/env_spec.rb
+++ b/spec/bundler/env_spec.rb
@@ -139,9 +139,13 @@ RSpec.describe Bundler::Env do
         expect(described_class.report).to include("Git         1.2.3 (Apple Git-BS)")
       end
     end
+  end
 
-    it "properly parses version of tools when shelling out to them" do
-      expect(described_class.send(:version_of, "ruby")).to_not include("\n")
+  describe ".version_of" do
+    let(:parsed_version) { described_class.send(:version_of, "ruby") }
+
+    it "strips version of new line characters" do
+      expect(parsed_version).to_not include("\n")
     end
   end
 end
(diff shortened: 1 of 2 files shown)


spec/bundler/env_spec.rb
```diff
@@ -139,5 +139,9 @@
         expect(described_class.report).to include("Git         1.2.3 (Apple Git-BS)")
       end
     end
+
+    it "properly parses version of tools when shelling out to them" do
+      expect(described_class.send(:version_of, "ruby")).to_not include("\n")
```
deivid-rodriguez: For the test I thought I could shell out directly to `ruby` since we know for sure it will be in the path?

colby-swandale:
Thanks! @bundlerbor r+

spec/bundler/env_spec.rb
```diff
@@ -139,5 +139,9 @@
         expect(described_class.report).to include("Git         1.2.3 (Apple Git-BS)")
       end
     end
+
+    it "properly parses version of tools when shelling out to them" do
```
colby-swandale: I would just say ` it "strips ruby version of new line characters"`

deivid-rodriguez: Absolutely, the description should be better. The thing is the test is not specific to `ruby', it's actually `rbenv` and `rvm` that are passed as parameters to the method in real life, but I used `ruby` since I'm sure it will be in PATH during the spec run. Have a look at the new version!

colby-swandale:
Thanks! @bundlerbot r+

PR #6686, opened 2018-09-06
Output OpenSSL information only when OpenSSL is available.

voxik:
It seems that only single OpenSSL availability check should be enough to output or not output the OpenSSL information.

I split this into two commits, so you can cherry-pick, in case you want to be more cautious about the availability of specific constants, but since they were introduced in Ruby 1.8.0 (https://github.com/ruby/ruby/commit/78ff3833fb67c8005a9b851037e7), I would not bother.

Please note that that this code was pointed out by Coverity scanner (although it is definitely not an error):

~~~
Error: FORWARD_NULL (CWE-476):
rubygem-bundler-1.16.1/usr/share/gems/gems/bundler-1.16.1/lib/bundler/env.rb:113: null_check: Calling "defined?(OpenSSL)" implies that "OpenSSL" might be null-like.
rubygem-bundler-1.16.1/usr/share/gems/gems/bundler-1.16.1/lib/bundler/env.rb:114: property_access: Accessing a property of null-like value "OpenSSL".
#  112|         out << ["  Bin Dir", Gem.bindir]
#  113|         out << ["OpenSSL"] if defined?(OpenSSL)
#  114|->       out << ["  Compiled", OpenSSL::OPENSSL_VERSION] if defined?(OpenSSL::OPENSSL_VERSION)
#  115|         out << ["  Loaded", OpenSSL::OPENSSL_LIBRARY_VERSION] if defined?(OpenSSL::OPENSSL_LIBRARY_VERSION)
#  116|         out << ["  Cert File", OpenSSL::X509::DEFAULT_CERT_FILE] if defined?(OpenSSL::X509::DEFAULT_CERT_FILE)
~~~

### DIFF
--- a/lib/bundler/env.rb
+++ b/lib/bundler/env.rb
@@ -110,11 +110,13 @@ module Bundler
       out << ["  Gem Path", ENV.fetch("GEM_PATH") { Gem.path.join(File::PATH_SEPARATOR) }]
       out << ["  User Path", Gem.user_dir]
       out << ["  Bin Dir", Gem.bindir]
-      out << ["OpenSSL"] if defined?(OpenSSL)
-      out << ["  Compiled", OpenSSL::OPENSSL_VERSION] if defined?(OpenSSL::OPENSSL_VERSION)
-      out << ["  Loaded", OpenSSL::OPENSSL_LIBRARY_VERSION] if defined?(OpenSSL::OPENSSL_LIBRARY_VERSION)
-      out << ["  Cert File", OpenSSL::X509::DEFAULT_CERT_FILE] if defined?(OpenSSL::X509::DEFAULT_CERT_FILE)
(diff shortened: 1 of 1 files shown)


voxik:
Hm, the 1.8.7 behavior is interesting indeed. There should be probably loaded some additional library because although OpenSSL is defined, the OpenSSL constants are not available. So either just pick the first commit or fix Bundler to load what is necessary to load, because I don't know why the versions and paths should not be reported correctly on 1.8.7.

voxik:
> The value of checking for the OpenSSL constant does not have much value.

The check was always there, it is just different now. Would it be acceptable if I dropped the second commit? Anyway feel free to reject this PR, I just thought I'll give it try.

segiddins:
Just the first commit would be perfect, thanks!

colby-swandale:
ping @voxik

voxik:
I have dropped the second commit.

Issue #6537, opened 2018-05-17
"cannot load such file" @bundler 1.16.2 with old ruby version.


LarsFronius:
Not sure if the issue is isolated to old ruby versions, but one of our customers does:
```
$ rvm use 2.0.0 --install
Using /home/rof/.rvm/gems/ruby-2.0.0-p648
YAML safe loading is not available. Please upgrade psych to a version that supports safe loading (>= 2.0).

$ gem install bundler
YAML safe loading is not available. Please upgrade psych to a version that supports safe loading (>= 2.0).
Fetching: bundler-1.16.2.gem (100%)
Successfully installed bundler-1.16.2
1 gem installed

$ bundle exec parallel_test
4 processes for 40 specs, ~ 10 specs per process
/home/rof/.rvm/gems/ruby-2.0.0-p648/bin/bundle:23:in `load': cannot load such file -- /home/rof/.rvm/rubies/ruby-2.0.0-p648/lib/ruby/gems/2.0.0/gems/bundler-1.16.2/exe/bundle (LoadError)
	from /home/rof/.rvm/gems/ruby-2.0.0-p648/bin/bundle:23:in `<main>'
	from /home/rof/.rvm/gems/ruby-2.0.0-p648/bin/ruby_executable_hooks:15:in `eval'
	from /home/rof/.rvm/gems/ruby-2.0.0-p648/bin/ruby_executable_hooks:15:in `<main>'
/home/rof/.rvm/gems/ruby-2.0.0-p648/bin/bundle:23:in `load': cannot load such file -- /home/rof/.rvm/rubies/ruby-2.0.0-p648/lib/ruby/gems/2.0.0/gems/bundler-1.16.2/exe/bundle (LoadError)
	from /home/rof/.rvm/gems/ruby-2.0.0-p648/bin/bundle:23:in `<main>'
	from /home/rof/.rvm/gems/ruby-2.0.0-p648/bin/ruby_executable_hooks:15:in `eval'
	from /home/rof/.rvm/gems/ruby-2.0.0-p648/bin/ruby_executable_hooks:15:in `<main>'
/home/rof/.rvm/gems/ruby-2.0.0-p648/bin/bundle:23:in `load': cannot load such file -- /home/rof/.rvm/rubies/ruby-2.0.0-p648/lib/ruby/gems/2.0.0/gems/bundler-1.16.2/exe/bundle (LoadError)
[...]

PR #6689, opened 2018-09-06
Fix loadError occurring in nested bundle exec calls

colby-swandale:

There is a LoadError occurring in Bundler when an application has its Gemfile installed with `--path`, and the user has a recent version of RubyGems that installs a version of Bundler as a default gem.

If the user has some code that they're running with `bundle exec` (like a test suite) that is shelling out and executing `bundle exec` again, the user will receive an error saying that Bundler could not be loaded.

    /home/travis/.rvm/gems/ruby-2.5.1/bin/bundle:23:in `load': cannot load such file -- /home/travis/.rvm/rubies/ruby-2.5.1/lib/ruby/gems/2.5.0/gems/bundler-1.16.3/exe/bundle (LoadError)

The problem that's happening is that when we run `bundle exec`, Bundler will set the Ruby environment to add the path of the current version of Bundler into the `LOAD_PATH` and also require `bundler/setup`. 

    RUBYOPT=-rbundler/setup
    RUBYLIB=/usr/local/bundle/gems/bundler-1.16.4/lib

This will have Ruby load the lib directory of the version of Bundler the user installed, but it's been loaded with the Gemspec from the default gem version of Bundler that RubyGems installed.

This gemspec is being loaded because RubyGems has a copy of the Bundler gemspec sitting in the default gems. And because we changed `GEM_HOME` when the user ran Bundler with `bundle install --path`, RubyGems just figures that seeing this is only Bundler gemspec it can find, then it should activate it but then `RUBYLIB` comes into play and just overrules RubyGems to load the newer Bundler src.

This is ultimately what's giving us the weird load path that doesn't exist.

mind blown

See #6537 and #6629

When we set the `BUNDLE_BIN_PATH` env var inside bundler, check that the file exists. If it doesn't, we will just set it to the bundle executable file.

This seems to be the solution that patches the LoadError issue without having to heavily refactor code or make the existing code more complicated.

### DIFF
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -300,7 +300,12 @@ module Bundler
 
     def set_bundle_variables
       begin
-        Bundler::SharedHelpers.set_env "BUNDLE_BIN_PATH", Bundler.rubygems.bin_path("bundler", "bundle", VERSION)
+        exe_file = Bundler.rubygems.bin_path("bundler", "bundle", VERSION)
+        unless File.exist?(exe_file)
+          exe_file = File.expand_path("../../../exe/bundle", __FILE__)
+        end
+
+        Bundler::SharedHelpers.set_env "BUNDLE_BIN_PATH", exe_file
       rescue Gem::GemNotFoundException
         Bundler::SharedHelpers.set_env "BUNDLE_BIN_PATH", File.expand_path("../../../exe/bundle", __FILE__)
       end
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
Just so it's not forgotten on a slack comment. I _think_ the spec added here passes without the fix?

colby-swandale:
I have amended the spec to correctly test this behavior.

colby-swandale:
@bundlerbot r+

Issue #6163, opened 2017-11-09
Missing or uninitialized constant: Bundler::Deprecate when installing a bundle with Rubinius


philr:
Running `bundle install` with Bundler version 1.16.0 on Rubinius can fail with a 'Missing or uninitialized constant: Bundler::Deprecate' error.

The error only occurs when the Gemfile references gems that are not already installed (or when forcing gems to be installed with `bundle install --force`).

I encounter the error with clean installs of Rubinius versions 3.82 and 3.86 using both RubyGems versions 2.6.14 and 2.7.2.

Running `bundle install --force` with Bundler version 1.15.4 works correctly on Rubinius 3.82 and 3.86.

This issue appears to have been caused by commit f9599c7e8c6775bb4044ad2b4809a957ef8fadea.

## Backtrace

```
NameError: Missing or uninitialized constant: Bundler::Deprecate
  core/module.rb:548:in `const_missing'
  /home/philr/.rbenv/versions/rbx-3.86/gems/gems/bundler-1.16.0/lib/bundler/deprecate.rb:22:in `__script__'
  /home/philr/.rbenv/versions/rbx-3.86/gems/gems/bundler-1.16.0/lib/bundler/deprecate.rb:10:in `__script__'
  core/code_loader.rb:233:in `require'
  core/kernel.rb:851:in `gem_original_require (require)'
  /home/philr/.rbenv/versions/rbx-3.86/site/rubygems/core_ext/kernel_require.rb:55:in `require'
  /home/philr/.rbenv/versions/rbx-3.86/gems/gems/bundler-1.16.0/lib/bundler/psyched_yaml.rb:30:in `__script__'
  core/code_loader.rb:233:in `require'
  core/kernel.rb:851:in `gem_original_require (require)'
  /home/philr/.rbenv/versions/rbx-3.86/site/rubygems/core_ext/kernel_require.rb:55:in `require'
  /home/philr/.rbenv/versions/rbx-3.86/gems/gems/bundler-1.16.0/lib/bundler/rubygems_integration.rb:115:in `configuration'
  /home/philr/.rbenv/versions/rbx-3.86/gems/gems/bundler-1.16.0/lib/bundler/fetcher.rb:242:in `connection'
  /home/philr/.rbenv/versions/rbx-3.86/gems/gems/bundler-1.16.0/lib/bundler/fetcher.rb:85:in `initialize'
[...]

PR #6692, opened 2018-09-09
Check that Bundler::Deprecate is not an autoload constant

eregon:
* Otherwise, it should be defined by this file.
* The issue is bundler/deprecate.rb checks `defined? Bundler::Deprecate`
  and this would normally return true since an autoload is defined for
  that constant. But since the autoload file is #require-d explicitly
  (by bundler/psyched_yaml and bundler/shared_helpers), MRI makes it
  undefined to save this pattern. This however requires a complex
  autoload implementation and is confusing to debug, so let's simplify.
* Related to https://github.com/bundler/bundler/issues/6163 and https://github.com/rubinius/rubinius/issues/3769.

Bundler fails in TruffleRuby without this change.
We plan to fix this case in TruffleRuby, but since at least 2 Ruby implementations did not expect such a corner case, we should make the code simpler since it's easy enough in this case.

See above.

It's simple and does not break if an extra `require "bundler/deprecate"` is later added in the codebase.

### DIFF
--- a/lib/bundler/deprecate.rb
+++ b/lib/bundler/deprecate.rb
@@ -8,7 +8,8 @@ rescue LoadError
 end
 
 module Bundler
-  if defined? Bundler::Deprecate
+  # If Bundler::Deprecate is an autoload constant, we need to define it
+  if defined?(Bundler::Deprecate) && !autoload?(:Deprecate)
     # nothing to do!
   elsif defined? ::Deprecate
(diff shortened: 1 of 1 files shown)


segiddins:
Cool, thanks!

@bundlerbot r+

eregon:
Could this also be backported to 1.16.x?

PR #6693, opened 2018-09-09
Add support for TruffleRuby

eregon:
Hello,
This PR adds support for TruffleRuby in Bundler.

I searched for `rbx` and `jruby` (case-insensitive) through the codebase to find all places which need changes and where tests use specific Ruby implementations. So hopefully this PR is complete, but please tell me if I missed something.

I'd also like to test TruffleRuby in Bundler's CI, but will do this as a separate PR, as we first need the next release of TruffleRuby so that it includes the fix for https://github.com/oracle/truffleruby/issues/1413.

### DIFF
--- a/spec/support/platforms.rb
+++ b/spec/support/platforms.rb
@@ -79,7 +79,7 @@ module Spec
       when "jruby"
         JRUBY_VERSION
       else
-        raise BundlerError, "That RUBY_ENGINE is not recognized"
+        RUBY_ENGINE_VERSION
       end
     end
 
--- a/lib/bundler/cli/exec.rb
+++ b/lib/bundler/cli/exec.rb
@@ -89,6 +89,7 @@ module Bundler
       possibilities = [
         "#!/usr/bin/env ruby\n",
         "#!/usr/bin/env jruby\n",
+        "#!/usr/bin/env truffleruby\n",
         "#!#{Gem.ruby}\n",
       ]
 
--- a/lib/bundler/current_ruby.rb
+++ b/lib/bundler/current_ruby.rb
@@ -32,11 +32,13 @@ module Bundler
       mswin64
       rbx
       ruby
+      truffleruby
       x64_mingw
     ].freeze
 
     def ruby?
-      !mswin? && (!defined?(RUBY_ENGINE) || RUBY_ENGINE == "ruby" || RUBY_ENGINE == "rbx" || RUBY_ENGINE == "maglev")
+      !mswin? && (!defined?(RUBY_ENGINE) || RUBY_ENGINE == "ruby" ||
+          RUBY_ENGINE == "rbx" || RUBY_ENGINE == "maglev" || RUBY_ENGINE == "truffleruby")
     end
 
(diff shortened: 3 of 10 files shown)


eregon:
This should also be backported to 1.16.x.

spec/support/platforms.rb
```diff
@@ -79,7 +79,7 @@ def local_engine_version
       when "jruby"
         JRUBY_VERSION
       else
-        raise BundlerError, "That RUBY_ENGINE is not recognized"
```
colby-swandale: Woo! This is awesome! 🎉 

My first question though why are you modifying this behavior instead of adding to it?

segiddins: we might also want to check that `RUBY_ENGINE_VERSION` is indeed defined? I'm not sure if all Rubies define it

eregon: My thinking was rather than adding a special case for `truffleruby` I'd make it general.
Every not ancient (=1.8) Ruby version should define `RUBY_ENGINE_VERSION`, it's a standard constant like `RUBY_ENGINE`, `RUBY_VERSION`, etc.
I can add a defined? it it causes a problem for an implementation but it seems unlikely as currently this code path was hardcoded to only support MRI, JRuby and Rubinius.

eregon: @colby-swandale What do you think? Is this code fine as-is or should I change it something else?

colby-swandale: I'm happy with the PR, so let's merge it

colby-swandale:
@bundlerbot r+

deivid-rodriguez:
@eregon This is great, thanks!

[This PR](https://github.com/rubygems/rubygems/pull/2344) I opened to rubygems a while ago is _somehow_ related. Maybe you want to have a look since you have experience accross several implementations and the constants they define.

eregon:
@bundlerbot timed out, should it be retried or is there an issue with the PR?

colby-swandale:
@bundlerbot retry

eregon:
Thanks for review and merging!

hsbt:
> This should also be backported to 1.16.x.

This pull-request is not bugfix. I think we need to bump 1.17.x maybe. 

Can we merge non-bugfix feature with tiny versionup?

colby-swandale:
@hsbt We have allowed non-bugfixes to go into minor releases before if it made sense. And this isn't a huge change - so i'm happy to backport this into the next patch release.

eregon:
1.16.x or 1.17.x is fine, what TruffleRuby needs is a (non-pre) release of Bundler so the patches in TruffleRuby can be removed and TruffleRuby can use vanilla Bundler.

Is there a schedule for the next Bundler release?
The next release of TruffleRuby has feature freeze on September 21.
That might be quite close, but it would be ideal so there is no overlap between patches in TruffleRuby and support for TruffleRuby in Bundler.

colby-swandale:
There is no scheduled time for a release, but i will try and get your changes released before the 21st.

PR #6694, opened 2018-09-09
Fix local spec failure

deivid-rodriguez:

The problem was that bundler specs won't pass on my system. One of the two failures I was getting is the spec added by #6502.

My diagnosis was that my system has a similar setup as TravisCI (Linux with a ruby version manager on top), but I don't set the `TRAVIS` environment variable to run the tests. This means some trickery used in the specs to get them green on TravisCI is not applied to my local run, so specs failed.

My fix is to make this spec green independently from whether it's run on TravisCI or not. My changes also make this spec not delete any files from the rubygems installation where the specs are running. In https://github.com/bundler/bundler/pull/6502#issuecomment-403713916 @indirect said that the files being deleted here are put in there by `RVM`. However, I don't think that's accurate, it's `gem update --system` putting these files in there. So if these files should not be there, we should not put them there in the first place instead of deleting them during bundler's specs.

I chose this fix because the alternative would be to set the `TRAVIS` environment variable before running the specs locally, but that would mean that `bundler`'s specs would delete files from my global `rubygems` installation, which seems... no good.

### DIFF
--- a/spec/runtime/setup_spec.rb
+++ b/spec/runtime/setup_spec.rb
@@ -884,7 +884,11 @@ end
     it "should successfully require 'bundler/setup'" do
       install_gemfile ""
 
-      ruby <<-R, :env => { "GEM_PATH" => symlinked_gem_home }, :no_lib => true
+      ruby <<-'R', :env => { "GEM_PATH" => symlinked_gem_home }, :no_lib => true
+        # Remove any bundler that's not the current bundler from $LOAD_PATH
+        $LOAD_PATH.each do |path|
+          $LOAD_PATH.delete(path) if File.exist?("#{path}/bundler.rb")
+        end
         puts (require 'bundler/setup')
       R
 
--- a/spec/runtime/setup_spec.rb
+++ b/spec/runtime/setup_spec.rb
@@ -886,14 +886,7 @@ end
 
       ENV["GEM_PATH"] = symlinked_gem_home.path
 
-      ruby <<-R
-        if $LOAD_PATH.include?("#{bundler_lib}")
-          # We should use bundler from GEM_PATH for this test, so we should
-          # remove path to the bundler source tree
-          $LOAD_PATH.delete("#{bundler_lib}")
-        else
-          raise "We don't have #{bundler_lib} in $LOAD_PATH"
-        end
+      ruby <<-R, :no_lib => true
         puts (require 'bundler/setup')
       R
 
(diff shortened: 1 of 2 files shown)


bin/with_rubygems
```diff
@@ -3,28 +3,30 @@
...
+  def run(*cmd)
+    return if system(*cmd)
+    raise "Running `#{cmd.join(" ")}` failed"
+  end
+
+  module_function :run
 end
 
-version = ENV.delete("RGV")
+version = ENV.delete("RGV") || "2.7.7"
```
segiddins: I'm not sure I like this change -- it means I need to have the local tmp checkout of rubygems present to be able to use any of the development binstubs

deivid-rodriguez: Yeah, it means that. Why is it a problem though? The checkout happens automatically so as a user of the binstubs nothing should change (other than a slightly slower first usage?). As I say in the commit message, I like it because it makes the specs more predictable and less dependent on the host system, and more similar to what's done in the CI.

Anyways, the reason I added this is that my alternative fix won't work without it, because when removing the entries from `LOAD_PATH` that include `bundler`, I'm also removing the entries that include `rubygems`. So unless `rubygems` is included in the `LOAD_PATH` separately, it doesn't work.

If you still don't want this, I can remove the change and make sure I always set `RGV` before running the specs, just let me know :+1:.

segiddins: It requires an internet connection to run the specs

deivid-rodriguez: Yeah, the first time you run the specs. But the first time you run the specs also requires an internet connection in order to download bundler's repo.

Anyways, I'll remove the change!

deivid-rodriguez: @segiddins I removed the commit! At this point I can also remove https://github.com/bundler/bundler/pull/6694/commits/e3395f44bceb2465fea882669f687bf678737377, since it's no longer necessary. Although it might be argued that it's a good change anyways, so I didn't (yet) remove it.

deivid-rodriguez: By the way, I must've messed up something while running the specs, the commit author of the last commit is [lolwut](https://github.com/bundler/bundler/blob/368d7594adbdf83032719011d7608545faf2d942/spec/support/builders.rb#L655)! :fearful:

deivid-rodriguez: I removed the mentioned commit and ammended the lolwut author! I got now a green local build :smile: :tada:

indirect:
@bundlerbot r+

deivid-rodriguez:
@bundlerbot r+

indirect:
hmm, I guess bundlerbot is having a bad day :/

@bundlerbot ping

indirect:
okay, tried rebuilding bundlerbot, let's see if the permissions are worked out...

@bundlerbot ping

indirect:
hooray!

@bundlerbot r=deivid-rodriguez

PR #6695, opened 2018-09-10
[GemVersionPromoter] Prefer non-pre-release versions

### LINKED ISSUE #6684: `bundle update --patch` unexpectedly updates to pre-release version (only 1.16.4)
#### Steps to reproduce

```ruby
# Gemfile
source "https://rubygems.org"
gem "mail"

# Gemfile.lock
GEM
  remote: https://rubygems.org/
  specs:
    mail (2.7.0)
      mini_mime (>= 0.1.1)
    mini_mime (1.0.1)

PLATFORMS
  ruby

DEPENDENCIES
  mail

BUNDLED WITH
   1.16.4
```

```
$ bundle update         # does nothing, which is fine
$ bundle update mail    # does nothing, which is fine
$ bundle update --patch # updates mail to 2.7.1.rc-1 which is unexpected
```

```diff
diff --git a/Gemfile.lock b/Gemfile.lock
index 799a80f..522550c 100644
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 GEM
   remote: https://rubygems.org/
   specs:
-    mail (2.7.0)
+    mail (2.7.1.rc1)
       mini_mime (>= 0.1.1)
     mini_mime (1.0.1)
```

segiddins:

The problem was `bundle update --patch` could cause a prerelease to be resolved, when it shouldn't have been.

Closes #6684.

My diagnosis was the resolver moved pre's to be first when no pre requirement was specified, but then the GVP did its own sorting.

My fix was to re-implement that "pre's go first unless they were explicitly requested" logic in the GVP.

I chose this fix because it allows the GVP to properly sort versions, making it entirely correct without dependence upon the resolver.

### DIFF
--- a/lib/bundler/gem_version_promoter.rb
+++ b/lib/bundler/gem_version_promoter.rb
@@ -24,6 +24,8 @@ module Bundler
     # existing in the referenced source.
     attr_accessor :strict
 
+    attr_accessor :prerelease_specified
+
     # Given a list of locked_specs and a list of gems to unlock creates a
     # GemVersionPromoter instance.
     #
@@ -39,6 +41,7 @@ module Bundler
       @locked_specs = locked_specs
       @unlock_gems = unlock_gems
       @sort_versions = {}
+      @prerelease_specified = {}
     end
 
     # @param value [Symbol] One of three Symbols: :major, :minor or :patch.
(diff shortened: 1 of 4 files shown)


greysteil:
Test failures are because [this line](https://github.com/bundler/bundler/pull/6695/files#diff-bd75110972eace20eb4674a8eb746bccR122) stops pre-release sorting from happening for locked specs. I don't understand the GVP logic well enough to immediately know what to change there.

One easy way to fix would be to remove the `unless @prerelease_specified[dependency.name] || @use_gvp` change, but I'm guessing you added that to avoid doing the pre-release sorting work twice...

colby-swandale:
Thanks! @bundlerbot r+

PR #6685, opened 2018-09-03
Remove Travis condition making test fail locally

deivid-rodriguez:
### What was the end-user problem that led to this PR?

I tried running the specs locally and I got two failures.

### What was your diagnosis of the problem?
 My diagnosis was that the tests were artificially only doing the right thing on Travis.

### What is your fix for the problem, implemented in this PR?

My fix was to remove the condition to check whether we're in Travis or not, so that the correct assertion is made.

### Why did you choose this fix out of the possible options?

I chose this fix because it works for me and it simplifies the test.

PR #6696, opened 2018-09-12
Better conditional assertion

deivid-rodriguez:
This is another try at #6685.

The problem was that I get a failure when running the specs locally.

My diagnosis was that the result of this specs depends, not on whether we are on Travis or not, but on which platform the spec runs. 

My fix is to check for the platform so that each platform gets its expected result, regardless of whether the spec is run on Travis CI or not.

I chose this fix because hopefully it doesn't break local specs for @colby-swandale, and it fixes them for me :)

### DIFF
--- a/spec/commands/exec_spec.rb
+++ b/spec/commands/exec_spec.rb
@@ -561,7 +561,7 @@ RSpec.describe "bundle exec" do
         ex << "raise SignalException, 'SIGTERM'\n"
         ex
       end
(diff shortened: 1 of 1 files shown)


colby-swandale:
This is a lot better! Great change! @bundlerbot r+

deivid-rodriguez:
Thanks @colby-swandale! By the way, did you try my suggestion in https://github.com/bundler/bundler/pull/6685#issuecomment-419742732? No big deal, I was just curious and I don't have an OSX machine to satisfy my curiosity :smile:

PR #6697, opened 2018-09-12
[CLI::GEM] bundle gem will generate homepage_uri and code and changelog section …

walf443:

The problem was I always have to remember how to add CHANGELOG.md  when I set up a new gem.

My diagnosis was that CHANGELOG.md has become enough of a community standard that it makes sense to offer it in bundle gem.

My fix is only added metadata[:changelog_uri] to generated gemspec file.
Changelog.md file is not generated by default.

I chose this fix because Changelog.md file format is different by projects.

### DIFF
--- a/lib/bundler/templates/newgem/newgem.gemspec.tt
+++ b/lib/bundler/templates/newgem/newgem.gemspec.tt
@@ -25,6 +25,10 @@ Gem::Specification.new do |spec|
   # to allow pushing to a single host or delete this section to allow pushing to any host.
   if spec.respond_to?(:metadata)
     spec.metadata["allowed_push_host"] = "TODO: Set to 'http://mygemserver.com'"
+
+    spec.metadata["homepage_uri"] = spec.homepage
+    spec.metadata["source_code_uri"] = "TODO: Put your gem's public repo URL here."
+    spec.metadata["changelog_uri"] = "TODO: Put your gem's CHANGELOG.md URL here."
   else
     raise "RubyGems 2.0 or newer is required to protect against " \
       "public gem pushes."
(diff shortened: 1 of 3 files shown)


walf443:
rubygems link in changelog_uri was introduced by https://github.com/rubygems/rubygems.org/pull/1553/files

walf443:
oh. I'll try to fix test.

hsbt:
@walf443 Thanks. I will merge this after test passes.

walf443:
@hsbt I'm sorry to be late to fix test.

hsbt:
@bundlerbot r+

kbrock:
@walf443 Thank you so much for putting together this PR.

/cc @olivierlacan not sure if you saw this

olivierlacan:
@kbrock I did in a new gem I built. It's awesome! Although I was a bit confused that the block appeared wrapped in the RubyGems.org push prevention conditional at first because of the comment location. Seems like that might confuse some folks into deleting the whole conditional block: 

```ruby
  # Prevent pushing this gem to RubyGems.org. To allow pushes either set the 'allowed_push_host'
  # to allow pushing to a single host or delete this section to allow pushing to any host.
  if spec.respond_to?(:metadata)
    spec.metadata["allowed_push_host"] = "TODO: Set to 'http://mygemserver.com'"

    spec.metadata["homepage_uri"] = spec.homepage
    spec.metadata["source_code_uri"] = "TODO: Put your gem's public repo URL here."
    spec.metadata["changelog_uri"] = "TODO: Put your gem's CHANGELOG.md URL here."
  else
    raise "RubyGems 2.0 or newer is required to protect against " \
      "public gem pushes."
  end
```

I think I might submit a PR to address that if it makes sense.

PR #6701, opened 2018-09-18
Fix incorrect year in changelog

dentarg:


greysteil:
Thanks! @bundlerbot r+

PR #6702, opened 2018-09-18
Better `--force` to `--redownload` transition

deivid-rodriguez:

The problem was. that, while trying to run `bundle install --force` today, I was getting the error `unknown switch --force` error, and I was not sure what was causing it and what I needed to do to fix it. Also, running `bundle install --help` would show `--force` as a supported option, so that would make things even more confusing.

My diagnosis was that:

* A lot of removal/renaming of options in bundler 2 are artificially linked to the `forget_cli_options` switch. To me, those changes even though they are related (stopping rememebering cli options prompted us to actually remove some unnecessary ones), they are independent, so it's not straighforward to find out that it's the `forget_cli_options` setting causing this rename to be in place.

* We should show a helpful deprecation message instead of a hard error, so the transition is easier for users. 

* We should update the commands help to document newer instead of deprecated options.

* The `bundle update` command has an equivalent `--force` option that should be renamed too for sanity.

My fix to the above problems is to:

* Always keep the `--force` alias, independently of whether the `forget_cli_options` is set or not.
* Show a deprecation message when `--force` is used in bundler 2, but still keep it working as before.
* Update the commands help to use `--redownload`.
* Make the equivalent changes to the `bundle update`

I chose this fix because it seems like the more user friendly way going forward in my opinion.


indirect:
hooray! this is perfect, and exactly the kind of thing I am hoping to accomplish with the 2.0 transition. 😄 I am extremely on board with the goals here, and I think this is a great change. thank you! 🙌 

@bundlerbot r+

deivid-rodriguez:
@bundlerbot r=indirect

colby-swandale:
Note that for 1.17.0, the documentation for `--force` cannot be removed. Either both options needs to be specified in the docs or just use `--force` for the time being until the next major release.

deivid-rodriguez:
You're totally right. I think we should document `--redownload`, and say `--force` is an alias?

deivid-rodriguez:
@colby-swandale Thoughts about doing something like [this](https://github.com/bundler/bundler/compare/still_mention_force), and removing the "deprecated" word when backporting it?

colby-swandale:
I like the idea but i would prefer saying that `--redownload` is the alias for 1.X releases.

deivid-rodriguez:
Sounds good to me, let me cook a PR!

deivid-rodriguez:
Actually, after another thought I don't think we should document deprecated options, so it's better if I target the PR directly to `1.17-stable`. Sounds good?

colby-swandale:
sounds great!

deivid-rodriguez:
Here you go! :point_right:  #6734

PR #6706, opened 2018-09-23
Clean up some specs

deivid-rodriguez:

The problem was that some specs could be more readable and clean.

My diagnosis was that some stuff could be more explicit, some TODO's could be removed because they're already fixed, and there should not be typos.

My fix is to make the obvious changes to fix what I diagnosed.

I chose this fix because it makes the specs clear, in my opinion.

### DIFF
--- a/spec/runtime/setup_spec.rb
+++ b/spec/runtime/setup_spec.rb
@@ -1323,7 +1323,6 @@ end
 
       it "activates no gems with bundle exec" do
         install_gemfile! ""
-        # ensure we clean out the default gems, bceause bundler's allowed to be activated
         create_file("script.rb", code)
         bundle! "exec ruby ./script.rb", :env => { :RUBYOPT => activation_warning_hack_rubyopt + " -rbundler/setup" }
         expect(last_command.stdout).to eq("{}")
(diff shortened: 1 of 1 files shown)


spec/runtime/setup_spec.rb
```diff
@@ -1318,22 +1317,18 @@ def lock_with(ruby_version = nil)
...
+        ruby! code, :env => { :RUBYOPT => activation_warning_hack_rubyopt + " -rbundler/setup" }
         expect(last_command.stdout).to eq("{}")
       end
 
       it "activates no gems with bundle exec" do
         install_gemfile! ""
-        # ensure we clean out the default gems, bceause bundler's allowed to be activated
         create_file("script.rb", code)
-        bundle! "exec ruby ./script.rb", :env => { :RUBYOPT => activation_warning_hack_rubyopt + " -rbundler/setup" }
+        bundle! "exec ruby ./script.rb", :env => { :RUBYOPT => activation_warning_hack_rubyopt }
```
colby-swandale: Why does this line no longer need `" -rbundler/setup"`?

deivid-rodriguez: Because this spec is checking activation via `bundle exec` not via `-rbundler/setup` (as opposed to the previous spec, that uses `ruby` directly but passes `-rbundler/setup`).

deivid-rodriguez: It's not that it no longer need it, it never actually needed it, as far as I can see.

colby-swandale: Thanks for the explanation!

colby-swandale:
@bundlerbot r+

PR #6707, opened 2018-09-24
Run more assertions in more cases 

deivid-rodriguez:

I noticed a couple of places where assertions were being excluded and they shouldn't:

* One was introduced by me in #6702, where the specs added (and some already present) started being tested only on bundler 2.x.
* The other one was introduced in f7414bcb17fe1bd67246021251b5f0527bd6afd1, where one assertion would be run only if a certain env variable was not set. I think it was because of a TravisCI environmental issue that now seems fixed.

My diagnosis was that none of these exclusions are necessary.

My fix is to restore the excluded assertions to all environments and branches.

I chose this fix because it seems best to avoid future problems.

### DIFF
--- a/spec/commands/exec_spec.rb
+++ b/spec/commands/exec_spec.rb
@@ -843,10 +843,7 @@ __FILE__: #{path.to_s.inspect}
           expect(bundle!("exec #{file}", :artifice => nil)).to eq(expected)
           expect(bundle!("exec bundle exec #{file}", :artifice => nil)).to eq(expected)
           expect(bundle!("exec ruby #{file}", :artifice => nil)).to eq(expected)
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
I noticed the issue with #6702 when I was going to milestone it for 1.17.0. The changes here will only actually be exercised when this PR and #6702 are backported. I already did that locally and the specs pass.

colby-swandale:
why is this in the 1.17.0 milestone?

deivid-rodriguez:
Because otherwise the specs I'm adding here for `bundler < 2` will never be run, and I'd like to make sure I'm not adding any deprecation messages here (yet).

colby-swandale:
thanks for the exaplanation

PR #6708, opened 2018-09-25
Fix only_update_to_newer_versions regression

### LINKED ISSUE #6529: only_update_to_newer_versions regression in v1.16
Hi,

I've noticed a few edge cases where the `only_update_to_newer_versions` setting doesn't get respected anymore, where it used to work in Bundler 1.15.

Here is an example to reproduce: https://gist.github.com/theflow/edd84a5318738b59f65cf42eaaede9f3

Bundler v1.16 will downgrade carrierwave-mongoid to an really old version which is missing the version requirement on mongoid:

```
+ BUNDLE_ONLY_UPDATE_TO_NEWER_VERSIONS=true
+ bundle update mongoid carrierwave-mongoid
Resolving dependencies....
Using concurrent-ruby 1.0.5
Using i18n 1.0.1
Using minitest 5.11.3
Using thread_safe 0.3.6
Using tzinfo 1.2.5
Using activesupport 5.2.0
Using activemodel 5.2.0
Using bson 4.3.0
Using bundler 1.16.1
Using mime-types-data 3.2016.0521
Using mime-types 3.1
Using carrierwave 1.2.2
Using mongo 2.5.3
Using mongoid 7.0.1 (was 6.4.1)
Using carrierwave-mongoid 0.1.0 (was 1.0.0)
Bundler attempted to update carrierwave-mongoid but its version regressed from 1.0.0 to 0.1.0
Bundle updated!
```

I'm pretty sure it was introduced in https://github.com/bundler/bundler/commit/47256d20cb05ebc724ee67173094682153b6b4aa, but I don't understand the reason behind that change.

Here https://github.com/bundler/bundler/blob/master/lib/bundler/definition.rb#L975 it's comparing a `Bundler::Dependency` with an `Array` from the group_by which will always be non-equal, which I guess is not the intention?

Comparing the first element of the array fixes this issue for me, but not sure if that's the correct way and what's supposed to happen if there are multiple `dependencies_by_name`.

https://github.com/theflow/bundler/commit/07942b1a6ccec4faff8716afc89288edad8496d2

Cheers,
Florian

theflow:
This is my attempt to fix #6529

Running `bundle update` with `BUNDLE_ONLY_UPDATE_TO_NEWER_VERSIONS: "true"` resulted in a gem getting downgraded to a really old version in a certain edge case. Ironically it wouldn't get downgraded when `BUNDLE_ONLY_UPDATE_TO_NEWER_VERSIONS` was set to false.

My diagnosis was that https://github.com/bundler/bundler/commit/47256d20cb05ebc724ee67173094682153b6b4aa tried to solve the problem of still allowing manual downgrades in the Gemfile while `only_update_to_newer_versions` is true. But introduced a regression that prevented the `additional_base_requirements_for_resolve` method to work as intended:

This is the relevant change from that commit that tries to avoid adding the `>=` requirement if the  requirement in the Gemfile is different than the requirement in the lockfile (as far as I understand it):

```ruby
next requirements if @locked_deps[name] != dependencies_by_name[name]
```

I identified two problems
 1. `dependencies_by_name[name]` returns an array of `Bundler::Dependency`, where as 
`@locked_deps[name]` just returns a single `Bundler::Dependency`. Comparing the two will always be false.
 1. `@locked_deps` is always empty in case of `bundle update`. See: https://github.com/bundler/bundler/blob/3d9e6167a7df9ca89a030dfe95c7cdff293e74a9/lib/bundler/definition.rb#L95

My fixes:

 1. Make sure `dependencies_by_name` is a hash with `Bundler::Dependency` as values
 1. Fetch the `@locked_gems.dependencies` again instead of using `@locked_deps`
 1. The existing test worked for me with and without the `only_update_to_newer_versions` set to true, I replaced it with a reproduction of the edge case I was investigating (this is as minimal as I could make it)
 1. I've added a test for the manual downgrading case.

This is the only way I could make these cases work. It's possible there are other edge cases I don't understand.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -974,10 +974,10 @@ module Bundler
 
     def additional_base_requirements_for_resolve
       return [] unless @locked_gems && Bundler.feature_flag.only_update_to_newer_versions?
-      dependencies_by_name = dependencies.group_by(&:name)
+      dependencies_by_name = dependencies.inject({}) {|memo, dep| memo.update(dep.name => dep) }
       @locked_gems.specs.reduce({}) do |requirements, locked_spec|
         name = locked_spec.name
-        next requirements if @locked_deps[name] != dependencies_by_name[name]
+        next requirements if @locked_gems.dependencies[name] != dependencies_by_name[name]
         dep = Gem::Dependency.new(name, ">= #{locked_spec.version}")
         requirements[name] = DepProxy.new(dep, locked_spec.platform)
         requirements
--- a/spec/commands/update_spec.rb
+++ b/spec/commands/update_spec.rb
@@ -134,33 +134,67 @@ RSpec.describe "bundle update" do
       before do
         bundle! "config only_update_to_newer_versions true"
       end
+
       it "does not go to an older version" do
         build_repo4 do
-          build_gem "a" do |s|
-            s.add_dependency "b"
-            s.add_dependency "c"
+          build_gem "tilt", "2.0.8"
+          build_gem "slim", "3.0.9" do |s|
+            s.add_dependency "tilt", [">= 1.3.3", "< 2.1"]
           end
-          build_gem "b"
-          build_gem "c"
-          build_gem "c", "2.0"
+          build_gem "slim_lint", "0.16.1" do |s|
+            s.add_dependency "slim", [">= 3.0", "< 5.0"]
+          end
+          build_gem "slim-rails", "0.2.1" do |s|
+            s.add_dependency "slim", ">= 0.9.2"
+          end
+          build_gem "slim-rails", "3.1.3" do |s|
+            s.add_dependency "slim", "~> 3.0"
+          end
         end
 
         install_gemfile! <<-G
           source "file:#{gem_repo4}"
-          gem "a"
+          gem "slim-rails"
+          gem "slim_lint"
         G
 
-        expect(the_bundle).to include_gems("a 1.0", "b 1.0", "c 2.0")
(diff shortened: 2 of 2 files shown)


lib/bundler/definition.rb
```diff
@@ -974,10 +974,11 @@ def compute_requires
 
     def additional_base_requirements_for_resolve
       return [] unless @locked_gems && Bundler.feature_flag.only_update_to_newer_versions?
-      dependencies_by_name = dependencies.group_by(&:name)
+      dependencies_by_name = dependencies.each_with_object({}) {|dep, memo| memo[dep.name] = dep }
+      all_locked_deps = @locked_gems.dependencies
```
greysteil: I'd rather lose this assignment and use the `@locked_gems.dependencies` directly in the loop

lib/bundler/definition.rb
```diff
@@ -974,10 +974,11 @@ def compute_requires
 
     def additional_base_requirements_for_resolve
       return [] unless @locked_gems && Bundler.feature_flag.only_update_to_newer_versions?
-      dependencies_by_name = dependencies.group_by(&:name)
+      dependencies_by_name = dependencies.each_with_object({}) {|dep, memo| memo[dep.name] = dep }
```
greysteil: We don't have `each_with_object` in Ruby 1.8.7, so this will need to use `reduce`

theflow:
I've incorporated your comments

greysteil:
@bundlerbot r+

greysteil:
Thanks for this @theflow! 🎉

PR #4272, opened 2016-02-04
Skip building git gem extensions if they're already built

indirect:

PR #6711, opened 2018-09-27
Avoid fetching an rebuilding git gems when it's not necessary

casperisfine:
Kind of a followup to https://github.com/bundler/bundler/pull/4272

### The issue

As demonstrated by the updated test case, whenever a gem is changed (even a non-git one), bundler re-fetch all git gems, and recompile their extensions (if any).

You can repro that issue by running the modified `git_spec.rb` against master.

### The proposed fix

In that patch I simply make `Source::Git` skip the `fetch` step if the `cached_revision` matches the install path.

Since `install_path` uses only the first 10 characters of the full revision, there is a very very small chance of a collision happening, but It's so small that I think it can be ignored.

However my understanding of the codebase is too limited to be 100% sure the git gem would be properly updated if it's definition (`branch` / `ref` / `etc`) is updated. I tried to write a test case for this but couldn't figure out how to create a repo with multiple revisions. I'll keep digging, but I figured I might as well ask for feedback at this stage.

@segiddins any thoughts on this ? (since you fixed #4272), any pointers on how to better test this?

cc @rafaelfranca @jules2689

### DIFF
--- a/lib/bundler/source/git.rb
+++ b/lib/bundler/source/git.rb
@@ -260,7 +260,11 @@ module Bundler
       end
 
       def requires_checkout?
-        allow_git_ops? && !local?
+        !cached_revision_checked_out? && allow_git_ops? && !local?
+      end
+
+      def cached_revision_checked_out?
+        cached_revision && cached_revision == revision && install_path.exist?
       end
 
       def base_name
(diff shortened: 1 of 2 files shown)


segiddins:
I have no idea why travis choked, by I hope you dont mind me adding some tests ;)

segiddins:
@bundlerbot r+

casperisfine:
🎉 thanks a lot for improving the coverage !

PR #6713, opened 2018-09-28
Make config location relative to the current working directory

### LINKED ISSUE #2048: --standalone generates file in incorrect PWD

### LINKED ISSUE #6589: `settings[:gemfile]` from `.bundle/config` file is not respected when doing `require 'bundle/setup'`
Gemfile configuration from `.bundle/config` is only respected when you use `Bundler::CLI`. Can someone tell me why, please?

### LINKED ISSUE #7445: Bundler 2 does not auto-switch versions when BUNDLE_GEMFILE is set via .bundle/config
Bundler 2 is supposed to detect if a project was bundled with v1 based on the `BUNDLED WITH` lines in the lockfile.

This is not working correctly if:
1. The Gemfile is in a subdirectory or with an alternate name
2. The alternate location is specified in `.bundle/config` with `BUNDLE_GEMFILE`
3. `bundle update` is run from the project root directory

Note: This is working correctly if one of the following is true:
- `bundle update` is run from the subdirectory that includes the Gemfile
- _or_ `bundle update` is run with `BUNDLE_GEMFILE` env var set to the alternate location

## Expected result

Running bundler 2.x on a project that was bundled with 1.x would be that  `BUNDLED_WITH` would not be updated.

## Observed result

`BUNDLED WITH` is updated to 2.x

## Example repo to reproduce

A repository that demonstrates this behavior is located here: https://github.com/jtokoph/bundler2bug

The readme has steps to reproduce.

deivid-rodriguez:

There are two related end-user problems.

* When a custom `Gemfile` is configured via `BUNDLE_GEMFILE` and we run `bundle config`, the local config is not generated in the current working directory. The reason is in the diagnosis section, but the end user problem is that it is very counterintuitive. On bundler 1 this is aggravated because of "sticky options". See https://github.com/bundler/bundler/issues/2048#issuecomment-7615264 for an example.

* When configuring a custom gemfile via settings, it is not respected when requiring `bundler/setup`. See #6589.

My diagnosis was that the config location is currently resolved from the current `Gemfile` location, and that leads to configuration being created on counterintuitive locations. For example, you have `BUNDLE_GEMFILE` set to a particular non-root location, and run `bundle config` on the root folder, but no local config is apparently generated... until you figure out it's being generated in the same folder `BUNDLE_GEMFILE` is pointing to.

Also, since the config location depends on the location of the current `Gemfile`, we can't consistently resolve the `Gemfile` from configuration, because of a chicken-and-egg problem. We need the `Gemfile` to figure out where the config is, but we need the config to figure out where the `Gemfile` is. This is why #6589 can't currently work.

My fix was to change how we resolve the configuration root, and instead look up in the directory hierarchy for an existing local config, and otherwise default to the current working directory. With that in place, fixing #6589 is easy. My changes led to a single spec failure. Even if it's just one, it is backwards incompatible if backwards compatibility is defined by our current set of specs. Since I'd like to be able to opt-in to the new behavior right now on bundler 1, I made a feature flag for this change. This feature flag needed to be different from the other feature flags we currently have, because it can't be config-based (chicken-and-egg problem again). So I introduced "env-based feature flags".

Once we have the new configuration root resolution in place, the fix for #6589 is to add knowledge to the `find_gemfile` method to read from settings (and not just from env).

I chose this fix because changing the way we resolve configuration seemed like the only way to resolve the problems that I wanted to fix here.

Fixes #6589.
Fixes #2048 (the remaining part of it, https://github.com/bundler/bundler/issues/2048#issuecomment-7615264 and https://github.com/bundler/bundler/issues/2048#issuecomment-47390038).
Fixes #7445.

### DIFF
--- a/lib/bundler/feature_flag.rb
+++ b/lib/bundler/feature_flag.rb
@@ -25,6 +25,15 @@ module Bundler
     end
     private_class_method :settings_method
 
+    def self.env_flag(name, &default)
+      define_method(:"#{name}?") do
+        value = ENV["BUNDLE_#{name.to_s.upcase}"]
+        value = instance_eval(&default) if value.nil?
+        value
+      end
+    end
+    private_class_method :env_flag
+
     (1..10).each {|v| define_method("bundler_#{v}_mode?") { major_version >= v } }
 
     settings_flag(:allow_bundler_dependency_conflicts) { bundler_3_mode? }
@@ -50,6 +59,8 @@ module Bundler
     settings_flag(:use_gem_version_promoter_for_major_updates) { bundler_3_mode? }
 
     settings_option(:default_cli_command) { bundler_3_mode? ? :cli_help : :install }
+
+    env_flag(:config_relative_to_cwd) { bundler_3_mode? }
 
     def initialize(bundler_version)
       @bundler_version = Gem::Version.create(bundler_version)
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -12,8 +12,12 @@ module Bundler
   module SharedHelpers
     def root
       gemfile = find_gemfile
-      raise GemfileNotFound, "Could not locate Gemfile" unless gemfile
-      Pathname.new(gemfile).tap{|x| x.untaint if RUBY_VERSION < "2.7" }.expand_path.parent
+      return Pathname.new(gemfile).tap{|x| x.untaint if RUBY_VERSION < "2.7" }.expand_path.parent if gemfile
+
+      bundle_dir = default_bundle_dir
+      raise GemfileNotFound, "Could not locate Gemfile or .bundle/ directory" unless bundle_dir
+
+      bundle_dir.parent
     end
 
     def default_gemfile
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -238,7 +238,11 @@ module Bundler
 
     def find_gemfile
       require_relative "../bundler"
-      given = Bundler.settings[:gemfile]
+      given = if Bundler.feature_flag.config_relative_to_cwd?
+        Bundler.settings[:gemfile]
+      else
+        ENV["BUNDLE_GEMFILE"]
+      end
       return given if given && !given.empty?
       find_file(*gemfile_names)
     end
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -237,7 +237,8 @@ module Bundler
     end
 
     def find_gemfile
-      given = ENV["BUNDLE_GEMFILE"]
+      require_relative "../bundler"
+      given = Bundler.settings[:gemfile]
       return given if given && !given.empty?
       find_file(*gemfile_names)
     end
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -20,6 +20,12 @@ module Bundler
       bundle_dir.parent
     end
 
+    def config_root
+      bundle_dir = default_bundle_dir
+
+      bundle_dir ? bundle_dir.parent : Pathname.pwd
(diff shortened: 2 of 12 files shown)


lib/bundler/feature_flag.rb
```diff
@@ -19,12 +19,21 @@ def self.settings_option(key, &default)
     def self.settings_method(name, key, &default)
       define_method(name) do
         value = Bundler.settings[key]
-        value = instance_eval(&default) if value.nil? && !default.nil?
+        value = instance_eval(&default) if value.nil?
```
deivid-rodriguez: This is kind of unrelated to the PR, but I saw it while working on "env based feature settings". Currently we're never passing `default` as `nil`here, and I don't think it would make sense that we did that?

lib/bundler/shared_helpers.rb
```diff
@@ -25,8 +25,18 @@ module Bundler
...
       gemfile = find_gemfile
-      raise GemfileNotFound, "Could not locate Gemfile" unless gemfile
-      Pathname.new(gemfile).untaint.expand_path.parent
+      return Pathname.new(gemfile).untaint.expand_path.parent if gemfile
+
+      bundle_dir = default_bundle_dir
+      raise GemfileNotFound, "Could not locate Gemfile or .bundle/ directory" unless bundle_dir
+
+      bundle_dir.parent
+    end
```
deivid-rodriguez: Here I'm changing a bit what this method does. I take it that this is not public API, and if any, the public API would be `Bundler.root` (which still works as it worked before).

deivid-rodriguez:
There were too many unrelated changes in specs in this PR, so I extracted those improvements to #6714. Once that PR is reviewed and merged, I'll rebase this one so that the core changes of this PR are smaller and easier to review.

deivid-rodriguez:
I cleaned up this PR, it should be ready for a review now!

lib/bundler/shared_helpers.rb
```diff
@@ -245,7 +255,12 @@ def validate_bundle_path
...
-      given = ENV["BUNDLE_GEMFILE"]
+      require "bundler"
+      given = if Bundler.feature_flag.config_relative_to_cwd?
+        Bundler.settings[:gemfile]
+      else
+        ENV["BUNDLE_GEMFILE"]
+      end
       return given if given && !given.empty?
       names = gemfile_names
       names.reverse! if order_matters && Bundler.feature_flag.prefer_gems_rb?
```
deivid-rodriguez: As a side note, it took me a while to figure out that I needed to `require "bundler"` in this method, because I assumed the presence of `Bundler.feature_flag.prefer_gems_rb?` here implied that it was already available at this point. Turns out that this condition is never executed by our specs (or at least removing it does not cause any failures). So, maybe we need some coverage there, and also maybe the addition of `require "bundler"` is fixing some issue with the `prefer_gems_rb` feature flag... :man_shrugging:.

lib/bundler/feature_flag.rb
deivid-rodriguez: I extracted this change to #6884 because it was unrelated, so it should be discussed separately.

deivid-rodriguez:
It'd be great if anybody could share some thoughts about this :)

deivid-rodriguez:
Closing in favour of https://github.com/rubygems/rubygems/pull/3963.

PR #6714, opened 2018-09-28
Spec reset state and refactorings

deivid-rodriguez:

The problem was that while working on #6713, I noticed several issues like order dependent failures, situations hard to debug, or specs doing too much and taking longer than they should.
 
My diagnosis was that:

* Sometimes specs leak settings and env variable modifications.
* When a hang happens inside a subprocess, it's hard to debug because logging is not printed out anywhere.
* Some specs create unnecessary gemfiles and run `bundle install` multiple times unnecessarily.

My fix for state leaks is to reset state after each spec, for the hangs is to reorder some lines in the `sys_exec` helper (https://github.com/bundler/bundler/commit/b305a5b2524a6457b05a9d39e9526f75c98a0752), and for the unnecessary operations, to refactor the specs to avoid them.

I chose this fix because it seems like the best way to alleviate the issues found.

### DIFF
--- a/spec/commands/config_spec.rb
+++ b/spec/commands/config_spec.rb
@@ -1,13 +1,6 @@
 # frozen_string_literal: true
 
 RSpec.describe ".bundle/config" do
-  before :each do
-    gemfile <<-G
-      source "file://#{gem_repo1}"
-      gem "rack", "1.0.0"
-    G
-  end
-
   describe "config" do
     before { bundle "config foo bar" }
 
@@ -43,6 +36,13 @@ RSpec.describe ".bundle/config" do
   end
 
   describe "location" do
+    before :each do
+      gemfile <<-G
+        source "file://#{gem_repo1}"
+        gem "rack", "1.0.0"
+      G
+    end
+
     it "can be moved with an environment variable" do
       ENV["BUNDLE_APP_CONFIG"] = tmp("foo/bar").to_s
       bundle "install", forgotten_command_line_options(:path => "vendor/bundle")
@@ -66,7 +66,14 @@ RSpec.describe ".bundle/config" do
   end
 
   describe "global" do
-    before(:each) { bundle :install }
+    before(:each) do
+      gemfile <<-G
+        source "file://#{gem_repo1}"
+        gem "rack", "1.0.0"
+      G
+
+      bundle :install
+    end
 
     it "is the default" do
       bundle "config foo global"
@@ -155,7 +162,14 @@ RSpec.describe ".bundle/config" do
   end
 
   describe "local" do
-    before(:each) { bundle :install }
+    before(:each) do
+      gemfile <<-G
+        source "file://#{gem_repo1}"
+        gem "rack", "1.0.0"
+      G
+
+      bundle :install
+    end
 
     it "can also be set explicitly" do
       bundle "config --local foo local"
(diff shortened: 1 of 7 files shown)


spec/commands/config_spec.rb
```diff
@@ -42,31 +35,46 @@
...
+      gemfile <<-G
+        source "file://#{gem_repo1}"
+        gem "rack", "1.0.0"
+      G
+    end
+
     it "can be moved with an environment variable" do
-      ENV["BUNDLE_APP_CONFIG"] = tmp("foo/bar").to_s
-      bundle "install", forgotten_command_line_options(:path => "vendor/bundle")
+      with_env_vars("BUNDLE_APP_CONFIG" => tmp("foo/bar").to_s) do
```
segiddins: we already reset `ENV` after each spec, is this necessary?

spec/commands/config_spec.rb
```diff
@@ -42,31 +35,46 @@
...
+      end
     end
 
     it "can provide a relative path with the environment variable" do
       FileUtils.mkdir_p bundled_app("omg")
-      Dir.chdir bundled_app("omg")
 
-      ENV["BUNDLE_APP_CONFIG"] = "../foo"
-      bundle "install", forgotten_command_line_options(:path => "vendor/bundle")
+      Dir.chdir bundled_app("omg") do
```
segiddins: don't we reset the working directory after each spec?

deivid-rodriguez:
Now I'm not sure any more but I'm sure that the changes in #6713 definitely surfaced state leakage between specs. I should have probably tracked which changes fixed which problem, but unless there's something in the commit messages, I didn't. Anyways, the kind of leakage I saw was:

* Specs failing due to order.
* Specs leaving a `.bundle` folder in the root of the repo.
* Current working directory change not happening at the right time. I was using around hooks, and the directory would change in the middle of the spec for some reason, I finally tracked that down to https://github.com/rspec/rspec-core/issues/2544 and I think I fixed that by not using around hooks.

I figured I could open a PR since in my opinion these changes in general make the specs more robust and improve readability, no matter what external hooks there are in place to globally reset the state.

Anyways, what I'm going to do is remove these changes from #6713, and see the exact issues that surface. And then come back here.

deivid-rodriguez:
Ok, so I did what I said and... no issues. So it seems that all the problems I found during the development of #6713 where specific to some intermediate state but are not present in the final version...

I removed the env variable and current workding directory resets since, as you pointed out, are already reset after each spec :+1:.

Let me know your thoughts about the other commits!

Thanks @segiddins!

deivid-rodriguez:
Sorry for the back and forth. This PR in its current state _does_ fix an issue when running specs in #6713. If these changes are not applied, specs leave around a `.bundle/config` with the following content:

```
---
BUNDLE_NO_INSTALL: "true"
BUNDLE_MIRROR__HTTPS://RUBYGEMS__ORG/: "https://username:password@rubygems-mirror.org/"
```

I'm guessing the state leakage has been always there but didn't really matter before, since config was generated in the folder of the Gemfile in the temporary app (`tmp/bundled_app` or something like that). In any case, the changes here should fix it.

spec/install/gemfile/git_spec.rb
```diff
@@ -1010,13 +1009,11 @@
       install_gemfile <<-G
         gem "foo", :git => "file://#{lib_path("foo-1.0")}", :ref => "#{revision}"
       G
-      bundle "install"
```
segiddins: these `bundle install`s probably shouldn't be removed?

deivid-rodriguez: Aren't they redundant given that `install_gemfile` already runs that?

segiddins: 👍

deivid-rodriguez:
@segiddins Shall I move forward with this?

segiddins:
lgtm!

PR #6716, opened 2018-09-29
Remove if branch that always evaluate to false

gabrielgiordano:
`loaded_spec = nil` always evaluate to false, so the code in this if is never executed.
 
I'm not sure if this is a bug or if the branch is not necessary. Since I don't know the codebase well enough, I'm opening the pull request to at least bring the attention to a possible problem.

### DIFF
--- a/lib/bundler/source/metadata.rb
+++ b/lib/bundler/source/metadata.rb
@@ -21,9 +21,7 @@ module Bundler
             # can't point to the actual gemspec or else the require paths will be wrong
             s.loaded_from = File.expand_path("..", __FILE__)
           end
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Thanks @gabrielgiordano. This is the alternative to #6687. I'm not sure what's best, but I think I lean torwards this alternative?

colby-swandale:
I agree, we should be fixing the clause rather than just removing it.

deivid-rodriguez:
He, I actually meant that I think I prefer _this_ PR, since we wouldn't change how bundler is currently working in any way... But I'm not really sure...

colby-swandale:
Ah, sorry i didn't read that correctly. But i still think we should fix it rather then just removing it.

PR #6718, opened 2018-10-02
Correct `bundle show` deprecation

deivid-rodriguez:

The problem was `bundle show` deprecation messages are incorrect:

```
$ bundle show yard
[DEPRECATED FOR 2.0] use `bundle list` instead of `bundle show`
Resolving dependencies...
/home/deivid/Code/activeadmin/.bundle/ruby/2.5.0/gems/yard-0.9.16
$ bundle list yard
ERROR: "bundle list" was called with arguments ["yard"]
Usage: "bundle list"
```

My diagnosis was that deprecation messages only mention `bundle list`, but in some cases, the replacement is `bundle info`.

My fix is to replace "show" in the original command with the appropriate alternative in each case.

I chose this fix because it was the most user friendly message, since it prints the exact command the user needs to type to get rid of the warning.

### DIFF
--- a/spec/bundler/cli_spec.rb
+++ b/spec/bundler/cli_spec.rb
@@ -71,17 +71,17 @@ RSpec.describe "bundle executable" do
     it "prints the running command" do
       gemfile ""
       bundle! "info bundler", :verbose => true
-      expect(last_command.stdout).to start_with("Running `bundle info bundler --no-color --verbose` with bundler #{Bundler::VERSION}")
+      expect(last_command.stdout).to start_with("Running `bundle info bundler --verbose` with bundler #{Bundler::VERSION}")
     end
 
     it "doesn't print defaults" do
       install_gemfile! "", :verbose => true
-      expect(last_command.stdout).to start_with("Running `bundle install --no-color --retry 0 --verbose` with bundler #{Bundler::VERSION}")
+      expect(last_command.stdout).to start_with("Running `bundle install --retry 0 --verbose` with bundler #{Bundler::VERSION}")
     end
 
     it "doesn't print defaults" do
       install_gemfile! "", :verbose => true
-      expect(last_command.stdout).to start_with("Running `bundle install --no-color --retry 0 --verbose` with bundler #{Bundler::VERSION}")
+      expect(last_command.stdout).to start_with("Running `bundle install --retry 0 --verbose` with bundler #{Bundler::VERSION}")
     end
   end
 
--- a/spec/support/helpers.rb
+++ b/spec/support/helpers.rb
@@ -99,9 +99,6 @@ module Spec
       with_sudo = options.delete(:sudo)
       sudo = with_sudo == :preserve_env ? "sudo -E" : "sudo" if with_sudo
 
-      no_color = options.delete("no-color") { cmd.to_s !~ /\A(e|ex|exe|exec|conf|confi|config)(\s|\z)/ }
-      options["no-color"] = true if no_color
-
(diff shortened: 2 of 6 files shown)


spec/bundler/cli_spec.rb
```diff
@@ -71,17 +71,17 @@
     it "prints the running command" do
       gemfile ""
       bundle! "info bundler", :verbose => true
-      expect(last_command.stdout).to start_with("Running `bundle info bundler --no-color --verbose` with bundler #{Bundler::VERSION}")
+      expect(last_command.stdout).to start_with("Running `bundle info bundler --verbose` with bundler #{Bundler::VERSION}")
```
segiddins: why did this change?

spec/support/helpers.rb
```diff
@@ -99,9 +99,6 @@ def bundle(cmd, options = {})
       with_sudo = options.delete(:sudo)
       sudo = with_sudo == :preserve_env ? "sudo -E" : "sudo" if with_sudo
 
-      no_color = options.delete("no-color") { cmd.to_s !~ /\A(e|ex|exe|exec|conf|confi|config)(\s|\z)/ }
-      options["no-color"] = true if no_color
```
segiddins: Why did you remove this?

spec/bundler/cli_spec.rb
deivid-rodriguez: See below.

spec/support/helpers.rb
deivid-rodriguez: I was trying to test the command line arguments that `bundle show` receives and for some reason I was getting the `--no-color` flag added to them? Then I found this flag and I added `s|sh|sho|show` as exceptions here and it worked.

But then I decided that doing this didn't make any sense to me, so I removed it so I could get some feedback from you :smiley:. Why are we doing that?

segiddins: So all the test output will never contain ANSI escape codes , and thus be consistent no matter how the tests are being run

deivid-rodriguez: Can you show an example of this inconsistency? Our current test environment seems to handle this just fine? (I think the explanation of why this just works is that an `open3` context is recognized by ruby as a "non-tty" situation, so it's automatically [turned off](https://github.com/bundler/bundler/blob/e115a839123a21365904fd86980452d2b2b0283e/lib/bundler/ui/shell.rb#L13) by `bundler`).

To me doing this is not good because the helper does not represent real world usage, but a slight modification of it, and it leads to weird assertions (like the ones I'm fixing here), and suprising behavior for the developer (when I first saw this I was like "what? where did this flag come from?", and I had to spend a while tracking it down and adding yet more exceptions here).

segiddins: You could also pass no-color => false to the bundle helper

deivid-rodriguez: Right. So is that what you're asking me to do?

deivid-rodriguez: I'm going to do that and make my case to remove this separately :slightly_smiling_face:.

deivid-rodriguez: This flag will be needed on most `bundle show` specs. Do you prefer that I explicitly pass the flag to every `bundle` invocation that needs it, that I add more exceptions inside the method, or that I leave this patch as it is?

segiddins: It's your call :) If it doesn't break the tests, I suppose removing it is fine?

deivid-rodriguez: Ok, yeah, I think it's fine!

deivid-rodriguez:
@bundlerbot r+

PR #6719, opened 2018-10-03
Update for new bors merge commit message

segiddins:

The problem was andre updated @bundlerbot and it makes merges with a new message now

My diagnosis was the merge message had changed

My fix update a regexp or two


colby-swandale:
@bundlerbot  r+

PR #6720, opened 2018-10-03
[WIP] Log all messages (at all levels) to a file

segiddins:

The problem was...

My diagnosis was...

My fix...

I chose this fix because...

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -66,8 +66,21 @@ module Bundler
       @configured ||= configure_gem_home_and_path
     end
 
+    def ui_logger
+      @ui_logger ||= begin
+        io = begin
+          path = tmp.join("output.log")
+          File.open(path, "w")
+        rescue
+          nil
+        end
+
+        UI::Logger.new(io)
+      end
+    end
+
     def ui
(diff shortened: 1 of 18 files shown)


indirect (changes requested):
This looks great! Add some tests and fill out the PR message and I’m happy. 😄 

Tiny nits: 
- trace is not a log level (it should be logged as debug)
- don’t use __method__ when every other line in the methods just use the literals (plus it leads to things like treating trace like a level)
- rename `post_log_progress` to `log_progress_end` or something like that

indirect:
Also, it should probably only do that log thing when there's an unhandled exception, rather than any time there's any exception at all. This looks a little silly. 😅 

```
$ bundle install
Could not locate Gemfile

The full bundler log is available in /var/folders/l5/qk8y24j13snb43ngkyzdgkp40000gn/T/bundler20181012-26085-15d6y5a26085/output.log
$ gs
## segiddins/log-everything-in-tmp...origin/segiddins/log-everything-in-tmp
 M lib/bundler/installer.rb
$ git nuke
HEAD is now at 290c44caa Log all messages (at all levels) to a file
```

PR #6721, opened 2018-10-03
http:// URLs in the gemspec

amatsuda:

When the users visit https://rubygems.org/gems/bundler and click the links, or access the gem metadata from scripts e.g. gem-src, we're seeing unneeded redirection from http://... to https://... because both github.com and bundler.io redirects http requests to https.

Rewrote all http:// URLs in the gemspec to https://.


colby-swandale:
Thanks!

@bundlerbot r+

Issue #6595, opened 2018-06-20
Bundler 1 crashes on gemfiles generated using bundler 2


deivid-rodriguez:
I think it could be reasonable to warn or even fail with an upgrade message, but the crash one currently gets seems like a no-go.

#### To reproduce

```
$ git clone bundler/bundler

$ cd bundler

$ bin/rake install
Could not find 'automatiek' (~> 0.1.0) among 562 total gem(s)
Checked in 'GEM_PATH=/home/deivid/.gem/ruby/2.5.0:/home/deivid/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0', execute `gem env` for more information (Gem::MissingSpecError)
Could not find 'mustache' (= 0.99.6) - did find: [mustache-1.0.5]
Checked in 'GEM_PATH=/home/deivid/.gem/ruby/2.5.0:/home/deivid/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0', execute `gem env` for more information (Gem::MissingSpecVersionError)
Could not find 'rdiscount' (~> 2.2) among 562 total gem(s)
Checked in 'GEM_PATH=/home/deivid/.gem/ruby/2.5.0:/home/deivid/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0', execute `gem env` for more information (Gem::MissingSpecError)
Could not find 'ronn' (~> 0.7.3) among 562 total gem(s)
Checked in 'GEM_PATH=/home/deivid/.gem/ruby/2.5.0:/home/deivid/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0', execute `gem env` for more information (Gem::MissingSpecError)
Install the ronn gem to build the help pages
bundler 2.0.0.dev built to pkg/bundler-2.0.0.dev.gem.
bundler (2.0.0.dev) installed.
install ran for   0.000000   0.004000   1.628000 (  2.361539)

$ bundle --version
2.0.0.dev

$ mkdir test_app

$ cd test_app

$ echo -e "source 'https://rubygems.org'\n\ngem 'byebug'" > Gemfile

$ bundle install
Fetching gem metadata from https://rubygems.org/.......
Resolving dependencies...
Installing byebug 10.0.2 with native extensions
Bundle complete! 1 Gemfile dependency, 2 gems now installed.
Bundled gems are installed into `./.bundle`

$ cat Gemfile.lock 
GEM
  remote: https://rubygems.org/
  specs:
    byebug (10.0.2)

[...]

PR #6725, opened 2018-10-05
Test bad rubygems binstub + locked bundler behavior

deivid-rodriguez:

The problem was that we have no specs to demostrate the issue in #6595, nor to prove the fix in https://github.com/rubygems/rubygems/pull/2426.

My diagnosis was that we need those specs.

My fix is to write the specs. They don't look very good though... :sweat_smile:.

I chose this fix because I couldn't think of any other ways to write them.

### DIFF
--- a/spec/lock/lockfile_bundler_1_spec.rb
+++ b/spec/lock/lockfile_bundler_1_spec.rb
@@ -201,6 +201,44 @@ RSpec.describe "the lockfile format", :bundler => "< 2" do
     G
   end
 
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
This fails, as expected, since the fix is not yet on rubygems master...

segiddins:
Should this test go in the rubygems repo?

deivid-rodriguez:
Yeah, most likely...

deivid-rodriguez:
Closing since I managed to write a spec in the rubygems PR that definitely feels more natural than what I'm doing here :+1:.

PR #6726, opened 2018-10-06
Add 1.16.6 release to master's CHANGELOG

deivid-rodriguez:

The problem was that I couldn't see the latest release's changelog from my copy pointing to master.

My diagnosis was that the release was missing.

My fix is to add it to the changelog.

I could've merged 1-16-stable into master and then push, but I'm not sure I can push to master, and I prefer going through a PR anyways?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,25 @@
+## 1.16.6 (2018-10-05)
+
+Changes:
+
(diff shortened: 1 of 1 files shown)


colby-swandale:
The CHANGELOG for minor releases is updated through `git merge` in the release process. I just didn't get around to it yet but i'll do that now.

deivid-rodriguez:
I see it's on master now. Thanks so much @colby-swandale!

Issue #6639, opened 2018-07-25
bundler 1.16.3 Bundler::HTTPError: Net::HTTPTooManyRequests 


mfmcdonagh:
Hello,
I'm running bundler in a ci enviorment, 

I've stared to see the following errors:

```
2018-07-24 23:21:21.857 GMT |  HTTP GET https://index.rubygems.org/info/hirb
2018-07-24 23:21:21.857 GMT |  Bundler::HTTPError: Net::HTTPTooManyRequests: 
2018-07-24 23:21:21.857 GMT |  
2018-07-24 23:21:21.857 GMT |  
2018-07-24 23:21:21.857 GMT |  
429 Too Many Requests

2018-07-24 23:21:21.857 GMT |  
nginx

2018-07-24 23:21:21.857 GMT |  
2018-07-24 23:21:21.857 GMT |  
2018-07-24 23:21:21.857 GMT |  
2018-07-24 23:21:21.857 GMT |  /usr/local/bundle/gems/bundler-1.16.3/lib/bundler/fetcher/downloader.rb:37:in `fetch'
2018-07-24 23:21:21.857 GMT |    /usr/local/bundle/gems/bundler-1.16.3/lib/bundler/fetcher/compact_index.rb:117:in `call'
2018-07-24 23:21:21.857 GMT |    /usr/local/bundle/gems/bundler-1.16.3/lib/bundler/compact_index_client/updater.rb:53:in `block in update'
2018-07-24 23:21:21.857 GMT |    /usr/local/lib/ruby/2.1.0/tmpdir.rb:88:in `mktmpdir'
2018-07-24 23:21:21.857 GMT |    /usr/local/bundle/gems/bundler-1.16.3/lib/bundler/compact_index_client/updater.rb:31:in `update'
2018-07-24 23:21:21.857 GMT |    /usr/local/bundle/gems/bundler-1.16.3/lib/bundler/compact_index_client.rb:82:in `update'
2018-07-24 23:21:21.857 GMT |    /usr/local/bundle/gems/bundler-1.16.3/lib/bundler/compact_index_client.rb:98:in `update_info'
2018-07-24 23:21:21.857 GMT |    /usr/local/bundle/gems/bundler-1.16.3/lib/bundler/compact_index_client.rb:55:in `block in dependencies'
2018-07-24 23:21:21.857 GMT |    /usr/local/bundle/gems/bundler-1.16.3/lib/bundler/fetcher/compact_index.rb:88:in `call'
[...]

Issue #6471, opened 2018-04-04
bundler attempts to install gems which conflict with an explicit ruby version requirement


lamont-granquist:
Simple example:

```
% cat Gemfile
source "https://rubygems.org"
ruby "2.3.7"
gem "berkshelf"
% bundle install
Fetching gem metadata from https://rubygems.org/.............
Fetching gem metadata from https://rubygems.org/..
Resolving dependencies...
Using public_suffix 3.0.2
Using addressable 2.5.2
Using buff-extensions 2.0.0
Using hashie 3.5.7
Using varia_model 0.6.0
Using buff-config 2.0.0
Using bundler 1.16.1
Using fuzzyurl 0.9.0
Using tomlrb 1.2.6
Using mixlib-config 2.2.6
Using mixlib-shellout 2.3.2
Using chef-config 14.0.190
Using libyajl2 1.2.0
Using ffi-yajl 2.3.1
Using mixlib-log 2.0.3
Using rack 2.0.4
Using uuidtools 2.1.5
Fetching chef-zero 14.0.1
Installing chef-zero 14.0.1
Gem::RuntimeRequirementNotMetError: chef-zero requires Ruby version >= 2.4.3. The current ruby version is 2.3.0.
An error occurred while installing chef-zero (14.0.1), and Bundler cannot continue.
Make sure that `gem install chef-zero -v '14.0.1'` succeeds before bundling.

In Gemfile:
  berkshelf was resolved to 6.3.1, which depends on
    chef was resolved to 14.0.190, which depends on
      chef-zero
% bundle --version
Bundler version 1.16.1
```

chef-14.0.190 is pinned >= 2.4.0 here:

https://github.com/chef/chef/blob/97c1dd6f1cac6d97e85d05039cad8b28596141ba/chef.gemspec#L16

so chef-14 shouldn't been in the solution set.  But it only pins chef-zero to >= 13.0 here:

https://github.com/chef/chef/blob/97c1dd6f1cac6d97e85d05039cad8b28596141ba/chef.gemspec#L35

That itself should have allowed chef-zero 13.x versions which supported 2.3.x to install, but it selected the latest chef-zero-14.x which has been bumped to >= 2.4.0

lamont-granquist:
Also note that rubygems doesn't get this right either:

```% gem install berkshelf
ERROR:  Error installing berkshelf:
[...]

PR #6728, opened 2018-10-08
Fallback to sequentially fetching specs on 429s

deivid-rodriguez:

The problem is that sometimes `bundler` is unable to resolve certain gemfiles. Specifically, sometimes it does not respect the `required_ruby_version` setting. This causes some people to assume that `bundler` will always try to install the latest version of any dependency, regardless of the ruby version being run, because as a matter of fact, this feature sometimes just doesn't work. See for example the discussion at https://github.com/rspec/rspec/issues/25.

The problem was consistently reproducible until a few minutes ago with the following `Gemfile` under `ruby 2.3.7`

```ruby
source "https://rubygems.org"
ruby "2.3.7"
gem "berkshelf", "= 6.3.1"
```

```
$ docker run -it --rm --volume $(pwd):/app ruby:2.3.7 sh -c "cd /app && rm -f Gemfile.lock && bundle install"
Fetching gem metadata from https://rubygems.org/.............
Fetching gem metadata from https://rubygems.org/..
Resolving dependencies....
Fetching public_suffix 3.0.3
Installing public_suffix 3.0.3
Fetching addressable 2.5.2
Installing addressable 2.5.2
Fetching buff-extensions 2.0.0
Installing buff-extensions 2.0.0
Fetching hashie 3.6.0
Installing hashie 3.6.0
Fetching varia_model 0.6.0
Installing varia_model 0.6.0
Fetching buff-config 2.0.0
Installing buff-config 2.0.0
Using bundler 1.16.5
Fetching fuzzyurl 0.9.0
Installing fuzzyurl 0.9.0
Fetching tomlrb 1.2.7
Installing tomlrb 1.2.7
Fetching mixlib-config 2.2.13
Installing mixlib-config 2.2.13
Fetching mixlib-shellout 2.4.0
Installing mixlib-shellout 2.4.0
Fetching chef-config 14.5.33
Installing chef-config 14.5.33
Fetching libyajl2 1.2.0
Installing libyajl2 1.2.0 with native extensions
Fetching ffi-yajl 2.3.1
Installing ffi-yajl 2.3.1 with native extensions
Fetching mixlib-log 2.0.4
Installing mixlib-log 2.0.4
Fetching rack 2.0.5
Installing rack 2.0.5
Fetching uuidtools 2.1.5
Installing uuidtools 2.1.5
Fetching chef-zero 14.0.6
Installing chef-zero 14.0.6
Gem::RuntimeRequirementNotMetError: chef-zero requires Ruby version >= 2.4.0. The current ruby version is 2.3.0.
An error occurred while installing chef-zero (14.0.6), and Bundler cannot continue.
Make sure that `gem install chef-zero -v '14.0.6' --source 'https://rubygems.org/'` succeeds before bundling.

In Gemfile:
  berkshelf was resolved to 6.3.1, which depends on
    chef was resolved to 14.5.33, which depends on
      chef-zero
```

Funny enough, I can no longer reproduce it at the moment, I guess it depends on the specific load conditions of the rubygems.org servers?

My diagnosis was that sometimes our resolution falls back to the old dependency API that didn't implement the `required_ruby_version` setting. In particular, this happens because the new API returns `Net::HTTPTooManyRequests`, so `bundler` gives up and defaults to the old API.

My fix is to, instead of directly fall back to the old API when rate limiting happens, try first to fetch the dependencies sequentially instead of in parallel still from the new API, so that rate limit does not affect us.

I chose this fix because it was the only idea that came up. As a matter of fact, #6471 and #6639 were closed because there was nothing we could do, so it seems like it's the only idea so far :)

### DIFF
--- a/lib/bundler/compact_index_client.rb
+++ b/lib/bundler/compact_index_client.rb
@@ -18,11 +18,6 @@ module Bundler
 
     attr_reader :directory
 
-    # @return [Lambda] A lambda that takes an array of inputs and a block, and
-    #         maps the inputs with the block in parallel.
-    #
-    attr_accessor :in_parallel
-
     def initialize(directory, fetcher)
       @directory = Pathname.new(directory)
       @updater = Updater.new(fetcher)
@@ -31,7 +26,28 @@ module Bundler
       @info_checksums_by_name = {}
       @parsed_checksums = false
       @mutex = Mutex.new
-      @in_parallel = lambda do |inputs, &blk|
+    end
+
+    def execution_mode=(block)
+      Bundler::CompactIndexClient.debug { "execution_mode=" }
+      @endpoints = Set.new
+
+      @execution_mode = block
+    end
+
+    # @return [Lambda] A lambda that takes an array of inputs and a block, and
+    #         maps the inputs with the block in parallel.
+    #
+    def execution_mode
+      @execution_mode || sequentially
+    end
+
+    def sequential_execution_mode!
+      self.execution_mode = sequentially
+    end
+
+    def sequentially
+      @sequentially ||= lambda do |inputs, &blk|
         inputs.map(&blk)
       end
     end
@@ -51,7 +67,7 @@ module Bundler
 
     def dependencies(names)
       Bundler::CompactIndexClient.debug { "dependencies(#{names})" }
-      in_parallel.call(names) do |name|
+      execution_mode.call(names) do |name|
         update_info(name)
         @cache.dependencies(name).map {|d| d.unshift(name) }
       end.flatten(1)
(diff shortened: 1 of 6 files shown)


indirect:
This seems like a great idea :)

lamont-granquist:
:+1:

deivid-rodriguez:
This should be good to go now that we've dropped 1.8 support! :smiley:

deivid-rodriguez:
My bad, I thought master had finally dropped support... I'll keep waiting :)

deivid-rodriguez:
I rebased this PR and it's now passing! :tada:

indirect:
@bundlerbot r+

Issue #6729, opened 2018-10-08
bundler should print warnings to STDERR instead of STDOUT


dgpokl:
Encountered an issue today where running a command like `bundle show rubocop-junit-formatter` prints a warning to STDOUT, screwing up a script. I think warnings should be printed to STDERR so that they won't pollute the expected output of `bundle` commands.

#### Version: 
`bundler-1.16.6`

#### Steps to reproduce:
Platform: Linux or Mac OS
Gemfile:
```
gem 'rubocop-junit-formatter'
# Windows does not include zoneinfo files, so bundle the tzinfo-data gem
gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
```
Run:
```
`bundle`
`bundle show rubocop-junit-formatter`
#### Expected results:
This should be printed to **STDERR**:
```
The dependency tzinfo-data (>= 0) will be unused by any of the platforms Bundler is installing for. Bundler is installing for ruby but the dependency is only for x86-mingw32, x86-mswin32, x64-mingw32, java. To add those platforms to the bundle, run `bundle lock --add-platform x86-mingw32 x86-mswin32 x64-mingw32 java`.
```
This should be printed to **STDOUT**:
/Users/dpritchard/.rvm/gems/ruby-2.5.1@search_api/gems/rubocop-junit-formatter-0.1.4

#### Actual results:
All the above is printed to STDOUT.

deivid-rodriguez:
This will be the default behavior in bundler 2 and you can opt in on bundler 1 via `bundle config error_on_stderr true` or the `BUNDLER_ERROR_ON_STDERR` environment variable.

Although I would honestly ship this as a non-configurable bug fix, not sure how other @bundler/maintainers feel about it.

greysteil:
I’m 👍 on shipping as a bug fix unless there’s history to this one I’m not aware of

deivid-rodriguez:
I think this might also make Rails maintainers happy because I think currently they keep a list of warnings they need to filter out from the output of their generator tests. Maybe doing this will make all that go away? /cc @y-yagi

colby-swandale:
Printing errors to `STDERR` is a breaking change, so it's slated to come out in Bundler 2

deivid-rodriguez:
Any bug fix can be considered a breaking change, and viceversa, if [worded properly](https://xkcd.com/1172/) :smiley:.

[...]

PR #6730, opened 2018-10-09
Print errors to stderr by default, and remove configuration option

greysteil:

The problem was #6729 - Bundler unexpectedly outputs error and warning messages to STDOUT.

My diagnosis was that whilst very minorly breaking, this is essentially a bug fix, and should be considered for inclusion for Bundler 2.0 even if very few other breaking changes are.

My fix was so switch output for warning and error messages to STDERR, and remove the configuration option (as is redundant once the setup is flipped - anyone wanting to redirect those message to STDOUT could do so in their shell).

I chose this fix because I think the new behaviour is what everyone would expect, and we should get it out from behind its feature switch sooner rather than later.

Alternatively, we might want to keep the Bundler 2.0 release "purer" by only dropping Ruby versions in it - that's totally fine too, but I figured we should have the code to discus #6729, rather than doing it in abstract.


colby-swandale:
Apologies but i'm confused, the existing code is already set to be enabled for Bundle 2. Why do we need to remove the feature flag?

spec/commands/exec_spec.rb
```diff
@@ -594,17 +579,6 @@ def bin_path(a,b,c)
...
-      let(:exit_code) { 1 }
-      let(:expected) { super() << "\nbundler: failed to load command: #{path} (#{path})" }
-      let(:expected_err) do
-        "RuntimeError: ERROR\n  #{path}:10" +
-          (Bundler.current_ruby.ruby_18? ? "" : ":in `<top (required)>'")
-      end
-      it_behaves_like "it runs"
-    end
-
     context "the executable raises", :bundler => "2" do
```
deivid-rodriguez: You should remove `:bundler => 2` so that this test actually runs, always.

spec/commands/exec_spec.rb
```diff
@@ -615,15 +589,6 @@ def bin_path(a,b,c)
...
-    context "the executable raises an error without a backtrace", :bundler => "< 2" do
-      let(:executable) { super() << "\nclass Err < Exception\ndef backtrace; end;\nend\nraise Err" }
-      let(:exit_code) { 1 }
-      let(:expected) { super() << "\nbundler: failed to load command: #{path} (#{path})" }
-      let(:expected_err) { "Err: Err" }
-
-      it_behaves_like "it runs"
-    end
-
     context "the executable raises an error without a backtrace", :bundler => "2" do
```
deivid-rodriguez: Same here.

deivid-rodriguez:
@greysteil I wouldn't work on this any more because there's no green light for it, but just in case we end up doing this, documentation about `BUNDLE_ERROR_ON_STDERR` would also need to be removed.

spec/commands/exec_spec.rb
```diff
@@ -570,21 +572,6 @@ def bin_path(a,b,c)
...
-      if LessThanProc.with(RUBY_VERSION).call("1.9")
-        # Kernel#exec in ruby < 1.9 will raise Errno::ENOEXEC if the command content is empty,
-        # even if the command is set as an executable.
-        pending "Kernel#exec is different"
-      else
-        it_behaves_like "it runs"
-      end
-    end
-
     context "the executable is empty", :bundler => "2" do
```
deivid-rodriguez: Missed this tag on my initial review that should also be removed...

greysteil:
I have no idea what's causing the final test failure here. I'll have another go at fixing it tomorrow, unless anyone can spot anything I'm missing.

greysteil:
@colby-swandale - any reason I can't base this off of `2-0-stable`?

colby-swandale:
Seeing this is not fixing anything broken on `2-0-0`, this will need to go through the normal release process after 2.0 is released.

greysteil:
Yep, at this point this PR is just code cleanup, and removing the `error_on_stderr` option.

deivid-rodriguez:
I rebased this PR and added some extra simplifications. Thanks @colby-swandale from preventing me from stealing attribution :sweat_smile:, and @greysteil for creating this PR :heart:.

greysteil:
Thanks @deivid-rodriguez! ❤️

deivid-rodriguez:
@greysteil Can't ask for a review from you since you're the author (I'm always bothered by this :grimacing:), but feel free to merge away if it looks good to you! :)

greysteil:
@bundlerbot r+

PR #6733, opened 2018-10-11
Check for file or folder when checking for git hash in build metadata

colby-swandale:

A change that was introduced in #6664 is breaking RubyGem's specs because the Build Metadata is not being generated correctly.

This is breaking because the change is checking for the `.git` to determine if the git hash should be fetched for the version string. This folder does not exist in the submodule in RubyGems' repo

See https://travis-ci.org/rubygems/rubygems/jobs/437576972

Check if `.git` exists either as a file or folder

### DIFF
--- a/lib/bundler/build_metadata.rb
+++ b/lib/bundler/build_metadata.rb
@@ -28,7 +28,9 @@ module Bundler
       # If Bundler has been installed without its .git directory and without a
       # commit instance variable then we can't determine its commits SHA.
       git_dir = File.join(File.expand_path("../../..", __FILE__), ".git")
-      return "unknown" unless File.directory?(git_dir)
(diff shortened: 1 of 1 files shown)


greysteil:
@colby-swandale - will the `git rev-parse --short HEAD` command that will then be run actually return in this case? Otherwise we should change the test, not the code.

colby-swandale:
Yes, it will return a hash (although it's RubyGem's git hash)

```
~/Projects/rubygems/bundler
❯ git rev-parse --short HEAD
a356d8cf9
```

greysteil:
Cool, that still sounds useful. 👍 from me.

colby-swandale:
@bundlerbot r+

PR #6734, opened 2018-10-11
Still document the `--force` option

deivid-rodriguez:

The problem was that `--force` was no longer mentioned in the 1-17-stable docs, even though it's a first-class option.

My diagnosis was that I completely replaced docs to mention `--redownload` instead, but in this series `--force` is not yet deprecated, so it should be documented.

My fix is to re-document it, and only mention `--redownload` as an alias. 

I chose this fix because that's @colby-swandale's preferred option, and I don't have a strong opinion (the alternative would be to document it only as an alias to `--redownload`, and not the order way around).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ env:
 matrix:
   include:
     - rvm: 2.4.1
-      env: RGV=v2.6.8 BUNDLER_SPEC_SUB_VERSION=2.0.0
+      env: RGV=v2.6.8
     # Ruby 2.4, Rubygems 2.6.8 and up
     # Ruby 2.3, Rubygems 2.5.1 and up
     - rvm: 2.2.6
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,17 @@
+## 11.7.0.pre.2 (2018-10-11)
+
+Features:
+
+  - Add `--optimistic` and `--strict` options to `bundle add` that sets a version constraint for new gems ([#6553](https://github.com/bundler/bundler/issues/6553), @agrim123)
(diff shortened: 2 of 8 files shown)


colby-swandale:
Thank you! @bundlerbot r+

deivid-rodriguez:
Failures are weird, I don't think they're caused by these doc updates, but they are related to the option deprecations, so I'll have a look!

deivid-rodriguez:
@colby-swandale I fixed some of the failures in 5618aaf025ea9084f2e8e829160e313306461325, we were testing in bundler 2 mode in this branch on a specific matrix entry, for some reason...

The other failures are related to `bundle remove` but the backport of `bundle remove` is not mentioned in the changelog so I'm not sure it's intended?

colby-swandale:
I've cherry-picked `221c64c` into `1-17-stable` which includes patches to resolve the failing specs.

deivid-rodriguez:
Colby, have a look at 5618aaf and https://github.com/bundler/bundler/pull/6734/commits/a875c7e25a2e59df6aec26701e35f18af91fc408 too. I think https://github.com/bundler/bundler/commit/4b40d4d62f98bd1e83d013fd1199c1aee8387f44 is wrong because the scoping to `bundler < 2` is already done on the inner specs that need it. The problem is that we shouldn't be running bundler 2 specs in this branch, so it the `BUNDLER_SPEC_SUB_VERSION=2.0.0` in `.travis.yml` that t's wrong.

deivid-rodriguez:
I don't understand... Feature flags are tested on 1.x stable, they just have different values on this branch. Even if that was needed, what makes the Rubygems 2.6.8 +  ruby 2.4.1 special to be the only matrix entry that needs this flag?

PR #6737, opened 2018-10-12
when running from source, find submodule shas too

indirect:
We recently merged #6664 to prevent Bundler from (wrongly) claiming the
git commit of any parent directory when it is run from source. However,
Bundler is also run from source as a submodule in RubyGems, and in that
case it does not have its own git directory.

This commit resolves the failure in version_spec.rb that only appears
when the Bundler tests are run from a submodule, by explicitly checking
for that submodule, and using `git ls-tree` to get the sha if so.

Arguably, this is a bugfix compared to the previous behavior, which
would blindly print the current sha from the _RubyGems_ repo, while
claiming that it was the sha of Bundler.

### DIFF
--- a/lib/bundler/build_metadata.rb
+++ b/lib/bundler/build_metadata.rb
@@ -28,14 +28,19 @@ module Bundler
       # If Bundler has been installed without its .git directory and without a
       # commit instance variable then we can't determine its commits SHA.
       git_dir = File.join(File.expand_path("../../..", __FILE__), ".git")
-      # Check for both a file or folder because RubyGems runs Bundler's test suite in a submodule
-      # which does not have a .git folder
-      return "unknown" unless File.exist?(git_dir)
-
-      # Otherwise shell out to git.
-      @git_commit_sha = Dir.chdir(File.expand_path("..", __FILE__)) do
-        `git rev-parse --short HEAD`.strip.freeze
+      if File.directory?(git_dir)
+        @git_commit_sha = Dir.chdir(git_dir) { `git rev-parse --short HEAD`.strip.freeze }
       end
+
+      # If Bundler is a submodule in RubyGems, get the submodule commit
+      git_sub_dir = File.join(File.expand_path("../../../..", __FILE__), ".git")
+      if File.directory?(git_sub_dir)
+        @git_commit_sha = Dir.chdir(git_sub_dir) do
+          `git ls-tree --abbrev=8 HEAD bundler | awk '{ print $3 }'`.strip.freeze
+        end
+      end
+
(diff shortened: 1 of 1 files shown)


lib/bundler/build_metadata.rb
```diff
@@ -28,14 +28,19 @@ def self.git_commit_sha
...
+      end
 
-      # Otherwise shell out to git.
-      @git_commit_sha = Dir.chdir(File.expand_path("..", __FILE__)) do
-        `git rev-parse --short HEAD`.strip.freeze
+      # If Bundler is a submodule in RubyGems, get the submodule commit
+      git_sub_dir = File.join(File.expand_path("../../../..", __FILE__), ".git")
+      if File.directory?(git_sub_dir)
+        @git_commit_sha = Dir.chdir(git_sub_dir) do
+          `git ls-tree --abbrev=8 HEAD bundler | awk '{ print $3 }'`.strip.freeze
```
deivid-rodriguez: Maybe it's best to process output in ruby so we don't introduce a dependency on `awk`?

hsbt:
What's status this and https://github.com/bundler/bundler/pull/6733?

When they are merged `1-16-stable` branch, I changed its HEAD revision on rubygems/rubygems repository and will release RG 2.7.8. Thanks!

indirect:
@bundlerbot r+

lib/bundler/build_metadata.rb
```diff
@@ -28,14 +28,19 @@ def self.git_commit_sha
...
+      if File.directory?(git_dir)
+        @git_commit_sha = Dir.chdir(git_dir) { `git rev-parse --short HEAD`.strip.freeze }
+      end
 
-      # Otherwise shell out to git.
-      @git_commit_sha = Dir.chdir(File.expand_path("..", __FILE__)) do
-        `git rev-parse --short HEAD`.strip.freeze
+      # If Bundler is a submodule in RubyGems, get the submodule commit
+      git_sub_dir = File.join(File.expand_path("../../../..", __FILE__), ".git")
+      if File.directory?(git_sub_dir)
```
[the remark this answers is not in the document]
indirect: good point, added `return`s.

lib/bundler/build_metadata.rb
indirect: good point! switched to `.split.fetch`

indirect:
@hsbt see #6752 for backport to 1-16-stable 👍🏻

hsbt:
@indirect Thanks a lot. I will prepare to release RG 2.7.8.

Issue #5992, opened 2017-08-31
Bundler installs unsuitable gem versions on Windows


leoarnold:
**Environments:** (all possible combinations were tested)
* Windows 10 or Windows Server 2012
* Ruby:
   * ruby 2.1.9p490 (2016-03-30 revision 54437) [x64-mingw32]
   * ruby 2.1.9p490 (2016-03-30 revision 54437) [i386-mingw32]
   * ruby 2.0.0p648 (2015-12-16) [x64-mingw32]
   * ruby 2.0.0p648 (2015-12-16) [i386-mingw32]
   * ruby 1.9.3p551 (2014-11-13) [i386-mingw32]
* Gem 2.6.13
* Bundler 1.15.4
* Gemfile:

```ruby
source 'https://rubygems.org'

gem 'nokogiri'
```

**Problem:** Bundler installs a version of Nokogiri which requires `< 2.5, >= 2.2` even though Bundler is running on Ruby `< 2.2`.

```
Fetching nokogiri 1.8.0 (x64-mingw32) # respectivly (x86-mingw32)
Installing nokogiri 1.8.0 (x64-mingw32)
Gem::InstallError: nokogiri requires Ruby version < 2.5, >= 2.2.
An error occurred while installing nokogiri (1.8.0), and Bundler cannot
continue.
Make sure that `gem install nokogiri -v '1.8.0'` succeeds before bundling.
In Gemfile:
  nokogiri
Command exited with code 5
```

**Suspicion:** Bundler chooses the `mingw32` version, but uses the metadata from the mainline gem. In the case of Nokogiri 1.8.0, those differ:
* [nokogiri 1.8.0](https://rubygems.org/gems/nokogiri/versions/1.8.0) requires Ruby `>= 2.1.0`
* [nokogiri 1.8.0 (x86-mingw32)](https://rubygems.org/gems/nokogiri/versions/1.8.0-x86-mingw32) requires Ruby `< 2.5, >= 2.2 `
* [nokogiri 1.8.0 (x64-mingw32)](https://rubygems.org/gems/nokogiri/versions/1.8.0-x64-mingw32) requires Ruby `< 2.5, >= 2.2 `

**Caveat:** Using the `nokogiri` example, the bug cannot be reproduced on Ruby `>= 2.2.0`. This does *NOT* mean that the bug is not there - it's just the case where the differing Ruby version requirements currently overlap.

segiddins:
Please share the output of `bundle env`

[...]

PR #6738, opened 2018-10-12
Make RakeTask spec:deps OS agnostic

leoarnold:
@segiddins encouraged contributions towards support for Windows
https://github.com/bundler/bundler/issues/5992#issuecomment-326809543

As a first step towards this goal this commit introduces safe and
OS agnostic directory traversal in the first pieces of Ruby code
called when setting up a test environment.

See #5992.

File path concatenation explicitly used UNIX style file system separators.

Use `File.join` and `Pathname.join` instead. Define null device explicitly since Ruby 1.8.7 does not know `File::NULL`.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -5,7 +5,8 @@ $:.unshift File.expand_path("../lib", __FILE__)
 require "shellwords"
 require "benchmark"
 
-RUBYGEMS_REPO = if `cd .. && git remote --verbose 2>/dev/null` =~ /rubygems/i
+NULL_DEVICE = (Gem.win_platform? ? "NUL" : "/dev/null")
+RUBYGEMS_REPO = if `git -C "#{File.expand_path("..")}" remote --verbose 2> #{NULL_DEVICE}` =~ /rubygems/i
   File.expand_path("..")
 else
   File.expand_path("tmp/rubygems")
--- a/spec/support/rubygems_ext.rb
+++ b/spec/support/rubygems_ext.rb
@@ -33,17 +33,17 @@ module Spec
 
       ENV["BUNDLE_PATH"] = nil
       ENV["GEM_HOME"] = ENV["GEM_PATH"] = Path.base_system_gems.to_s
-      ENV["PATH"] = [Path.bindir, "#{Path.system_gem_path}/bin", ENV["PATH"]].join(File::PATH_SEPARATOR)
+      ENV["PATH"] = [Path.bindir, Path.system_gem_path.join("bin"), ENV["PATH"]].join(File::PATH_SEPARATOR)
 
       manifest = DEPS.to_a.sort_by(&:first).map {|k, v| "#{k} => #{v}\n" }
-      manifest_path = "#{Path.base_system_gems}/manifest.txt"
+      manifest_path = Path.base_system_gems.join("manifest.txt")
       # it's OK if there are extra gems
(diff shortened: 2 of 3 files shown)


leoarnold:
This contribution has been tested on Windows using AppVeyor:
https://ci.appveyor.com/project/leoarnold/bundler/builds/19449495

Coincidentally, Travis CI just now announced Windows support:
https://blog.travis-ci.com/2018-10-11-windows-early-release

deivid-rodriguez:
@leoarnold Thanks for this! I have a question regarding Windows and path separators. I used to think that best practice was `File.join` indeed, but recently I discovered that ther's apparently no prove that `/` doesn't work on Windows. See https://github.com/rubocop-hq/rubocop/issues/5823, for example.

Does it not work without the `File.join` changes. I'm not saying we shouldn't do it, but if the discussion in the above link is correct, it would only be a matter of style.

spec/support/rubygems_ext.rb
```diff
@@ -33,17 +33,17 @@ def self.setup
...
+      manifest_path = Path.base_system_gems.join("manifest.txt")
       # it's OK if there are extra gems
-      if !File.exist?(manifest_path) || !(manifest - File.readlines(manifest_path)).empty?
+      if !manifest_path.file? || !(manifest - manifest_path.readlines).empty?
         FileUtils.rm_rf(Path.base_system_gems)
         FileUtils.mkdir_p(Path.base_system_gems)
         puts "installing gems for the tests to use..."
         install_gems(DEPS)
-        File.open(manifest_path, "w") {|f| f << manifest.join }
+        File.open(manifest_path.to_s, "w") {|f| f << manifest.join }
```
segiddins: this can now use `Pathname#open`

Rakefile
```diff
@@ -1,14 +1,15 @@
...
 
-$:.unshift File.expand_path("../lib", __FILE__)
+$:.unshift File.expand_path(File.join("..", "lib"), __FILE__)
 require "shellwords"
 require "benchmark"
 
-RUBYGEMS_REPO = if `cd .. && git remote --verbose 2>/dev/null` =~ /rubygems/i
-  File.expand_path("..")
+NULL_DEVICE = (Gem.win_platform? ? "NUL" : "/dev/null")
+RUBYGEMS_REPO = if `git -C #{File.expand_path(File.join("..", ".."), __FILE__)} remote --verbose 2> #{NULL_DEVICE}` =~ /rubygems/i
```
segiddins: I think we need to now quote the path, since it could contain spaces

Rakefile
```diff
@@ -1,14 +1,15 @@
...
+  File.expand_path(File.join("..", ".."), __FILE__)
 else
-  File.expand_path("tmp/rubygems")
+  File.expand_path(File.join("..", "tmp", "rubygems"), __FILE__)
```
segiddins: expanding relative to `__FILE__` in a `Rakefile` isn't necessary, rake chdirs to the directory containing the rakefile

leoarnold:
@deivid-rodriguez Great hint. I was able to confirm this on Windows 10 - even using Ruby 1.8.7. I will adapt the PR.

leoarnold:
@segiddins I added your suggestions to the PR. Thank you for reviewing.

colby-swandale:
@bundlerbot r+

segiddins:
@bundlerbot r+

PR #6740, opened 2018-10-16
Removed win32-open3 workaround.

hsbt:

`win32-open3` gem was required Ruby 1.8 with Windows environment. Because RubyInstaller2 only provides the version of Ruby 2.1+.

Removed needless workaround.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -34,17 +34,6 @@ end
 
 require "bundler"
 
-# Require the correct version of popen for the current platform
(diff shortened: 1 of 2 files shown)


hsbt:
See original commit. https://github.com/ruby/ruby/commit/f35fb6d36c3218988b17dbeb4412922c23745f0e

colby-swandale:
Great commit! it's nice to be able to cleanup 1.8

@bundlerbot r+

PR #6741, opened 2018-10-16
Support Windows platform with ruby core repository.

hsbt:

In Windows environment, ':' is always contained path variable.

See our original commit and its message 

https://github.com/ruby/ruby/commit/f35fb6d36c3218988b17dbeb4412922c23745f0e


colby-swandale:
Thanks! @bundlerbot r+

hsbt:
@bundlerbot r+

PR #6742, opened 2018-10-16
Use helper method instead of hard-code path.

hsbt:

In ruby core repository, We need to replace executable file like `ruby`, `rake`, `gem`. 

This pull request makes hard-coded executable file to the helper methods.


hsbt:
After that, I update `rake` helper method supporting ruby core repository in another pull request.

colby-swandale:
@bundlerbot r+

hsbt:
Thanks!

PR #6743, opened 2018-10-16
Support file structure of ruby core repository.

hsbt:

In the ruby core repository, I put bundler executable and bundler libraries under the `bin` and `lib` directories. It breaks the current behavior.

Support the structure of ruby core repository.

### DIFF
--- a/bundler.gemspec
+++ b/bundler.gemspec
@@ -1,12 +1,12 @@
 # coding: utf-8
 # frozen_string_literal: true
 
-version = File.expand_path("../lib/bundler/version", __FILE__)
+version = File.expand_path("../lib/bundler/version.rb", __FILE__)
 if File.file?(version)
   require version
 else
   # for Ruby core repository
-  require File.expand_path("../bundler/version", __FILE__)
+  require File.expand_path("../bundler/version.rb", __FILE__)
 end
 
 require "shellwords"
--- a/bundler.gemspec
+++ b/bundler.gemspec
@@ -1,7 +1,13 @@
 # coding: utf-8
 # frozen_string_literal: true
 
-require File.expand_path("../lib/bundler/version", __FILE__)
+begin
+  require File.expand_path("../lib/bundler/version", __FILE__)
+rescue LoadError
+  # for Ruby core repository
+  require File.expand_path("../bundler/version", __FILE__)
+end
+
 require "shellwords"
 
 Gem::Specification.new do |s|
--- a/bundler.gemspec
+++ b/bundler.gemspec
@@ -1,9 +1,10 @@
 # coding: utf-8
 # frozen_string_literal: true
 
-begin
-  require File.expand_path("../lib/bundler/version", __FILE__)
-rescue LoadError
+version = File.expand_path("../lib/bundler/version", __FILE__)
+if File.file?(version)
+  require version
+else
   # for Ruby core repository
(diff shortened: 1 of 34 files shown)


hsbt:
I will add another commits later.

bundler.gemspec
```diff
@@ -1,7 +1,13 @@
 # coding: utf-8
 # frozen_string_literal: true
 
-require File.expand_path("../lib/bundler/version", __FILE__)
+begin
+  require File.expand_path("../lib/bundler/version", __FILE__)
+rescue LoadError
+  # for Ruby core repository
+  require File.expand_path("../bundler/version", __FILE__)
```
colby-swandale: Is there a way to avoid this? Using a `rescue` to support Ruby core does not seem nice.

hsbt:
I finished to fix or skip failing examples on ruby core repository.  the all of examples is working fine on ruby core repository with https://github.com/ruby/ruby/pull/1991.

bundler.gemspec
segiddins: we could use `File.file?`

spec/spec_helper.rb
```diff
@@ -107,6 +114,13 @@ module Gem
     mocks.allow_message_expectations_on_nil = false
   end
 
+  config.before :suite do
+    if ENV["BUNDLE_RUBY"]
+      @orig_ruby = Gem.ruby
```
segiddins: could we use `around` to avoid needing an ivar?

spec/bundler/ssl_certs/certificate_manager_spec.rb
```diff
@@ -11,13 +11,18 @@
 
   # Pretend bundler root is rubygems root
   before do
+    # Backing up rubygems ceriticates
+    FileUtils.mv(rubygems_certs_dir, rubygems_certs_dir + ".back") if ENV["BUNDLE_RUBY"] && ENV["BUNDLE_GEM"]
```
segiddins: can this use the helper method for the conditional?

spec/support/helpers.rb
```diff
@@ -212,7 +212,8 @@ def gem_command(command, args = "", options = {})
         args = args.gsub(/(?=")/, "\\")
         args = %("#{args}")
       end
-      sys_exec("#{Gem.ruby} -rrubygems -S gem --backtrace #{command} #{args}")
+      gem = ENV["BUNDLE_GEM"] || "#{Gem.ruby} -rrubygems -S gem --backtrace"
```
segiddins: when won't `ruby -rrubygems -S gem` invoke ruby gems?

hsbt: `ENV["BUNDLE_GEM"]` was stored `$(srcdir)/bin/gem` from ruby source repository. In ruby core repository, It needs to works working directory without source directory.

bundler.gemspec
hsbt: `lib/bundler/version` string was used the other example directory. This string can be only once in gemspec.

hsbt: see https://github.com/bundler/bundler/blob/master/spec/runtime/setup_spec.rb#L877

hsbt:
@segiddins @colby-swandale Do you have additional comments or feedback? I will merge this a few days later.

hsbt:
@bundlerbot r+

PR #6745, opened 2018-10-17
Bump rake version

Schwad:

We are working to [effectively deprecate `rubyforge_project` in RubyGems](https://github.com/rubygems/rubygems/pull/2436) and remove other extraneous references to the attribute. This caused tests tied to bundler to fail.

Ultimately the failure can be found here: https://travis-ci.org/rubygems/rubygems/jobs/442597902

Due to a reference in the `rake` gem to `rubyforge_project`, and the spec treating a deprecation notice as an error and a failure.

The `rake` gem removed the last of its references to `rubyforge_project` by [December 2016](https://github.com/ruby/rake/search?q=rubyforge&type=Commits), so this update points the development dependency to the nearest minor upgrade to `rake` after those changes.

I am not absolutely certain that this is the exact final wrinkle needed to fix this bug and pass this test, but it made the most sense to me. In the stacktrace the `10.x` rake version was referenced, which matched with the development dependency, which would cause this failure. Look forward to thoughts and feedback. :)

### DIFF
--- a/bundler.gemspec
+++ b/bundler.gemspec
@@ -38,7 +38,7 @@ Gem::Specification.new do |s|
 
   s.add_development_dependency "automatiek", "~> 0.1.0"
   s.add_development_dependency "mustache",   "0.99.6"
-  s.add_development_dependency "rake",       "~> 10.0"
+  s.add_development_dependency "rake",       "~> 11.1"
   s.add_development_dependency "rdiscount",  "~> 2.2"
   s.add_development_dependency "ronn",       "~> 0.7.3"
   s.add_development_dependency "rspec",      "~> 3.6"


colby-swandale:
We can't upgrade to the latest version of Rake because we're still testing changes against Ruby 1.8.7. But we're close to releasing Bundler 2 that will remove 1.8.7. We can leave this open until then.

Schwad:
@colby-swandale thanks for the information! I referenced this issue over in [rubygems](https://github.com/rubygems/rubygems/pull/2436#discussion_r227926006) and what I think I'll do is keep from 'officially' executing `deprecate` on `rubyforge_project` until Bundler 2.0 is out like you say. 

Then once we have 2.0 release, this merged, I'll put that PR forward and hopefully we'll be all square. Thanks!

Schwad:
@colby-swandale would this be a candidate for merging into the `2-0-stable` branch?

colby-swandale:
This will eventually be merged into in a stable branch when we decide to release it.

hsbt:
I update its version on https://github.com/bundler/bundler/pull/6923 and https://github.com/bundler/bundler/pull/6924

Schwad:
Great work, thanks! 🙏

PR #6747, opened 2018-10-18
update ruby to the latest versions in travis-ci

colby-swandale:
### Overview

New versions of all the currently supported versions of Ruby we're released today. This PR is just updating .travis.yml to use these new versions of Ruby.


colby-swandale:
@bundlerbot r+

PR #6748, opened 2018-10-18
Enable gemfile flag for bundle lock

### LINKED ISSUE #6735: bundle lock should accept most of the options of bundle install
Such as --gemfile and also --binstubs, --force, --with, --without

jekuta:

`bundle lock` does not accept the `--gemfile` flag that allows to define a custom Gemfile.

The option is not defined as available.

Adds the option to the list available options for `lock` command. 

I wasn't sure though, in which order to display this the new option. I just added it where it felt the most natural to me, but please let me know if there is a better order.

Also do you think it would make sense to add this flag to the other commands as well? 
For example: `add`, `remove`, `console`, `outdated`, `clean`?
For now it is available for: `check`, `exec`, `install`, `update`, `package`, `doctor`. 

I don't think there are any other options to implement this.

Closes #6735

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -618,6 +618,8 @@ module Bundler
       "do not attempt to fetch remote gemspecs and use the local gem cache only"
     method_option "print", :type => :boolean, :default => false, :banner =>
       "print the lockfile to STDOUT instead of writing to the file system"
+    method_option "gemfile", :type => :string, :banner =>
+      "Use the specified gemfile instead of Gemfile"
     method_option "lockfile", :type => :string, :default => nil, :banner =>
       "the path the lockfile should be written to"
     method_option "full-index", :type => :boolean, :default => false, :banner =>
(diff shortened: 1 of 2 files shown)


lib/bundler/cli.rb
```diff
@@ -618,6 +618,8 @@ def inject(name, version)
       "do not attempt to fetch remote gemspecs and use the local gem cache only"
     method_option "print", :type => :boolean, :default => false, :banner =>
       "print the lockfile to STDOUT instead of writing to the file system"
+    method_option "gemfile", :type => :string, :banner =>
+      "Use the specified gemfile instead of Gemfile"
```
colby-swandale:
```suggestion
      "Use the specified Gemfile instead of Gemfile"
```

colby-swandale:
@bundlerbot r+

colby-swandale:
> Also do you think it would make sense to add this flag to the other commands as well?

Yes! it would be great if we could make the UI consistent across the commands that operate on the Gemfile

PR #6749, opened 2018-10-19
Add local git repository source option (`--local_git`) to plugin installation

### LINKED ISSUE #5446: Add flag `--ignore-cache` to `plugin install`
Similar to #5445 in that I ran into some developer annoyance when developing a plugin, where changes weren't being shown because `bundle plugin install` decided that the cached version of the plugin took precedence over the updated remote copy. If there were a flag like `--ignore-cache` that would force a load of the remote version, I could've saved a couple "wait, did I not save the file?" moments.

indirect:
Reopening #6338 to close #5446.


indirect:
@bundlerbot r+

PR #6751, opened 2018-10-20
Fix BUNDLE_PATH_RELATIVE_TO_CWD env variable name

deivid-rodriguez:

The problem was that an environment variable in the docs does not work.

My diagnosis was that its name is incorrect.

My fix is to fix the typo.

I chose this fix because it's the only non absurd fix.


deivid-rodriguez:
It sounds like Github needs a break :)

greysteil:
GitHub isn't sending webhooks, so we won't be able to use bundlerbot for the moment.

deivid-rodriguez:
Yeah, I guess everyone has noticed by now :sweat_smile:. It's amazing how tightly we depend on certain services :open_mouth:.

deivid-rodriguez:
@bundlerbot r=greysteil

PR #6757, opened 2018-10-24
Fix relative_to_original_destination_root

mgrachev:

When I create a gem with the command `bundle gem`, I see the wrong output.
For example:
```bash
$ bundle gem test-gem
Creating gem 'test-gem'...
MIT License enabled in config
      create  est-gem.Gemfile
      create  est-gem.lib.test.gem.rb
      create  est-gem.lib.test.gem.version.rb
      create  est-gem.test-gem.gemspec
      create  est-gem.Rakefile
      create  est-gem.README.md
      create  est-gem.bin.console
      create  est-gem.bin.setup
      create  est-gem..gitignore
      create  est-gem.LICENSE.txt
Initializing git repo in /test-gem
Gem 'test-gem' was successfully created. For more information on making a RubyGem visit https://bundler.io/guides/creating_gem.html
```

The problem in the wrong file name and path of new gem files.

I updated the method `relative_to_original_destination_root`.
I took as a basis this commit https://github.com/erikhuda/thor/commit/506950f59245ff0db3c7d6589fbfbd182cb20f0c#diff-f8ad773f327e75ac0dae88f03b0098d1 and I changed it a little.

### DIFF
--- a/lib/bundler/vendor/thor/lib/thor/actions.rb
+++ b/lib/bundler/vendor/thor/lib/thor/actions.rb
@@ -113,9 +113,11 @@ class Bundler::Thor
     # the script started).
     #
     def relative_to_original_destination_root(path, remove_dot = true)
-      path = path.dup
-      if path.gsub!(@destination_stack[0], ".")
-        remove_dot ? (path[2..-1] || "") : path
+      root = @destination_stack[0]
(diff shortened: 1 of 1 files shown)


colby-swandale:
Thanks for looking into this issue but this PR needs to be made against the [thor](https://github.com/erikhuda/thor) library. Bundler vendors the library into it's src and changes would need to be made against thor first.

Issue #6760, opened 2018-10-25
Bundler 1.17.0 fails if no path set


bpollock-vida:
Trying to install my Gemfile with bundler 1.17.0 always fails, but downgrading bundler to 1.16.1 succeeds.

Also the recommended resolution is not in proper format:
 `gem install CFPropertyList -v '3.0.0' --source 'https://rubygems.org/'
does not work, but 
 `gem install CFPropertyList --source 'https://rubygems.org/' -v '3.0.0' 
does.

### Gemfile
source "https://rubygems.org"

gem 'fastlane', '~> 2.104.0'
gem 'earlgrey', '~> 0.1.5'
gem 'cocoapods', '~> 1.5.3'
gem 'versionomy', '~> 0.5.0'

### Output

Brices-MacBook-Pro% bundle install
Fetching gem metadata from https://rubygems.org/.........
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies...
Using CFPropertyList 3.0.0
ArgumentError: comparison of Pathname with String failed
An error occurred while installing CFPropertyList (3.0.0), and Bundler cannot continue.
Make sure that `gem install CFPropertyList -v '3.0.0' --source 'https://rubygems.org/'` succeeds before bundling.

In Gemfile:
  cocoapods was resolved to 1.5.3, which depends on
    xcodeproj was resolved to 1.7.0, which depends on
      CFPropertyList

bpollock-vida:
thanks for the speedy fix!

PR #6761, opened 2018-10-25
fix error with path objects in array

alexggordon:

The problem was that `Pathname` instances in an array can not be sorted when there are string instances in the array also. Because of this, calling `.sort` before `.to_s` resulted in the error

```
[!] There was an error parsing `Gemfile`: comparison of Pathname with String failed. Bundler cannot continue.
```

you can easily see this issue doing
```
> require 'rails'
=> true
> [Pathname.new("/tmp/bundle")]
=> [#<Pathname:/tmp/bundle>]
> [Pathname.new("/tmp/bundle"), "test"]
=> [#<Pathname:/tmp/bundle>, "test"]
> [Pathname.new("/tmp/bundle"), "test"].sort
ArgumentError: comparison of Pathname with String failed
```

`sort` was called before `map` in the warn message. 

We should call `map(&:to_s)` before calling sort, and add a test case for this scenario. 

Because it broke our production deploys.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -371,7 +371,7 @@ EOF
         unwritable_files = files.reject {|f| File.writable?(f) }
         sudo_needed = !unwritable_files.empty?
         if sudo_needed
-          Bundler.ui.warn "Following files may not be writable, so sudo is needed:\n  #{unwritable_files.sort.map(&:to_s).join("\n  ")}"
+          Bundler.ui.warn "Following files may not be writable, so sudo is needed:\n  #{unwritable_files.map(&:to_s).sort.join("\n  ")}"
         end
       end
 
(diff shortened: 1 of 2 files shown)


alexggordon:
This also fixes #6760

greysteil:
@bundlerbot r+

alexggordon:
You're welcome for the speedy fix! Any chance of this fix getting backported into 17.0 or getting a patch release pretty soon?

greysteil:
@colby-swandale is probably the best person to ask on that one, but we've had a few people run into this already so I wouldn't be opposed to a patch release in the next week.

mateuscruz:
This broke our deploy pipeline earlier today. It took me 6 hours to identify the source and normalize all of our environments. I had to rollback to bundler 1.16. Thanks for the quick fix!

greysteil:
Sorry for the bug @mateuscruz! I can't promise a quick release, but I'll talk to the team and we'll do our best. We'll give it a few days to hoover up any other bugs either way.

mateuscruz:
No worries, @greysteil . It was entirely our fault. We didn't have a locked version of bundler and it would update at every deploy. We've locked it to 1.16 for the time being. Keep up the good work.

greysteil:
@alexggordon @mateuscruz - this is now released in 1.17.1. Apologies for the regression (and all credit to @indirect for the incredibly quick release!).

PR #6763, opened 2018-10-26
Removed jobs with old ruby versions from Travis CI.

hsbt:
I removed the ruby versions under the Ruby 2.2. We need to fix gemspec and other parameters of .travis.yml. But this is the first step of Bundler-2.0.0 reboot.

PR #6764, opened 2018-10-26
Remove unnecessary stuff from specs

deivid-rodriguez:
I wrote this patch a while ago but I didn't propose it because I didn't think it would be accepted. Now I think we're ready.

The problem was that the specs have a lot of conditional behavior on `Gem::VERSION`.

My diagnosis was that this was necessary because of testing against a big range of rubygems versions on the latest stable branch.

My fix is to remove all the conditionals since we'll no longer test against rubygems versions that old on the latest stable, once #6763 lands.

I chose this fix because it makes test code simpler.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -51,17 +51,10 @@ namespace :spec do
       deps.delete("rdiscount")
     end
 
-    if Gem::VERSION < "2.0.0"
-      deps.sort_by {|name, _| name }.map do |name, version|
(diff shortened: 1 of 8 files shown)


colby-swandale:
@bundlerbot r+

deivid-rodriguez:
@colby-swandale I forgot to note that this is not only removing compatibility code from the specs but also from the default gem template.

Issue #6768, opened 2018-10-29
`bundle remove` error


dduugg:
Given a simple Gemfile consisting of:
```ruby
gem "rack"
group :test do
  gem "coffee-script-source"
end
```
When invoking `bundle remove rack`, the expected result is that the Gemfile looks like:
```ruby
group :test do
  gem "coffee-script-source"
end
```
The actual result is an error:
```
$ bundle remove rack
Removing gems from /path/to/Gemfile

[!] There was an error parsing `Gemfile`: syntax error, unexpected end-of-input - group :test do
              ^. Bundler cannot continue.

 #  from /Users/duggiestyle/Projects/bundler/Gemfile:1
 #  -------------------------------------------
 >  group :test do
 #  -------------------------------------------
```

PR #6769, opened 2018-10-29
Fix `remove` when block method appears in gem name

dduugg:

https://github.com/bundler/bundler/issues/6768

`remove_nested_blocks` considers any appearance of text matching a block method (`group source env install_if`) to be an occurrence of a nested block.

`remove_nested_blocks` should only reduce the scope of where a nested block method can occur.

Another approach would use word delimiters rather than `starts_with?` to find nested block methods, but this is probably fine.

### DIFF
--- a/lib/bundler/injector.rb
+++ b/lib/bundler/injector.rb
@@ -123,7 +123,7 @@ module Bundler
       end
     end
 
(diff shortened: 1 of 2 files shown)


segiddins:
Thanks!
@bundlerbot r+

PR #6770, opened 2018-10-31
delete merged branches automatically

indirect:
Use a fancy new Bors feature to delete merged branches automatically, hooray.


hsbt:
@bundlerbot r+

PR #6773, opened 2018-11-06
Fix indentation of bundler executable

meganemura:

`bundle binstubs bundler` generates the file like followings.

```ruby
require "rubygems"

m = Module.new do
    module_function

  def invoked_as_script?
    File.expand_path($0) == File.expand_path(__FILE__)
  end
(snip)
end
```

My diagnosis was...

the template's `module_function` line is not match to other lines.

I fix the template.

no other options.

### DIFF
--- a/lib/bundler/templates/Executable.bundler
+++ b/lib/bundler/templates/Executable.bundler
@@ -11,7 +11,7 @@
 require "rubygems"
 
 m = Module.new do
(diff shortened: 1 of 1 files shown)


segiddins:
Thanks!

@bundlerbot r+

segiddins:
@bundlerbot retry

PR #6774, opened 2018-11-06
[Definition] Dont pin path deps to newer versions

### LINKED ISSUE #6750: ONLY_UPDATE_TO_NEWER_VERSIONS also affects `bundle install` when `path:` is used
When you have a `path:` dependency in your Gemfile, checkout a reference of the dependency older that the one previously bundled, and run `bundle install`, you get an error like the following

```
$ BUNDLE_ONLY_UPDATE_TO_NEWER_VERSIONS=true bundle install
Could not find gem 'rake' in source at `../../../rake`.
The source contains 'rake' at: 12.3.0
```

I used the environment variable so it reproduces on 1.x, but this is the default behavior on master.

This is pretty likely to happen during a bisection where you jump references up and down and run a bundled process on each of them.

## Environment

```
Bundler       1.16.6
  Platforms   ruby, x86_64-linux
Ruby          2.5.1p57 (2018-03-29 revision 63029) [x86_64-linux]
  Full Path   /home/deivid/.rbenv/versions/2.5.1/bin/ruby
  Config Dir  /home/deivid/.rbenv/versions/2.5.1/etc
RubyGems      2.7.7
  Gem Home    /home/deivid/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0
  Gem Path    /home/deivid/.gem/ruby/2.5.0:/home/deivid/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0
  User Path   /home/deivid/.gem/ruby/2.5.0
  Bin Dir     /home/deivid/.rbenv/versions/2.5.1/bin
Tools         
  Git         2.19.1
  RVM         not installed
  rbenv       rbenv 1.1.1
  chruby      not installed
Gem.ruby      /home/deivid/.rbenv/versions/2.5.1/bin/ruby
bundle #!     /home/deivid/.local/src/rbenv-1.1.1/versions/2.5.1/bin/ruby

```

## Bundler Build Metadata

```
Built At          2018-10-05
Git SHA           f66c33467
Released Version  true
```

## Bundler settings

```
gem.test
  Set for the current user (/home/deivid/.bundle/config): "rspec"
gem.mit
  Set for the current user (/home/deivid/.bundle/config): true
gem.coc
  Set for the current user (/home/deivid/.bundle/config): true
```

### Gemfile

```ruby
source "https://rubygems.org"

gem "rake", path: "~/Code/rake"
```

### Gemfile.lock

```
PATH
  remote: ../../../rake
  specs:
    rake (12.3.1)

GEM
  remote: https://rubygems.org/
  specs:

PLATFORMS
  ruby

DEPENDENCIES
  rake!

BUNDLED WITH
   1.16.6
```

segiddins:

The problem was setting `only_update_to_newer_versions` would break using path gems.

Closes #6750

My diagnosis was path gems should _always_ get their version used, since there's 0 alternative versions from a path source.

My fix should circuits the list of additional requirements when the dependency's source is a path source.

I chose this fix because it solves the problem in the same place that has the feature flag conditional

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -977,7 +977,9 @@ module Bundler
       dependencies_by_name = dependencies.inject({}) {|memo, dep| memo.update(dep.name => dep) }
       @locked_gems.specs.reduce({}) do |requirements, locked_spec|
         name = locked_spec.name
-        next requirements if @locked_gems.dependencies[name] != dependencies_by_name[name]
+        dependency = dependencies_by_name[name]
+        next requirements if @locked_gems.dependencies[name] != dependency
+        next requirements if dependency && dependency.source.is_a?(Source::Path)
         dep = Gem::Dependency.new(name, ">= #{locked_spec.version}")
         requirements[name] = DepProxy.new(dep, locked_spec.platform)
         requirements
--- a/spec/install/gemfile/gemspec_spec.rb
+++ b/spec/install/gemfile/gemspec_spec.rb
@@ -263,6 +263,20 @@ RSpec.describe "bundle install from an existing gemspec" do
     expect(out).to eq("WIN")
   end
 
+  it "works with only_update_to_newer_versions" do
+    build_lib "omg", "2.0", :path => lib_path("omg")
+
+    install_gemfile <<-G
+      gemspec :path => "#{lib_path("omg")}"
+    G
+
+    build_lib "omg", "1.0", :path => lib_path("omg")
+
+    bundle! :install, :env => { "BUNDLE_BUNDLE_ONLY_UPDATE_TO_NEWER_VERSIONS" => "true" }
+
+    expect(the_bundle).to include_gems "omg 1.0"
+  end
+
   context "in deployment mode" do
     context "when the lockfile was not updated after a change to the gemspec's dependencies" do
       it "reports that installation failed" do
--- a/spec/install/gemfile/path_spec.rb
+++ b/spec/install/gemfile/path_spec.rb
@@ -132,6 +132,20 @@ RSpec.describe "bundle install with explicit source paths" do
     expect(the_bundle).to include_gems "foo 1.0"
   end
 
+  it "works with only_update_to_newer_versions" do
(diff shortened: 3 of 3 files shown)


lib/bundler/definition.rb
```diff
@@ -978,6 +978,7 @@ def additional_base_requirements_for_resolve
       @locked_gems.specs.reduce({}) do |requirements, locked_spec|
         name = locked_spec.name
         next requirements if @locked_gems.dependencies[name] != dependencies_by_name[name]
+        next requirements if dependencies_by_name[name].source.is_a?(Source::Path)
```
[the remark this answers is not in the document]
segiddins: Ugh you’re right

segiddins:
This is green, @greysteil please take another look?

lib/bundler/definition.rb
```diff
@@ -977,7 +977,9 @@ def additional_base_requirements_for_resolve
       dependencies_by_name = dependencies.inject({}) {|memo, dep| memo.update(dep.name => dep) }
       @locked_gems.specs.reduce({}) do |requirements, locked_spec|
         name = locked_spec.name
-        next requirements if @locked_gems.dependencies[name] != dependencies_by_name[name]
+        dependency = dependencies_by_name[name]
+        next requirements if @locked_gems.dependencies[name] != dependency
+        next requirements if dependency && dependency.source.is_a?(Source::Path)
```
greysteil: I think we still need to make this an `instance_of?` rather than a `is_a?` to avoid affecting `Source::Gemspec` and `Source::Git` dependencies.

segiddins: Why though? Those subclasses also dont have any way to resolve versions

greysteil: You're totally right - in fact we *want* downgrades of gemspec dependencies and changing this would break them. Have added a test (broken on master, fixed on this branch)

segiddins: Thanks for adding a test!

segiddins:
@bundlerbot r+

PR #6775, opened 2018-11-06
[Plugin::Index] Only register each plugin once for a given hook

### LINKED ISSUE #6771: bundle plugin install appends to the list of hooks instead of replacing the originals
### What you're trying to accomplish
Install a custom bundler plugin that adds an `after-install` hook.
### The command you ran
`bundle plugin install my_plugin --git file://$HOME/my_plugin`
`bundle plugin install my_plugin --git file://$HOME/my_plugin`
### What you expected to happen
The registered hook to be replaced on a second install.
### What actually happened
The hook was appended to the list of hooks instead of replaced.  This is evidenced by the `~/.bundle/plugin/index` file.  This causes the hook to run multiple times after a gem is installed.  Obviously this applies to any hook, not just `after-install`.
```
---
commands:
hooks:
  after-install:
  - "my_plugin"
  - "my_plugin"
load_paths:
  my_plugin:
  - "/home/matt/.bundle/plugin/bundler/gems/my_plugin-1a2941af577c/lib"
plugin_paths:
  my_plugin: "/home/matt/.bundle/plugin/bundler/gems/my_plugin-1a2941af577c"
sources:
```
The hook was added following the directions at [How to write a Bundler plugin](https://bundler.io/v1.17/guides/bundler_plugins.html).  My `plugins.rb`:
```
require 'my_plugin/hooks'

Bundler::Plugin.add_hook(
  'after-install',
  &My_Plugin.hooks.method(:after_install))
```
### Environment
```
Bundler       1.17.1
  Platforms   ruby, x86_64-linux
Ruby          2.5.3p105 (2018-10-18 revision 65156) [x86_64-linux-gnu]
  Full Path   /opt/ruby/2.5.3/bin/ruby
  Config Dir  /opt/ruby/2.5.3/etc
RubyGems      2.7.6
  Gem Home    /opt/ruby/2.5.3/lib/ruby/gems/2.5.0
  Gem Path    /home/matt/.gem/ruby/2.5.0:/opt/ruby/2.5.3/lib/ruby/gems/2.5.0
  User Path   /home/matt/.gem/ruby/2.5.0
  Bin Dir     /opt/ruby/2.5.3/bin
Tools         
  Git         2.18.0
  RVM         not installed
  rbenv       not installed
  chruby      not installed
```
### Bundler Build Metadata
```
Built At          2018-10-25
Git SHA           2760d72d3
Released Version  true
```

segiddins:

The problem was running `plugin install` twice for a plugin withs hooks would cause that hook to be registered twice.

Closes #6771.

My diagnosis was a plugin's hooks should only be run once per event.

My fix is to `uniq` the list of plugins registered for each event.

### DIFF
--- a/spec/bundler/plugin/index_spec.rb
+++ b/spec/bundler/plugin/index_spec.rb
@@ -88,7 +88,12 @@ RSpec.describe Bundler::Plugin::Index do
 
     it "only registers a gem once for an event" do
       path = lib_path(plugin_name)
-      index.register_plugin(plugin_name, path.to_s, [path.join("lib").to_s], commands, sources, hooks)
+      index.register_plugin(plugin_name,
+                            path.to_s,
+                            [path.join("lib").to_s],
+                            commands,
+                            sources,
+                            hooks + hooks)
       expect(index.hook_plugins("after-bar")).to eq([plugin_name])
     end
 
--- a/spec/bundler/plugin/index_spec.rb
+++ b/spec/bundler/plugin/index_spec.rb
@@ -89,11 +89,11 @@ RSpec.describe Bundler::Plugin::Index do
     it "only registers a gem once for an event" do
       path = lib_path(plugin_name)
       index.register_plugin(plugin_name,
-                            path.to_s,
-                            [path.join("lib").to_s],
-                            commands,
-                            sources,
-                            hooks + hooks)
+        path.to_s,
+        [path.join("lib").to_s],
+        commands,
+        sources,
+        hooks + hooks)
       expect(index.hook_plugins("after-bar")).to eq([plugin_name])
     end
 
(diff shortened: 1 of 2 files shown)


spec/bundler/plugin/index_spec.rb
```diff
@@ -86,6 +86,12 @@
       expect(new_index.hook_plugins("after-bar")).to eq([plugin_name])
     end
 
+    it "only registers a gem once for an event" do
+      path = lib_path(plugin_name)
+      index.register_plugin(plugin_name, path.to_s, [path.join("lib").to_s], commands, sources, hooks)
```
olleolleolle: As a way of expressing the redundancy of the hooks listed, the `hooks` argument could be doubled:

```suggestion
      index.register_plugin(plugin_name,
                            path.to_s,
                            [path.join("lib").to_s],
                            commands,
                            sources,
                            hooks + hooks)
```

colby-swandale:
@bundlerbot r+

PR #6786, opened 2018-11-16
Followed up #6743

hsbt:

#6743 was failed with ruby core repository and ignored working example.

I fixed them.


hsbt:
@bundlerbot r+

PR #6788, opened 2018-11-16
Fix multiple source warning message from `error` to `warn`

jlw:

I've been working on a few repos for most of this year that use a privately-hosted gem with the same name (but completely different purpose) as a gem hosted at rubygems.org - whenever I start with a clean copy and `bundle install` for the first time, I am presented with a warning in red that the gem was found in multiple sources. The red text makes me think that something went wrong, even though bundler did exactly what I wanted it to do.

The user experience is wrong - a warning should not be red.

When I looked at the source, I saw immediately that the warning was using `Bundler.ui.error` instead of `Bundler.ui.warn`.

This is a bug fix with a limited lifespan - after setting up a local copy of bundler I see that 2.0's handling of multiple sources A) is smarter and bypasses the error in most of the repos where I saw it with bundler 1.x; and B) fails early with a different error message in the remaining repos.

This fix **doesn't** need to be merged into the bundler 2.0 branch(es).

The fix is simply switching to `Bundler.ui.warn` in this case.

_I didn't see any other possible options._

### DIFF
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -203,14 +203,14 @@ module Bundler
 
     def warn_ambiguous_gems
       Installer.ambiguous_gems.to_a.each do |name, installed_from_uri, *also_found_in_uris|
-        Bundler.ui.error "Warning: the gem '#{name}' was found in multiple sources."
-        Bundler.ui.error "Installed from: #{installed_from_uri}"
-        Bundler.ui.error "Also found in:"
-        also_found_in_uris.each {|uri| Bundler.ui.error "  * #{uri}" }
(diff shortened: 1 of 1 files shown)


jlw:
FYI, the test around this behavior only looks at the content of the output and not the style, so I left this change without any new test coverage.

greysteil:
I'm totally 👍 on this change, but will leave @colby-swandale or others to comment on whether merging directly into `1-17-stable` is the correct approach.

Thanks for the contribution!

jlw:
I initially noticed that the master branch ran as v2, so I verified this change from the 1-17-stable branch without digging any deeper. I see now handling for different version modes in master and am verifying this basic change there now. I'll open a new PR against master.

Issue #6785, opened 2018-11-14
Change `github` to HTTPS in Bundler 2+


indirect:
As part of shipping Bundler 2 with "no brainer" changes that are technically breaking, we should update the `github` Gemfile DSL shortcut to use HTTPS instead of HTTP, and we should remove `git_source(:github) {|repo_name| "https://github.com/#{repo_name}" }` from the Gemfile generated by `bundle init` and generated by `bundle gem foo`.

gavrieltal:
I am looking for an open source project to get involved in, and I was recommended by a friend to start with Bundler. I'll try solving this one.

gavrieltal:
a few questions:

is running `$ bin/rake spec` on a clean fork of the development master supposed to spit out hundreds of errors?

Do you want the relevant code simply commented out, saying that it used to be there, or do you want it outright removed?

Do you also want to remove `git_source(:bitbucket)` and `git_source(:gist)`?

colby-swandale:
@gavrieltal I'm currently working on this issue. You should checkout our [help wanted/good first issue](https://github.com/bundler/bundler/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22+label%3A%22good+first+issue%22) tags that has a list of issues that we think are good first new contributors to get started on.

deivid-rodriguez:
> is running $ bin/rake spec on a clean fork of the development master supposed to spit out hundreds of errors?

If I recall correctly, you need to run `bin/rake spec:deps` first.

gavrieltal:
I got these error messages, don't know what it means entirely:
```Could not find 'rdiscount' (~> 2.2) among 47 total gem(s)
Checked in 'GEM_PATH=/home/gavriel/.gem/ruby/2.5.0:/var/lib/gems/2.5.0:/usr/lib/x86_64-linux-gnu/rubygems-integration/2.5.0:/usr/share/rubygems-integration/2.5.0:/usr/share/rubygems-integration/all', execute `gem env` for more information (Gem::MissingSpecError)
Could not find 'ronn' (~> 0.7.3) among 47 total gem(s)
Checked in 'GEM_PATH=/home/gavriel/.gem/ruby/2.5.0:/var/lib/gems/2.5.0:/usr/lib/x86_64-linux-gnu/rubygems-integration/2.5.0:/usr/share/rubygems-integration/2.5.0:/usr/share/rubygems-integration/all', execute `gem env` for more information (Gem::MissingSpecError)
[...]

PR #6791, opened 2018-11-17
set https to the default protocol for git

colby-swandale:
This PR is setting Git gems to be fetched over `https` by default.

See #6785

### DIFF
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -305,7 +305,7 @@ module Bundler
         #   end
         repo_name = "#{repo_name}/#{repo_name}" unless repo_name.include?("/")
         if Bundler.settings["github.https"] || Bundler.feature_flag.github_https_source?
-          Bundler::SharedHelpers.major_deprecation 2, "The `github.https` setting will ba removed" if Bundler.settings["github.https"]
+          Bundler::SharedHelpers.major_deprecation 2, "The `github.https` setting will be removed" if Bundler.settings["github.https"]
           "https://github.com/#{repo_name}.git"
         else
           "git://github.com/#{repo_name}.git"
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -304,9 +304,8 @@ module Bundler
         #     "https://github.com/#{repo_name}.git"
         #   end
         repo_name = "#{repo_name}/#{repo_name}" unless repo_name.include?("/")
-        # TODO: 2.0 upgrade this setting to the default
-        if Bundler.settings["github.https"]
-          Bundler::SharedHelpers.major_deprecation 2, "The `github.https` setting will be removed"
+        if Bundler.settings["github.https"] || Bundler.feature_flag.github_https_source?
+          Bundler::SharedHelpers.major_deprecation 2, "The `github.https` setting will ba removed" if Bundler.settings["github.https"]
           "https://github.com/#{repo_name}.git"
         else
           "git://github.com/#{repo_name}.git"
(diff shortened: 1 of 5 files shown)


segiddins:
This breaks the 1.x specs

colby-swandale:
Yea, i wasn't really sure what to do about this but a feature flag sounds ideal.

lib/bundler/dsl.rb
```diff
@@ -304,9 +304,8 @@ def add_git_sources
         #     "https://github.com/#{repo_name}.git"
         #   end
         repo_name = "#{repo_name}/#{repo_name}" unless repo_name.include?("/")
-        # TODO: 2.0 upgrade this setting to the default
-        if Bundler.settings["github.https"]
-          Bundler::SharedHelpers.major_deprecation 2, "The `github.https` setting will be removed"
+        if Bundler.settings["github.https"] || Bundler.feature_flag.github_https_source?
+          Bundler::SharedHelpers.major_deprecation 2, "The `github.https` setting will ba removed" if Bundler.settings["github.https"]
```
olleolleolle:
```suggestion
          Bundler::SharedHelpers.major_deprecation 2, "The `github.https` setting will be removed" if Bundler.settings["github.https"]
```

spec/bundler/dsl_spec.rb
```diff
@@ -25,8 +25,18 @@
       expect { subject.git_source(:example) }.to raise_error(Bundler::InvalidOption)
     end
 
-    context "default hosts (git, gist)", :bundler => "< 2" do
+    context "default hosts (git, gist)" do
+      # Note: There is a feature flag that will disabled these soruces by
```
olleolleolle:
```suggestion
      # Note: There is a feature flag that will disable these sources by
```

deivid-rodriguez:
Hi! I haven't looked at the CI failures, but the new feature flag and the setting `github.https` seem redundant to me. Can we instead change the default value of the `github.https` setting in bundler 2 by making it a feature flag and making it have the `bundler_2_mode` value?

Also, the deprecation warning does not makes a lot of sense right now. We're saying that this setting will be deprecated but introducing another one that does the same thing. I think we want to delay the deprecation horizon to bundler 3.

Finally, I see two directions here but I'm not sure which one we're heading to:

* Completely remove the `github:` option. I thought this was what we were doing and instead we were encouraging users to include the shortcuts they need (for `github` or other similar systems) right in the `Gemfile`. This is what https://github.com/bundler/bundler/pull/5728 explains.

* Keep the `github` option but change the default to `https`. 

Are we doing the latter for bundler 2, and still planning the former for later versions, or are we only doing the latter?

colby-swandale:
> Hi! I haven't looked at the CI failures, but the new feature flag and the setting github.https seem redundant to me. Can we instead change the default value of the github.https setting in bundler 2 by making it a feature flag and making it have the bundler_2_mode value?

But `github.https` can't be a feature flag unless we change the config name.

> Also, the deprecation warning does not makes a lot of sense right now. We're saying that this setting will be deprecated but introducing another one that does the same thing. I think we want to delay the deprecation horizon to bundler 3.

Keep in mind that `master` is still accepting pull requests for Bundler 1 and is still only v2. The feature flags won't be bumped to Bundler 3 until after 2.0.0 has been released. When i merge this into `2-0-stable` i will adjust the version being used in the major deprecation.

> Are we doing the latter for bundler 2, and still planning the former for later versions

For Bundler 2 we're using `https` by default, but we're thinking about removing it in a future major release.

segiddins:
We can definitely use github.https as a feature flag name, I think you’ll just need to manually construct the accessor method for it

colby-swandale:
hmm i would prefer avoid doing that but i will adjust it

PR #6793, opened 2018-11-20
Removed exe/bundle_ruby and its example.

hsbt:
Originated from #3489

`exe/bundle_ruby` was already removed from bundler's gemspec. We can remove it now.


hsbt:
Bundler 2 was out now. We can merge this.

hsbt:
@bundlerbot r+

PR #6798, opened 2018-11-22
Use https in the Github shortcut by default in Bundler 2

### LINKED ISSUE #6785: Change `github` to HTTPS in Bundler 2+

colby-swandale:
This PR is setting Git gems to be fetched over https by default.

Note: This PR is successing #6791 

Closes #6785

### DIFF
--- a/lib/bundler/feature_flag.rb
+++ b/lib/bundler/feature_flag.rb
@@ -60,6 +60,8 @@ module Bundler
 
     settings_option(:default_cli_command) { bundler_2_mode? ? :cli_help : :install }
 
+    settings_method(:github_https?, "github.https") { bundler_2_mode? }
+
     def initialize(bundler_version)
       @bundler_version = Gem::Version.create(bundler_version)
     end
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -305,7 +305,7 @@ module Bundler
         #   end
         repo_name = "#{repo_name}/#{repo_name}" unless repo_name.include?("/")
         # TODO: 2.0 upgrade this setting to the default
-        if Bundler.settings["github.https"]
+        if Bundler.feature_flag.github_https?
           Bundler::SharedHelpers.major_deprecation 2, "The `github.https` setting will be removed"
           "https://github.com/#{repo_name}.git"
         else
--- a/spec/bundler/dsl_spec.rb
+++ b/spec/bundler/dsl_spec.rb
@@ -25,7 +25,23 @@ RSpec.describe Bundler::Dsl do
       expect { subject.git_source(:example) }.to raise_error(Bundler::InvalidOption)
     end
 
+    context "github_https feature flag" do
+      it "is true when github.https is true" do
+        bundle "config github.https true"
+        expect(Bundler.feature_flag.github_https?).to eq "true"
+      end
+    end
+
     context "default hosts (git, gist)", :bundler => "< 2" do
+      context "when github.https config is true" do
+        before { bundle "config github.https true" }
+        it "converts :github to :git using https" do
+          subject.gem("sparks", :github => "indirect/sparks")
+          github_uri = "https://github.com/indirect/sparks.git"
+          expect(subject.dependencies.first.source.uri).to eq(github_uri)
+        end
+      end
+
       it "converts :github to :git" do
         subject.gem("sparks", :github => "indirect/sparks")
         github_uri = "git://github.com/indirect/sparks.git"


colby-swandale:
Note that for this to me merged into `master` i couldn't check for `https` outside of the `:bundler => "< 2"` scope. But this can be cleaned up once 2.0.0 is released.

lib/bundler/feature_flag.rb
```diff
@@ -64,6 +65,12 @@ def initialize(bundler_version)
       @bundler_version = Gem::Version.create(bundler_version)
     end
 
+    def github_https(&block)
+      value = Bundler.settings["github.https"]
+      value = instance_eval(&default) if value.nil? && !default.nil?
+      value
+    end
```
deivid-rodriguez: Instead of having a specific reader for this feature flag, we can maybe add general support for "dot settings" and use a similar criteria to the one for translating environment variables into settings and viceversa. Something like

```diff
diff --git a/lib/bundler/feature_flag.rb b/lib/bundler/feature_flag.rb
index 83e7ff038..c39814c34 100644
--- a/lib/bundler/feature_flag.rb
+++ b/lib/bundler/feature_flag.rb
@@ -18,7 +18,7 @@ module Bundler
 
     def self.settings_method(name, key, &default)
       define_method(name) do
-        value = Bundler.settings[key]
+        value = Bundler.settings[key.gsub("__", ".")]
         value = instance_eval(&default) if value.nil? && !default.nil?
         value
       end
@@ -57,6 +57,7 @@ module Bundler
     settings_flag(:update_requires_all_flag) { bundler_2_mode? }
     settings_flag(:use_gem_version_promoter_for_major_updates) { bundler_2_mode? }
     settings_flag(:viz_command) { !bundler_2_mode? }
+    settings_flag(:github__https) { bundler_2_mode? }
 
     settings_option(:default_cli_command) { bundler_2_mode? ? :cli_help : :install }
```

lib/bundler/settings.rb
```diff
@@ -35,6 +35,7 @@ class Settings
       frozen
       gem.coc
       gem.mit
+      github_https
```
deivid-rodriguez: This is the list of settings, not feature flags, so the `.` should be used here, right?

lib/bundler/feature_flag.rb
colby-swandale: I don't want to make this something that is supported in Bundler, this PR is very much an exception to the usual rule.

deivid-rodriguez: Ok @colby-swandale, if you think that's best, it's no big deal.

spec/quality_spec.rb
```diff
@@ -178,6 +178,7 @@ def check_for_specific_pronouns(filename)
       console_command
       deployment_means_frozen
       forget_cli_options
+      github_https
```
deivid-rodriguez: You didn't apply my suggestion of changing `github_https` to `github.https` in the list of settings, and instead removed it from the list. So you should now remove it from the list of exemptions as well.

lib/bundler/feature_flag.rb
```diff
@@ -64,6 +64,12 @@ def initialize(bundler_version)
       @bundler_version = Gem::Version.create(bundler_version)
     end
 
+    def github_https
```
segiddins: `github_https?`

segiddins: This could also probably be 

```ruby
settings_method(:github_https?, "github.https") { bundler_2_mode? }
```

That way you don't have to replicate the logic

colby-swandale: oh awesome, i didn't realize that. Thanks!

lib/bundler/feature_flag.rb
```diff
@@ -60,6 +60,8 @@ def self.settings_method(name, key, &default)
 
     settings_option(:default_cli_command) { bundler_2_mode? ? :cli_help : :install }
 
+    settings_method(:github_https?, "github.https") { bundler_2_mode? }
```
deivid-rodriguez: Of course, it looks great now! :100:

colby-swandale:
@bundlerbot r+

jordanmichaelrushing:
Well that latest merge just screwed us over for about 5 hours. If there's any googlers out there with deployments that are now all of a sudden broken, it's because your source is changing from git to https on deployment and your local was expecting git. So rebuild your gemfile.lock again locally and you should be fine.
Specifically locally I changed any instance of `github:` to `git:`and re bundled. GL.

indirect:
@jordanmichaelrushing Hi.

First, the meta: we have a [code of conduct](https://bundler.io/conduct.html), and the language in your message isn't okay. Please reword your message to be less shitty towards the people who put a huge amount of effort into providing you with Bundler, for free.

Second, to your point about "sudden" breakage: Bundler has been printing out warnings about this change for literally _years_. More than one. Your application may have broken suddenly, in a way that surprised you, but this change in Bundler was not sudden.

PR #6803, opened 2018-11-24
Todo code removal for bundler 2.0

gavrieltal:
more to-do comments heeded


segiddins:
I think all these changes need to come in Bundler 3, @colby-swandale, since we changed the meaning of Bundler 2 :/

colby-swandale:
It seems that this PR is being replaced by #6808

PR #6805, opened 2018-11-24
[Plugin] Ignore the deployment setting

### LINKED ISSUE #6795: Plugin can't be installed through Gemfile when bundler is frozen
Hello guys 👋 ,

We recently decided to create a Bundler plugin to help our many applications at Shopify dualboot their app. #6777 talks more about the "dualbooting" part and why it's useful.

After finishing writing the plugin, I realized that plugins can't be installed while bundler is frozen, which in our case is always true in production (the heroku ruby buildpack runs `bundle install --deployment`). This makes the plugin feature unusable if an app declare `plugin 'some_plugin'` in their Gemfile.

The error is `Unable to find a spec satisfying bootboot (>= 0) in the set. Perhaps the lockfile is corrupted?`, which makes sense since plugins don't get written as part of the Gemfile.lock snapshot.

While I'm not very familiar with Bundler, this seems to be a bug to me as otherwise the `plugin` and the source feature that comes with it is completely useless.
Shall plugins be part of the lockfile snapshot or should we allow their installation when bundler is frozen even when they aren't in the snapshot?

For some context and if you are curious you can find the "bootboot" plugin here https://github.com/shopify/bootboot

segiddins:

The problem was installing a new plugin in deployment mode would fail with a suggestion the lock file was corrupted, when it installed fine without deployment mode.

Fixes #6795.

My diagnosis was deployment mode was interfering with installing a new plugin.

My fix overrides the deployment and frozen settings when creating the plugin Definition.

I chose this fix because having a "frozen" plugin bundle makes no sense, since plugins aren't written to the lockfile.

### DIFF
--- a/lib/bundler/plugin.rb
+++ b/lib/bundler/plugin.rb
@@ -73,20 +73,22 @@ module Bundler
     # @param [Pathname] gemfile path
     # @param [Proc] block that can be evaluated for (inline) Gemfile
     def gemfile_install(gemfile = nil, &inline)
-      builder = DSL.new
-      if block_given?
-        builder.instance_eval(&inline)
-      else
-        builder.eval_gemfile(gemfile)
+      Bundler.settings.temporary(:frozen => false, :deployment => false) do
+        builder = DSL.new
+        if block_given?
+          builder.instance_eval(&inline)
+        else
+          builder.eval_gemfile(gemfile)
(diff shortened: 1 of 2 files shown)


colby-swandale:
Awesome!

@bundlerbot r+

PR #6806, opened 2018-11-24
Todo remove require for Bundler 2.0

gavrieltal:


colby-swandale:
Thanks!

@bundlerbot r+

colby-swandale:
Is this replacing #6803 ?

gavrieltal:
It looks like there's one more todo that's explicitly Ruby < 1.9 related that I can figure out (as a newbie). Yes, it's basically a replacement of #6803.

PR #6808, opened 2018-11-25
deleting portion of code supporting Ruby < 1.9 for Bundler 2.0

gavrieltal:
I think this should be the last of the intended (valid) changes from #6803


gavrieltal:
The only failure here is for Ruby 1.8.7.

hsbt:
@gavrieltal Thanks! But the current master branch already fixed this pull-request.

PR #6809, opened 2018-11-25
Fix using gemspec & force_ruby_platform on windows

### LINKED ISSUE #6801: force_ruby_platform misses gemspec dependencies on Windows
If you use `force_ruby_platform` on Windows, gemspec dependencies seem to be ignored. I was using this setting on byebug because I had [problems](https://ci.appveyor.com/project/deivid-rodriguez/byebug/build/746/job/5im93y1wd05629fr) in the past when installing some default gems under Windows, and setting that [fixed it](https://github.com/deivid-rodriguez/byebug/pull/343).

Since I encountered this problem I have now [removed the setting](https://github.com/deivid-rodriguez/byebug/pull/512), and things seem to be working fine, but I thought I would report it anyways.

I created a [sample project](https://github.com/deivid-rodriguez/force_ruby_platform_test) to demostrate [the issue](https://ci.appveyor.com/project/deivid-rodriguez/force-ruby-platform-test/builds/20503688). Basically, when I specify a runtime dependency on the gemspec (`rouge` in this case), and I use `force_ruby_platform` on Windows, this new dependency is ignored by `bundle install` even if the `gemspec` DSL is properly specified in the `Gemfile`.

segiddins:

The problem was using `gemspec` and `force_ruby_platform` on Windows would lead to gems not being requirable.

Fixes #6801.

My diagnosis was there was a place where `force_ruby_platform` wasn't being taken into account, namely the query methods on `Bundler.current_ruby`

My fix was to add a check for the local platform in `current_ruby`, so the `force_ruby_platform` override would be taken into account.

I chose this fix because it avoids hard-coding knowledge of the setting in more places.

### DIFF
--- a/lib/bundler/current_ruby.rb
+++ b/lib/bundler/current_ruby.rb
@@ -38,6 +38,8 @@ module Bundler
     ].freeze
 
     def ruby?
+      return true if Bundler::GemHelpers.generic_local_platform == Gem::Platform::RUBY
+
       !mswin? && (RUBY_ENGINE == "ruby" || RUBY_ENGINE == "rbx" || RUBY_ENGINE == "maglev" || RUBY_ENGINE == "truffleruby")
     end
 
--- a/lib/bundler/gem_helpers.rb
+++ b/lib/bundler/gem_helpers.rb
@@ -2,7 +2,7 @@
 
 module Bundler
   module GemHelpers
-    GENERIC_CACHE = {} # rubocop:disable MutableConstant
+    GENERIC_CACHE = { Gem::Platform::RUBY => Gem::Platform::RUBY } # rubocop:disable MutableConstant
     GENERICS = [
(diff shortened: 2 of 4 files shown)


deivid-rodriguez:
@segiddins I reverted 2 of the commits in this PR, and that fixed the new test that you added. I also squashed the added test, with the commit that's making it pass. https://github.com/bundler/bundler/pull/6809/commits/493ff581b8892a3c6603fae38cd35cbe2d308b05 is not needed for this PR, but I left it there because I think it's a good refactoring :)

deivid-rodriguez:
By the way, this PR as it is now seems to reduce the number of Windows spec failures from 356 to 349 :smiley:

deivid-rodriguez:
I did try this PR, and it fixed the original problem. It also includes a regression spec, so I plan to merge it tomorrow unless there's feedback.

deivid-rodriguez:
@bundlerbot r+

PR #6810, opened 2018-11-25
fixed 'critical' codereview warning about complex conditional

gavrieltal:
I went to codereview.com and made an adjustment to how a conditional statement was written because it was judged too complicated.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -425,7 +425,7 @@ module Bundler
       deleted_deps = @locked_deps.values - @dependencies
 
       # Check if it is possible that the source is only changed thing
-      if (new_deps.empty? && deleted_deps.empty?) && (!new_sources.empty? && !deleted_sources.empty?)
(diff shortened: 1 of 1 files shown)


colby-swandale:
Can you please fill in the pull request template with about what the problem is and what this is fixing.

segiddins:
I’m not sure this is correct, you’re no longer checking the changes to the sources?

gavrieltal:
Oops, I've totally jumped the gun on this one! My change is probably simply incorrect. It appears that this portion of the code is simply not covered by tests, which led me to think my submission was fine. I suppose I'll make writing some tests my next task. Thanks!

segiddins:
I usually just run rspec on a single ruby version for specs I suspect cover what I’m changing, or the whole suite if I’m convinced I’ve broken something.

colby-swandale:
This PR can't be merged as this is breaking functionality.

PR #6811, opened 2018-11-27
use '|' to preserve line breaks in welcome message

colby-swandale:

I used the wrong block in the welcome message YAML. `>` strips the new line breaks, we want to use `|` instead.


colby-swandale:
@bundlerbot r+

PR #6818, opened 2018-11-28
Following up ruby core changes.

hsbt:

These changes enabled to use the same configuration of executables in ruby core repository.

  * To use libexec instead of bin directory.
  * Restore exe direcotry for bundler.gemspec while running test suite.

The ruby core repository uses `exe` directory for other usage. I chose `libexec` instead of `exe` in ruby core. And, The location of `bundler.gemspec` is not the root of the repository on ruby core. We should restore `libexec` to `exe` under the `lib` dir.

see. https://github.com/ruby/ruby/blob/trunk/lib/bundler.gemspec


hsbt:
@bundlerbot r+

PR #6819, opened 2018-11-30
Fix FIPS detection

frezbo:
Signed-off-by: Noel Georgi <git@frezbo.com>

Ruby compiled with latest OpenSSL FIPS would not raise a `DigestError` rather raise a `RuntimeError`, so it better to detect as how the ruby OpenSSL library sets values. 

detect as how the ruby OpenSSL library sets values.

### DIFF
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -210,14 +210,11 @@ module Bundler
 
     def md5_available?
       return @md5_available if defined?(@md5_available)
-      @md5_available = begin
-        require "openssl"
-        OpenSSL::Digest::MD5.digest("")
-        true
-      rescue LoadError
-        true
-      rescue OpenSSL::Digest::DigestError
+      require 'openssl'
+      @md5_available = if OpenSSL::OPENSSL_FIPS && OpenSSL.fips_mode
         false
+      else
+        true
       end
     end
 
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -210,7 +210,7 @@ module Bundler
 
     def md5_available?
       return @md5_available if defined?(@md5_available)
-      require 'openssl'
+      require "openssl"
       @md5_available = if OpenSSL::OPENSSL_FIPS && OpenSSL.fips_mode
         false
       else
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -222,10 +222,10 @@ module Bundler
           OpenSSL::Digest::MD5.digest("")
           true
         end
-        rescue LoadError
-          true
-        rescue OpenSSL::Digest::DigestError
-          false
+      rescue LoadError
+        true
+      rescue OpenSSL::Digest::DigestError
+        false
       end
     end
 
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -210,11 +210,22 @@ module Bundler
 
     def md5_available?
       return @md5_available if defined?(@md5_available)
-      require "openssl"
-      @md5_available = if defined?(OpenSSL::OPENSSL_FIPS) && OpenSSL::OPENSSL_FIPS && OpenSSL.fips_mode
-        false
-      else
-        true
+      @md5_available = begin
+        require "openssl"
+        if defined?(OpenSSL::OPENSSL_FIPS) && OpenSSL.respond_to?(:fips_mode)
+          if OpenSSL.fips_mode
+            false
+          else
+            true
+          end
+        else
+          OpenSSL::Digest::MD5.digest("")
+          true
+        end
+        rescue LoadError
+          true
+        rescue OpenSSL::Digest::DigestError
+          false
       end
     end
 
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -211,7 +211,7 @@ module Bundler
     def md5_available?
       return @md5_available if defined?(@md5_available)
       require "openssl"
-      @md5_available = if OpenSSL::OPENSSL_FIPS && OpenSSL.fips_mode
+      @md5_available = if defined?(OpenSSL::OPENSSL_FIPS) && OpenSSL::OPENSSL_FIPS && OpenSSL.fips_mode
         false
       else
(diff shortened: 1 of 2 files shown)


lib/bundler/shared_helpers.rb
```diff
@@ -210,14 +210,11 @@ def pretty_dependency(dep, print_source = false)
...
     def md5_available?
       return @md5_available if defined?(@md5_available)
-      @md5_available = begin
-        require "openssl"
-        OpenSSL::Digest::MD5.digest("")
-        true
-      rescue LoadError
-        true
-      rescue OpenSSL::Digest::DigestError
+      require "openssl"
```
colby-swandale: We should still support old versions of OpenSSL that does not define `OpenSSL.fips_mode`

```suggestion
      @md5_available = begin
        require "openssl"
        if defined?(OpenSSL::OPENSSL_FIPS) && OpenSSL.respond_to?(:fips_mode)
          OpenSSL.fips_mode
        else
          OpenSSL::Digest::MD5.digest("")
          true
        end
      rescue LoadError
        true
      rescue OpenSSL::Digest::DigestError
        false
      end
```

spec/bundler/fetcher/compact_index_spec.rb
```diff
@@ -59,8 +59,7 @@ def remove_cached_md5_availability
 
         context "when FIPS-mode is active" do
           before do
-            allow(OpenSSL::Digest::MD5).to receive(:digest).
-              and_raise(OpenSSL::Digest::DigestError)
+            allow(OpenSSL).to receive(:fips_mode).and_return(true)
```
colby-swandale: this should return `false` but i'm also not sure if this should be changed, i don't think this will work on ruby 1.X versions

lib/bundler/shared_helpers.rb
frezbo: This makes sense, I will update it, glad you thought of backward compatibility.

spec/bundler/fetcher/compact_index_spec.rb
frezbo: I don't get why it should return `false`, the context is `context "when FIPS-mode is active" do`. Can you please explain?

frezbo:
@colby-swandale I have kept the old code and fixed the `require` stub.

frezbo:
Seems the test suite failed for `ruvy 1.8.7` starting `remove_instance_variable` is private. Is this related to the code changes?

spec/bundler/fetcher/compact_index_spec.rb
colby-swandale: Because we're raising an exception with `and_raise(OpenSSL::Digest::DigestError)` in the `before` block.  We're also already testing the happy path on line 70, so the context seems to be inaccurate.

colby-swandale: Oh i got mixed up here, you're right.

spec/bundler/fetcher/compact_index_spec.rb
```diff
@@ -35,11 +35,8 @@
       end
 
       context "when OpenSSL is not available" do
-        before do
-          allow(compact_index).to receive(:require).with("openssl").and_raise(LoadError)
-        end
-
         it "returns true" do
+          allow(Bundler::SharedHelpers).to receive(:require).with("openssl").and_raise(LoadError)
```
colby-swandale: Does this need to be moved into the `it` block for any specific reason?

spec/bundler/fetcher/compact_index_spec.rb
```diff
@@ -57,6 +54,22 @@ def remove_cached_md5_availability
...
+            expect(compact_index).to_not be_available
+          end
+        end
+
+        it "returns true" do
+          expect(compact_index).to be_available
+        end
+      end
+
+      context "when OpenSSL is FIPS-enabled", :ruby => "< 2.0.0" do
```
colby-swandale: I'm guessing that Ruby 1.8.7 in Travis has a version of OpenSSL that has `fips_mode` available, so the `:ruby` tag is not necessary. But we should try and have a test that covers when `fips_mode` is not available.

spec/bundler/fetcher/compact_index_spec.rb
frezbo: yeh, it's rspec specific, unless its `suibject` it needs to be in the `it` block.

deivid-rodriguez: Can you clarify your answer @frezbo? I didn't understand it, and I wonder the same thing @colby-swandale wonders.

spec/bundler/fetcher/compact_index_spec.rb
```diff
@@ -74,6 +70,19 @@ def remove_cached_md5_availability
       end
     end
 
+    context "when OpenSSL is FIPS-enabled with OPENSSL_FIPS not available" do
+      context "when FIPS-mode is active" do
+        before do
+          OpenSSL.send(:remove_const, "OPENSSL_FIPS") if defined? OpenSSL::OPENSSL_FIPS
```
deivid-rodriguez: Should we restore the const to the previous value it had before this test?

deivid-rodriguez (changes requested):
Hi @frezbo! Sorry for letting this forgotten for such a long time. I added a couple of comments, but the PR looks fine to me in general. Can you rebase and force-push to run CI against it?

frezbo:
@deivid-rodriguez Thanks for looking into this, I had also completely forgotten this issue. My system's currently brokem and lost all data. I will come bck to this once I'm setup with my new one, probably in the next 3-4 weeks.

deivid-rodriguez:
Great, thanks! And good luck with restoring your system :)

deivid-rodriguez:
Hi @frezbo! Any news? :smiley:

frezbo:
@deivid-rodriguez Unfortunately, I still don't have setup where I can re-produce the issue and do development. My workstation eco-system has completely switched. I might be able to take this up probably after 3-4 months only. Requesting any other OSS contributor to take this up.

frezbo:
@deivid-rodriguez Let me see if I can pick this up, I am getting a setup slowly done where I can further work on this.

deivid-rodriguez:
That is great, thanks @frezbo!

PR #6823, opened 2018-12-03
Filter credentials from gemfile for bundler 2.0

### LINKED ISSUE #6489: Credentials are not properly filtered out in 'bundle install'
#4704 fixed some of this, but apparently there still seems to be an issue under certain circumstances.

Here is my Gemfile, running with bundler `1.16.1`:

```ruby
git_source(:acme) { |repo_name| "https://#{ENV['GITHUB_TOKEN']}:x-oauth-basic@github.com/acme/#{repo_name}" }

gem 'foobar', acme: 'foobar'
```

When I run `GITHUB_TOKEN=my-secret-token bundle install`, this is part of the output I get:

```
Using foobar 0.1.0 from https://my-secret-token:x-oauth-basic@github.com/acme/foobar (at master@abc1234)
```

This is clearly less than optimal, especially if being used on Travis or other CI services. (In our case, we are using private repos so the criticality is "less severe" _for us_, but it's still an issue since potential credential leakage is never good.)

(Rationale for using https git sources rather than SSH - this is being used in Dockerfiles, where passing in an ENV variable for a GitHub OAuth token is somewhat more convenient than having to pass in a custom SSH key.)

gavrieltal:
### What was the end-user problem that led to this PR

The problem was that private credentials in the gemfile were sent to the lockfile to preserve backwards-compatibility in bundler < 2, but for bundler >= 2 this security should be offered to users.

#6489 

My fix was to filter credentials on the remotes of sources.

I chose this fix because it is minimal and mimics #6490 in behavior, I believe.

Fixes #6489.

### DIFF
--- a/lib/bundler/lockfile_generator.rb
+++ b/lib/bundler/lockfile_generator.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require "bundler/uri_credentials_filter.rb"
+require "bundler/uri_credentials_filter"
 
 module Bundler
   class LockfileGenerator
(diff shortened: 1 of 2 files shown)


zoras:
@gavrieltal Can make sure the gem installs even when the credentials are filtered out? The filter is needed while logging only.

PR #6828, opened 2018-12-05
Added quiet flag to inline bundler

ipepe:

I'm using inline bundler on my CI minitest file. I wanted to make it quiet so output is clean and meaningful.

Bundler inline's gemfile method only accepted :ui option.

I added quiet option flag to gemfile method.

I chose this fix because it looked simple and clean.

### DIFF
--- a/lib/bundler/inline.rb
+++ b/lib/bundler/inline.rb
@@ -36,6 +36,7 @@ def gemfile(install = false, options = {}, &gemfile)
 
   opts = options.dup
   ui = opts.delete(:ui) { Bundler::UI::Shell.new }
+  ui.level = "silent" if opts.delete(:quiet)
   raise ArgumentError, "Unknown options: #{opts.keys.join(", ")}" unless opts.empty?
 
   old_root = Bundler.method(:root)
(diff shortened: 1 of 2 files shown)


colby-swandale:
This PR will need a test before we will consider merging it.

ipepe:
👍 I'm on it!

ipepe:
How do You like it now @colby-swandale ?

lib/bundler/inline.rb
```diff
@@ -36,6 +36,7 @@ def gemfile(install = false, options = {}, &gemfile)
 
   opts = options.dup
   ui = opts.delete(:ui) { Bundler::UI::Shell.new }
+  ui.level = "silent" if opts.delete(:quiet) { false }
```
colby-swandale: We don't need to set a default value here.

spec/runtime/inline_spec.rb
```diff
@@ -112,6 +112,20 @@ def confirm(msg, newline = nil)
...
+      require 'bundler'
+
+      gemfile(true, :quiet => true) do
+        source "https://notaserver.com"
+        gem "activesupport", :require => true
+      end
+    RUBY
+
+    expect(out).to be_empty
+    expect(exitstatus).to be_zero if exitstatus
```
colby-swandale: This assertion can be removed, this test only cares about the `quiet` flag.

ipepe:
`some small minor details` fixed!

colby-swandale:
Looks great! Thank you!

@bundlerbot r+

ipepe:
Thank You

PR #6830, opened 2018-12-06
move rubocop into a separate stage in Travis-CI

colby-swandale:
## Overview

This PR is pretty much a copy of https://github.com/rubygems/rubygems/pull/2510. It moves rubocop into a separate stage that is run before any tests are executed. This will allow for faster feedback and mean that we don't have to run rubocop on all our available instances in Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,16 @@ rvm:
   - 2.4.5
   - 2.3.8
 
+jobs:
+  include:
+    - stage: linting
+      rvm: 2.5
+      script: rake rubocop
+
+stages:
+  - linting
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
@bundlerbot r+

deivid-rodriguez:
Actually, I hate to say it since I feel so happy about it, but I think this is unintentional: CI is no longer running against 1.8.7... :kissing_smiling_eyes:.

deivid-rodriguez:
@colby-swandale Just let me know, and I'll revert this if you think it's necessary, since I merged it :)

PR #6831, opened 2018-12-06
Modify gem build warnings spec

bronzdoc:
To make awareness of  Rubygems security issues that may have, we are adding a set of warnings to Rubygems when a gem is built,  so i'm modifying this spec to adhere with Rubygems.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,26 +50,17 @@ jobs:
     - rvm: 2.6.1
       script: rake rubocop
       stage: linting
-    # Ruby 2.5, Rubygems 2.7 and up
+    # Ruby 2.5, Rubygems 2.7
     - rvm: 2.5.3
       env: RGV=v2.7.8
       stage: test
-    # Ruby 2.4, Rubygems 2.6 and up
+    # Ruby 2.4, Rubygems 2.6
     - rvm: 2.4.5
       env: RGV=v2.6.14
       stage: test
-    - rvm: 2.4.5
-      env: RGV=v2.7.8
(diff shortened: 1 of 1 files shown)


bronzdoc:
@deivid-rodriguez yes, i should have renamed the spec to `bundler gem can be built` i really don't understand this spec tbh with you, bundler gem should be able to be built regardles of Rubygems warnings. If you don't want warnings when building it use the `--silent` flag.  If we have to modify the output of rubygems stderr to make the spec pass i don't see the point of having it. That's my point of view i may have not getting something right?

bronzdoc:
@deivid-rodriguez yes, i should have renamed the spec to `bundler gem can be built` i really don't understand this spec tbh with you, bundler gem should be able to be built regardles of Rubygems warnings. If you don't want warnings when building it use the `--silent` flag.  If we have to modify the output of rubygems stderr to make the spec pass i don't see the point of having it. That's my point of view i may have not getting something right?

deivid-rodriguez:
The way I see it, we have some `gem build` linting in rubygems that detects what we consider bad practices when defining gemspecs. To me, it makes perfect sense that we stick to our own good practices and make sure that the `bundler` gem builds without warnings. Obviously sometimes there's no way around things, so it's good to have a list of exceptions, but I think this check is good to have in general.

bronzdoc:
what a hell^ 😖

bronzdoc:
why this got closed in a force push?

bronzdoc:
nvm...

PR #6832, opened 2018-12-06
Bring CI matrix back to its previous state

deivid-rodriguez:

The problem was that we lost some CI matrix entries after #6830.

My diagnosis was that extra entries should be defined under `jobs:` now, specifying their stage.

My fix is to merge the `jobs` and `matrix` section, and specify the `stage` of each entry.

I chose this fix because it seems to bring back the matrix to the way it was before (except for the extra linting stage).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,12 +31,6 @@ rvm:
   - 2.4.5
   - 2.3.8
 
-jobs:
-  include:
-    - stage: linting
-      rvm: 2.5
-      script: rake rubocop
-
 stages:
   - linting
   - test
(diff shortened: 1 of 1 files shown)


colby-swandale:
Thanks for picking up on this 🙇 

@bundlerbot r+

deivid-rodriguez:
Is there a way to cancel the bundlerbot run? No big deal, but I still wanted to add some tiny changes and I'm curious if it can be done.

deivid-rodriguez:
I'm going to try, just for fun!

deivid-rodriguez:
@bundlerbot r-

deivid-rodriguez:
@bundlerbot r=colby-swandale

PR #6834, opened 2018-12-08
Allow files other than Gemfile.lock to be opened in definition specs

segiddins:

The problem was the specs are brittle.

My diagnosis was  we should assume code can open other files.

My fix allows other files to be opened, and only raises when opening the file under test.


hsbt:
@bundlerbot r+

hsbt:
@colby-swandale Can you backport this into 2-0-stable branch? This needs to be rubygems change that is https://github.com/rubygems/rubygems/blob/master/lib/rubygems/stub_specification.rb#L114

PR #6843, opened 2018-12-18
Improve `clean_env` deprecation path

deivid-rodriguez:

The problem was that I noticed that there's still some use cases where the current implementation of `Bundler.with_clean_env` can be useful. Whereas the most common use case is to go back to the original environment before bundler was loaded, sometimes you actually want a fully "unbundled" environment. For example, I wanted to test a rails template in isolation (`rails new my_app -m my_template.rb`) in a library where we specifically set `BUNDLE_GEMFILE` in CI. In that context, `with_original_env` won't do the trick for me, because to properly test my template i need to make sure no bundler environment is set at all.

My diagnosis was that we should probably undeprecate `(with_)clean_env`. But @indirect suggested that we should still deprecate the method name, because it's confusing what each developer understands by "clean". That's a very good point.

My fix is to move the functionality of `(with_)clean_env` to `(with_)unbundled_env`, and deprecate `(with_)clean_env`.

In addition, I noticed that the current deprecation message for `with_clean_env` actually mentioned the method `clean_env`, which is not a method the end user is using directly. So I changed the deprecation messages to suggest the proper alternative for each case (suggest `unbundled_env` as a replacement for `clean_env`; and `with_unbundled_env` as a replacement for `with_clean_env`).

I chose this fix because it undeprecates the functionality previously provided by `clean_env`, while still deprecating the `clean_env` name because of being confusing.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -280,10 +280,19 @@ EOF
       ORIGINAL_ENV.clone
     end
 
-    # @deprecated Use `original_env` instead
-    # @return [Hash] Environment with all bundler-related variables removed
+    # @deprecated Use `unbundled_env` instead
     def clean_env
-      Bundler::SharedHelpers.major_deprecation(2, "`Bundler.clean_env` has weird edge cases, use `Bundler.original_env` instead")
+      Bundler::SharedHelpers.major_deprecation(
+        2,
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
@bundlerbot r+

PR #6850, opened 2018-12-22
Change default setting allow_sudo is 'false'

a-wing:

The problem was very Danger

If your `/etc/sudoers` set `NOPASSWD`
look like:
```
%wheel ALL=(ALL) NOPASSWD: ALL
```
bunlde install default install to my `/usr/lib`

But I don't know, there is no need for me to confirm

So. very Danger

default setting disable sudo

### DIFF
--- a/lib/bundler/settings.rb
+++ b/lib/bundler/settings.rb
@@ -282,8 +282,7 @@ module Bundler
 
     def allow_sudo?
       key = key_for(:path)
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Hi @a-wing, could you please explain this a bit more in detail, get specs passing and add a regression spec that proves your fix? Thanks!

deivid-rodriguez:
Closing for lack of feedback.

PR #6856, opened 2018-12-28
Test against Ruby 2.6 and RubyGems 3

segiddins:

The problem was we weren't testing our compatibility with the latest and greatest.


segiddins:
Looks like we can no longer use forwardable (it's now a gem), and a spec needs to be updated for bundler being a default gem

segiddins:
That last commit doesn’t work, we need to avoid using the shell so we don’t need to shellescape the directory

deivid-rodriguez:
Maybe we can use a manifest file like in rubygems.

colby-swandale:
i would prefer avoiding that

colby-swandale:
I've ended up just copying the what the `Shellwords.escape` method does into the gemspec

segiddins:
👍

colby-swandale:
@bundlerbot r+

PR #6860, opened 2018-12-29
Add missing space inside block braces

EmilioCristalli:

When using `bundle gem GEM_NAME`, the Gemfile bundler generated was missing a space in the `git_source(:github)` block

### DIFF
--- a/lib/bundler/templates/newgem/Gemfile.tt
+++ b/lib/bundler/templates/newgem/Gemfile.tt
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-git_source(:github) {|repo_name| "https://github.com/#{repo_name}" }
+git_source(:github) { |repo_name| "https://github.com/#{repo_name}" }
 
 # Specify your gem's dependencies in <%= config[:name] %>.gemspec
 gemspec


deivid-rodriguez (commented):
This is the [preferred style we use in `bundler`](https://github.com/bundler/bundler/blob/a63a39d738865665f714f179fe42dd10006da26d/.rubocop.yml#L39-L41). A more standard normalization was proposed in https://github.com/bundler/bundler/pull/3850, but it was fixed via https://github.com/bundler/bundler/commit/14a7eb25ee03c9fb4ecf3e4c6b1f44a8267a413a by enabling the opposite style even if it was less common in the code base. It sounds like it's a personal preference of some bundler developers.

In any case, it's unclear what we should do about the style of code generated by bundler, but the current proposal is to also generate a rubocop config against which the generated code is offense free. This is being worked in https://github.com/bundler/bundler/pull/6455.

So, I think we can close this PR and focus on #6455 instead?

EmilioCristalli:
oh, didn't notice it was bundler's preferred style.

I guess I assumed it would make sense to add the space to make it consistent with the [newgem.gemspec](https://github.com/bundler/bundler/blob/460f2f9633d7db05f526c4fa1c6d2f914415f921/lib/bundler/templates/newgem/newgem.gemspec.tt#L36), but I'm ok with closing this if you are leaning towards keeping bundler style for generated code.

Or I can fix the newgem.gemspec.tt if anyone thinks it's worth it (I personally prefer a space before block params, but having consistent styles is usually a good idea, so I'm ok either way)

deivid-rodriguez:
I feel similarly (I also prefer a space, and I also like consistency), let's wait for other maintainer's input.

indirect:
I personally prefer it without a space, but it seems like most Ruby devs have landed on using a space. I'm okay with switching Bundler's style to reduce fighting with linters that run on generated Gemfiles. But I will defend double quotes to the death! 😅 

If it helps, this should stop being a problem soon, when 2.0 final ships. 😄

hsbt:
@EmilioCristalli Bundler repo was merged into rubygems. If you still interested in this PR, Can you re-open this on https://github.com/rubygems/rubygems? Thanks.

PR #6862, opened 2019-01-02
stop circular require warnings - guard require "rubygems"

MSP-Greg:

Testing rubygems/rubygems with warnings on, specifically, `test_gem.rb`.  Several 'circular require' warnings.

RubyGems gets required, then requires Bundler.  Several places in Bundler then require RubyGems.

Guard the RubyGems requires.  Note that rubygems/specification is one of the few files loaded by RubyGems, most are autoload.  I kept adding guards until all warnings were removed.

Unaware of other possible options

### DIFF
--- a/lib/bundler/compatibility_guard.rb
+++ b/lib/bundler/compatibility_guard.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: false
 
-require "rubygems"
+require "rubygems" unless Object.const_defined? :Gem
 require "bundler/version"
 
 if Bundler::VERSION.split(".").first.to_i >= 2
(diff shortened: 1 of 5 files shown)


MSP-Greg:
To repo, from RubyGems repo:
```
rake test RUBYOPT=-v TEST=test/rubygems/test_gem.rb
```

MSP-Greg:
ruby/ruby just switched on warnings in their tests.  See for similar circular require warnings, starting around line 2912:

https://travis-ci.org/ruby/ruby/jobs/474816395

MSP-Greg:
@segiddins

That's a very deep rabbit hole.

What if someone modified $LOAD_PATH so the require statement loaded a different rubygems.rb file?   

I haven't checked, but `autoload` might have similar issues with a improperly defined constant...

I could update the PR where, instead of using the guard, one could do something similar to:
```
verbose, $VERBOSE = $VERBOSE, false
require "rubygems"
$VERBOSE = verbose
```

PR #6864, opened 2019-01-03
Include URL in Bundler::Fetcher::FallbackError message for Net::HTTPNotFound

greysteil:

It was really painful to debug persistent NotFound errors, as it wasn't clear where they were coming from,

Bundler was obfuscating the URL that wasn't found unnecessarily.

My fix is to add the URL to the `Bundler::Fetcher::FallbackError` message for `Net::HTTPNotFound` errors.

I chose this fix because it was simple and easy to test.

### DIFF
--- a/lib/bundler/fetcher/downloader.rb
+++ b/lib/bundler/fetcher/downloader.rb
@@ -37,7 +37,7 @@ module Bundler
         when Net::HTTPUnauthorized
           raise AuthenticationRequiredError, uri.host
         when Net::HTTPNotFound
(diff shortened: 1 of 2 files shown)


segiddins:
Can you add a spec that makes sure that passwords are removed from the URI in the error message?

greysteil:
Good shout - updated! The CI failure on ruby-head is hopefully unrelated...

segiddins:
Nice!
@bundlerbot r+

PR #6867, opened 2019-01-03
Lower required RubyGems to 2.5 or greater

indirect:
It turns out that every existing version of RubyGems will ignore the required RubyGems version and try to install the newest version of Bundler... until the middle of the install, when it will raise an exception and fail. This PR tries to fix that in the least-bad way possible.

The end user problem is that running `gem install bundler` now produces an exception on any Ruby and RubyGems older than 2.6 or 3.0 respectively.

Unfortunately, RubyGems does not use the `required_rubygems_version` to select an older version of Bundler to install. Instead, RubyGems tries to install the newest version even though that version is not compatible, and then raises an exception during installation because it is not compatible.

Lower the `required_rubygems_version` in Bundler 2 back down to RubyGems 2.5 or greater.

I chose this fix because it will stop RubyGems from raising an exception while installing Bundler 2, and instead allow the installation to complete successfully.

There are some bugfixes in RubyGems 3 that improve the quality of life for Bundler 2 users, but they are less bad than RubyGems simply exploding anytime you try to run `gem install bundler`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,21 @@ matrix:
     # 3.x mode
     - rvm: 2.5.3
       env: RGV=v3.0.1 BUNDLER_SPEC_SUB_VERSION=3.0.0
+    # Ruby 2.5, Rubygems 2.7 and up
+    - rvm: 2.5.3
+      env: RGV=v2.7.7
+    # Ruby 2.4, Rubygems 2.6 and up
+    - rvm: 2.4.5
+      env: RGV=v2.6.14
+    - rvm: 2.4.5
+      env: RGV=v2.7.7
+    # Ruby 2.3, Rubygems 2.5 and up
+    - rvm: 2.3.7
+      env: RGV=v2.5.2
+    - rvm: 2.3.7
+      env: RGV=v2.6.14
+    - rvm: 2.3.7
+      env: RGV=v2.7.7
     # Ruby-head (we want to know how we're doing, but not fail the build)
     - rvm: ruby-head
       env: RGV=master
--- a/bundler.gemspec
+++ b/bundler.gemspec
@@ -34,7 +34,7 @@ Gem::Specification.new do |s|
 
   if s.version >= Gem::Version.new("2.a".dup)
     s.required_ruby_version     = ">= 2.3.0"
-    s.required_rubygems_version = ">= 3.0.0"
+    s.required_rubygems_version = ">= 2.5.0"
   else
     s.required_ruby_version     = ">= 1.8.7"
     s.required_rubygems_version = ">= 1.3.6"


indirect:
Summarizing some relevant conversation with @segiddins and @colby-swandale:

- It seems like required rubygems version was set to 3.0 with the idea that rubygems would install the newest version that stated it was compatible. Instead, every rubygems version including 3.0 tries to install the newest bundler regardless, and throws an exception if the newest version happens to say it is not compatible (@indirect)
- Running `gem install bundler` or `gem install rails` now fails on macOS default Ruby, and on any Ruby older than 2.6 (@colby-swandale)
- If we support Ruby 2.3, we should also support RubyGems 2.5.1 (which shipped with Ruby 2.3.0) and greater (@segiddins)
- While RubyGems 3.0 contains some quality of life fixes and improvements for Bundler 2, those improvements are not worth the cost of `gem install bundler` failing on every older RubyGems version (@indirect)

indirect:
@bundlerbot r=greysteil

colby-swandale:
For the record, The main motivator behind requiring 3.0.0 was https://github.com/rubygems/rubygems/pull/2426 that fixed an issue with the BundlerVersionFinder that was released in v3.

colby-swandale:
@bundlerbot r=greysteil

colby-swandale:
@bundlerbot retry

colby-swandale:
@bundlerbot r=greysteil

colby-swandale:
I'm just going to merge this manually, i don't know why bundlerbot is not playing nice.

mvz:
As predicted in https://github.com/bundler/bundler/pull/6867#issuecomment-451298324, this change 'broke' the BundlerVersionFinder on systems with older versions of RubyGems.

(Well, not 'broke', because it wasn't working to begin with, but the sudden silent updating of versions in Gemfile.lock was a tad surprising, given the changelog entry).

deivid-rodriguez:
Hi @mvz, just making sure. You mean whatever change bumped the requirement to 3.0.0, not this specific changeset that lowered it back to 2.5. Right?

mvz:
I do mean that I think this particular pull request introduced a problem: Given a system with pre-3.0 rubygems version, and a Gemfile.lock with bundler 1.17.3 specified inside,
- Bundler 2.0.0 will complain about needing a newer rubygems.
- Bundler 2.0.1 will overwrite the Gemfile.lock with its own version number.

With rubygems 3.0, both versions keep the existing bundler version in the Gemfile.lock as it is.

deivid-rodriguez:
Mmmm. But that sounds like some bug fixed in rubygems 3.0, right? Or would you expect us to change anything in bundler related to this?

mvz:
Well, as I see it, Bundler 2.0.0 required rubygems 3.0, to make one important feature work (version switching). Then, in 2.0.1, it lowered the requirement, and this important feature broke, without much warning (https://github.com/bundler/bundler/blob/master/CHANGELOG.md#201-2019-01-04).

So yes, it's a bug fixed in rubygems 3.0, but isn't that why the dependency was changed in the first place?

Some things Bundler could do:
* Document that keeping the version in Gemfile.lock intact is broken with pre-3.0 versions of rubygems.
* Warn about use of older versions of rubygems.
* Up the version requirements on rubygems. I'll readily admit that I have absolutely no idea how feasible that is.

deivid-rodriguez:
> So yes, it's a bug fixed in rubygems 3.0, but isn't that why the dependency was changed in the first place?

Yeah, but we realized that was too agressive, because it would make `gem install bundler` fail on older rubygems. That's why we reverted it.

I believe this is a problem only on ruby 2.3 and ruby 2.4, because the newer rubies we support provide rubygems 2.7 or higher by default, and I don't think this is an issue on rubygems >= 2.7. 

I would be good with adding a documentation bnit about recommended rubygems & ruby support vs official rubygems & ruby support.

PR #6879, opened 2019-01-04
Merge 2-0-stable into master and get specs passing

deivid-rodriguez:

The problem is that master is still running specs against older rubies even if 2-0-stable is not. That means quite a lot of PRs are blocked because they don't pass tests on old rubies.

My diagnosis is that we need to officially start bundler 3 development.

My fix is to merge the 2-0-stable branch into master, bump the version to 3.0.0.dev in `version.rb`, stop testing against old rubies and get specs passing.

I chose this fix because it gets master green while including everything currently in the 2-0-stable branches.


deivid-rodriguez:
I got specs passing locally against 2.6, let's see what TravisCI says for other rubies.

Also, we have delayed deprecations to bundler 3 but I see no reason to not release the pending major deprecations on a future 2.x release. But we can discuss this separately, make a PR to 2-0-stable if we decide to do it, and then forwardport it to master.

colby-swandale:
Let me sort out the deprecation issue, which should make the merge into master a lot easier.

deivid-rodriguez:
Sure!

hsbt:
@deivid-rodriguez How about the status of this branch? I'm okay to merge this into master branch.

deivid-rodriguez:
I'll try to get this ready to merge today! :+1:

deivid-rodriguez:
I created #6933 to 2-0-stable to sort out deprecations. Then I can forward port those changes here and rebase this PR.

deivid-rodriguez:
Closing in favor of #6957, @colby-swandale will manually cherry-pick any commits he needs from here.

PR #6881, opened 2019-01-05
Fix Travis head, remove RG 2.7.7

MSP-Greg:

As of 2019-01-05, Travis is installing RG 2.7.7 in ruby trunk.  Fix removes it.

My diagnosis was...

My fix...

I chose this fix because...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,15 @@
 language: ruby
 script: rake spec:travis
 before_script:
+  # fix Travis trunk build as of 2019-01-05
+  - |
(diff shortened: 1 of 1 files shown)


colby-swandale:
Can you please explain what this is trying to fix? We pull our own copy of RubyGems that Bundler is tested against (the `RGV` variables in `.travis.yml`)

colby-swandale:
Oh i see, the version of RubyGems installed by travis is getting loaded.

MSP-Greg:
@colby
This won't work with the rakefile, so I'm working in my fork, updating rakefile and the test matrix...

PR #6884, opened 2019-01-07
Remove unnecessary condition

deivid-rodriguez:

The problem was that some code in the feature flags implementation seemed uncovered.

My diagnosis was that the code should be either covered or removed.

My fix was to remove the code.

I chose this fix because the scenario where that code would get covered is more likely to be unintentional and risk introducing a bug. Defining a feature flag without a default block is the same as don't defining a feature flag and just relying on the default value for the setting. In these circumstances, it's more likely that the developer forgot to include a default block, so letting that case crash would be better, I think.


hsbt:
seems reasonable.

hsbt:
@bundlerbot r+

PR #6885, opened 2019-01-07
Don't check for the existence of a writable home directory if BUNDLE_USER_HOME is set

jturkel:
#6024 added support for specifying an alternate bundler home directory via the `BUNDLE_USER_HOME` environment variable. Unfortunately bundler still checks for a writable user home directory even if the `BUNDLE_USER_HOME` override is specified resulting in warnings like this:
```
`/home/appuser` is not a directory.
Bundler will use `/tmp/bundler/home/unknown' as your home directory temporarily.
```
The fix for this problem is to only evaluate bundler path fallbacks if environment variable overrides are not specified.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -198,19 +198,19 @@ module Bundler
     def user_bundle_path(dir = "home")
       env_var, fallback = case dir
                           when "home"
-                            ["BUNDLE_USER_HOME", Pathname.new(user_home).join(".bundle")]
+                            ["BUNDLE_USER_HOME", proc { Pathname.new(user_home).join(".bundle") }]
                           when "cache"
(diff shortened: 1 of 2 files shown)


jturkel:
@segiddins - I believe testing is covered by the change I made to `spec/bundler/bundler_spec.rb` to raise an exception if the user home is requested (at least that makes the test fail without the corresponding code fix). Is there additional test coverage you think is needed?

deivid-rodriguez:
No further feedback so let's do this!

@bundlerbot r+

semaperepelitsa:
Bundler 2.0.2 (released 5 days ago) does not include this change. Are you planning it for future versions?

Issue #6841, opened 2018-12-18
`bundle add` for `git:` and `branch:`?


janpio:
Is there a way to add a dependency with `git:` and `branch:` ([How to install gems from git repositories](https://bundler.io/guides/git.html)) using the `bundle` CLI? 

This should be the result in my `Gemfile`:
```
gem 'nokogiri', git: 'https://github.com/rack/rack.git', branch: 'rack-1.5'
```

`bundle add` doesn't seem to have the necessary parameters: https://bundler.io/v1.17/man/bundle-add.1.html

janpio:
I faked this with a (probably terrible) shell script:

```
#!/bin/sh

if [ -n "$REPO" ] && [ -n "$BRANCH" ]; then
    SEARCH='gem "fastlane".*'
    REPLACE="gem \"fastlane\", :git => \"$REPO\", :branch => \"$BRANCH\""
    sed -i "s@$SEARCH@$REPLACE@" Gemfile
fi
```

If I set the `REPO` and `BRANCH` env variables, this will replace the `gem "fastlane" ...` line (that I know is in the Gemfile) with another line that specified the repo and branch for the fastlane gem.

gkpacker:
Maybe it can help you: https://github.com/rdp/specific_install

janpio:
I don't think so, this seems to install the gem directly - I want to manipulate the `Gemfile` as per the example I added in my original post.

colby-swandale:
There is no options to add a gem with the `git` or `branch` options, but a PR for this would be greatly welcomed.

janpio:
I am looking into this. 

Probably relevant code:

https://github.com/bundler/bundler/blob/0ab0dc7ea999743ccab128f5ab582be834de427a/lib/bundler/cli.rb#L357-L371
https://github.com/bundler/bundler/blob/master/lib/bundler/cli/add.rb
https://github.com/bundler/bundler/blob/d44d803357506895555ff97f73e60d593820a0de/lib/bundler/dependency.rb#L80-L92
https://github.com/bundler/bundler/blob/master/man/bundle-add.ronn
https://github.com/bundler/bundler/blob/master/spec/commands/add_spec.rb

Further inspiration and to research:
https://github.com/bundler/bundler/pull/6556/files
https://github.com/bundler/bundler/search?utf8=%E2%9C%93&q=git&type=

PR #6899, opened 2019-01-09
Windows CI: Azure Pipelines

### LINKED ISSUE #4665: Bundler test suite should run on Appveyor against PRs
We're looking to run the Bundler test suite against new/updated Bundler PRs on Windows environments via [Appveyor](https://ci.appveyor.com/project/AndrArko/bundler).
- A work-in-progress PR was made early on at #4299, but I don't currently have the time bandwidth to undertake this work right now.
- Additionally, @indirect, think we should be using an Appveyor account that isn't a personal account (it's currently under your `AndreArko` user?

janpio:

When I wanted to start to work on #6841 by adding a few tests, I quickly learned that bundler's test suite does not pass on Windows: #6868

I decided to start working on fixing this, and created a few issues already: https://github.com/bundler/bundler/search?q=%22%5BWindows+test+failure%5D%22&unscoped_q=%22%5BWindows+test+failure%5D%22&type=Issues 

But it become clear to me, that the only way to really track the improvement was to add CI for Windows - otherwise opening PRs with improvements would not really make sense.

This PR first makes it possible to run the test suite on Windows at all by "hiding" a bit of code behind `Gem.win_platform?` (`Spec::Manpages.setup` made it super complicated to run on Windows - and is not really necessary to run the test suite). 

It also works around a strange bug concerning `thor`'s (a vendored dependency) use of `readline` which crashes in strange ways on Windows (see #6902). And it adds applies a patch to the `readline` implementation used on CI (via RubyInstaller ruby) that otherwise blocks the tests (see #6907).

Then it adds CI configuration for Azure Pipelines, a free for OSS CI platform. I configured the Azure Pipelines build to create a JUnit compatible report if the test suite and display it in the "Tests" tab of the build.

I think disabling this code for Windows is a valid choice to get the ball rolling, later this will undone and the underlying problem will be fixed.

I chose the workarounds for #6902 and #6907 as I couldn't find a proper bugfix yet - so this will be one of the things to be fixed using Windows CI.

Azure Pipelines is a great CI platform for Windows (and later even multiple platforms if we choose to also run the tests on Linux and macOS).

I chose not to adapt the `rake travis` task (yet) because it is _much_ more complex than the "simple" `bin/rspec` test suite run which I could directly execute in the CI scripts of Azure Pipelines.


hsbt:
Why should we use Azure Pipeline? Its configuration is out of scope for Windows support.

hsbt:
> Any reason why you think Azure Pipelines is not a good idea?

Cost of Maintainance. I'm okay to use Travis and Appveyor only. I'm against to add the new CI platform that was duplicated process with existence CI.

janpio:
Then I suggest removing AppVeyor from this PR and only add the Azure Pipelines configuration. It will enable me much better to fix the Windows test failures as I get feedback on all 3 relevant platforms at the same time, in much less time.

Agree?

hsbt:
> Agree?

Don't agree. If you enabled only Windows, I will accept to merge this PR. It's better to make the small start at first.

janpio:
I removed AppVeyor and the Linux and macOS builds on Azure Pipelines and updated the initial PR description to match this new reality. (This also makes it unnecessary to fix stuff like #6900, so this is good to merge right now.)

lib/bundler/vendor/thor/lib/thor/line_editor/readline.rb
```diff
@@ -1,5 +1,5 @@
 begin
-  require "readline"
+  require "readline" unless Gem.win_platform?
```
deivid-rodriguez: This is vendored code, so this change should be proposed to [the thor project](https://github.com/erikhuda/thor) first. And that will probably involve adding Windows CI in there as well, and thus keep going down the rabbit hole... :sweat_smile:.

Maybe the problem could also be worked around by lazily loading `readline` in thor only when the functionality that needs it is loaded. That way you could probably fix this problem (bundler crashing because of a library not used by bundler being loaded) without having to add Windows support in `thor`.

Thanks for what your doing, by the way!

janpio: Could we maybe accept/merge this change here anyway, while I start working on getting this understood and fixed in `thor` directly? I wanted to start sending PRs for the first ~10 issues I fixed on bundler on Windows, but need working Windows CI to be able to track the progress :/

janpio: (The thor tests almost pass on Windows, so I will definitely look into what really is happening with `readline` here in bundler's usage of thor [and how to fix the rest of those tests I'm at it... 🤷‍♀️ ])

janpio: I will continue my investigation into this at https://github.com/bundler/bundler/issues/6902 👷‍♀️

janpio:
To pull it out of the discussion on a code line at https://github.com/bundler/bundler/pull/6899/files/e182fcbeaa564f0d06a3a3d7fa62e3d8e132a873#diff-501fea893f314bd6414dabb1718c8127: This PR is currently on hold until I figure out what is going on with `readline`. I will report back.

hsbt:
@janpio Thanks. I'm not familiar with Azure Pipeline yet. It's a good time to learn it for me. I also help to prepare to support the Windows environment.

lib/bundler/vendor/thor/lib/thor/line_editor/readline.rb
janpio: I replaced this change in the vendored code by a folder creation in the CI configuration. See more about the reasons for this in #6902. (I could use the help over there)

azure-pipelines/steps.yml
```diff
@@ -0,0 +1,32 @@
+steps:
+
+- task: UseRubyVersion@0
+  inputs:
+    versionSpec: '= 2.4'
+  condition: not(eq(variables['Agent.OS'], 'Darwin')) # not possible on Mac
```
simi: Can be this removed for now since only Windows build is triggered?

janpio: Yep!

```suggestion
```

janpio:
Azure Pipelines can be pretty awesome - short wait times for builds and ok build times as well. And loads of different functionality if you need it.

---

This PR is now good to merge as I found a workaround for #6902 that does not require any code changes in vendored `thor`! Also #6907 has a workaround implemented in the CI script as well:

https://dev.azure.com/janpio/bundler/_build/results?buildId=606
https://dev.azure.com/janpio/bundler/_build/results?buildId=606&view=ms.vss-test-web.test-result-details

---

To activate Azure Pipelines for `bundler/bundler`, one of the maintainers (@hsbt maybe?) should follow these steps: https://samcogan.com/free-azure-build-pipelines-for-open-source-projects/ 

(For the template step, there should be a simple ruby template or just select an empty template - the actual build configuration is part of this PR and will come from here. No manual configuration should be necessary at all.)

Then we should be able to start a build with the configuration of this PR by closing and reopening the PR or just merge this into `master` and see the first build on `master` (that this PR creates a good configuration can be seen at https://dev.azure.com/janpio/bundler/_build/results?buildId=606). 

(Ping me on Slack, username `janpio`, if you need any help with activating Azure Pipelines.)

azure-pipelines/steps.yml
```diff
@@ -0,0 +1,35 @@
...
+- script: |
+    git apply --ignore-space-change --ignore-whitespace azure-pipelines\rbreadline.diff --directory=C:/hostedtoolcache/windows/Ruby/2.4.3/x64/lib/ruby/site_ruby --unsafe-paths
+  displayName: 'patch local readline implementation (for https://github.com/bundler/bundler/issues/6907)'
+
+- script: |
+    ruby bin/rake spec:deps
+  displayName: 'ruby bin/rake spec:deps'
+
+- script: |
+    ruby bin/rspec --format progress -r rspec_junit_formatter --format RspecJunitFormatter -o rspec/bundler-junit-results.xml
```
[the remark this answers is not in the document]
janpio: Yes, I decided to not include this complexity already when there are >600 failing tests on Windows. I thought I would add that later, when the tests are at least semi stable. Should I change this now already?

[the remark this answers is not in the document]
deivid-rodriguez: I also think it's fine to start small and get bundler CI passing against _any_ rubygems version, and later take care about completing the matrix and make sure bundler CI passes against _all_ rubygems versions.

bundler.gemspec
```diff
@@ -47,6 +47,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency "ronn",       "~> 0.7.3"
   s.add_development_dependency "rspec",      "~> 3.6"
   s.add_development_dependency "rubocop",    "= 0.50.0"
+  s.add_development_dependency "rspec_junit_formatter", "~> 0.2.3"
```
colby-swandale: I'm not sure about adding a new dev dependency. Does azure require this to work?

colby-swandale:
When Azure announced Ruby support for their Azure CI service, i was quick to jump onboard and looked into seeing if we could use it ourselves. But thinking about it over the past few weeks, i've been wondering if we're better off using AppVeyour instead.

Appveyor is being used by most OSS Ruby projects on Github at the moment, and a lot of those Appveyor pipelines are being maintained by @MSP-Greg, who is a semi-regular contributor to the RubyGems and Bundler projects. If there was ever an issue with testing Bundler on Windows, we would have someone that is quite familiar with both domains to help resolve any issues.

\cc @indirect thoughts?

MSP-Greg:
@colby-swandale

> I've been wondering if we're better off using AppVeyor instead.

When Azure Pipelines (AP) was first announced, I thought I'd check into building Ruby trunk there.  AP was new, MSFT obviously has resources far beyond what Travis & Appveyor have, and most importantly, it would run a lot of jobs parallel.

I was trying code based on ruby-loco (largely PowerShell) and the ruby/ruby appveyor build code.  A few notes:

1. At the time, I had to install everything.  A version of Ruby (you need Ruby to build Ruby), the full MSYS2/MinGW build system, etc.  The code isn't brittle, but it's not something a non-windows user would want to jump into.

2. Both Appveyor (AV) & Travis 'hold' environment variables as one moves between pre-defined script steps.  AP does not.  So there's a bit of monkeying around with using a AP specific API to share them between script steps.

3. There were a lot of other quirky things, like logs showing absolute time, not relative, etc.  I even had to build a special package of 7-zip, since the normal windows installer had an issue.  At the time, there was not a lot of pre-installed software, at least compared to AV.

Form bouncing around encouraging and helping with AV tests, the main issue I come across is the 'it takes too long'.  AP will help with that.  But, if I ported some repos over to AP, and the script doubled or tripled in size, most people would run away very quickly, as it's just too different from Travis.

I'd certainly like to see bundler testing on Windows.  I looked at it a long time ago, now I don't recall the difference between the 'bundler' testing in RG, vs the testing here.  I haven't really looked at this PR, nor what work @janpio and @hsbt may have done.

**Off topic:**

I think many people in the OSS community don't like seeing a big guy (MSFT) rolling over a small guy (AV).  AV has always been receptive to changes, but it takes forever to update their images (Ruby 2.6 isn't available yet), and they don't allow most accounts to test parallel (although Ruby CI runs two or three).  At one point I posted several suggestions for AP, and several MSFT people were participating.  They are supporting AP and they do want to see it adopted by the OSS community....

Sorry for the long post.  I'm interesting in helping with whatever.

deivid-rodriguez:
I support trying out Azure Pipelines and see how it goes. We have a volunteer, @janpio, who has already done this setup and was (hopefully, still is) willing to fix the test suite on Windows after this. I volunteer to set this up on the bundler organization so @janpio can proceed with his work!

bundler.gemspec
janpio: Not really.

This enables the use of `-r rspec_junit_formatter --format RspecJunitFormatter -o rspec/bundler-junit-results.xml` as a param for `ruby bin/rspec` which creates this file, which can then be used by the `PublishTestResults@2` task to generate the test result analysis at https://dev.azure.com/janpio/bundler/_build/results?buildId=606&view=ms.vss-test-web.build-test-results-tab. That would be very handy to follow the improvements I make when the Windows testing bugfixes get merged, but of course one could also manually track the fixed tests. This is just a convenience thing.

I thought that it would worth that convenience as it is just a development dependency. But if it is easier to merge this PR without this added dependency, I can of course remove it and the part for the test result publishing from the CI config.

indirect: I would prefer that you install `rspec_junit_formatter` as part of the setup steps in Azure Pipelines, since Bundler does not itself depend on the gem for development.

indirect:
I'm also happy to try out Azure Pipelines, since it seems like it will help us with our goal of running the Bundler tests on Windows. Once we have passing tests, we can compare Appveyor and Azure and see which one meets our needs best in the long run.

I've added `bundler/bundler` on GitHub to `bundler/bundler` on Azure.

bundler.gemspec
janpio: Umm, I feel a bit dumb now, but: How does one do that? Just `gem install rspec_junit_formatter`?

janpio:
Thanks for adding the Azure Pipelines integration @indirect.

As I removed the `rspec_junit_formatter` dependency (see https://github.com/bundler/bundler/pull/6899/files/485bbf1e2b32ed603dbd0c790adf35607b8c9ac5#diff-6ac292c8e0d03698a046322989f7e57c for the ongoing discussion about that), Azure Pipelines picked up the commit and ran it - and gave this PR a big red "X". The Azure Pipelines tests of course will fail until I have all the Windows bugs fixes, which can take quite some time I suppose. Should I investigate if it is possible to configure Azure Pipelines to report "success" although the tests are failing? (That could maybe keep all the PRs and commits with a green checkmark, although Azure Pipelines has test failures on the builds.)

deivid-rodriguez:
Maybe if you add `|| exit 0` for now after the command that runs the specs, it will have that effect for now. We can remove it after we manage to get it green.

bundler.gemspec
janpio: Hm, `ruby bin/rake spec:deps` would probbaly be one way, because that basically runs `gem install ...`. But that would also install it on travis, which is probably not wanted.

janpio: Yeah, I am out of my depth here - I made a few commits but always get `C:/hostedtoolcache/windows/Ruby/2.4.3/x64/lib/ruby/2.4.0/rubygems/core_ext/kernel_require.rb:55:in `require': cannot load such file -- rspec_junit_formatter (LoadError)`. All help appreciated.

janpio:
That worked @deivid-rodriguez ❤️ 

Only the `rspec_junit_formatter` problem/suggestion left: https://github.com/bundler/bundler/pull/6899/files/485bbf1e2b32ed603dbd0c790adf35607b8c9ac5#diff-6ac292c8e0d03698a046322989f7e57c

indirect:
@janpio does the version of the gem matter? When you put it in the gemspec, you said `~> 0.2.3`, which means `>= 0.2.0 && < 0.3.0`. When you run `gem install rspec_junit_formatter`, it seems to be installing version 0.4.1.

On my machine, I was able to make it work by re-ordering the require. The script is currently running this command:

```
ruby bin/rspec --format progress -r rspec_junit_formatter --format RspecJunitFormatter -o rspec/bundler-junit-results.xml 
```

When I changed it to this command instead, it seemed to work:

```
ruby -r rspec_junit_formatter bin/rspec --format progress --format RspecJunitFormatter -o rspec/bundler-junit-results.xml 
```

Maybe give that a shot?

janpio:
> does the version of the gem matter?

Nope, that was just a copy paste thing from fastlane where we seem to be using that version.

> When I changed it to this command instead, it seemed to wor

That did indeed work ❤️ But why?

bundler.gemspec
janpio: Now is installed in an additional step in the Azure Pipelines configuration, so fully independent of the general codebase.

indirect:
It worked because our `bin/rspec` script loads some gems and then blocks off any more rubygems, to try to avoid issues when our specs run. The change loads the gem before our script
disables additional gem loads.

janpio:
Understood, it's always in the details ;)

---

And that's a wrap, everything is working as it should now (I think):

image
https://dev.azure.com/bundler/bundler/_build/results?buildId=9&view=ms.vss-test-web.build-test-results-tab
https://dev.azure.com/bundler/bundler/_build/results?buildId=9

I cleaned up the commits a bit, so this is good to merge now 🚀 
Thanks everyone that gave feedback and helped me finish this.

janpio:
Maybe you noticed that the "bundler.bundler" check is still in status "In progress - This check has started..." although on Azure Pipelines it has finished after ~40 minutes just fine. This seems to be a problem on Azure Pipelines caused by the numbers of tests failures of our build. I am in contact with their development team, who are working on a fix right now. Will update when there is news on this.

ethomson:
👋  Apologies for the problem with the check not getting updated from Azure Pipelines.  We had a bug in the check updater code when making a call to our Test Case Manager API.  We sorted that out and we think that things should be 👍 from now on.

janpio:
Thanks @ethomson for the support with this, the Azure Pipelines check is indeed green now and this PR is finally really ready to merge 🚀 

(I so want to work on decreasing that "671 failing tests" number!)

colby-swandale:
@bundlerbot r+

PR #6906, opened 2019-01-13
Add deployment flag to check command

### LINKED ISSUE #6459: Feature request: `bundle check --deployment`
I'd like a command to tell me if my lock file is "out of sync", i.e., if running `bundle install` or `bundle lock` would cause changes on it. I think this can be handy, for example, in CI situations to implement "remember to update the lock file" checks.

I asked about this feature in bundler's slack and @indirect noticed that this was "conceptually `bundle check --deployment`, which I do not think currently exists".

So I'm proposing that we add this feature! :)

NOTE: If someone is to implement something like this now, manually, a couple of one-liners were mentioned in slack:

```bash
bundle_lock_path=mktemp; bundle lock --print > $bundle_lock_path; diff $bundle_lock_path ./Gemfile.lock
```

```bash
a=$(bundle lock --print) ; b=$(cat Gemfile.lock); [ "$a" = "$b" ]
```

ankitkataria:

Fixes #6459

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -156,7 +156,7 @@ module Bundler
     method_option "path", :type => :string, :banner =>
       "Specify a different path than the system default ($BUNDLE_PATH or $GEM_HOME).#{" Bundler will remember this value for future installs on this machine" unless Bundler.feature_flag.forget_cli_options?}"
     method_option "deployment", :type => :boolean, :required => false, :default => false, :banner =>
-      "To Check the difference in the existing and new gemfile if deployed"
+      "Check the difference in the existing and new lockfile when deployed"
     map "c" => "check"
     def check
       require "bundler/cli/check"
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -155,6 +155,8 @@ module Bundler
       "Use the specified gemfile instead of Gemfile"
     method_option "path", :type => :string, :banner =>
       "Specify a different path than the system default ($BUNDLE_PATH or $GEM_HOME).#{" Bundler will remember this value for future installs on this machine" unless Bundler.feature_flag.forget_cli_options?}"
+    method_option "deployment", :type => :boolean, :required => false, :default => false, :banner =>
+      "To Check the difference in the existing and new gemfile if deployed"
     map "c" => "check"
     def check
       require "bundler/cli/check"
--- a/lib/bundler/cli/check.rb
+++ b/lib/bundler/cli/check.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "tempfile"
+
 module Bundler
   class CLI::Check
     attr_reader :options
--- a/lib/bundler/cli/check.rb
+++ b/lib/bundler/cli/check.rb
@@ -14,12 +14,17 @@ module Bundler
       begin
         definition = Bundler.definition
 
+        ui = Bundler.ui
+        Bundler.ui = UI::Silent.new
+
         if options[:deployment]
           definition.resolve_remotely!
-          final_lockfile = definition.to_lock.split(/(?<=\n)/).to_set
-          current_lockfile = IO.readlines(Bundler.default_lockfile).to_set
-          diff = (current_lockfile ^ final_lockfile).to_a.join
+          final_lockfile = definition.to_lock
+          current_lockfile = IO.readlines(Bundler.default_lockfile).join
+          diff = Bundler::Diff.new(current_lockfile, final_lockfile)
         end
+
+        Bundler.ui = ui
 
         definition.validate_runtime!
         not_installed = definition.missing_specs
--- a/lib/bundler/cli/check.rb
+++ b/lib/bundler/cli/check.rb
@@ -13,6 +13,14 @@ module Bundler
 
       begin
         definition = Bundler.definition
+
+        if options[:deployment]
+          definition.resolve_remotely!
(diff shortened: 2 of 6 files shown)


lib/bundler/cli.rb
```diff
@@ -160,6 +160,8 @@ def init
       "Use the specified gemfile instead of Gemfile"
     method_option "path", :type => :string, :banner =>
       "Specify a different path than the system default ($BUNDLE_PATH or $GEM_HOME).#{" Bundler will remember this value for future installs on this machine" unless Bundler.feature_flag.forget_cli_options?}"
+    method_option "deployment", :type => :boolean, :required => false, :default => false, :banner =>
+      "To Check the difference in the existing and new gemfile if deployed"
```
agrim123: `Check the difference in the existing and new lockfile when deployed`

ankitkataria:
@deivid-rodriguez @segiddins is there anything left to be done in this one?

deivid-rodriguez:
Hei @ankitkataria. I feel the "diff implementation" is quite complicated (not that you made it complicated, it is complicated by itself). As I said in the feature proposal, I would prefer to start small with a tiny feature that only detects the mismatch and suggests a solution.

@colby-swandale @indirect How do you feel about this?

lib/bundler/diff.rb
```diff
@@ -0,0 +1,71 @@
...
+      @string2 = string2
+    end
+
+    def diff
+      @diff ||= begin
+          @paths = [tempfile(@string1), tempfile(@string2)]
+
+          cmd = format '"%s" %s %s', diff_bin, "-U10000", @paths.map {|s| %("#{s}") }.join(" ")
+          diff = `#{cmd}`
+          diff.force_encoding("ASCII-8BIT") if diff.respond_to?(:valid_encoding?) && !diff.valid_encoding?
```
deivid-rodriguez: This sounds like old times legacy code. I would remove it.

lib/bundler/diff.rb
```diff
@@ -0,0 +1,71 @@
...
+          diff
+        end
+    ensure
+      # unlink the tempfiles explicitly now that the diff is generated
+      if defined? @tempfiles # to avoid Ruby warnings about undefined ins var.
+        Array(@tempfiles).each do |t|
+          begin
+            # check that the path is not nil and file still exists.
+            # REE seems to be very agressive with when it magically removes
```
deivid-rodriguez: REE is a very old implementation. I think we can just remove the file without any rescue or checks?

lib/bundler/diff.rb
```diff
@@ -0,0 +1,71 @@
...
+      regexp = /^(--- "?#{@paths[0]}"?|\+\+\+ "?#{@paths[1]}"?|@@|\\\\)/
+      @diff.split("\n").reject {|x| x =~ regexp }.map {|line| line + "\n" }.join
+    end
+
+  private
+
+    def tempfile(string)
+      t = Tempfile.new("diffy")
+      # ensure tempfiles aren't unlinked when GC runs by maintaining a
+      # reference to them.
```
deivid-rodriguez: Can this really happen, that a tempfile is removed suddenly while you're using it? I would remove all these hacks from the `diff` gem itself, and run its test suite to see if something breaks.

indirect: I think I would prefer an implementation that holds references to the tempfiles until the diff is generated, prints the diff, and then lets the tempfiles get deleted as part of garbage collection when the process is shutting down. It feels like a waste of code and effort to explicitly unlink the tempfiles ourselves when Ruby is about to do it for us when the process ends.

lib/bundler/diff.rb
indirect: Ruby Enterprise Edition was discontinued in 2012, which is 7 years ago now. I agree that we should not support REE in the next version of Bundler. 🙂

lib/bundler/diff.rb
indirect: I agree that this seems like code intended to support older versions of Ruby. Can you either explain what this line is needed for in Ruby 2.3+ or remove it?

indirect (changes requested):

lib/bundler/cli/check.rb
```diff
@@ -13,6 +15,19 @@ def run
...
       begin
         definition = Bundler.definition
+
+        ui = Bundler.ui
+        Bundler.ui = UI::Silent.new
+
+        if options[:deployment]
+          definition.resolve_remotely!
+          final_lockfile = definition.to_lock
+          current_lockfile = IO.readlines(Bundler.default_lockfile).join
```
segiddins: Maybe just `File.read` instead of joining all the lines?

lib/bundler/cli/check.rb
```diff
@@ -13,6 +15,19 @@ def run
...
+          diff = Bundler::Diff.new(current_lockfile, final_lockfile)
+        end
+
+        Bundler.ui = ui
```
segiddins: should this be in an `ensure`?

lib/bundler/diff.rb
```diff
@@ -0,0 +1,71 @@
...
+      @string1 = string1
+      @string2 = string2
+    end
+
+    def diff
+      @diff ||= begin
+          @paths = [tempfile(@string1), tempfile(@string2)]
+
+          cmd = format '"%s" %s %s', diff_bin, "-U10000", @paths.map {|s| %("#{s}") }.join(" ")
+          diff = `#{cmd}`
```
segiddins: Can we avoid using a shell?

PR #6911, opened 2019-01-16
Actually switch the default behavior of github sources.

### LINKED ISSUE #6910: GitHub HTTPS is not actually enabled by default in Bundler 2?
I just upgraded to Bundler 2, and tried cleaning my Gemfile to remove the workaround from https://github.com/bundler/bundler/issues/4978#issuecomment-272248627. However, it looks like Bundler is still using `git://...` urls instead of `https://...` urls for me.

I dug into the code in the `2.0.1` branch, and it looks like there's a leftover TODO about making this switch: https://github.com/bundler/bundler/blob/v2.0.1/lib/bundler/dsl.rb#L307-L308.

I dug into the code history a bit to try to understand this better, and things get a little weird. My understanding:

1. In df2956dcc52fbf241988bd51ff6cbf5244198995, @indirect changed the default from `git://` to `https://`.
2. 1cc3d97a7f71726e16f57462cba2814b1237123e reverted that change.
3. 61e00d657174e84f3b27570becdfe9bccc3b3a0c added a config option and warned that the default for this option would be changing in version 2.
3. Then, in d9844e8d0f8d365c8750cf214a678602f29e4926, a new `github_https?` method was introduced, but the method isn't actually used anywhere except tests. The commit message claims that this changes the default behavior, but it doesn't. It looks like that commit was *trying* to double commit the code from https://github.com/bundler/bundler/pull/6798/files, but it's missing this key part:

image

I'm pretty confident think the right fix is to add this missing line of code to the `2.*` release series. I'll send in a PR implementing this soon.

jfly:
This fixes #6910. I also took this opportunity to remove some other code
that the comments imply should have been removed in 2.0, but it looks
like never actually got removed.

### DIFF
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -290,25 +290,11 @@ module Bundler
         warn_deprecated_git_source(:github, <<-'RUBY'.strip, 'Change any "reponame" :github sources to "username/reponame".')
 "https://github.com/#{repo_name}.git"
         RUBY
-        # It would be better to use https instead of the git protocol, but this
-        # can break deployment of existing locked bundles when switching between
-        # different versions of Bundler. The change will be made in 2.0, which
-        # does not guarantee compatibility with the 1.x series.
-        #
-        # See https://github.com/bundler/bundler/pull/2569 for discussion
-        #
-        # This can be overridden by adding this code to your Gemfiles:
-        #
-        #   git_source(:github) do |repo_name|
-        #     repo_name = "#{repo_name}/#{repo_name}" unless repo_name.include?("/")
-        #     "https://github.com/#{repo_name}.git"
-        #   end
         repo_name = "#{repo_name}/#{repo_name}" unless repo_name.include?("/")
-        # TODO: 2.0 upgrade this setting to the default
-        if Bundler.settings["github.https"]
-          Bundler::SharedHelpers.major_deprecation 3, "The `github.https` setting will be removed"
+        if Bundler.feature_flag.github_https?
           "https://github.com/#{repo_name}.git"
         else
+          Bundler::SharedHelpers.major_deprecation 3, "Setting `github.https` to false is deprecated and won't be supported in the future."
           "git://github.com/#{repo_name}.git"
         end
       end
--- a/spec/bundler/dsl_spec.rb
+++ b/spec/bundler/dsl_spec.rb
@@ -44,7 +44,7 @@ RSpec.describe Bundler::Dsl do
 
       it "converts :github to :git" do
         subject.gem("sparks", :github => "indirect/sparks")
-        github_uri = "git://github.com/indirect/sparks.git"
+        github_uri = "https://github.com/indirect/sparks.git"
         expect(subject.dependencies.first.source.uri).to eq(github_uri)
       end
 
@@ -62,7 +62,7 @@ RSpec.describe Bundler::Dsl do
 
       it "converts 'rails' to 'rails/rails'" do
         subject.gem("rails", :github => "rails")
-        github_uri = "git://github.com/rails/rails.git"
+        github_uri = "https://github.com/rails/rails.git"
         expect(subject.dependencies.first.source.uri).to eq(github_uri)
       end
 
@@ -253,7 +253,7 @@ RSpec.describe Bundler::Dsl do
         end
 
         subject.dependencies.each do |d|
-          expect(d.source.uri).to eq("git://github.com/spree/spree.git")
+          expect(d.source.uri).to eq("https://github.com/spree/spree.git")
         end
       end
     end
--- a/spec/other/major_deprecation_spec.rb
+++ b/spec/other/major_deprecation_spec.rb
@@ -209,8 +209,8 @@ The :github git source is deprecated, and will be removed in Bundler 3.0. Change
         subject.gem("sparks", :github => "indirect/sparks")
       end
 
-      it "upgrades to https on request" do
-        Bundler.settings.temporary "github.https" => true
+      it "downgrades to http on request" do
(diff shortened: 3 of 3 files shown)


lib/bundler/dsl.rb
```diff
@@ -490,7 +469,7 @@ def check_primary_source_safety(source_list)
     end
 
     def warn_deprecated_git_source(name, replacement, additional_message = nil)
-      # TODO: 2.0 remove deprecation
+      # TODO: 3.0 remove deprecation
```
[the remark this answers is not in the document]
deivid-rodriguez: I think we should instead keep the sources, and enable the deprecations instead (`Bundler::SharedHelpers.major_deprecation 2`). I can do that in #6933, since that's what that PR is about.

lib/bundler/dsl.rb
```diff
@@ -304,34 +304,13 @@ def add_git_sources
         #     "https://github.com/#{repo_name}.git"
         #   end
         repo_name = "#{repo_name}/#{repo_name}" unless repo_name.include?("/")
-        # TODO: 2.0 upgrade this setting to the default
-        if Bundler.settings["github.https"]
+        if Bundler.feature_flag.github_https?
```
deivid-rodriguez: :+1:

deivid-rodriguez:
In general, I'm not sure of how we should do this transition, and other similar ones.

In particular, this one is even more complicated because we are removing two different things: the source itself, and using http by default for it. If we are removing the source altogether, then the setting about using https by default seems redundant? In my opinion, the sources are fine and handy, as long as they use a secure default protocol?

Thoughts?

jfly:
Thanks for the feedback, @deivid-rodriguez! Sorry for the delayed response, I was travelling.

I have simplified this PR to focus just on the HTTPS default switch. I'll let you take care of the rest in #6933.

lib/bundler/dsl.rb
```diff
@@ -290,25 +290,11 @@ def add_git_sources
...
-        #     "https://github.com/#{repo_name}.git"
-        #   end
         repo_name = "#{repo_name}/#{repo_name}" unless repo_name.include?("/")
-        # TODO: 2.0 upgrade this setting to the default
-        if Bundler.settings["github.https"]
-          Bundler::SharedHelpers.major_deprecation 3, "The `github.https` setting will be removed"
+        if Bundler.feature_flag.github_https?
           "https://github.com/#{repo_name}.git"
         else
+          Bundler::SharedHelpers.major_deprecation 3, "The `github.https` setting will be removed"
```
deivid-rodriguez: I think deprecating this setting now needs to be moved to https://github.com/bundler/bundler/blob/66e4215dbf87fedf2fdff3532a0742340c4e02f2/lib/bundler/feature_flag.rb, since here we can't be sure whether we are using the default value for the setting, or whether the user is actually setting it. The deprecation warning needs to be displayed only if the user is actually configuring the setting.

jfly: Well, I moved this because I happen to know that the default for this setting is now true, so if we end up in this `else` block, then it means somebody changed it away from the default. This would not detect the case where someone has set it to "true", but that also doesn't seem terribly important to warn people about.

Alternatively, I could refactor `feature_flag.rb` a bit to add a warning if `value` is not nit here: https://github.com/bundler/bundler/blob/66e4215dbf87fedf2fdff3532a0742340c4e02f2/lib/bundler/feature_flag.rb#L22, but that would be a bit of shuffling, and I'm not sure it's worth the effort.

Whatever you think is best!

deivid-rodriguez: Maybe the easiest (and still correct) thing to do for now is to change the deprecation text to: "Setting the `github.https` to false is deprecated and won't be supported in the future", or something like that?

jfly: I like that! I've made this change.

lib/bundler/dsl.rb
```diff
@@ -290,25 +290,11 @@ def add_git_sources
...
-        # does not guarantee compatibility with the 1.x series.
-        #
-        # See https://github.com/bundler/bundler/pull/2569 for discussion
-        #
-        # This can be overridden by adding this code to your Gemfiles:
-        #
-        #   git_source(:github) do |repo_name|
-        #     repo_name = "#{repo_name}/#{repo_name}" unless repo_name.include?("/")
-        #     "https://github.com/#{repo_name}.git"
-        #   end
```
deivid-rodriguez: I'm fine with removing this comment, since in my opinion it actually adds more confusion to the current situation...

jfly: I agree. Just to confirm: no further action required here?

deivid-rodriguez: Not from me!

I'll wait a bit to merge though in case there are other reviews or further discussion, specially since @segiddins approved this PR before I added my review.

deivid-rodriguez: Actually... Can you fix the TravisCI build?

jfly: Oh whoops! I didn't check how the build did. Should be fixed now, but I'll keep an eye on it.

deivid-rodriguez:
I have now have a plan on how to tackle these deprecations. For starters let's do this and I'll follow up from here!

deivid-rodriguez:
@bundlerbot r+

deivid-rodriguez:
@bundlerbot retry

jfly:
Sounds good, thanks!

deivid-rodriguez:
We don't seem to have a setup to merge PRs to 2-0-stable. So for the moment I'm going to close this and I'll backport upon the next release.

This is fixed (and your attribution kept) in #6975 anyways.

jfly:
Many thanks, @deivid-rodriguez!

PR #6922, opened 2019-01-23
Remove TODO label from default options of gemspec metadata

hsbt:

When invoking `bundle install` on project directory, It shows the error of bundler.

Bundler put the needless `TODO` label from gemspec template.

I removed TODO label from gemspec. They are not always provided at all times.

I understood this metadata was useful, But `TODO:` is unnecessary at first time of `bundle gem`.

### DIFF
--- a/lib/bundler/templates/newgem/newgem.gemspec.tt
+++ b/lib/bundler/templates/newgem/newgem.gemspec.tt
@@ -24,8 +24,8 @@ Gem::Specification.new do |spec|
   spec.metadata["allowed_push_host"] = "TODO: Set to 'http://mygemserver.com'"
 
   spec.metadata["homepage_uri"] = spec.homepage
-  spec.metadata["source_code_uri"] = "TODO: Put your gem's public repo URL here."
-  spec.metadata["changelog_uri"] = "TODO: Put your gem's CHANGELOG.md URL here."
+  spec.metadata["source_code_uri"] = "Put your gem's public repo URL here."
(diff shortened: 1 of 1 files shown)


ixti:
`TODO` drags attention that gemspecs needs to be fixed to avoid publishing gem with *Put your description* as gem's description.

niku:
Let me share my opinion to make this issue moving forward.

I received your messages following like:

- @hsbt "`bundle install` should work well even if a gem just created using `bundle gem`."
- @ixti "Publishing gem should avoid unless putting user descriptions to gemspecs."

Is it right? If so, I agree with both your opinions.

I think responsibility of "Publishing gem" is not a part of `bundler` command. It should be `gem push` command https://guides.rubygems.org/publishing/#publishing-to-rubygemsorg .

On the other hand, I'm worried if just removing `TODO`s in gemspecs which is generated from `bundle gem`, publishing gem will increase with no user descriptions.

So, I wonder if `gem push` avoided pushing gem that includes `TODO` s, we don't need to check `TODO` s in `bundler`.

How do you think?

hsbt:
I don't have a strong opinion about this. But It makes to reduce the developer experience with `bundle install` definitely.

deivid-rodriguez:
I understand the rationale of the TODO's, but I agree with @hsbt that they harm the developer experience in many cases. 

Maybe we can find a middle ground by prompting for this information when the gem is created, and adding optional flags to fill in the fields directly if given.

PR #6923, opened 2019-01-23
Ignore `Object =~` warnings.

hsbt:
### What was the end-user problem that led to this PR?

We faced a lot of warnings with `rake spec` of bundler repository.

### What was your diagnosis of the problem?

Ruby 2.7.0-dev added the warnings for `Object =~` comparison. 

### What is your fix for the problem, implemented in this PR?

I added the additional condition and bump a version of the development dependency.

PR #6924, opened 2019-01-23
Bump rake version to 12.0 on gemspec template

hsbt:

Related with https://github.com/bundler/bundler/pull/6923

rake-10.x displayed the warnings of `Object =~` with Ruby 2.7.0-dev.

Rake 10.0 is EOL. I'm a maintainer of Rake. I have no plan to avoid `Object =~` warnings with Rake 10.x.


hsbt:
@bundlerbot r+

PR #6925, opened 2019-01-23
Add bundle lock --update --bundler

alyssais:

The problem was that I couldn't generate a Bundler 2.0 lockfile on a machine that doesn't have the libraries available to be able to build native extensions, which is part of my normal workflow.

My diagnosis was that `bundle lock --update`, which I can usually use instead of `bundle update`, didn't support the `--bundler` switch like `bundle update` does.

My fix implements the `--bundler` flag for `bundle lock --update`.

I chose this fix because I think it makes sense for `bundle lock --update` to remain consistent with `bundle update`, except for not actually installing gems.

### DIFF
--- a/spec/commands/lock_spec.rb
+++ b/spec/commands/lock_spec.rb
@@ -360,3 +360,24 @@ RSpec.describe "bundle lock" do
     end
   end
 end
+
+RSpec.describe "bundle update --bundler" do
+  it "updates the bundler version in the lockfile without re-resolving" do
+    build_repo4 do
+      build_gem "rack", "1.0"
+    end
+
+    install_gemfile! <<-G
+      source "file:#{gem_repo4}"
+      gem "rack"
+    G
+    lockfile lockfile.sub(/(^\s*)#{Bundler::VERSION}($)/, '\11.0.0\2')
+
+    FileUtils.rm_r gem_repo4
+
+    bundle! :update, :bundler => true, :verbose => true
+    expect(the_bundle).to include_gem "rack 1.0"
+
+    expect(the_bundle.locked_gems.bundler_version).to eq v(Bundler::VERSION)
+  end
+end
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -638,6 +638,8 @@ module Bundler
       "If updating, do not allow any gem to be updated past latest --patch | --minor | --major"
     method_option "conservative", :type => :boolean, :banner =>
       "If updating, use bundle install conservative update behavior and do not allow shared dependencies to be updated"
+    method_option "bundler", :type => :string, :lazy_default => "> 0.a", :banner =>
+      "Update the locked version of bundler"
     def lock
       require "bundler/cli/lock"
       Lock.new(options).run
--- a/lib/bundler/cli/lock.rb
+++ b/lib/bundler/cli/lock.rb
@@ -23,7 +23,7 @@ module Bundler
       update = options[:update]
       if update.is_a?(Array) # unlocking specific gems
         Bundler::CLI::Common.ensure_all_gems_in_lockfile!(update)
-        update = { :gems => update, :lock_shared_dependencies => options[:conservative] }
+        update = { :gems => update, :lock_shared_dependencies => options[:conservative], :bundler => options[:bundler] }
       end
       definition = Bundler.definition(update)
 


alyssais:
For this to be useful, RubyGems will also need a patch:

```diff
diff --git a/lib/rubygems/bundler_version_finder.rb b/lib/rubygems/bundler_version_finder.rb
index e74baca1..e7148d3f 100644
--- a/lib/rubygems/bundler_version_finder.rb
+++ b/lib/rubygems/bundler_version_finder.rb
@@ -49,7 +49,8 @@ To install the missing version, run `gem install bundler:#{vr.first}`

   def self.bundle_update_bundler_version
     return unless File.basename($0) == "bundle".freeze
-    return unless "update".start_with?(ARGV.first || " ")
+    return if ARGV.first.nil?
+    return if %w[update lock].none? { |c| c.start_with?(ARGV.first) }
     bundler_version = nil
     update_index = nil
     ARGV.each_with_index do |a, i|
```

Makes sense to do it here first though, I think.

spec/commands/lock_spec.rb
```diff
@@ -360,3 +360,24 @@ def read_lockfile(file = "Gemfile.lock")
...
+
+    install_gemfile! <<-G
+      source "file:#{gem_repo4}"
+      gem "rack"
+    G
+    lockfile lockfile.sub(/(^\s*)#{Bundler::VERSION}($)/, '\11.0.0\2')
+
+    FileUtils.rm_r gem_repo4
+
+    bundle! :update, :bundler => true, :verbose => true
```
deivid-rodriguez: Should be actually testing `bundle lock --update --bundler` here? Maybe we can add a test similar to 

https://github.com/bundler/bundler/blob/d0103bf094fbde0f38955862872c245e847612d4/spec/commands/lock_spec.rb#L78-L85

but for `bundle lock --update --bundler` instead?

deivid-rodriguez (changes requested):

deivid-rodriguez:
Sorry for the delay @alyssais.

So... if `bundle lock --update` always updates the bundler version, why do we need this new flag? I didn't fully understand the usefulness in the "bundler version finder" side. I think the only reason for the patch you posted in https://github.com/bundler/bundler/pull/6925#issuecomment-457156864 would be to skip usages of this new option when selecting the bundler version to use, but if we don't need the new option, we don't need to patch the bundler version finder either... right?

alyssais:
So, currently, if I have Bundler 1.x and 2.x installed, and I want to update a lockfile from 1.x to 2.x, my only option is to run `bundle update --bundler`. If I run anything else, RubyGems' `BundlerVersionFinder` will intercept, and run Bundler 1.x, so Bundler won't be updated. There is currently no way to update the Bundler version of a lockfile without installing all the gems.

I should be able to run _something_ that has the same Bundler-updating effect, without installing all the gems in my Gemfile. Normally the command for this would be `bundle lock --update`, but because of `BundlerVersionFinder`, that will run Bundler 1.x and not update Bundler.

So, `BundlerVersionFinder` needs to be modified in some way to allow some command that can handle this case, hence the patch.

Now, initially I had planned to add `bundle lock --update --bundler` for consistency with `bundle update --bundler`, but it seems that, unlike `bundle update`, `bundle lock --update` _always_ updates Bundler anyway. So, there's no need for `--bundler`. There is still a need for `BundlerVersionFinder` to be modified to allow running `bundle lock` as well as `bundler update`, though.

So, _I think_, what should happen here is I should close this PR, and just send the `BundlerVersionFinder` patch to RubyGems. Does that sound right?

It is, however, extremely possible I've got something wrong here – I find the implementation of autoswitching across Bundler and RubyGems quite difficult to keep track of in my head, especially given it's been a month since I last had to think about it, and several months since my initial work on this.

deivid-rodriguez:
@alyssais That makes sense to me, I understand now :+1:.

What the bundler version finder in rubygems seems to be doing is autoswitching on `bundle update`_unless_ you're running `bundle update --bundler`. So I think it makes sense to add the flag here as proposed initially, that only updates the bundler version in the lock file, and explicitly guard for that case in the bundler version finder.

I plan to review some issues with the autoswitching like the one you mentioned about `bundle lock --update` updating the locked bundler version, but I think this PR is good to go once we get it properly tested. I think it's better to keep this consistent, and we can keep the autoswitching in rubygems as it's working now for `bundle update`, without excluding all `bundle lock` variants from it.

> especially given it's been a month since I last had to think about it, and several months since my initial work on this.

I'm so sorry I've been so slow getting back to you here... :pray:.

alyssais:
> I plan to review some issues with the autoswitching like the one you mentioned about `bundle lock --update` updating the locked bundler version, but I think this PR is good to go once we get it properly tested.

Given the `bundle lock --update` behaviour you mentioned, I don't think it will be possible to write a test for `bundle lock --update --bundler` that doesn't also pass for `bundle lock --update` – I'm guessing this is okay for now?

> I'm so sorry I've been so slow getting back to you here... 🙏.

No worries! Wasn't trying to criticise (I've been slow too), just to explain why my replies are so confused!

deivid-rodriguez:
> Given the bundle lock --update behaviour you mentioned, I don't think it will be possible to write a test for bundle lock --update --bundler that doesn't also pass for bundle lock --update – I'm guessing this is okay for now?

I would expect the new flag to _only_ update the bundler version. If that's the idea, it should be possible to write a test that includes some other potential updates that are skipped when running `bundle lock --update --bundler`?

alyssais:
> I would expect the new flag to _only_ update the bundler version.

I think I would expect `bundle lock --bundler` to do this, but `bundle lock --update --bundler` to behave as `bundle lock --update` does now.

The reason being that `bundle update --bundler` updates all your gems in addition to Bundler, so `bundle lock --update --bundler` should be consistent with that.

deivid-rodriguez:
That makes sense, yeah :+1:. No problem if `bundle lock --update --bundler`, and `bundle lock --update` can't currently be distinguished in the tests.

alyssais:
Cool. I'll see what I can do. Thanks for bearing with me!

deivid-rodriguez:
Hello again @alyssais! Did you find time to work on this?

PR #6926, opened 2019-01-23
Move Rubocop development dependency into gemspec

alyssais:

The problem was that tools like RubyGems.org or Bundix that use the gemspec as the source of truth for dependencies didn't pick up on the development dependency on Rubocop.

My diagnosis was that the reason for this, which was that Rubocop had a minimum Ruby version of 2.0.0, was no longer an issue, since Bundler 2 doesn't support Rubies that old anyway.

My fix was to move the Rubocop requirement into the gemspec.

I chose this fix because it addresses the problem I was having, feels right, and the old workaround doesn't look like it's necessary any more.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -43,7 +43,6 @@ namespace :spec do
     deps = Hash[bundler_spec.development_dependencies.map do |d|
       [d.name, d.requirement.to_s]
     end]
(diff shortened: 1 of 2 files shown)


hsbt:
@bundlerbot r+

PR #6927, opened 2019-01-23
Stop gemspec requiring dependencies under Bundler

alyssais:

The problem was that I couldn't use Bundler to set up an environment for hacking on Bundler, making it compatible with other Bundler-derived tools like Bundix.

My diagnosis was that the gemspec requires bundler/version, which then requires rubygems. When running under Bundler, requiring rubygems tries to load all of Bundler's development dependencies.

My fix is to ignore missing development dependencies in bundler/version, so that it can be loaded by the gemspec under Bundler.

I chose this fix because I don't see any other ones, and I don't think this should have any impact outside of this edge case.

### DIFF
--- a/lib/bundler/version.rb
+++ b/lib/bundler/version.rb
@@ -14,6 +14,9 @@ module Bundler
       require "rubygems"
     rescue LoadError
       return
+    rescue StandardError => e
+      # We don't need Bundler's dependencies solely for VERSION in the gemspec.
+      raise unless defined?(Bundler::GemNotFound) && e.is_a?(Bundler::GemNotFound)
     end
     return unless bundler_spec = Gem.loaded_specs["bundler"]
     return if bundler_spec.version == VERSION


segiddins:
I’m not sure this is safe — we later call methods on `Gem` in this file, so we need to have `rubygems.rb` loaded

alyssais:
It should be obvious through an undefined constant or method error, right? It worked fine in by testing, and would definitely have executed `Gem.loaded_specs` at least?

deivid-rodriguez:
@alyssais Can you provide reproduction steps for this issue? Using bundler for hacking on bundler itself sounds... interesting :smiley:

alyssais:
Sure!

Here's a Gemfile:

```ruby
source "https://rubygems.org"
gemspec
```

I really just want to use this to install the gems. The reason why I want to do this is a little weird, but it has to do with NixOS, and specifically its [Bundix](https://github.com/manveru/bundix) tool. I can go into more detail if required, but I don't think it's that relevant.

If I run `bundle lock`, with that Gemfile, I get this error:

> ```
> [!] There was an error parsing `Gemfile`:
> [!] There was an error while loading `bundler.gemspec`: Could not find gem 'automatiek (~> 0.1.0)' in any of the gem sources listed in your Gemfile.. Bundler cannot continue.
> 
>  #  from /bundler/bundler.gemspec:5
>  #  -------------------------------------------
>  #  begin
>  >    require File.expand_path("../lib/bundler/version", __FILE__)
>  #  rescue LoadError
>  #  -------------------------------------------
> . Bundler cannot continue.
> 
>  #  from /bundler/Gemfile:2
>  #  -------------------------------------------
>  #  source "https://rubygems.org"
>  >  gemspec
>  #  -------------------------------------------
> ```

But, with my fix applied, I can successfully install the gems.

deivid-rodriguez:
@alyssais I created #7062 that I think should also fix this problem!

deivid-rodriguez:
I'm pretty sure #7062 is a better approach to this, and I also tried that PR for your situation and it worked, so I'll close this, ok?

Issue #5994, opened 2017-08-31
Add a setting for default update level


chrismo:
https://github.com/bundler/bundler/blob/57f5ef656bb1845079225dde42b2d91ecb07ee2b/lib/bundler/cli.rb#L256-L257

These lines show the options, we may need to do a little more work to make this happen - but I think it's just one location to do this. 

I can take this one on (if we agree we still want to do this - @indirect @segiddins)

indirect:
@chrismo I'm not sure if I'm qualified to weigh in here... for me, what to update is a gem-by-gem thing. I don't even want Rails patch level changes unless I change my Gemfile. I (usually) watch minor releases from gems that are 1.0+. I (usually) want only patch releases from gems that are < 1.0. What are the use cases motivating the change in defaults?

indirect:
After some discussion, we arrived at a loose consensus: we will implement this change as a configuration setting, allowing users who want a default of `--patch` to opt-in. In the future, based on usage and other feedback, we may eventually default that setting to `on` rather than `off`. 👍

ankitkataria:
@indirect @segiddins Can I work on this issue?

segiddins:
Go for it!

ankitkataria:
What I understood from the issue is that we have to implement a configuration option for patch level changes. It'll be off by default and can be turned to on if required.
Am I correct? @segiddins

colby-swandale:
Feature has been merged. Closing.

PR #6931, opened 2019-01-24
Add patch option in bundle config

ankitkataria:

Issue #5994 

As mentioned by @indirect I added a `patch` option in the bundler config, which by default is false. If set to true, patch level update are preferred if no update levels are specified. 

A patch level config option which can be opted into.

### DIFF
--- a/man/bundle-config.ronn
+++ b/man/bundle-config.ronn
@@ -220,8 +220,6 @@ learn more about their operation in [bundle install(1)](bundle-install.1.html).
 * `only_update_to_newer_versions` (`BUNDLE_ONLY_UPDATE_TO_NEWER_VERSIONS`):
    During `bundle update`, only resolve to newer versions of the gems in the
    lockfile.
-* `patch` (BUNDLE_PATCH):
-   Whether to use patch update level, by default set to false.
 * `path` (`BUNDLE_PATH`):
    The location on disk where all gems in your bundle will be located regardless
    of `$GEM_HOME` or `$GEM_PATH` values. Bundle gems not found in this location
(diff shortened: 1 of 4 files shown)


spec/commands/update_spec.rb
```diff
@@ -809,6 +809,15 @@
       G
     end
 
+    context "with patch set as default update level in config" do
+      it "should do a patch level update" do
+        bundle! "config --local patch true"
+        bundle! "update --patch foo"
```
greysteil: This shouldn't include the `--patch`, right?

```suggestion
        bundle! "update foo"
```

man/bundle-config.ronn
```diff
@@ -220,6 +220,8 @@ learn more about their operation in [bundle install(1)](bundle-install.1.html).
 * `only_update_to_newer_versions` (`BUNDLE_ONLY_UPDATE_TO_NEWER_VERSIONS`):
    During `bundle update`, only resolve to newer versions of the gems in the
    lockfile.
+* `patch` (BUNDLE_PATCH):
+   Whether to use patch update level, by default set to false.
```
greysteil: Coming to the conversation fresh, this description doesn't give me enough information. Perhaps the option could be renamed to `prefer_patch`, and the description could read something like:

```
Prefer updating only to next patch version during updates. Makes `bundle update` calls equivalent to `bundler update --patch`.
```

spec/commands/update_spec.rb
ankitkataria: Yes, corrected that. :sweat_smile:

man/bundle-config.ronn
ankitkataria: Yes, prefer_patch seems to be a better and the text you've given makes much more sense to a new user. I've updated that as well.

greysteil:
@chrismo are you also 👍 ?

indirect:
@bundlerbot r=greysteil :)

PR #6933, opened 2019-01-25
Restore 2.0 deprecations

deivid-rodriguez:

The problem was that deprecation messages for stuff we want to change in bundler 3 are not yet being displayed.

My diagnosis was that we need to make sure we enable deprecations as soon as possible.

My fix includes several improvements regarding deprecations:

* ~Change the deprecation strategy. Previously deprecations would be always hidden unless the user explicitly turn them on via configuration, or the user is running a "deprecation release", namely, a release with minor level version equal to 99. The new deprecation strategy is simpler: no deprecation releases, and deprecations are turned on by default unless the user turns them off.~ Moved to #6965 and targeted to master.

* Move deprecations back to bundler 2. We moved them to bundler 3, but I think it's fine to deprecate all the stuff in bundler 2, and actually change the behavior in bundler 3. Exceptions are:
  - Prefering gems.rb and deprecating Gemfile. Because I think we decided to delay this due to the ommipresence of Gemfile's everywhere.
  - The one about [executable mismatches](https://github.com/bundler/bundler/blob/2e74a1a41916c15a0dbc6d55e194d829986178d3/lib/bundler/rubygems_integration.rb#L468-L472), because I'm not sure I understand that one, how to reproduce it, and whether it should be a deprecation message or just a warning. 
  - The ones about https by default and deprecated sources (github, and others). Because it's partially handled in #6911 and I want to give that process some more thought. 

* ~Change deprecation message wording. Previous they would read "[DEPRECATED FOR 2.0] \<Message about the deprecation>\". I find the version in the message a bit confusing so I changed it to just "[DEPRECATED] \<Message about the deprecation>\".~ Moved to #6965 and targeted to master.

I included all of the improvements in this PR because just moving back the deprecation horizon didn't really fix what I wanted to fix, namely, get the stuff really deprecated in the sense that end users get actual deprecation messages.


deivid-rodriguez:
I'll now retarget this PR to master, including only the behavior changes (bullet points number 1 and 3 above).

deivid-rodriguez:
Moved bullet points 1 and 3 to #6965 and changed the description of this PR to reflect that.

deivid-rodriguez:
We'll most likely cut 2.1 from master, so I'm closing this.

PR #6934, opened 2019-01-28
Include gem name in native extension build directory path

joelnordell:

The problem was that when including multiple gems from the same git repo, and more than one of them has native extensions, Bundler would only build the _first_ one and the other native extensions would not get built.

Example Gemfile:

```
git "git@github.com:/foo/bar_and_baz.git", :branch => "master" do
  gem "bar"
  gem "baz"
end
```

(In this scenario, gems `bar` and `baz` are both located in the `bar_and_baz.git` repo, in different subdirectories, each with their own `gemspec` and both provide a native extension.)

Our diagnosis was that Bundler would build the first native extension in a path named for the whole repo, and when finished place a file there indicating that no more extensions should be built. Subsequently, when the 2nd gem is installed, the native extension is then skipped.

Our fix was to include the name of the subdir containing each gem in the `extension_dir` so that each separate gem within the repo is treated separately with regard to building their native extensions.

We chose this fix because it solved the problem!

---

Thanks to @crugg for digging into this and coming up with the fix.

### DIFF
--- a/lib/bundler/rubygems_ext.rb
+++ b/lib/bundler/rubygems_ext.rb
@@ -80,7 +80,8 @@ module Gem
       alias_method :rg_extension_dir, :extension_dir
       def extension_dir
         @bundler_extension_dir ||= if source.respond_to?(:extension_dir_name)
-          File.expand_path(File.join(extensions_dir, source.extension_dir_name))
+          unique_extension_dir = [source.extension_dir_name, File.basename(full_gem_path)].uniq.join("-")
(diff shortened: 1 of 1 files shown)


joelnordell:
I would like to add a test covering this scenario (by extending the `spec/install/global_cache_spec.rb` test to also build two more extensions provided by two gems that are both in the same git repo under different subdirs) before this is merged.

segiddins:
Will this require rebuilding all existing extensions?

deivid-rodriguez:
@joelnordell Are you still planning to add that test?

PR #6938, opened 2019-01-31
Travis to Xenial, rake install RG 2.7.8

MSP-Greg:

1) Travis testing is using Trusty/14.04/OpenSSL 1.0.1

2) Rakefile - gem update --system 2.6.13

as above

fix above

I chose this fix because...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: ruby
+dist: xenial
 script: rake spec:travis
 before_script:
   - travis_retry rake -E 'module ::Bundler; VERSION = "0.0.0"; end' override_version
--- a/Rakefile
+++ b/Rakefile
@@ -73,7 +73,7 @@ namespace :spec do
       sh "sudo apt-get install graphviz -y 2>&1 | tail -n 2"
 
       # Install the gems with a consistent version of RubyGems
-      sh "gem update --system 2.6.13"
+      sh "gem update --system 2.7.8 --no-document --conservative"
 
       $LOAD_PATH.unshift("./spec")
       require "support/rubygems_ext"


MSP-Greg:
Closing, see PR #6952 Bump TravisCI build to Xenial

deivid-rodriguez:
If you want, you can rebase and reopen, leaving the rubygems update, and I can investigate why it's failing directly in this branch.

MSP-Greg:
@deivid-rodriguez

master is green, so I'm okay.  I'm biased towards latest-greatest, so I updated from 2.6.13 to 2.7.8.  The failure seemed like an odd one anyway.

MSP-Greg:
@deivid-rodriguez

Took another look at this.  I replaced:

https://github.com/bundler/bundler/blob/1a63384c807f164bfc2fe26ce34ac781a974c7e6/Rakefile#L75-L76

With:
```ruby
# Install the gems with a consistent version of RubyGems, except for
# Ruby 2.7+ and RubyGems 2.7.8 due to Object#=~ being deprecated
if Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.7.0.a")
  sh("gem update --system 2.7.8")
end
```

This removes most of the test failures (from 25 to 8), as shown in 'lib/ruby/site_ruby/2.7.0/rubygems/version.rb:212', examples [here](https://travis-ci.org/bundler/bundler/jobs/490971393#L800).  I believe the rest of the failures are due to `etc` being made a gem, and some other odd version issues.

The thing I haven't had a chance to look at is that the job has 'RGV=master' set, so why currently is it using a version of `version.rb` that's from 2.6.13?  The fix for this in RG happened in November...

deivid-rodriguez:
> The thing I haven't had a chance to look at is that the job has 'RGV=master' set, so why currently is it using a version of `version.rb` that's from 2.6.13? The fix for this in RG happened in November...

I noticed this when testing a bug fix I'm working on. It seems unexpected, I'm going to investigate.

deivid-rodriguez:
@MSP-Greg I fixed this in #6962, can you confirm?

MSP-Greg:
@deivid-rodriguez

Thanks for tracking this down.  Reading your description in the PR, it doesn't seem like the issue is order dependent. JFYI, I'm not really familiar with the bundler test suite...

I grabbed the PR, changed:
https://github.com/bundler/bundler/blob/cb5883d00b608c8672b4e360226641ebd4838b66/Rakefile#L72-L73

to use 2.6.14, which caused any use of the 'normal' RubyGems install in trunk to show the warnings from the `Object#=~` deprecation caused by `lib/ruby/site_ruby/2.7.0/rubygems/version.rb:212`.

Ran it in my fork, and there was no mention of the file/location in the trunk log.

Put simply, I can confirm that the PR fixes the issue.

deivid-rodriguez:
Yeah! Thanks for confirming!

PR #6939, opened 2019-01-31
Multiple without-group support for bundler list

denispasin:

The problem was that I needed to list all gems and dependencies that weren't in group :development or :test (or both of the together) but `bundler list --without-group` was only accepting one group at a time.

My diagnosis was I should just make it work for multiple groups

My fix is I split groups by `,` in args.

I chose this fix because it's simple and look like you can find in other cli.

I have two questions: Should I take care of the case where you have multiple `--without-group` specified ? Should I support `--without-groups` ?

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -314,7 +314,7 @@ module Bundler
       desc "list", "List all gems in the bundle"
       method_option "name-only", :type => :boolean, :banner => "print only the gem names"
       method_option "only-group", :type => :string, :banner => "print gems from a particular group"
-      method_option "without-group", :type => :string, :banner => "print all gems expect from a group"
+      method_option "without-group", :type => :string, :banner => "print all gems expect from 1 or multiple groups"
       method_option "paths", :type => :boolean, :banner => "print the path to each gem in the bundle"
       def list
         require "bundler/cli/list"
--- a/lib/bundler/cli/list.rb
+++ b/lib/bundler/cli/list.rb
@@ -31,8 +31,16 @@ module Bundler
 
   private
 
+    def without_groups
+      @without_groups ||= @options["without-group"].split(/\s*,\s*/)
+    end
+
     def verify_group_exists(groups)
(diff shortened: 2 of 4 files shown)


lib/bundler/cli.rb
```diff
@@ -314,7 +314,7 @@ def show(gem_name = nil)
       desc "list", "List all gems in the bundle"
       method_option "name-only", :type => :boolean, :banner => "print only the gem names"
       method_option "only-group", :type => :string, :banner => "print gems from a particular group"
-      method_option "without-group", :type => :string, :banner => "print all gems expect from a group"
+      method_option "without-group", :type => :string, :banner => "print all gems expect from 1 or multiple groups"
```
colby-swandale: how does `a given set of groups` sound instead?

lib/bundler/cli/list.rb
```diff
@@ -45,13 +53,14 @@ def filtered_specs_by_groups
...
-          groups.reject {|g| g == @options["without-group"].to_sym }
+          without_groups.inject(groups) do |memo, without_group|
+            memo.reject {|g| g == without_group.to_sym }
+          end
         elsif @options["only-group"]
           groups.select {|g| g == @options["only-group"].to_sym }
         else
           groups
         end.map(&:to_sym)
-
```
colby-swandale: Please don't remove the spacing. It helps with readability.

lib/bundler/cli/list.rb
```diff
@@ -45,13 +53,14 @@ def filtered_specs_by_groups
 
       show_groups =
         if @options["without-group"]
-          groups.reject {|g| g == @options["without-group"].to_sym }
+          without_groups.inject(groups) do |memo, without_group|
```
colby-swandale: I'm not sure what `memo` means here, can you clarify please?

esasse:
@denispasin hi, are you still interested in this PR? There are pending review questions. If we don’t get an update we may have to close it. Thanks.

PR #6941, opened 2019-02-04
Ignore to add bundler lib direcotry if it is same as rubylibdir

hsbt:

In ruby core, the bundled bundler was located same as rubylibdir. When user used the specific version of default gems like json, Bundler uses default gems under the rubylibdir, not a specific version.

Fixed  https://bugs.ruby-lang.org/issues/15469

See the details: https://bugs.ruby-lang.org/issues/15469#note-4

I avoid adding bundler directory from RUBYLIB environmental variable Because its directory was already added $LOAD_PATH.

### DIFF
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -3,6 +3,7 @@
 require "bundler/compatibility_guard"
 
 require "pathname"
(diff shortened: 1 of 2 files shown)


hsbt:
@bundlerbot r+

Issue #6937, opened 2019-01-30
Bundler 2 is already installed, still getting "You must use Bundler 2 or greater with this lockfile" error


stuartweir:
After installing Bundler following the steps from this [page](https://bundler.io/blog/2019/01/03/announcing-bundler-2.html), I cannot seem to run a test in rails.

The Gemfile.lock states at the bottom:
```
BUNDLED WITH
   2.0.1
```
Bundler Version:
```bash
$ bundler -v
Bundler version 2.0.1
```
**Rails Test failing even though Bundler 2 is installed**
```bash
$ bundle exec bin/rails test test/lib/my_rails_test_test.rb
You must use Bundler 2 or greater with this lockfile.
```

I have tried uninstalling and reinstalling Bundler:
```bash
$ gem uninstall bundler && gem install bundler
Successfully uninstalled bundler-2.0.1
Fetching bundler-2.0.1.gem
Successfully installed bundler-2.0.1
Parsing documentation for bundler-2.0.1
Installing ri documentation for bundler-2.0.1
Done installing documentation for bundler after 5 seconds
1 gem installed
```

Ruby version:
```bash
$ ruby -v
ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-darwin15]
```

I'm really not sure what seems to be the issue here, or why it's failing. Any help appreciated.

stuartweir:
Looks like it might be somehow falling back on an old bundler?
```bash
$ gem list bundler

*** LOCAL GEMS ***

bundler (2.0.1, default: 1.17.3, 1.7.3)
capistrano-bundler (1.5.0, 1.3.0, 1.2.0)```

not sure how to set/force bundler to use 2.0.1

**Comment by hewrin** [comment]
I am seeing a similar issue as well, if I delete Gemfile.lock and run bundle again, it says 
```
BUNDLED WITH
   1.16.0
```
Which is my old version of bundler

morgoth:
I have the same issue when using Rails and `bin/*` scripts in app.

Using bundler 2.0.1 with "2.0" mode on Rails 5.2.2 on Ruby 2.6.1

deivid-rodriguez:
Hello! I tried reproducing this with a fresh Rails app and could, can someone provide reproduction steps? Thanks!

deivid-rodriguez:
Reproduced now! Thanks! I'll look into it!

deivid-rodriguez:
[...]

PR #6948, opened 2019-02-07
Bump rubygems to 3.0.2 and ruby to 2.6.1 in TravisCI

deivid-rodriguez:

The problem was that I have some spec failures locally, and they don't seem to be reproduced in TravisCI.

My diagnosis was that either my environment is messed up or there's some issues with the latest rubygems + ruby combination.

My fix is to use latest rubygems and ruby in Travis, so I can double check whether the problem is just my environment or a real problem.

I chose this fix because it's always good practice to test against the latest versions of your dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ env:
   # We need to know if changes to rubygems will break bundler on release
   - RGV=master
   # Test the latest rubygems release with all of our supported rubies
-  - RGV=v3.0.1
+  - RGV=v3.0.2
 
 jobs:
   include:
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
Apparently TravisCI reproduces the same failure I'm getting locally. I'm going to investigate it.

deivid-rodriguez:
Ok, so.... It's apparently https://github.com/bundler/bundler/issues/6937. Suprisingly, 2.6.0 does not have this bug whereas 2.6.1 has it. As in, the `spec.files` property in the default bundler gemspec shipped with ruby is correct in 2.6.0, but incorrect in 2.6.1.

deivid-rodriguez:
Oh god, this was tough, it should be ready now.

deivid-rodriguez:
Merging away because I want to move on fast. Feel free to still comment and I'll do any further changes on a separate PR!

deivid-rodriguez:
@bundlerbot r+

PR #6949, opened 2019-02-08
Split test assertions by stream

deivid-rodriguez:

The problem was that once we enable deprecations, they will start getting printed to the screen and will affect existing output assertions, thus breaking unrelated tests.

My diagnosis was that we started printing errors to the stderr stream in bundler 2, but our `out` test helper [checks both streams combined](). That means deprecations affect the result of this helper, and any test runninng commands that print deprecations and checking for the exact output will be affected.

My fix is to consider only output in the `out` helper, and error in the `err` helper. That involves a simple change in the helper to return only `stdout` instead of `stdboth`, and fixing a bunch of tests that used to check `out` for errors and now need to check `err`.

I chose this fix because even if it was tedious, it simplifies enabling deprecations, and slightly improves the readability of tests.

### DIFF
--- a/spec/support/helpers.rb
+++ b/spec/support/helpers.rb
@@ -41,11 +41,11 @@ module Spec
     end
 
     def out
-      last_command.stdboth
+      last_command.stdout
     end
 
     def err
-      last_command.stderr
+      Bundler.feature_flag.error_on_stderr? ? last_command.stderr : last_command.stdout
     end
 
     def exitstatus
--- a/lib/bundler/cli/show.rb
+++ b/lib/bundler/cli/show.rb
@@ -24,7 +24,7 @@ module Bundler
           return unless spec
           path = spec.full_gem_path
           unless File.directory?(path)
-            Bundler.ui.warn "The gem #{gem_name} has been deleted. It was installed at:"
+            return Bundler.ui.warn "The gem #{gem_name} has been deleted. It was installed at: #{path}"
           end
         end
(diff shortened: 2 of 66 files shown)


spec/support/helpers.rb
```diff
@@ -41,7 +41,7 @@ def last_command
     end
 
     def out
-      last_command.stdboth
+      last_command.stdout
```
deivid-rodriguez: This is the key change in this PR. The rest is fixing the tests this change breaks.

deivid-rodriguez:
It was a bit harder than I thought but it surfaced some little issues with output streams that I fixed:

* The show command was printing half of the sentence of an error to stderr and the other half to stdout. https://github.com/bundler/bundler/pull/6949/commits/4e152685a9c81d661ea94351cfc5ba47713a61e9.
* One error message in the git proxy was not going through `Bundler.ui.warn`. https://github.com/bundler/bundler/pull/6949/commits/25da5ee198fef7bd5165a5df2b3ff4ba0fe44fa5.
* `bundler/setup` was printing errors to stdout. https://github.com/bundler/bundler/pull/6949/commits/8a64f1c5de9d7c3ce276b5530492fa3defc7337b.

colby-swandale:
This looks great!

colby-swandale:
@bundlerbot r+

deivid-rodriguez:
Wooooooow, thanks so much @colby-swandale!! :tada:

PR #6951, opened 2019-02-09
Fix symlink spec on ruby 2.6

deivid-rodriguez:

The problem was that we were skipping this test on ruby 2.6. Skipping tests on old versions is not much of a problem, but doing it on the latest version is opening the door for regressions. So I want to fix it.

My diagnosis was that since bundler is now a default gem, that bundler version is being activated by the spec, but since we are on a 2.0.1 locked bundle, the spec runs into a version mismatch error.

My fix is to enable the `Kernel.require` rubygems monkeypatch for this case (the `bundle` gem itself), that requires default gem files without activating the specific default version, and thus not forcing end users to use that. Since... that's exactly the problem here.

I chose this fix because it makes sense to me.

### DIFF
--- a/spec/runtime/setup_spec.rb
+++ b/spec/runtime/setup_spec.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "tmpdir"
+
 RSpec.describe "Bundler.setup" do
   describe "with no arguments" do
     it "makes all groups available" do
--- a/spec/runtime/setup_spec.rb
+++ b/spec/runtime/setup_spec.rb
@@ -807,6 +807,7 @@ end
     let(:gem_home) { Dir.mktmpdir }
     let(:symlinked_gem_home) { Tempfile.new("gem_home").path }
     let(:bundler_dir) { ruby_core? ? File.expand_path("../../../..", __FILE__) : File.expand_path("../../..", __FILE__) }
+    let(:full_name) { "bundler-#{Bundler::VERSION}" }
 
     before do
       FileUtils.ln_sf(gem_home, symlinked_gem_home)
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Nah, it's more complicated than that since the lockfile parser doesn't care about whether bundler is activated or not.

spec/spec_helper.rb
```diff
@@ -12,6 +12,13 @@ module Gem
   end
 end
 
+if ENV.select {|k, _v| k =~ /TRAVIS/ }.any?
```
olleolleolle: Is this the same thing?

```suggestion
if ENV.keys.any? { |k| k.start_with?('TRAVIS_') }
```

deivid-rodriguez: I think so, or we could even check for an exact value: https://docs.travis-ci.com/user/environment-variables/#default-environment-variables.

olleolleolle: By the way: lots of fanmail your way for all the great work and good cheer you bring to Bundler.

deivid-rodriguez:
Regarding this PR, I'm not sure what to do with it, since I'm not sure whether the current failure means that the original issue can actually happen again since `bundler` became a default gem, or whether it's just not as "easy" to test it.

The current "fix" is just a workaround consisting of moving `bundler` to its own isolated location, where manipulating the `$LOAD_PATH` for it won't affect the standard library and cause issues with `$LOAD_PATH` precedence. This kind of issues (other than this failing spec) have also appeared in the past, see:

* https://bugs.ruby-lang.org/issues/15469
* https://github.com/rubygems/rubygems/issues/2188

It is true that the first of these issues was already fixed by https://github.com/bundler/bundler/pull/6941, and that might also have fixed the other issue too. But maybe it's worth bringing this to ruby-core and ask whether they think moving bundler & rubygems to an isolated location (like `RbConfig::CONFIG["sitelibdir"]`) makes sense.

@hsbt Do you have any thoughts about this?

deivid-rodriguez:
Yay! I managed to finally get this spec passing in a reliable way, and also made sure that it is still a regression test for the issue that originated it: https://github.com/bundler/bundler/pull/6502. :tada:

Essentially nothing changed about the issue: bundler was removing itself from the LOAD_PATH when installed in a symlinked location, and we fixed that in the PR I just linked to. The manifestation of the problem did change when bundler started being shipped as a default gem. Before, requiring `bundler/setup` would just fail under this bug, but when `bundler` is present as a default gem, it would instead "leak" to the default bundler installation rather than failing.

So, I changed the test to verify that we don't leak to a bundler installation different from the current one.

spec/runtime/setup_spec.rb
```diff
@@ -813,32 +816,30 @@ def clean_load_path(lp)
...
 
-      File.open(File.join(specifications_dir, "bundler.gemspec"), "wb") do |f|
+      File.open(File.join(specifications_dir, "#{full_name}.gemspec"), "wb") do |f|
         f.write(gemspec)
       end
     end
 
-    # Can't make this pass on 2.6 since the ruby standard library has the same $LOAD_PATH
-    # entry as bundler (since it's a default gem)
-    it "should successfully require 'bundler/setup'", :ruby_repo, :ruby => "< 2.6" do
```
deivid-rodriguez: Ooops, unintentionally removed `:ruby_repo` from here. Will add it back.

deivid-rodriguez:
I can't believe this is green!! :tada:

deivid-rodriguez:
@bundlerbot r+

PR #6952, opened 2019-02-09
Bump TravisCI build to Xenial

deivid-rodriguez:

The problem was that CI in #6938 is failing, and I'm not sure why.

My diagnosis was that I should split the two changes made in that PR to find out the culprit.

My fix is to split the TravisCI update to xenial, because my guess is that this is not the culprit, so if the build is green, we can merge this and upgrade :tada: 

I chose this fix because it means no work for me, just delegate to TravisCI. :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: ruby
+dist: xenial
 script: rake spec:travis
 before_script:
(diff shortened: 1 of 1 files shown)


colby-swandale:
👍


@bundlerbot r+

colby-swandale:
@bundlerbot r+

PR #6954, opened 2019-02-10
Bump test MRI 2.3 version in TravisCI

deivid-rodriguez:

The problem was that we don't know whether bundler works against the latest patch level version of MRI 2.3.

My diagnosis was that we should test it.

My fix is to bump the 2.3 matrix entries in TravisCI config to 2.3.8. This MRI version comes with rubygems 2.7.8, so bump that too.

I chose this fix because it's the only one, given the CI setup we currently have.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,14 +52,14 @@ jobs:
       stage: linting
     # Ruby 2.5, Rubygems 2.7 and up
     - rvm: 2.5.3
-      env: RGV=v2.7.7
+      env: RGV=v2.7.8
       stage: test
     # Ruby 2.4, Rubygems 2.6 and up
     - rvm: 2.4.5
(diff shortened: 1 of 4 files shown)


deivid-rodriguez:
The two failures happen because they hit the bug fixed by https://github.com/rubygems/rubygems/pull/1527, so I restricted those specs to `rubygems` higher than 2.6.2 which is the first rubygems version that shipped that fix.

As per why those specs are currently passing on master... I have absolutely no idea, I can reproduce the failures with both rvm 2.3.7 and rvm 2.3.8 on fresh rvm installations. :man_shrugging:

deivid-rodriguez:
Moving on with these tiny CI PRs! :rocket:

deivid-rodriguez:
@bundlerbot r+

PR #6958, opened 2019-02-11
Remove rdiscount indirect development dependency

deivid-rodriguez:

The problem was that I was confused about the `rdiscount` dependency since I didn't see it used anywhere.

My diagnosis was that it's just an indirect dependency (through ronn).

My fix is to remove the dependency.

I chose this fix because it will no longer confuse devs into thinking we're actually using it directly.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -44,11 +44,8 @@ namespace :spec do
       [d.name, d.requirement.to_s]
     end]
 
-    # JRuby can't build ronn or rdiscount, so we skip that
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
This [was introduced back in ruby 1.9.2 times](https://github.com/bundler/bundler/commit/2868f492b6ddaba0f3913412d86a99a122443b9c), we no longer need to pin it.

deivid-rodriguez:
It should be one of those straightforward PRs, so going in!

@bundlerbot r+

PR #6959, opened 2019-02-12
Bump base rubygems version to 3.0.2

deivid-rodriguez:

The problem was that we are using old rubygems versions internally.

My diagnosis was that we should use the newest version of our own software.

My fix is to upgrade the rubygems version used in the Rakefile for development tasks.

I chose this fix because it's good for us.


deivid-rodriguez:
Seems straightforward, merging!

@bundlerbot r+

PR #6962, opened 2019-02-12
Don't leak to system rubygems during specs

deivid-rodriguez:

The problem was that sometimes one get warnings on the system rubygems files, or maybe when modifying the `tmp` copy of rubygems, the changes seem not applied.

My diagnosis was our specs sometimes use the system rubygems version, not the one controlled by the `RVG` environment variable.

My fix is to remove most occurrences of `:env => { :RUBYOPT => "-r#{spec_dir.join("support/hax")}" }` when spawning subprocesses. The reason is that in order to use the tmp copy of rubygems, our specs need to include `-I tmp/rubygems/lib` in the `RUBYOPT` environment variable. If we overwrite it in subprocesses, it won't use it. We can remove those however, since the "hax file" is already included in the main spec helper, while also keeping the previous value of `RUBYOPT`.

https://github.com/bundler/bundler/blob/189d56f2341d64ec5440242f42dbf84002be1238/spec/spec_helper.rb#L47

I chose this fix because it simplifies the spec setup while also making it more correct.

### DIFF
--- a/spec/commands/exec_spec.rb
+++ b/spec/commands/exec_spec.rb
@@ -38,7 +38,7 @@ RSpec.describe "bundle exec" do
       gem "rack"
     G
 
-    bundle "exec 'cd #{tmp("gems")} && rackup'", :env => { :RUBYOPT => "-r#{spec_dir.join("support/hax")}" }
+    bundle "exec 'cd #{tmp("gems")} && rackup'"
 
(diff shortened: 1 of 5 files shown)


deivid-rodriguez:
@MSP-Greg confirmed the fix, so merging!

@bundlerbot r+

PR #6963, opened 2019-02-13
Allow to `bundle exec` default gems

### LINKED ISSUE #6929: can't run bundle exec irb with Ruby 2.6
To reproduce...

```sh
> mkdir vanilla; cd vanilla
> touch Gemfile
> bundle update
The Gemfile specifies no dependencies
Resolving dependencies...
Bundle updated!
> gem list bundler

*** LOCAL GEMS ***

bundler (2.0.1, default: 1.17.2)
> cat Gemfile.lock
GEM
  specs:

PLATFORMS
  ruby
  x86_64-darwin-17

DEPENDENCIES

BUNDLED WITH
   2.0.1
> bundle exec irb
bundler: failed to load command: irb (/Users/jfrey/.rubies/ruby-2.6.0/bin/irb)
Gem::Exception: can't find executable irb for gem irb. irb is not currently included in the bundle, perhaps you meant to add it to your Gemfile?
  /Users/jfrey/.gem/ruby/2.6.0/gems/bundler-2.0.1/lib/bundler/rubygems_integration.rb:462:in `block in replace_bin_path'
  /Users/jfrey/.gem/ruby/2.6.0/gems/bundler-2.0.1/lib/bundler/rubygems_integration.rb:482:in `block in replace_bin_path'
  /Users/jfrey/.rubies/ruby-2.6.0/bin/irb:23:in `<top (required)>'
```

Similarly, I can't do the manual approach either, but I get a different error because it seems to want to load 1.17.2 regardless.

```sh
> irb
irb(main):001:0> require 'bundler/setup'
You must use Bundler 2 or greater with this lockfile.
# Note that this immediately exits irb

> irb
irb(main):001:0> require 'bundler'
=> true
irb(main):002:0> Bundler::VERSION
=> "1.17.2"
```

Probably related to #6913

deivid-rodriguez:

The problem was that since `irb` and others were moved to default gems, users cannot directly use them in a bundler context, unless they add it to their gemfiles. In my opinion, this completely defeats the purpose of default gems, and makes bundler degrade the user experience instead of making it better.

My diagnosis was that when bundler replaces the set of gems known to rubygems, it restricts the world to what's in the Gemfile (or resolved from it), and that doesn't include default gems.

My fix is to also include the set of default gems, unless they are already included in the gemfile dependencies.

I chose this fix because it's reasonably simple and I think it shouldn't affect how other parts of bundler function.

Fixes #6929.
Fixes https://bugs.ruby-lang.org/issues/15503.

### DIFF
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -512,7 +512,7 @@ module Bundler
         h
       end
 
-      Gem::Specification.send(:default_stubs, "*.gemspec").each do |stub|
+      Bundler.rubygems.default_stubs.each do |stub|
         default_spec = stub.to_spec
         default_spec_name = default_spec.name
         next if specs_by_name.key?(default_spec_name)
@@ -854,6 +854,16 @@ module Bundler
           Gem::Specification.stubs.find_all do |spec|
             spec.name == name
           end.map(&:to_spec)
+        end
+      end
+
+      if Gem::Specification.respond_to?(:default_stubs)
+        def default_stubs
+          Gem::Specification.default_stubs("*.gemspec")
+        end
+      else
+        def default_stubs
+          Gem::Specification.send(:default_stubs, "*.gemspec")
         end
       end
 
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -512,6 +512,15 @@ module Bundler
         h
       end
 
+      Gem::Specification.send(:default_stubs, "*.gemspec").each do |stub|
+        default_spec = stub.to_spec
+        default_spec_name = default_spec.name
+        next if specs_by_name.key?(default_spec_name)
+
+        specs << default_spec
+        specs_by_name[default_spec_name] = default_spec
+      end
+
       replace_gem(specs, specs_by_name)
       stub_rubygems(specs)
       replace_bin_path(specs, specs_by_name)
--- a/spec/commands/exec_spec.rb
+++ b/spec/commands/exec_spec.rb
@@ -140,6 +140,83 @@ RSpec.describe "bundle exec" do
     end
   end
 
+  context "with default gems" do
+    let(:system_gems_to_install) { [] }
+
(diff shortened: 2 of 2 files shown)


lib/bundler/rubygems_integration.rb
```diff
@@ -514,6 +514,15 @@ def replace_entrypoints(specs)
         h
       end
 
+      Gem::Specification.send(:default_stubs, "*.gemspec").each do |stub|
```
deivid-rodriguez: Using `send` here, because `default_stubs` is private. We should probably make it public in `rubygems`, or make a separate method public that calls `default_stubs` with the "*.gemspec" pattern.

colby-swandale:
I think this may have already been patched by @hsbt in ruby-src IIRC.

colby-swandale:
This is the issue i'm talking about https://bugs.ruby-lang.org/issues/15469

deivid-rodriguez:
They are similar issues, but different I think.

The issue you mention is about versions of default gems included in the Gemfile being ignored, and activating the default version instead. This was fixed by @hsbt in both ruby-core and https://github.com/bundler/bundler/pull/6941. Ruby core ticket: https://bugs.ruby-lang.org/issues/15469.

This one is about default gems not being activated at all unless they are included in the Gemfile. Ruby core ticket: https://bugs.ruby-lang.org/issues/15503.

hsbt:
I also agree with this issue was different from https://bugs.ruby-lang.org/issues/15469.

@deivid-rodriguez I'll check this later. Thanks.

lib/bundler/rubygems_integration.rb
hsbt: make sense. I applied it: https://github.com/rubygems/rubygems/pull/2675

deivid-rodriguez:
@hsbt I added a compatibility layer so that the new public method gets used if available, and otherwise we use `send`.

deivid-rodriguez:
I removed one unnecessary change in specs from this PR, and also rebased it! :+1:

deivid-rodriguez:
This has waited for a while now, and we're separately discussing other minor things about default gem handling when they don't even work at the moment... :sweat_smile:.

So merging away!

@bundlerbot r+

PR #6964, opened 2019-02-13
Dependency review

deivid-rodriguez:

The problem was that we were using some outdated test dependencies.

My diagnosis was that we can upgrade them, since the comments justifying them being pinned are no longer relevant.

My fix is to unpin the dependencies, and remove the comments no longer applying.

I chose this fix because it was the only fix I could think of.

### DIFF
--- a/spec/support/rubygems_ext.rb
+++ b/spec/support/rubygems_ext.rb
@@ -8,9 +8,7 @@ module Spec
     DEPS = begin
       deps = {
         # artifice doesn't support rack 2.x now.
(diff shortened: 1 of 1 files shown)


segiddins:
I think we should be able to use rack 2 now?

deivid-rodriguez:
According to the comment I left in there, the version of the `artifice` gem we're using does not support it...

segiddins:
Ah sorry I didn’t see that

@bundlerbot r+

deivid-rodriguez:
Thanks for reviewing @segiddins!

Issue #4855, opened 2016-08-09
1.99: the one that complains at you a lot


indirect:
- [x] hardcode the deprecations flags as turned on
- [x] deprecate `inject` in favor of `add`
- [ ] release final at the same time as 2.0

segiddins:
> hardcode the deprecations flags as turned on

This can just be changing `SharedHelpers.prints_major_deprecations?` to be `return false if !Bundler.settings[:major_deprecations] &&  !Bundler::VERSION.split(".").include?("99")` and can be done now

segiddins:
> deprecate inject in favor of add

this is dependent upon https://github.com/bundler/bundler/issues/4901

colby-swandale:
There won't be a 1.99, so this can be closed. I'll also be writing docs about major releases after 2.0 is out that will cover this as well.

PR #6965, opened 2019-02-14
Turn on deprecations by default

deivid-rodriguez:

The problem was we've had deprecations in place for a long time, but we've never displayed them to users by default. 

My diagnosis was that the current strategy doesn't work because:

* Printing deprecations as an optin feature almost never get enabled, so most users don't know about the stuff we'll be deprecating in the future.

* Printing deprecations in "deprecation releases" doesn't work well either, because it's unclear how long the deprecation release should last and thus how long we need to hold the final release (that will inhibit installation of the deprecation release).

My fix is to remove the concept of deprecation releases, and to add a feature flag for printing major deprecations that it's enabled by default.

As a extra related change, I also reworded the deprecation messages, because I find the current message "[DEPRECATED FOR 2.0] <Message about the deprecation>" a bit confusing because it's unclear what the version printed is referring to (deprecation horizon? current running version?), so I changed it to just "[DEPRECATED] <Message about the deprecation>".

I chose this fix because it (once released) finally makes it so that users will know about our deprecations.

### DIFF
--- a/lib/bundler/feature_flag.rb
+++ b/lib/bundler/feature_flag.rb
@@ -44,7 +44,6 @@ module Bundler
     settings_flag(:init_gems_rb) { bundler_2_mode? }
     settings_flag(:list_command) { bundler_2_mode? }
     settings_flag(:lockfile_uses_separate_rubygems_sources) { bundler_2_mode? }
-    settings_flag(:major_deprecations) { bundler_2_mode? }
     settings_flag(:only_update_to_newer_versions) { bundler_2_mode? }
     settings_flag(:path_relative_to_cwd) { bundler_2_mode? }
     settings_flag(:plugins) { @bundler_version >= Gem::Version.new("1.14") }
--- a/lib/bundler/feature_flag.rb
+++ b/lib/bundler/feature_flag.rb
@@ -44,6 +44,7 @@ module Bundler
     settings_flag(:init_gems_rb) { bundler_2_mode? }
     settings_flag(:list_command) { bundler_2_mode? }
     settings_flag(:lockfile_uses_separate_rubygems_sources) { bundler_2_mode? }
+    settings_flag(:major_deprecations) { bundler_2_mode? }
     settings_flag(:only_update_to_newer_versions) { bundler_2_mode? }
     settings_flag(:path_relative_to_cwd) { bundler_2_mode? }
     settings_flag(:plugins) { @bundler_version >= Gem::Version.new("1.14") }
--- a/lib/bundler/settings.rb
+++ b/lib/bundler/settings.rb
@@ -41,7 +41,6 @@ module Bundler
       init_gems_rb
       list_command
       lockfile_uses_separate_rubygems_sources
-      major_deprecations
       no_install
       no_prune
       only_update_to_newer_versions
@@ -52,6 +51,7 @@ module Bundler
       prefer_patch
       print_only_version_number
       setup_makes_kernel_gem_public
+      silence_deprecations
       silence_root_warning
       skip_default_git_sources
       specific_platform
@@ -76,7 +76,7 @@ module Bundler
     ].freeze
 
     DEFAULT_CONFIG = {
-      :major_deprecations => true,
+      :silence_deprecations => false,
       :disable_version_check => true,
       :prefer_patch => false,
       :redirect => 5,
--- a/lib/bundler/settings.rb
+++ b/lib/bundler/settings.rb
@@ -76,6 +76,7 @@ module Bundler
     ].freeze
 
     DEFAULT_CONFIG = {
+      :major_deprecations => true,
       :disable_version_check => true,
       :prefer_patch => false,
       :redirect => 5,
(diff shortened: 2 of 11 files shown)


deivid-rodriguez:
@indirect @segiddins You ok with this path forward? I know this was not the original plan, but I read https://github.com/bundler/bundler/pull/6586#issuecomment-397516354 and https://github.com/bundler/bundler/issues/4855#issuecomment-446802605 and I think we all agree about this now?

colby-swandale:
It's normal for users to have many versions of Bundler installed, so i find the deprecation message with the version number pretty handy to know which major version of Bundler has deprecated/removed what.

lib/bundler/feature_flag.rb
```diff
@@ -44,6 +44,7 @@ def self.settings_method(name, key, &default)
     settings_flag(:init_gems_rb) { bundler_2_mode? }
     settings_flag(:list_command) { bundler_2_mode? }
     settings_flag(:lockfile_uses_separate_rubygems_sources) { bundler_2_mode? }
+    settings_flag(:major_deprecations) { bundler_2_mode? }
```
colby-swandale: IMO this should not be a feature flag as this is not enabling any particular Bundler feature and would also never be removed either. Can we just set the config option to have a default value instead?

deivid-rodriguez: That makes sense to me, I'm going to change it as you suggest :+1:.

deivid-rodriguez:
> It's normal for users to have many versions of Bundler installed, so i find the deprecation message with the version number pretty handy to know which major version of Bundler has deprecated/removed what.

It was confusing for me because the first time I saw it I didn't know whether it meant the deprecation horizon (when the feature will be removed), the currently running version, or the version that removed the feature.

Anyways, I don't really find the version the deprecated a feature very useful for users, but I can restore it if other feel differently. If I do restore it, should I at least remove the trailing ".0"? If we deprecate a feature on bundler 2.1, won't that be incorrect?

deivid-rodriguez:
By the way, thanks so much for having a look at this @colby-swandale !

deivid-rodriguez:
Ok, so it turns out the feature flag is needed so that we can keep the 1.x specs passing. The way we've been using feature flags is so that we can virtually maintain a single branch, and the only difference between branches is the running bundler version (which triggers a different set of specs).

Enabling deprecations inconditionally made the deprecations be also enabled in bundler 1, and that causes tests to fail. Doing this revealed a few problems with our tests and deprecations that we should address though, so tentatively removing the feature flag was a good thing :smiley:.

In particular, our tests are using deprecated behaviour. If we don't use the new behavior ourselves, how can we encourage users to do it? :sweat_smile:. I'll be addressing this on a separate PR, so that our tests always use preferred non deprecated behaviour _unless_ we are testing the deprecations themselves.

Creating a PR to fix the above also surfaced a few problems with some of the deprecations themselves. I'll create issues about these problems that I found as well.
 
As per what to do with this PR. I see several options:

* Restore the feature flag, at least for the time being.
* Remove the 1.x build, and get ready to maintain several branches. Our feature flag strategy didn't fully work after all, and we're maintaining several branches anyways.
* Hold the PR until I address the above issues.

deivid-rodriguez:
Oh, another idea: explicitly toggle off deprecations via environment for the 1.x build.

indirect:
Adding those updates to the tests sounds great! Also turning off deprecations for the versions before the deprecations are enabled sounds good to me. :)

deivid-rodriguez:
> Also turning off deprecations for the versions before the deprecations are enabled sounds good to me. 

Ok, so that means turning them off via environment as suggested in https://github.com/bundler/bundler/pull/6965#issuecomment-465128430, right? Just double checking because... well, that is essentially what the feature flag was doing, so maybe I can just restore that.

> As for the setting, maybe we can make it always default to on, with a setting named `silence_deprecations` that you can set to true if you want to hide deprecations?

Regarding this, it sounds like a good idea because it's great naming, but it implies introducing yet another setting to do the same thing. Would we remove (or deprecate :scream:) the current setting, or just add it as a "nice redundancy"?

indirect:
> that means turning them off via environment as suggested in [#6965 (comment)](https://github.com/bundler/bundler/pull/6965#issuecomment-465128430), right?

Yup, that seems super reasonable to me. 👍 

> Would we remove (or deprecate 😱) the current setting, or just add it as a "nice redundancy"?

I would be okay with either renaming the setting or adding it as the "user-facing" setting with good UX.

deivid-rodriguez:
I changed the name of the setting to `silence_deprecations`, and regarding the failing 1.x specs, it was due to a bug in deprecations that I fixed :point_right: https://github.com/bundler/bundler/pull/6965/commits/cf7fa53e0e672f9db75e9d48f88ae4a8526a5312.

deivid-rodriguez:
@indirect May I ask for a final look here?

indirect:
Awesome! :)

@bundlerbot r+

PR #6968, opened 2019-02-15
Travis "bundle gem" template: Drop unused sudo: false option

olleolleolle:

I reviewed a PR in an Open Source project, and it had this defunct setting from Travis.

I was told the `bundle gem` command outputs this directive, and I looked here, and lo, it was.

See https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration for historical detail about when it was removed.

My fix is to **drop** this **unused directive** from the template.

I chose this fix because it works the same for the user, with less configuration.

### DIFF
--- a/lib/bundler/templates/newgem/travis.yml.tt
+++ b/lib/bundler/templates/newgem/travis.yml.tt
@@ -1,5 +1,4 @@
 ---
-sudo: false
 language: ruby
 cache: bundler
 rvm:


olleolleolle:
Kevin in the Bundler Slack channel added the input of "just run 1 Ruby to begin with".

deivid-rodriguez:
This looks great to me, you just need to fix the failing specs. They seem to be pretty exact on checking what the generated TravisCI config should look like :sweat_smile:.

lib/bundler/templates/newgem/travis.yml.tt
```diff
@@ -1,7 +1,5 @@
 ---
-sudo: false
 language: ruby
 cache: bundler
-rvm:
-  - <%= RUBY_VERSION %>
```
colby-swandale: I don't agree with this change. It doesn't offer any benefit, and we're adding more steps for users that want to test multiple ruby versions

deivid-rodriguez: Yeah, I agree with @colby-swandale here actually. The current configuration (aesthetics appart) was great for a single ruby too, so should work for every one. Let's leave it as it was, with the added benefit that you don't need to fix the failing spec :)

olleolleolle: There's a majority, I'll revert this bit - thanks for looking at it.

deivid-rodriguez:
@bundlerbot r+

deivid-rodriguez:
@bundlerbot r+

PR #6969, opened 2019-02-17
Fix redefinition warning

deivid-rodriguez:

The problem was that the test suite spits some warnings

```
/home/deivid/.rbenv/versions/2.6.1/lib/ruby/gems/2.6.0/gems/rspec-expectations-3.8.2/lib/rspec/matchers/dsl.rb:72: warning: method redefined; discarding old be_well_formed
/home/deivid/.rbenv/versions/2.6.1/lib/ruby/gems/2.6.0/gems/rspec-expectations-3.8.2/lib/rspec/matchers/dsl.rb:72: warning: previous definition of be_well_formed was here
```

My diagnosis was that we are redefining an RSpec matcher.

My fix is to define the matcher just once.

I chose this fix because moving it to the centralized place where custom matchers are defined made sense to me.

### DIFF
--- a/spec/quality_es_spec.rb
+++ b/spec/quality_es_spec.rb
@@ -43,14 +43,6 @@ RSpec.describe "La biblioteca si misma" do
     failing_line_message unless failing_line_message.empty?
   end
 
-  RSpec::Matchers.define :be_well_formed do
-    match(&:empty?)
-
-    failure_message do |actual|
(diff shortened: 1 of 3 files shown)


hsbt:
@bundler/bundlerbot r+

hsbt:
@bundlerbot r+

PR #6971, opened 2019-02-18
Fix warning message about bundler version

### LINKED ISSUE #6909: The warning for running an older version of Bundler 1.x shows an incorrect solution
Given I have Bundler `1.17.1` installed, and a `Gemfile.lock` specifying Bundler `1.17.3`, when I run `bundle check`, I expect the solution in the warning shown to be helpful. Instead, the warning is:

```
Warning: the running version of Bundler (1.17.1) is older than the version that created the lockfile (1.17.3). We suggest you upgrade to the latest version of Bundler by running `gem install bundler`.
```

The solution, `gem install bundler`, will not upgrade Bundler 1.x, but instead install 2.x. Thanks to the new version autoswitching, it still uses Bundler `1.17.1`, and the warning remains.

Issue #6908 makes this confusing to debug.

As documented, upgrading to Bundler 2 is not always convenient, so upgrading 1.x should be supported.

I think the solution shown should instead be:

```
We suggest you upgrade to the latest version of Bundler by running `gem update --system && gem install bundler && bundle update --bundler`, or upgrade to the latest version of Bundler 1.x by running `gem install bundler --version '~> 1'.
```

deivid-rodriguez:
Fixes #6909.

The problem was that we're recommending a command to make a warning go away that doesn't really make the warning go away.

My diagnosis was that the current recommendation installs the latest bundler, which might not be the same as the version the lock file was created with. If it's not the same, the warning will not go away.

My fix is to recommend installing exactly the version the lockfile was created with.

I chose this fix because it's really simple and it should work.

### DIFF
--- a/lib/bundler/lockfile_parser.rb
+++ b/lib/bundler/lockfile_parser.rb
@@ -107,9 +107,9 @@ module Bundler
       when 0
         if current_version < bundler_version
           Bundler.ui.warn "Warning: the running version of Bundler (#{current_version}) is older " \
-               "than the version that created the lockfile (#{bundler_version}). We suggest you " \
-               "upgrade to the latest version of Bundler by running `gem " \
-               "install bundler#{prerelease_text}`.\n"
+               "than the version that created the lockfile (#{bundler_version}). We suggest you to " \
+               "upgrade to the version that created the lockfile by running `gem install " \
+               "bundler:#{bundler_version}#{prerelease_text}`.\n"
         end
       end
     end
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
For some reason TravisCI didn't send back the green status to Github, but this should be ready for a review. Next is to downgrade the hard error "You must use Bundler 2 or greater with this lockfile." to this message for the time being, but first I want to get the message right!

ZimbiX:
Not bad. Thanks for looking at this =)

deivid-rodriguez:
@bundlerbot r=colby-swandale

deivid-rodriguez:
Thanks for having a look, @colby-swandale!

PR #6974, opened 2019-02-19
Use newer list interface

deivid-rodriguez:

The problem was that our specs emit a lot of deprecations about the list command.

My diagnosis was that we are using the deprecated interface to the list command everywhere.

My fix is to start using the new interface. If we don't use it ourselves, how could we be in a position to tell our users to start using it?

I chose this fix because we should do this, at same point, and doing it now simplifies enabling deprecations.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -365,7 +365,7 @@ EOF
       # system binaries. If you put '-n foo' in your .gemrc, RubyGems will
       # install binstubs there instead. Unfortunately, RubyGems doesn't expose
       # that directory at all, so rather than parse .gemrc ourselves, we allow
-      # the directory to be set as well, via `bundle config bindir foo`.
+      # the directory to be set as well, via `bundle config set bindir foo`.
       Bundler.settings[:system_bindir] || Bundler.rubygems.gem_bindir
     end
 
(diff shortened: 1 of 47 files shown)


deivid-rodriguez:
I still need to update the documentation of list to document the preferred interface, but given the amount of changes this required I'd like to make sure we are in the same page about this deprecation. The new interface is superior and way more clear, but I feel some of the old shortcuts could stay, such as `bundle config` as a shortcut to `bundle list`, or `bundle config <key> <value>` as a shortcut to `bundle config set <key> <value>`?

deivid-rodriguez:
@janpio The azure build seems to be hanging? It sounds like a problem with Azure not being able to pickup the latest code after a force-push... maybe?

This is ready for a review, by the way. I updated the `bundle config` docs to document the new interface to the config command, and alos fixed an error message where the recommended command was not clear and would not work as specified. See https://github.com/bundler/bundler/pull/6974/commits/4bf058b56c8d9ab37f22dbd4a0c70b11a1950d2d.

deivid-rodriguez:
> @janpio The azure build seems to be hanging? It sounds like a problem with Azure not being able to pickup the latest code after a force-push... maybe?

Actually disregard this, looks like it was still running. Now it finalized properly :+1:.

janpio:
Hm, Azure Pipelines taking longer than Travis is of course not good, I will try to get a few PRs in fixing test failures on Windows which should hopefully speed up the build.

indirect:
Awesome, thanks!

@bundlerbot r+

PR #6893, opened 2019-01-08
Add output to bundle info to show the dependent gems

ankitkataria:
### What was the end-user problem that led to this PR?

Fixes #6800 
- adds functionality to show the dependent gems

### What is your fix for the problem, implemented in this PR?

- Replicated the functionality from `bundler-stats` gem using jmmastey/bundler-stats#1 as a reference as mentioned in the issue

PR #6975, opened 2019-02-19
Reenable deprecation specs (and make up a plan)

deivid-rodriguez:

The problem was that need to deprecate stuff and we are not yet ready (not for removing it, for deprecating it!)

My diagnosis was we need to make a plan at least for the most important deprecations, since there's some of them are not clear how they will happen.

My fix is to:

* [x] ~Get deprecation specs passing.~ Done in #6991. 

* [ ] ~Make a plan for stopping remembering options. For this, I propose the following:~ Extracted to #7006.
  - ~In bundler 2 (next release, 2.1 for example), print a deprecation message to warn about the removal of all the flags that currently depend on the options being remembered.~
  - ~In bundler 3, remove the options and deprecate remembering command line flags.~
 
* [ ] ~Make a plan for the removal of custom sources :github, :gist, and :bitbucket, and switching to https. I also propose to do this in 2 steps.~ Extracted to #7000.
  - ~In bundler 2 (next release, 2.1 for example), start using https by default for those sources, and deprecate setting the `github.https` setting to false.~
  - ~In bundler 3, make the setting `github.https` a noop, using `https` nevertheless, and go ahead and deprecate the sources.~

* [x] ~Remove the deprecations about old rubies and rubygems, since that code will never be run.~ Extracted to (and expanded in) #6982 and #6983.

* [ ] ~Remove the deprecation about prefering `gems.rb` to `Gemfile`. The deprecation was unclear, only showed when there were multiple gemfiles in the same folder, and we had decided to delay this anyways. I changed this deprecation to instead warn about having multiple gemfiles in the same folder.~ Extracted to #6999.

I chose this fix because although a bit slow, it's the smoothest path for our users that I could think of.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -61,11 +61,6 @@ module Bundler
       end
     end
 
-    def self.deprecated_option(*args, &blk)
-      return if Bundler.feature_flag.forget_cli_options?
-      method_option(*args, &blk)
-    end
-
     check_unknown_options!(:except => [:config, :exec])
     stop_on_unknown_option! :exec
 
@@ -142,7 +137,7 @@ module Bundler
       Gemfile to a gem with a gemspec, the --gemspec option will automatically add each
       dependency listed in the gemspec file to the newly created Gemfile.
     D
-    deprecated_option "gemspec", :type => :string, :banner => "Use the specified .gemspec to create the Gemfile"
+    method_option "gemspec", :type => :string, :banner => "Use the specified .gemspec to create the Gemfile"
     def init
       require "bundler/cli/init"
       Init.new(options.dup).run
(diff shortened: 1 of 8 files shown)


deivid-rodriguez:
I gave it a little more thought and I think the deprecation is fine with the updated message.

But maybe we should also stop printing the deprecation if the user explicitly sets the feature flag to `false`. This could be handy if the user wants to get rid of the deprecation message because she is aware of the change, but doesn't have time to review every flag given to every bundler command and decide whether a permanent config is needed, or whether the flag is fine as it is.

deivid-rodriguez:
Ok, I've augmented the scope of this PR to handle the most important deprecations and make a plan for each of them (including the deprecation of sticky options). This is because I need to make a compromise between fully isolated PRs, and keeping a sane list of opened PRs and not overwhelming other devs with review requests.

This PR is still a WIP, and it's created on top of #6965.

I changed the description to describe where I'm heading with this.

indirect:
I really like this plan. I am 👍

deivid-rodriguez:
Great, specs are passing but I want to give another pass at the flags deprecation and simplify the diff a bit. For now I extracted some changes to a couple of separate PRs: #6982 and #6893.

deivid-rodriguez:
I have now extracted each of the steps underlined in the description to separate PRs, so they can be reviewed and discussed independently! :+1:

Closing this!

PR #6973, opened 2019-02-19
Backport the latest commits on ruby core repository.

hsbt:
### What was the end-user problem that led to this PR?

RSpec example of the bundled bundler was failed on Ruby 2.7 from ruby core repository. I fixed some of the issues on ruby core repository.

### What was your diagnosis of the problem?

1. I fixed the location of `lib/bundler.gemspec` to `lib/bundler/bundler.gemspec` for fixing gemspec issue of Ruby 2.6.1: https://github.com/ruby/ruby/commit/7c9771be02ddc707fea8ec96089c8a100c59f806
2. The current Bundler examples relied on the same versions of Ruby that are the first binary of PATH and rspec invocation binary. But the ruby core repository uses the different versions for them.

### What is your fix for the problem, implemented in this PR?

I fixed the location path of bundler.gemspec and uses `ENV['BUNDLE_GEM']` instead of hard-coded `gem` command.

### Why did you choose this fix out of the possible options?

There is no options on ruby core repository. I welcome to another options.

PR #6977, opened 2019-02-20
Update the location of bundler gemspec.

hsbt:

On the ruby core repository has the original patch for the test suite. The current bundler repository was inconsistency status with it.

This pull-request was the missing commit of https://github.com/bundler/bundler/pull/6973

The test suite of bundler break the inconsistency gemspec location,


deivid-rodriguez:
Looks good, thanks!

@bundlerbot r+

PR #6980, opened 2019-02-21
Better code coverage

deivid-rodriguez:

The problem was that we are using an out of date solution for code coverage and that might cause issues or inaccuracies in the future.

My diagnosis was that the codeclimate gem has been deprecated in favor of an external executable. 

My fix is to replace the code coverage setup with a modern one.

I chose this fix because it's what codeclimate recommends.


deivid-rodriguez:
This one should be ready for a review now, in case someone wants to have a look. Not sure what I did but the codeclimate checks got enabled. The regular one about "quality" I would remove it, but the ones about coverage might be interesting to keep?

deivid-rodriguez:
Ok, so I noticed that our current coverage is wrong and completely misleading. It should be much higher than 73%, which is what the badge currently says.

This is due to most specs in `bundler` running under subprocesses, and thus not getting coverage because of `simplecov` not being available there. I fixed this but it is superslow and things time out in Travis. I'm not sure whether the slowness is related to the overhead of simplecov, or to the high number of coverage reports that need to be merged (one per subprocess).

I don't want to spend any more time on this, so I would personally completely remove `simplecov` until we have a correct solution for tracking coverage because the current situation is just misleading.

indirect:
@deivid-rodriguez iirc we added the coverage badge as part of our push to implement unit tests. way back in the day, maybe 5 years ago, we _only_ had subprocess integration tests, and that meant we often ran into bugs that could have been caught by unit tests. we used the coverage number to track a lot of added unit tests, covering more and more of the code base with tests that were much faster to run and covered a lot more edge cases than the very slow integration tests.

I don't really have an opinion one way or another about the coverage badge, so I'm fine with removing it now that it has served its purpose to track and encourage adding lots of unit tests that we didn't have before.

deivid-rodriguez:
I see... it's interesting to see what purpose it served. It's also good that I don't think we're currently showing a coverage badge anywhere right now, so this is only misleading for developers using `simplecov` locally, not for other users visiting the repo.

I don't have a strong opinion on removing it either. If someone finds the current setup still useful (for example, for observing the increases when adding coverage through tests that don't spawn subprocesses), we can keep it. I can leave this PR there in case future me or anyone else wants to investigate why the setup here makes tests so unbearably slow. Or I can go back to the previous version of the PR that only migrated to the new coverage tool by `codeclimate`, without attempting to make coverage tracking actually correct.

deivid-rodriguez:
I'm going to close this and remove coverage tracking for the time being instead. Doing that should make the repository merge easier, since it's one less issue to care about when merging both CIs.

PR #6982, opened 2019-02-22
Remove unnecessary rubygems filters

deivid-rodriguez:

The problem was there's a lot of unnecessary code, specially in specs, that's never run.

My diagnosis was that since dropping support for old rubygems versions, this code is no longer necessary.

My fix is to remove the code.

I chose this fix because red is great on diffs.


hsbt:
Seems good.

deivid-rodriguez:
@bundlerbot r=hsbt

PR #6983, opened 2019-02-22
Remove unnecessary ruby filters

deivid-rodriguez:

The problem was there's a lot of unnecessary code, specially in specs, that's never run.

My diagnosis was that since dropping support for old ruby versions, this code is no longer necessary.

My fix is to remove the code.

I chose this fix because red is great on diffs.

### DIFF
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -164,9 +164,6 @@ module Bundler
           "gems.rb and gems.locked will be preferred to Gemfile and Gemfile.lock."
       end
 
-      if RUBY_VERSION < "2"
-        major_deprecation(2, "Bundler will only support ruby >= 2.0, you are running #{RUBY_VERSION}")
-      end
       return if Bundler.rubygems.provides?(">= 2")
       major_deprecation(2, "Bundler will only support rubygems >= 2.0, you are running #{Bundler.rubygems.version}")
     end
--- a/spec/bundler/fetcher/compact_index_spec.rb
+++ b/spec/bundler/fetcher/compact_index_spec.rb
@@ -44,7 +44,7 @@ RSpec.describe Bundler::Fetcher::CompactIndex do
         end
       end
 
-      context "when OpenSSL is FIPS-enabled", :ruby => ">= 2.0.0" do
+      context "when OpenSSL is FIPS-enabled" do
         def remove_cached_md5_availability
           return unless Bundler::SharedHelpers.instance_variable_defined?(:@md5_available)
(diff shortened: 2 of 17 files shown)


spec/other/compatibility_guard_spec.rb
```diff
@@ -4,15 +4,6 @@
   context "when the bundler version is 2+" do
     before { simulate_bundler_version "2.0.a" }
 
-    context "when running on Ruby < 2.3", :ruby => "< 2.3" do
```
simi: Any chance to keep this spec, but just "simulate" unsupported Ruby version?

deivid-rodriguez: The thing is... this spec will never happen in real life I think. Running `gem install bundler` on an unsupported ruby would be short-circuit first by rubygems, wouldn't it?

deivid-rodriguez: In other words, this situation will only happen if bundler 2 is installed on rubies below 2.3 and that is... impossible I think?

simi: I think it is still possible (in theory, not tested) with some old ruby version and old rubygems version not supporting Ruby version constraint. But you're right. It is probably not worth it to keep this spec.

deivid-rodriguez: Yeah, that's a good point. Maybe it's worth keeping actually.It's just three lines of lib code and this spec.

shushugah:
Why are VCR files enclosed in this PR?

deivid-rodriguez:
@shushugah I could've removed the realworld specs that were specific to old rubies, but I ended up adding them to the current suite instead. The VCR files are the result of the API responses recorded by those tests.

I did this because I thought maybe it was worth running these high level specs in our test suite (even if they test cases that only been problematic on old rubies). But I'm not really sure, I guess we could just remove them.

spec/other/compatibility_guard_spec.rb
deivid-rodriguez: Let's not get too agressive with the removing, I restored the compatibility guard for now :+1:.

deivid-rodriguez:
Finally went ahead and removed that too, so no VCR files included in this PR anymore.

spec/other/compatibility_guard_spec.rb
segiddins: You can also clone and run Bundler locally, so I think at the bare minimum we should keep the check in lib/

deivid-rodriguez:
@bundlerbot r=colby-swandale

PR #6984, opened 2019-02-22
Refresh all VCR cassettes

deivid-rodriguez:

The problem was our VCR cassettes contain responses recorded a long time ago, but the real APIs might have changed or give different results since then.

My diagnosis was that we should re-record the cassettes every once in a while.

My fix removes old the previous cassettes and re-records them.

I chose this fix because I'm not sure how to do it otherwise.


deivid-rodriguez:
@bundlerbot r+

PR #6985, opened 2019-02-22
Remove git subshelling from gemspec

deivid-rodriguez:

The initial problem was that the default bundler gemspec integrated in ruby-core was shipped nearly empty: https://bugs.ruby-lang.org/issues/15582. That caused issues in bundler: https://github.com/bundler/bundler/issues/6937  

I looked into ruby-core history, and still not fully sure how it happened but something I noticed is that bundler does not make this integration easy because ruby-core doesn't have a git environment available, so it needs to do a bit of juggling: https://github.com/ruby/ruby/blob/trunk/tool/rbinstall.rb#L802-L803. We could make things easier.

My fix is to stop subshelling to `git` from our own gemspec. I think we do this in the default generated gemspec so that newbie users don't accidentally ship generated files with their gems (although the current solution would still ship generated files that newbie users accidentally commit to source control :smile:). But we shouldn't consider ourselves newbies, so I think we can avoid that. 

I chose this fix because it avoids shelling out to git inside the gemspec while still keeping some assurances about the correctness of the shipped set of files through a quality spec.

### DIFF
--- a/bundler.gemspec
+++ b/bundler.gemspec
@@ -46,8 +46,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency "rspec",      "~> 3.6"
   s.add_development_dependency "rubocop",    "= 0.65.0"
 
-  base_dir = File.dirname(__FILE__).gsub(%r{([^A-Za-z0-9_\-.,:\/@\n])}, "\\\\\\1")
-  s.files = IO.popen("git -C #{base_dir} ls-files -z", &:read).split("\x0").select {|f| f.match(%r{^(lib|exe)/}) }
+  s.files = Dir.glob("{lib,exe}/**/*", File::FNM_DOTMATCH).reject {|f| File.directory?(f) }
 
   # we don't check in man pages, but we need to ship them because
   # we use them to generate the long-form help for each command.
--- a/spec/quality_spec.rb
+++ b/spec/quality_spec.rb
@@ -237,6 +237,18 @@ RSpec.describe "The library itself" do
     end
   end
 
+  it "ships the correct set of files" do
+    Dir.chdir(root) do
+      git_list = IO.popen("git ls-files -z", &:read).split("\x0").select {|f| f.match(%r{^(lib|exe)/}) }
+      git_list += %w[CHANGELOG.md LICENSE.md README.md bundler.gemspec]
+      git_list += Dir.glob("man/**/*")
+
+      gem_list = Gem::Specification.load(gemspec.to_s).files
+
+      expect(git_list.to_set).to eq(gem_list.to_set)
+    end
+  end
+
   it "does not contain any warnings" do
     Dir.chdir(root) do
       exclusions = %w[
--- a/task/release.rake
+++ b/task/release.rake
@@ -4,9 +4,33 @@ require "bundler/gem_tasks"
 task :build => ["build_metadata", "man:build", "generate_files"] do
   Rake::Task["build_metadata:clean"].tap(&:reenable).real_invoke
 end
-task :release => ["man:require", "man:build", "release:verify_github", "build_metadata"]
+task :release => ["man:require", "man:build", "release:verify_files", "release:verify_github", "build_metadata"]
 
 namespace :release do
+  task :verify_files do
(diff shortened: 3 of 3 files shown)


deivid-rodriguez:
Yeah, the our specs already depend on `git`.

The only problem with this PR is that for now I removed the possibility of reading the gemspec correctly from outside of the root folder of the gem, because the `base_dir:` parameter to `Dir.glob` only exists since ruby 2.5. I guess I could check the running ruby and use this solution only for the latest rubies, but I'm not sure if this is worth it.

I'd like @hsbt's opinions on whether something like this would make the ruby-core installer easier to implement, and less likely to run into issues.

hsbt:
The ruby core collect files of bundler when installing the ruby binaries with https://github.com/ruby/ruby/blob/trunk/tool/rbinstall.rb#L661

This change is helpful to us. The ruby core needs to support a non-git environment like a docker.

colby-swandale:
As the person that has been releasing Bundler for the past year, i'm not comfortable with this PR being merged. I heavily depend on git to know exactly what files are going to be shipped in a release.

Each release will often require some sort of change that i need to make. This ranges from updating the Changelog, to fixing broken tests, or just quality improvements. See the merge commits from the stable branches into master that i make every time a release is published.

During the release process, Git provides a small but critical barrier from accidentally introducing files into a release. This obviously helps keep quality control for bundler (one of the most used ruby tools ever) up and helps ensure we don't ship broken releases.

I understand that ruby has had some issues with Bundler & Git but that should not be given priority over our own team's release flow.

indirect:
Maybe there's a way to get both things? It makes sense to me to want to provide ruby-core with a way to package Bundler without git installed, but it also seems like an important and helpful release process thing to have git as a backstop against accidentally including the wrong files.

The quality spec that checks the gemspec against the git file list is a good step in the direction of providing some assurance. Maybe we could also add something to the `gem build` step that checks the files that are going to be released against the files that are checked in to git, and errors out listing the files that don't match if there are files that don't match?

I'd be happy to hear more thoughts from both of you about possible solutions. :+1:

deivid-rodriguez:
Yeah, by no means I want to break our own workflows, let's make this better for everyone.

I realize that the current quality spec is not super-helpful since the environment in CI will be usually clean. My idea was to introduce the same check I added to the specs, but as a prerrequisite for the `:release` task. How does that sound?

@indirect If I understand your idea right, you suggest moving this check to `rubygems` itself? That's interesting, I hadn't considered that... :thinking:

deivid-rodriguez:
I gave it a little more thought and I think making this validation on `gem build` is not going to work, since this behavior is not always what users want. For example, developers of fat binary gems will want generated artifacts to be included with the gem even if they're not on source control. Right?

I think the best place for this check would be in a custom prerequisite to `rake release`.

indirect:
Sorry my explanation wasn’t clear! I also meant as a dependency of the rake task named :build 🙂

deivid-rodriguez:
I added the release file check in  https://github.com/bundler/bundler/pull/6985/commits/3470aef3514be8e38a94f8ea44c23add98e887d0. This PR is not yet ready (I need to restore building the gemspec from folders different from the root folder), but I want to make sure first. @colby-swandale are you good with this given the checks we are adding?

deivid-rodriguez:
Actually, I misunderstood what `git -C` was trying to do here. It was fixing https://github.com/bundler/bundler/issues/6029, which still works with this PR.

Building the current `bundler.gemspec` from a different directory to where the files are doesn't work with or without this PR, because `git -C` returns relative paths, which are interpreted by `gem build` as local to the cwd. Actually it does work with rubygems 3 due to https://github.com/rubygems/rubygems/pull/2204, but that's being [reverted and moved to `gem build -C` in the next rubygems release](https://github.com/rubygems/rubygems/pull/2596) anyways. And it would stil work with this PR even if that change was not being reverted.

To sum up, this PR should be good to go as is!

deivid-rodriguez:
Thank you @colby-swandale! Fixed conflicts, will merge later :+1:.

deivid-rodriguez:
Going in!

@bundlerbot r+

PR #6991, opened 2019-02-26
Fix deprecation specs

deivid-rodriguez:

The problem was the current deprecation specs are broken. Not the deprecations themselves, their specs.

My diagnosis was we need to fix the specs for the deprecations before actually making sure that the deprecations themselves are properly enabled.

My fix is to:

* Get the target version corrected. Deprecation specs should run on bundler 2, not on bundler 1, since deprecations are disabled on bundler 1.
* Get the deprecation matcher fixed. The previous version was passing every time any deprecation was printed and a positive expectation was set. That means that on a spec printing a single deprecation, the assertion `expect(warnings).to have_major_deprecation("bananas will be deprecated in favor of potatoes")`, or any other message, would just pass. This explains why the deprecation specs were currently passing on bundler 1 and it's fixed by https://github.com/bundler/bundler/commit/33383f2faae10cd307b86466921b1a17391bae80.

This fix changes the target for the deprecation specs to bundler 2 (https://github.com/bundler/bundler/commit/e54ddb60298036e3a25c95deda6d575d384ff48a) and uses the following strategy to get CI green after that:

* If the deprecation is already correct on bundler 2, do nothing, the specs already pass.
* If the deprecation is incorrect on bundler 2, but the fix is a no-brainer, fix it here. For example, the `bundle update --all` deprecation (968fe965c), or the `bundle console` deprecation (5bece2f0f).
* If the deprecation is incorrect on bundler 2, but fixing it requires a separate discussion, skip the spec for now, and stage the fix for a separate PR.

I chose this fix because it keeps this PR simple and focused, while unblocking fixing the rest of the deprecations by at least having correct (although disabled) specs for them.


deivid-rodriguez:
This is ready.

I also have some other PRs ready for each of the remaining deprecations, but I want to get this in first so those PRs can really prove what they are fixing by correcting the skipped specs that this PR is leaving.

deivid-rodriguez:
I don't think there's anything that can cause issues here, and it unblocks another PRs, so I'm going to go ahead and merge!

Please feel free to leave feedback afterwards and I can amend anything later! :+1:

deivid-rodriguez:
@bundlerbot r+

Issue #6900, opened 2019-01-09
source_spec.rb: "with color" tests only work on CI that return TTY=true (e.g. Travis, not AppVeyor or Azure Pipelines)


janpio:
When creating #6899 I noticed some 3 tests failing for both macOS and Linux on both AppVeyor and Azure Pipelines that should not really fail:

```
2019-01-09T21:37:36.5421043Z Failures:
2019-01-09T21:37:36.5421043Z 
2019-01-09T21:37:36.5421243Z   1) Bundler::Source#version_message when there are locked gems that contain the relevant gem spec with a different version with color should return a string with the spec name and version and locked spec version
2019-01-09T21:37:36.5421494Z      Failure/Error: expect(subject.version_message(spec)).to eq("nokogiri >= 1.6\e[32m (was < 1.5)\e[0m")
2019-01-09T21:37:36.5421644Z 
2019-01-09T21:37:36.5422113Z        expected: "nokogiri >= 1.6\e[32m (was < 1.5)\e[0m"
2019-01-09T21:37:36.5422288Z             got: "nokogiri >= 1.6 (was < 1.5)"
2019-01-09T21:37:36.5422423Z 
2019-01-09T21:37:36.5422594Z        (compared using ==)
2019-01-09T21:37:36.5423477Z      # ./spec/bundler/source_spec.rb:63:in `block (7 levels) in <top (required)>'
2019-01-09T21:37:36.5423891Z      # ./spec/spec_helper.rb:122:in `block (2 levels) in <top (required)>'
2019-01-09T21:37:36.5424038Z 
2019-01-09T21:37:36.5424214Z   2) Bundler::Source#version_message when there are locked gems that contain the relevant gem spec with a more recent version with color should return a string with the locked spec version in yellow
2019-01-09T21:37:36.5424407Z      Failure/Error: expect(subject.version_message(spec)).to eq("nokogiri 1.6.1\e[33m (was 1.7.0)\e[0m")
2019-01-09T21:37:36.5424523Z 
2019-01-09T21:37:36.5424667Z        expected: "nokogiri 1.6.1\e[33m (was 1.7.0)\e[0m"
2019-01-09T21:37:36.5424799Z             got: "nokogiri 1.6.1 (was 1.7.0)"
2019-01-09T21:37:36.5424935Z 
2019-01-09T21:37:36.5425056Z        (compared using ==)
[...]

PR #6957, opened 2019-02-10
Move on to bundler 3

colby-swandale:
This PR contains the merge of `2-0-stable` to `master`

PR #6994, opened 2019-02-27
skip colorize examples when the environment was mswin or non tty.

hsbt:

Fixes https://github.com/bundler/bundler/issues/6900

The current bundler examples were not supported on mswin and non-tty environments.

I make to skip them when Thor detected unsupported environment.

Thor is vendoered now. It's easy to fix bundler's example directly.

### DIFF
--- a/spec/bundler/source_spec.rb
+++ b/spec/bundler/source_spec.rb
@@ -60,6 +60,10 @@ RSpec.describe Bundler::Source do
             before { Bundler.ui = Bundler::UI::Shell.new }
 
             it "should return a string with the spec name and version and locked spec version" do
+              if Bundler.ui.instance_variable_get(:@shell).is_a?(Bundler::Thor::Shell::Basic)
+                skip "tty color is not supported with Thor::Shell::Basic environment."
+              end
+
               expect(subject.version_message(spec)).to eq("nokogiri >= 1.6\e[32m (was < 1.5)\e[0m")
             end
           end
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
@hsbt I think this is related to the fix @colby-swandale and I worked on in #6957. Can you double check whether https://github.com/bundler/bundler/pull/6957/commits/577b5221fb76fe4ad2cb7b2c3eebdb547e21383c fixes this issue?

hsbt:
It's related. I'm not sure that  https://github.com/bundler/bundler/commit/577b5221fb76fe4ad2cb7b2c3eebdb547e21383c is the correct approach. 

Because `Thor::Base.shell.new` has a original condition that is https://github.com/bundler/bundler/blob/master/lib/bundler/vendor/thor/lib/thor/shell.rb#L11 . https://github.com/bundler/bundler/commit/577b5221fb76fe4ad2cb7b2c3eebdb547e21383c skip the above condition like `ENV["THOR_SHELL"]` variables.

deivid-rodriguez:
I think both fixes are correct.

Ours is preventing the thor's shell to be memoized at the class level, and thus preventing our specs from using the right shell for what they are testing.

Yours is acknowledging that a colorized shell is not supported on Windows, trusting `thor`'s opinion. But, if the specs pass on Azure with our fix, maybe it means that colors are now supported :man_shrugging:.

hsbt:
@bundlerbot r+

Issue #6913, opened 2019-01-16
How to use Bundler 2.0 on MRI 2.6.0?


eregon:
I can install the gem with `gem install bundler`, but it's not used:
```
$ gem list

*** LOCAL GEMS ***

bigdecimal (default: 1.4.1)
bundler (2.0.1, default: 1.17.3)
...

$ bundle --version
Bundler version 1.17.3

$ bundle _2.0.1_ --version
Traceback (most recent call last):
	2: from /home/eregon/.gem/ruby/2.6.0/bin/bundle:23:in `<main>'
	1: from /home/eregon/.rubies/ruby-2.6.0/lib/ruby/site_ruby/2.6.0/rubygems.rb:302:in `activate_bin_path'
/home/eregon/.rubies/ruby-2.6.0/lib/ruby/site_ruby/2.6.0/rubygems.rb:283:in `find_spec_for_exe': Could not find 'bundler' (1.16.1) required by your /home/eregon/code/mspec/Gemfile.lock. (Gem::GemNotFoundException)
To update to the latest version installed on your system, run `bundle update --bundler`.
To install the missing version, run `gem install bundler:1.16.1`
```

And running that (or `bundle _2.0.1_ update --bundler`) doesn't change the Gemfile.lock.

I also tried to update RubyGems with `gem update --system`, but that doesn't help either.

deivid-rodriguez:
We definitely need to better communicate this, but it sounds like your lock file is locked to bundler 1.16.1 (`BUNDLED WITH 1.16.1`). You have 2 options: either upgrade your lock file to the bundler version you are using (`bundle update --bundler`), or use the bundler version you're locked to (`gem install bundler:1.16.1`).

eregon:
It's quite confusing imho, I would expect `bundler` to be "a gem" and so when I install version 2.0.1, the latest version is used like for other gems.
Compatibility would then be achieved by having code in Bundler 2.0 that cares about that, rather than relying on a `bundler 1.x` gem being installed, and having RubyGems or Bundler magically choose a different version based on the `Gemfile.lock` in the current directory.
This also means releases of 1.x would be less needed, but right now 1.x has to be maintained because 2.x cannot run 1.x Gemfile.lock, right?

eregon:
[...]

PR #6996, opened 2019-02-27
Downgrade version mismatches to warnings

### LINKED ISSUE #6993: Difference between v1 and v2 lock files
Given the issues people have with bundler 2 (https://github.com/bundler/bundler/issues/6784, https://github.com/rubygems/rubygems/issues/2661), what exactly are the differences between v1 and v2 lock files that v2 lock files cannot be used by bundler v1?

Because I have this diff that made my lock file unusable:

<pre>
diff --git a/mongoid-app/Gemfile.lock b/mongoid-app/Gemfile.lock
index b87758e..52eda2d 100644
--- a/mongoid-app/Gemfile.lock
+++ b/mongoid-app/Gemfile.lock
@@ -93,7 +93,7 @@ GEM
     mini_mime (1.0.1)
     mini_portile2 (2.4.0)
     minitest (5.11.3)
-    mongo (2.7.0)
+    mongo (2.7.1)
       bson (>= 4.4.2, < 5.0.0)
     mongoid (7.0.2)
       activemodel (>= 5.1, < 6.0.0)
@@ -208,4 +208,4 @@ DEPENDENCIES
   web-console (>= 3.3.0)
 
 BUNDLED WITH
-   1.17.3
+   2.0.1
</pre>

According to https://bundler.io/blog/2019/01/03/announcing-bundler-2.html Bundler 2 does not have any new features.

deivid-rodriguez:
Closes #6993.
Complements rubygems/rubygems#2621.

The problem was that bundler 2 has become too unfriendly because it errors out when the running version does not match the version in the BUNDLED WITH section in the lockfile.

My diagnosis is that we still believe that making sure the exact same bundler version is always run for a specific application is a good thing, because it ensures resolution is always the same because it's resolved by the exact same code that generated the lockfile.

BUT, we need to ensure this in a more user friendly way, for example by auto-switching and auto-installing the right bundler version without any user intervention.

My fix is to downgrade this hard error to a warning that does not prevent bundler from running.

I chose this fix because it still shows the mismatch but allows bundler to run, thus being a bit friendlier. Hopefully this won't be needed once bundler is smart enough to autoswitch and autoinstall.

### DIFF
--- a/lib/bundler/lockfile_parser.rb
+++ b/lib/bundler/lockfile_parser.rb
@@ -101,9 +101,6 @@ module Bundler
       return unless bundler_version
       prerelease_text = bundler_version.prerelease? ? " --pre" : ""
       current_version = Gem::Version.create(Bundler::VERSION)
-      if current_version.segments.first < bundler_version.segments.first
-        raise LockfileError, "You must use Bundler #{bundler_version.segments.first} or greater with this lockfile."
-      end
       return unless current_version < bundler_version
       Bundler.ui.warn "Warning: the running version of Bundler (#{current_version}) is older " \
            "than the version that created the lockfile (#{bundler_version}). We suggest you to " \
--- a/lib/bundler/lockfile_parser.rb
+++ b/lib/bundler/lockfile_parser.rb
@@ -101,17 +101,14 @@ module Bundler
       return unless bundler_version
       prerelease_text = bundler_version.prerelease? ? " --pre" : ""
       current_version = Gem::Version.create(Bundler::VERSION)
-      case current_version.segments.first <=> bundler_version.segments.first
-      when -1
+      if current_version.segments.first < bundler_version.segments.first
         raise LockfileError, "You must use Bundler #{bundler_version.segments.first} or greater with this lockfile."
-      when 0
-        if current_version < bundler_version
-          Bundler.ui.warn "Warning: the running version of Bundler (#{current_version}) is older " \
-               "than the version that created the lockfile (#{bundler_version}). We suggest you to " \
-               "upgrade to the version that created the lockfile by running `gem install " \
-               "bundler:#{bundler_version}#{prerelease_text}`.\n"
-        end
       end
+      return unless current_version < bundler_version
+      Bundler.ui.warn "Warning: the running version of Bundler (#{current_version}) is older " \
+           "than the version that created the lockfile (#{bundler_version}). We suggest you to " \
+           "upgrade to the version that created the lockfile by running `gem install " \
+           "bundler:#{bundler_version}#{prerelease_text}`.\n"
     end
 
   private
--- a/spec/lock/lockfile_spec.rb
+++ b/spec/lock/lockfile_spec.rb
@@ -155,6 +155,9 @@ RSpec.describe "the lockfile format" do
   end
 
   it "outputs a warning if the current is older than lockfile's bundler version" do
+    current_version = Bundler::VERSION
+    newer_minor = bump_minor(current_version)
+
     lockfile <<-L
       GEM
         remote: file://localhost#{gem_repo1}/
@@ -162,27 +165,26 @@ RSpec.describe "the lockfile format" do
           rack (1.0.0)
 
       PLATFORMS
-        #{generic_local_platform}
+        #{lockfile_platforms}
 
       DEPENDENCIES
         rack
 
       BUNDLED WITH
-         9999999.1.0
+         #{newer_minor}
     L
 
-    simulate_bundler_version "9999999.0.0" do
-      install_gemfile <<-G
-        source "file://localhost#{gem_repo1}"
+    install_gemfile <<-G
+      source "file://localhost#{gem_repo1}"
 
-        gem "rack"
-      G
-    end
+      gem "rack"
+    G
 
-    warning_message = "the running version of Bundler (9999999.0.0) is older " \
-                      "than the version that created the lockfile (9999999.1.0). " \
+    pre_flag = prerelease?(newer_minor) ? " --pre" : ""
+    warning_message = "the running version of Bundler (#{current_version}) is older " \
+                      "than the version that created the lockfile (#{newer_minor}). " \
                       "We suggest you to upgrade to the version that created the " \
-                      "lockfile by running `gem install bundler:9999999.1.0`."
+                      "lockfile by running `gem install bundler:#{newer_minor}#{pre_flag}`."
     expect(err).to include warning_message
 
     lockfile_should_be <<-G
@@ -192,18 +194,20 @@ RSpec.describe "the lockfile format" do
           rack (1.0.0)
 
       PLATFORMS
-        #{generic_local_platform}
-        #{specific_local_platform}
+        #{lockfile_platforms}
 
       DEPENDENCIES
         rack
 
       BUNDLED WITH
-         9999999.1.0
+         #{newer_minor}
     G
   end
 
   it "warns when updating bundler major version" do
+    current_version = Bundler::VERSION
+    older_major = previous_major(current_version)
+
     lockfile <<-L
       GEM
         remote: file://localhost#{gem_repo1}/
@@ -211,43 +215,41 @@ RSpec.describe "the lockfile format" do
           rack (1.0.0)
 
       PLATFORMS
-        #{generic_local_platform}
+        #{lockfile_platforms}
 
       DEPENDENCIES
         rack
 
       BUNDLED WITH
-         1.10.0
+         #{older_major}
     L
 
-    simulate_bundler_version "9999999.0.0" do
-      install_gemfile <<-G
-        source "file://localhost#{gem_repo1}/"
+    install_gemfile <<-G
+      source "file://localhost#{gem_repo1}/"
 
-        gem "rack"
-      G
+      gem "rack"
+    G
 
-      expect(err).to include(
-        "Warning: the lockfile is being updated to Bundler " \
-        "9999999, after which you will be unable to return to Bundler 1."
-      )
+    expect(err).to include(
+      "Warning: the lockfile is being updated to Bundler " \
+      "#{current_version.split(".").first}, after which you will be unable to return to Bundler #{older_major.split(".").first}."
+    )
 
-      lockfile_should_be <<-G
-        GEM
-          remote: file://localhost#{gem_repo1}/
-          specs:
-            rack (1.0.0)
+    lockfile_should_be <<-G
+      GEM
+        remote: file://localhost#{gem_repo1}/
+        specs:
+          rack (1.0.0)
 
-        PLATFORMS
-          #{lockfile_platforms}
+      PLATFORMS
+        #{lockfile_platforms}
 
-        DEPENDENCIES
-          rack
+      DEPENDENCIES
+        rack
 
-        BUNDLED WITH
-           9999999.0.0
-      G
-    end
+      BUNDLED WITH
+         #{current_version}
+    G
   end
 
   it "generates a simple lockfile for a single source, gem with dependencies" do
@@ -1468,5 +1470,23 @@ RSpec.describe "the lockfile format" do
 
     expect(err).to match(/your Gemfile.lock contains merge conflicts/i)
     expect(err).to match(/git checkout HEAD -- Gemfile.lock/i)
+  end
+
+private
+
+  def prerelease?(version)
+    Gem::Version.new(version).prerelease?
+  end
+
+  def previous_major(version)
+    version.split(".").map.with_index {|v, i| i == 0 ? v.to_i - 1 : v }.join(".")
+  end
+
+  def bump_minor(version)
+    bump(version, 1)
+  end
+
+  def bump(version, segment)
+    version.split(".").map.with_index {|v, i| i == segment ? v.to_i + 1 : v }.join(".")
   end
 end
--- a/spec/lock/lockfile_spec.rb
+++ b/spec/lock/lockfile_spec.rb
@@ -230,7 +230,7 @@ RSpec.describe "the lockfile format" do
     expect(err).to include("You must use Bundler 9999999 or greater with this lockfile.")
   end
 
-  it "warns when updating bundler major version", :bundler => "< 3" do
+  it "warns when updating bundler major version" do
     lockfile <<-L
       GEM
         remote: file://localhost#{gem_repo1}/
@@ -239,54 +239,6 @@ RSpec.describe "the lockfile format" do
 
       PLATFORMS
         #{generic_local_platform}
-
-      DEPENDENCIES
-        rack
-
-      BUNDLED WITH
-         1.10.0
-    L
-
-    simulate_bundler_version "9999999.0.0" do
-      install_gemfile <<-G
-        source "file://localhost#{gem_repo1}/"
-
-        gem "rack"
-      G
-    end
-
-    expect(err).to include(
-      "Warning: the lockfile is being updated to Bundler " \
-      "9999999, after which you will be unable to return to Bundler 1."
-    )
-
-    lockfile_should_be <<-G
-      GEM
-        remote: file://localhost#{gem_repo1}/
-        specs:
-          rack (1.0.0)
-
-      PLATFORMS
-        #{generic_local_platform}
-        #{specific_local_platform}
-
-      DEPENDENCIES
-        rack
-
-      BUNDLED WITH
-         9999999.0.0
-    G
-  end
-
-  it "warns when updating bundler major version", :bundler => "3" do
-    lockfile <<-L
-      GEM
-        remote: file://localhost#{gem_repo1}/
-        specs:
-          rack (1.0.0)
-
-      PLATFORMS
-        #{lockfile_platforms}
 
       DEPENDENCIES
         rack
--- a/spec/lock/lockfile_spec.rb
+++ b/spec/lock/lockfile_spec.rb
@@ -230,44 +230,6 @@ RSpec.describe "the lockfile format" do
     expect(err).to include("You must use Bundler 9999999 or greater with this lockfile.")
   end
 
-  it "shows a friendly error when running with a new bundler 2 lockfile" do
-    lockfile <<-L
-      GEM
-        remote: https://rails-assets.org/
-        specs:
-         rails-assets-bootstrap (3.3.4)
-           rails-assets-jquery (>= 1.9.1)
-         rails-assets-jquery (2.1.4)
-
-      GEM
-        remote: https://rubygems.org/
-        specs:
-         rake (10.4.2)
-
-      PLATFORMS
-        ruby
-
-      DEPENDENCIES
-        rails-assets-bootstrap!
-        rake
-
-      BUNDLED WITH
-         9999999.0.0
-    L
-
-    install_gemfile <<-G
-      source 'https://rubygems.org'
-      gem 'rake'
-
-      source 'https://rails-assets.org' do
-        gem 'rails-assets-bootstrap'
-      end
-    G
-
-    expect(last_command).to be_failure
-    expect(err).to include("You must use Bundler 9999999 or greater with this lockfile.")
-  end
-
   it "warns when updating bundler major version", :bundler => "< 3" do
     lockfile <<-L
       GEM
--- a/spec/lock/lockfile_spec.rb
+++ b/spec/lock/lockfile_spec.rb
@@ -154,7 +154,7 @@ RSpec.describe "the lockfile format" do
     G
   end
 
-  it "outputs a warning if the current is older than lockfile's bundler version" do
+  it "warns if the current is older than lockfile's bundler version" do
     current_version = Bundler::VERSION
     newer_minor = bump_minor(current_version)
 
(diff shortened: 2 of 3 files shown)


p-mongo:
+1 for downgrade, automatic installation of software from network? Sounds like something I would like to NOT have on my machines (I am not a windows/apple user, thanks).

deivid-rodriguez:
I don't yet have a clear idea of how it will exactly work, but my initial intention was that this "self auto-installation" would happen on `bundle install`. So I think it wouldn't be surprising for users because we are installing software from the network in any case.

colby-swandale:
I'm kinda torn on this. I'm worried about how users could be using a version of Bundler that includes breaking changes without realising it. I get the argument that lots of users were running into this error, but it made users aware that they were trying to use a Gemfile against a new version of Bundler that had breaking changes.

deivid-rodriguez:
But don't you think aborting is too much? Specially for the bundler 1 -> bundler 2 transition, since those are not intended to be really backwards incompatible (except for dropping old rubies, which does not affect supported rubies).

As a matter of fact, the only thing that makes bundler 2 incompatible with bundler 1 is this issue... :sweat_smile:.

indirect:
I'm also torn. I agree with @deivid-rodriguez that it is not user-friendly to throw an error when the installation might succeed, and I also agree with @colby-swandale that it's worrying to let users do things that we know sometimes won't work at all.

If we are going to allow users to do something that we know will sometimes fail and/or throw mysterious errors, I think that probably deserves a different kind of warning than "you have 1.17.2, and this lockfile was created with 1.17.3".

@deivid-rodriguez @colby-swandale how do you feel about these two options?

1. We could add something to the friendly exception catcher that look for a major version mismatch. That way if there is an exception, the end user will see something like "MAJOR VERSION MISMATCH: you are using Bundler 2 on a Bundler 1 lockfile. You might be able to resolve this error by running `gem install bundler -v '~> 1' and trying again."
2. We could change the version mismatch warning when it is a major version. Instead of saying "We suggest you upgrade", we could say something like "MAJOR VERSION MISMATCH: you are using Bundler 2 on a Bundler 1 lockfile, which is not supported by the Bundler team. This might not work. If you have problems, install Bundler 1 by running `gem install bundler -v '~> 1'` and try again.

What do you think? I'm trying to figure out how to balance the way user-friendliness means getting out of the way and "just working" with the way that Bundler 2 doesn't always actually work the same way as Bundler 1. 😬 Also super open to ideas. 😄

deivid-rodriguez:
Wait, just to clarify.

In this PR, I'm not actually changing how this works, I'm only delaying it to the _next_ major version bump. This is because the bundler 2 major release was special in the sense that there were not strict intended backwards incompatible changes. We decided to release it in order to get rid of old rubies support, but other than that there should've been no breaking changes (other than this issue being overlooked).

So, we _do_ kwow that installation will work, at least with the same certainty that a minor or patch level version bump will keep things working. So, I downgraded the hard error to the warning we already print when there's a mismatch in the minor or the patch level version.

I do think that the "hard error" message can be improved with actionable suggestions (the rubygems part of this already has that), but for now my idea was just to delay the hard errors.

indirect:
Yep, I definitely understand that this is delaying the hard error. To me, this issue feels like tension between being user friendly (not throwing a hard error because it MIGHT work) and being Bundler-dev-team friendly (not letting users do this, because it might NOT work).

Bundler 2 should hopefully be almost entirely compatible, but we aren't supporting it that way. I'm worried that if we merge this and ship it, we'll start getting new issues where Bundler 2 doesn't work the same as Bundler 1. The options I suggested above are me trying to solve that worry, while still reducing the v2 vs v1 error to a warning.

deivid-rodriguez:
Sure, I can understand that.

Should I close this PR and it's rubygems counterpart, and focus on more actionable error messages. Just note that the case this was solving is not the one you mention in your suggestions, but the inverse (running bundler 1 on a bundler 2 lockfile). But for this fix to make sense in that case, we would need to backport this to Bundler 1... :thinking:.

indirect:
Hmm. Maybe we should explicitly write out our warning/error goals, and then work on PRs to get there? I don't think we need to get rid of this PR, but I want to make sure it will have the end results that we actually want before we merge it. There are four cases (I think): Bundler 1 or Bundler 2, with a v1 lock or v2 lock. Here is my proposal for how we ideally want each case to work:

- Bundler 1, v1 lock: just works, nothing special
- Bundler 1, v2 lock: prints a yellow warning `MAJOR VERSION MISMATCH: This is a Bundler 2 lockfile, but you have Bundler 1. This may not work, and the Bundler team cannot support you if you have problems. Please upgrade to Bundler 2 to use this lockfile.`
- Bundler 2, v1 lock: prints a yellow warning `MAJOR UPGRADE: This is a Bundler 1 lockfile, but you are running Bundler 2. This may not work, and the Bundler team cannot support you if you have problems. Upgrade your lockfile by running \`bundle update --bundler\` to resolve this warning.`
- Bundler 2, v2 lock: just works, nothing special

How do you all feel about that, @deivid-rodriguez @colby-swandale?

deivid-rodriguez:
Thanks @indirect. Sounds good, I will have a look at the different cases and what kind of message we are giving in each case.

But just to make sure we are all on the same page, because I feel we are going a bit back and forth here. Do we agree all agree with the intention of this PR, namely, downgrading hard errors about this to warnings, _only for the bundler 2_ release?

indirect:
I think I agree! One last question: will this PR cause Bundler 2 to automatically upgrade the lockfile to a Bundler 2 lockfile? Or will it print the warning and leave the lockfile alone?

I would prefer that it print the warning and not upgrade the lock unless the user explicitly runs `bundle update --bundler`.

deivid-rodriguez:
Yup, that's the intention, no lock file upgrade, just a warning printed and lock file left alone :+1:. I will double check when I check all the different combinations you mentioned before.

deivid-rodriguez:
Sorry for the delay here. rubygems/rubygems#2621 has been merged so this is next on my plate :+1:.

colby-swandale:
@indirect In the current implementation the Bundler 1 `Gemfile.lock` will automatically be updated to Bundler 2 without the user's consent.

deivid-rodriguez:
You mean in which implementation? This PR, latest release, master, 2-0-stable?

colby-swandale:
Everything. Bundler will just lock to whatever version is running `bundle install`. So going from `1.17.2 -> 2.0.1` with `bundle install` will happen.

deivid-rodriguez:
Really? I don't think I've seen that bevavior. I have several apps/libraries locked to bundler 1, and they don't get upgraded silently unless I use `bundle update --bundler`. Have you double checked this?

deivid-rodriguez:
```
$ docker run --rm -it ruby:2.6.2 sh

# bundle --version
Bundler version 1.17.2

# gem --version
3.0.3

# touch Gemfile              

# bundle
The Gemfile specifies no dependencies
Resolving dependencies...
Bundle complete! 0 Gemfile dependencies, 1 gem now installed.
Bundled gems are installed into `./usr/local/bundle`

# cat Gemfile.lock	
GEM
  specs:

PLATFORMS
  ruby

DEPENDENCIES

BUNDLED WITH
   1.17.2

# gem install bundler:2.0.1
Fetching bundler-2.0.1.gem
Successfully installed bundler-2.0.1
1 gem installed

# bundle --version
Bundler version 1.17.2

# bundle install
The Gemfile specifies no dependencies
Bundle complete! 0 Gemfile dependencies, 1 gem now installed.
Bundled gems are installed into `./usr/local/bundle`

# cat Gemfile.lock
GEM
  specs:

PLATFORMS
  ruby

DEPENDENCIES

BUNDLED WITH
   1.17.2
```

colby-swandale:
```
$ cat Gemfile.lock
GEM
  remote: https://rubygems.org/
  specs:
    rack (2.0.6)

PLATFORMS
  ruby

DEPENDENCIES
  rack

BUNDLED WITH
   1.17.2

$ bundle --version
Bundler version 1.17.2
```
```
$ git checkout master
$ ruby -I lib exe/bundle install
Resolving dependencies...
Fetching rack 2.0.6
Installing rack 2.0.6
Warning: the lockfile is being updated to Bundler 2, after which you will be unable to return to Bundler 1.
Bundled gems are installed into `./.bundle`
```

```
$ cat Gemfile.lock
GEM
  remote: https://rubygems.org/
  specs:
    rack (2.0.6)

PLATFORMS
  ruby
  x86_64-darwin-18

DEPENDENCIES
  rack

BUNDLED WITH
   2.0.0.dev
```

colby-swandale:
Note that this PR doesn't change any of this behavior already.

deivid-rodriguez:
Ok, so the answer to my [previous question](https://github.com/bundler/bundler/pull/6996#issuecomment-475881186) would be that this is the behaviour in master, but not in any released version?

deivid-rodriguez:
Yeah, I reproduced this on master. This is unexpected to me. Do we all agree that we don't want this?

deivid-rodriguez:
Ok, I see what's going on now.

The selection of the right `bundler` executable for an application is currently done by `rubygems`. If we run the bundler executable directly without going through a rubygems binstub, we skip that step, that's why running `ruby -Ilib exe/bundle` upgrades the lockfile. But if you do `gem build bundler.gemspec; gem install bundler-2.0.0-dev.gem`, then things will happen as expected.

deivid-rodriguez:
What's the plan here? Should I rebase this PR and review the diffferent situations @indirect mentioned so that we can ship this with the next patch level release?

indirect:
@deivid-rodriguez that sounds good to me. my understanding is that this change will resolve one specific situation:

- Application lockfile says Bundler 1
- User has only Bundler 2 installed
- On running `bundle install`:
  - before this change, an error
  - after this change, a warning, gems installed, and lockfile updated to Bundler 2

Let me know if there are other scenarios I missed!

deivid-rodriguez:
These are the different cases, and how they are handled. Currently, and after this PR (provided it is properly backported to 1-17-stable and 2-0-stable):

## Currently

|   | Locked to Bundler 1 | Locked to Bundler 2 | Locked to Bundler 3+ |
| - | ---------------------------------------------- | ------------- | -------------- |
| Running Bundler 1  | WARN only on minor/patch mismatch | ERR | ERR |
| Running bundler 2 | WARN and locked version upgraded  | WARN only on minor/patch mismatch | ERR 
| Running bundler 3+ | WARN, and locked version upgraded | WARN, and locked version upgraded | WARN only on minor/patch mismatch |

## After this PR

|   | Locked to Bundler 1 | Locked to Bundler 2 | Locked to Bundler 3+ |
| - | ---------------------------------------------- | ------------- | -------------- |
| Running Bundler 1  | WARN only on minor/patch mismatch | WARN | WARN |
| Running bundler 2 | WARN and locked version upgraded  | WARN only on minor/patch mismatch | WARN 
| Running bundler 3+ | WARN, and locked version upgraded | WARN, and locked version upgraded | WARN only on minor/patch mismatch |

After writing this summary, I'm realizing that the point of this PR was to turn the "ERR" on the first-row-second-column cell of the first table into "WARN", but that can only be done in the 2-0-stable / 1-17-stable branches, since master no longer supports bundler 1.

So, this PR might not be necessary for master, it depends on which behavior we expect for the last two rows of the tables.

indirect:
This seems like an improvement to me, including porting this change to 1.x, 2.x, and the RubyGems version chooser as needed.

We might want to evaluate the possibility of forcing users with a Bundler 3 lock to upgrade to Bundler 3, because installing a Bundler 3 lock with Bundler <=2 could expose users to new security issues that they thought they were safe from by having a Bundler 3 lock.

eregon:
I'm a bit confused by the table. Here are my thoughts:
* Bundler `n` cannot support Gemfile/Gemfile.lock of Bundler `n+1`. This is simply impossible if the format of these files changes. So I'm OK to fail hard in that case. It seems this PR is actually trying to solve this. Maybe it's worth having a special case for Bundler 1/2, since their Gemfile/Gemfile.lock are very compatible.
This makes me think versioning of Bundler files (like a protocol version) could be separate from Bundler versions. If there was, the Bundler files version would probably not change between Bundler 1 and Bundler 2, the only change for Bundler 2 being dropping  support for ancient Rubies (and not requiring any changes to Bundler files).
* The opposite case, backwards compatibility, I think should work ([it currently doesn't](https://github.com/bundler/bundler/issues/6913#issuecomment-482967583)). Specifically, Bundler `n` should support Gemfile/Gemfile.lock of Bundler `n-1` and maybe also of `n-2`. This would allow a much smoother migration to newer Bundler versions, which is otherwise prevented until all users of a given Gemfile.lock switch (or are forced to) at the same time to a newer Bundler, which is obviously very disruptive. That would also remove the need to magically switch which Bundler version is used based on files in the current directory, which IMHO is confusing and unexpected (e.g., I don't know any other software doing this). Furthermore, such switching cannot work when Bundler/RubyGems become a single library with a single version, as there is always only one active RubyGems version.

deivid-rodriguez:
@eregon You're right, I updated the left-down part of the first table, that indeed fails, and that should be fixed by the rubygems PR.

deivid-rodriguez:
Yeah, I just tried both PRs in combination and it seems to just work. Running a bundler 2 gemfile with bundler 3 prints this warning: "Warning: the lockfile is being updated to Bundler 3, after which you will be unable to return to Bundler 2.", but works.

eregon:
> Running a bundler 2 gemfile with bundler 3 prints this warning: "Warning: the lockfile is being updated to Bundler 3, after which you will be unable to return to Bundler 2.", but works.

I think it should not update the lockfile, otherwise it makes it incompatible with Bundler 2.
Bundler 3 should be free to change the lockfile format if it needs to, and Bundler 2 would not be able to understand that.

eregon:
> "Warning: the lockfile is being updated to Bundler 3, after which you will be unable to return to Bundler 2."

This message also seems to imply `Running bundler 2` / `Locked to Bundler 3` will result in an error, which I think is OK in this direction.

deivid-rodriguez:
In general, the lock file will be updated with the bundler version that created it, that's why it's called the `BUNDLED WITH` section. I think the only exception to this is when an older version is running a newer lockfile. In that case, bundler is allowed to run but the locked version is not changed and the user gets a warning message. This is already the case for minor/patch level version mismatches.

This PR intends to extend that ONLY for bundler 1/2 major version mismatches, as a way to try to alleviate the issues we've been having while the lockfile format having no intended incompatibilities. But keeps the error for bundler 3 or higher.

(By the way, I think the implementation of the PR is incorrect :sweat_smile:, thanks for making me revisiting it)

deivid-rodriguez:
I corrected the table once again to reflect this :see_no_evil:.

deivid-rodriguez:
> We might want to evaluate the possibility of forcing users with a Bundler 3 lock to upgrade to Bundler 3, because installing a Bundler 3 lock with Bundler <=2 could expose users to new security issues that they thought they were safe from by having a Bundler 3 lock.

@indirect The table was incorrect, this PR only adds a special case for bundler 1 and 2, but this is still the same otherwise. Users running a bundler 3+ lockfile with bundler 1 or 2 will get an error telling them to install bundler 3.

eregon:
The table makes sense to me now :smile: 

> In general, the lock file will be updated with the bundler version that created it, that's why it's called the `BUNDLED WITH` section.

Right, I understand `BUNDLED WITH` as the version of bundler that did the resolution and created the lockfile.
However, running `bundle _2.0.1_ install` with a lockfile of Bundler 1 does *not* create the lockfile and does *not* resolve, so I think it should not change the lockfile `BUNDLED WITH`.

> I think the only exception to this is when an older version is running a newer lockfile (same major version).

Right, that warns and does not change the lockfile, as I expect.
Another case is same major, but bundler newer than the lockfile. This is of course fine, and the lockfile is not changed.
I think Bundler 2 and lockfile 1 is a very similar situation. IMHO `bundle install` could warn that you might want to redo resolution with Bundler 2, but should not change the lockfile.

> This PR intends to extend that ONLY for bundler 1/2 major version mismatches, as a way to try to alleviate the issues we've been having while the lockfile format having no intended incompatibilities. But keeps the error for bundler 3 or higher.

I see, makes sense.

lib/bundler/lockfile_parser.rb
```diff
@@ -101,17 +101,16 @@ def warn_for_outdated_bundler_version
...
-        raise LockfileError, "You must use Bundler #{bundler_version.segments.first} or greater with this lockfile."
-      when 0
-        if current_version < bundler_version
-          Bundler.ui.warn "Warning: the running version of Bundler (#{current_version}) is older " \
-               "than the version that created the lockfile (#{bundler_version}). We suggest you to " \
-               "upgrade to the version that created the lockfile by running `gem install " \
-               "bundler:#{bundler_version}#{prerelease_text}`.\n"
-        end
+      current_major_version = current_version.segments.first
+      major_bundler_version = bundler_version.segments.first
```
eregon: `bundler_version` being the `BUNDLED WITH` version is a very confusing name (it's not Bundler's version!).
The code was already like that, but it definitely doesn't make it easy to understand what happens.

lib/bundler/lockfile_parser.rb
```diff
@@ -101,17 +101,16 @@ def warn_for_outdated_bundler_version
...
+      if current_major_version < major_bundler_version && current_major_version > 2
```
eregon: Should be `major_bundler_version > 2`, right?

lib/bundler/lockfile_parser.rb
deivid-rodriguez: Yes, absolutely agree. I left it there because it was already like that, but should be changed. I can do that.

lib/bundler/lockfile_parser.rb
deivid-rodriguez: I had the exact same thought, but maybe it's because of the naming confusion? I had a second look before and decided it was correct.

deivid-rodriguez:
> Right, I understand BUNDLED WITH as the version of bundler that did the resolution and created the lockfile.
However, running bundle _2.0.1_ install with a lockfile of Bundler 1 does not create the lockfile and does not resolve, so I think it should not change the lockfile BUNDLED WITH.

Yeah, that's what I meant too. Version should be changed only if the `bundle` command ends up rewriting the lockfile. That part is currently confusing in the table, so I'll just remove it.

lib/bundler/lockfile_parser.rb
eregon: `current_major_version` is the Bundler (major) version, right?
And `major_bundler_version` is the (major) `BUNDLED WITH` version.
We want to error out if Bundler version < `BUNDLED WITH` and `BUNDLED WITH` > 2 (all `BUNDLED WITH` <= 2 lockfiles are compatible).

I believe if Bundler version < `BUNDLED WITH` and Bundler version > 2 is incorrect, for instance for Bundler version 2 (or 1) and `BUNDLED WITH` 3, which should error out but this condition is `false`.

deivid-rodriguez: Right, this is so confusing... The thing (I believe) is that the lower left part of the table should be already caught by the rubygems. The error would be:

> Could not find 'bundler' (3) required by your Gemfile.lock. To update to the latest version installed on your system, run `bundle update --bundler`. To install the missing version, run `gem install bundler:3`

Downgrading that to a warning as well is the complementary part of this in rubygems: https://github.com/rubygems/rubygems/pull/2696.

This brings up other issues:

* We should probably remove the `bundle update --bundler` recommendation from that advice, since I don't think `bundle update --bundler` works for downgrading the `BUNDLED WITH` version.

* We might want to remove the check in those cases, because otherwise, with both PRs applied, you get duplicated warnings, like 

```
$ bundle 
Could not find 'bundler' (3.0.0) required by your /home/deivid/Code/playground/bundler/bundle_with_dance/Gemfile.lock.
To update to the latest version installed on your system, run `bundle update --bundler`.
To install the missing version, run `gem install bundler:3.0.0`

Warning: the running version of Bundler (2.1.0.pre.1) is older than the version that created the lockfile (3.0.0).
We suggest you to upgrade to the version that created the lockfile by running `gem install bundler:3.0.0`.
The Gemfile specifies no dependencies
Bundle complete! 0 Gemfile dependencies, 1 gem now installed.
```

deivid-rodriguez: Maybe we should actually remove this check altoghether from `bundler` and unify everything in `rubygems`.

indirect:
After further discussions, I am going to write up a proposal for Bundler to always accept existing lockfiles, without errors. It seems like the best way to achieve our compatibility goals is to only “upgrade” the lockfile when asked to do so by the user.

deivid-rodriguez:
@indirect Whatever the proposal ends up being, I noticed that this hard error shouldn't be here. Whether you require `bundler/setup`, or run bundler standalone, both cases should be caught by the bundler version finder in rubygems, so I don't think we need the hard error here. So, I removed it.

indirect:
Works for me. :) I'm r+ on this and we can do a follow-up once the RFC is reviewed and accepted.

deivid-rodriguez:
Ok! Let's do this then!

@bundlerbot r=indirect

PR #6998, opened 2019-02-28
Add unavoidable warnings to test against build

bronzdoc:
This is the same as https://github.com/bundler/bundler/pull/6831 i can't reopen that one, so i added the changes here...

What you think @deivid-rodriguez ?

### DIFF
--- a/spec/quality_spec.rb
+++ b/spec/quality_spec.rb
@@ -228,6 +228,10 @@ RSpec.describe "The library itself" do
 
         # there's no way around this warning
         last_command.stderr.sub!(/^YAML safe loading.*/, "")
+        last_command.stderr.sub!("WARNING:  No cert chain found. Please consider signing your gem to improve security.", "")
+        last_command.stderr.sub!("See https://guides.rubygems.org/security for more information.", "")
+        last_command.stderr.sub!("WARNING:  See http://guides.rubygems.org/specification-reference/ for help", "")
+        last_command.stderr.sub!("\n\n", "")
 
         expect(last_command.stderr).to be_empty, "bundler should build as a gem without warnings, but\n#{err}"
       ensure


spec/quality_spec.rb
```diff
@@ -223,6 +223,11 @@ def check_for_specific_pronouns(filename)
         if Bundler.rubygems.provides?(">= 2.4")
```
deivid-rodriguez: This has been removed from latest master, that why the PR shows conflicts, I think.

spec/quality_spec.rb
```diff
@@ -226,8 +226,13 @@ def check_for_specific_pronouns(filename)
           gem_command! :build, gemspec
         end
 
-        # there's no way around this warning
+        # there's no way aroudn this warning
```
deivid-rodriguez: Typo!

spec/quality_spec.rb
```diff
@@ -226,8 +226,13 @@ def check_for_specific_pronouns(filename)
...
         last_command.stderr.sub!(/^YAML safe loading.*/, "")
+        last_command.stderr.sub!(/^YAML safe loading.*/, "")
```
deivid-rodriguez: Duplicated line!

bronzdoc:
@deivid-rodriguez done :)

spec/quality_spec.rb
```diff
@@ -228,6 +228,10 @@ def check_for_specific_pronouns(filename)
 
         # there's no way around this warning
         last_command.stderr.sub!(/^YAML safe loading.*/, "")
+        last_command.stderr.sub!("WARNING:  No cert chain found. Please consider signing your gem to improve security.", "")
+        last_command.stderr.sub!("See https://guides.rubygems.org/security for more information.", "")
+        last_command.stderr.sub!("WARNING:  See http://guides.rubygems.org/specification-reference/ for help", "")
+        last_command.stderr.sub!("\n\n", "")
```
deivid-rodriguez: Nitpick / curiosity, why do we need to remove two new line characters if we're removing three warning lines? I would instead include the line character with the previous `gsub`'s.

bronzdoc:
> Shouldn't we sign the bundler gem with a certificate before warning others about not doing it? I think it's important that we do the right thing before blaming others for not doing it.

IMHO It is out of the scope of this PR to decide what is the right thing of bundler to do or any other gem in fact, Rubygems client will do it's part in recommending what to do but not imposing what to do.

deivid-rodriguez:
I don't know... I honestly don't like spitting warnings at users about "best practices" that we don't follow ourselves. Not only talking about `bundler`, the `rubygems-update` gem is also not signed.

I'd like to hear more opinions.

bronzdoc:
@indirect @segiddinsd @djberg96 thoughts?

djberg96:
@deivid-rodriguez Well, what about adding a cert to bundler then? But basically I agree with @bronzdoc, it's outside the scope of this PR. It's just a suggestion, not an order.

djberg96:
:+1:

deivid-rodriguez:
Yeah, absolutely, I agree it's out of the scope of this PR.

I'm just saying that, in my opinion, adding certs for our own gems is something we should do before adding these warnings to the gemspec validator.

djberg96:
@deivid-rodriguez Well then, add them. :)

indirect:
For this particular PR: seems totally fine to me.

For suggesting signing gems in general: Gem signing has been fundamentally broken and insecure for almost the entire time that it has existed (finally noticed and fixed in CVE-2018-1000076). I do not know of any time that the RubyGems or Bundler teams have actively gone out of their way to tell gem authors to sign gems to make them more secure.

If we want stronger assurance about the contents of gems, we should be trying to implement [The Update Framework](https://theupdateframework.github.io/) or something like it, not telling people to sign gems using GPG.

deivid-rodriguez:
I'm not sure I understand @indirect. Does that mean you are against printing a warning in rubygems when the gem being built is not signed? If that's the case, the only reason for this PR if to get rubygems tests passing when implementing that feature in https://github.com/rubygems/rubygems/pull/2503, so this PR would be unnecessary.

indirect:
@deivid-rodriguez I personally think it's pointless to suggest signing gems. If the rest of the RubyGems team wants to print warnings about unsigned gems, I also don't think it's worth fighting over. 🤷‍♂️

deivid-rodriguez:
I see. I don't really have an opinion on signing gems. I have never done it myself and I probably won't bother regardless of this warning is finally aded.

I do think, as I have stated several times in this thread, that not following the best practices we are supposed to be driving is itself a bad practice.

deivid-rodriguez:
So what should we do here?

It sounds like @indirect and myself are against (although @indirect doesn't have a strong opinion), and @bronzdoc and @djberg96 are in favor.

Any more opinions, so we can either move this PR and rubygems/rubygems#2503 forward or close both?

bronzdoc:
Lets close both, if there's is interest by more people in this, we can re think it and open it.

PR #7000, opened 2019-03-01
Git based sources and custom sources deprecation

deivid-rodriguez:

The problem was that our plan for deprecating using `git` instead of `https` for custom sources under the hood, and to deprecating those custom sources itself was unclear.

My diagnosis was that we should give contradictory/confusing deprecation warnings to users, so we should do this is several steps.

My fix is to implement the following plan:

* In bundler 2 (next release, 2.1 for example), start using https by default for the custom sources (`:github`, `:gist`, `:bitbucket`), and deprecate setting the `github.https` setting to false.
* In bundler 3, make the setting `github.https` a noop, using `https` nevertheless, and go ahead and deprecate the sources.

I chose this fix because it's the friendlier path I could think of for our users.

### DIFF
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -480,7 +480,7 @@ repo_name ||= user_name
       end
 
       Bundler::SharedHelpers.major_deprecation 3, <<-EOS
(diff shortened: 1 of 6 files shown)


deivid-rodriguez:
I haven't got any reviews here, but @indirect agreed with this plan on #6975, so I'm doing this. As with any PRs that I self-merge, please feel free to leave any feedback after-merge and I'll try to address them.

@bundlerbot r+

PR #7002, opened 2019-03-01
Respect color option when instantiating shells

deivid-rodriguez:

The problem was that some specs have started to fail randomly in several unrelated PRs. See for example, https://travis-ci.org/bundler/bundler/jobs/500322731. Note that currently bundler specs always run in the same order ( :crying_cat_face: ), but enabling or disabling specs can still affect the final set of specs, and thus make this issue manifest.

My diagnosis was that the base thor's shell is being memoized the first time bundler instantiates a shell. That means further instantiations will not respect the "--no-color" option.

My fix is to reset the base thor's shell the proper color / non-color one, every time we instantiate a new shell, not only when "--color" is given. 

I chose this fix because it felt like a real issue that should be fixed in lib/, and not only with some state resetting hack inside the specs.

### DIFF
--- a/lib/bundler/ui/shell.rb
+++ b/lib/bundler/ui/shell.rb
@@ -10,7 +10,7 @@ module Bundler
       attr_writer :shell
 
       def initialize(options = {})
-        if options["no-color"] || !$stdout.tty?
+        if options["no-color"]
           Thor::Base.shell = Thor::Shell::Basic
         end
         @shell = Thor::Base.shell.new
--- a/lib/bundler/ui/shell.rb
+++ b/lib/bundler/ui/shell.rb
@@ -10,9 +10,7 @@ module Bundler
       attr_writer :shell
 
       def initialize(options = {})
-        if options["no-color"]
-          Thor::Base.shell = Thor::Shell::Basic
-        end
+        Thor::Base.shell = options["no-color"] ? Thor::Shell::Basic : nil
         @shell = Thor::Base.shell.new
         @level = ENV["DEBUG"] ? "debug" : "info"
         @warning_history = []
--- a/spec/bundler/source_spec.rb
+++ b/spec/bundler/source_spec.rb
@@ -65,6 +65,8 @@ RSpec.describe Bundler::Source do
           end
 
           context "without color" do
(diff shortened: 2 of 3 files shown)


colby-swandale:
I'm happy to see this change merged, but how does this effect #6994?

deivid-rodriguez:
I think both fixes are fine. With or without this patch, a colorized shell might not be supported on Windows, so #6994 would be necessary.

deivid-rodriguez:
I propose that we merge this one, and see if [the failing source specs](https://dev.azure.com/bundler/bundler/_build/results?buildId=284&view=logs&jobId=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&taskId=f0948d91-f29e-5a90-437b-021b74373f23&lineStart=1596&lineEnd=1626&colStart=1&colEnd=1) are fixed. If they are fixed, we can skip #6994.

deivid-rodriguez:
I meant to request a review from @hsbt here, since he expressed hesitations about this in #6994.

hsbt:
I still wonder that we should enforce to use `Thor::Shell::Color`.  When given mswin environment, What happens with `Thor::Shell::Color` ?

I think we should handle only `--no-color` option, and rely on `Thor::Shell` behavior with test changes like https://github.com/bundler/bundler/pull/6994

deivid-rodriguez:
@hsbt I think you are right. I'll propose a different solution for these flaky failures.

deivid-rodriguez:
@hsbt I (hopefully) finished this PR.  You were right, ANSI codes don't work on Windows. But the flaky I mentioned before also exists. So my fix for it is:

* Skip specs that rely on ANSI color codes on Windows.
* Fix the flaky by, instead of resetting thor's shell to `Thor::Shell::Color` when `no-color` is not specified, reset it to `nil`. That means `thor`'s logic to choose its default shell will be run again the next time because the default shell is no longer memoized.

deivid-rodriguez:
@hsbt Could you have another check at the updated PR if you have time?

deivid-rodriguez:
Yay!

@bundlerbot r+

PR #6608, opened 2018-06-28
Support rubocop-0.57.2

hsbt:
### What was the end-user problem that led to this PR?

When users use rubocop-0.57.2, They faced following errors.

```
.rubocop.yml: Lint/EndAlignment has the wrong namespace - should be Layout
Error: The `Style/TrailingCommaInLiteral` cop no longer exists. Please use `Style/TrailingCommaInArrayLiteral` and/or `Style/TrailingCommaInHashLiteral` instead.
(obsolete configuration found in .rubocop_todo.yml, please update it)
The `Lint/RescueWithoutErrorClass` cop has been replaced by `Style/RescueStandardError`.
(obsolete configuration found in .rubocop_todo.yml, please update it)
The `Performance/HashEachMethods` cop has been removed since it no longer provides performance benefits in modern rubies.
(obsolete configuration found in .rubocop_todo.yml, please update it)
The `Style/MethodMissing` cop has been split into `Style/MethodMissingSuper` and `Style/MissingRespondToMissing`.
(obsolete configuration found in .rubocop_todo.yml, please update it)
```

### What was your diagnosis of the problem?

.rubocop.yml and .rubocop_todo.yml still uses old configuration.

### What is your fix for the problem, implemented in this PR?

I fixed style declarations to modern style.

Issue #7003, opened 2019-03-01
Use a `disabledByDefault` approach to rubocop


deivid-rodriguez:
Over the years, I've come to prefer enabling `rubocop` one rule at a time, but enabling consistency across the whole repo, rather that enabling the whole set of default rules, and maintaining a TODO file that gets a lot of churn over time and tends to get outdated.

I feel this approach gets less in the middle of developers with complaints that are not even enforced accross the whole repo. For example, rubocop yielded at me at https://github.com/bundler/bundler/pull/7002, but the offending rule is not fully enforced and [has exceptions](https://github.com/bundler/bundler/blob/4820acf3dbea172c6ec4f165b875acbe85f095cd/.rubocop_todo.yml#L213-L228). I feel it's easier for new developers to follow the current style if they "see it everywhere".

Also, maintaining the TODO file is not fun, and it obscures the point of other changes. For example, in the [2-0-stable merge](https://github.com/bundler/bundler/pull/6957), a big part of the diff is updating the `.rubocop_todo.yml` that doesn't really add much other than merge conflicts.

So, I propose to use the same approach we use in [rubygems](https://github.com/rubygems/rubygems/blob/77c3a99cc473c549bd322c7871d01fb6b8e5c580/.rubocop.yml).

hsbt:
+1

I always enabled `disabledByDefault` at my project.

PR #7004, opened 2019-03-01
Bump rubocop to 0.65.0

deivid-rodriguez:

The problem was that running rubocop (to fix style issues in other PRs) prints several warnings in the parser gem.

My diagnosis was that we should upgrade rubocop.

My fix is to upgrade rubocop, fix the config to be compatible with the new version, and regenerate the TODO config.

I chose this fix because it does the job. There's still some stuff I don't like about the current "rubocop maintainance" that I've proposed in #7003. But for the current situation I think this works.

Closes #6608.


colby-swandale:
Closes #6608

deivid-rodriguez:
Indeed, I missed that PR!

deivid-rodriguez:
@bundlerbot r+

PR #7005, opened 2019-03-01
Allow OTP code to be passed through bundler

kddeisz:
Hi bundler friends!

Currently, you can't pass your OTP code for 2FA through the `bundle exec rake release` command because it just sits and hangs.

The subshell used to spawn the `gem` command that does the push is hanging because it's waiting for the user to enter their OTP code. 

I switched the spawning from calling `read` immediately (which hangs, because it's waiting for EOF) to instead hook together stdins and call `gets` in a separate thread.

It works!

I have no idea how to actually test this with existing testing infra. Could use some guidance on that. Thanks!

### DIFF
--- a/lib/bundler/gem_helper.rb
+++ b/lib/bundler/gem_helper.rb
@@ -100,13 +100,25 @@ module Bundler
   protected
 
     def rubygem_push(path)
-      gem_command = %W[gem push #{path}]
-      gem_command << "--key" << gem_key if gem_key
-      gem_command << "--host" << allowed_push_host if allowed_push_host
+      args = ["gem", "push", path]
+      args += ["--key", gem_key] if gem_key
+      args += ["--host", allowed_push_host] if allowed_push_host
+
       unless allowed_push_host || Bundler.user_home.join(".gem/credentials").file?
         raise "Your rubygems.org credentials aren't set. Run `gem push` to set them."
       end
-      sh(gem_command)
+
+      Bundler.ui.debug(args.join(" "))
+      SharedHelpers.chdir(base) do
+        IO.popen(args, :in => :in) do |io|
+          Thread.new { puts io.gets until io.eof? }.join
+          io.close
+        end
+      end
+
+      code = $?.exitstatus
+      exit code unless code.zero?
+
       Bundler.ui.confirm "Pushed #{name} #{version} to #{gem_push_host}"
     end
 


deivid-rodriguez:
Thanks for fixing this @kddeisz! Tests are failing, can you have a look?

As per how to test this, I haven't looked at it in detail, but some initial advice I could give is to have a look at the `bundle gem` specs, because that command picks up user input, so they might help here.

lib/bundler/gem_helper.rb
```diff
@@ -100,13 +100,25 @@ def install_gem(built_gem_path = nil, local = false)
...
-      gem_command << "--key" << gem_key if gem_key
-      gem_command << "--host" << allowed_push_host if allowed_push_host
+      args = ["gem", "push", path]
+      args += ["--key", gem_key] if gem_key
+      args += ["--host", allowed_push_host] if allowed_push_host
+
       unless allowed_push_host || Bundler.user_home.join(".gem/credentials").file?
         raise "Your rubygems.org credentials aren't set. Run `gem push` to set them."
       end
-      sh(gem_command)
```
colby-swandale: It would be good if we could fix this problem in the `sh` method rather then just have the release task do its own output/input.

kddeisz: I'm happy to push this into `sh`, I just wasn't sure if you would want to switch all of these functions over to that. Not sure of any other implications it may have.

deivid-rodriguez: @colby-swandale @kddeisz I had a look at this PR and the related one (#7108), and I tend to think that @kddeisz's approach is better. The fix in #7108 changes all the methods using the `sh` helper to print all of there stdout and stderr to the screen. That's quite verbose, specially in the case of `rake build`. We _might_ still want to change that and print all those messages to the screen everytime, but for the shake of fixing this particular issue, I would only modify this specific method, which is also the only case where we should actually expect input from the user, I think.

What about extracting a `sh_with_input` method with the modified logic, and use it only here, for now?

kddeisz: Gosh yeah sorry this has languished for a while. I can take a look at this again later this week.

deivid-rodriguez: No worries, let's wait for @colby-swandale's input first so that you don't do unnecessary work.

deivid-rodriguez:
@kddeisz I created #7199 to try and finally fix this. I started with your code though, so I kept your attribution in there. I'm closing this to reduce the number of PRs fixing the same thing, thanks so much for getting this fix started!!

kddeisz:
No worries sounds good

PR #7006, opened 2019-03-01
Fix sticky options deprecation

deivid-rodriguez:

The problem was that deprecating sticky options is currently "useless" since we are removing the flags where the stickyness makes a difference at the same time. So, we don't help users adapt to (and understand) our changes. 

My diagnosis was that we should do this change in two steps, so that first users get a explanation about why we are removing the flags and what is the preferred alternative for them, and once nobody is no longer using these flags, we deprecate sticky options in general, since they are surprising and no longer useful, since we don't depend on them. 

My fix is to:

* In bundler 2 (next release, 2.1 for example), print a deprecation message to warn about the removal of all the flags that currently depend on the options being remembered.

* In bundler 3, remove the options and deprecate remembering command line flags.

I chose this fix because it properly and gradually informs our users about the changes that we intend to make and why we intend to make them.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -795,7 +795,7 @@ module Bundler
       return unless name_index
 
       value = options[name]
-      value = "#{value.join(" ")}" if option.type == :array
+      value = value.join(" ").to_s if option.type == :array
 
       Bundler::SharedHelpers.major_deprecation 2,\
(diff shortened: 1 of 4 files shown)


deivid-rodriguez:
Feedback appreciated in general, but specially with wording of the deprecations!

deivid-rodriguez:
Going in as per the plan outlined in #6975.

@bundlerbot r+

PR #7008, opened 2019-03-03
`bundle update` shows help instead of raising

deivid-rodriguez:

The problem was that after we change the behavior of `bundle update` to require the `--all` flag to update the whole bundle, it will start raising with the message "To update everything, pass the \`--all\` flag.". In my opinion, this is not very user friendly because it makes the simplest version of the command an error.

My diagnosis was that we should instead teach users how to use `bundle update` when they run `bundle  update`. This is the same rationale as showing `bundle` help instead of running `bundle install` when `bundle` is run. It's more user friendly than just saying "To install gems, use `bundle install`".

My fix is to delegate to `bundle help update` when `bundle update` is run.

I chose this fix because it makes sense from the end user point of view as I explained above.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -101,29 +101,10 @@ module Bundler
                              :desc => "Specify the number of times you wish to attempt network commands"
     class_option "verbose", :type => :boolean, :desc => "Enable verbose output mode", :aliases => "-V"
 
-    def help(topic = nil)
-      case topic
-      when "gemfile" then command = "gemfile"
-      when nil       then command = "bundle"
-      else command = "bundle-#{topic}"
-      end
-
-      man_path  = File.expand_path("../../../man", __FILE__)
-      man_pages = Hash[Dir.glob(File.join(man_path, "*")).grep(/.*\.\d*\Z/).collect do |f|
-        [File.basename(f, ".*"), f]
-      end]
-
-      if man_pages.include?(command)
-        if Bundler.which("man") && man_path !~ %r{^file:/.+!/META-INF/jruby.home/.+}
-          Kernel.exec "man #{man_pages[command]}"
-        else
-          puts File.read("#{man_path}/#{File.basename(man_pages[command])}.txt")
-        end
-      elsif command_path = Bundler.which("bundler-#{topic}")
-        Kernel.exec(command_path, "--help")
-      else
-        super
-      end
+    def help(topic = nil, plain = nil)
+      return super if plain
+      require "bundler/cli/help"
+      Help.new(topic, self).run
     end
 
     def self.handle_no_command_error(command, has_namespace = $thor_runner)
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -101,11 +101,11 @@ module Bundler
                              :desc => "Specify the number of times you wish to attempt network commands"
     class_option "verbose", :type => :boolean, :desc => "Enable verbose output mode", :aliases => "-V"
 
-    def help(cli = nil)
-      case cli
+    def help(topic = nil)
+      case topic
       when "gemfile" then command = "gemfile"
       when nil       then command = "bundle"
-      else command = "bundle-#{cli}"
+      else command = "bundle-#{topic}"
       end
 
       man_path  = File.expand_path("../../../man", __FILE__)
@@ -119,7 +119,7 @@ module Bundler
         else
           puts File.read("#{man_path}/#{File.basename(man_pages[command])}.txt")
         end
-      elsif command_path = Bundler.which("bundler-#{cli}")
+      elsif command_path = Bundler.which("bundler-#{topic}")
         Kernel.exec(command_path, "--help")
       else
         super
--- a/lib/bundler/cli/help.rb
+++ b/lib/bundler/cli/help.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+module Bundler
+  class CLI::Help
+    attr_reader :topic, :cli
+    def initialize(topic = nil, cli = nil)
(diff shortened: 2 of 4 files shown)


deivid-rodriguez:
By the way, this was the original proposal in https://github.com/bundler/bundler-features/issues/18, but it didn't get implemented when the feature landed in https://github.com/bundler/bundler/pull/5722.

colby-swandale:
I'm not in favor of this change. Showing the help page instead of giving a small error message could be too complicated for many users (especially new users) in trying to figure out what to do. You have to know where to look on manual pages.

I would prefer we keep to having the error simple and consistent with other commands like `bundle open` & `bundle gem`.

deivid-rodriguez:
But, why would new users assume `bundle update` will update all gems? And if this is the intuitive way of using bundler or other version managers, why are we changing this?

We shouldn't look at this from the perspective of a new user who migrates to bundler 2 (those will already have deprecation warnings about this for the whole 2.x series), we should look at it from the perspective a new user that installs bundler and uses it for the first time.

colby-swandale:
We could defiently improve the current error message to be more informative for users.

deivid-rodriguez:
Ok, so we consider that running `bundle update` is an error, and we should inform users how to properly use bundler, because they're doing it wrong, is that right?

I thought that the idea was to consider `bundle update` a supported command that the user runs when she wants to get information about how to update gems.

Note that I don't have a strong opinion about this, I just want to understand our reasoning since the original feature request proposed to print usage instructions when `bundle update` is run, not assume that the user wants to update all gems, and only explain how to do that.

deivid-rodriguez:
Ok, after a bit more thought, I think I agree with you @colby-swandale. I'm going to work on a better error message here.

deivid-rodriguez:
I've been thinking about this a bit more and I'm quite torn between:

* Undoing this deprecation.
* Display help when `bundle update` is run.

If we assume new users will intuitively run `bundle update` to update all gems at the same time, then this change is not a good change in my opinion. In that case, I would revert this change.

On the other hand, if we believe this is a good change then we should educate users and what can be better than showing help when a command has no other usage? Printing the error we're printing now is assuming new users will intuitively want to update all gems by running `bundle update`. And that's just getting in the middle and being annoying in my opinion.

Note that I'm not talking about the deprecation message in bundler 2 here, but about the error message in bundler 3. This kind of message is good for a deprecation message since that's intended for old users who are used to running `bundle update` to update all gems.

I think I slightly tend towards reverting this change and restore the old `bundle update` right now... :thinking:.

deivid-rodriguez:
Sorry for going back and forth with this.

I do understand @colby-swandale's point about man pages being hard to read for non experienced users, so I'm ok with showing a more general help message. I actually worked on a patch to `thor` to be able to specify multiple synopsis to a command, so that the error would look like

```
ERROR: "bundle update" was called with no arguments
Usage: bundle update *GEMS [OPTIONS]
       bundle update --group *GROUPS [OPTIONS]
       bundle update --source *SOURCES [OPTIONS]
       bundle update --ruby
       bundle update --bundler
```

But my hesitations about this deprecation are still there.

indirect:
@deivid-rodriguez everything that you said here makes sense to me. I can try to explain the "users run `bundle update` and are surprised" thing. Basically, we have gotten a lot of comments from users who skim the Bundler docs and somehow come to the conclusion that you run `bundle install` the _first_ time you check out a project, and then try to run `git pull && bundle update` at some time later to "update to the latest". This has extremely unexpected results, and also takes _forever_ to run because Bundler is trying to update every single gem and sometimes these projects have hundreds of gemfile entries and many hundreds of gems total.

If we can save people from accidentally trying to update every gem when what they actually wanted was to update their installed bundle, I think that would be a big win. In my estimation, users aren't going to read the docs for the `install` and `update` commands before trying `update` to see if it does what they want. Because of that, it seems like users who _do_ know what they want can start running `bundle update --all`, and hopefully users who are confused can get a helpful message redirecting them to `bundle install`. :)

deivid-rodriguez:
Ok. I think I'm satisfied with the rationale about the deprecation :+1:. `bundle update` is potentially dangerous because it can change a lot the code you will run in your production environment. Gems are best updated one at a time.

Now, for the migration process. My thoughts are:

* We should at least do some research to find out the real culprit of

>  comments from users who skim the Bundler docs and somehow come to the conclusion that you run bundle install the first time you check out a project, and then try to run git pull && bundle update at some time later to "update to the latest".

Maybe we have some misleading information in our docs, or there's a well know blog post that recommends this, or a common error message that suggests `bundle update` and should suggest something less disruptive, or something?

* Maybe we can use the deprecation period & message to try teach confused users about `bundle install` being better than `bundle update` in most situations.

indirect:
If we can find a source for the wrong idea about `bundle update`, I would love to fix it! My impression from talking to users has always been something like:

> Well, I saw that the `update` command existed, and what I wanted was to "update" the installed gems on my machine to match the list in the Gemfile that I just fetched by running `git pull`! I already installed this bundle in the past, so I wouldn't want to `install` again. I just want to update my installed bundle of gems, so that must mean I should run `update`. Surely that is logical.

That said, if anyone knows of a place contributing to this confusion, I would love to hear about it. 😬 

I definitely think that adding some kind of message to the output when you run `bundle update` without `--all` explaining that you probably don't want to run `bundle update` at all is a great idea. 😄

deivid-rodriguez:
For starters I opened [a PR to `thor`](https://github.com/erikhuda/thor/pull/652) to be able to improve the error messages for commands with complicated synopsis such as `bundle update`.

deivid-rodriguez:
My PR to thor was merged. I'll try to give this PR another thought and see what's best to do here.

PR #7009, opened 2019-03-03
Rubocop disabled by default

### LINKED ISSUE #7003: Use a `disabledByDefault` approach to rubocop

deivid-rodriguez:
Closes #7003.

The problem was that the current rubocop config gets in the middle too much for my taste.

My diagnosis was that we should enable one cop at a time instead of maintaining a hard too maintain, easily falling out of date `.rubocop_todo.yml` config.

My fix is to remove the TODO file, and only keep enabled the cops that current have 0 offenses. We can progressively globally enable more cops as we see fit in the future.

I chose this fix because it allows us to enforce _full_ consistency for specific rules, without getting in the middle at all otherwise.


.rubocop.yml
```diff
@@ -10,134 +8,54 @@ AllCops:
...
 # Layout
 
-Layout/AccessModifierIndentation:
-  EnforcedStyle: outdent
-
 Layout/AlignParameters:
+  Enabled: true
   EnforcedStyle: with_fixed_indentation
 
-Layout/EndAlignment:
```
colby-swandale: We shouldn't be removing cops that we have explicitly configured in our code base, I would propose that change in a separate PR/issue. Are these being removed because there are TODOs in `rubocop_todo.yml`?

deivid-rodriguez: > We shouldn't be removing cops that we have explicitly configured in our code base, I would propose that change in a separate PR/issue.

Well, technically these are not explicitly enabled, only explicitly configured. The only reason they appear here is that their configuration deviates from rubocop's default. The way I see it, they are no different from other cops that were enabled until now, but with the default configuration.

> Are these being removed because there are TODOs in rubocop_todo.yml?

Yes, my strategy here was to remove the TODO file and also remove (instead of explicitly enabling) all cops in `.rubocop.yml` that cause offenses without the TODO file.

That said, the `Layout` cops in particular should be safe to fully enable, so I could do that in a separate PR if you want.

deivid-rodriguez: Or, I can move the exceptions of these cops to the `.rubocop.yml`, although that's not something I particularly like because if I move the code around without changing its style, rubocop will shout at me because the exception no longer applies (like it happened on https://travis-ci.org/bundler/bundler/jobs/501054430, for example).

colby-swandale: Could we resolve the TODOs for these rules?

deivid-rodriguez: As I explained, I don't find these rules more special than others currently enabled implicitly. But I fixed the issues for them anyways in #7014. I left `Layout/AccessModifierIndentation` out, because the auto-correction is crashing and I don't want to do it manually. If don't want to remove this rule, I can move the exception list to the `.rubocop.yml` file.

deivid-rodriguez:
I finally took a different approach. The new version of this PR keeps all previous cops that we had enabled still enabled, and only removes the ones that had offenses. So the only difference now is that we no longer maintain a TODO config, and that we only explicitly opt-in to new cops, we don't automatically start using them when we upgrade rubocop. 

I did some manually intervention on some cops. In particular, I wonder why the `Style/MutableConstant` cop had exclusions. I think they were probably there for a reason, they all seem to be regexps. Any insights @colby-swandale?

deivid-rodriguez:
Ended up moving the specific cop fixes to #7016. This PR contains only the migration to a `DisabledByDefault` configuration.

deivid-rodriguez:
Wohohooooo! So happy we're doing this!

@bundlerbot r=colby-swandale

Issue #6888, opened 2019-01-08
[Windows test failure] rspec ./spec/bundler/plugin/installer_spec.rb:64 # Bundler::Plugin::Installer cli install with actual installers git plugins returns the installed spec after installing


janpio:
As described in https://github.com/bundler/bundler/issues/6868#issuecomment-451418573 currently 665 out of 2509 tests of bundler are failing on my Windows environment. I am working on fixing this.

---

Next one:

```

C:\Projects\bundler (windows -> origin)
λ ruby bin/rspec ./spec/bundler/plugin/installer_spec.rb:64
Run options:
  include {:locations=>{"./spec/bundler/plugin/installer_spec.rb"=>[64]}}
  exclude {:ruby_repo=>false, :bundler=>"=< 2.0", :rubygems_master=>true, :git=>"=< 2.20.1", :rubygems=>"=< 2.6.14.1", :ruby=>"=< 2.4.4", :realworld=>true, :sudo=>true}

Bundler::Plugin::Installer
  cli install
    with actual installers
      git plugins
error: pathspec 'INITIAL' did not match any file(s) known to git
error: pathspec 'COMMIT'' did not match any file(s) known to git
warning: You appear to have cloned an empty repository.
fatal: Needed a single revision
$ C:/Ruby24-x64/bin/ruby.exe -rrubygems -S gem --backtrace generate_index
Generating Marshal quick index gemspecs for 65 gems
.................................................................
Complete
Generated Marshal quick index gemspecs: 0.256s
Generating specs index
Generated specs index: 0.010s
Generating latest specs index
Generated latest specs index: 0.000s
Generating prerelease specs index
Generated prerelease specs index: 0.010s
Compressing indices
Compressed indices: 0.023s
C:/Projects/bundler/tmp/gems/base/gems/builder-2.1.2/lib/builder/xchar.rb:22: warning: constant ::Fixnum is deprecated
C:/Projects/bundler/tmp/gems/base/gems/builder-2.1.2/lib/builder/xchar.rb:89: warning: constant ::Fixnum is deprecated
# $? => 0
        returns the installed spec after installing (FAILED - 1)


Retried examples: 0

Failures:

  1) Bundler::Plugin::Installer cli install with actual installers git plugins returns the installed spec after installing
     Failure/Error: raise MissingGitRevisionError.new(ref, URICredentialsFilter.credential_filtered_uri(uri))

     Bundler::Source::Git::MissingGitRevisionError:
[...]

PR #7010, opened 2019-03-03
Use double quotes for `git commit -m`

deivid-rodriguez:

The problem was that Azure logs are full of "error: pathspec 'COMMIT'' did not match any file(s) known to git." and other similar ones.

My diagnosis was that Windows' doesn't like single quotes.

My fix is to use double quotes, which is also consistent with our global style.

I chose this fix because it brings the number of spec failures on Windows from 657 to 474! :tada:

### DIFF
--- a/spec/install/gemfile/git_spec.rb
+++ b/spec/install/gemfile/git_spec.rb
@@ -221,7 +221,7 @@ RSpec.describe "bundle install with git sources" do
       end
 
       Dir.chdir(lib_path("foo-1.0")) do
-        `git update-ref -m 'Bundler Spec!' refs/bundler/1 master~1`
+        `git update-ref -m "Bundler Spec!" refs/bundler/1 master~1`
       end
 
       # want to ensure we don't fallback to HEAD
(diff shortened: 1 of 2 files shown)


janpio:
> My diagnosis was that Windows' git doesn't like single quotes.

Windows in general doesn't use single quotes for escaping.

Besides that: 👍

deivid-rodriguez:
> Windows in general doesn't use single quotes for escaping command parameters.

So, you mean the patch is correct but the explanation of the PR and commit message are not? If so, I'll correct them with your suggestion.

> closes #6888

Sorry, I missed the related ticket! Didn't mean to steal credit... :flushed:.

janpio:
> So, you mean the patch is correct but the explanation of the PR and commit message are not? If so, I'll correct them with your suggestion.

Yes, patch is perfect and does more than my unfinished one was doing (which only fixed the INITIAL COMMIT one). Commit message is fine as well, just wanted to point out that it's not Git's fault but Windows ;)

> Sorry, I missed the related ticket! Didn't mean to steal credit... 😳.

Ha, no offense taken at all. Seems Sunday afternoons are a good time to work on some PRs ;) (Really, I just started to create a few PRs to fix Windows stuff when this PR came up as the newest build on Azure - perfect timing)

colby-swandale:
@bundlerbot r+

PR #7013, opened 2019-03-03
Move azure-pipeline folder to a hidden folder

colby-swandale:

We have a folder whose purpose is to contain support files for the Azure CI called `azure-pipelines`. The issue is that this folder does not relate to the Bundler development environment and is showing up in `ls` when it doesn't really need to be.

Move `azure-pipelines` to a hidden folder with the same name


colby-swandale:
Yea, that might help 😆

deivid-rodriguez:
@bundlerbot r+

janpio:
Thanks for fixing this @colby-swandale - totally valid issue and idea.

PR #7014, opened 2019-03-03
Fix all offenses of some easy cops

deivid-rodriguez:
References https://github.com/bundler/bundler/pull/7009/files#r261892716.

The problem was just some leftover consistency offenses.

My diagnosis was that we can easily fix these globally.

My fix is to enable the three cops and run `rubocop --auto-correct`.

I chose this fix because these cops depend on each other, so enabling just one of them leaves weird indented code behind.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -595,16 +595,6 @@ Style/StderrPuts:
 Style/SymbolArray:
   EnforcedStyle: brackets
 
-# Offense count: 3
-# Cop supports --auto-correct.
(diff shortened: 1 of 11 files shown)


deivid-rodriguez:
I fixed all issues for the cops currently configured differently from rubocop's default. Except for `Layout/AccessModifierIndentation` because auto-correcting it crash and I don't feel like spending time on correcting it manually.

deivid-rodriguez:
@bundlerbot r+

PR #7016, opened 2019-03-04
Fix offenses for more cops

deivid-rodriguez:

The problem was that a few cops with exceptions in the TODO file have been bothering me recently when moving files around.

My diagnosis was that we can just fix these offenses once and for all, or alternatively disable them inline, or moving them to the permanent config.

My fix is to:
* Auto-correct a few simple cops.
* Permanently disable `Style/GuardClause`, because I don't think it always leads to better code, and sometimes it forces artificial refactorings when making changes to methods, making the intention of the changes more obscure.
* Move `rescue Exception` disabling inline, so that when moving that code around, rubocop doesn't complain.

I chose this fix because this is my opinionated approach to this. :)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -50,6 +50,9 @@ Style/Alias:
 Style/FrozenStringLiteralComment:
   EnforcedStyle: always
 
+Style/GuardClause:
+  Enabled: false
+
 Style/MultilineBlockChain:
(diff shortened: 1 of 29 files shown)


deivid-rodriguez:
Going in, thanks @hsbt!

@bundlerbot r=hsbt

PR #7017, opened 2019-03-05
Development environment cleanup

deivid-rodriguez:

The problem was making changes in the development setup was brittle and led to hard to debug errors. For example, the ones in #6980. Also, it was hard to reproduce exactly what's going on in the current CI environment. Which is useful to setup other CI, or run the specs under a docker image.

My diagnosis was that we can simplify a lot of this stuff.

My fix is... a lot of simplifications, but the most important ones being:

* Now each binstub activates only its own dependency, not every development dependency. That means a contributor can perfectly work on manual pages installing only `rake` and `ronn` without having to install all development dependencies. Same with styling (`rubocop`), or with specs (`rspec`).
* Error messages are better now.
* The rake tasks for each main `Rakefile` section (man / spec / vendor) have been extracted to their own files.
* Rake tasks now shell out to the proper binstub, so gem activation is only needed there.

I chose this fix because it simplifies a lot the development environment in my opinion. Also, specs now pass under a docker image!

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -97,7 +97,6 @@ begin
 
   desc "Run specs"
   RSpec::Core::RakeTask.new
-  task :spec => "man:build"
 
   require "rubocop/rake_task"
   rubocop = RuboCop::RakeTask.new
--- a/Rakefile
+++ b/Rakefile
@@ -172,8 +172,7 @@ namespace :spec do
         end
 
         puts "Checked out rubygems '#{rg}' at #{hash}"
-        ENV["RUBYOPT"] = "-I#{File.join(RUBYGEMS_REPO, "lib")} #{rubyopt}"
-        puts "RUBYOPT=#{ENV["RUBYOPT"]}"
+        ENV["RGV"] = rg
       end
 
       task rg => ["clone_rubygems_#{rg}"]
--- a/Rakefile
+++ b/Rakefile
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 $:.unshift File.expand_path("../lib", __FILE__)
-require "shellwords"
 require "benchmark"
 
 NULL_DEVICE = (Gem.win_platform? ? "NUL" : "/dev/null")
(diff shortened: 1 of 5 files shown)


hsbt:
I agreed with your direction. But This pull-request was a few large changes. Should we split the followings?

1. Remove to support the old rubies like Ruby 1.8.7.
2. Split Rakefile
3. Others

deivid-rodriguez:
Absolutely, you're totally right.

deivid-rodriguez:
@hsbt I extracted the 1.8.7 changes to #7024, and removed the Rakefile split from this PR. I can also extract some improvements to the spec themselves, and not to the `rake` tasks (like https://github.com/bundler/bundler/pull/7017/commits/f8d6025039320b8866c44a1dec14ed8804047e3b, or https://github.com/bundler/bundler/pull/7017/commits/15aebd32697c93a6e4d90de1ab1abbeb667dd5d4) to a separate PR if you want.

deivid-rodriguez:
I extracted #7027 from this PR too.

deivid-rodriguez:
@hsbt This is ready for another review, it should be much smaller and focused now. The diff is still somewhat big, but it's much simpler if you view it without whitespace changes: https://github.com/bundler/bundler/pull/7017?w=1. I still recommend commit per commit reviewing for the rationale of each change! Thanks!

deivid-rodriguez:
At this point I addressed @hsbt's feedback and this PR is super low is risk since it only affects `Rakefile`, binstubs, and spec helper. So I'm doing it.

If any bundler dev has any trouble after merging this, please let me know and I'll fix it! :+1:

@bundlerbot r+

deivid-rodriguez:
I forgot to mention one change this PR introduced.

Now CI uses the "documentation" RSpec format, just like the local test run. I prefer this format because it is more informative for a CI system than "just dots", but if someone disagrees with this change I can revert it.

Sorry for not mentioning it in the PR, I forgot :pray:.

colby-swandale:
I don't like the change to the rspec format. It now takes forever to scroll through the build output to see what errors are occurring in builds.

deivid-rodriguez:
You're right about the inconvenient scrolling in TravisCI. Thanks for catching that and fixing it! :+1:

PR #7023, opened 2019-03-08
Shellsplit build config

### LINKED ISSUE #6940: bundle config build.<gem> doesn't send multiple parameters to extconf.rb correctly
### What works (`bundler 1.7.2`)
`bundle config build.<gem> --with-opt-lib=... --with-client-lib=...`, then 
`bundle install`
This succeeds, giving the same results as:
`gem install gem -v x.y.z -- --with-opt-lib=... --with-client-lib=...`.   

From `mkmf.log`:
```
gcc ... -L"/path/to/lib" -L"/another/path"
```

### What happens now (tried with bundler `1.17.1` and `2.0.1`)
After the exact same steps as above:
```
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.  

...

/usr/bin/ruby -I /usr/lib64/ruby/2.6.0 -r ./siteconf20190130-29495-12z1ot5.rb \ 
 extconf.rb --with-opt-lib\=<...>\ --with-client-lib\=<...>
```  
Notice that the build parameters are separated by an escaped space, because `extconf.rb` is being given the build parameters as one big string.  Verified this using `pry`.   

From `mkmf.log`:

```
gcc ... -L"/path/to/lib --with-client-lib=/another/path"
```
And the native extension fails to install because we can no longer find the required headers.

deivid-rodriguez:

The problem was #6940. Build configurations with multiple whitepaced parameter are no longer properly handled.

My diagnosis was not mine, was @jeremy's: Since [shellscaping / shelljoining build arguments to the rubygems gem installer](https://github.com/bundler/bundler/issues/6940), these are no longer properly passed

My fix is not mine, it's @jeremy's: shellsplit settings before passing them.

I chose this fix because it works and it sounds like the proper followup to https://github.com/rubygems/rubygems/pull/2441.

Fixes #6940.

### DIFF
--- a/lib/bundler/installer/gem_installer.rb
+++ b/lib/bundler/installer/gem_installer.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "shellwords"
+
 module Bundler
   class GemInstaller
     attr_reader :spec, :standalone, :worker, :force, :installer
@@ -56,7 +58,9 @@ module Bundler
 
     def spec_settings
       # Fetch the build settings, if there are any
-      Bundler.settings["build.#{spec.name}"]
+      if settings = Bundler.settings["build.#{spec.name}"]
+        Shellwords.shellsplit(settings)
+      end
     end
 
     def install
--- a/spec/bundler/installer/gem_installer_spec.rb
+++ b/spec/bundler/installer/gem_installer_spec.rb
@@ -26,4 +26,15 @@ RSpec.describe Bundler::GemInstaller do
       subject.install_from_spec
     end
   end
+
+  context "spec_settings is build option with spaces" do
+    it "invokes install method with build_args" do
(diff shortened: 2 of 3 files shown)


jeremy:
Might be nice to have stored these args as an array in the first place, too! With backward compatibility support for single-string build args. Separate PR though, I imagine.

For this PR, could broaden test coverage in `spec/install/gems/native_extensions_spec.rb` as well, since that tests actually passing the build args to extconf.rb. That uses a single build arg now, but should be proven to work with multiple args.

deivid-rodriguez:
I added the suggested spec, sorry for the delay!

deivid-rodriguez:
Feedback was addressed and @hsbt already approved the initial version, so going in!

@bundlerbot r+

PR #7024, opened 2019-03-08
Remove old 1.8.7 remanents

deivid-rodriguez:

The problem was that we have some really legacy code in our specs.

My diagnosis was that we forgot to remove/update these bits when dropping support.

My fix is to remove the legacy code.

I chose this fix because it's the only fix, really.


hsbt:
@bundlerbot r+

PR #7025, opened 2019-03-09
Bump rubygems in CI

deivid-rodriguez:

The problem was not testing against the latest and greatest rubygems.

My diagnosis was rubygems is outdated since the latest security releases.

My fix is to update the tested rubygems.

I chose this fix because it's the only fix.


colby-swandale:
@bundlerbot r+

PR #7027, opened 2019-03-09
Spec simplifications

deivid-rodriguez:

The problem was that the current specs don't pass under docker.

My diagnosis was that we should use 0.0.0.0 instead of 127.0.0.1 for mirror specs, since docker likes it better and the change doesn't affect the intention of the specs.

My fix is to do that change, but also add a few extra simplifications that I created while setting up the specs to run inside a docker image.

I chose this fix because it works and makes the test setup simpler.

### DIFF
--- a/spec/bundler/gem_helper_spec.rb
+++ b/spec/bundler/gem_helper_spec.rb
@@ -210,10 +210,10 @@ RSpec.describe Bundler::GemHelper do
 
       context "succeeds" do
         before do
-          Dir.chdir(gem_repo1) { `git init --bare` }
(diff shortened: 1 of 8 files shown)


deivid-rodriguez:
Well, just by looking at the description one can see that this PR is not really focused on a single subject either :rofl:... @hsbt Let me know if I should also split this into: general spec improvements vs docker fixes.

deivid-rodriguez:
The stuff here is explained commit per commit, affects only specs, and makes them more robust, so I'm going to merge it.

@bundlerbot r+

hsbt:
👍  Looks good. When we found the regression or issues, We will fix it.

PR #7029, opened 2019-03-12
Make deprecation specs less messy

deivid-rodriguez:

The problem was that deprecation specs, even though they are now passing, are very brittle. This is because most of them create multiple gem files (`gems.rb` and `Gemfile`), and thus display deprecation messages by default about that.

My diagnosis was that each deprecation should test its own message, instead of having the same deprecation printed by most specs.

My fix is to make the specs create a single gem file, except for the ones that test the deprecation about multiple gem files.

I chose this fix because it makes the tests less brittle and do less unnecessary things.

### DIFF
--- a/spec/commands/show_spec.rb
+++ b/spec/commands/show_spec.rb
@@ -157,7 +157,7 @@ RSpec.describe "bundle show" do
 
     it "does not output git errors" do
       bundle :show
-      expect(last_command.stderr).to eq_err("")
+      expect(err_without_deprecations).to be_empty
     end
   end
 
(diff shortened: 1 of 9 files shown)


deivid-rodriguez:
Ok, so I found another bug in the `have_major_deprecation` matcher. That bug was making all the deprecation specs about `.with_clean_env` and friends pass on bundler 1, even if those are not deprecated on bundler 1.

I fixed all the broken deprecation specs, and I was fixing the matcher too but since it's the second bug I find on it I decided to replace it with a simpler but equally readable alternative.

Besides that, I also replaced the `eq_err` matcher, which was pretty obscure in my opinion, with a more readable alternative.

deivid-rodriguez:
I think this arguably improves our specs readability, and since I want to unblock #6957 soon, I'm merging it!

@bundlerbot r+

PR #7030, opened 2019-03-12
Remove `LessThanProc`

deivid-rodriguez:

The problem was that the `LessThanProc` class we use for filtering specs is confusing, in my opinion. For example, if I tag a spec with `:bundler => "2"`, I would expect it to run only against bundler 2. However, that's not how it works, it instead runs against any bundler higher than or equal to 2. 

My diagnosis was that the `LessThanProc` class could be much simpler, and just check that the passed requirement matches the version the class was created with.

My fix is to simplify the `LessThanProc` class to only do requirement matching, and remove the "less than" functionality.

I chose this fix because in my opinion it's simpler to understand and less surprising.


deivid-rodriguez:
Thanks @colby-swandale!

@bundlerbot r=colby-swandale

PR #7033, opened 2019-03-13
Try fix azure flakyness

deivid-rodriguez:

The problem was that sometimes Azure is timing out. My guess is that this is not Azure slowness, but the big amount of failures currently present on the Windows build making it very slow.

My diagnosis was that we don't want red Azure builds for now, since we don't yet really fully support Windows.

My fix is to wrap the job that runs the specs with a [bash timeout](https://linux.die.net/man/1/timeout) with a limit under Azure's one hour limit. 

I chose this fix because I think it should prevent Azure statuses from getting in the middle for now.

### DIFF
--- a/.azure-pipelines/steps.yml
+++ b/.azure-pipelines/steps.yml
@@ -22,7 +22,7 @@ steps:
 - script: |
     ruby bin/rake spec:deps
   displayName: 'ruby bin/rake spec:deps'
-  
+
 - script: |
     gem install --no-document --conservative rspec_junit_formatter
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
These failures kind of failure last happened at #7030. See https://dev.azure.com/bundler/0e797e3c-1d87-40b4-ae08-813af36de214/_build/results?buildId=472.

deivid-rodriguez:
Merging this because I just got another timeout on #7037.

@bundlerbot r+

PR #7035, opened 2019-03-14
Unify the certificates for rubygems.org to rubygems store.

hsbt:

The problem was the users have duplicated certificates in their boxes. 

I wonder why bundler has a certificates copy from rubygems. It raised up in the tracker of ruby core. 

https://bugs.ruby-lang.org/issues/15384

I removed the duplicated certificates from bundler, and fallback to rubygems's one.


lib/bundler/fetcher.rb
```diff
@@ -295,7 +296,13 @@ def bundler_cert_store
         end
       else
         store.set_default_paths
-        certs = File.expand_path("../ssl_certs/*/*.pem", __FILE__)
+
+        rubygems_certs_dir = File.expand_path("../../rubygems/ssl_certs", __FILE__)
+        unless File.exists?(rubygems_certs_dir)
+          rubygems_certs_dir = File.join(RbConfig::CONFIG["rubylibdir"], "rubygems", "ssl_certs")
+        end
```
deivid-rodriguez: Rubygems has `Gem::RUBYGEMS_DIR` to inform of its own installation location:

https://github.com/rubygems/rubygems/blob/8b81173e0463fe362187319d4459c563b9b848c8/lib/rubygems.rb#L116that

Maybe it's a better fit here since it will always be correct and doesn't make us check for different possible installation structures?

lib/bundler/fetcher.rb
```diff
@@ -295,7 +296,13 @@ def bundler_cert_store
         end
       else
         store.set_default_paths
-        certs = File.expand_path("../ssl_certs/*/*.pem", __FILE__)
+
+        rubygems_certs_dir = File.expand_path("../../rubygems/ssl_certs", __FILE__)
+        unless File.exists?(rubygems_certs_dir)
```
deivid-rodriguez:
```suggestion
        unless File.exist?(rubygems_certs_dir)
```

lib/bundler/fetcher.rb
hsbt: Thanks, I didn't know it. It always provides in user's environment. We should use it.

voxik: There is even `Gem::Request.get_cert_files` available since RubyGems 2.3, so it should be even better ...

voxik:
> Why were they copied originally and why can we now use the cert store in RubyGems?

For backward compatibility with older RubyGems? The certs were introduced by these two commits:

* RubyGems: https://github.com/rubygems/rubygems/commit/5e16a5428f973667cabfa07e94ff939e7a83ebd9
* Bundler: https://github.com/bundler/bundler/commit/290b2e3c03a3c489de4e96315e6c700fc7813dbe

I assume that Bundler wanted to be secure but at the same time to support older RubyGems, where the certificates were not available yet.

Nevertheless, shipping your own certificates is fishy IMO. It just means RubyGems are not using globally trusted certificates, which is wrong IMO. Why they should?

On Fedora/RHEL, we remove all certificates shipped with Ruby/RubyGems/Bundler and everything works just fine. If it stopped working, then we would have to discuss the issue with our security guys and I guess with Mozilla subsequently, because the Fedora/RHEL cacerts are coming from Mozilla if I am not mistaken.

The tool to check that no certificates are shipped reports issues such as:

https://fedoraproject.org/wiki/Common_Rpmlint_issues#pem-certificate

But this might be different on other platforms (Windows?)

voxik:
BTW https://github.com/bundler/bundler/pull/4380 seems to be the latest PR touching the certificates and it does the same changes to Bundler as well as to RubyGems.

deivid-rodriguez:
The historical reason @voxik gave makes sense to me, although not explicitly stated in commit messages.

I don't see any risk in this PR, because we are using the exact same certificate contents after all, but maybe I'm missing something?

As per whether we should stop shipping certificates altogether, I think it makes sense to raise that as a separate issue?

lib/bundler/fetcher.rb
deivid-rodriguez: I'm fine with using `Gem::Request.get_cert_files` :+1:.

hsbt: Thanks again. Update done.

hsbt:
>I don't see any risk in this PR, because we are using the exact same certificate contents after all, but maybe I'm missing something?

I also don't see any risk.

voxik:
LGTM

hsbt:
Does anyone have another concern? I'm going to merge this in a few days later.

deivid-rodriguez:
No objections, let's get this in!

@bundlerbot r+

hsbt:
@bundlerbot retry

PR #7036, opened 2019-03-14
Bump travis rubies

deivid-rodriguez:

The problem was that hacks create confusion, even if they include TODO notes.

My diagnosis was that we can should upgrade to the latest ruby releases to reduce the number of hacks we need to maintain.

My fix is to upgrade rubies and remove hacks.

I chose this fix because it's a good change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ addons:
 
 rvm:
   - 2.6.2
-  - 2.5.4
+  - 2.5.5
   - 2.4.5
   - 2.3.8
 
@@ -51,7 +51,7 @@ jobs:
       script: rake rubocop
       stage: linting
     # Ruby 2.5, Rubygems 2.7
-    - rvm: 2.5.4
+    - rvm: 2.5.5
       env: RGV=v2.7.9
       stage: test
     # Ruby 2.4, Rubygems 2.6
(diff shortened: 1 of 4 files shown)


colby-swandale:
Restarted the build. Seems that the tgz for Ruby 2.6.2 wasn't up at the time of opening the PR.

deivid-rodriguez:
I still need to restore some of the hacks. The `gem install --defaul etc:1.0.1` one is still needed :(

deivid-rodriguez:
Hei!

I managed to get this green and removed all hacks. After closer examination, it might be `gem install etc:1.0.1 --default` (in general, `gem install <gem_with_extension> --default`), and not the ruby-core installer that was wrong. So what the hack was doing was "breaking" the installed specification, and thus preventing the default gem from being activated.

This sounds like a better fix. It needs a patch to `fileutils` that I also [proposed upstream](https://github.com/ruby/fileutils).

deivid-rodriguez:
We should wait for the upstream PR, @hsbt could you have a look?

hsbt:
@bundlerbot r+

hsbt:
@deivid-rodriguez I released https://github.com/ruby/fileutils/releases/tag/v1.2.0 with https://github.com/ruby/fileutils/pull/34

deivid-rodriguez:
Great! I opened https://github.com/bundler/bundler/pull/7068 to vendor that!

PR #7037, opened 2019-03-14
Test `bundle viz` command deprecation

deivid-rodriguez:

The problem was that we were not printing any deprecation messages for `bundle viz`.

My diagnosis was that we are removing `bundle viz` at the same time as deprecating it. Thus, the deprecations never actually show up. 

My fix is to first deprecate the command on bundler 2, then remove it in bundler 3.

I chose this fix because it's the most sensible approach to this removal.


deivid-rodriguez:
Thanks @colby-swandale!

@bundlerbot r=colby-swandale

deivid-rodriguez:
Had to rebase this on top of #7029 and adapt it. It should be ready now!

@bundlerbot r=colby-swandale

PR #7038, opened 2019-03-14
Strip unknown regexp option line from error format

k0kubun:

On Gemfile syntax error like `gem 'foo', :path => /unquoted/string/syntax/error`, Ruby 2.7 will show strange error message like:

```
[!] There was an error parsing `Gemfile`: unknown regexp options - trg - ...foo', :path => /unquoted/string/syntax/error
       ...                        ^~~~~~~. Bundler cannot continue.
```

due to recent change in Ruby trunk https://github.com/ruby/ruby/commit/b6468b01f7288789ef2e9bf548d81cdadb8ef053. This caused Bundler's [CI failure on ruby/ruby repository](https://dev.azure.com/rubylang/ruby/_build/results?buildId=180&view=logs&jobId=b6208202-37e9-506d-3aed-7f8fc7b76c87&taskId=a8620f99-3e9e-5782-41a8-f55e2d53059e&lineStart=94&lineEnd=95&colStart=1&colEnd=1).

Bundler's Travis ruby-head has not failed yet because Travis ruby-head is still a little old. It will fail later and this is fixing the future failure.

SyntaxError's message was changed to have an additional message indicating a position of an unknown regexp option like:

```
foo /bar/baz
         ^~~
```

or 

```
...oooooooooooooooooooooo /bar/baz
...                            ^~~
```

My fix tries to strip the new message part because the error line is already shown in another part of `DSLError#to_s`.

To keep the best compatibility with older Bundler versions.

### DIFF
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -582,7 +582,7 @@ The :#{name} git source is deprecated, and will be removed in the future.#{addit
         description = self.description
         if dsl_path && description =~ /((#{Regexp.quote File.expand_path(dsl_path)}|#{Regexp.quote dsl_path.to_s}):\d+)/
           trace_line = Regexp.last_match[1]
-          description = description.sub(/#{Regexp.quote trace_line}:\s*/, "").sub("\n", " - ")
+          description = description.sub(/\n.*\n(\.\.\.)? *\^~+$/, "").sub(/#{Regexp.quote trace_line}:\s*/, "").sub("\n", " - ")
         end
         [trace_line, description]
       end


colby-swandale:
This will need a test to make sure we don't introduce a regression in the future.

k0kubun:
I intentionally left it as is because having a test that fails by ruby's SyntaxError behavior change is the most important here and the failed test is doing the role.

That said, which did you mean by the reply:

* You need to write a unit test that fails in older rubies rather than testing against actual Ruby behavior
* You need to increase the number of test cases like the failed one to verify Bundler's behavior against the running Ruby better

deivid-rodriguez:
I tend to agree with @k0kubun here in the sense that fixing the ruby-head test failure while keeping the rest of the specs green should be enough here? Ruby changed in a way that affects bundler and our specs caught that. Fixing the failure that caught this should be enough?

Without having looked into this, only concern is @k0kubun's comment in the ruby repo: "Possible fix would be bundler/bundler#7038, but Ruby script ending with `^~~` could be wrongly stripped by that?". Maybe it's worth adding a test that checks just that.

hsbt:
@bundlerbot r+

hsbt:
We will find the regression about new feature of Ruby head on https://dev.azure.com/rubylang/ruby/_build?definitionId=3.

PR #7040, opened 2019-03-15
Remove `RUBY_VERSION` leftovers

deivid-rodriguez:

The problem was that find old ruby compatibility code can confuse contributors into thinking those versions are still supported.

My diagnosis was that we can remove this code, since we no longer support those versions.

My fix is to remove the code.

I chose this fix because we can do it :)


hsbt:
Nice clean-up 👍

hsbt:
@bundlerbot r+

PR #7043, opened 2019-03-15
Remove old rubygems compatibility code

deivid-rodriguez:

The problem was that I was unsure where I needed to add the compatibility layer to #6963, and it took me a bit to scan through the compatibility file and figure it out.

My diagnosis was that all this compatibility code is unused but makes this file harder to understand and scan through.

My fix is to remove the code.

I chose this fix because we can do it, I think.

### DIFF
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -20,6 +20,7 @@ module Bundler
 
     def initialize
       @replaced_methods = {}
+      backport_ext_builder_monitor
     end
 
     def version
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
Diff best seen with whitespace removed: https://github.com/bundler/bundler/pull/7043/files?w=1.

lib/bundler/rubygems_integration.rb
```diff
@@ -620,296 +539,160 @@ def method_visibility(klass, method)
...
-    elsif RubygemsIntegration.provides?(">= 1.8.0")
-      RubygemsIntegration::AlmostModern.new
-    elsif RubygemsIntegration.provides?(">= 1.7.0")
-      RubygemsIntegration::Transitional.new
-    elsif RubygemsIntegration.provides?(">= 1.4.0")
-      RubygemsIntegration::Legacy.new
-    else # RubyGems 1.3.6 and 1.3.7
-      RubygemsIntegration::Ancient.new
-    end
+    @rubygems ||= RubygemsIntegration.new
```
deivid-rodriguez: This is a good summary of this PR. Since currently `bundler` only supports rubygems 2.5 or higher, we can stick to the `MoreFuture` interface, which I unified in the `RubygemsIntegration` class.

deivid-rodriguez:
This PR might be hard to review if you look at the final diff, but if you go commit per commit, it should be straightforward I think.

deivid-rodriguez:
Can we get this in? Every time I need to read through this file I'm like :see_no_evil:

bronzdoc:
Sorry for taking too long @deivid-rodriguez!
 
@bundlerbot r+

PR #7044, opened 2019-03-15
Unskip `--binstubs` flag deprecation

deivid-rodriguez:

The problem was that some deprecation specs were still skipped.

My diagnosis was that I should get all of them passing.

My fix is to uncomment the spec and get it passing. The only fix was to include the full deprecation message so that it matches the assertion.

I could've regexp-matched the existing message but I preferred to use the full message in the assertion instead. I chose this fix because although it makes it more brittle since changing the message _will_ break the test, having the whole messages inside the deprecation specs makes it easier to get a quick view and understand the stuff that we are deprecating and the quality of the existing deprecation messages.

### DIFF
--- a/spec/other/major_deprecation_spec.rb
+++ b/spec/other/major_deprecation_spec.rb
@@ -74,9 +74,16 @@ RSpec.describe "major deprecations" do
     end
 
     describe "bundle install --binstubs" do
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Thanks @colby-swandale!

@bundlerbot r=colby-swandale

PR #7045, opened 2019-03-15
Review old `bundle config` interface deprecation

deivid-rodriguez:

The problem was that the `bundle config` command interface changes needed a few tweaks.

My diagnosis was that:

* `bundle config` deprecation was not getting enabled until bundler 3.
* `bundle config` documentation was incorrect about `unset` not supporting `--global` or `--local`.
* `bundle config` deprecation didn't give actionable suggestions.

My fix is to move the version to deprecate the old interface to `bundler 2`, to fix the documentation, and to improve the deprecation messages to give actionable suggestions.

I chose this fix because it makes `bundle config` better.

### DIFF
--- a/lib/bundler/cli/config.rb
+++ b/lib/bundler/cli/config.rb
@@ -14,8 +14,20 @@ module Bundler
     scope_options
     method_option :delete, :type => :boolean, :banner => "delete"
     def base(name = nil, *value)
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
:heart:

@bundlerbot r=colby-swandale

PR #7047, opened 2019-03-17
Remove remembered options deprecation message

deivid-rodriguez:

The problem was that this deprecation message is almost always a false positive. I realized about this when being bitten by this deprecation message during on our own specs (see https://travis-ci.org/bundler/bundler/builds/506776354 for example).

My diagnosis was that people running things like `bundle install --retry 3` will be annoyed by this deprecation message, because they're not doing anything wrong, really. They have the option to set the flag permanently through configuration, but passing it per command is fine too. The fact that the flag will no longer be remembered is unlikely to cause any problems in my opinion.

My fix is to remove this message altogether. We already print deprecation messages for the flags that we are removing because they only are guaranteed to work when remembering options (`--with`, `--without`, and friends). For the rest of the flags not being removed from `bundle install`, printing a deprecation message is just annoying since users are not doing anything wrong by using them.

I chose this fix because we shouldn't overwhelm users with deprecation messages unless we really need to.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -798,7 +798,7 @@ module Bundler
       value = value.join(" ").to_s if option.type == :array
 
       Bundler::SharedHelpers.major_deprecation 2,\
-        "The `#{flag_name}` flag is deprecated because it relied on being " \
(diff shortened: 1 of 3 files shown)


indirect:
Good catch!

@bundlerbot r+

PR #7049, opened 2019-03-19
Remove mustache dep

deivid-rodriguez:

The problem is that I just can't find where this development dependency is used. Furthermore, I can't find where it was used in the past :S

My diagnosis was that maybe it's never been used?

My fix is to remove the dependency.

I chose this fix because the less dependency, the better.

### DIFF
--- a/bundler.gemspec
+++ b/bundler.gemspec
@@ -40,7 +40,6 @@ Gem::Specification.new do |s|
   end
 
   s.add_development_dependency "automatiek", "~> 0.1.0"
-  s.add_development_dependency "mustache",   "0.99.6"
   s.add_development_dependency "rake",       "~> 12.0"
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Ok, I get it now. This is a transitive dependency ([of `ronn`](https://github.com/rtomayko/ronn/blob/c12e25d200d4851918c8d39d796091e6635cab9d/ronn.gemspec#L90)). At some point we [had to lock it](4652ad659d385b128014abc66ee31c89a0463ef6) because `mustache` 1.0 [dropped support for old rubies](https://github.com/mustache/mustache/blob/master/HISTORY.md#100-2015-01-11).

So, just another bonus for not having to support old rubies :)

hsbt:
@bundlerbot r+

PR #7051, opened 2019-03-20
Migrate git proxy helpers to use Open3

deivid-rodriguez:

The problem was that git proxy helpers had a few pending TODOs.

My diagnosis was that we should get the TODOs done, because now they can be done.

My fix is to fix the TODOs by migrating git proxy helpers to use `Open3` instead of backticks.

I chose this fix because it works.

### DIFF
--- a/lib/bundler/source/git/git_proxy.rb
+++ b/lib/bundler/source/git/git_proxy.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require "open3"
 require "shellwords"
 require "tempfile"
 module Bundler
@@ -77,8 +78,8 @@ module Bundler
 
         def contains?(commit)
           allowed_in_path do
-            result = git_null("branch --contains #{commit}")
-            $? == 0 && result =~ /^\* (.*)$/
+            result, status = git_null("branch --contains #{commit}")
+            status.success? && result =~ /^\* (.*)$/
           end
         end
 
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Any reviews anybody? :)

lib/bundler/source/git/git_proxy.rb
```diff
@@ -235,26 +238,15 @@ def allowed_in_path
...
-            Tempfile.open("captured_stderr") do |f|
-              STDERR.reopen(f)
-              return_value = yield
-              f.rewind
-              captured_err = f.read
-            end
-          ensure
-            STDERR.reopen backup_stderr
-          end
+        def capture_and_filter_stderr(uri, cmd)
```
hsbt: Should we define argument order same as `capture_and_ignore_stderr` like ```capture_and_filter_stderr(cmd, uri)```?

I prefer to same arguments order both. But It\'s not strong opinion.

deivid-rodriguez: But `capture_and_ignore_stderr` receives no `uri`, so there's no really a "defined order". `cmd` is the first argument of that method, but it's also the last argument :)

I have no strong opinion either, I only defined it like this so that it was more "visually similar" to the old way:

```diff
-            capture_and_filter_stderr(uri) { `git #{command}` }
+            capture_and_filter_stderr(uri, "git #{command}")
```

deivid-rodriguez: These are all private utilities, so we can always change them in the future if we find a different way is more readable. So I'll merge this! Thanks for having a look @hsbt!!

deivid-rodriguez:
@bundlerbot r+

deivid-rodriguez:
@bundlerbot retry

deivid-rodriguez:
@bundlerbot r=hsbt

PR #7052, opened 2019-03-20
Introduce  `original_system`, `original_exec`, `unbundled_system`, and `unbundled_exec` 

deivid-rodriguez:

The problem was that the `clean_system` and `clean_exec` would print deprecation messages, but there was not alternative for them.

My diagnosis was that the helpers are using deprecated behavior and that we should provide non deprecated alternatives.

My fix is to introduce `original_system`, `original_exec`, `unbundled_system`, and `unbundled_exec` for consistency with the rest of the helpers, and deprecate `clean_system` and `clean_exec`.

I chose this fix because while maybe not super pretty names, they offer an alternative consistent with the other helpers.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -335,12 +335,46 @@ EOF
       with_env(unbundled_env) { yield }
     end
 
-    def clean_system(*args)
-      with_clean_env { Kernel.system(*args) }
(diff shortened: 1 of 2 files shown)


indirect:
@bundlerbot r+

PR #7054, opened 2019-03-20
Remove --no-hardlinks and see what happens

deivid-rodriguez:

The problem was that cloning bit git repos can be very slow.

My diagnosis was that we are cloning the full history of the repo.

My fix is to use `--depth 1` for remote clones to only download what's necessary.

I chose this fix because it is much faster, according to my tests.

### DIFF
--- a/lib/bundler/source/git/git_proxy.rb
+++ b/lib/bundler/source/git/git_proxy.rb
@@ -105,7 +105,11 @@ module Bundler
             SharedHelpers.filesystem_access(path.dirname) do |p|
               FileUtils.mkdir_p(p)
             end
-            git_retry %(clone #{uri_escaped_with_configured_credentials} "#{path}" --bare --no-hardlinks --quiet)
+            if remote?(uri)
+              git_retry %(clone #{uri_escaped_with_configured_credentials} "#{path}" --bare --quiet --depth 1)
+            else
+              git_retry %(clone #{uri_escaped_with_configured_credentials} "#{path}" --bare --no-hardlinks --quiet)
+            end
             return unless extra_ref
           end
 
@@ -217,7 +221,7 @@ module Bundler
 
         # Adds credentials to the URI as Fetcher#configured_uri_for does
         def configured_uri_for(uri)
-          if /https?:/ =~ uri
+          if remote?(uri)
             remote = URI(uri)
             config_auth = Bundler.settings[remote.to_s] || Bundler.settings[remote.host]
             remote.userinfo ||= config_auth
@@ -225,6 +229,10 @@ module Bundler
           else
             uri
           end
+        end
+
+        def remote?(uri)
+          /https?:/ =~ uri
         end
 
         def allow?
--- a/lib/bundler/source/git/git_proxy.rb
+++ b/lib/bundler/source/git/git_proxy.rb
@@ -114,7 +114,11 @@ module Bundler
           end
 
           in_path do
-            git_retry %(fetch --force --quiet --tags #{uri_escaped_with_configured_credentials} "refs/heads/*:refs/heads/*" #{extra_ref})
+            if remote?(uri)
+              git_retry %(fetch --force --quiet --depth 1 --tags #{uri_escaped_with_configured_credentials} "refs/heads/*:refs/heads/*" #{extra_ref})
+            else
+              git_retry %(fetch --force --quiet --tags #{uri_escaped_with_configured_credentials} "refs/heads/*:refs/heads/*" #{extra_ref})
+            end
           end
         end
 
--- a/spec/install/gemfile/git_spec.rb
+++ b/spec/install/gemfile/git_spec.rb
@@ -876,7 +876,7 @@ RSpec.describe "bundle install with git sources" do
     update_git "foo"
     update_git "foo"
 
-    install_gemfile <<-G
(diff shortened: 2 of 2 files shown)


colby-swandale:
I doubt any test is covering this exact behaviour. We will need to dig into the git/github history to figure out why we have this option.

deivid-rodriguez:
Yup, I also doubted there was any unit test covering it, but I wasn't sure about higher level behavior, for example, some bundler functionality relying on some extra files / git objects being created only with this option. The purpose of this CI run was to confirm just that.

Let me dig into git history now and turn this into a proper PR if appropriate.

deivid-rodriguez:
So, the `--no-hardlinks` has always been there since the introduction of git sources, on a 2010 commit title ["More work towards awesomeness"](https://github.com/bundler/bundler/commit/f31835ec73e90db52f1fe9c5426c971c5723a60b) :smile: 

So, we should drop the option since the speed improvements are... a lot.

deivid-rodriguez:
I'm experimenting a bit more here and what actually seems to make a difference is cloning the entire repository history and not just the last revision. In order words, the `--depth 1` option.

Adding this option _might_ have an effect, not sure. I'm going to try adding it and see if CI complains.

deivid-rodriguez:
Ok, so CI complained but from a quick look at the failures, I think we can get the speed up without breaking existing functionality. I'll work on it :+1:.

deivid-rodriguez:
This is the difference I see (under my admittely suuuuuuuper slow network). 5 minutes (15MiB) vs 35 minutos (71MiB to get to 22%).

```
$ time git clone --bare --depth 1 https://github.com/rails/rails rails_bare
Clonando en un repositorio vacío 'rails_bare'...
remote: Enumerating objects: 4691, done.
remote: Counting objects: 100% (4691/4691), done.
remote: Compressing objects: 100% (3958/3958), done.
remote: Total 4691 (delta 297), reused 2671 (delta 156), pack-reused 0
Recibiendo objetos: 100% (4691/4691), 15.84 MiB | 51.00 KiB/s, listo.
Resolviendo deltas: 100% (297/297), listo.

real	5m45,306s
user	0m3,458s
sys	0m2,079s
```

```
$ time git clone --bare https://github.com/rails/rails rails_bare_full
Clonando en un repositorio vacío 'rails_bare_full'...
remote: Enumerating objects: 29, done.
remote: Counting objects: 100% (29/29), done.
remote: Compressing objects: 100% (29/29), done.
^Ccibiendo objetos:  22% (153275/681994), 71.89 MiB | 7.00 KiB/s     

real	35m56,429s
user	0m0,015s
sys	0m0,019s
```

deivid-rodriguez:
We should still test that this behaves properly when changing or updating git references in Gemfiles.

localhostdotdev:
ok, so with `Bundler version 2.0.1`, I got a checked out `rails/rails` (somehow it worked at some point, maybe because I used ssh, not sure).

I then install bundler from this branch, and I get:

```
Fetching https://github.com/rails/rails.git
fatal: Could not parse object '1da9a7e410f15859939eb4158421846638a41e71'.
Revision 1da9a7e410f15859939eb4158421846638a41e71 does not exist in the repository
https://github.com/rails/rails.git. Maybe you misspelled it?
```

for reference, the commit does exist: https://github.com/rails/rails/commit/1da9a7e410f15859939eb4158421846638a41e71

going to rails folder in `.bundle` and doing `git log` there:

```
commit ae3cc9daa94ea3db3c1d9e4ddcac508d88caa11e (grafted, HEAD -> master, origin/master, origin/HEAD)
```

for reference this commit is actually more recent than the one bundler is trying to get (2 hours difference).

I remove the `.bundle` directory, I'm now getting (with `bundle install`):

```
Resolving dependencies....
Could not find activesupport-6.0.0.beta3 in any of the sources
```

I guess there is just a lot of changes in this new version, anyway, I run `bundle update --all` and the git clone is still stuck.

I patch bundler to show the commands output and I get:

```
fetch --force --quiet --tags 'https://github.com/rails/rails.git' "refs/heads/*:refs/heads/*"
```

doing the command without `--quiet`:

```
Receiving objects:   8% (54810/685114), 14.57 MiB | 1.46 MiB/s
```

(this thing takes a while, would be really nice to show progress)

next command:

```
clone --no-checkout "/my/home/.bundle/cache/git/rails-fcf0202857b07db1a0f6220dae5ca99319ca0f32" "/my/home/src/updates/.bundle/ruby/2.6.0/bundler/gems/rails-ae3cc9daa94e"
```

not sure what's going on but I get:

```
Bundler could not find compatible versions for gem "zeitwerk":
  In snapshot (Gemfile.lock):
    zeitwerk (>= 2.1.2)

  In Gemfile:
    rspec-rails was resolved to 3.8.2, which depends on
      activesupport (>= 3.0) was resolved to 6.0.0.beta3, which depends on
        zeitwerk (~> 1.3, >= 1.3.1)

Running `bundle update` will rebuild your snapshot from scratch, using only
the gems in your Gemfile, which may resolve the conflict.
```

(when using this branch without my changes, dependencies resolved fine before)

what's weird is I never got the `--depth=1` command.

deivid-rodriguez:
Closed in favour of https://github.com/rubygems/rubygems/pull/4475.

PR #7055, opened 2019-03-21
Revert the RSpec format in CI back to dots

colby-swandale:

#7017 made a change that sets the RSpec format to `documentation` in CI. I'm against this change because it now takes _a lot_ more time to scroll to the build errors, and makes the build log have too much information. Maintainers care much more about the failing specs and their errors than the test descriptions in CI.

Explicitly set the RSpec format to `progress` in CI.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -134,7 +134,7 @@ namespace :spec do
     (branches + releases).each do |rg|
       desc "Run specs with RubyGems #{rg}"
       task rg do
-        sh("bin/rspec")
+        sh("bin/rspec --format progress --color")
       end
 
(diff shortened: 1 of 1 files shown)


colby-swandale:
@bundlerbot r+

deivid-rodriguez:
Actually I discovered a nice TravisCI feature thanks to this: the "Scroll to end of log" button :smile:

ScrollToBottom

Anyways, no big deal, I can always turn it on specific branches/PR when I need it :+1:

deivid-rodriguez:
And still, since all of our suites (regular, sudo, and realworld) run in the same job, scrolling is still inconvenient even with the button. So, :+1: to dots in CI, at least for now.

PR #7057, opened 2019-03-21
Review multiple sources deprecation

deivid-rodriguez:

The problem was that I had not yet reviewed the deprecation about multiple global sources being present on a Gemfile, and thus the specs were skipped.

My diagnosis was that we need to delay the removal of multiple sources support to bundler 3, so that we can show the deprecations in the 2.x series.

I also noticed that part of the deprecation message was inaccurate. In order to upgrade the warning to an error, you would also need to configure the `lockfile_uses_separate_rubygems_sources` setting. Otherwise you will get an error that these two settings depend on each other and can't be enabled separatedly.

My fix is to delay these feature flags to bundler 3 so that the deprecation specs pass. Also, since before giving this advice I'd like to study why we have two different settings that can't be enabled separately, and why the can't be merged to a single one, I have removed that part of the message for now.

I chose this fix because it keeps me moving with reviewing all the deprecations for bundler 3 breaking changes, and makes sure that the deprecations for this change of behavior are tested (and passing).

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -643,7 +643,7 @@ module Bundler
     end
 
     def converge_rubygems_sources
-      return false if Bundler.feature_flag.lockfile_uses_separate_rubygems_sources?
+      return false if Bundler.feature_flag.disable_multisource?
 
       changes = false
 
@@ -915,7 +915,7 @@ module Bundler
       # look for that gemspec (or its dependencies)
       default = sources.default_source
       source_requirements = { :default => default }
-      default = nil unless Bundler.feature_flag.lockfile_uses_separate_rubygems_sources?
+      default = nil unless Bundler.feature_flag.disable_multisource?
       dependencies.each do |dep|
         next unless source = dep.source || default
         source_requirements[dep.name] = source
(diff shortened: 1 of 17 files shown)


deivid-rodriguez:
This PR also includes some extra changes to refactor deprecation specs, and to move the tests for multiple sources support deprecation to the file that tests all deprecations.

As always, I can move those to a separate PR if reviewers feel it's needed :+1:.

deivid-rodriguez:
Ok, so this is failing because this introduces the lockfile incompatibility mentioned in #7007. I would love a review of that PR.

deivid-rodriguez:
> This PR also includes some extra changes to refactor deprecation specs, and to move the tests for multiple sources support deprecation to the file that tests all deprecations.

I ended up extracting these changes to #7105, so I'm going to ask for reviews there first.

deivid-rodriguez:
This PR is now ready. I ended up keeping the instructions to opt-in to the fix in the deprecation message, but merging the `lockfile_uses_separate_rubygems_sources` and `disable_multisource` settings together.

Also, there was a failing spec in the list command specs, but it was not related to the list command, but to a slight change in behavior due to the multisource fix.

I'm unsure whether we want this new behavior or not, but since it was unrelated to the list command and I don't want it to block this PR, I refactored the list command specs to not hit the behavior change, and documented the change separately on a different spec.  See https://github.com/bundler/bundler/pull/7057/commits/a918651d1c810639f7168ae1eb383d38499043a5.

indirect:
@bundlerbot r+

PR #7061, opened 2019-03-23
Fix circular require warnings

deivid-rodriguez:

The problem was that testing rubygems in ruby-core with warnings enabled prints a lot of ruby warnings.

My diagnosis was that while requiring `rubygems`, bundler is required [here](https://github.com/rubygems/rubygems/blob/dad8d252f2b1562c4517fa6d73eb39dfbd0c4f37/lib/rubygems.rb#L1184), and then while requiring `bundler`, `rubygems` is required in several places. Thus the circular require warnings are printed.

My fix is to not require `rubygems` from `bundler`. Rubygems is made available by default by `ruby` anyways unless you use the `--disable-gems` option. And if you use that option, requiring `bundler` makes no sense.

I chose this fix because it's cleaner than #6862, in my opinion. This PR closes #6862 by superseding it.

### DIFF
--- a/lib/bundler/compatibility_guard.rb
+++ b/lib/bundler/compatibility_guard.rb
@@ -1,6 +1,5 @@
 # frozen_string_literal: false
 
-require "rubygems"
 require "bundler/version"
 
 if Bundler::VERSION.split(".").first.to_i >= 2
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
Sounds like merging https://github.com/rubygems/rubygems/pull/2621 broke our build. I'll look into it.

deivid-rodriguez:
Ooops, sorry I forgot to use bundlerbot :grimacing:.

PR #7062, opened 2019-03-23
Remove version overwriting hacks

deivid-rodriguez:

The problem was that we are doing some hacks in our version file that are... dangerous. At very least they cause "circular problems" like #6927, but I also have bad feelings about it.

My diagnosis was we're overwriting the version of the currently loaded bundler :grimacing:. This hack is _almost_ unnecessary (see the spec run here with the hack fully removed: https://travis-ci.org/bundler/bundler/builds/509497021. Only three jobs failed, and for old rubies/rubygems/bundler). 

My fix is to limit the hack to `bundler`'s spec run, since it's only needed there.

I chose this fix because fully getting the build green with the hack fully removed is a bit of work. I dedicated a bit of time to investigate one of the failures and it all comes down to the priority of loading default gems vs `-I`. So I think something like https://github.com/rubygems/rubygems/pull/1868 should fix it. But that's out of the scope of this PR.

Closes #6927.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -22,43 +22,43 @@ module Bundler
   ENV.replace(environment_preserver.backup)
   SUDO_MUTEX = Mutex.new
 
-  autoload :Definition,             "bundler/definition"
-  autoload :Dependency,             "bundler/dependency"
-  autoload :DepProxy,               "bundler/dep_proxy"
-  autoload :Deprecate,              "bundler/deprecate"
-  autoload :Dsl,                    "bundler/dsl"
-  autoload :EndpointSpecification,  "bundler/endpoint_specification"
-  autoload :Env,                    "bundler/env"
-  autoload :Fetcher,                "bundler/fetcher"
-  autoload :FeatureFlag,            "bundler/feature_flag"
-  autoload :GemHelper,              "bundler/gem_helper"
-  autoload :GemHelpers,             "bundler/gem_helpers"
-  autoload :GemRemoteFetcher,       "bundler/gem_remote_fetcher"
-  autoload :GemVersionPromoter,     "bundler/gem_version_promoter"
-  autoload :Graph,                  "bundler/graph"
-  autoload :Index,                  "bundler/index"
-  autoload :Injector,               "bundler/injector"
-  autoload :Installer,              "bundler/installer"
-  autoload :LazySpecification,      "bundler/lazy_specification"
-  autoload :LockfileParser,         "bundler/lockfile_parser"
-  autoload :MatchPlatform,          "bundler/match_platform"
-  autoload :ProcessLock,            "bundler/process_lock"
-  autoload :RemoteSpecification,    "bundler/remote_specification"
-  autoload :Resolver,               "bundler/resolver"
-  autoload :Retry,                  "bundler/retry"
-  autoload :RubyDsl,                "bundler/ruby_dsl"
-  autoload :RubyGemsGemInstaller,   "bundler/rubygems_gem_installer"
-  autoload :RubyVersion,            "bundler/ruby_version"
-  autoload :Runtime,                "bundler/runtime"
-  autoload :Settings,               "bundler/settings"
-  autoload :SharedHelpers,          "bundler/shared_helpers"
-  autoload :Source,                 "bundler/source"
-  autoload :SourceList,             "bundler/source_list"
-  autoload :SpecSet,                "bundler/spec_set"
-  autoload :StubSpecification,      "bundler/stub_specification"
-  autoload :UI,                     "bundler/ui"
-  autoload :URICredentialsFilter,   "bundler/uri_credentials_filter"
-  autoload :VersionRanges,          "bundler/version_ranges"
+  autoload :Definition,             File.expand_path("bundler/definition", __dir__)
+  autoload :Dependency,             File.expand_path("bundler/dependency", __dir__)
+  autoload :DepProxy,               File.expand_path("bundler/dep_proxy", __dir__)
+  autoload :Deprecate,              File.expand_path("bundler/deprecate", __dir__)
+  autoload :Dsl,                    File.expand_path("bundler/dsl", __dir__)
+  autoload :EndpointSpecification,  File.expand_path("bundler/endpoint_specification", __dir__)
+  autoload :Env,                    File.expand_path("bundler/env", __dir__)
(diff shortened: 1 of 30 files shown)


deivid-rodriguez:
Ok, so I looked a bit more into this, and it turns out that this "version hack" is only needed for rubygems versions the doesn't provide the `BundlerVersionFinder`. So I added a check for the rubygems version, and removed the hack.

While debugging this, I added a few other changes that are not directly related to this, so I have extracted them to separate PRs (#7136, #7137, #7138, #7139).

@segiddins What do you think about this? I really don't like changing the version of a loaded gem, if we can avoid it :grimacing:...

lib/bundler.rb
```diff
@@ -27,36 +27,36 @@ module Bundler
   autoload :DepProxy,               "bundler/dep_proxy"
   autoload :Deprecate,              "bundler/deprecate"
   autoload :Dsl,                    "bundler/dsl"
-  autoload :EndpointSpecification,  "bundler/endpoint_specification"
+  autoload :EndpointSpecification,  File.expand_path("bundler/endpoint_specification", __dir__)
```
segiddins: this seems... not right to me

.travis.yml
```diff
@@ -13,6 +13,7 @@ branches:
     - trying
     - /.+-dev$/
     - /.+-stable$/
+    - more_relative_code_loading
```
segiddins: revert

spec/bundler/shared_helpers_spec.rb
```diff
@@ -261,15 +261,6 @@
       subject.set_bundle_environment
     end
 
-    it "ignores if bundler_ruby_lib is same as rubylibdir" do
```
segiddins: why is this spec removed?

spec/runtime/load_spec.rb
```diff
@@ -56,19 +57,6 @@
         Bundler.load
       end.to raise_error(Bundler::GemfileNotFound, /omg\.rb/)
     end
-
-    it "does not find a Gemfile above the testing directory" do
```
segiddins: why is this removed?

deivid-rodriguez:
Sorry @segiddins. This is not ready for review. A general idea of whether you think this is a good idea is appreciated though.

lib/bundler.rb
deivid-rodriguez: This is just an experiment not meant to stay in this PR, but I actually planned to open a separate PR to migrate our autoloads to require absolute paths. I think it would be a nice follow up to https://github.com/bundler/bundler/pull/7100 since we would make sure that we never use the $LOAD_PATH for requiring our own internal code.

What do you think it is not right?

By the way, it would be prettier if we had this: https://github.com/bundler/bundler/pull/7100 :smiley:.

.travis.yml
deivid-rodriguez: WIP code, not ready for review.

spec/bundler/shared_helpers_spec.rb
deivid-rodriguez: This is part of an experiment of the hypothesis that if we never use the $LOAD_PATH, we can revert several patches and workarounds for issues introduced in the past, and we no longer need to special case bundler when manipulating the $LOAD_PATH.

But it's not meant to stay in this PR.

spec/runtime/load_spec.rb
deivid-rodriguez: This is part of https://github.com/bundler/bundler/pull/7138, can you review that PR instead for the rationale? That one is ready for review :+1:.

deivid-rodriguez:
@segiddins This should be ready for a review now!

lib/bundler.rb
deivid-rodriguez: Marking this as solved because I removed the changes from this PR. I'll open a separate discussion for migrating our autoloads to not use the load path.

deivid-rodriguez:
These version overwriting hacks bit me again in #7166, but I think, at least some of them can be fixed for all supported rubygems versions, so I'm back to working on this PR :sweat_smile:. I'll re-request a review when I'm ready :)

deivid-rodriguez:
I'm so happy I finally managed to keep this part of the diff

```diff
diff --git a/lib/bundler/version.rb b/lib/bundler/version.rb
index cfd630095..21eeeb0fd 100644
--- a/lib/bundler/version.rb
+++ b/lib/bundler/version.rb
@@ -1,23 +1,7 @@
 # frozen_string_literal: false
 
 module Bundler
-  # We're doing this because we might write tests that deal
-  # with other versions of bundler and we are unsure how to
-  # handle this better.
-  VERSION = "2.1.0.pre.1".freeze unless defined?(::Bundler::VERSION)
-
-  def self.overwrite_loaded_gem_version
-    begin
-      require "rubygems"
-    rescue LoadError
-      return
-    end
-    return unless bundler_spec = Gem.loaded_specs["bundler"]
-    return if bundler_spec.version == VERSION
-    bundler_spec.version = Bundler::VERSION
-  end
-  private_class_method :overwrite_loaded_gem_version
-  overwrite_loaded_gem_version
+  VERSION = "2.1.0.pre.1".freeze
 
   def self.bundler_major_version
     @bundler_major_version ||= VERSION.split(".").first.to_i
```

while also keeping CI green for all supported rubygems & rubies :tada:

The price to pay is the admittedly ugly style of using absolute paths with autoloads, but I think it's definitely worth it. We could make it prettier if [autoload_relative ever lands into core](https://bugs.ruby-lang.org/issues/15330).

This should fix a bunch of problems with manually installed bundler versions coexisting with bundler versions as default gems :smiley:.

As per vendored dependencies, I created https://github.com/CocoaPods/Molinillo/pull/130 for `Molinillo`, and `fileutils` has already been updated but changes have not yet been incorporated here (I created #7156 to do that separately). The more complicated one would be `thor` because it still supports 1.8.7. I requested that to be dropped in https://github.com/erikhuda/thor/pull/643#issuecomment-490611243. In any case, the activity in the `thor` repo is really low, so I think it's fine to incorporate these changes directly, and propose them upstream once they drop support.

I'd like other to chime in here :smiley:

deivid-rodriguez:
Also, I think this might get green several specs that are currently skipped when running with the "ruby-core" flag.

deivid-rodriguez:
Thanks @indirect!

@bundlerbot r+

PR #7063, opened 2019-03-23
Small fix with plural.

ryanfox1985:
**The problem was** when is outdated is ran always if the gem belongs to one group or more always show "groups" in plural.

**My diagnosis is** if the gem belongs to one group it's incorrect to show "groups" in plural.

**My fix was** if the gem belongs to one group it should show "group" in singular.

### DIFF
--- a/lib/bundler/cli/outdated.rb
+++ b/lib/bundler/cli/outdated.rb
@@ -133,7 +133,7 @@ module Bundler
 
             unless options[:parseable]
               if groups
-                Bundler.ui.info "===== Group #{groups} ====="
+                Bundler.ui.info "===== #{groups_text("Group", groups)} ====="
               else
                 Bundler.ui.info "===== Without group ====="
               end
@@ -166,6 +166,10 @@ module Bundler
     end
 
   private
+
+    def groups_text(group_text, groups)
+      "#{group_text}#{groups.split(",").size > 1 ? "s" : ""} \"#{groups}\""
+    end
 
     def retrieve_active_spec(strict, definition, current_spec)
       if strict
@@ -212,7 +216,7 @@ module Bundler
       elsif options_include_groups || !groups
         "  * #{spec_outdated_info}"
       else
-        "  * #{spec_outdated_info} in groups \"#{groups}\""
+        "  * #{spec_outdated_info} in #{groups_text("group", groups)}"
       end
 
       Bundler.ui.info output_message.rstrip
--- a/spec/commands/outdated_spec.rb
+++ b/spec/commands/outdated_spec.rb
@@ -59,15 +59,19 @@ RSpec.describe "bundle outdated" do
       install_gemfile <<-G
         source "file://#{gem_repo2}"
 
+        gem "terranova", '8'
+
         group :development, :test do
           gem 'activesupport', '2.3.5'
         end
       G
 
       update_repo2 { build_gem "activesupport", "3.0" }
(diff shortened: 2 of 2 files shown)


lib/bundler/cli/outdated.rb
```diff
@@ -91,16 +91,12 @@ def run
           groups = dependency.groups.join(", ")
         end
 
-        outdated_gems_list << { :active_spec => active_spec,
```
colby-swandale: Is there any reason why you're changing the style?

lib/bundler/cli/outdated.rb
```diff
@@ -133,7 +129,7 @@ def run
 
             unless options[:parseable]
               if groups
-                Bundler.ui.info "===== Group #{groups} ====="
+                Bundler.ui.info "===== Group#{groups.split(",").size > 1 ? "s" : ""} #{groups} ====="
```
colby-swandale: This should be put into a variable before, to help keep the code clean.

spec/commands/outdated_spec.rb
```diff
@@ -186,7 +186,7 @@ def test_group_option(group = nil, gems_list_size = 1)
       bundle "outdated --groups"
       expect(out).to include("===== Group default =====")
       expect(out).to include("terranova (newest 9, installed 8, requested = 8)")
-      expect(out).to include("===== Group development, test =====")
+      expect(out).to include("===== Groups development, test =====")
```
colby-swandale: You will need to add a test for the single non-plural header.

lib/bundler/cli/outdated.rb
ryanfox1985: yes duplicated code, when I did the feature group I duplicated the hash https://github.com/bundler/bundler/pull/4841/files#diff-1dacde23b0f908c590b785187df11b00R77

lib/bundler/cli/outdated.rb
ryanfox1985: I will put in a method to no duplicate code.

spec/commands/outdated_spec.rb
ryanfox1985: line 187?

lib/bundler/cli/outdated.rb
ryanfox1985: I will push something better

ryanfox1985:
done, I pushed the lasts changes, this PR is ready for code review

lib/bundler/cli/outdated.rb
deivid-rodriguez: What @colby-swandale meant is that you shouldn't change the current style. First, because we should first agree on any style changes, and apply them consistently, but mainly because it hides the actual point of your PR, namely, fixing the small issue with output pluralization. So the idea would be to undo these particular changes and leave this part of the code untouched because it's not needed for this fix.

ryanfox1985: I agree.

deivid-rodriguez:
Thanks @ryanfox1985!

@bundlerbot r+

deivid-rodriguez:
Actually, can you squash your commits to a single one to keep the commit history a bit cleaner?

@bunderbot r-

deivid-rodriguez:
@bundlerbot r-

ryanfox1985:
squashed.

deivid-rodriguez:
@bundlerbot r+

PR #7066, opened 2019-03-25
Remove bundled Molinillo in favor of RubyGems version.

voxik:
Avoid duplication between Bundler and RubyGems.

I have not really tested this, so let's take it as a start of a discussion ...

Just a few remarks:

1. All supported versions of RubyGems already ships with Molinillo
2. Molinillo was pushed into RubyGems but it is not well maintained. This should stop.


deivid-rodriguez:
We need to upgrade molinillo to 0.6.6 on rubygems first: https://github.com/rubygems/rubygems/pull/2026. Then find a solution for https://github.com/rubygems/rubygems/pull/2026#issuecomment-334768804.

voxik:
> Then find a solution for [rubygems/rubygems#2026 (comment)](https://github.com/rubygems/rubygems/pull/2026#issuecomment-334768804).

That is easy. Stop shipping independent Bundler. That is 10 year old plan AFAIR.

deivid-rodriguez:
The first bundler release was 10 years ago. I doubt the plan when bundler was first released was to stop shipping it independently... Oh well.

Merging bundler & rubygems would fix this issue, yeah. I guess we should close this PR then since this will ultimately get fixed when we finally manage to merge the libraries.

> Molinillo was pushed into RubyGems but it is not well maintained. This should stop.

Not sure what you meant by this exactly but in my opinion molinillo is one of the most stable components right now. Most of the issues we get either here or in rubygems are not related to the resolver.

voxik:
> The first bundler release was 10 years ago. I doubt the plan when bundler was first released was to stop shipping it independently... Oh well.

If not 10 yers ago, then the plan was hashed at least 5 years ago:

https://bundler.io/blog/2013/12/07/the-rumors-of-bundlers-death-have-been-greatly-exaggerated.html

That was the reason for introducing bundler submodule into RubyGems and Molinillo, etc. Now some believes that RubyGems cannot be shipped without Bundler, I believe that the integration is mostly nonexistent and it seems there is some pushback, because even the first obvious steps such as Bundler using Molinillo from RubyGems are postponed for already two years.
 
> Merging bundler & rubygems would fix this issue, yeah. I guess we should close this PR then since this will ultimately get fixed when we finally manage to merge the libraries.

This should be the first step IMO, otherwise it will never materialize.

> > Molinillo was pushed into RubyGems but it is not well maintained. This should stop.
> 
> Not sure what you meant by this exactly but in my opinion molinillo is one of the most stable components right now. Most of the issues we get either here or in rubygems are not related to the resolver.

Molinillo was introduced into RubyGems to have the same resolver as Bundler. When there is different version, it is not the same resolver, therefore this goal failed.

deivid-rodriguez:
Yeah, 5 years would sound like a less exaggerated statement :laughing: 

> This should be the first step IMO, otherwise it will never materialize.

Do we at least agree that upgrading the `molinillo` version used by `rubygems` is a prerequisite for this step? Once that's done, how should we handle future releases of molinillo if we move this unification forward?

voxik:
> Do we at least agree that upgrading the `molinillo` version used by `rubygems` is a prerequisite for this step?

Definitely. Actually first step could be to resolve the 17 test suite errors and possible provide some compatibility layer, as it is the case for other RubyGems features, which were missing in older releases.

> Once that's done, how should we handle future releases of molinillo if we move this unification forward?

Molinillo in both Bundler and RubyGems must be updated together IMO. If Molinillo is stable as you said, then there should be 1.0 release which won't break API anymore.

deivid-rodriguez:
Sounds good, yeah.

I'm pretty sure upgrading `rubygems`  to use the version of `molinillo` that `bundler` uses at the moment would fix the test suite here. So moving rubygems/rubygems#2026 forward, and then providing that compatibility layer here could work. :+1:

segiddins:
I’m very, very strongly against this. Bundler already maintains a whole file of backwards compatibility shims to deal with differences in RubyGems versions, and I really don’t want to have to do that with Molinillo.

Molinillo does not, and probably will not, ever have a stable API, since at it’s root it is a single algorithm — and I reserve the right to change how that algorithm works at any time. And both bundler and rubygems depend (directly and indirectly) on the particulars of how a version of Molinillo’s algorithm works.

hsbt:
We should unify them when integrating RubyGems and Bundler. We are going to discuss this in RubyKaigi days.

colby-swandale:
I'm going to close this. The Bundler/RubyGems core team have an idea to merge Bundler and RubyGems into the same repository in the future. When that occurs we will then look at having them use the same vendor code.

Right now though, merging this PR will create some major and complex problems such as Bundler having to support many different versions of Molinillo. Which IMO, greatly outweighs the benefits.

simi:
> ...and **I reserve the right** to change how that algorithm works at any time...

@segiddins is it personal?

deivid-rodriguez:
The plan sounds good to me too! :+1:

greysteil:
I’m 👍 on what @segiddins said. Unless we’re very clear about the definition of backwards compatibility *not* including changes to the algorithm (even ones that change the output) then pretty much every molinillo change would be a major version bump.

PR #7067, opened 2019-03-25
Tweaking vendoring

deivid-rodriguez:

The problem was that we were not using the latest versions of some of our vendored dependencies.

My diagnosis was that we should upgrade them.

My fix is to upgrade them using `automatiek`, and add a few tweaks to our vendoring setup.

I chose this fix because.... I didn't really considered other options.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -297,9 +297,24 @@ namespace :man do
   end
 end
 
-begin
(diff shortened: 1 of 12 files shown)


hsbt:
@bundlerbot r+

PR #7068, opened 2019-03-27
Bump vendored fileutils to 1.2.0

deivid-rodriguez:

The problem was we manually updated some code in the vendored fileutils instead of using a released version.

My diagnosis was that we should use the recently released fileutils 1.2.0.

My fix is to vendor it through `bin/rake vendor:fileutils[v1.2.0]`.

I chose this fix because it properly upgrades the vendored code to match what's last released.


hsbt:
@bundlerbot r+

PR #7070, opened 2019-03-27
Improve cross repo integration

deivid-rodriguez:

The problem was that we had on `rubygems`, we're not testing bundler against the rubygems checkout of each PR, but against master. That is confusing because developers assume that the PRs tests will run against the PRs code, but this is currently not true.

My diagnosis was that on rubygems we are [specifying `RGV=master`](https://github.com/rubygems/rubygems/blob/fa6e547330e97b417ed11f262d68d03c57abeeda/.travis.yml#L12) in our TravisCI matrix. And what `bundler` does with that is to clone rubygems master under `tmp/rubygems` and use that. Thus your rubygems PR is not really tested.

My fix is to rework the setup to accept an `RGV` environment variable when it contains. In that case, no git operations will be performed, and that path will be used directly and assumed to contain a rubygems checkout.

I also refactored a few things about the setup while at it.

There were existing tasks to test bundler against a rubygems checkout, but they were unused and they expected a `RG` environment variable instead. I decided to reuse these tasks, but make them pick up the `RGV` as long as it contains a valid path. I chose this fix because it works and it doesn't make the existing setup too different.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -145,7 +145,7 @@ namespace :spec do
 
     desc "Run specs under a RubyGems checkout (set RG=path)"
     task "co" do
-      sh("bin/rspec")
+      sh("bin/rspec --format progress")
     end
 
     task "setup_co" do
--- a/Rakefile
+++ b/Rakefile
@@ -3,8 +3,7 @@
 $:.unshift File.expand_path("../lib", __FILE__)
 require "benchmark"
 
-NULL_DEVICE = (Gem.win_platform? ? "NUL" : "/dev/null")
-RUBYGEMS_REPO = if `git -C "#{File.expand_path("..")}" remote --verbose 2> #{NULL_DEVICE}` =~ /rubygems/i
+RUBYGEMS_REPO = if `git -C "#{File.expand_path("..")}" remote --verbose 2> #{IO::NULL}` =~ /rubygems/i
   File.expand_path("..")
 else
   File.expand_path("tmp/rubygems")
--- a/Rakefile
+++ b/Rakefile
@@ -126,7 +126,7 @@ namespace :spec do
     (branches + releases).each do |rg|
       desc "Run specs with RubyGems #{rg}"
       task rg do
-        sh("bin/rspec --format progress --color")
+        sh("bin/rspec --format progress")
       end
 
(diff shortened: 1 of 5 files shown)


deivid-rodriguez:
After this PR, we can change `RGV=master` to `RGV=..` in the TravisCI config in rubygems, and our PRs will start checking the rubygems code in the PR as expected.

Rakefile
```diff
@@ -136,43 +129,26 @@ namespace :spec do
...
-          raise "need to be running against master with bundler as a submodule"
-        end
-
-        puts "Checked out rubygems '#{rg}' at #{hash}"
+      task "set_rubygems_#{rg}" do
         ENV["RGV"] = rg
       end
 
-      task rg => ["clone_rubygems_#{rg}"]
+      task rg => ["set_rubygems_#{rg}"]
```
bronzdoc: It isn't redundant to name this `set_rubygems_#{rg}` since we already in the `rubygems` namespace? what about just `set_#{rg}` so we can call something like  `spec:rubygems:set_#{rg}`

spec/runtime/setup_spec.rb
```diff
@@ -1213,7 +1213,7 @@ def lock_with(ruby_version = nil)
 
     describe "default gem activation" do
       let(:exemptions) do
-        if Gem::Version.new(Gem::VERSION) >= Gem::Version.new("2.7") || ENV["RGV"] == "master"
```
bronzdoc: why we don't need this check anymore?

bronzdoc:
LGTM @deivid-rodriguez just some minor comments

spec/runtime/setup_spec.rb
deivid-rodriguez: Rubygems master is 3.1 nowadays, so the first condition should suffice, I think.

But, yeah, this change is actually independent from this PR, but since I was at it I reviewed usages of the `RGV` environment variable and found this one was unnecessary.

Rakefile
deivid-rodriguez: I renamed it as you suggested, thanks! :+1:

deivid-rodriguez:
Thanks for having a look @bronzdoc!

@bundlerbot r+

PR #7071, opened 2019-03-27
Correction to : bundle outdated --source not working as expected

### LINKED ISSUE #4881: bundle outdated --source not working as expected

samyak-jn:

The problem was bundle outdated --source not working as expected.

My diagnosis was to add --source in the directory of spec/commands/outdated_spec.rb and add a basic spec to it, I'm a beginner at work, trying to learn.

My fix was to add the basic spec for outdated, I need guidance on the same to enhance and correct it.

I chose this fix because it was needed for the issue, I guess.

Closes #4881.

### DIFF
--- a/spec/commands/outdated_spec.rb
+++ b/spec/commands/outdated_spec.rb
@@ -792,3 +792,13 @@ RSpec.describe "bundle outdated" do
     end
   end
 end
+
+
+describe "with --source" do
+      bundle "outdated --source #{source}"
+      expect(out).to include("Bundle source !")
+      # Gem names are one per-line, between "*" and their parenthesized version.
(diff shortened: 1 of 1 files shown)


indirect:
Hi @samyak-jn, please check on the failing build on Travis and resolve the issues there. Thanks!

PR #7072, opened 2019-03-28
Review `bundle show` deprecation

deivid-rodriguez:

The problem was there was some work to do regarding the deprecation of `bundle show` and the `bundle list` introduction.

* The command should error out in bundler 3 and was still there.
* The command would print erroneous deprecation messages for the `--verbose` and `--outdated` flags. These flags to `bundle show` are undocumented and will be removed in bundler 3.
* The `bundle list` help would show the help for the new `list` command that lists groups of (or all) gems in the Gemfile, but its implementation would just be an alias to `bundle show`.

My fix is to remove the command in bundler 3, to show a different deprecation message when `--outdated` or `--verbose` are used to inform about their removal without replacement, and the inconditionally introduce the new list command without any feature flag. 

I chose this fix because:

* The `--outdated` and `--verbse` options are undocumented, so nobody should be using them. Still, I give a correct deprecation message for any users they might have used them.
* The `list` command was super confusing because it was an alias to `show` but that was not documented anywhere. So, I consider find to introduce the new `bundle list` now to match the documentation.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -320,7 +320,7 @@ module Bundler
     desc "list", "List all gems in the bundle"
     method_option "name-only", :type => :boolean, :banner => "print only the gem names"
     method_option "only-group", :type => :string, :banner => "print gems from a particular group"
-    method_option "without-group", :type => :string, :banner => "print all gems expect from a group"
+    method_option "without-group", :type => :string, :banner => "print all gems except from a group"
     method_option "paths", :type => :boolean, :banner => "print the path to each gem in the bundle"
     def list
       require "bundler/cli/list"
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -282,7 +282,7 @@ module Bundler
       end
     end
 
-    unless Bundler.feature_flag.list_command?
+    unless Bundler.feature_flag.bundler_3_mode?
       desc "show GEM [OPTIONS]", "Shows all gems that are part of the bundle, or the path to a given gem"
       long_desc <<-D
         Show lists the names and versions of all gems that are required by your Gemfile.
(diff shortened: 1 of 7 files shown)


deivid-rodriguez:
I still want to add a few things to this PR so I marked it as "status: working", not sure if it's the right way to mark a PR as WIP.

deivid-rodriguez:
Ok, this should be ready now, let me know any thoughts about this.

lib/bundler/cli.rb
```diff
@@ -282,50 +282,51 @@ def update(*gems)
...
-        new_command = rest.find {|arg| !arg.start_with?("--") } ? "info" : "list"
-
-        new_arguments = rest.map do |arg|
-          next arg if arg != "--paths"
-          next "--path" if new_command == "info"
-        end
+    desc "list", "List all gems in the bundle"
+    method_option "name-only", :type => :boolean, :banner => "print only the gem names"
+    method_option "only-group", :type => :string, :banner => "print gems from a particular group"
+    method_option "without-group", :type => :string, :banner => "print all gems expect from a group"
```
indirect: tiny typo, `expect` → `except`

deivid-rodriguez:
> If I'm remembering our deprecation plan correctly, we should keep `show` and make it print a warning in Bundler 2.1 before removing it in Bundler 3, right? Or maybe I am remembering wrong.

No, that's exactly right :+1:. This PR keeps that, but it actually removes the command unles `Bundler.feature_flag.bundler_3_mode?`. The deprecation was already there, but the removal was only a TODO note but not actually done.

The diff is a bit confusing in this regard, I'll simplify it while I fixed the typo you found!

lib/bundler/cli.rb
deivid-rodriguez: Fixed in https://github.com/bundler/bundler/pull/7072/commits/a080d9bc2c08c6d6c799b600b40f1df3dd502300.

deivid-rodriguez:
I ended up leaving the diff as it was, I think it's fine.

deivid-rodriguez:
Actually ended up changing the diff a bit, although I'm not sure it makes it more clear. Anyways, it should be ready now.

indirect:
@bundlerbot r+

PR #7073, opened 2019-03-28
Unskip passing spec

deivid-rodriguez:

The problem was that I discovered by chance that one of the skipped specs seems to be passing.

My diagnosis was that something has changed for the better that made this spec green.

My fix is to remove the skip.

I chose this fix because it's good :)


colby-swandale:
@bundlerbot r+

PR #7074, opened 2019-03-28
Remove unnecessary test stuff

deivid-rodriguez:

The problem was that our test setup could be simpler.

My diagnosis was that some lines installing test gem dependencies could be removed from the Rakefile, because the tests already detect this situation and auto-install test gems appropriately.

My fix is to remove the lines. However, removing the lines led to a single test failure that can be explained becase `FileUtils` ends up being required too late, thus skipping one call to `Gem.clear_paths` done in the tests setup so that stuff memoized by `rubygems` is cleared and tests can assume is not loaded yet. The situation is very intricate but I tried to explain it in the commit message.

I chose this fix because it fixes the test failure, but I also plan to propose to eliminate the `autoload` calls in rubygems that also contributed to this issue.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -54,11 +54,6 @@ namespace :spec do
       "'#{name}:#{version}'"
     end.join(" ")
     sh %(#{Gem.ruby} -S gem #{gem_install_command})
-
-    # Download and install gems used inside tests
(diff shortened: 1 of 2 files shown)


colby-swandale:
@bundlerbot r+

PR #7075, opened 2019-03-28
Convert binstub conflict deprecation to a warning

deivid-rodriguez:

The problem was that this specific deprecation I don't think it makes sense as a deprecation.

My diagnosis was that nothing is changing in bundler so that this needs to be a deprecation. This is just a potentially undesired situation that can be caused by gems shipping executables conflicting with the executables of other gems, and that can be fixed by using project specific binstubs.

My fix is to change this deprecation message to a warning, and to get it tested.

I chose this fix because it makes sense to me. With this, we can start printing this warning now without thinking of a deprecation path because I don't think we're changing anything here, really.

### DIFF
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -442,14 +442,10 @@ module Bundler
         raise ArgumentError, "you must supply exec_name" unless exec_name
 
         spec_with_name = specs_by_name[gem_name]
-        spec =
-          if spec_with_name && spec_with_name.executables.include?(exec_name)
-            spec_with_name
-          else
-            specs.find {|s| s.executables.include?(exec_name) }
-          end
+        matching_specs_by_exec_name = specs.select {|s| s.executables.include?(exec_name) }
+        spec = matching_specs_by_exec_name.delete(spec_with_name)
 
-        unless spec
+        unless spec || !matching_specs_by_exec_name.empty?
           message = "can't find executable #{exec_name} for gem #{gem_name}"
           if spec_with_name.nil?
             message += ". #{gem_name} is not currently included in the bundle, " \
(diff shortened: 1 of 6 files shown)


deivid-rodriguez:
I noticed that this warning is actually hardly ever useful, because the only case it was detecting wouldn't cause any problems because even if it was loading an incorrect binstub, the gem finally loaded is the one the user would expect.

However, the situation were multiple conflicting executables were being used in the same app was going unnoticed. I'll try to fix this.

deivid-rodriguez:
Ok. I'm quite satisfied with this PR now. This is what it does:

* Simplify the `find_spec_for_exe` to assume `exec_name` is always passed. The previous version was defaulting to `Gem::Specification.default_executable` (an executable named just like the gem), but that method has been long deprecated in rubygems. This changes goes together with https://github.com/rubygems/rubygems/pull/2706.

* Convert the previous deprecation to a warning as explained in the PR body.

* Add a new warning about conflicting executables with the same name in the same application. The warning text I added is (for example, taking from the new spec):

    > The `rackup` executable in the `fake` gem is being loaded, but it's also present in other gems (`rack`). If you meant to run the executable for another gem, make sure you use a project specific binstub (`bundle binstub <gem_name>`). If you plan to actually use _both_ conflicting executables, generate binstubs for both and disambiguate their names.

As a note, I had to change a previously existing spec, but in my opinion the spec I changed didn't make any sense. It's this one:

https://github.com/bundler/bundler/blob/53989a9812338b804f2c951b6077dfe296d72a99/spec/install/binstubs_spec.rb#L23-L42

The situation described in the test is pretty much undefined. There's no information to know which executable is the good one. As a matter of fact, I run the spec several times under bundler 2 and the results as per which binstub is loaded were undefined, I guess it comes down to which gem is downloaded and installed in the last place.

deivid-rodriguez:
Finally got this one green!

PR #7078, opened 2019-03-28
Allow `update` to install when `--no-install` used

### LINKED ISSUE #7077: `update <gem>` is not installing gems (merely packaging), when BUNDLE_NO_INSTALL: "true"
- `package --no-install` writes a `.bundle/config` with `BUNDLE_NO_INSTALL: "true"`
- which causes `update <gem>` to not install

this is not the documented behavior of `BUNDLE_NO_INSTALL`

> [no_install (BUNDLE_NO_INSTALL): Whether bundle package should skip installing gems.](https://bundler.io/v2.0/bundle_config.html)

bundler 2.0.1, RubyGems      3.0.3, ruby 2.6.2

deivid-rodriguez:
Fixes #7077.

The problem was #7077. When the `no_install` configuration is activated, this prevents `bundle update` from installing gems, but `no_install` is only meant to affect `bundle package`.

My diagnosis was that `bundle update` needs to ignore this setting.

My fix is to the same `bundle install` does to fix this problem.

I chose this fix because it's the most straightforward solution, although the handling of this flag could probably use some refactoring.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -277,7 +277,9 @@ module Bundler
     def update(*gems)
       SharedHelpers.major_deprecation(2, "The `--force` option has been renamed to `--redownload`") if ARGV.include?("--force")
       require "bundler/cli/update"
-      Update.new(options, gems).run
+      Bundler.settings.temporary(:no_install => false) do
+        Update.new(options, gems).run
+      end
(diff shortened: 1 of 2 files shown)


colby-swandale:
@bundlerbot r+

PR #7079, opened 2019-03-29
Remove `bundle config` deprecations from specs

deivid-rodriguez:

The problem was that some specs still use the deprecated `bundle config interface`

My fix is to use the newer interface.


hsbt:
@bunderbot r+

deivid-rodriguez:
Previous comment had a typo and was ignored.

@bundlerbot r=hsbt

PR #7080, opened 2019-03-29
Always pass the `--all` flag when needed

deivid-rodriguez:

The problem was that some specs were still using `bundle update` to update all gems, which is deprecated.

My fix is to update all specs to use the new preferred command, namely, `bundle update --all`.


deivid-rodriguez:
Another standard PR removing deprecations from our specs, merging in!

@bundlerbot r+

PR #7082, opened 2019-03-30
Complete some missing specs

deivid-rodriguez:
This is a follow up to #7052.

The problem was that the `Bundle.with_clean_env` & `Bundle.with_original_env` family was missing some deprecation specs, and also the specs were spread out across different files.

My fix is to add the missing specs, and centralize them in the deprecation specs file.


deivid-rodriguez:
Straightforward and affects tests only, so merging!

@bundlerbot r+

Issue #6930, opened 2019-01-24
full_gem_path is invalid for default gems on Ruby 2.5+


Fryguy:
Using both bundler 2.0.1 and 1.17.2, with Ruby 2.5, it seems that if the Gemfile contains a gem that also happens to be a default gem (i.e. those that are packaged with ruby), then rubygems creates a Bundler::StubSpecification instance with an invalid full_gem_path.

Reproducer:

Gemfile:
```ruby
gem 'json'
```

Shell with Ruby 2.5.3:
```sh
> bundle update
> bundle exec irb

irb(main):001:0> s = Bundler.environment.specs.detect { |s| s.full_gem_path.include?("json-2.1.0") }
=> #<Bundler::StubSpecification:0x00007fc78508b548 @name="json", ... >
irb(main):002:0> s.full_gem_path
=> "/Users/jfrey/.rubies/ruby-2.5.3/lib/ruby/gems/2.5.0/gems/json-2.1.0"
irb(main):003:0> Dir.exist?(s.full_gem_path)
=> false
```

Notice the gem path is pointing to the ruby installation directory, but it's a directory that doesn't exist. However, this seems to work on Ruby 2.4.5, as it refers me to the installed gem in the .gem directory.

```sh
irb(main):001:0> s = Bundler.environment.specs.detect { |s| s.full_gem_path.include?("json-2.1.0") }
=> #<Bundler::StubSpecification:0x00007fd4bb8a7a68 @name="json", ... >
irb(main):002:0> s.full_gem_path
=> "/Users/jfrey/.gem/ruby/2.4.5/gems/json-2.1.0"
irb(main):003:0> Dir.exist?(s.full_gem_path)
=> true
```

FWIW, I am using chruby/ruby-install, which installs gems to ~/.gem and rubies to ~/.rubies

Thanks to @NickLaMuro for helping me track this down.

Fryguy:
I've tried this with a number of system gems, and it happens only with the "default" ones.  That is, gems like `did_you_mean`, do not exhibit this problem because they are fully extracted and actually exist in the directory:

```sh
> ls -1 /Users/jfrey/.rubies/ruby-2.5.3/lib/ruby/gems/2.5.0/gems/
did_you_mean-1.2.0
minitest-5.10.3
net-telnet-0.1.1
power_assert-1.1.1
rake-12.3.0
rdoc-6.0.1
test-unit-3.2.7
xmlrpc-0.3.0
```

[...]

PR #7083, opened 2019-03-31
Don't consider default gems on `bundle info`

deivid-rodriguez:

The problem was that while old `bundle show` did not consider default gems, the new `bundle info` does.

My diagnosis was that we should keep the same behavior as `bundle show`.

My fix is remove the default gems specific code from the command.

I chose this fix because it's consistent with other commands. If users want this in the feature, we can always add it under a command flag or whatever.

### DIFF
--- a/lib/bundler/cli/info.rb
+++ b/lib/bundler/cli/info.rb
@@ -26,13 +26,7 @@ module Bundler
 
     def spec_for_gem(gem_name)
       spec = Bundler.definition.specs.find {|s| s.name == gem_name }
-      spec || default_gem_spec(gem_name) || Bundler::CLI::Common.select_spec(gem_name, :regex_match)
-    end
-
-    def default_gem_spec(gem_name)
-      return unless Gem::Specification.respond_to?(:find_all_by_name)
-      gem_spec = Gem::Specification.find_all_by_name(gem_name).last
-      return gem_spec if gem_spec && gem_spec.respond_to?(:default_gem?) && gem_spec.default_gem?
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
References https://github.com/bundler/bundler/pull/7026#issuecomment-478303327.

colby-swandale:
This was an intentional decision and is defiantly not a bug. Default gems are libraries that can be loaded by the application which the user may want to get more information about.

Yes these gem may not be in the Bundle, but pointing out to the user of a gem being one that has shipped with ruby is something i consider a suitable compromise compared to just showing the user an error.

deivid-rodriguez:
As @indirect pointed out, these gems are not part of the bundle, and other commands, such as `bundle list` already ignore them.

But on top of that, due to https://github.com/bundler/bundler/issues/6930, the information we are giving about the location of the gem is also wrong so I think this is going to be more confusing than helpful.

Ideally this should be consistent, and maybe optionally toggleable through a flag and config options. But given where we are now, I would prefer to remove it for the time being.

colby-swandale:
> these gems are not part of the bundle

Yes, but it's common for applications to use default gems, so i saw this as being a use case that a lot uses may find useful, instead of just saying the gem is not in the bundle. We should fix #6930 instead of just straight up removing this.

deivid-rodriguez:
This is really minor. I only opened this PR because @indirect and I seemed to agree on this, but I don't think none of us have a strong opinion here. At least I don't, so closing! :+1:

PR #7088, opened 2019-04-01
Bump version to 2.1.0.pre.1.

hsbt:

When we released the new version, We got the test failure or an error with a version number like 2.0.0.

We didn't test with new version number until release time.

I did bump to the minor version in the master branch.

We should develop with a new version number like 2.1.0.

### DIFF
--- a/lib/bundler/version.rb
+++ b/lib/bundler/version.rb
@@ -7,7 +7,7 @@ module Bundler
   # We're doing this because we might write tests that deal
   # with other versions of bundler and we are unsure how to
   # handle this better.
-  VERSION = "2.1.0.beta1" unless defined?(::Bundler::VERSION)
+  VERSION = "2.1.0.pre.1" unless defined?(::Bundler::VERSION)
 
   def self.overwrite_loaded_gem_version
     begin
--- a/lib/bundler/version.rb
+++ b/lib/bundler/version.rb
@@ -7,7 +7,7 @@ module Bundler
   # We're doing this because we might write tests that deal
   # with other versions of bundler and we are unsure how to
   # handle this better.
-  VERSION = "2.0.0.dev" unless defined?(::Bundler::VERSION)
+  VERSION = "2.1.0.beta1" unless defined?(::Bundler::VERSION)
 
   def self.overwrite_loaded_gem_version
     begin


deivid-rodriguez:
Let me sort out the test failure here because I already investigated it in #6957. :+1:

colby-swandale:
Sorry, i didn't mean to mark as a review as "request changes". It would be good to merge #6957 that will bump the version instead of making a separate change.

deivid-rodriguez:
For what it's worth, I prefer this PR. Smaller changes means understanding them better.

For example, in this case the test failure caused by this PR helped me understand the issue better and taught me why the changes I proposed in #7089 are needed. While they were buried in #6957 I knew they fixed some issue but I couldn't recall why they were needed.

lib/bundler/version.rb
```diff
@@ -7,7 +7,7 @@ module Bundler
   # We're doing this because we might write tests that deal
   # with other versions of bundler and we are unsure how to
   # handle this better.
-  VERSION = "2.0.0.dev" unless defined?(::Bundler::VERSION)
+  VERSION = "2.1.0.beta1" unless defined?(::Bundler::VERSION)
```
hsbt: Should I change this to `.dev`? I think we should set the real case like `.beta1`.

colby-swandale: we don't use `beta` in releases. It would `2.1.0.pre.1`

deivid-rodriguez: :+1:

I don't have a strong opinion, but I slightly prefer `.beta1`, yeah.

Given how rubygems version comparison works, `2.1.0.dev` ranks higher than `2.1.0.beta1`, so it could potentially lead to some confusion. I prefer to set this to the minimum target version that we might release next. If we end up deciding to go straight to `rc1` for example, we can change it again when we make that decision.

deivid-rodriguez: > we don't use beta in releases. It would 2.1.0.pre.1

That's true. Let's go with `2.1.0.pre.1` then :+1:.

deivid-rodriguez:
@hsbt #7089 landed, you can update the version and rebase now :+1:

colby-swandale:
@bundlerbot r+

hsbt:
Thanks!

PR #7089, opened 2019-04-01
Normalize file URIs in spec's lockfiles

deivid-rodriguez:

The problem was that bumping bundler's version in #7088 made a lockfile spec fail.

My diagnosis was that there was a combination of bugs that was making this spec pass, but the spec was incorrect.

My fix is to change the test version the spec uses to not hit https://github.com/rubygems/rubygems/issues/2595, and thus reproduce the failure. And then fix the spec that was incorrect because the lockfile written had different URLs (`file://localhost/<stuff>`) from the lockfile we were checking against (`file://<stuff>`), thus tricking `bundle install` into thinking something had changed, and thus making it rewrite it with an incorrect bundler version.

I chose this fix because it makes #7088 pass and it reduces surprises.

### DIFF
--- a/spec/lock/lockfile_spec.rb
+++ b/spec/lock/lockfile_spec.rb
@@ -76,7 +76,7 @@ RSpec.describe "the lockfile format", :bundler => "2" do
   end
 
   it "does not update the lockfile's bundler version if nothing changed during bundle install" do
-    version = "#{Bundler::VERSION.split(".").first}.0.0.0.a"
+    version = "#{Bundler::VERSION.split(".").first}.0.0.a"
 
     lockfile <<-L
       GEM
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
This PR only affects tests, but it's very intricate. I rebased [the bug fix in rubygems](https://github.com/rubygems/rubygems/pull/2651) for one of the issues involved here, and [exactly this spec failed](https://travis-ci.org/rubygems/rubygems/jobs/514210152) when testing against master. Seeing that failure can help understanding the bug in this test.

hsbt:
👍  After merging this, I'm going to rebase #7088

hsbt:
@bundlerbot r+

PR #7090, opened 2019-04-01
Make sure spec fails if gems cannot be installed

deivid-rodriguez:

The problem was if the installation if system gems failed during specs failed, that was being silently ignored.

My diagnosis was that if this installation fails, the specs that use this helper are not probably testing what they should

My fix is to use the bang method instead.


hsbt:
@bundlerbot r+

PR #7091, opened 2019-04-02
Change info command spec to use `bundle info` instead of `bundle show`

colby-swandale:

A test in the info command spec was using `bundle show` instead of `bundle info`.

This is causing specs in Bundler 3 to fail. See https://travis-ci.org/bundler/bundler/jobs/514507744

Change the test to use `bundle info` instead.


colby-swandale:
@bundlerbot r+

PR #7093, opened 2019-04-03
Merge lockfile spec files

deivid-rodriguez:

The problem was current lockfile specs are hard to maintain.

My diagnosis was that we were maintaining separate files per version for the lockfile specs. I imagine we did this because at some point the lockfiles were too different to make it worth maintaining a single file with some filters on the bundler version. I don't think this is the case anymore.

My fix is to merge both files.

I chose this fix because I think it simplifies maintenance and it gives a quick view of the differences in the lock file format across versions in a single place.


deivid-rodriguez:
Thanks @colby-swandale!

@bundlerbot r=colby-swandale

PR #7100, opened 2019-04-04
Prefer `require_relative` for internal requires

deivid-rodriguez:

The problem was that `bundler` seems to, in some very rare cases, leak to the copy of itself installed as a default gem. I have been able to reproduce this only for stuff that we have already fixed. For example: https://github.com/bundler/bundler/pull/6502. However, I have the gut feeling that this can still happen under some conditions, because sometimes we still get reports from people using bundler 2, and getting the error "You must user Bundler 2 or greater with this Gemfile". 

My diagnosis was that somehow, due to the complicated LOAD_PATH manipulation bundler does, we may endup requiring bundler files in another copy of bundler.

My fix is not really a fix, although it _might_ prevent the potential issue from happening. As @colby-swandale would say, we should fix the real culprit instead. However, I think using `require_relative` is a better practice anyways, because it makes it clear that you are requiring "internal" files and not files from some dependencies. And it should also be faster because it does not search the LOAD_PATH. And it skips the rubygems monkeypatches to `require`, which seems also good.

I chose this fix because I think it's a good practice.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -1,20 +1,20 @@
 # frozen_string_literal: true
 
-require "bundler/compatibility_guard"
+require_relative "bundler/compatibility_guard"
 
-require "bundler/vendored_fileutils"
+require_relative "bundler/vendored_fileutils"
 require "pathname"
 require "rbconfig"
 
-require "bundler/errors"
-require "bundler/environment_preserver"
-require "bundler/plugin"
-require "bundler/rubygems_ext"
-require "bundler/rubygems_integration"
-require "bundler/version"
-require "bundler/constants"
-require "bundler/current_ruby"
-require "bundler/build_metadata"
+require_relative "bundler/errors"
+require_relative "bundler/environment_preserver"
+require_relative "bundler/plugin"
+require_relative "bundler/rubygems_ext"
+require_relative "bundler/rubygems_integration"
+require_relative "bundler/version"
+require_relative "bundler/constants"
+require_relative "bundler/current_ruby"
+require_relative "bundler/build_metadata"
 
 module Bundler
   environment_preserver = EnvironmentPreserver.new(ENV, EnvironmentPreserver::BUNDLER_KEYS)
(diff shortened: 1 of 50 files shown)


deivid-rodriguez:
Oh, I also introduced a separate change, but that I think it goes in the same direction. It's https://github.com/bundler/bundler/pull/7100/commits/e9a39a907d049d145604972578fddf315b98f7c0. In few words, if we're loading a file inside the bundler installation, we should directly load that instead of going through rubygems... stuff.

Cruzinho10:
Unsubscribe..

segiddins:
Looks good to me! Only thing I’d ask for would be adding a quality spec that asserts there are no `bundler/*` regular requires, so we can stay consistent

walf443:
require_relative seems to got some problem on some ruby version https://github.com/rails/rails/pull/30627

Most libraries does not need to consider this problem, I think. but bundler is widely used. So, it may be in trouble.

deivid-rodriguez:
Thanks for pointing that out @walf443! That's too bad, I wish we could do this without risk... :disappointed:.

@amatsuda Is `bundler` potentially vulnerable to run into this problem? We do have some specs for GEM_HOME being a symlinked location, but I don't think we have specs for just the `bundler` gem being installed at a symlinked location, which is the potential situation where this problem would happen, I think?

deivid-rodriguez:
The ruby fix was backported and [released with 2.4.4](https://github.com/ruby/ruby/commit/eaba9da1d13fac380fe94f977d7f8d89bd40cde0), and it was not backported to 2.3, but 2.3 is already EOL'd. So... I'm on the fence.

deivid-rodriguez:
Ideally, I should justify that doing this actually fixes a default gem activation issue, so this change would be a bug fix. I'll see what I can do!

indirect:
IIRC we said that we would stop supporting 2.3 at the same time as ruby-core, so I think I'm okay with officially dropping support for 2.3 from master, if you want to do that soon-ish?

deivid-rodriguez:
Ok, then I say let's get this in? I can open a separate PR for the support drop if/when we decide to do it.

indirect:
Let's do it. 👍🏻

deivid-rodriguez:
Going in!

@bundlerbot r+

PR #7102, opened 2019-04-05
Remove explicitly loading thread library

hsbt:

It's provided by an embedded module in the modern Ruby.


hsbt:
@bundlerbot r+

PR #7103, opened 2019-04-05
Set RSpec test order to be random

colby-swandale:
### What was the end-user problem that led to 

Bundler's test suite is currently being executed in the same order in both local development and in CI. This can sometimes introduce test dependency where a test depends on state from a previous test.

Set the RSpec order to be random on every execution. If there is a test that is failing because of a test depending on existing state, then we should fix that. If we do come across this problem, the RSpec seed can be used to reproduce the test failure. Example:

```
$ ./bin/rspec

Randomized with seed 33081
```

Run the test suite again in the same order:
```
$ ./bin/rspec --seed 33081
```

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -2,3 +2,4 @@
 --color
 --warnings
 --require spec_helper
+--order random
--- a/Rakefile
+++ b/Rakefile
@@ -97,15 +97,10 @@ namespace :spec do
   end
 
   desc "Run the spec suite with the sudo tests"
-  task :sudo => %w[set_sudo spec clean_sudo]
+  task :sudo => %w[set_sudo spec]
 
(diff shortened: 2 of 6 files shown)


deivid-rodriguez:
This much closer to green that I would've expected! :tada:

colby-swandale:
I've run the entire suite a few times and haven't encountered any other breaking tests. But we should re-run the tests travis a few times to make sure we don't introduce any breaking specs.

deivid-rodriguez:
I worked a bit on this PR a few days ago, but there are still some flakies. We're closer though.

colby-swandale:
I'm closing this, if anyone wants to work on it then feel free to clone/fork.

PR #7104, opened 2019-04-05
Add missing method to rake file

deivid-rodriguez:

The problem was any Rake task depending on the `:build_metadata` task is currently failing with  ``NameError: undefined local variable or method `bundler_spec' for main:Object``. For example, the release or build tasks.

My diagnosis was that I probably broke this during the development environment refactoring.

My fix is to add the missing method to the file that needs it.

I chose this fix because it works and it restores working tasks.

### DIFF
--- a/task/build_metadata.rake
+++ b/task/build_metadata.rake
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
 
+def bundler_spec
+  Gem::Specification.load("bundler.gemspec")
+end
+
 def write_build_metadata(build_metadata)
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Straightforward, going in!

@bundlerbot r+

simi:
Any chance to add spec for this as well to prevent the same problem in the future?

deivid-rodriguez:
If someone has time to add some infrastructure to start testing our rake tasks, I'm definitely not opposed to it! For now I went with the quick fix so that release and local installation tasks work again.

PR #7105, opened 2019-04-05
Simplify deprecation specs

deivid-rodriguez:

The problem was deprecation specs have become convoluted. The spawn many unnecessary subprocesses and do unnecessary work, and have complicated and redundant descriptions.

My fix is to reduce the nesting level of the specs, simplify descriptions, and refactor test setups so that each test only runs the bare minimum that it needs.


deivid-rodriguez:
@bundlerbot r+

PR #7106, opened 2019-04-05
Cache rubocop in travis-ci

colby-swandale:

We run rubocop on every build that goes through every file looking for any breaking rules. Currently this takes about ~30s to execute. We also have the linting step required to pass before any other steps are executed, so having this step execute as fast as possible is important to developers.

I think there is a small optimize we can make to the build time by caching rubocop between builds.

Have travis cache the results from rubocop between builds to speed up build times


deivid-rodriguez:
@bundlerbot r+

PR #7108, opened 2019-04-08
Fix GemHelper to allow reading Input/Output

### LINKED ISSUE #6854: Enabling TOTP on rubygems.org breaks `rake release`
If you enable 2FA for both UI and API, you can't release new gems with `rake release`. ~AFAICT, there's no way one can inform a TOTP code when issuing a `gem push`.~ Upgrading to latest rubygems with `gem update --system` makes `--otp` available to `gem push`.

```
$ gem -v     
3.0.1

$ gem push -h
Usage: gem push GEM [options]

  Options:
    -k, --key KEYNAME                Use the given API key
                                     from ~/.gem/credentials
        --otp CODE                   Digit code for multifactor authentication
        --host HOST                  Push to another gemcutter-compatible host
                                       (e.g. https://rubygems.org)


  Local/Remote Options:
    -p, --[no-]http-proxy [URL]      Use HTTP proxy for remote operations


  Common Options:
    -h, --help                       Get help on this command
    -V, --[no-]verbose               Set the verbose level of output
    -q, --quiet                      Silence command progress meter
        --silent                     Silence RubyGems output
        --config-file FILE           Use this config file instead of default
        --backtrace                  Show stack backtrace on errors
        --debug                      Turn on Ruby debugging
        --norc                       Avoid loading any .gemrc file


  Arguments:
    GEM       built gem to push up

  Summary:
    Push a gem up to the gem server

  Description:
    The push command uploads a gem to the push server (the default is
    https://rubygems.org) and adds it to the index.
    
    The gem can be removed from the index and deleted from the server using the
    yank
    command.  For further discussion see the help for the yank command.
```

colby-swandale:

Note: This PR is a super seed of #7005.

Currently, users that have enabled 2FA in RubyGems.org with both Web+API are not able to upload their gems via `rake release`, due to a bug in `GemHelper#sh_with_status` not handling input/output correctly.

See https://github.com/bundler/bundler/issues/6854

See  https://github.com/bundler/bundler/issues/6854 as this PR is mostly the same.

The major difference here is that i've merged the change into `#sh_with_status` and that i'm using `IO#getc` instead of `IO#gets`. This is because commands could potentially be asking for input on the same line as output, resulting in output that still hasn't been printed.

Example using `passwd`

```
# Using IO#gets
➜ ruby -I lib -rbundler -e 'Bundler::GemHelper.new.send(:sh, "passwd")'
Changing password for colby.
<INPUT>
```

```
# Using IO#getc
$ ruby -I lib -rbundler -e 'Bundler::GemHelper.new.send(:sh, "passwd")'
Changing password for colby.
Old Password:<INPUT>
```

Closes #6854 #7005

### DIFF
--- a/lib/bundler/gem_helper.rb
+++ b/lib/bundler/gem_helper.rb
@@ -100,13 +100,25 @@ module Bundler
   protected
 
     def rubygem_push(path)
-      gem_command = %W[gem push #{path}]
-      gem_command << "--key" << gem_key if gem_key
-      gem_command << "--host" << allowed_push_host if allowed_push_host
+      args = ["gem", "push", path]
+      args += ["--key", gem_key] if gem_key
+      args += ["--host", allowed_push_host] if allowed_push_host
+
       unless allowed_push_host || Bundler.user_home.join(".gem/credentials").file?
         raise "Your rubygems.org credentials aren't set. Run `gem push` to set them."
       end
-      sh(gem_command)
+
+      Bundler.ui.debug(args.join(" "))
+      SharedHelpers.chdir(base) do
+        IO.popen(args, :in => :in) do |io|
+          Thread.new { puts io.gets until io.eof? }.join
+          io.close
+        end
+      end
+
+      code = $?.exitstatus
+      exit code unless code.zero?
+
       Bundler.ui.confirm "Pushed #{name} #{version} to #{gem_push_host}"
     end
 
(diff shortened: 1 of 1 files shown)


colby-swandale:
I'm not particularly sure how to go about testing this, any ideas welcome.

deivid-rodriguez:
I made one suggestion [in the other PR](https://github.com/bundler/bundler/pull/7005#issuecomment-468790151).

In particular, have a look at the specs that call the `bundle` helper with a block

https://github.com/bundler/bundler/blob/48e767da514bb2280b7694a97797eab3904f0442/spec/commands/newgem_spec.rb#L829-L871

deivid-rodriguez:
Superseded by #7199.

PR #7110, opened 2019-04-08
Cleanup old rubies stuff

deivid-rodriguez:

The problem was that there was still some code specific for old rubies in our code base.

My fix is to remove that code.


hsbt:
@bundlerbot r+

PR #7112, opened 2019-04-10
Remove `prefer_gems_rb` setting

deivid-rodriguez:

The problem was that we currently have a feature flag and setting, `prefer_gems_rb`, that currently does very very little, and it's not worth the maintenance cost in my opinion.

My diagnosis was that the only situation where this setting makes a difference at the moment is when a project has both a `Gemfile` _and_ a `gems.rb` file. In that case, if `prefer_gems_rb` is enabled, the warning will tell the user to remove the `Gemfile` because it's being ignored, whereas if not enabled, it will tell the user to remove the `gems.rb` file because it's being ignored.

I think this setting might've made sense when we actually planned to really deprecate Gemfile's in the short term. Since I think we're not planning to deprecate Gemfile's at the moment, I think it's better to remove the setting.

My fix is to remove the setting and keep supporting both names. In the weird case of both types being found in the same project, always prefer `gems.rb` and tell the user to remove the `Gemfile`.

### DIFF
--- a/lib/bundler/feature_flag.rb
+++ b/lib/bundler/feature_flag.rb
@@ -44,7 +44,6 @@ module Bundler
     settings_flag(:only_update_to_newer_versions) { bundler_3_mode? }
     settings_flag(:path_relative_to_cwd) { bundler_3_mode? }
     settings_flag(:plugins) { @bundler_version >= Gem::Version.new("1.14") }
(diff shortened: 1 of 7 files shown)


indirect:
seems like the right approach to me 👍 

@bundlerbot r+

PR #7113, opened 2019-04-10
Make `init_gems_rb` a regular setting

deivid-rodriguez:

The problem was that we have a feature flag to change the file name of the file where gems are defined from `Gemfile` to `gems.rb`, and it is unclear whether we will actually do this, and when.

My diagnosis was that a feature flag is not a good fit here. It's perfectly fine to configure this and allow users to use `gems.rb` by default for their gems, but we don't know when/if we will actually change the default so a plain setting feels better than a feature flag.

My fix is to convert the `init_gems_rb` feature flag into a plain setting.

### DIFF
--- a/lib/bundler/cli/init.rb
+++ b/lib/bundler/cli/init.rb
@@ -41,7 +41,7 @@ module Bundler
   private
 
     def gemfile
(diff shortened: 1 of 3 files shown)


indirect:
great 👍 

@bundlerbot r+

PR #7115, opened 2019-04-11
Allow local install on jruby

deivid-rodriguez:

The problem was that `bin/rake install:local` fails on jruby.

My diagnosis was that the `ronn` gem does not support `jruby`. Since the `install:local` task builds docs on the fly, the task crashes.

My fix is to make the `man:build` task a noop on jruby, so that `bin/rake install:local` can be run on jruby, so jruby users can easily try out bundler's master.

I chose this fix because it sounds like a good workaround.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -199,68 +199,72 @@ task :rubocop do
 end
 
 namespace :man do
-  ronn_dep = development_dependencies.find do |dep|
-    dep.name == "ronn"
-  end
-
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
This affects the development environment only and allows to try out bundler on jruby locally, particularly important since we don't yet have CI for jruby.

@bundlerbot r+

deivid-rodriguez:
@bundlerbot r+

PR #7116, opened 2019-04-11
Refactors.

ryanfox1985:
no problem

My diagnosis was the possibility to reduce useless code and make the ruby code more readable following the Rubocop style guide.

My fix it's just refactors.

### DIFF
--- a/lib/bundler/cli/outdated.rb
+++ b/lib/bundler/cli/outdated.rb
@@ -2,17 +2,21 @@
 
 module Bundler
   class CLI::Outdated
-    attr_reader :options, :gems
+    attr_reader :options, :gems, :options_include_groups, :filter_options_patch, :sources
+    attr_accessor :outdated_gems_by_groups, :outdated_gems_list
 
     def initialize(options, gems)
       @options = options
       @gems = gems
+      @outdated_gems_by_groups = {}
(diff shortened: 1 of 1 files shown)


ryanfox1985:
22 lines less.

colby-swandale:
Can you please fill-out the PR template https://github.com/bundler/bundler/blob/master/.github/PULL_REQUEST_TEMPLATE.md. It really helps us try to understand the intent to your change. You might also want to rework your commit messages a bit better to communicate what they're fixing/changing.

ryanfox1985:
@colby-swandale ok, I rewrote the PR subject.

deivid-rodriguez:
Hei @ryanfox1985! Thanks for contributing! :heart:

As @colby-swandale points out, it's hard to understand the intention of your changes at the moment, even with the new PR description. It seems that you are driven by the number of lines this PR removes, but that's not really a justification for changing anything. Short code is not necessarily better than long code...

Right now this PR is a mixture of style changes and instance variable extractions put together in a single commit. What do you think about getting started with the style changes you want applied, and opening separate PRs enabling the corresponding rubocop rule(s) globally?

ryanfox1985:
ok I understand, I will submit new PRs applying separated changes.

PR #7117, opened 2019-04-12
Remove `error_on_stderr` setting

deivid-rodriguez:
Closes #6730 (supersedes it).

The problem was that this is not a good fit for a feature flag nor a setting since it was just a bug fix.

My diagnosis was that we can remove the setting, specially since the behavior has already changed.

My fix is to remove the setting.


colby-swandale:
This is a bit nit-picky but it would be nice to keep the attribution by either adding these commits to either the original PR. Or copying the original PR's commits into here.

Bundler is open source, so any work/effort on it should be recognized

deivid-rodriguez:
It's not nitpicky, you are absolutely 100% right. The only reason I didn't use the original PR was that it's pointed to 2-0-stable, but I should've reused the commits or add @greysteil as co-author if not reusing them (like in this case).

Thanks so much for pointing this out :heart:. I'm going to instead change the target branch of #7117 and rebase it.

PR #7118, opened 2019-04-12
Refactors to `bundle add`

ryanfox1985:

The problem was no problem just code refactors.

My diagnosis was the possibility to reorganize the code to make clean and more readable.

My fix reorganize the code, good usage of the attributes.

### DIFF
--- a/lib/bundler/cli/add.rb
+++ b/lib/bundler/cli/add.rb
@@ -2,19 +2,21 @@
 
 module Bundler
   class CLI::Add
+    attr_reader :gems, :options
+
     def initialize(options, gems)
       @gems = gems
       @options = options
-      @options[:group] = @options[:group].split(",").map(&:strip) if !@options[:group].nil? && !@options[:group].empty?
+      @options[:group] = options[:group].split(",").map(&:strip) if !options[:group].nil? && !options[:group].empty?
     end
 
     def run
-      raise InvalidOption, "You can not specify `--strict` and `--optimistic` at the same time." if @options[:strict] && @options[:optimistic]
+      raise InvalidOption, "You can not specify `--strict` and `--optimistic` at the same time." if options[:strict] && options[:optimistic]
 
       # raise error when no gems are specified
-      raise InvalidOption, "Please specify gems to add." if @gems.empty?
+      raise InvalidOption, "Please specify gems to add." if gems.empty?
 
-      version = @options[:version].nil? ? nil : @options[:version].split(",").map(&:strip)
+      version = options[:version].nil? ? nil : options[:version].split(",").map(&:strip)
 
       unless version.nil?
         version.each do |v|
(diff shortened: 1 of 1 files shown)


ryanfox1985:
@deivid-rodriguez is it needed to split this PR too?

lib/bundler/cli/add.rb
```diff
@@ -2,34 +2,38 @@
 
 module Bundler
   class CLI::Add
+    attr_reader :gems, :options, :version
```
ryanfox1985: I created the attributes reader, this allow to avoid to call every time the variables with @

ryanfox1985: I added version, multiples usages.

lib/bundler/cli/add.rb
```diff
@@ -2,34 +2,38 @@
...
       @gems = gems
       @options = options
-      @options[:group] = @options[:group].split(",").map(&:strip) if !@options[:group].nil? && !@options[:group].empty?
+      @options[:group] = options[:group].split(",").map(&:strip) if !options[:group].nil? && !options[:group].empty?
+      @version = options[:version].split(",").map(&:strip) unless options[:version].nil?
     end
 
     def run
-      raise InvalidOption, "You can not specify `--strict` and `--optimistic` at the same time." if @options[:strict] && @options[:optimistic]
+      validate_options!
```
ryanfox1985: I moved all validations to `validate_options!` I added the exclamation mark because it can raise an exception.

lib/bundler/cli/add.rb
```diff
@@ -2,34 +2,38 @@
...
 
-      Injector.inject(dependencies,
-        :conservative_versioning => @options[:version].nil?, # Perform conservative versioning only when version is not specified
-        :optimistic => @options[:optimistic],
-        :strict => @options[:strict])
+      # raise error when no gems are specified
+      raise InvalidOption, "Please specify gems to add." if gems.empty?
 
-      Installer.install(Bundler.root, Bundler.definition) unless @options["skip-install"]
+      version.to_a.each do |v|
```
ryanfox1985: I removed the if converting version to an array, it's possible to reach here with a nil value.

deivid-rodriguez:
I would like that, per every comment you have made explaining what you're doing, you extracted that particular change to a separate commit, and explain in the commit message what you explained in the comments. No need for a separate PR in this case, but in my opinion, splitting the commits is nice!

Thanks!

ryanfox1985:
@deivid-rodriguez done.

lib/bundler/cli/add.rb
```diff
@@ -2,34 +2,45 @@
 
 module Bundler
   class CLI::Add
+    attr_reader :gems, :options, :version
+
     def initialize(options, gems)
       @gems = gems
       @options = options
-      @options[:group] = @options[:group].split(",").map(&:strip) if !@options[:group].nil? && !@options[:group].empty?
+      @options[:group] = options[:group].split(",").map(&:strip) if !options[:group].nil? && !options[:group].empty?
```
deivid-rodriguez: I guess since we are at it we can parse this just like we parse `@version` in the following line.

```suggestion
      @options[:group] = options[:group].split(",").map(&:strip) unless options[:group].nil?
```

ryanfox1985: Agree, also I noticed some params are parsing in some commands at the initializer (duplicated code), I think it's a good idea to move to precommand action.

deivid-rodriguez:
@ryanfox1985 This looks good to me. Only request, can you amend the last commit message to describe what the commit is actually doing instead of calling it "deivid suggestion."? Thanks!

ryanfox1985:
@deivid-rodriguez done!

deivid-rodriguez:
Thanks @ryanfox1985!

@bundlerbot r+

PR #7120, opened 2019-04-12
Fix spec calling incorrect helper

deivid-rodriguez:

The problem was the subprocess spawed by one spec was crashing, but that was not making the spec fail.

My diagnosis was that since the spec was checking for a failure status of the subprocess, that was not specific enough to detect the particular bug in this spec (it was calling `clean_system`, removed from bundler 3, instead of `unbundled_system`).

My fix is to restore the previous strategy for these specs, namely, check for a more specific status than 0
or 1.


deivid-rodriguez:
Affects specs only and it's straighforward, so merging.

@bundlerbot r+

PR #7122, opened 2019-04-12
Remove standard output from `with_rubygems`

deivid-rodriguez:
This PR might be very specific to my setup, but it would certainly help me.

The problem was that sometimes I don't find out early about rubocop issues, so I have to push extra commits to fix them, and that means time and CI resources.

My diagnosis was that:

* I run rubocop automatically on save from `vim using [ale](https://github.com/w0rp/ale).
* I set `ENV["RGV"] = master`, so that I always run the latest rubygems.

Because of setting `ENV["RGV"]`, the `with_rubygems` script is run everytime a binstub is loaded (including `rubocop`). This script prints some git information to the standard output. Since ALE expects a very specific output from `rubocop`, the fact that the `with_rubygems` script prints stuff to the standard output means that ALE no longer works, and I don't get to early fix style issues.

My fix is to make the `with_rubygems` script be silent.

I could've removed loading the `with_rubygems` script from the `rubocop` binstub, but I chose this fix because other subcommands in the script already seemed to try to be silent, like the `--quiet` flag passed to `git checkout`.

### DIFF
--- a/bin/with_rubygems
+++ b/bin/with_rubygems
@@ -4,7 +4,7 @@
 require "pathname"
 
 def run(*cmd)
-  return if system(*cmd)
+  return if system(*cmd, :out => IO::NULL)
   raise "Running `#{cmd.join(" ")}` failed"
 end
 
@@ -20,8 +20,6 @@ unless rubygems_path.directory?
     run("git remote update")
     version = "v#{version}" if version =~ /\A\d/
     run("git", "checkout", version, "--quiet")
-    hash = `git rev-parse HEAD`.chomp
-    puts "Checked out rubygems '#{version}' at #{hash}"
   end
 end
 


deivid-rodriguez:
Unless anyone has a very strong opinion here, I'd like to get this in... Otherwise I feel I need to choose between setting `RGV=master` or having a working linter configuration, and I'd like to have both :sweat_smile:.

deivid-rodriguez:
```
$ bin/rubocop
Fetching origin
Checked out rubygems 'master' at b2ec76a447f3670b115c1dd1267869427d968b7b
Inspecting 368 files
...........................C....................................................................................................................................................................................................................................................................................................................................................

Offenses:

spec/bundler/env_spec.rb:57:7: C: Layout/AccessModifierIndentation: Outdent access modifiers like private.
      private
      ^^^^^^^

368 files inspected, 1 offense detected
```

The first two lines of the output is what I'm removing here, so that tools relying on `rubocop`'s output having a specific shape don't get tripped up.

deivid-rodriguez:
Doing this since it's handy for me and there's no strong opposition.

@bundlerbot r+

PR #5114, opened 2016-10-23
Improve and fixed guide style code from outdated cmd

ryanfox1985:
Hi, 

I separate these part from this PR https://github.com/bundler/bundler/pull/5058.

Regards.

PR #7123, opened 2019-04-12
Refactors to `bundle outdated`

ryanfox1985:

Code in `bundle outdated` code be simpler and more readable.

My fix is to refactor the command.

### DIFF
--- a/lib/bundler/cli/outdated.rb
+++ b/lib/bundler/cli/outdated.rb
@@ -2,7 +2,7 @@
 
 module Bundler
   class CLI::Outdated
-    attr_reader :options, :gems, :options_include_groups, :filter_options_patch, :sources
+    attr_reader :options, :gems, :options_include_groups, :filter_options_patch, :sources, :strict
     attr_accessor :outdated_gems_by_groups, :outdated_gems_list
 
     def initialize(options, gems)
@@ -19,6 +19,11 @@ module Bundler
       @options_include_groups = [:group, :groups].any? do |v|
         options.keys.include?(v.to_s)
       end
+
+      # the patch level options imply strict is also true. It wouldn't make
+      # sense otherwise.
+      @strict = options["filter-strict"] ||
+        Bundler::CLI::Common.patch_level_options(options).any?
     end
 
     def run
@@ -47,11 +52,6 @@ module Bundler
         options
       )
 
-      # the patch level options imply strict is also true. It wouldn't make
-      # sense otherwise.
-      strict = options["filter-strict"] ||
-        Bundler::CLI::Common.patch_level_options(options).any?
-
       definition_resolution = proc do
         options[:local] ? definition.resolve_with_cache! : definition.resolve_remotely!
       end
@@ -79,7 +79,7 @@ module Bundler
         next if !gems.empty? && !gems.include?(current_spec.name)
 
         dependency = current_dependencies[current_spec.name]
-        active_spec = retrieve_active_spec(strict, definition, current_spec)
+        active_spec = retrieve_active_spec(definition, current_spec)
 
         next if active_spec.nil?
         next if filter_options_patch.any? &&
@@ -168,7 +168,7 @@ module Bundler
       end
     end
 
-    def retrieve_active_spec(strict, definition, current_spec)
+    def retrieve_active_spec(definition, current_spec)
       return unless current_spec.match_platform(Bundler.local_platform)
 
       if strict
--- a/lib/bundler/cli/outdated.rb
+++ b/lib/bundler/cli/outdated.rb
@@ -105,11 +105,7 @@ module Bundler
         display_nothing_outdated_message
       else
         unless options[:parseable]
-          if options[:pre]
-            Bundler.ui.info "Outdated gems included in the bundle (including pre-releases):"
-          else
-            Bundler.ui.info "Outdated gems included in the bundle:"
-          end
+          Bundler.ui.info(header_outdated_message)
         end
 
         if options_include_groups
@@ -125,11 +121,7 @@ module Bundler
             next if (!options[:groups] && !contains_group) || gems.nil?
 
             unless options[:parseable]
-              if groups
-                Bundler.ui.info "===== #{groups_text("Group", groups)} ====="
-              else
-                Bundler.ui.info "===== Without group ====="
-              end
+              Bundler.ui.info(header_group_message(groups))
             end
 
             print_gems(gems)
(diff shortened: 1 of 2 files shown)


ryanfox1985:
I closed the PR https://github.com/bundler/bundler/pull/7116 and I committed separately to follow better the changes.

deivid-rodriguez:
Hei @ryanfox1985, thanks! Are you planning to follow up with more PRs?

I'm really happy that you want to get involved, but I think I would prefer that you focus on fixing existing issues, instead of just making the code more readable. Making the code more readable is great but it takes time to review and there's always the risk of introducing regressions, so we need to be very careful.

ryanfox1985:
I see, I just want to make clean my old contribution https://github.com/bundler/bundler/pull/5114

deivid-rodriguez:
Aha, I understand, I didn't know this was coming from a previous contribution :+1:.

spec/commands/outdated_spec.rb
```diff
@@ -72,7 +72,7 @@
   end
 
   describe "with --group option" do
-    def test_group_option(group = nil, gems_list_size = 1)
+    before :each do
```
deivid-rodriguez: We're not consistent here, but the preferred style should be `before do` instead of `before :each do`

lib/bundler/cli/outdated.rb
```diff
@@ -2,18 +2,23 @@
 
 module Bundler
   class CLI::Outdated
-    attr_reader :options, :gems
+    attr_reader :options, :gems, :options_include_groups, :filter_options_patch,
+      :sources
```
deivid-rodriguez: Can we keep the readers on a single line?

lib/bundler/cli/outdated.rb
```diff
@@ -215,7 +209,7 @@ def print_gem(current_spec, active_spec, dependency, groups, options_include_gro
         "  * #{spec_outdated_info} in groups \"#{groups}\""
       end
 
-      Bundler.ui.info output_message.rstrip
+      Bundler.ui.info(output_message.rstrip)
```
deivid-rodriguez: Can we keep the same style as previously?

lib/bundler/cli/outdated.rb
```diff
@@ -59,18 +57,15 @@ def run
         definition_resolution.call
       end
 
-      Bundler.ui.info ""
+      Bundler.ui.info("")
```
deivid-rodriguez: Let's keep the old style.

lib/bundler/cli/outdated.rb
```diff
@@ -59,18 +57,15 @@ def run
...
 
       # Loop through the current specs
       gemfile_specs, dependency_specs = current_specs.partition do |spec|
-        current_dependencies.key? spec.name
+        current_dependencies.key?(spec.name)
```
deivid-rodriguez: Let's keep the old style of parentheses everywhery, and focus on the refactoring.

lib/bundler/cli/outdated.rb
```diff
@@ -59,18 +57,15 @@ def run
...
       end
 
-      specs = if options["only-explicit"]
-        gemfile_specs
-      else
-        gemfile_specs + dependency_specs
-      end
+      specs = gemfile_specs
+      specs += dependency_specs unless options["only-explicit"]
```
deivid-rodriguez: I think this change is only stylistic, and does not add much to readability. Can you remove it?

lib/bundler/cli/outdated.rb
```diff
@@ -79,45 +74,41 @@ def run
...
         end
 
-        outdated_gems_list << { :active_spec => active_spec,
-                                :current_spec => current_spec,
-                                :dependency => dependency,
-                                :groups => groups }
+        outdated_gems_list << {
+          :active_spec => active_spec, :current_spec => current_spec,
+          :dependency => dependency, :groups => groups
+        }
```
deivid-rodriguez: Can you keep the previous 4 lines as they were before?

ryanfox1985:
@deivid-rodriguez I did your requested changes.

lib/bundler/cli/outdated.rb
ryanfox1985: not agree, I my opinion the parentheses should be mandatory, this case with one parameter there is no confusion or doubt but when people add more parameters and logic in the parameters can be confused.

spec/commands/outdated_spec.rb
ryanfox1985: I changed all the `before :each do` occurrences from this spec to be consistent.

lib/bundler/cli/outdated.rb
```diff
@@ -97,105 +93,81 @@ def run
...
+      active_specs = definition.find_indexed_specs(current_spec)
+      if !current_spec.version.prerelease? && !options[:pre] && active_specs.size > 1
+        active_specs.delete_if {|b| b.respond_to?(:version) && b.version.prerelease? }
+      end
+      active_specs.last
+    end
+
+    def print_gems(gems_list)
+      gems_list.each do |g|
+        print_gem(g[:current_spec], g[:active_spec], g[:dependency], g[:groups])
```
deivid-rodriguez: I agree with extracting the logic, but I like the previous style, and also the more descriptive variable name (`gem`, instead of `g`). Let's avoid the style changes and only extract the common logic the method, keeping it as:

```ruby
def print_gems(gems_list)
  gems_list.each do |gem|
    print_gem(
      gem[:current_spec],
      gem[:active_spec],
      gem[:dependency],
      gem[:groups]
    )
  end
end
```

lib/bundler/cli/outdated.rb
```diff
@@ -162,17 +153,13 @@ def run
...
-        active_spec = active_specs.last
-      end
+      return definition.find_resolved_spec(current_spec) if strict
 
-      active_spec
+      active_specs = definition.find_indexed_specs(current_spec)
+      if !current_spec.version.prerelease? && !options[:pre] && active_specs.size > 1
+        active_specs.delete_if {|b| b.respond_to?(:version) && b.version.prerelease? }
+      end
+      active_specs.last
```
deivid-rodriguez: Regarding this method, I can see how removing the `active_spec` local makes it simpler, but in general I don't think early returns always make the code better. In this case, I prefer the old version because it gives both branches the same importance.

lib/bundler/cli/outdated.rb
ryanfox1985: what do you think in kwargs usage?

lib/bundler/cli/outdated.rb
ryanfox1985: do you mean rollback of `retrieve_active_spec` method?

deivid-rodriguez: Yeah

lib/bundler/cli/outdated.rb
deivid-rodriguez: You mean here? No, let's keep it as it is. For new code it's good, but I don't want to convert all of our code base to use it.

deivid-rodriguez:
I rebased this PR and tried to minimize the set of changes being made, and clarify the history of commits. This should be ready now.

ryanfox1985:
LGTM

deivid-rodriguez:
This is a unlikely to cause important problems, and I want to have a look at other PRs improving `bundle outdated` so I plan to merge it in a few days.

Reviews appreciated!

deivid-rodriguez:
@bundlerbot r+

PR #7124, opened 2019-04-12
Skip platform warnings in inline mode

### LINKED ISSUE #6822: Platform warnings should be skipped when using bundler/inline
I noticed that when using `bundler/inline`, I get the usual platform warnings for dependencies not meant for my platform:

```
The dependency activerecord-jdbcsqlite3-adapter (~> 51.0) will be unused by any of the platforms Bundler is installing for. Bundler is installing for ruby but the dependency is only for java. To add those platforms to the bundle, run `bundle lock --add-platform java`.
The dependency jruby-openssl (~> 0.10.1) will be unused by any of the platforms Bundler is installing for. Bundler is installing for ruby but the dependency is only for java. To add those platforms to the bundle, run `bundle lock --add-platform java`.
```

However, since in the case of `bundler/inline` there's no lock file, I think these messages should be avoided?

deivid-rodriguez:
Fixes #6822.

The problem was that inline mode prints platform warnings recommending to run `bundle lock` commands.

My diagnosis was that the platform is always missing in the lock file since there's no lock file here.

My fix is to skip platform warnings in this case.

I chose this fix because it's simple.

### DIFF
--- a/lib/bundler/inline.rb
+++ b/lib/bundler/inline.rb
@@ -60,7 +60,7 @@ def gemfile(install = false, options = {}, &gemfile)
 
     Bundler.ui = ui if install
     if install || missing_specs.call
-      Bundler.settings.temporary(:inline => true) do
+      Bundler.settings.temporary(:inline => true, :disable_platform_warnings => true) do
         installer = Bundler::Installer.install(Bundler.root, definition, :system => true)
         installer.post_install_messages.each do |name, message|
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
Yay! Thanks @indirect!

@bundlerbot r=indirect

deivid-rodriguez:
@bundlerbot retry

PR #7125, opened 2019-04-12
Ignore `frozen` setting in inline mode

### LINKED ISSUE #7114: [help wanted] script using bundler/inline in "frozen" capistrano-deployed app
hello.  I need some help understanding if I'm just being dumb about my implementation of something.  I already have a solution that fixes what I'm about to describe but it feels wrong. Apologies for the length.

## Problem
I have an app I just recently converted to deploying with Capistrano.  As such it freezes my bundle with a .bundle/config file setting of `BUNDLE_FROZEN: "true"`.  All well and good for the app.  However I have a standalone ruby script, in `[RAILS_ROOT]/bin`, that gets deployed along with the app. That script uses `bundler/inline` to load gems that are newer than the ones specified in my app Gemfile.

Since I'm in a "frozen" deployment state I get the following when trying to run the bin script from the Rails root directory…

````
/home/app/.rbenv/versions/2.4.5/lib/ruby/site_ruby/2.4.0/bundler/definition.rb:506:in `validate_platforms!': Your bundle only supports platforms [] but your local platforms are ["ruby", "x86_64-linux"], and there's no compatible match between those two lists. (Bundler::ProductionError)
	from /home/app/.rbenv/versions/2.4.5/lib/ruby/site_ruby/2.4.0/bundler/definition.rb:471:in `validate_runtime!'
	from /home/app/.rbenv/versions/2.4.5/lib/ruby/site_ruby/2.4.0/bundler/inline.rb:53:in `gemfile'
	from bin/test:5:in `<main>'
````
_please ignore the use of 2.4.5.  we're working on upgrading._ 😄

Here's an example script that fails…

````ruby
#!/usr/bin/env ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'

  gem 'awesome_print'
end

ap 'foo'
````

to get it to fail add `BUNDLE_FROZEN: "true"` to your `.bundle/config` file.  I can do this even in my local dev environment.

## How I got it work
Like I said at the start.  I figured out how to make the script run just fine.  To get it to work I had to wrap the `gemfile` call like so…

````ruby
#!/usr/bin/env ruby
require 'bundler/inline'
require 'bundler'

Bundler.settings.temporary(frozen: false) do
  gemfile do
    source 'https://rubygems.org'

    gem 'awesome_print'
  end
end

ap 'foo'
````

It feels weird to have to do that.  I want bundler inline to ignore the frozen status from the get-go.  But perhaps the Bundler devs have reasons.

## Plea for help
Does anyone have thoughts on this?  I would prefer to keep the bin script inside my Rails app since it and the app share .env settings.  I can't believe I'm the first person to run a ruby script using inline bundler inside a capistrano-deployed Rails app.  But perhaps I am.  🤔😄 

Thanks in advance.

(e)

deivid-rodriguez:
Fixes #7114.

The problem was that the inline mode wouldn't work if `BUNDLE_FROZEN` is set.

My diagnosis was that since we're in frozen mode, the current platform doesn't get added to the definition, and thus platform validation fails.

My fix is to temporary ignore `BUNDLE_FROZEN` for the inline mode. The inline mode can't really be frozen since it has no lock file.

### DIFF
--- a/lib/bundler/inline.rb
+++ b/lib/bundler/inline.rb
@@ -49,26 +49,28 @@ def gemfile(install = false, options = {}, &gemfile)
   builder = Bundler::Dsl.new
   builder.instance_eval(&gemfile)
 
-  definition = builder.to_definition(nil, true)
-  def definition.lock(*); end
-  definition.validate_runtime!
+  Bundler.settings.temporary(:frozen => false) do
+    definition = builder.to_definition(nil, true)
+    def definition.lock(*); end
+    definition.validate_runtime!
 
-  missing_specs = proc do
-    definition.missing_specs?
-  end
+    missing_specs = proc do
+      definition.missing_specs?
+    end
 
-  Bundler.ui = ui if install
-  if install || missing_specs.call
-    Bundler.settings.temporary(:inline => true) do
-      installer = Bundler::Installer.install(Bundler.root, definition, :system => true)
-      installer.post_install_messages.each do |name, message|
-        Bundler.ui.info "Post-install message from #{name}:\n#{message}"
+    Bundler.ui = ui if install
+    if install || missing_specs.call
+      Bundler.settings.temporary(:inline => true) do
+        installer = Bundler::Installer.install(Bundler.root, definition, :system => true)
+        installer.post_install_messages.each do |name, message|
+          Bundler.ui.info "Post-install message from #{name}:\n#{message}"
+        end
       end
     end
-  end
 
-  runtime = Bundler::Runtime.new(nil, definition)
-  runtime.setup.require
+    runtime = Bundler::Runtime.new(nil, definition)
+    runtime.setup.require
+  end
 ensure
   bundler_module = class << Bundler; self; end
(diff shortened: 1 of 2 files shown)


indirect:
Yeah, this seems legit. It strikes me as weird/unsupported to be running an inline script while BUNDLE_FROZEN is set, but if we can make it work, why not. :) r+ from me.

deivid-rodriguez:
Thanks!

@bundlerbot r=indirect

silent-e:
re: @indirect 
> Yeah, this seems legit. It strikes me as weird/unsupported to be running an inline script while BUNDLE_FROZEN is set, but if we can make it work, why not. :) r+ from me.

I was the one that submitted the original issue bringing this up.  The use case I have is a Rails 4 app but I have a standalone script in the project that is using inline to load a 5.x version of ActiveSupport.  After deploy capistrano froze the bundle and the standalone fails without this fix.

FYI

(e)

PR #7128, opened 2019-04-14
Backport to workaround from ruby core.

hsbt:

The current master branch couldn't invoke with the ruby core repository.

1. We need to add explicitly declare `rspec` in spec_helper.rb
2. Some examples were failed on ruby core repository.

1. simply added.
2. update the `ruby_repo` labels for skipping to run.


deivid-rodriguez:
@bundlerbot r+

Issue #7096, opened 2019-04-03
Bundle should not require that the user grant write permissions to paths in /var/folders/


calebstdenis:
When I run the following,

```
bundle gem my-gem
cd my-gem
bundle
```

I get the following error:
```
There was an error while trying to write to `/var/folders/04/42312kk53v37n0w58p857gfw0hh8c4/T/bundler-compact-index-20190403-88495-1hv2d9b/versions`. It is likely that you need
to grant write permissions for that path.
```

Why do I need to grant write permissions to a path that changes every time I `bundle`?

calebstdenis:
Apologies in advance for any misunderstanding from my end. I'm trying to be a Grand Wizard but am still a mere elemental mage.

deivid-rodriguez:
What's the output of `bundle env`? It sounds like the folder configured for `bundler` to store its cache is not writable by the user you're running bundler with.

calebstdenis:
@deivid-rodriguez 

I am impressed that the output is in Markdown. Good call! 🙂
## Environment

```
Bundler       1.17.2
  Platforms   ruby, x86_64-darwin-18
Ruby          2.6.2p47 (2019-03-13 revision 67232) [x86_64-darwin18]
  Full Path   /Users/cstdenis/.rbenv/versions/2.6.2/bin/ruby
  Config Dir  /Users/cstdenis/.rbenv/versions/2.6.2/etc
RubyGems      3.0.3
  Gem Home    /Users/cstdenis/.rbenv/versions/2.6.2/lib/ruby/gems/2.6.0
  Gem Path    /Users/cstdenis/.gem/ruby/2.6.0:/Users/cstdenis/.rbenv/versions/2.6.2/lib/ruby/gems/2.6.0
  User Path   /Users/cstdenis/.gem/ruby/2.6.0
  Bin Dir     /Users/cstdenis/.rbenv/versions/2.6.2/bin
Tools         
  Git         2.17.2 (Apple Git-113)
  RVM         not installed
  rbenv       rbenv 1.1.2
  chruby      not installed
```

## Bundler Build Metadata

```
Built At          2018-12-19
Git SHA           3fc4de72b
Released Version  false
```

## Bundler settings

```
gem.test
  Set for the current user (/Users/cstdenis/.bundle/config): false
gem.mit
  Set for the current user (/Users/cstdenis/.bundle/config): false
gem.coc
  Set for the current user (/Users/cstdenis/.bundle/config): false
```

## Gemfile

[...]

PR #7129, opened 2019-04-14
Bundle env improvements

deivid-rodriguez:
This came from debugging #7096.

The problem was that `bundle env` is great, but could be more useful and accurate.

My fix is to start improving it, by dumping more paths, and more accurate information, and getting the paths tested.

I have a few more improvements in mind, but I want to keep my PRs small.


indirect:
I like adding tests for this. 😄 Can you explain more about what you mean by "accurate"? I vaguely recall that there was some time when `GEM_PATH` could be set to something different than `Gem.path`, probably because `Gem.path` got set and then Bundler changed `GEM_PATH` and an error happened before `Gem.path` was also updated to match? Or something like that. Maybe in the future we can also print `GEM_PATH/GEM_HOME` if they don't match `Gem.path/Gem.home`?

deivid-rodriguez:
Once rubygems/bundler loads its paths from `GEM_PATH` / `GEM_HOME`, those paths are always used internally, so it seems to me that those values are what we want here. But I definitely agree that we should also display `GEM_PATH` / `GEM_HOME` if they are different from the value originally set. I'm going to add that :+1:

deivid-rodriguez:
Actually... How can these be different? I guess if a plugin is somehow messing with the environment variables after bundler is loaded? But I'm not sure whether `bundle env` loads plugins?

All in all, "accurate" is not a good word, this should always be the same so it is just simplifying the existing code and making it simpler.

indirect:
I think I am vaguely remembering some wild debugging situation where the end vars did not match the method results, but maybe we can wait to improvement that until we see it crop up again in the wild in the future :)

deivid-rodriguez:
Awesome, so this is good to go? :)

indirect:
Yeah, go for it 👍🏻

deivid-rodriguez:
:+1:

@bundlerbot r+

PR #7136, opened 2019-04-23
Bump CI rubies to latest patch levels

deivid-rodriguez:

The problem was that new rubies have been released, and we're not yet testing against them,

My fix is to update our CI matrix.


deivid-rodriguez:
@bundlerbot r+

PR #7137, opened 2019-04-23
Gemspec require relative

deivid-rodriguez:

This is a follow up to #7100. Gemspec is another place where `require_relative` sounds like a win.

My fix changes bundler's gemspec to use `require_relative`, and extracts the relevant change from #4397 to also change the generated gemspec from `bundle gem`.

### DIFF
--- a/bundler.gemspec
+++ b/bundler.gemspec
@@ -1,10 +1,10 @@
 # frozen_string_literal: true
 
 begin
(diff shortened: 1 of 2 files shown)


duckinator:
`require_relative` is definitely a huge win here. :+1:

deivid-rodriguez:
Going in!

@bundlerbot r+

PR #7138, opened 2019-04-23
Remove unnecessary `BUNDLER_SPEC_RUN` env variable

deivid-rodriguez:

The problem was that our library code contains logic that's only necessary for bundler's specs. I think library code should not contain this kind of code, since it affects end users.

My fix is to remove the logic, and instead ensure the specs satisfy what the code wanted to ensure (that specs pass even when there's a Gemfile further up in the directory hierarchy outside of the bundler's development copy).


deivid-rodriguez:
@indirect What do you think about this? I think it's better to not pollute library code with test specific code.

PR #7139, opened 2019-04-23
Tweak rubygems integration

deivid-rodriguez:

The problem was that the rubygems integration file could be more consistent.

My fix is to lazily require `monitor`, just like it's done with the other requires in the file, and to define the methods that check the rubygems version a single time, just like the rest of the methods in the file.

### DIFF
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -304,27 +304,17 @@ module Bundler
       end
     end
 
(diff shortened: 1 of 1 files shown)


duckinator:
The removal of `if/else` with nested method definitions is always nice. This looks good. :slightly_smiling_face:

deivid-rodriguez:
Thanks! I'll do this!

@bundlerbot r+

PR #7141, opened 2019-04-24
Remove coverage tracking for the time being

deivid-rodriguez:

The problem was that current coverage numbers are incorrect since our specs rely on checking the results of bundler subprocesses and coverage is not tracked there. We don't know the actual numbers but coverage should be much higher than the [73% codeclimate reports today](https://codeclimate.com/github/bundler/bundler/test_coverage).

I tried to get this fixed in #6980, but couldn't get it working.

My fix is to remove coverage tracking for now. This should make merging CIs with rubygems easier, and removes the misleading coverage numbers.

Coverage has helped in the past when the test suite was "unitary test based" as explained by @indirect in #6980, but now it's not useful. Let's revisit it in the future.


indirect:
@bundlerbot r+

PR #7145, opened 2019-04-27
More compatibility cleanup

deivid-rodriguez:

The problem was we still have unnecessary code for compatiblity with rubygems versions we no longer support.

My diagnosis was that we can remove this code.

My fix is to remove it.


hsbt:
@bundlerbot r+

PR #7147, opened 2019-04-28
Fix specs with rubygems master installed globally

deivid-rodriguez:

The problem was if you have rubygems's master installed globally, bundler specs no longer pass.

My diagnosis was that having rubygems master installed, installs bundler's master as a default gem. That actually makes bundler behave better. In particular, `bundle show` and `bundle licenses` start properly handling the `bundler` gem itself.

My fix is to change bundler's code to fix this specs in every situation, and not only when the bundler version being tested is also installed as a default gem.

NOTE: I also included some changes to unify `bundle show` deprecation specs in the file that tests all deprecations.

### DIFF
--- a/lib/bundler/source/metadata.rb
+++ b/lib/bundler/source/metadata.rb
@@ -13,6 +13,7 @@ module Bundler
           idx << Gem::Specification.new do |s|
             s.name     = "bundler"
             s.version  = VERSION
+            s.license  = "MIT"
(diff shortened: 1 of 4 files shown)


deivid-rodriguez:
This fixes several realworld issues, like `bundle` not being available to show certain informations about itself, like website or license. So I'm merging it!

@bundlerbot r+

PR #7148, opened 2019-04-28
Use `Gem.add_to_load_path` if available

deivid-rodriguez:
This is a follow up to https://github.com/rubygems/rubygems/pull/2749.

My fix is to unify all LOAD_PATH manipulation in a single place, that uses rubygems logic if available.

### DIFF
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -39,10 +39,6 @@ module Bundler
       Gem::Command.build_args = args
     end
 
-    def load_path_insert_index
-      Gem.load_path_insert_index
-    end
-
     def loaded_specs(name)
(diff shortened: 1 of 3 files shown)


lib/bundler/rubygems_integration.rb
```diff
@@ -39,14 +39,22 @@ def build_args=(args)
...
-    def load_path_insert_index
-      Gem.load_path_insert_index
-    end
-
     def loaded_specs(name)
       Gem.loaded_specs[name]
     end
 
+    def add_to_load_path(paths)
+      return Gem.add_to_load_path(*paths) if Gem.respond_to?(:add_to_load_path)
```
hsbt: 👍

hsbt:
@bundlerbot r+

PR #7150, opened 2019-04-28
Proper group skip message for bundle update

deivid-rodriguez:

The problem was that when `bundle update` skipped groups, it shows a message about "gems not being installed". The message could be better and say "gems not being updated" instead, since that's what the user was trying to do.

My diagnosis was that helper method was reused by the update and the install commands, but its message was specific to install.

My fix is to pass a parameter to the method with the calling command, so that the message is customized properly.

I chose this fix because it's easy.


deivid-rodriguez:
Seems straightforward, I'll merge it!

@bundlerbot r+

PR #7151, opened 2019-04-29
Bump rubocop to 0.68.0

deivid-rodriguez:

The problem was that our rubocop version is getting outdated. Also, I added [a PR to rubocop](https://github.com/rubocop-hq/rubocop/pull/6935) so that we can remove the exclusions in our configuration, and it has now been released.

My fix is to update rubocop to the latest version, update the configuration, and fix new offenses.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,3 +1,5 @@
+require: rubocop-performance
+
 AllCops:
   DisabledByDefault: true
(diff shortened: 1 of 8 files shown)


hsbt:
👍

hsbt:
@bundlerbot r+

PR #7143, opened 2019-04-24
Fix to update conservatively for gemspec update

mtsmfm:
Fix https://github.com/bundler/bundler/issues/6967

### What was the end-user problem that led to this PR?

https://github.com/bundler/bundler/issues/6967

### What was your diagnosis of the problem?

My diagnosis was https://github.com/bundler/bundler/blob/ec8f98574a90de1e87a50e00141180d761f161dc/lib/bundler/definition.rb#L806 is loose

### What is your fix for the problem, implemented in this PR?

My fix makes it strict

### Why did you choose this fix out of the possible options?

I think I don't have any other options

PR #7153, opened 2019-05-01
Rename local variables for readability

mtsmfm:
Extracted from https://github.com/bundler/bundler/pull/7143#discussion_r279187838

> Also, I would split the variable rename, which is great because it makes the code more readable, to a separate commit

No end-user is affected because it's just local var name.

As a bundler contributor, I think it's much easier to understand if we rename them.
When I was investigating https://github.com/bundler/bundler/pull/7143, it's hard to understand what `other` stuff and `deps2` is.


deivid-rodriguez:
Thanks!

@bundlerbot r+

PR #7156, opened 2019-05-03
Improve fileutils require

deivid-rodriguez:

The problem was there's [upstream changes in fileutils](https://github.com/ruby/fileutils/pull/35).

My fix is to port the upstream changes, which consist of using `require_relative` everywhere.

### DIFF
--- a/lib/bundler/vendor/fileutils/lib/fileutils.rb
+++ b/lib/bundler/vendor/fileutils/lib/fileutils.rb
@@ -6,7 +6,7 @@ rescue LoadError
   # for make mjit-headers
 end
 
-require "bundler/vendor/fileutils/lib/fileutils/version"
+require_relative "fileutils/version"
 
 #
 # = fileutils.rb


deivid-rodriguez:
So this broke on ruby 2.3, just like the conditional claimed.

I thought this would be compatible judging by the [`required_rubygems_version` restriction in the `fileutils` version we're using](https://github.com/ruby/fileutils/blob/400574d80d1d5363816574a88bbebf3b38a3ac86/fileutils.gemspec#L1).

The restriction was changed here: https://github.com/ruby/fileutils/pull/19. At first I thought the motivation was to at least allow installation on `jruby`, but I think the currently supported jruby series (9.2) are ruby 2.5+ compatible, so I'm not sure why the restriction was relaxed if fileutils doesn't really work on ruby 2.3. @hsbt?

Anyways, I'll drop the related commit since the conditional is indeed necessary.

deivid-rodriguez:
This is part of #7062, which has already been approved, but since I had already opened this one, I'll merge this and rebase #7062 on top.

@bundlerbot r+

PR #7015, opened 2019-03-04
Expand comparison path for Windows compat

janpio:
### What was the end-user problem that led to this PR?

A specific test (#6890) was failing...

### What was your diagnosis of the problem?

... because the comparison was missing the drive letter on Windows. 

### What is your fix for the problem, implemented in this PR?

I added expansion of the path that was already present in the implementation but not the test.

### Why did you choose this fix out of the possible options?

Makes the test pass and looks right.

---

closes #6890 

fixes 1 failing test on Windows

PR #7157, opened 2019-05-03
Improve azure configuration

deivid-rodriguez:

The problem was that failures in Azure are not logged, so it hard to improve the situation.

My fix is to configure Azure to run until completion without any timeouts, and remove intercepting the global timeout (which we did to make sure build is still green when it times out), because it's no longer necessary. Now we should be able to see the failures on every PR.

### DIFF
--- a/.azure-pipelines/steps.yml
+++ b/.azure-pipelines/steps.yml
@@ -28,7 +28,7 @@ steps:
   displayName: 'gem install rspec_junit_formatter'
 
 - script: |
-    timeout 50m bash -c 'ruby -r rspec_junit_formatter bin/rspec --format progress --format RspecJunitFormatter -o rspec/bundler-junit-results.xml' || exit 0
+    ruby -r rspec_junit_formatter bin/rspec --format progress --format RspecJunitFormatter -o rspec/bundler-junit-results.xml || exit 0
   displayName: 'ruby bin/rspec'
 
 - task: PublishTestResults@2
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -4,3 +4,4 @@ jobs:
     vmImage: 'vs2017-win2016'
   steps:
   - template: .azure-pipelines/steps.yml
+  timeoutInMinutes: 0


deivid-rodriguez:
This PR shows that many failures on Windows are due to improper handling of the `file://` protocol. I had the idea of copying https://github.com/sporkmonger/addressable/blob/5894c95a7768435cb46d1355954611dbd194832e/lib/addressable/uri.rb#L235-L301.

/cc Windows contributors @janpio @MSP-Greg

MSP-Greg:
@deivid-rodriguez

1. Set CI=true in azure - LGTM

2. Quite a while ago, I looked at bundler spec & windows.  I recall that there were a lot of instances where a windows path with a drive reference was added to something else, not just related to the `file://` protocol.  Somewhere there should probably a 'global spec method for converting a windows path starting with `C:\` or `C:/` to `/c/`...

4. Re Azure Pipelines, I haven't checked recently to see what's available in their standard configurations.  I recall that there were several commands in the test suite that were not available in a windows console unless something like MSYS2 is installed.  I've got code to install it in Pipelines somewhere...

janpio:
> The problem was that failures in Azure are not logged, so it hard to improve the situation.

Strange, something changed here. Compare this old result:
https://github.com/bundler/bundler/pull/7015/checks
to this newer one: 
https://github.com/bundler/bundler/runs/117572561

Unfortunately the old run is too old and the build data is not available on Azure Pipelines directly any more, just the check on GitHub.

---

Back when I created the PR I chose to not limit to 25 failures as it is easier to find fixable ones (and see which happen most often) by looking at all test failures, instead of just 25 (which also required the 50 minute limit though).

deivid-rodriguez:
That's interesting... I guess it was not timing out at that point? Or maybe it was timing out, but at least the failures so far were being correctly dumped before timing out? :man_shrugging:

janpio:
Oh right, it is hitting the 50 minute limit now.
Here is a build from not too long ago that just made it below 50: https://dev.azure.com/bundler/bundler/_build/results?buildId=1142

I saw the "total number of failures" as an interesting enough information to let the build run for 50 minutes (plus important to actually suggest what has the most impact when fixed). Form that standpoint, maybe it might be worth to increase the time limit to something that still works.

On the other hand, I don't have the time to contribute here right now, so this is just an random opinion.

deivid-rodriguez:
I fully agree with you. The problem is that most times the build reaches the maximum time for an Azure job and we don't get any information at all, only dots and F's, but nothing else.

If there's a way to increase the maximum time an Azure job can run to, for example, 2 hours, then I'm all for it, but if not, I think it's better to get some useful information every time, than no information at all most times.

.azure-pipelines/steps.yml
```diff
@@ -28,7 +28,7 @@ steps:
   displayName: 'gem install rspec_junit_formatter'
 
 - script: |
-    timeout 50m bash -c 'ruby -r rspec_junit_formatter bin/rspec --format progress --format RspecJunitFormatter -o rspec/bundler-junit-results.xml' || exit 0
```
janpio: `timeout 120m bash -c ...` maybe?

deivid-rodriguez: When I wrapped this with `timeout`, it was to intercept Azure's global timeout of 1h, and being able to always produce a green build. But maybe the global timeout is higher now? I'll give it a try.

janpio: Oh, 1h limit? Meh.

Did it work? Can't see because of the force push :/

If not, the best solution is probably to just deal with it and fix those failures :D

deivid-rodriguez: I didn't try, actually. All I did was increasing the allowed failures to 250 on windows and it took 40 minutes to reach to that number of failures, so it didn't time out.

deivid-rodriguez: Found the proper configuration, now all Azure jobs will run until the end! :+1:

janpio:
`  timeoutInMinutes: 0`, interesting - and it even works (unless there is a global cutoff at 60 minutes).

deivid-rodriguez:
It timed out though... :disappointed:. Until we figure out why the build is getting stalled, I prefer to keep the current measures to have the check green.

janpio:
Oh right, but only for 2.5. Did this break it maybe?

I'll see if I can manage to find a few hours to get rid of a few hundred of those test failures ;)

deivid-rodriguez:
That's a good point, I'll revert the 2.5 upgrade and see what happens.

janpio:
That looks good to me.

deivid-rodriguez:
@bundlerbot r+

deivid-rodriguez:
@bundlerbot r-

I'm cancelling because the PR description no longer matches what's landing, so the commit message will be so confusing. I'll update the PR description and then reapprove.

deivid-rodriguez:
@bundlerbot r+

PR #7160, opened 2019-05-05
Bump Rubocop version

natesholland:

There was not a problem, just some housekeeping.

I wanted to see if bumping the latest Rubocop version would bring in any new fixes or updates. It did not but I figured it didn't hurt to bump the version in case new PRs are affected by the new cops.

I bumped the patch version of RubCop by 1.

I chose this fix because it was the only logical thing to do.


deivid-rodriguez:
Thank you!

@bundlerbot r+

PR #7161, opened 2019-05-06
Fix file URIs on Windows

deivid-rodriguez:

The problem was that file:// URIs were not properly handled on Windows, and those are heavily used by our specs.

My diagnosis was that Windows paths do not start with a slash, so we need to add an extra slash to
separate the host from the path in file:// urls. Otherwise "D:" is parsed as the host segment in the URI.

As a consequence of that change, we also need to strip that leading character when using the path segment for those URIs, so that it's a valid file path on Windows.

This change reduces Windows CI spec failures from 429 to 355.

In order to apply this fix for all file URI's in tests, I extracted a helper method and searched and replaced all file:// ocurrences with the method. The diff is big because of that, but I think it works fine.

### DIFF
--- a/lib/bundler/fetcher.rb
+++ b/lib/bundler/fetcher.rb
@@ -99,7 +99,7 @@ module Bundler
 
       uri = URI.parse("#{remote_uri}#{Gem::MARSHAL_SPEC_DIR}#{spec_file_name}.rz")
       if uri.scheme == "file"
-        path = Gem.win_platform? ? uri.path[1..-1] : uri.path
+        path = Bundler.rubygems.correct_for_windows_path(uri.path)
         Bundler.load_marshal Bundler.rubygems.inflate(Gem.read_binary(path))
       elsif cached_spec_path = gemspec_cached_path(spec_file_name)
         Bundler.load_gemspec(cached_spec_path)
--- a/lib/bundler/fetcher.rb
+++ b/lib/bundler/fetcher.rb
@@ -99,7 +99,8 @@ module Bundler
 
       uri = URI.parse("#{remote_uri}#{Gem::MARSHAL_SPEC_DIR}#{spec_file_name}.rz")
       if uri.scheme == "file"
-        Bundler.load_marshal Bundler.rubygems.inflate(Gem.read_binary(uri.path))
+        path = Gem.win_platform? ? uri.path[1..-1] : uri.path
+        Bundler.load_marshal Bundler.rubygems.inflate(Gem.read_binary(path))
       elsif cached_spec_path = gemspec_cached_path(spec_file_name)
         Bundler.load_gemspec(cached_spec_path)
       else
--- a/lib/bundler/fetcher/index.rb
+++ b/lib/bundler/fetcher/index.rb
@@ -30,7 +30,7 @@ module Bundler
 
         uri = URI.parse("#{remote_uri}#{Gem::MARSHAL_SPEC_DIR}#{spec_file_name}.rz")
         if uri.scheme == "file"
-          path = Gem.win_platform? ? uri.path[1..-1] : uri.path
+          path = Bundler.rubygems.correct_for_windows_path(uri.path)
           Bundler.load_marshal Bundler.rubygems.inflate(Gem.read_binary(path))
         elsif cached_spec_path = gemspec_cached_path(spec_file_name)
           Bundler.load_gemspec(cached_spec_path)
--- a/lib/bundler/fetcher/index.rb
+++ b/lib/bundler/fetcher/index.rb
@@ -30,7 +30,8 @@ module Bundler
 
         uri = URI.parse("#{remote_uri}#{Gem::MARSHAL_SPEC_DIR}#{spec_file_name}.rz")
         if uri.scheme == "file"
-          Bundler.load_marshal Bundler.rubygems.inflate(Gem.read_binary(uri.path))
+          path = Gem.win_platform? ? uri.path[1..-1] : uri.path
+          Bundler.load_marshal Bundler.rubygems.inflate(Gem.read_binary(path))
         elsif cached_spec_path = gemspec_cached_path(spec_file_name)
           Bundler.load_gemspec(cached_spec_path)
         else
--- a/lib/bundler/source/rubygems/remote.rb
+++ b/lib/bundler/source/rubygems/remote.rb
@@ -25,8 +25,7 @@ module Bundler
 
             cache_uri = original_uri || uri
 
-            # URI::File of Ruby 2.6 returns empty string when given "file://".
-            host = defined?(URI::File) && cache_uri.is_a?(URI::File) ? nil : cache_uri.host
+            host = cache_uri.to_s.start_with?("file://") ? nil : cache_uri.host
 
             uri_parts = [host, cache_uri.user, cache_uri.port, cache_uri.path]
             uri_digest = SharedHelpers.digest(:MD5).hexdigest(uri_parts.compact.join("."))
--- a/spec/install/gemfile/sources_spec.rb
+++ b/spec/install/gemfile/sources_spec.rb
@@ -20,8 +20,8 @@ RSpec.describe "bundle install with gems on multiple sources" do
 
       before do
         gemfile <<-G
-          source "file://localhost#{gem_repo3}"
-          source "file://localhost#{gem_repo1}"
+          source "#{file_uri_for(gem_repo3)}"
+          source "#{file_uri_for(gem_repo1)}"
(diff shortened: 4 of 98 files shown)


lib/bundler/fetcher/index.rb
```diff
@@ -30,7 +30,8 @@ def fetch_spec(spec)
 
         uri = URI.parse("#{remote_uri}#{Gem::MARSHAL_SPEC_DIR}#{spec_file_name}.rz")
         if uri.scheme == "file"
-          Bundler.load_marshal Bundler.rubygems.inflate(Gem.read_binary(uri.path))
+          path = Gem.win_platform? ? uri.path[1..-1] : uri.path
```
ryanfox1985: duplicated code, is it possible to avoid these duplication?

deivid-rodriguez: Maybe, I'm not sure why this code is currently duplicated, but refactoring that is out of the scope of this PR.

janpio: I think this comment referred specifically to the code `path = Gem.win_platform? ? uri.path[1..-1] : uri.path` you added - but as it seems the whole `fetch_spec` is duplicated this is indeed out of scope to extract this, considering that `fetch_spec` itself will probably be dedpulicated sooner or later (hopefully).

lib/bundler/source/rubygems/remote.rb
```diff
@@ -25,8 +25,7 @@ def cache_slug
 
             cache_uri = original_uri || uri
 
-            # URI::File of Ruby 2.6 returns empty string when given "file://".
-            host = defined?(URI::File) && cache_uri.is_a?(URI::File) ? nil : cache_uri.host
+            host = cache_uri.to_s.start_with?("file://") ? nil : cache_uri.host
```
janpio: What is this change? Just a different way to achieve the same thing?

lib/bundler/fetcher/index.rb
deivid-rodriguez: Correct :+1:

spec/install/gemfile/sources_spec.rb
```diff
@@ -588,26 +569,6 @@
           gem "rack-obama"
         G
 
-        # It creates this lockfile.
```
janpio: unrelated change?

spec/install/gemfile/sources_spec.rb
```diff
@@ -471,58 +471,39 @@
...
-        source 'file://#{gem_repo1}'
+        source '#{file_uri_for(gem_repo1)}'
         gem 'rack'
-        gem 'foo', '~> 0.1', :source => 'file://#{gem_repo4}'
-        gem 'bar', '~> 0.1', :source => 'file://#{gem_repo4}'
+        gem 'foo', '~> 0.1', :source => '#{file_uri_for(gem_repo4)}'
+        gem 'bar', '~> 0.1', :source => '#{file_uri_for(gem_repo4)}'
       G
 
-      # 3. this lockfile.
```
janpio: unrelated change?

janpio:
- Sometimes the exisiting code uses `file:`, sometimes `file://` - is this important?
- Is there something present to avoid "plain" file URLs being used again? Or would this be caught by a new test failure on Windows?

lib/bundler/source/rubygems/remote.rb
deivid-rodriguez: Very good question.

My idea is to fix the issue that when `file://` URI is used with "localhost" (`file://localhost/path` instead of `file:///path`), then `localhost` would be prepended to the cache path, breaking [this spec](https://travis-ci.org/bundler/bundler/jobs/528808519).

Unfortunately I still need to use "localhost-type" file URIs for older rubies, because otherwise I can't seem to get a consistent behavior across versions (I guess because of bugs fixed in recent rubies):

https://github.com/bundler/bundler/blob/cdcde3e3e5d064dbcad7d20570de1732ba68ef9b/spec/support/path.rb#L69

The previous code use `file:/`style URI (single colon) to bypass this and get the spec passing:

https://github.com/bundler/bundler/blob/3d94636113db3f373be05514b7ae982db26b42e2/spec/install/global_cache_spec.rb#L197

But those don't work well on Windows.

The change makes sense to me since it makes the behavior consistent across ruby versions: we never include the host for `file://` URIs.

spec/install/gemfile/sources_spec.rb
deivid-rodriguez: Yeah, I can extract this changes to a separate PR. :+1:

deivid-rodriguez:
> Sometimes the exisiting code uses file:, sometimes file:// - is this important?

The `file://` protocol specification is quite complicated, but in principle both are correct variants. I think the variability used in specs is because the standard library behaves quite differently accross rubies & platforms, I think most likely because of bugs fixed in recent versions. See https://github.com/bundler/bundler/pull/7161#discussion_r282456588.

> Is there something present to avoid "plain" file URLs being used again? Or would this be caught by a new test failure on Windows?

Can you clarify, I didn't understand what do you mean.

spec/install/gemfile/sources_spec.rb
deivid-rodriguez: Extracted to #7167. Will rebase this PR once that one is merged, so that these changes disappear from the diff.

deivid-rodriguez:
As I understand the this, both `file:` and `file://` variants are correct. Then, if there's no host, we get a third slash meaning "absolute path starts". On Linux, this is the implicit behavior because absolute paths actually start with a slash, but on Windows it has to be added explicitly. Essentially, `file:///` URIs are correct, but the third slash is not part of the protocol.

janpio:
> > Is there something present to avoid "plain" file URLs being used again? Or would this be caught by a new test failure on Windows?
> 
> Can you clarify, I didn't understand what do you mean.

I effectively answered this myself: If someone introduces a spec change that uses `file://` again without the new helper method, this spec will just fail on Windows CI. Correct?

deivid-rodriguez:
> I effectively answered this myself: If someone introduces a spec change that uses file:// again without the new helper method, this spec will just fail on Windows CI. Correct?

Ah, I see. Yes, I'm pretty sure it would fail on Windows :+1:.

deivid-rodriguez:
PR was rebased and now it doesn't have unrelated changes.

As a side note, I think this can probably be fixed in ruby-core. In my opinion, `ruby -e 'puts URI.parse("file:///c:/path/to/file").path'` should print `c:/path/to/file` instead of `/c:/path/to/file` as it does not.

ESA-Security-Group:
Check

lib/bundler/fetcher.rb
```diff
@@ -99,7 +99,8 @@ def fetch_spec(spec)
 
       uri = URI.parse("#{remote_uri}#{Gem::MARSHAL_SPEC_DIR}#{spec_file_name}.rz")
       if uri.scheme == "file"
-        Bundler.load_marshal Bundler.rubygems.inflate(Gem.read_binary(uri.path))
+        path = Gem.win_platform? ? uri.path[1..-1] : uri.path
```
deivid-rodriguez: So, I wanted to make this logic a bit more robust (not always assume windows parsed paths start with `/`), and I found that this logic is already implemented in rubygems:

https://github.com/rubygems/rubygems/blob/d470f1787a94f751c907e8299ec44f61b7689432/lib/rubygems/remote_fetcher.rb#L320-L327

So I guess I can reuse that logic here?

deivid-rodriguez: Opened rubygems/rubygems#2767 to make this easier.

deivid-rodriguez: I included using the linked PR here :+1:

deivid-rodriguez:
ping @hsbt since you assigned this to yourself :)

deivid-rodriguez:
#7215 [failed](https://travis-ci.org/bundler/bundler/jobs/547819071) because of varying behaviour of `file://` URL parsing on different rubies. That should also be fixed by this PR.

deivid-rodriguez:
Thanks @hsbt!

@bundlerbot r=hsbt

ESA-Security-Group:
😂😂

Sincerely - 

Higher Commander - DPO
U.S AirForce - Armed Forces in Europe 
Unit: 527th - Space ࿇ Aggressor ࿇ Squadron

Ambjoern Skoeld

DPO16G4
Armed Forces of Europe - “EA”
APO/DPO Air/Army
 


19 juni 2019 kl. 18:38 skrev David Rodríguez <notifications@github.com>:

ping @hsbt since you assigned this to yourself :)

PR #7163, opened 2019-05-07
Remove `global_path_appends_ruby_scope` setting

deivid-rodriguez:

The problem was we're artificially holding shipping bug fixes like the fact the ruby scope is not included in the install path when `BUNDLE_PATH` is used as an environment variable or global setting.

My diagnosis was that we can remove the setting, and that it's fine that affected users rerun `bundle install` to get their installation paths fixed.

My fix is to remove the setting.


indirect:
Hopefully nothing is depending on the buggy behavior here 😅 

@bundlerbot r+

deivid-rodriguez:
Yeah, me too :smiley:. Thanks @indirect!

PR #7166, opened 2019-05-07
Better test specific file searching

deivid-rodriguez:

The problem was that in #7138 I removed some test specific code from lib, but it's not quite working.

My diagnosis was that the mocking I used to fix the issue of bundler looking for files outside of bundler's clone doesn't work inside subprocesses, which many of our specs use. That means some specs were currently failing if you have Gemfile's on `.bundle` folders up in the directory hierarchy ascending from your bundler's clone

My fix is revert to the previous approach of using a test specific environment variable to prevent search up higher than the test directory when running our specs.

I initially tried to monkeypatch the gemfile finding helpers only inside specs, but that caused load order issues (I monkeypatched the helpers insde `support/hax.rb`, that caused the monkeypatched lib code to be loaded _before_ the rest of `bundler`, which caused issues in some specs).

Overall, the old approach is the only one that works without issues, so I'm reverting to it, at least for now.

### DIFF
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -253,6 +253,19 @@ module Bundler
       current  = File.expand_path(SharedHelpers.pwd).untaint
 
       until !File.directory?(current) || current == previous
+        if ENV["BUNDLE_SPEC_RUN"]
+          # avoid stepping above the tmp directory when testing
+          gemspec = if ENV["BUNDLE_RUBY"] && ENV["BUNDLE_GEM"]
+            # for Ruby Core
+            "lib/bundler/bundler.gemspec"
+          else
+            "bundler.gemspec"
+          end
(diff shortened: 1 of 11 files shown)


deivid-rodriguez:
This is now passing but I was getting a spec failure and 1f1162d fixed and I have no idea why. So I'd like to investigate that before asking for reviews.

deivid-rodriguez:
I investigated the issues and finally decided to revert #7138, because I couldn't find a good solution that does not cause issues.

deivid-rodriguez:
Oh, this might fix #7169, by the way.

deivid-rodriguez:
This fixes several failures when testing locally on systems with a per-user bundler configuration, so I'm merging it. We can revisit if we find a solution that does not run into those issues.

@bundlerbot r+

PR #7167, opened 2019-05-09
Minor spec improvements

deivid-rodriguez:

The problem was that some specs had harcoded paths from @indirect's file system. :sweat_smile:. These lockfiles are clearly incorrect but specs are not failing.

My diagnosis is that these lockfiles are unnecessary for these tests and don't affect what these tests are testing.

My fix is to remove the lockfiles.

As a consequence of the change, the "comment numbering" got messed up. So I removed the numbering while keeping the comments to avoid that problem from happening again in the future.


hsbt:
@bundlerbot r+

indirect:
👍

Issue #7171, opened 2019-05-12
Prevent specific gem from locking


annikoff:
Hello. Here is my case. I have a core application and gem with extensions. For different installations, I use different branches of the extension's gem. Like this:
```ruby
# Gemfile
gem 'my_extension', git: 'url_to_repo', branch: ENV['MY_EXTENTION_BRANCH']
```
`Gemfile.lock` is in core repo. So, I wonder if there any convenient way to skip specific gem from locking in `Gemfile.lock`? For example by adding an option `skip_locking`:
```ruby
# Gemfile
gem 'my_extension', git: 'url_to_repo', branch: ENV['MY_EXTENTION_BRANCH'], skip_locking: true
```

As an experiment, I've changed bundler's source code on my development environment and it is working.

```diff
# lib/bundler/source/git.rb
# Bundler::Source::Git
- attr_reader :uri, :ref, :branch, :options, :submodules
+ attr_reader :uri, :ref, :branch, :options, :submodules, :skip_locking

def initialize(options)
+ @skip_locking = options["skip_locking"] || false
  @options = options
  @glob = options["glob"] || DEFAULT_GLOB

  @allow_cached = false
  @allow_remote = false
  # ....
end

def self.from_lock(options)
- new(options.merge("uri" => options.delete("remote"))
+ new(options.merge("uri" => options.delete("remote"), "skip_locking" => options["skip_locking"]))
end

```
```diff
# lib/bundler/dependency.rb
# Bundler::Dependency
+ attr_reader :skip_locking

def initialize(name, version, options = {}, &blk)
  type = options["type"] || :runtime
  super(name, version, type)

+ @skip_locking   = options["skip_locking"] || false
  @autorequire    = nil
  # ...
end
```
```diff
# lib/bundler/lockfile_generator.rb
# Bundler::LockfileGenerator
def add_dependencies
  out << "\nDEPENDENCIES\n"

  handled = []
  definition.dependencies.sort_by(&:to_s).each do |dep|
    next if handled.include?(dep.name)
+   next if dep.skip_locking
    out << dep.to_lock
    handled << dep.name
  end
end
```
```diff
# lib/bundler/source_list.rb
# Bundler::SourceList
def lock_sources
+ git_sources = self.git_sources.reject(&:skip_locking)
  if Bundler.feature_flag.lockfile_uses_separate_rubygems_sources?
    [[default_source], @rubygems_sources, git_sources, path_sources, plugin_sources].map do |sources|
[...]

Issue #4015, opened 2015-09-25
Bundler.require when Gemfile.lock isn't present


quoideneuf:
I'm not sure if this behavior is intentional or not. (If not, I guess this is a an objection rather than a bug report).

If I have a gem in a non-default group like "test" and run

````
bundle install --without test
````

I get a gemset with just my defaults. If my app uses Bundler.require I would expect it to run fine without the test gems (i.e., if I'm bundling the app for distribution). However, this isn't the case if I don't include Gemfile.lock.

I wrote a simple script to illustrate this:

https://github.com/quoideneuf/ImBunderCurious

Thanks

segiddins:
That's intentionally, you should be running `bundle install` before calling `Bundler.require`

quoideneuf:
Right, but my script does that...

segiddins:
And then you're removing the lockfile... Don't do that.

quoideneuf:
So, does that mean that Bundler.require requires Gemfile.lock to be present? If that is the case, the overall behavior still seems counter-intuitive. Here I've added some lines to the script:

https://github.com/quoideneuf/ImBunderCurious/blob/master/script.sh#L23-L30

Also, if Gemfile.lock IS required for Bundler.require, why not have a message similar to the one you get when Gemfile is missing?
````
Could not locate Gemfile or .bundle/ directory (Bundler::GemfileNotFound)
````

philnash:
I'm intrigued by this error. Could you post the error output from your example app? (I don't have rvm availlable right now.)

quoideneuf:
@philnash Sure, here it is - thanks for taking a look:

https://github.com/quoideneuf/ImBunderCurious/blob/master/out.txt

segiddins:
That is all the expected bundler behavior...

quoideneuf:
If that is the case, I think it's rather un user-friendly. If presence of Gemfile.lock is a hard requirement for Bundler.require, it should fail the same way it does when Gemfile is absent. No?

segiddins:
It actually isn't the lock file that's needed, but rather that all of the gems that running `bundle lock` would resolve to have to be installed

quoideneuf:
Assuming that the Gemfile includes default and test groups, Bundle.require:

succeeds if all default and test gems are present, but Gemfile.lock is absent
succeeds if only default gems are present, and Gemfile.lock is present
fails if only default gems are present, but Gemfile.lock is absent

[...]

PR #7173, opened 2019-05-17
Add Bundler::Definition.no_lock accessor for skipping lock file creation/update

jeremyevans:

I would like a way to avoid creating a `Gemfile.lock` file when doing `require "bundler/setup"` when I have a `Gemfile` without a `Gemfile.lock`.

`Bundler::Definition#lock` is called when requiring `bundler/setup`, and attempts to create `Gemfile.lock`.  I was able to avoid the problem by doing this:

```ruby
require 'bundler'
Bundler::Definition.send(:define_method, :lock){|*|}
require "bundler/setup"
```

However, that is a very brittle approach that is prone to breakage in future updates to bundler if Bundler's internal API changes.

Add `Bundler::Definition.no_lock` accessor.  Set to `true`, this makes `Bundler::Defintion#lock` be a no-op.  Then you can do this to avoid the creation of the lock file:

```ruby
require 'bundler'
Bundler::Definition.no_lock = true
require "bundler/setup"
```

The advantage of this is that it can handle future changes to the bundler internal API.

This seemed to be the simplest approach that would work.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -7,6 +7,11 @@ module Bundler
   class Definition
     include GemHelpers
 
+    class << self
+      # Do not create or modify a lockfile (Makes #lock a noop)
+      attr_accessor :no_lock
+    end
+
     attr_reader(
       :dependencies,
       :locked_deps,
@@ -325,6 +330,8 @@ module Bundler
     end
 
     def lock(file, preserve_unknown_sections = false)
+      return if Definition.no_lock
+
       contents = to_lock
 
       # Convert to \r\n if the existing lock has them
--- a/spec/bundler/definition_spec.rb
+++ b/spec/bundler/definition_spec.rb
@@ -31,6 +31,16 @@ RSpec.describe Bundler::Definition do
           to raise_error(Bundler::TemporaryResourceError, /temporarily unavailable/)
       end
     end
+    context "when Bundler::Definition.no_lock is set to true" do
+      subject { Bundler::Definition.new(nil, [], Bundler::SourceList.new, []) }
+      before { Bundler::Definition.no_lock = true }
+      after { Bundler::Definition.no_lock = false }
+
+      it "does not create a lock file" do
+        subject.lock("Gemfile.lock")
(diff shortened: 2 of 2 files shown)


indirect:
@jeremyevans hi Jeremy! Thanks for sending a PR. Can you tell us more about _why_ you want to run bundler/setup without creating a lock file? This is the first time I've heard someone say they wanted to do that, so I'm interested to hear more about what your goals are and how this change helps you reach your goals.

jeremyevans:
As to why I don't use a `Gemfile.lock`,  in the application where I need this, my `Gemfile` is already locked down to specific version ranges for most of my gems. Gems with compiled components come from the operating system's package manager, and other gems are installed manually (not via bundler). I need to be able to tolerate differences in the version of gems between multiple systems that run this application in different environments (including isolated environments with no internet access), to the extent that I specified them in my `Gemfile`.

Another situation where this change is beneficial would be for testing libraries.  IMO, libraries should not commit a `Gemfile.lock`, they should only commit a `Gemfile`, in order to get the widest possible testing of different supported dependency versions.  Without a way to prevent the creation of the lockfile, the lockfile will be created the first time the tests are run, and not updated afterward.  One of the reasons I currently do not use bundler in any of my libraries is because it forces the use of `Gemfile.lock`.

One significant problem with the current behavior is that if `Gemfile` exists and `Gemfile.lock` does not, Bundler will try to create `Gemfile.lock`, and raises an exception if it cannot.  This can happen if the user has read access to the directory but not write access, or if write access is prevented using system call filtering (in high security environments).

I should note that `bundler/setup` does not need the lockfile it creates in order to perform its function.  IMO, `bundler/setup` should use `Gemfile.lock` if it exists, and then try `Gemfile` if it does not, but I don't think it should try to write a `Gemfile.lock`.  However, that may present backwards compatibility issues, and this explicit approach should not present such issues.

The idea to avoid locking in certain cases seems to have popped up a couple of times before (#7171, #4015).  I'll agree with you that the need is not common, but when you need it, it would be nice to have it in a way that you know will not break in future bundler updates.

PR #7174, opened 2019-05-18
Revert "Remove unused filter"

deivid-rodriguez:

The problem was that I removed the `:git` filter from our specs in ac282dc9de5e20db85e9cdf23b97763a2a2cc292 because I thought it was unused, but it's not.

My diagnosis was that the filter is actually used once:

https://github.com/bundler/bundler/blob/3ec8165eec0a1af8c45ee258d3e7cb61fba875a2/spec/update/git_spec.rb#L162

My fix is to revert the commit.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -69,8 +69,11 @@ RSpec.configure do |config|
     config.filter_run_excluding :realworld => true
   end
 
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Merging this, because it's really straightforward.

@bundlerbot r+

PR #7177, opened 2019-05-23
`bundle gem` spec refactor

deivid-rodriguez:

The problem was that before I figured I needed to revert https://github.com/bundler/bundler/pull/7138 due to the specs no longer being stable, I debugged some failures in the `bundle gem` suite and ended up writing some refactorings.

My diagnosis was that something was wrong with these tests but in the end they were just flaky because of #7038.

In any case, I think the refactorings I wrote improve readabily. In particular, extending the global `reset!` method called before each spec, and reusing it again in the middle of some of these specs was kind of confusing.

### DIFF
--- a/spec/commands/newgem_spec.rb
+++ b/spec/commands/newgem_spec.rb
@@ -1,11 +1,6 @@
 # frozen_string_literal: true
 
 RSpec.describe "bundle gem" do
-  def reset!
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
@bundlerbot r+

PR #7179, opened 2019-05-23
Random test order

deivid-rodriguez:
This is a follow-up to the work started by @colby-swandale in #7103.

The problem was that our specs do not run in random order, so the tests might be depending on state set by other tests to pass.

My diagnosis was that we should enable random test ordering.

My fix is to randomize the specs and fix the resulting failures.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -2,3 +2,4 @@
 --color
 --warnings
 --require spec_helper
+--order random
--- a/Rakefile
+++ b/Rakefile
@@ -97,15 +97,10 @@ namespace :spec do
   end
 
   desc "Run the spec suite with the sudo tests"
-  task :sudo => %w[set_sudo spec clean_sudo]
+  task :sudo => %w[set_sudo spec]
 
   task :set_sudo do
     ENV["BUNDLER_SUDO_TESTS"] = "1"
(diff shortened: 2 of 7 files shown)


deivid-rodriguez:
This seems to be passing, but I'll run CI a couple more times to make sure it's reasonably stable.

deivid-rodriguez:
I restarted CI one more time. If this passes, that would be the 3rd time in a row, plus the one by our bot, that would be 4 greens, which seems enough to me for getting this in.

So... let's see.

deivid-rodriguez:
And yet another flaky surfaced... :sweat_smile:

deivid-rodriguez:
Unless CI fails, this is good for a review now.

I also included changes to fix `bin/rspec --bisect` so we can easily fix flakies in the future. With the previous setup, it would just hang.

deivid-rodriguez:
No reviews, but this should be good to go. Please tag me if you see any flakies :+1:.

@bundlerbot r+

indirect:
omg amazing working bisect this is the best <3

deivid-rodriguez:
First master build after this was merged failed, so I already need it :rofl:

PR #7185, opened 2019-05-29
Try to help GitHub recognize the MIT license

indirect:
It seems like GitHub can't tell what license this is, despite explicitly
naming it inside the file. These changes make our license file closer to
other license files that GitHub does successfully recognize as MIT.
Hopefully it'll work.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,23 +1,22 @@
-Portions copyright (c) 2010 Andre Arko  
+The MIT License
+
+Portions copyright (c) 2010-2019 André Arko
 Portions copyright (c) 2009 Engine Yard
 
-MIT License
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
 
-Permission is hereby granted, free of charge, to any person obtaining
-a copy of this software and associated documentation files (the
-"Software"), to deal in the Software without restriction, including
-without limitation the rights to use, copy, modify, merge, publish,
-distribute, sublicense, and/or sell copies of the Software, and to
-permit persons to whom the Software is furnished to do so, subject to
-the following conditions:
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
 
-The above copyright notice and this permission notice shall be
-included in all copies or substantial portions of the Software.
-
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
-EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
-MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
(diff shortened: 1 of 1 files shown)


hsbt:
👍 

Also need to update https://github.com/bundler/bundler/blob/master/bundler.gemspec#L50

indirect:
@hsbt ah, thank you! fixed.

deivid-rodriguez:
CI should get green here after https://github.com/bundler/bundler/pull/7186 is merged.

tieguy:
FWIW, the .md file name extension doesn't matter to GH - it scans anything with LICENSE in the name. The problem is purely in the content. (To test, install and run https://github.com/licensee/licensee)

Looks like `licensee` will still be confused by the "portions copyright" - `licensee --diff` reports only a 94% similarity, and IIRC GitHub's internal settings require 98%. 

Unfortunately, the problem here looks to be with "portions" - `licensee` ignores lines that start with 'copyright', so if you delete simply "portions" from the current LICENSE.md you get a 98+% match. The rearrangement proposed in this patch doesn't address that problem, unfortunately.

One possible solution is to move the "portions" lines to a COPYRIGHT file, as discussed in [this document from Software Freedom Law Center](https://softwarefreedom.org/resources/2012/ManagingCopyrightInformation.html#maintaining-centralized-copyright-notices).

tieguy:
(Oh, and I'm just hearing that with the new license announcements last week they may have changed their license scanning setup, so take all the above with a grain of salt - I'm sure the file extension still won't matter, but the rest may be less relevant soon. Investigating more, will update as soon as I know more.)

tieguy:
Weirdly, `licensee --diff` and `licensee --detect` have different matches for LICENSE.md. :/ Maybe a rerun of https://github.com/licensee/licensee/issues/347

(And it does look like `licensee` is still what is powering this API. Will dig in more this afternoon.)

deivid-rodriguez:
@tieguy I think I got it, thanks to your help! :) Look:

```
$ licensee diff --license MIT LICENSE.md 
Comparing to MIT License:
Input Length:      1032
License length:    1020
Similarity:      98.91%
diff --git a/LICENSE b/LICENSE
index 0ce42af..d287560 100644
--- a/LICENSE
+++ b/LICENSE
@@ -1,14 +1,14 @@
{+mit license+} permission is hereby granted, free of charge, to any person
obtaining a copy of this software and associated documentation files (the
"software"), to deal in the software without restriction, including without
limitation the rights to use, copy, modify, merge, publish, distribute,
sublicense, and/or sell copies of the software, and to permit persons to whom
the software is furnished to do so, subject to the following conditions: the
above copyright notice and this permission notice shall be included in all
copies or substantial portions of the software. the software is provided "as
is", without warranty of any kind, express or implied, including but not limited
to the warranties of merchantability, fitness for a particular purpose and
noninfringement. in no event shall the authors or copyright holders be liable
for any claim, damages or other liability, whether in an action of contract,
tort or otherwise, arising from, out of or in connection with the software or
the use or other dealings in the software.

$ sed -i '/MIT License/d' LICENSE.md # Delete the mismatch

$ licensee diff --license MIT LICENSE.md 
Comparing to MIT License:
Input Length:       1020
License length:     1020
Similarity:      100.00%
Exact match!
```

deivid-rodriguez:
Actually, with the specific content change introduced by this PR, `licensee` also reports 100%. So I would say this PR is good to go, except we don't need the rename, so it will become much simpler.

indirect:
Updated without the rename, `licensee` shows 100% certainty when I run it. Let's do this. 👍 

@bundlerbot r=hsbt

PR #7186, opened 2019-05-29
Fix azure CI issue

deivid-rodriguez:

The problem was that Azure CI is now failing.

My diagnosis was that the ruby version was upgraded, and as a consequence, a hardcoded path in our configuration no longer exists.

My fix is to update the path and to be more explicit about the ruby version we want to use, so that this doesn't happen again.

I chose this fix because it fixes the problem, and prevents it from happening in the future.

### DIFF
--- a/.azure-pipelines/steps.yml
+++ b/.azure-pipelines/steps.yml
@@ -16,7 +16,7 @@ steps:
   displayName: 'work around readline crash (for https://github.com/bundler/bundler/issues/6902)'
 
 - script: |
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Merging to unblock to other PRs.

@bundlerbot r+

deivid-rodriguez:
Well, they are not technically blocked, since bors only runs TravisCI, but they all have a permanent red cross and that goes away with this PR.

deivid-rodriguez:
Well, that didn't work after all... I'll think of a solution.

PR #7187, opened 2019-05-29
Add debugging information to flaky test

deivid-rodriguez:

The problem was that last master build failed with a test failure I had never seen: https://travis-ci.org/bundler/bundler/jobs/538089338.

My diagnosis was that this failure _is_ indeed order dependent, since it only appeared after enabling test suite randomization. BUT, this failure is not consistently reproducible.

This PR does not fix the problem. It only adds some debugging information, so we have more information to figure this out when it happens again.

I chose this ~fix~ change because it's no clear to me why this happened or how to fix it.

### DIFF
--- a/spec/bundler/fetcher/compact_index_spec.rb
+++ b/spec/bundler/fetcher/compact_index_spec.rb
@@ -11,11 +11,14 @@ RSpec.describe Bundler::Fetcher::CompactIndex do
   end
 
   describe "#specs_for_names" do
+    let(:thread_list) { Thread.list.select {|thread| thread.status == "run" } }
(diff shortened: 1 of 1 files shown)


indirect:
@bundlerbot r+

indirect:
I’m a little worried that the `let` might cache the result of the select and produce unexpected results later, but it should work regardless for this particular test.

deivid-rodriguez:
You are right, this kind of thing is better to compute every time you need it.

I actually saved it in a local variable initially, but then I thought that since this information was only relevant in the sad path, it was better to not compute it every time. However, after a second thought, I think you are right. Should I move it back to a local variable inside the test body?

PR #7188, opened 2019-05-29
Second try at azure fix

deivid-rodriguez:
This is a followup to #7186.

The problem was that Azure CI is still unstable.

My diagnosis was that sometimes MRI 2.4.3 is used, and sometimes MRI 2.4.5 is used instead.

My fix, for now, is to make the steps independent from the ruby included in the container that it's run. 

I chose this fix because other better fixes require more work and I don't want to spend a lot of time on this at the moment.


deivid-rodriguez:
The Travis CI status was not properly received by Github, but this should be good regardless.

@bundlerbot r+

PR #7190, opened 2019-06-04
Remove incorrect `man:clobber` task

deivid-rodriguez:

The problem was that `man:clobber` was listed under `bin/rake -T` but it wasn't working because it was trying to remove a non existent folder.

My diagnosis was that the correct task is actually `bin/rake man:clean`.

My fix is to remove `man:clobber` and move its description on top of `man:clean`, so that it's correctly listed under `bin/rake -T`.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -242,6 +242,7 @@ namespace :man do
       end
       task :build_all_pages => "index.txt"
 
+      desc "Remove all built man pages"
       task :clean do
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
@bundlerbot r+

deivid-rodriguez:
@bundlerbot retry

PR #7193, opened 2019-06-08
More `require_relative`

deivid-rodriguez:
This is a follow up to #7062 and #7100, migrating the last missing internal requires to use `require_relative`. I thought I had migrated everything already, but I had not.

As a note, I'm migrating thor's code here. I will open a PR to upstream's thor to incorporate this changes, but thor is still stuck with 1.8 support, so we have to wait a bit. Given the very low activity thor has, it's fine to make the changes here first, and wait until we can incorporate them upstream.
 
The problem was that sometimes we can end up requiring the default version of bundler, instead of the current copy that's being run, and that's dangerous and can cause version mismatch issues.

My fix is to always use `require_relative` for internal requires.

### DIFF
--- a/exe/bundle
+++ b/exe/bundle
@@ -7,7 +7,7 @@ Signal.trap("INT") do
   exit 1
 end
 
-require "bundler"
+require_relative "../lib/bundler"
 # Check if an older version of bundler is installed
 $LOAD_PATH.each do |path|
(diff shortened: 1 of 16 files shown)


deivid-rodriguez:
Thanks @hsbt!

@bundlerbot r+

PR #7196, opened 2019-06-10
Backport the commits from ruby/ruby

hsbt:

I merged the current master branch into ruby/ruby repository. Some of the tests are failing because the current code is not working with ruby core repository.

The failing examples specified the bundler/bundler repository structure and not worked with a recent version of Ruby.

Added the condition of the structure of ruby core repository and better handles the exception.

There are no other options.


hsbt:
@bundlerbot r+

hsbt:
@bundlerbot retry

PR #7200, opened 2019-06-11
Bump net-http-persistent to 3.1.0

deivid-rodriguez:

The problem was that I wanted to propose some changes to this vendored gem, but I found out that we are using an old version of it.

My diagnosis was that we should upgrade.

My fix is to upgrade the dependency. Since it's a major update, it required some changes. Also, I had to:

* Add a new artifice task to vendorize new `connection_pool` dependency. This is the main downside of this PR, that the new version adds a dependency on this gem. But this gem is very stable, and rarely changes and releases new versions, as can be seen by its [releases](https://github.com/mperham/connection_pool/releases). 

* Cherry-pick a Windows fix not yet merged into master branch of `net-http-persistent`: https://github.com/drbrain/net-http-persistent/pull/90.


colby-swandale:
I'm ✅ on upgrading the gem but i would much prefer we have the open PR in net-http-presistent merged first before merging this.

deivid-rodriguez:
Yeah. It's been ignored for two years now, though... :grimacing:

colby-swandale:
Looking into the current status with https://github.com/drbrain/net-http-persistent/pull/90. We probably don't have a choice, and the PR is fairly simple that i doubt it will cause an issue. So i'm 👍 on this.

Can we just document (even just a comment in the Rakefile) that we've cherry-picked this change incase we need to reference it in the future.

deivid-rodriguez:
A comment about the cherry-pick sounds good, yeah. I'll add it :+1:.

deivid-rodriguez:
Another downside of this PR is that the automatiek task for vendoring `net-http-persistent` now needs manual intervention because its `connection_pool` requires need to be edited to point to the vendored version.

deivid-rodriguez:
I'll try to fix the last issue I mentioned in `automatiek`.

deivid-rodriguez:
I created a patch for automatiek and described in the PR's description how it would be used for this specific case: https://github.com/segiddins/automatiek/pull/3.

Rakefile
```diff
@@ -269,6 +276,10 @@ else
     lib.vendor_lib = "lib/bundler/vendor/thor"
   end
 
+  # We currently cherry-pick changes to use `require_relative` internally
+  # instead of regular `require`. These changes have already landed to fileutils
+  # master, but have not yet been released, so we cherry-pick them for the time
+  # being.
```
deivid-rodriguez: @hsbt If you have time, it would be nice to release `fileutils` master as fileutils-1.2.1, so we don't need to cherry-pick anything.

deivid-rodriguez:
The net-http-persistent PR has been merged! Maybe we'll be able to jump straight to the next release :tada:

deivid-rodriguez:
Updated to use net-http-persistent 3.1.0, which was released yesterday, including the PR we needed.

Only thing left here would be an `automatiek` release including https://github.com/segiddins/automatiek/pull/3, so that `net-http-persistent` can be vendored without manual intervention.

But I don't think it's a hard requirement for this PR.

deivid-rodriguez:
This was approved by @hbst and @colby-swandale, so I'll merge it.

Vendoring a future version of `net-http-persistent` is not automatic, but almost (only `connection_pool` references need manual intervention).

So no big deal, we can wait for the PR to be merged and reviewed, and update `automatiek` later.

@bundlerbot r+

deivid-rodriguez:
Actually, the comment about `thor` is incorrect, let me fix it.

@bundlerbot r-

deivid-rodriguez:
Done, and also added a comment about the current `automatiek` situation.

@bundlerbot r+

PR #7201, opened 2019-06-12
More spec failure debugging

deivid-rodriguez:

The problem was this spec failure is still happening... https://travis-ci.org/bundler/bundler/jobs/544642885.

My diagnosis was that we should add more information to the failure to help figuring it out.


deivid-rodriguez:
Thanks to this debugging, I found the culprit! See #7202.

Closing here.

PR #7202, opened 2019-06-13
Fix thread leak in compact index specs

deivid-rodriguez:

The problem was that sometimes CI fails randomly. For example, https://travis-ci.org/bundler/bundler/jobs/544642885.

My diagnosis was that the spec after it was mocking `Worker#stop` and thus preventing the threads to be actually stopped. That leak would after the spec checking for the number of threads.

My fix is to also call the original `#stop` method, so that no threads are left behind.

There's an alternative option, which is to completely remove this spec. In my opinion, this spec is checking exactly the same thing as the previous spec (the whole purpose of `Worker#stop` is to stop threads), but in a more brittle way. So I'm tempted to completely kill the spec instead.

### DIFF
--- a/spec/bundler/fetcher/compact_index_spec.rb
+++ b/spec/bundler/fetcher/compact_index_spec.rb
@@ -22,7 +22,7 @@ RSpec.describe Bundler::Fetcher::CompactIndex do
     end
 
     it "calls worker#stop during the run" do
-      expect_any_instance_of(Bundler::Worker).to receive(:stop).at_least(:once)
+      expect_any_instance_of(Bundler::Worker).to receive(:stop).at_least(:once).and_call_original
 
       compact_index.specs_for_names(["lskdjf"])
     end


deivid-rodriguez:
Thanks @colby-swandale. Any opinion on fully removing the test vs this PR?

deivid-rodriguez:
I finally went with removing the spec, and also fixed an incorrect spec nesting since I was at it. Let me know what you think!

colby-swandale:
Hmmm i'm a bit on the fence here. IMO the Compact Index spec knowing that `Bundler::Worker` is creating threads is leaking concerns a bit. Could we improve the test suite by moving that spec into the `Bundler::Worker` specs and leave the spec that's checking `Worker#stop` the same?

colby-swandale:
I should be more clear, when i say moving the spec into the specs for `Bundler::Worker`. I mean the idea of testing a worker with a set amount of threads, and checking that they have been closed afterwards.

deivid-rodriguez:
I decided to go back to the previous version of the fix, just to get the flaky fix. We can improve the test organization in a separate set of changes :+1:.

I did keep the nesting fix, but I can also remove it if you want.

deivid-rodriguez:
Went back to the approved version, so merging!

@bundlerbot r+

PR #7206, opened 2019-06-14
Allow bundle activation if v2.0 >= than required

mcelicalderon:

If a newer version of bundler ins installed on the system, but the one on the Gemfile.lock is not, running commands using binstubs will fail for bundler >= 2.0

The template is too strict with v2.0

Activate bundler versions as long as they are greater than the one on the lock file

This is just a quick fix to the problem I describe, other improvements can be done I just wanted to validate first that this would be a desired behavior.

I tested locally that the problem is fixed. If you think this should work like that (v1 used to and still does) I can fix all the broken specs.

### DIFF
--- a/lib/bundler/templates/Executable.bundler
+++ b/lib/bundler/templates/Executable.bundler
@@ -75,17 +75,19 @@ m = Module.new do
   end
 
   def activate_bundler(bundler_version)
-    if Gem::Version.correct?(bundler_version) && Gem::Version.new(bundler_version).release < Gem::Version.new("2.0")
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
We're relaxing this in #7259, so it should be better.

We're still failing on major version mismatches, so I'm going to close this, since at least for the time being, that's expected behavior.

Let me know if I'm missing something and thanks for your contribution anyways! :heart: 
Closing this since

mcelicalderon:
No problem. Looks like #7259 will take care of that too.

PR #7209, opened 2019-06-17
Add a minimum ruby version to the generated gemspec 

deivid-rodriguez:
This PR is a finished version of #4397.

The problem was that the generated gemspec for new gems does not impose a minimum ruby version.

My fix is to set the minimum ruby version in the generated gemspec to the minimum ruby version supported by bundler itself.

I chose this fix because it's simple.

### DIFF
--- a/lib/bundler/templates/newgem/newgem.gemspec.tt
+++ b/lib/bundler/templates/newgem/newgem.gemspec.tt
@@ -12,6 +12,7 @@ Gem::Specification.new do |spec|
 <%- if config[:mit] -%>
   spec.license       = "MIT"
 <%- end -%>
+  spec.required_ruby_version = Gem::Requirement.new(">= 2.3.0")
 
   spec.metadata["allowed_push_host"] = "TODO: Set to 'http://mygemserver.com'"
 
--- a/spec/commands/newgem_spec.rb
+++ b/spec/commands/newgem_spec.rb
@@ -274,6 +274,12 @@ RSpec.describe "bundle gem" do
         to match(/mygemserver\.com/)
     end
 
+    it "sets a minimum ruby version" do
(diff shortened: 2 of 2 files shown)


deivid-rodriguez:
@bundlerbot r=hsbt

lib/bundler/templates/newgem/newgem.gemspec.tt
```diff
@@ -12,6 +12,7 @@ Gem::Specification.new do |spec|
 <%- if config[:mit] -%>
   spec.license       = "MIT"
 <%- end -%>
+  spec.required_ruby_version = Gem::Requirement.new(">= 2.3.0")
```
simi: If I understand it will, this should be the same as https://github.com/bundler/bundler/blob/c5dcdfbc3617d1ecf551f047456d4ea03a18f930/bundler.gemspec#L34. But to prevent loading bundler gemspec during generating new gem, this is hardcoded in here, but covered by spec ensuring those two values are the same.

deivid-rodriguez: Yes, although I didn't really consider your alternative solution: load bundler's gemspec, and load the minimum version from there. It sounds like it might be a better approach since it's more DRY and we don't need a spec to keep the minimum versions in sync... :thinking:.

simi: Ahh, I was thinking loading bundler's gemspec is overkill during generating new gem. Anyway I can inspect that option if needed. Feel free to ping me.

deivid-rodriguez: Yeah, I don't have a strong opinion, this is probably good enough.

PR #7210, opened 2019-06-17
Upgrade to molinillo's master

deivid-rodriguez:
The reason for this PR is that I was adding some comments about the current state of our vendored dependencies as requested in https://github.com/bundler/bundler/pull/7200, and I noticed that we are currently using [the latest commit in molinillo's master](https://github.com/CocoaPods/Molinillo/pull/130) but not some prior work.

This is not necessarily a problem, but it might be better to include the whole upstream repo as it is on its master branch now, because it gets us ready to support a future release.

My fix is to include the latest changes in molinillo's master in our vendored copy.

Another option would be to do nothing here and just wait for the next molinillo's release. Maybe `molinillo` is even ready to actually release its master branch.

### DIFF
--- a/lib/bundler/vendor/molinillo/lib/molinillo/compatibility.rb
+++ b/lib/bundler/vendor/molinillo/lib/molinillo/compatibility.rb
@@ -8,7 +8,7 @@ module Bundler::Molinillo
     if [].respond_to?(:flat_map)
       # Flat map
       # @param [Enumerable] enum an enumerable object
-      # @block the block to flat-map with
+      # @param blk the block to flat-map with
       # @return The enum, flat-mapped
       def flat_map(enum, &blk)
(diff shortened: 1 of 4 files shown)


segiddins:
We should only be using releases of Molinillo, rather than pulling straight from master

hsbt:
I also agree with @segiddins. But continuous integration is helpful for us. Can we add the master branch of molinillo to CI matrix?

deivid-rodriguez:
Yeah, I agree too in general.

In this particular case, I cherry-picked the `require_relative` changes because they seemed to fix issues with bundler installed as a default gem, and also seemed safe.

What's the state of current molinillo's master? Would it be possible to make a new release that the next version of bundler can use? As I recall, there were only some changes for better error reporting.

PR #7211, opened 2019-06-17
Use real paths for `bundle clean`

### LINKED ISSUE #7208: "bundle clean" removes git dependencies with RubyGems 3+ when path involves symlink
The `bundle clean` command removes gems that are still being used if all 3 of these criteria are true:

1. RubyGems 3.0+ is being used.
2. The bundler installation path contains a symlink anywhere in the path (it could be the final directory, or any intermediate).
3. A git dependency is present in your Gemfile.

If all 3 of these situations apply, then `bundle clean` will always remove the git dependencies present in your Gemfile, despite the fact that those gems are still in use. This occurs in both Bundler 2 and Bundler 1.7.

I'm wondering if this is related to these symlink changes in RubyGems 3.0: https://github.com/rubygems/rubygems/pull/2352 However, since this issue is specific to git dependencies supported by Bundler, that's why I'm filing the ticket here, but let me know if I should file this on RubyGems instead.

Here's a demonstration of the issue (with Bundler 2.0.2, but the same behavior is seen with Bundler 1.17.3 as long as RubyGems 3+ is used):

```sh
# Create a Gemfile with a git dependency.
mkdir test-bundler
cd test-bundler
printf 'source "https://rubygems.org"\ngem "rack", :git => "https://github.com/rack/rack.git"' > Gemfile

# Create a symlink
mkdir real-path
ln -s real-path symlink-path

# Ruby, RubyGems, and Bundler versions
ruby --version
# ruby 2.6.3p62 (2019-04-16 revision 67580) [x86_64-darwin16]
gem update --system 3.0.4
gem install bundler --version 2.0.2
gem --version
# 3.0.4
bundle --version
# Bundler version 2.0.2

# Install to symlink
bundle install --path=./symlink-path

# Run "bundle clean", which removes the git dependency.
bundle clean --verbose
# Running `bundle clean --verbose` with bundler 2.0.2
# Found no changes, using resolution from the lockfile
# Removing rack (70c838f7204f)

bundle exec ruby -e 'puts "Hello, world."'
# https://github.com/rack/rack.git (at master@70c838f) is not yet checked out. Run `bundle install` first.
```

Note that this issue does not occur if you're not using a symlink in RubyGems 3+:

```sh
gem --version
# 3.0.3
bundle --version
# Bundler version 2.0.2

bundle install --path=./real-path

bundle clean --verbose
# Running `bundle clean --verbose` with bundler 2.0.2
# Found no changes, using resolution from the lockfile

bundle exec ruby -e 'puts "Hello, world."'
# Hello, world.
```

It also does not occur if you *are* using a symlink, but are using RubyGems 2:

```sh
gem update --system 2.7.10

gem --version
# 2.7.10
bundle --version
# Bundler version 2.0.2

bundle install --path=./symlink-path

bundle clean --verbose
# Running `bundle clean --verbose` with bundler 2.0.2
# Found no changes, using resolution from the lockfile

bundle exec ruby -e 'puts "Hello, world."'
# Hello, world.
```

deivid-rodriguez:
Fixes #7208.

The problem was that since rubygems 3.0, `bundle clean` incorrectly cleans git gems when they are installed to a symlinked location, but still being used.

My diagnosis was that since https://github.com/rubygems/rubygems/pull/2352, `Gem.dir` returns an array of realpaths, not symlinked ones. However, we don't do the same resolution on the `bundler` side, so in this situation git gems are not correctly skipped from the cleanup.

My fix is to resolve the array of paths `bundle clean` uses to do its thing into an array of real paths, just like rubygems does now.

I chose this fix because it fixes the problem.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -317,7 +317,7 @@ module Bundler
     end
 
     def spec_git_paths
-      sources.git_sources.map {|s| s.path.to_s }
+      sources.git_sources.map {|s| File.realpath(s.path) }
     end
 
     def groups
--- a/spec/commands/clean_spec.rb
+++ b/spec/commands/clean_spec.rb
@@ -183,6 +183,32 @@ RSpec.describe "bundle clean" do
     expect(vendored_gems("bin/rackup")).to exist
   end
 
+  it "keeps used git gems even if installed to a symlinked location" do
+    build_git "foo", :path => lib_path("foo")
+    git_path = lib_path("foo")
+    revision = revision_for(git_path)
+
+    gemfile <<-G
+      source "file://#{gem_repo1}"
+
+      gem "rack", "1.0.0"
+      git "#{git_path}", :ref => "#{revision}" do
+        gem "foo"
+      end
+    G
+
+    FileUtils.mkdir_p(bundled_app("real-path"))
+    FileUtils.ln_sf(bundled_app("real-path"), bundled_app("symlink-path"))
+
+    bundle "install", forgotten_command_line_options(:path => bundled_app("symlink-path"))
+
+    bundle :clean
+
+    expect(out).not_to include("Removing foo (#{revision[0..11]})")
+
+    expect(bundled_app("symlink-path/#{Bundler.ruby_scope}/bundler/gems/foo-#{revision[0..11]}")).to exist
+  end
+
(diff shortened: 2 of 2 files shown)


deivid-rodriguez:
@bundlerbot r=colby-swandale

PR #7212, opened 2019-06-18
Don't re-resolve locked platform specific gems

deivid-rodriguez:

The problem was that when you have locked platform gems in your lockfile for a platform different than the running one, the lockfile was always getting re-resolved.

My diagnosis was that the dependencies passed on to the `SpecSet#for` call that's supposed to check that no dependencies are missing for the current platform, always include all dependencies for all platforms. And thus platform missing dependencies are always detected. The reason for that is that `expand_dependencies` was being passed `remote = true`

https://github.com/bundler/bundler/blob/0b5da910e71760eb585b34d08d43ea31b988ca24/lib/bundler/definition.rb#L815-L816

and thus not excluding dependencies exclusive for other platforms

https://github.com/bundler/bundler/blob/0b5da910e71760eb585b34d08d43ea31b988ca24/lib/bundler/definition.rb#L887

My fix is to skip the second argument to `expand_dependencies` in this case, so that it works in its default "local mode" and the dependencies for other platforms are properly filtered out.

I chose this fix because it fixes the issue in a simple way.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -813,7 +813,7 @@ module Bundler
 
       resolve = SpecSet.new(converged)
       expanded_deps = expand_dependencies(deps, true)
-      @locked_specs_incomplete_for_platform = !resolve.for(expanded_deps, @unlock[:gems], true, true)
+      @locked_specs_incomplete_for_platform = !resolve.for(expand_dependencies(deps), @unlock[:gems], true, true)
       resolve = resolve.for(expanded_deps, @unlock[:gems], false, false, false)
       diff    = nil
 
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
A bit of context on how I discovered this issue.

* I've been running bundler's master for a while.
* On repos where I have "multi platform gemfiles", I started noticing that bundler's master would print "Resolving dependencies..." to the screen a lot of times.
* After investigation, this bug is not new but became apparent since ccb5eae03bb29f167dd165c7ad14b3254fc53912, because since that commit, bundler's UI is initialized earlier, and as a consequence any logging bundler does starts earlier (right after `require "bundler/setup"`). This made the "Resolving dependencies..." messages appear and that's how I discovered the issue.

deivid-rodriguez:
I'll merge this fix tomorrow unless anybody has objections.

indirect:
@bundlerbot r+

PR #7213, opened 2019-06-18
Replace `bundler_2:install` task with `bundler_3:install`

deivid-rodriguez:

The problem was that the `bundler_2:install` task is no longer meaningful.

My diagnosis was that master points to 2.x now, so this task is useless.

My fix is move it to `bundler_3:install`.

I chose this fix because it keeps the shortcut, but for a useful future version.


deivid-rodriguez:
@bundlerbot r+

PR #7215, opened 2019-06-19
Auto multiplatform

deivid-rodriguez:

The problem was that currently if the gemfile include gems for specific platforms, like the default `gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby] ` Rails uses, bundler will _always_ print a noisy warning, but I don't think users are doing anything wrong, just running `bundle install` on it.

Also, unless they run `bundle lock --add-platform x86-mingw32 x86-mswin32 x64-mingw32 java`, and commit the resulting lockfile, they will continue to get the same warning over and over again.

My diagnosis was that the right thing to do is that instead of warning users, bundler does the right thing by default. That is:

* Resolve the gemfile and lock it for all platforms present in the Gemfile.
* Install gems for the current platform if requested.

My fix is to do just that.

I chose this fix because I think it's better for our users. We currently have a specific setting to avoid this warning, which I guess we added given the complaints. We no longer need that setting, nor the warning, so I removed both.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -519,9 +519,7 @@ module Bundler
     end
 
     def add_current_platform
-      current_platform = Bundler.local_platform
-      add_platform(current_platform) if Bundler.feature_flag.specific_platform?
-      add_platform(generic(current_platform))
+      current_platforms.each {|platform| add_platform(platform) }
     end
 
     def find_resolved_spec(current_spec)
@@ -544,6 +542,14 @@ module Bundler
     end
 
   private
+
+    def current_platforms
+      current_platform = Bundler.local_platform
+      platforms = []
+      platforms << current_platform if Bundler.feature_flag.specific_platform?
+      platforms << generic(current_platform)
+      platforms
+    end
 
     def change_reason
       if unlocking?
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -113,7 +113,7 @@ module Bundler
       end
       @unlocking ||= @unlock[:ruby] ||= (!@locked_ruby_version ^ !@ruby_version)
 
-      add_current_platform unless Bundler.frozen_bundle?
+      add_platforms unless Bundler.frozen_bundle?
 
       converge_path_sources_to_gemspec_sources
       @path_changes = converge_paths
@@ -542,6 +542,12 @@ module Bundler
     end
 
   private
+
+    def add_platforms
+      (@dependencies.flat_map(&:expanded_platforms) + current_platforms).uniq.each do |platform|
+        add_platform(platform)
+      end
+    end
 
     def current_platforms
       current_platform = Bundler.local_platform
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -518,10 +518,6 @@ module Bundler
       raise InvalidOption, "Unable to remove the platform `#{platform}` since the only platforms are #{@platforms.join ", "}"
     end
 
-    def add_current_platform
-      current_platforms.each {|platform| add_platform(platform) }
-    end
-
     def find_resolved_spec(current_spec)
       specs.find_by_name_and_platform(current_spec.name, current_spec.platform)
     end
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -892,7 +892,7 @@ module Bundler
         next if !remote && !dep.current_platform?
         platforms = dep.gem_platforms(sorted_platforms)
         if platforms.empty? && !Bundler.settings[:disable_platform_warnings]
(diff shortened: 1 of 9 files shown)


deivid-rodriguez:
To expand a bit more on the rationale of this PR:

When a team adds a platform specific dependency to their Gemfile, they should know and consider that the dependency will be unused for developers under other platforms. But once the decision is made, I don't see the need to warn the developers of the other platforms every time they use bundler.

deivid-rodriguez:
Marked this as "ready for review" because... I think it's ready.

In my opinion, doing this makes bundler more pleasant to use for multiplatform gemfiles, because it's transparent to users and does not require any special treatment.

lib/bundler/definition.rb
```diff
@@ -545,6 +539,20 @@ def unlocking?
...
 
+    def add_platforms
+      (@dependencies.flat_map(&:expanded_platforms) + current_platforms).uniq.each do |platform|
+        add_platform(platform)
+      end
+    end
+
+    def current_platforms
+      current_platform = Bundler.local_platform
+      platforms = []
```
simi: What about to use `tap` in here?

```ruby
[].tap do |platforms|
  platforms << current_platform if Bundler.feature_flag.specific_platform?
  platforms << generic(current_platform)
end
```

man/bundle-config.ronn
```diff
@@ -176,8 +176,6 @@ learn more about their operation in [bundle install(1)](bundle-install.1.html).
    When set, Gemfiles containing multiple sources will produce errors
    instead of warnings.
    Use `bundle config unset disable_multisource` to unset.
-* `disable_platform_warnings` (`BUNDLE_DISABLE_PLATFORM_WARNINGS`):
```
simi: Is there any way how to inform users using this setting that it is not valid anymore and can be removed?

simi:
Is this change worth it creating new minor release?

lib/bundler/definition.rb
deivid-rodriguez: Soungs good, yeah. I'll update it.

man/bundle-config.ronn
deivid-rodriguez: I've been planning to work on setting validation & deprecation for a while, but I haven't yet found the time. For the time being, these non-existing settings are just ignored, which doesn't sound like a specially bad behavior. But I plan to improve on this in the future :+1:.

deivid-rodriguez:
Yes, we'll cut the next release as 2.1.0 from master. It this PR lands, it should be included with it.

lib/bundler/definition.rb
deivid-rodriguez: Done in https://github.com/bundler/bundler/pull/7215/commits/d7b5cf1fcf5cd86cd579450414cb9ff5ab03cf5e.

indirect:
Hmm. I guess my impression of the warning was that it was telling you that you should remove the useless gem from your Gemfile, because it was never going to be installed. Is there a case where you would actively want to have the gem in your Gemfile, but also want to only install it on platforms you have never locked to? That seems contradictory to me.

deivid-rodriguez:
In my opinion, the fact that some gems will be unused under some platforms is inherent to the multiplatform feature itself, so we shouldn't warn about it because it's expected.

Take the following Gemfile for example (a simplification of the default Gemfile Rails generators create):

```ruby
source "https://rubygems.org"

gem "rails"

gem "tzinfo-data", platforms: "jruby"
```

If I type that Gemfile, it means that I'm explicitly opting into the multiplatform feature. So the behavior I would want as a user when I run `bundle install` on it is:

* Resolve and lock the Gemfile for all platforms (`jruby` and whatever platform I'm running).
* Install the resolution for the platform that I'm currently running.

That way, when the other developers of my team using `jruby` install the `Gemfile`, they install a predictable, deterministic resolution.

Currently, the only way to get that behavior is to run:

```
$ bundle install
$ bundle lock --add-platform java
```

But there's no way to do it without getting a warning on the first `bundle install`, which annoys people because it makes you think you're doing something wrong.

If you only plan to use MRI, you shouldn't specify any jruby-specific gems in your Gemfile and write instead:

```ruby
source "https://rubygems.org"

gem "rails"
```

If on the other hand you only plan to use jruby, then you should not specify any `platform` option at all and write

```ruby
source "https://rubygems.org"

gem "rails"

gem "tzinfo-data"
```

So, to sum up, I think the range of platforms users expect to support can be guessed implicit from the Gemfile and the running platform. So we should resolve by default for all those platforms and don't warn about behavior that's most likely expected.

The behavior I described is what this PR implements.

indirect:
Ooooooooohhhh. That's a good description. I'm convinced. 😄

indirect:
@bundlerbot r+

PR #7217, opened 2019-06-20
Fix a typo on a spec comment

deivid-rodriguez:

The problem was there's a typo on a comment.

My diagnosis was that typos are bad because they difficult reading and distract devs.

My fix is to spell bundler properly.

I chose this fix because `bundler` is the only word that makes sense in the context :)


deivid-rodriguez:
@bundlerbot r+

PR #7218, opened 2019-06-20
Bump rubocop to 0.71

deivid-rodriguez:

The problem was that [our build broke](https://travis-ci.org/bundler/bundler/jobs/548144536) due to a dependency incompatibility caused by a new release of `rubocop-performance`.

My diagnosis was that our requirements were very loose for `rubocop-performance`, but very strict for `rubocop`, so very likely to break on `rubocop-performance` releases.

My fix is to upgrade `rubocop` to fix the current issue, and to lock `rubocop-performance` so it doesn't happen again in the future.

I chose this fix because it works now and in the future.


deivid-rodriguez:
Quick merge to get CI back to green. Happy to address any feedback later :+1:.

@bundlerbot r+

PR #7221, opened 2019-06-26
Bump rubygems in CI to 3.0.4 and 2.7.10

deivid-rodriguez:

The problem was that we should make sure everything is ok with the most recent rubygems releases.

My fix is to upgrade the rubygems versions in our CI.


hsbt:
@bundlerbot r+

PR #7222, opened 2019-06-26
Remove `add_development_dependency` from new gems

deivid-rodriguez:

The problem was that a lot of gems in the wild use `add_development_dependency` for their development dependencies, but using `Gemfile`'s for that is more useful.

My diagnosis was that the current situation is due to the fact that bundler generates a gem skeleton that uses `add_development_dependency` by default.

My fix is to stop using `add_development_dependency` in generated gems, and instead use the Gemfile.

I chose this fix because it encourages better practices.

### DIFF
--- a/lib/bundler/templates/newgem/Gemfile.tt
+++ b/lib/bundler/templates/newgem/Gemfile.tt
@@ -2,3 +2,11 @@ source "https://rubygems.org"
 
 # Specify your gem's dependencies in <%= config[:name] %>.gemspec
 gemspec
+
+gem "rake", "~> 12.0"
+<%- if config[:ext] -%>
+gem "rake-compiler"
+<%- end -%>
+<%- if config[:test] -%>
+gem "<%= config[:test] %>", "~> <%= config[:test_framework_version] %>"
+<%- end -%>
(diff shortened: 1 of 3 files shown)


simi:
Does this mean `add_development_dependency` is going to be deprecated and removed in the future? Also why not to use common groups (`development`, `test`) in `Gemfile` for rake, ext-compiler and test framework?

deivid-rodriguez:
Deprecating it is kind of tough at the moment, it's too heavily used so it would be too annoying for the community in my opinion. I think reducing its usage by making it opt-in, not opt-out, on new gems is a good start. We could revisit it in the future, and see if deprecating it makes sense by then.

Regarding using groups, I'm not sure whether the distinction is useful here. Generally speaking, I think only `:development` and `:test` are useful on a gem context, and all three gems we include by default would probably make sense in both groups, so I feel it's easier to skip any grouping.

deivid-rodriguez:
@simi Are you ok with this? Thanks for all your late reviews, by the way, they are really helpful.

deivid-rodriguez:
@bundlerbot r+

PR #7223, opened 2019-06-26
RUBY_ENGINE should always be defined after 1.8.7

deivid-rodriguez:

The problem was that we have code checking that `RUBY_ENGINE` everytime it's used and I think it's unnecessary.

My diagnosis was that every ruby implementation newer than 1.8 should define this.

My fix is to remove the unnecessary code.


deivid-rodriguez:
@bundlerbot r+

PR #7224, opened 2019-06-27
Remove unnecessary `order_matters` param

deivid-rodriguez:

The problem was that the `find_gemfile` method was receiving an `order_matters` parameter that's no longer used anywhere.

My fix is to remove the parameter.

### DIFF
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -20,7 +20,7 @@ module Bundler
     end
 
     def default_gemfile
-      gemfile = find_gemfile(:order_matters)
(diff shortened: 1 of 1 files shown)


hsbt:
👍 

I confirmed to not use `find_gemfile` with `order_matters` parameter on existence gems.

https://gist.github.com/hsbt/469fbe590754295dfce59aaada838cb5

hsbt:
@bundlerbot r+

deivid-rodriguez:
Thanks, I assumed this was safe because it's a private method, but it's great that you checked :heart:

PR #7225, opened 2019-06-27
Clearer `MissingRevision` git errors

deivid-rodriguez:

The problem was that sometimes we get the error "Revision xxxxxx does not exist in the repository https://github.com/my_user/my_repo. Maybe you misspelled?"

My diagnosis was that it's not always easy to troubleshoot this error, because you don't even know which git command failed in the first place.

My fix is to also include in the error the command that originally failed.

I chose this fix because I think it will help users with troubleshooting.

### DIFF
--- a/spec/bundler/source/git/git_proxy_spec.rb
+++ b/spec/bundler/source/git/git_proxy_spec.rb
@@ -128,17 +128,18 @@ RSpec.describe Bundler::Source::Git::GitProxy do
 
     context "when given a SHA as a revision" do
       let(:revision) { "abcd" * 10 }
+      let(:command) { "reset --hard #{revision}" }
 
       it "fails gracefully when resetting to the revision fails" do
         expect(subject).to receive(:git_retry).with(start_with("clone ")) { destination.mkpath }
         expect(subject).to receive(:git_retry).with(start_with("fetch "))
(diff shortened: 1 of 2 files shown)


spec/bundler/source/git/git_proxy_spec.rb
```diff
@@ -136,8 +136,11 @@
         expect(subject).not_to receive(:git)
 
         expect { subject.copy_to(destination, submodules) }.
-          to raise_error(Bundler::Source::Git::MissingGitRevisionError,
-            "Revision #{revision} does not exist in the repository #{uri}. Maybe you misspelled it?")
+          to raise_error(
+            Bundler::Source::Git::MissingGitRevisionError,
+            "Git error: command `git command` in directory #{destination} has failed.\n" \
```
simi: :thinking: to make this assertion closer to the real scenario you can maybe change line 135 to (command can be stored to variable to be used twice):

```ruby
git_command = "reset --hard #{revision}"
expect(subject).to receive(:git).with(git_command).and_raise(Bundler::Source::Git::GitCommandError, git_command)
```

And assert `git reset --hard...` is part of the error message.

deivid-rodriguez: Done in https://github.com/bundler/bundler/pull/7225/commits/ca96316b97353b4344ba283be0051ef7e9d9a013.

deivid-rodriguez:
Going in with @simi's blessing :)

@bundlerbot r+

PR #7226, opened 2019-06-27
Improve quality specs

deivid-rodriguez:

I'm not really sure of the exact context when I made these changes, but I recall I was getting some invalid encoding errors somehow, and they turned out to be an estraneous `’` quote in one of the documentation files.
 
My diagnosis was that we should not include this character anywhere, but instead use regular quotes.

My fix is to add a quality spec for this, and while I was at it, I added some improvements to quality specs in general.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ If you'd like to contribute to Bundler, that's awesome, and we <3 you. We've put
 
 If you'd like to request a substantial change to Bundler or to the Bundler documentation, refer to the [Bundler RFC process](https://github.com/bundler/rfcs) for more information.
 
-While some Bundler contributors are compensated by Ruby Together, the project maintainers make decisions independent of Ruby Together. As a project, we welcome contributions regardless of the author’s affiliation with Ruby Together.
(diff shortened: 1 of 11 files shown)


simi:
Hmm, I have never seen quality specs before. Is that "poor man's rubocop alternative"? Do we really still need it since there's rubocop check done on CI?

deivid-rodriguez:
It's similar but not the same. Quality specs are used for tasks that are outside of rubocop's scope, mainly because they are quality checks that apply to all (or most) files in the projects, not only to files containing ruby code.

deivid-rodriguez:
I fixed this PR since it had failed to detect a lot of those non-standard quotes. Now, I can close this since I can't reproduce the original problem, and maybe this symbol is more common when typing on English keyboards than I thought? I do find the consistency of using always the same symbol for single quotes nice and less distracting.

deivid-rodriguez:
I'll merge this one tomorrow unless anybody has objections.

deivid-rodriguez:
:heart:

@bundlerbot r=hsbt

PR #7229, opened 2019-07-03
Correct default cache path

Confusion:
The default path where `bundle package` puts the gems is `vendor/cache`, not `vendor/bundle`.


deivid-rodriguez:
Thanks so much @Confusion and @simi!

@bundlerbot r+

PR #7231, opened 2019-07-04
Simplify and improve documentation tasks

deivid-rodriguez:

The problem was that running bundler's tests, or installing the development bundler locally leaves generated documentation files around, and that's not neat and can affect developers in various ways. For example, by incorrectly including [bundler's generated documentation inside rubygems's manifest](https://github.com/rubygems/rubygems/pull/2813#issuecomment-507046061).

My diagnosis was that we can make this less likely to happen by not generating docs during our specs (it's not needed), and by cleaning them up after `rake build` or `rake install` (they are needed in the installed copy, but not in the development copy).

My fix does that, and cleans up other related stuff.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,3 @@ man/*
 
 # rspec failure tracking
 .rspec_status
-
-# files generated during packaging
-/lib/bundler/generated/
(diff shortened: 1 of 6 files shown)


simi:
Special :+1: for introducing manual pages build to CI linting stage.

Can be https://github.com/bundler/bundler/blob/391b11966efb140a9ba204ca4e05ec7694296ce0/.travis.yml#L7 removed now?

deivid-rodriguez:
Oooops, I had missed that `before_script` line. Maybe some tests need it after all... I'll remove the line and see.

deivid-rodriguez:
Closing in favor of #7245.

PR #7232, opened 2019-07-04
Parallelize test suite

deivid-rodriguez:

The problem was that tests are slow.

My diagnosis was that an easy improvement would be to run them in parallel.

My fix is to add the `parallel_tests` gem, and use that to run specs.

I chose this fix because it requires virtually no changes and it speeds up tests around 30%. Further work could be done to equally balance test processes by runtime, to make it faster.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -35,6 +35,7 @@ namespace :spec do
   desc "Ensure spec dependencies are installed"
   task :deps do
     Spec::Rubygems.dev_setup
+    Spec::Rubygems.build
   end
 
   namespace :travis do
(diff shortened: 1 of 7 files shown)


hsbt:
I prefer this solution. How about the improvement of time duration for test execution?

deivid-rodriguez:
Hi @hsbt!

Balancing test times would involve using the `RuntimeLogger` formatter and managing the `tmp/parallel_runtime_rspec.log` through TravisCI cache. See https://github.com/grosser/parallel_tests#even-test-group-run-times.

Do you prefer that I work on that before merging this PR, or as a further improvement on a separate PR?

hsbt:
I'm okay to merge this. Let's try to work for further improvement after that.

deivid-rodriguez:
Ok, thank you!

@bundlerbot r=hsbt

deivid-rodriguez:
There's one randomly failing spec that's most likely caused by this PR. I'll investigate it.

deivid-rodriguez:
Closed in favor of #7317.

PR #7233, opened 2019-07-04
Add missing expectation

deivid-rodriguez:

The problem was I noticed that this test was missing one assertion. See 

https://github.com/bundler/bundler/blob/566cce0629a0f1a4d147195a34d494306269e1cf/lib/bundler/shared_helpers.rb#L77-L84

My diagnosis was that it should be added.

My fix is to add the missing expectation.

I chose this fix because regardless of how implementation dependent this spec is, it should probably be complete.


deivid-rodriguez:
@bundlerbot r=hsbt

PR #7235, opened 2019-07-05
Spec improvements

deivid-rodriguez:
This is a bunch of miscelaneous commits that fix and improve some stuff in our test suite. I could open separate PRs for them but they are so small that I figured I could compile them all on a single PR.


deivid-rodriguez:
@bundlerbot r+

PR #7236, opened 2019-07-07
Fix triple spacing in lockfile

### LINKED ISSUE #7087: Lockfile generates 3 spaces
https://github.com/bundler/bundler/blob/d44d803357506895555ff97f73e60d593820a0de/lib/bundler/lockfile_generator.rb#L89 seems to generate lines with 3 spaces. 

E.g. The `BUNDLED WITH` section has 3 spaces before `2.0.1`

```
GEM
  remote: https://rubygems.org/
  specs:
    diff-lcs (1.3)
    rspec (3.8.0)
      rspec-core (~> 3.8.0)
      rspec-expectations (~> 3.8.0)
      rspec-mocks (~> 3.8.0)
    rspec-core (3.8.0)
      rspec-support (~> 3.8.0)
    rspec-expectations (3.8.2)
      diff-lcs (>= 1.2.0, < 2.0)
      rspec-support (~> 3.8.0)
    rspec-mocks (3.8.0)
      diff-lcs (>= 1.2.0, < 2.0)
      rspec-support (~> 3.8.0)
    rspec-support (3.8.0)

PLATFORMS
  ruby

DEPENDENCIES
  rspec

BUNDLED WITH
   2.0.1
```

Not sure if this is intentional or not ?

ajwann:

The BUNDLED WITH and RUBY VERSION section of the Gemfile.lock have an extra space before each line of content.

This is happening in the ```add_section``` method of the ```LockfileGenerator``` class, any time
that method receives a string for it's ["value" argument](https://github.com/bundler/bundler/blob/d44d803357506895555ff97f73e60d593820a0de/lib/bundler/lockfile_generator.rb#L77)

Remove this extra space from line 89.

Closes #7087.


lib/bundler/lockfile_generator.rb
```diff
@@ -86,7 +86,7 @@ def add_section(name, value)
           out << "  #{key}: #{val}\n"
         end
       when String
-        out << "   #{value}\n"
```
ajwann: This is causing the ```BUNDLED_WITH``` and ```RUBY_VERSION``` sections of the lockfile to have an extra space in from of each line of content.

ajwann:
I'll take care of these specs

colby-swandale:
I'm a bit cautious about accepting this PR because this is technically a breaking change and may break systems that parses the Gemfile.lock in a strict manner. I also don't think this is worth going through the entire feature flags process, so I'm a bit torn on how to approach this. Maybe we just leave this until Bundler 3 or even just let it have the extra space.. ¯\_(ツ)_/¯

simi:
@colby-swandale do you have any real example why anyone should parse Gemfile.lock outside of bundler itself?

ajwann:
@colby-swandale I'm also interested to know why someone would parse the lockfile.

deivid-rodriguez:
I tend to think this is fine too?

@ajwann Can you squash all commits into one for bisectability?

ajwann:
@deivid-rodriguez sure, no problem.

ajwann:
I had to rebase off master to pick up a new (failing) spec and fix it, but it should be all good now.

PR #7237, opened 2019-07-08
Remove `add_development_dependency` from gemspec

deivid-rodriguez:

The problem was that our development setup is complicated, and that we are using `add_development_dependency` even though in my opinion its usage should be discouraged.

My diagnosis was that our development dependencies are split between the gemspec and the `spec/support/rubygems_ext` file. We can simplify this and centralize everything in the support file.

My fix is to simplify our development setup by moving everything to `spec/support/rubygems_ext`, while also getting rid of `add_development_dependency` in our gemspec.

I chose this fix because it simplifies our setup and it further discourages usage of `add_development_dependency` (after #7222).

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -3,9 +3,7 @@
 $:.unshift File.expand_path("../lib", __FILE__)
 require "benchmark"
 
-def development_dependencies
-  @development_dependencies ||= Gem::Specification.load("bundler.gemspec").development_dependencies
-end
+require_relative "spec/support/rubygems_ext"
 
 # Benchmark task execution
 module Rake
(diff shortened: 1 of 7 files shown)


colby-swandale:
I don't favour this change as it is breaking away from typical convention of gems. Also the new location of development gems is in a place that's not very intuitive for devs/users.

indirect:
Too bad we can't do what we suggest to everyone else, and use a Gemfile 😅

deivid-rodriguez:
Yeah... I did try to use a Gemfile, but I just couldn't get it working. Maybe I'll try again in the future :)

@bundlerbot r=indirect

indirect:
sorry! didn’t mean to imply that you should use a gemfile, I think this is the best we can do for now :+1:

hsbt:
I prefer this change. It’s better to separate `DEV_DEPS` and `DEPS` from `rubygems_ext.rb`. I sometimes forgot they are placed on `rubygems_ext.rb`.

indirect:
@hsbt I think using a Gemfile does not work? but if you know how to make it work, I would be happy to use it.

PR #7238, opened 2019-07-08
Point to CoC which contains the contributor covenant

### LINKED ISSUE #6905: Generated README contains two different links to contributor covenant CoC
I just generated a new gem with `--coc --no-mit`, and this results in a README that points 'contributors' to the Contributor Covenant website, but also points 'Everyone' to the code of conduct included in the gem's repository. That these are in fact the same is not immediately obvious, so perhaps the README should just point to the gem's CoC?

ajwann:

The generated README contains two different links to contributor covenant. Since the generated Code of Conduct *contains* the text of the Contributor Covenant, we can just point the generated README to the generated CoC, and explicitly refer to it as the code of conduct. 

The generated readme now points to the gems CoC (which contains the Contributor Covenant)

Closes #6905.

### DIFF
--- a/lib/bundler/templates/newgem/README.md.tt
+++ b/lib/bundler/templates/newgem/README.md.tt
@@ -32,7 +32,8 @@ To install this gem onto your local machine, run `bundle exec rake install`. To 
 
 ## Contributing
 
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Commits squashed! Thanks so much @ajwann!

@bundlerbot r+

deivid-rodriguez:
@bundlerbot r+

PR #7239, opened 2019-07-09
Fully remove compatibility guard

deivid-rodriguez:

The problem was that this code is untested and never run.

My diagnosis was that we can remove it.

My fix is to remove the code.

I actually intented to remove this in https://github.com/bundler/bundler/pull/6983, but after a discussing it with @simi I decided to keep it (see https://github.com/bundler/bundler/pull/6983#discussion_r259327593). After a second though I think this is safe to remove and that the potential use case (that the latest bundler is allowed to be installed by a really really old rubygems that didn't support ruby version constraints) is not a problem. The `required_ruby_version` constraint landed in [rubygems 0.6, 15 years ago](https://github.com/rubygems/rubygems/commit/240a9d3a3dc7211cad9680ce2579f502e215b519).


deivid-rodriguez:
Moving on with @simi's blessing :heart:

@bundlerbot r+

PR #7242, opened 2019-07-10
Restore remembered options in manpages

### LINKED ISSUE #7198: bundle-config(1) manpage says `bundle install` doesn't remember options but it does
From https://github.com/bundler/bundler/blob/v2.0.1/man/bundle-config.ronn#remembering-options

> Flags passed to `bundle install` or the Bundler runtime,
such as `--path foo` or `--without production`, are not remembered between commands.
If these options must be remembered,they must be set using `bundle config`
(e.g., `bundle config path foo`).

IIRC it was true in 2.0.0.dev but had been reverted back to the original behavior. Maybe we should reverse the change to manpages in this commit? https://github.com/bundler/bundler/commit/e120f40f72fe53f55e05db9204f98d28f1065e81

uasi:

The problem was that the remembered options of `bundle install` were not documented in manpages.

Restore the section on remembered options from https://github.com/bundler/bundler/commit/e120f40f72fe53f55e05db9204f98d28f1065e81 and add description for some options.

They should be fully documented.

Closes #7198.

### DIFF
--- a/man/bundle-config.1
+++ b/man/bundle-config.1
@@ -60,7 +60,10 @@ Executing bundle with the \fBBUNDLE_IGNORE_CONFIG\fR environment variable set wi
 Executing \fBbundle config set disable_multisource true\fR upgrades the warning about the Gemfile containing multiple primary sources to an error\. Executing \fBbundle config unset disable_multisource\fR downgrades this error to a warning\.
 .
 .SH "REMEMBERING OPTIONS"
-Flags passed to \fBbundle install\fR or the Bundler runtime, such as \fB\-\-path foo\fR or \fB\-\-without production\fR, are not remembered between commands\. If these options must be remembered, they must be set using \fBbundle config\fR (e\.g\., \fBbundle config set path foo\fR)\.
+Flags passed to \fBbundle install\fR or the Bundler runtime, such as \fB\-\-path foo\fR or \fB\-\-without production\fR, are remembered between commands and saved to your local application\'s configuration (normally, \fB\./\.bundle/config\fR)\.
+.
+.P
+However, this will be changed in bundler 3, so it\'s better not to rely on this behavior\. If these options must be remembered, it\'s better to set them using \fBbundle config\fR (e\.g\., \fBbundle config set path foo\fR)\.
 .
 .P
 The options that can be configured are:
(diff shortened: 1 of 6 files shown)


deivid-rodriguez:
What if we documented both things? I mean, document how things work on bundler 2.x, but also start spreading the word about the upcoming changes in bundler 3.x by keeping the current documentation as well. Of course, making it obvious that it's future, not current, behavior.

uasi:
Didn't know that the plan to remove remembered options is still on the table. @deivid-rodriguez Is there a tracking issue for the upcoming changes? I think we can say "see this issue for the future changes on remembered options" in the documentation.

deivid-rodriguez:
The future changes are what's documented right now, I think. The changes are already implemented and included in master, but they are not yet enabled by default.

uasi:
@deivid-rodriguez okay, added some notes for future changes.

deivid-rodriguez:
@uasi Thanks a lot for your efforts. So, in my opinion, the current documentation is _almost_ fine, because the advice it gives it's correct and a good practice for _current_ and _future_ versions of bundler, so we should stick to that.

Only thing we need to do here in my opinion, is to clarify when are options to commands remembered, so my proposed changes here would be along the following lines:

```diff
diff --git a/man/bundle-config.ronn b/man/bundle-config.ronn
index ffecc12f6..069f3b95f 100644
--- a/man/bundle-config.ronn
+++ b/man/bundle-config.ronn
@@ -51,10 +51,13 @@ config unset disable_multisource` downgrades this error to a warning.
 
 ## REMEMBERING OPTIONS
 
-Flags passed to `bundle install` or the Bundler runtime,
-such as `--path foo` or `--without production`, are not remembered between commands.
-If these options must be remembered, they must be set using `bundle config`
-(e.g., `bundle config set path foo`).
+Flags passed to `bundle install` or the Bundler runtime, such as `--path foo` or
+`--without production`, are remembered between commands and saved to your local
+application's configuration (normally, `./.bundle/config`).
+
+However, this will be changed in bundler 3, so it's better not to rely on this
+behavior. If these options must be remembered, it's better to set them using
+`bundle config` (e.g., `bundle config set path foo`).
 
 The options that can be configured are:
 
```

deivid-rodriguez:
Ping @uasi.

deivid-rodriguez:
I fixed this PR with my own feedback, because I really want to include this in the next release!

deivid-rodriguez:
Going in! Thanks @uasi!

@bundlerbot r+

PR #7245, opened 2019-07-10
Commit generated docs to source control

deivid-rodriguez:

The problem was that generating docs on the fly every time they are needed (tests, release, installation) is complicated and prone to errors. For example, currently the `rubygems-update` gem that takes care of updating rubygems and install the default bundler does not properly include bundler docs. Also, managing the manifest as a mix of generated and source controlled files is complicated too. Finally, ruby-core already keeps man pages as part of its sources, so this would make us follow their lead.

My fix is to keep generated man pages source controlled, and add a task to make sure they are always kept in sync with their "source of truth".

I chose this fix because I think it should simplify things for us.


deivid-rodriguez:
Any opinions about this?

I know the diff is big, but I don't see this as being too hard to maintain. Whenever we update docs, we need to regenerate them and I added a task to make sure they don't fall out of sync. And everything else becomes much simpler in my opinion. Rubocop documentation is also managed similarly and I find it easy to work with.

Somewhere where this would also improve my experience is that I usually try out bundler's master on personal projects by installing it with `bin/rake install:local`, and then making debugging changes and reinstalling. This is currently quite slow because docs are rebuilt every time.

hsbt:
I'm 👍 to this.

Can we add the automation task for generating the documentation each changes?

deivid-rodriguez:
I added a task to make sure that generated docs never fall out of sync from their sources, that it's run as a prerequisite to `bin/rake build`.

Configuring a file watcher that rebuilds each file every time its source changes would be nice, but maybe it's something more adequate for each developer environment? It would be a matter of configure your editor to run `bin/rake man/gemfile.5` everytime `man/gemfile.5.ronn` changes, and so on.

deivid-rodriguez:
@hsbt Are you ok with this then, can I merge? Opinions anybody else?

deivid-rodriguez:
Let's do it!

@bundlerbot r=hsbt

PR #7246, opened 2019-07-11
No need to make gem refresh a noop

deivid-rodriguez:

The problem was that rubygems/rubygems#2711 fails because the test in bundler that checks that `Gem.refresh` reveals no system gems fails.

My diagnosis was that we don't really need to make `Gem.refresh` a noop at all, because we already setup a `post_reset` hook to automatically go back to the set of specs bundler knows about, everytime `Gem::Specification.reset` is called. See https://github.com/bundler/bundler/blob/847db56c6b5d2156d27990d73a9949507693985a/lib/bundler/rubygems_integration.rb#L551-L554.

And `Gem::Specification.reset` is what `Gem.refresh` does: https://github.com/rubygems/rubygems/blob/564bb0a5bb310ccceaf6cd6391b3e67e0712edd5/lib/rubygems.rb#L861-L867.

My fix is to remove the redefinition of `Gem.refresh`. This makes sense to me not only because it's unnecessary, but because the monkeypatch could still be very easily workarounded by calling `Gem::Specification.reset` directly.

After undoing the `Gem.refresh` redefinition, I changed the test to still make sure that the (not overriden) `Gem.refresh` still does not change the set of specs bundler knows about.

This is a nice simplification, but doesn't really fix the spec for https://github.com/rubygems/rubygems/pull/2711, because after the first call to `Bundler.rubygems.find_name("rack")`, the `@stubs_by_name` array in rubygems gets populated and thus is no longer empty.

We can test the same thing without using `@stubs_by_name` under the hood, so I changed the test to do that, so that the rubygems PR is not affected.

I chose this fix because it simplifies the integration between bundler & rubygems, and it unblocks https://github.com/rubygems/rubygems/pull/2711.

### DIFF
--- a/lib/bundler/rubygems_integration.rb
+++ b/lib/bundler/rubygems_integration.rb
@@ -441,13 +441,6 @@ module Bundler
       end
     end
 
-    # Because Bundler has a static view of what specs are available,
-    # we don't #refresh, so stub it out.
-    def replace_refresh
-      gem_class = (class << Gem; self; end)
-      redefine_method(gem_class, :refresh) {}
-    end
-
     # Replace or hook into RubyGems to provide a bundlerized view
     # of the world.
     def replace_entrypoints(specs)
@@ -468,7 +461,6 @@ module Bundler
       replace_gem(specs, specs_by_name)
       stub_rubygems(specs)
       replace_bin_path(specs_by_name)
-      replace_refresh
 
       Gem.clear_paths
     end
--- a/spec/runtime/setup_spec.rb
+++ b/spec/runtime/setup_spec.rb
@@ -845,7 +845,7 @@ end
     end
   end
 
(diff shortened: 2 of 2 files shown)


voxik:
It makes sense to me in the context of https://github.com/rubygems/rubygems/pull/2711. Thx

indirect:
Seems good. 👍 

@bundlerbot r+

indirect:
(for what it's worth, I think we overrode Gem.refresh back before we had a post_reset hook. so it makes sense to me that it might not be needed now.)

deivid-rodriguez:
> (for what it's worth, I think we overrode Gem.refresh back before we had a post_reset hook. so it makes sense to me that it might not be needed now.)

Yeah, it makes sense that the override because unneeded when the post_reset hook was added :+1:.

PR #7247, opened 2019-07-11
Drop rubygems 2.5.0 and 2.5.1 support

deivid-rodriguez:

The problem was that we have some conditional code in bundler dependent on the rubygems version being used, that could be easily removed if we dropped support for rubygems 2.5.0 and 2.5.1.

My diagnosis was that we could drop support for these versions now, to remove all of those conditionals and simplify the integration.

My fix is to drop support for rubygems 2.5.0 and 2.5.1.

I chose this fix because the oldest ruby we currently support, 2.3, ships with rubygems 2.5.2, so it's very unlikely that any user of the next bundler version will be using these old rubygems versions, because they would need to explicitly downgrade rubygems!

### DIFF
--- a/spec/commands/exec_spec.rb
+++ b/spec/commands/exec_spec.rb
@@ -851,7 +851,7 @@ __FILE__: #{path.to_s.inspect}
         bundle :install, :system_bundler => true, :path => "vendor/bundler"
       end
 
-      it "overrides disable_shared_gems so bundler can be found", :ruby_repo, :rubygems => ">= 2.6.2" do
+      it "overrides disable_shared_gems so bundler can be found", :ruby_repo, :rubygems => ">= 2.5.2" do
         system_gems :bundler
         file = bundled_app("file_that_bundle_execs.rb")
         create_file(file, <<-RB)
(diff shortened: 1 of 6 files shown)


simi:
:thinking: If I understand it well even downgrade will not be possible since required rubygems constraint in gemspec. Am I right?

deivid-rodriguez:
Yeah, right. What I meant is that in order for a ruby 2.3 user to not be allowed to install the next version of bundler, she would need to explicitly downgrade rubygems.

spec/commands/exec_spec.rb
```diff
@@ -851,7 +851,7 @@ def bin_path(a,b,c)
         bundle :install, :system_bundler => true, :path => "vendor/bundler"
       end
 
-      it "overrides disable_shared_gems so bundler can be found", :ruby_repo, :rubygems => ">= 2.6.2" do
```
hsbt: Is it collect? (We will support only >= RG 2.5.2)

deivid-rodriguez: I tried to run this spec on rubygems 2.5.2 and it passed, so I assumed this was a typo in the filter... :man_shrugging:.

hsbt: 👍  I see.

hsbt:
@bundlerbot r+

PR #7248, opened 2019-07-13
Fix nested bundle exec's when bundler is a default gem

### LINKED ISSUE #7244: nested bundle exec, not working in Ruby trunk/head
Only on Ruby trunk/head, there's a Travis CI test issue with Puma that generates a `LoadError`, where a `bundle exec` command is generated within a process started via `bundle exec`.

So, I used the following command:
```
bundle exec "bundle exec 'ruby -v'"
```

Works fine with Ruby 2.6.3, RG 3.0.3, & bundler 2.0.1.

With Ruby trunk/master/head/2.7.0 (RG 3.1.0.pre1, bundler 2.1.0.pre.1), it fails on the line in the bin stub:
```ruby
load Gem.activate_bin_path('bundler', 'bundle', version)
```
Correct value is:
```
lib/ruby/gems/2.7.0/gems/bundler-2.1.0.pre.1/exe/bundle
```
nested value is the following, which doesn't exist, and generates the LoadError:
```
lib/bin/bundle
```

**EDIT:**

Sorry for not making it explicitly clear, but `bundle exec "ruby -v"` works fine in both.  The only issue is nested `bundle exec`'s, which may be common in testing started with 'bundle exec rake \<test\>`...

MSP-Greg:

The problem was that when bundler is a default gem, nested `bundle exec` commands generate a LoadError.

```
/home/travis/.rvm/rubies/ruby-head/bin/bundle:30:in `load': cannot load such file --
/home/travis/.rvm/rubies/ruby-head/lib/bin/bundle (LoadError)
```

Not accounting for Bundler being installed as a default gem. When it's a default, the lib and exe folders do not share the same root folder.

This was the result of a change in https://github.com/bundler/bundler/commit/e742c3d5f458a4a59cf0eaab2567eca844f956d1 (#7100).

### Repo Example

Using Ruby master/trunk/ruby-head (as of https://github.com/ruby/ruby/commit/0c6c937904aafc1809386bd892a2d114d22d01fe), from a folder where `bundle exec` can be ran:

```
bundle exec "bundle exec 'ruby -v'"
```

Small adjustment to logic for finding the correct exe/bundle file.

I chose this fix because it's similar to previous code.

Fixes #7244.

### DIFF
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -287,9 +287,15 @@ module Bundler
     public :set_env
 
     def set_bundle_variables
+      # bundler exe & lib folders have same root folder, typical gem installation
       exe_file = File.expand_path("../../../exe/bundle", __FILE__)
-      # for Ruby core repository
-      exe_file = File.expand_path("../../../../bin/bundle", __FILE__) unless File.exist?(exe_file)
+
+      # for Ruby core repository testing
+      exe_file = File.expand_path("../../../bin/bundle", __FILE__) unless File.exist?(exe_file)
+
+      # bundler is a default gem, exe path is separate
+      exe_file = Bundler.rubygems.bin_path("bundler", "bundle", VERSION) unless File.exist?(exe_file)
+
       Bundler::SharedHelpers.set_env "BUNDLE_BIN_PATH", exe_file
       Bundler::SharedHelpers.set_env "BUNDLE_GEMFILE", find_gemfile.to_s
       Bundler::SharedHelpers.set_env "BUNDLER_VERSION", Bundler::VERSION
--- a/spec/bundler/shared_helpers_spec.rb
+++ b/spec/bundler/shared_helpers_spec.rb
@@ -402,8 +402,10 @@ RSpec.describe Bundler::SharedHelpers do
 
       it "sets BUNDLE_BIN_PATH to the bundle executable file" do
         subject.set_bundle_environment
-        bundle_exe = ruby_core? ? "../../../../../bin/bundle" : "../../../exe/bundle"
-        expect(ENV["BUNDLE_BIN_PATH"]).to eq(File.expand_path(bundle_exe, __FILE__))
+        bundle_exe = ruby_core? ? "../../../bin/bundle" : "../../../exe/bundle"
+        bin_path = ENV["BUNDLE_BIN_PATH"]
+        expect(bin_path).to eq(File.expand_path(bundle_exe, __FILE__))
+        expect(File.exist?(bin_path)).to be true
       end
     end
 


MSP-Greg:
In Ruby master/trunk/ruby-head, the bundler lib folder exists at:
```
lib\ruby\2.7.0\bundler
```
and the exe file is located at:
```
lib\ruby\gems\2.7.0\gems\bundler-2.1.0.pre.1\exe
```
This is true for both ruby-loco and a Travis build downloaded from http://rubies.travis-ci.org/


Hence, the line
```ruby
exe_file = Bundler.rubygems.bin_path("bundler", "bundle", VERSION) unless File.exist?(exe_file) 
```

must remain in the logic to find the exe file., but should be used only after the relative check of:
```ruby
 exe_file = File.expand_path("../../../exe/bundle", __FILE__) 
```

is done...

lib/bundler/shared_helpers.rb
```diff
@@ -287,8 +287,11 @@ def set_env(key, value)
     public :set_env
 
     def set_bundle_variables
+      # bundler exe & lib folders have same root folder
       exe_file = File.expand_path("../../../exe/bundle", __FILE__)
-      # for Ruby core repository
+      # bundler is a default gem, exe path is separate
+      exe_file = Bundler.rubygems.bin_path("bundler", "bundle", VERSION) unless File.exist?(exe_file)
+      # for Ruby core repository testing
       exe_file = File.expand_path("../../../../bin/bundle", __FILE__) unless File.exist?(exe_file)
```
deivid-rodriguez: Isn't this path incorrect though? I look at current's repository structure in ruby-core and it's like

```
|--- bin/bundle
|--- lib/bundler/shared_helpers.rb
```

So, I think one level of `../` is incorrect?

MSP-Greg: Yes, sort of.  It's incorrect from the source folder.

But, we have three folders, source, build, and install.  With ruby-loco, since it is now used in so many repos for testing, I patch the test files and run all testing from install.  I also do a CLI check from install  (I'll add a nested 'bundle exec' test sometime soon).  So I can't really check it from ruby-loco, as it's using the install folder, which finds the file with the addition.

Anyway, long story short, I did a similar PR in Ruby, and tests passed.  So I don't know if
```ruby
File.expand_path("../../../../bin/bundle", __FILE__)
```

is run from the 'build' folder or the 'source' folder.

I left It as is, I could test it in my Ruby fork on Travis & Appveyor?  Am I making sense?

deivid-rodriguez: Yes, you are. Running the tests from the install folder makes sense to me and it helps catching issues like this one, so thanks.

I'll let @hsbt answer the question about ruby-core testing, but the fix for bundler installed as a default gem looks good to me.

MSP-Greg: I've got a build with a patch with debug output running on Travis now...  20 minutes maybe...

MSP-Greg: @deivid-rodriguez 

Thanks for looking at that, as I ignored the existing code.  Using three levels, it finds the bin file, and it's in the build folder.  So, I pushed that change also.  Maybe core testing doesn't test it...

MSP-Greg:
Re testing and Ruby trunk/master, I added the following to shared_helpers.rb in Ruby:
```ruby
TEMP = File.expand_path("../../../bin/bundle", __FILE__)
STDERR.puts("", "shared_helpers.rb find bundle", TEMP, File.exist?(TEMP), "")
```

Output on Travis was:
```
shared_helpers.rb find bundle
/home/travis/build/MSP-Greg/ruby/bin/bundle
true
```

Hence, @deivid-rodriguez is correct, only three `../` should be used.  See [Travis](https://travis-ci.org/MSP-Greg/ruby/jobs/559054188#L2696)

MSP-Greg:
Sorry for all the damn testing, but, when I first tried this in Ruby core, I used an assembled path (File.join), and it passed.  But, after reviewing the history, I saw that the following was used:

```ruby
Bundler.rubygems.bin_path("bundler", "bundle", VERSION)
```

So, with the 1st & 2nd versions of this PR, I switched to that.  But, unlike File.join, bin_path will raise an error if the path isn't found, which happens when running from the core build folder.  So, the third version flips the last two 'path searches' so that the bin_path is ran last.

I tested locally with ruby-loco with `bundle exec "bundle exec 'ruby -v'"`.

Ruby Core Travis is [here](https://travis-ci.org/MSP-Greg/ruby/builds/559145078), which passed regarding this issue...

deivid-rodriguez:
I think you also need to update the expectation in https://github.com/bundler/bundler/blob/36ce7ccf84968d2a7f0eba98c605a5eac8e17e68/spec/bundler/shared_helpers_spec.rb#L405, so that the test doesn't break when run from ruby-core.

ioquatix:
Can we please merge this?

deivid-rodriguez:
Sorry for the delay, I'm away from home at the moment.

@bundlerbot r+

deivid-rodriguez:
Thanks so much @MSP-Greg!

ioquatix:
Thanks for this! A bunch of my specs which use ruby-head and bundler were broken and I didn't know why!

MSP-Greg:
You're welcome, and thanks for all your work in core and the 'socketry' gems.

We still need it pushed into the bundler copy in core for some repo's CI.  Similar to:

https://github.com/ruby/ruby/commit/8f37629519ad330032a38ac0e871b2912ed38a1b

It's tough to test without a real installation, but I think I can add a test for it in ruby-loco.  It's not 'helpful' when ruby-head CI fails in external repos...

ioquatix:
> It's tough to test without a real installation, but I think I can add a test for it in ruby-loco. It's not 'helpful' when ruby-head CI fails in external repos...

Yeah, it started failing for me and I didn't know what was going on. Maybe can discuss it with core team.

MSP-Greg:
The ruby 'install' has broken before.  Previously, the CLI bin scripts broke.  It got fixed quickly, I don't recall where the breakage was or whether it was platform specific.  It was quickly fixed.

See https://bugs.ruby-lang.org/issues/16012

MSP-Greg:
@hsbt

Thanks for your work pushing bundler updates to Ruby trunk/master.  The current Travis ruby-head build includes the updates, and I checked the repo with the nested bundle commands.

Travis was green.

JFYI, ruby-loco had a problem with `Bundler.rubygems.bin_path`, as after RG's `test_case.rb` is loaded...

I added another statement that found the file using RbConfig.  Probably won't work with prefix/suffix, but I've never tried using them on Windows, as I suspect the *.dll mapping isn't set up to deal with it.  Thanks again...

PR #7249, opened 2019-07-15
Add `bundle package` functionality to `bundle cache`

deivid-rodriguez:

The problem was that the previous plan provided no migration path from the current `bundle cache` to the future one (an alias to `bundle package`).

My diagnosis was that we should probably tell users first that in bundler 3, `bundle cache` will have a different implementation. However, after playing around with it a bit, I noticed that the additions to `bundle cache` provided by `bundle package` are fully backwards compatible. Or maybe I'm missing something but at least all `bundle cache` tests still pass when changing the implementation under the hood to use `bundle package`. As far as I can see, the only backwards incompatible change (start caching `git` and `path` gems by default) is already covered by the exisiting `cache_all` flag), so we should be good.

My fix is to remove the current implementation of `bundle cache`, and replace it with the current implementation of `bundle package`.

I chose this fix because it allows us to remove code and it reduces confusion about `bundle cache` vs `bundle package` making them work exactly the same.

On a future PR, I plan to clean this up so that everything (docs, code, tests) use `bundle cache`, and `bundle package` is left only as an alias.


deivid-rodriguez:
Thoughts about this, @indirect? I was a bit confused by the changes that were going happen here, I think this makes the transition simpler.

indirect:
Oh! That's awesome, thank you for this. :)

@bundlerbot r+

PR #7250, opened 2019-07-16
Fix ruby core dsl spec.rb failure, warning on build_metadata.rb

MSP-Greg:

Travis CI job on Ruby master was failing & had a nuisance warning.

1. dls_spec.rb - Minor change to the error msg regexp match string

2. build_metadata.rb - change `if @git_commit_sha` to `if instance_variable_defined? :@git_commit_sha`

### DIFF
--- a/spec/bundler/dsl_spec.rb
+++ b/spec/bundler/dsl_spec.rb
@@ -271,7 +271,7 @@ RSpec.describe Bundler::Dsl do
     it "will raise a Bundler::GemfileError" do
       gemfile "gem 'foo', :path => /unquoted/string/syntax/error"
       expect { Bundler::Dsl.evaluate(bundled_app("Gemfile"), nil, true) }.
-        to raise_error(Bundler::GemfileError, /There was an error parsing `Gemfile`:( compile error -)? unknown regexp options - trg. Bundler cannot continue./)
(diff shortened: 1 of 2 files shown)


MSP-Greg:
Travis CI is **all green**, including Ruby master/trunk/2.7.0...

spec/bundler/dsl_spec.rb
```diff
@@ -271,7 +271,7 @@
     it "will raise a Bundler::GemfileError" do
       gemfile "gem 'foo', :path => /unquoted/string/syntax/error"
       expect { Bundler::Dsl.evaluate(bundled_app("Gemfile"), nil, true) }.
-        to raise_error(Bundler::GemfileError, /There was an error parsing `Gemfile`:( compile error -)? unknown regexp options - trg. Bundler cannot continue./)
+        to raise_error(Bundler::GemfileError, /There was an error parsing `Gemfile`:( compile error -)? unknown regexp options - trg.+? Bundler cannot continue./)
```
deivid-rodriguez: Isn't the `?` character unnecessary? `+` already expresses optionality.

MSP-Greg: Can't keep regex engine differences straight, so I often use lazy (non-greedy) ...  Regardless, removed, passed.

deivid-rodriguez:
@bundlerbot r+

Thanks!

PR #7251, opened 2019-07-19
Removed the environmental variables that are BUNDLE_USER_*.

hsbt:

When the users uses `BUNDLE_USER_*` environmental variables like `BUNDLE_USER_CONFIG`, the rspec example of Bundler affected by them failed.

The current spec_helper didn't clear and restore them.

I remove and restore them same as `Bundler::EnvironmentPreserver::BUNDLER_PREFIX`

This is the most simple solution with the current implementation.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -101,6 +101,7 @@ RSpec.configure do |config|
     Spec::Rubygems.setup
     ENV["RUBYOPT"] = "#{ENV["RUBYOPT"]} -r#{Spec::Path.spec_dir}/support/hax.rb"
     ENV["BUNDLE_SPEC_RUN"] = "true"
+    ENV["BUNDLE_USER_CONFIG"] = ENV["BUNDLE_USER_CACHE"] = ENV["BUNDLE_USER_PLUGIN"] = nil
 
     # Don't wrap output in tests
     ENV["THOR_COLUMNS"] = "10000"


deivid-rodriguez:
@hsbt I didn't like the fact that we are adding a new public constant in `lib/` that it's unused everywhere, except to fix a problem with the specs under some environments. So I presented an alternative solution in #7255.

hsbt:
#7255 seems good to prepare for this. I will rebase and update this pr after merging #7255

deivid-rodriguez:
@hsbt I think #7255 makes this PR unnecessary because the specs no longer modify the environment?

hsbt:
In my understanding, the examples of bundler modify configurations defined by `BUNDLER_USER_*`. Therefore, Even if bundler doesn't touch the current environment, it will modify `BUNDLE_USER_CONFIG` such as `.bundle/config`.

But It's my miss understanding maybe. I will investigate after merging #7255.

deivid-rodriguez:
Yeah, but every test resets the environment back to the original:

https://github.com/bundler/bundler/blob/a9beb66ed23d9ea5bb1b07256400b503fad80a62/spec/spec_helper.rb#L120

So I think it shouldn't affect?

I'll merge #7255 now.

hsbt:
@deivid-rodriguez I confirmed to the current master is still broken with `BUNDLE_USER_*`. When it's defined on the test environment, the bundler examples modify it and was affected by them.

hsbt:
ah, I should refresh browser tab before submit...

hsbt:
@bundlerbot r+

deivid-rodriguez:
Thanks, I understood the problem better after seeing the final version of this PR. :+1:

PR #7255, opened 2019-07-23
Improve environment backup and restore

deivid-rodriguez:

The problem was that the specs were removing some environment variables for every tests, but not all of them, causing problems on some environments (see #7251).

My diagnosis was that we don't really need to do this.

My fix is to stop doing it, and fix the very few specs that failed after the change. One of the failures actually led to a bug fix (I think), see 6c27ebb.

I chose this fix because I think it's cleaner than adding a public constant in `lib/` only to fix a problem with the specs, like it's done in #7251.


deivid-rodriguez:
@bundlerbot r=hsbt

PR #7256, opened 2019-07-23
Document changelog handling for releases

deivid-rodriguez:

The problem was that I'm not sure how to release bundler.

My diagnosis was that the part about how to handle the changelog is the most unclear step.

My fix is to _guess_ how we are currently managing that, and document that.

I chose this fix because it improves our docs, while starting a PR at the same time where I can incorporate feedback if this is not the right way.

While I read the release scripts, I also added a couple of improvements to the `rake release:open_unreleased_prs` task.

### DIFF
--- a/doc/playbooks/RELEASING.md
+++ b/doc/playbooks/RELEASING.md
@@ -54,12 +54,17 @@ The `rake release:patch` command will automatically handle cherry-picking, and i
 
 Bundler maintains a list of changes present in each version in the `CHANGELOG.md` file.
 Entries should not be added in pull requests, but are rather written by the Bundler
-maintainers in the [bundler-changelog repo](https://github.com/bundler/bundler-changelog).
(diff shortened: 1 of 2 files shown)


indirect:
Nice! I haven’t released a minor since we started doing this, so @segiddins and @colby-swandale are probably the right reviewers 🙂 r+ from me!

deivid-rodriguez:
Thanks @indirect, I'll wait for feedback from the latest releasers :)

deivid-rodriguez:
I'll wait for feedback a few more days. If I don't get any feedback I'll merge this PR and start working on the release assuming this is the right way :)

deivid-rodriguez:
I started working on the changelog for the next version :)

deivid-rodriguez:
It sound like there's no reviews, so I guess I'll just go ahead with this way?

indirect:
👍🏻

indirect:
@bundlerbot r+

PR #7257, opened 2019-07-23
Get some more Windows specs passing

deivid-rodriguez:

The problem was that Windows build is still red.

My diagnosis was that some functionality is actually working, but specs need to be adapted, because some error messages include the platform on Windows.

My fix is to use `force_ruby_platform` for those specs.

I chose this fix because... I'm not sure, maybe it's better to adapt the error messages to include the platform when run on Windows.

### DIFF
--- a/spec/bundler/definition_spec.rb
+++ b/spec/bundler/definition_spec.rb
@@ -62,7 +62,7 @@ RSpec.describe Bundler::Definition do
             rack (1.0.0)
 
         PLATFORMS
-          ruby
+          #{lockfile_platforms}
 
         DEPENDENCIES
           foo!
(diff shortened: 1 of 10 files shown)


deivid-rodriguez:
Down to 336 failures now.

deivid-rodriguez:
I'd like some feedback specially on b5900e5. Not sure if it's better to force using ruby platform for these specs, so that the error messages are consistent, or to adapt the error message assertions to take into account that the platform is sometimes displayed.

Since this affects tests only though, I'll merge in a couple of days if there's no feedback.

deivid-rodriguez:
@bundlerbot r+

PR #7258, opened 2019-07-23
Restore the removal of the `--all` flag for `bundle cache` in Bundler 3

deivid-rodriguez:

The problem was that I accidentally restored this flag on Bundler 3 in #7249, but the plan still is to remove it.

My diagnosis was that I need to remove it again.

My fix is to remove it.

I chose this fix because it's the original plan and makes sense.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -411,9 +411,10 @@ module Bundler
     end
 
     desc "#{Bundler.feature_flag.bundler_3_mode? ? :cache : :package} [OPTIONS]", "Locks and then caches all of the gems into vendor/cache"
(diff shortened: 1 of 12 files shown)


deivid-rodriguez:
This is a regression of #7249, by the way.

deivid-rodriguez:
@indirect I tagged you because you reviewed the PR that caused the regression (and because your feedback is always great :)), sorry if I'm a bit overwhelming :sweat_smile:.

indirect:
👍🏻

@bundlerbot r+

indirect:
ugh wrong button

@bundlerbot r+

PR #7259, opened 2019-07-23
Relax bundler binstub

### LINKED ISSUE #7243: Bundler 2.x binstub incorrectly raises an error when run with a different bundler 2.x version
## Behavior

When using Bundler 2.0.2 with Ruby 2.5.5 and a Gemfile.lock that specifies a different 2.x version such as 2.0.1, bundler will generate a binstub that cannot be executed, instead it errors out and says that it expects to find the exact 2.0.1 version (even though 2.0.2 is fine and present).

```
Activating bundler (2.0.1) failed:
Could not find 'bundler' (2.0.1) required by your /Gemfile.lock.
To update to the latest version installed on your system, run `bundle update --bundler`.
To install the missing version, run `gem install bundler:2.0.1`
Checked in 'GEM_PATH=/root/.gem/ruby/2.5.0:/usr/local/lib/ruby/gems/2.5.0:/usr/local/bundle', execute `gem env` for more information

To install the version of bundler this project requires, run `gem install bundler -v '2.0.1'`
```

## Expected Behavior

Expected the binstub generated by bundler match version selector logic of rubygems and allow any major version of bundler to be able to execute a Gemfile.lock with a similar major version (i.e. 2.0.2 should be able to execute a Gemfile.lock with 2.0.1 or 2.0.3 or 2.1.0 etc.)

## Reproduction

In a fresh docker instance:

```
docker run -it --rm ruby:2.5 bash
```

Using bundler 2.0.2 with a Gemfile that contains a gem with an executable:

```
gem install bundler -v 2.0.2
cat << EOM > Gemfile
source 'https://rubygems.org'

gem 'explain_shell'
EOM

cat << EOM > Gemfile.lock
GEM
  remote: https://rubygems.org/
  specs:
    explain_shell (1.0.2)

PLATFORMS
  ruby

DEPENDENCIES
  explain_shell

BUNDLED WITH
   2.0.1
EOM
```

> Note that the BUNDLED WITH points at a different 2.x version, but it should be fine

Now install the gems and specify you want binstubs:

```
bundle install --without development:test --path vendor/bundle --binstubs vendor/bundle/bin -j4 --deployment
```

This will generate binstubs for all gems as well as for bundler, if you try to execute this binstub directly it fails:

```
$ vendor/bundle/bin/bundle install

Activating bundler (2.0.1) failed:
Could not find 'bundler' (2.0.1) required by your /Gemfile.lock.
To update to the latest version installed on your system, run `bundle update --bundler`.
To install the missing version, run `gem install bundler:2.0.1`
Checked in 'GEM_PATH=/root/.gem/ruby/2.5.0:/usr/local/lib/ruby/gems/2.5.0:/usr/local/bundle', execute `gem env` for more information

To install the version of bundler this project requires, run `gem install bundler -v '2.0.1'`
```

I originally found this in the Ruby buildpack, here's my original PR with the original Heroku specific repro (https://github.com/heroku/heroku-buildpack-ruby/pull/900)

deivid-rodriguez:

The problem was that the logic to select bundler versions was different when bundler is run from a rubygems binstub that when bundler is run from a bundler binstub.

My diagnosis was that we should unify the logic.

My fix is to use the same logic implemented in the rubygems version finder, namely, only fail if the major version of bundler does not match.

I chose this fix because it makes things consistent.

Fixes #7243.

### DIFF
--- a/lib/bundler/templates/Executable.bundler
+++ b/lib/bundler/templates/Executable.bundler
@@ -61,10 +61,9 @@ m = Module.new do
   end
 
   def bundler_version
-    @bundler_version ||= begin
+    @bundler_version ||=
       env_var_version || cli_arg_version ||
         lockfile_version || "#{Gem::Requirement.default}.a"
-    end
   end
 
   def load_bundler!
--- a/lib/bundler/templates/Executable.bundler
+++ b/lib/bundler/templates/Executable.bundler
@@ -87,7 +87,7 @@ m = Module.new do
       require "bundler/version"
     end
     return if require_error.nil? && Gem::Requirement.new(bundler_requirement).satisfied_by?(Gem::Version.new(Bundler::VERSION))
-    warn "Activating bundler (#{bundler_version}) failed:\n#{gem_error.message}\n\nTo install the version of bundler this project requires, run `gem install bundler -v '#{bundler_version}'`"
+    warn "Activating bundler (#{bundler_requirement}) failed:\n#{gem_error.message}\n\nTo install the version of bundler this project requires, run `gem install bundler -v '#{bundler_requirement}'`"
     exit 42
   end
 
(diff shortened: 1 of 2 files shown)


indirect:
👍🏻

@bundlerbot r+

PR #7261, opened 2019-07-23
Enable `Style/UnneededInterpolation` cop

deivid-rodriguez:

The problem was that sometimes we have some unneeded string interpolations in our code base.

My diagnosis was that we could remove them, because interpolating string variables seems weird and redundant.

My fix is to remove the unnecessary interpolations, and enable the corresponding rubocop cop.

I chose this fix because it fixes the current issues and it enforces the style rule from now on.


deivid-rodriguez:
@bundlerbot r=hsbt

Issue #7262, opened 2019-07-24
update's lock file order should same with install


SolaWing:
after bundle update, though the dependency not change, the path order is change. this make me have to run install to ensure the order consistency and avoid unneccesary commit
![图片](https://user-images.githubusercontent.com/3897953/61778856-22fb4080-ae32-11e9-8e9e-f0c8c56a5e87.png)

deivid-rodriguez:
This is unexpected, can you show the result of `bundle env`, and ideally a way to reproduce this?

SolaWing:
## Environment

```
Bundler       2.0.2
  Platforms   ruby, x86_64-darwin-18
Ruby          2.6.3p62 (2019-04-16 revision 67580) [x86_64-darwin18]
  Full Path   /Users/wang/.rbenv/versions/2.6.3/bin/ruby
  Config Dir  /Users/wang/.rbenv/versions/2.6.3/etc
RubyGems      3.0.3
  Gem Home    /Users/wang/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0
  Gem Path    /Users/wang/.gem/ruby/2.6.0:/Users/wang/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0
  User Path   /Users/wang/.gem/ruby/2.6.0
  Bin Dir     /Users/wang/.rbenv/versions/2.6.3/bin
Tools
  Git         2.20.1 (Apple Git-117)
  RVM         not installed
  rbenv       rbenv 1.1.2
  chruby      not installed
```

## Bundler Build Metadata

```
Built At          2019-06-13
Git SHA           496bca538
Released Version  true
```


I will try to reproduce it in a demo project

SolaWing:
Reproduce Steps:
bundle gem demo
cd demo
bundle gem aaa
bundle gem subgem
# in demo's Gemfile, specify
```
gem 'subgem', path: './subgem'
gem 'aaa', path: './aaa'
```
# modify gemspec to make bundle update happy.
bundle update
# commit lock file
bundle install
# bundle update and install have different order

here is the result files
[demo.zip](https://github.com/bundler/bundler/files/3426137/demo.zip)

deivid-rodriguez:
Reproduced, thanks! I'm going to have a look :+1:.

deivid-rodriguez:
[...]

PR #7263, opened 2019-07-24
Add `--[no-]git` option to `bundle gem`

deivid-rodriguez:

The problem was that while reproducing #7262, I actually wanted a single git repo, and every subgem source controlled under a single git repository. However, this cannot be done by default.

My diagnosis was that we could add a `--no-git` option to `bundle gem`, so that git repository initialization is skipped. I think this can be specially useful for monorepos.

My fix is to implement the flag.

I chose this fix because it's quite simple, and it solves my request.


deivid-rodriguez:
I rebased this PR.

I'll request a review from @hsbt and @indirect, because this is a new feature after all, so I'd like to make sure we all find this useful :smiley:.

deivid-rodriguez:
I'll assume no reviews mean no objections, and will go ahead with @simi's approval :)

@bundlerbot r=simi

deivid-rodriguez:
No idea why this wasn't closed automatically after merge... :/

PR #7264, opened 2019-07-24
Consistent lockfiles

### LINKED ISSUE #7262: update's lock file order should same with install

deivid-rodriguez:

The problem was that when a Gemfile would specify path sources with paths starting with "./", the generated lockfile would be inconsistent on `bundle install` and `bundle update`.

My diagnosis was that when running `bundle update`, the bundle would be unlocked, and bundler would order the path sources according to the path present in the Gemfile ("./aaa" for example). On the other hand, when running `bundle install`, bundler would read the relative paths from the lockfile, where they are normalized ("aaa" for example).

My fix is to normalize relative paths when creating the source list, so that "aaa" is always used over "./aaa".

I chose this fix because it fixes the problem and it's backwards compatible (as in, it doesn't break any specs that assert for specific lockfiles).

Fixes #7262.

### DIFF
--- a/lib/bundler/source/path.rb
+++ b/lib/bundler/source/path.rb
@@ -24,7 +24,12 @@ module Bundler
 
         if options["path"]
           @path = Pathname.new(options["path"])
-          @path = expand(@path) unless @path.relative?
+          expanded_path = expand(@path)
+          @path = if @path.relative?
+            expanded_path.relative_path_from(root_path.expand_path)
+          else
+            expanded_path
+          end
         end
 
         @name    = options["name"]
--- a/lib/bundler/source/path.rb
+++ b/lib/bundler/source/path.rb
@@ -20,7 +20,7 @@ module Bundler
         @allow_cached = false
         @allow_remote = false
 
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
Not sure who to request a review from here, specially since @segiddins is usually very busy. If I don't get any reviews in a few days, I'll go ahead and merge, since this fixes a bug and includes a regression spec.

@SolaWing Did you confirm that this fixes your issue, by the way?

SolaWing:
@deivid-rodriguez yes, I have patched the code and confirmed it

deivid-rodriguez:
@bundlerbot r+

deivid-rodriguez:
I really need to investigate this flaky... :angry: 

@bundlerbot retry

deivid-rodriguez:
@bundlerbot retry

deivid-rodriguez:
@bundlerbot r+

PR #7266, opened 2019-07-26
Bump to thor's master

deivid-rodriguez:

The problem is that we are currently cherry-picking some `thor` changes that are currently living on its master branch, but have not been released yet. The changes are about using `require_relative`and not relying on the LOAD_PATH. I cherry-picked them, because they were needed to fix some issues with `bundler` installed as a default gem.

Anyways, it's always better to depend on released versions of our dependencies.

My diagnosis was...

My fix is to do the same thing that I did for `Molinillo`in https://github.com/bundler/bundler/pull/7210, but for `thor`. I don't intend to merge it right now, but I want to make sure that we are ready to support a new version of `thor` if they released right now.

So, I can block this PR until a release is out.


deivid-rodriguez:
Hi again!

I would like to reconsider this. Unfortunately, we don't control the release schedule of thor, and vendoring our own copy gives us liberty of which patches to include or not. I think the current master of thor is more stable than the latest released version because it includes a bunch of fixes that we can actually use. I happen to need https://github.com/erikhuda/thor/commit/8c57f1dee1d6cb9d0071c793eae05278ece4d8c3 to get https://github.com/rubygems/rubygems/pull/2981 green, for example.

So, I propose to include thor's master in the next version of bundler.

deivid-rodriguez:
I fixed a few more issues in `thor` regarding Windows compatibility and I included them here, so things should be good now.

I'm merging this to unblock the next release!

@bundlerbot merge

PR #7268, opened 2019-07-29
Fix typo in binstubs documentation

scotttesler:

There is a typo in the [binstubs documentation](https://bundler.io/v2.0/man/bundle-binstubs.1.html).

It currently reads 

> This gives you a file that can **by** run directly

It should read 

> This gives you a file that can **be** run directly

Screen Shot 2019-07-29 at 3 14 24 PM

Fix the typo in the RONN file that is used to generate this documentation.


hsbt:
Thanks!

@bundlerbot r+

PR #7269, opened 2019-07-31
Bump rubocop to 0.74.0

deivid-rodriguez:

The problem was that we are using an outdated rubocop version.

My diagnosis was that we should upgrade, specially because the newest version fixes an issue that made us disable a cop under some situations.

My fix is to upgrade rubocop and remove the now unneeded exclusions.


deivid-rodriguez:
Thanks!

@bundlerbot r=hsbt

PR #7272, opened 2019-08-03
Revert "Remove version overriding stuff"

hsbt:

This reverts commit c675b59e8f7e3b4102cf0e83598b4df23691f209.

c675b59e8f7e3b4102cf0e83598b4df23691f209 was caused by https://travis-ci.org/rubygems/rubygems/jobs/566926815.

I only investigate with bisect.

I'm not sure why this change affects with test environment of rubygems with Travis.

@deivid-rodriguez Can we move this hack to only test suite?

### DIFF
--- a/lib/bundler/version.rb
+++ b/lib/bundler/version.rb
@@ -1,7 +1,23 @@
 # frozen_string_literal: false
 
 module Bundler
-  VERSION = "2.1.0.pre.1".freeze
+  # We're doing this because we might write tests that deal
+  # with other versions of bundler and we are unsure how to
+  # handle this better.
+  VERSION = "2.1.0.pre.1".freeze unless defined?(::Bundler::VERSION)
+
+  def self.overwrite_loaded_gem_version
+    begin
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
@hsbt If you give me some time, I will fix it the right way. This hack is bad because it allows bundler to jump from one copy of itself to another by "lying" about its own version.

I was working on a similar problem yesterday and created this branch: https://github.com/bundler/bundler/compare/remove_unneded_things?expand=1. Can you try the patches in there and see if they fix the issues?

deivid-rodriguez:
The branch should be green now, I'll propose a PR shortly and I expect it to fix the issues you're seeing in rubygems.

deivid-rodriguez:
May I close this PR? I really don't want to reintroduce this hack.

hsbt:
@deivid-rodriguez I agreed with your opinion and don't rush this.

Note: I really hope to integrate rubygems and bundler repos for fixing this situation soon 😇

PR #7277, opened 2019-08-05
Rerecord cassettes

deivid-rodriguez:

The problem was that we have not rerecorded cassettes in a while.

My diagnosis was that it's a good practice to rerecord cassettes once in a while.

My fix is to rerecord the cassettes.


deivid-rodriguez:
Straightforward, so going in!

@bundlerbot r+

PR #7278, opened 2019-08-05
Debug API fallback

deivid-rodriguez:

The problem was that I managed to find out that the current flaky failures are related to falling back sometimes to the old dependency API (for which we have no VCR recorded responses).

My diagnosis was that this is because our tests are sometimes messing up user's home making it not writable, and that makes the compact index API unusable.

My fix is to start by adding some debugging information to be able to confirm that this is the culprit. Once we do that, we can investigate how to fix it. The fix is composed of:

* Some debugging messages informing about when we fallback to the old API and why.
* Some puts during the offending realworld spec to dump to the screen all the information about the requests. 

I chose this fix because it should get us closer to fixing this.


deivid-rodriguez:
The temporary solution in #7276 seems to have worked, so I will try to fix the real culprit in this PR and not just introduce debugging information.

deivid-rodriguez:
This unstability will only trigger when TravisCI has network issues, and that's no longer the case. I can reopen this PR for debugging if it happens again in the future.

PR #7281, opened 2019-08-06
Bump test deps

deivid-rodriguez:

The problem was that some test dependencies, sinatra and rack in particular, were outdated.

My diagnosis was that we should use the latest versions.

My fix is to upgrade them. Both of them seem 100% compatible with respect to our test suite because there seems to be now failures, even though there was a note suggesting the opposite.


deivid-rodriguez:
Just improving the test suite, and I don't want to overwhelm reviewers, so going in!

@bundlerbot r+

deivid-rodriguez:
TravisCI was having network connectivity issues today, but they've reported them as fixed, so I'll retry.

@bundlerbot r+

PR #7283, opened 2019-08-07
Fix spec working by chance

deivid-rodriguez:

The problem was that this particular spec is like a nightmare that keeps on failing whenever we make changes to bundler, and we keep adding workarounds to it.

My diagnosis was that I should investigate closer what's going on with this spec, and fix it in a better way.

My fix is to fix several issues with the spec, and to essentially revert it back to the way it looked before https://github.com/bundler/bundler/commit/e63e844bc7444c6a489fcde0dc7011c6c4807edd. See each commit for more details.

I chose this fix because it should make it less painful to keep CI green as we make changes to bundler.


deivid-rodriguez:
@bundlerbot r+

PR #7284, opened 2019-08-08
Less UI silencing

deivid-rodriguez:

The problem was that that test suite prints some deprecation messages to the screen and I think that's unintented. See for example https://travis-ci.org/bundler/bundler/jobs/568969249.

My diagnosis was that bundler is using a different UI than the one intented sometimes. While fixing that, I also discovered that bundler is unintentionally silenced sometimes. For example, when you run `bundle exec`, `bundler` will automatically resolve dependencies in case you have never run `bundle install` before. In this case, I think it is expected that the "Resolving dependencies..." message is printed to the screen. 

My fix is to simplify UI handling while fixing these issues.

I chose this fix because it makes UI handling simpler.

### DIFF
--- a/lib/bundler/deployment.rb
+++ b/lib/bundler/deployment.rb
@@ -1,6 +1,11 @@
 # frozen_string_literal: true
 
 require_relative "shared_helpers"
+
+require_relative "../bundler"
+
+Bundler.ui = Bundler::UI::Shell.new
+
 Bundler::SharedHelpers.major_deprecation 2, "Bundler no longer integrates with " \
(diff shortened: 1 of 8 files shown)


deivid-rodriguez:
I had another look at this, and it sounds like some of this silencing is actually intended, as per @indirect's comment in https://github.com/bundler/bundler/pull/3722#discussion_r31890873.

So I'll change this PR to fix that.

deivid-rodriguez:
This PR was trying to do too much. Closing in favor of https://github.com/bundler/bundler/pull/7294.

PR #7285, opened 2019-08-08
Bump rspec dependency to 3.8

deivid-rodriguez:

The problem was that tests fail using rspec 3.7 or older.

My diagnosis was that we're using `config.bisect_runner` which is only available from 3.8.

My fix is to bump the requirement.


deivid-rodriguez:
@bundlerbot r+

PR #7288, opened 2019-08-10
Give a more clear error when user tries to `bundle open` a default gem

### LINKED ISSUE #4436: `bundle open` fails when the gem is a default gem

deivid-rodriguez:

The problem was that `bundle open` doesn't currently support opening default gems, but crashes with a very verbose and unclear error when the user tries to do it. 

My diagnosis was that we should give a more clear error in this case.

My fix is a rebased and improved version of #4882, that detects the situation and gives a better error. 

I chose this fix because it was already created and it's simpler than adding support for opening default gems, which is more problematic because they can't be pristine'd easily if edited and might not be in a writable location.

Closes #4882.
Fixes #4436.

### DIFF
--- a/lib/bundler/cli/open.rb
+++ b/lib/bundler/cli/open.rb
@@ -14,10 +14,10 @@ module Bundler
       editor = [ENV["BUNDLER_EDITOR"], ENV["VISUAL"], ENV["EDITOR"]].find {|e| !e.nil? && !e.empty? }
       return Bundler.ui.info("To open a bundled gem, set $EDITOR or $BUNDLER_EDITOR") unless editor
       return unless spec = Bundler::CLI::Common.select_spec(name, :regex_match)
-      path = spec.full_gem_path
       if spec.default_gem?
         Bundler.ui.info "Unable to open #{name} because it's a default gem, so the directory it would normally be installed to does not exist."
       else
+        path = spec.full_gem_path
         Dir.chdir(path) do
           command = Shellwords.split(editor) + [path]
           Bundler.with_original_env do
--- a/lib/bundler/cli/open.rb
+++ b/lib/bundler/cli/open.rb
@@ -15,11 +15,15 @@ module Bundler
       return Bundler.ui.info("To open a bundled gem, set $EDITOR or $BUNDLER_EDITOR") unless editor
       return unless spec = Bundler::CLI::Common.select_spec(name, :regex_match)
       path = spec.full_gem_path
-      Dir.chdir(path) do
-        command = Shellwords.split(editor) + [path]
-        Bundler.with_original_env do
-          system(*command)
-        end || Bundler.ui.info("Could not run '#{command.join(" ")}'")
+      if spec.default_gem?
+        Bundler.ui.info "Unable to open #{name} because it's a default gem, so the directory it would normally be installed to does not exist."
+      else
+        Dir.chdir(path) do
+          command = Shellwords.split(editor) + [path]
+          Bundler.with_original_env do
+            system(*command)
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
Straighforward, and the PR this comes from was already approved, pending specs to get green. The approach here is slightly different, but the result is the same, so going in!

@bundlerbot r+

PR #7290, opened 2019-08-10
add 'minitest-spec' test framework option

jethrodaniel:

The problem was that a minitest spec option wasn't present for `bundle gem --test`, and that I found myself manually converting the `--test=minitest` setup into a spec format.

Seems easy enough to change,, and doesn't add any dependencies.

My fix is to add the `--test=minitest-spec` option to the `gem` command.

I chose this fix because it was pretty easy and I was tired of setting up the minitest spec boilerplate.


jethrodaniel:
meant to create a draft, closing this

PR #7291, opened 2019-08-10
Add minitest-spec test option for gem command

jethrodaniel:

The problem was that a minitest spec option wasn't present for `bundle gem --test`, and that I found myself manually converting the `--test=minitest` setup into a spec format.

Seems easy enough to change, and doesn't add any dependencies.

My fix is to add the `--test=minitest-spec` option to the `gem` command.

I chose this fix because it was pretty easy and I was tired of setting up the minitest spec boilerplate.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -540,7 +540,7 @@ module Bundler
     method_option :ext, :type => :boolean, :default => false, :desc => "Generate the boilerplate for C extension code"
     method_option :mit, :type => :boolean, :desc => "Generate an MIT license file. Set a default with `bundle config set gem.mit true`."
     method_option :test, :type => :string, :lazy_default => "rspec", :aliases => "-t", :banner => "rspec",
-                         :desc => "Generate a test directory for your library, either rspec or minitest. Set a default with `bundle config set gem.test rspec`."
+                         :desc => "Generate tests for your library (rspec/minitest/minitest-spec). Set a default with `bundle config set gem.test rspec`."
     def gem(name)
     end
 
--- a/lib/bundler/cli/gem.rb
+++ b/lib/bundler/cli/gem.rb
@@ -12,6 +12,7 @@ module Bundler
     TEST_FRAMEWORK_VERSIONS = {
       "rspec" => "3.0",
       "minitest" => "5.0",
+      "minitest-spec" => "5.0",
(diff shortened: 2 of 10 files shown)


jethrodaniel:
Travis should be passing on the last commit, `83c3b1d25`

deivid-rodriguez:
Can you try cherry-picking https://github.com/bundler/bundler/pull/7293/commits/76f868d3a326d9fe5bbd75d764f37eeb04967c72, and pushing again to see if that fixes the CI?

jethrodaniel:
https://github.com/bundler/bundler/commit/76f868d3a326d9fe5bbd75d764f37eeb04967c72 fixed the month issue, but `man:check` still outputted some extra tabs (see [this travis build](https://travis-ci.org/bundler/bundler/jobs/570342495)).

So i ran ran the commands in an ubuntu 16.04 vm (like the travis builds), which worked fine, then copied over the generated man page. Not sure why it didn't work for me (on ubuntu 18.04).

deivid-rodriguez:
I think it might be because of the different `groff` versions generating different output. We should definitely make this better for contributors, thanks for taking the time to fix it :+1:

deivid-rodriguez:
I take it there's no strong opinion either in favour or against here?

hsbt:
Is there `minitest-spec` user in bundler/rubygems maintainers? If It's no, it's hard to maintain continuity for us.

indirect:
For the last 3 years or so, the previous core team held the position that `gem` can’t make everyone happy, so we did not accept new features, asking anyone wanting other defaults to use `ore` to create a gem template that met their needs instead.

I think I am the only core team member left from when that policy was put in place, and I personally could go either way. I would lean against accepting changes to the gem template because it isn’t core to bundler’s functionality, but if other maintainers are okay with adding to the maintenance burden I would not block.

deivid-rodriguez:
I don't personally use `minitest-spec`, but since we support the other two options, and I don't think there's many other test framework choices in the ruby community, it feels a bit like closing the gap for  `minitest-spec` users.

Anyways, I don't have a strong opinion either, so let's close this since the general sentiment between maintainers goes towards rejecting this.

deivid-rodriguez:
Sorry @jethrodaniel, thanks for working on this anyways!

Issue #7292, opened 2019-08-10
[feature-request] add minitest-spec test option for gem command


jethrodaniel:
I often use minitest spec with new gems, and find myself manually converting the `--test=minitest` setup from `bundle gem` into a spec format.

It'd be nice if we could use `bundle gem --test=minitest-spec` to do this automatically.

PR: https://github.com/bundler/bundler/pull/7291

jethrodaniel:
Added `[feature-request]` to the title because Github won't let me add the label.

deivid-rodriguez:
I'm good with adding this feature, personally.

deivid-rodriguez:
As per the related PR, this feature request was rejected some I'm closing this. I'm sorry :(

PR #7293, opened 2019-08-11
Fix `man:check`

deivid-rodriguez:

The problem was than `ronn` behaves weird sometimes and updates the month in pages to an old month. This can be seen in #7292, but I've seen it too myself. 

My fix is to explicitly pass the current date when generating man pages. This has the disadvantage that the man pages will become out of sync if they haven't been updated for more than a month, because the month will change.

But I'd like to introduce it until I think of a better alternative.


deivid-rodriguez:
I fixed up this PR with several change to try not to get in the middle of contributors too much with documentation rebuilding. Will merge tomorrow unless I get feedback earlier.

deivid-rodriguez:
@bundlerbot r+

PR #7294, opened 2019-08-11
Stop printing deprecation messages during specs

deivid-rodriguez:

The problem was that that test suite prints some deprecation messages to the screen and I think that's unintented. See for example https://travis-ci.org/bundler/bundler/jobs/568969249.

My diagnosis was that bundler is using a different UI than the one intented sometimes.

My fix is to make sure all deprecation messages go through bundler's UI.

I chose this fix because it stops deprecation messages from being interleaved with other output during the specs.


deivid-rodriguez:
@bundlerbot r+

PR #7295, opened 2019-08-12
Changelog for first 2.1.0 pre-release 

deivid-rodriguez:

The problem was that I want to release a first pre-release of bundler 2.1.0, but I don't have a changelog for it.

My diagnosis was that I needed to compile a changelog.

My fix is write the changelog.

I chose this fix because we used to have some automated change log handling, but I'm not sure how it work, so I instead just through each PR including user facing changes manually, with the aid of some helper tasks we have in our Rakefile.

This is still a WIP, because I might include some more PRs, and I also want to elaborate a bit on the deprecations that will get enabled, but it should be ready for an initial review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,10 @@
 ## 2.1.0.pre.1
 
 One of the biggest changes in bundler 2.1.0 is that deprecations for upcoming
-breaking changes in bundler 3 will be turned out by default. We do this to grab
-feedback and comunicate early to our users to kind of changes we're intending to
-make in bundler 3. See [#6965](https://github.com/bundler/bundler/pull/6965).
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
The initial draft is ready. I added an upgrading guide where I explain the deprecations, their motivations, and how to fix them.

Not sure whether that kind of document might be a better fit for a blog post or somewhere else.

What do you think?

hsbt:
It's okay to the initial pre-release. We can copy the guide to bundler blog after that.

deivid-rodriguez:
Thanks @hsbt! I'll update this with the latest PRs and get it merged later today! :+1:

deivid-rodriguez:
@bundlerbot r+

PR #7296, opened 2019-08-12
Improve exec'ing to the proper rubygems version

deivid-rodriguez:

The problem was that the `with_rubygems` script is hard to work with, because it does complicated ARGV manipulation before exec'ing.

My diagnosis was that we don't need it.

My fix is to implement the same thing in a simpler way.

I chose this fix because it simplifies things.


Rakefile
```diff
@@ -21,7 +21,7 @@ end
 
 desc "Run specs"
 task :spec do
-  sh("bin/rspec")
+  sh("bin/rspec --format progress")
```
hsbt: ❤️  I like this option.

deivid-rodriguez:
This PR is now ready. The ideas here are:

* Centralize `ENV["RGV"]` handling to a single place. This allows us, for example, to keep the list of tested rubies in a single place, and to not have to define special tasks for each TravisCI matrix entry.

* Allows to reproduce what TravisCI runs more easily. Currently, if you do `export RGV=master; bin/rake spec:travis`, it doesn't work. Now it does.

* Makes working with a local copy of rubygems more efficient by only performing git operations (like updating remotes or switching versions) when needed.

deivid-rodriguez:
@bundlerbot r+

PR #7297, opened 2019-08-13
Add initial Bundler::BuildMetadata Spec

kazu9su:

The problem was there is no spec of `Bundler::BuildMetadata`

My diagnosis was it's better to prepare spec

My fix is just adding spec file, has no impact on production codes.

I chose this fix because I saw the code coverage and there is enough space to contribute to this product.


hsbt:
@bundlerbot r+

PR #7302, opened 2019-08-15
Bundler displays a duplicate gem entries warning even if gems only appear once per group

davidstosik:

I have a Gemfile where some gems appear in two distinct groups. When I run Bundle, the following warning is displayed:

> Your Gemfile lists the gem #{gem} more than once.
> You should probably keep only one of them.
> Remove any duplicate entries and specify the gem **only once (per group)**.
> While it's not a problem now, it could cause errors if you change the version of one of them later.                                                           

Example Gemfile:
```rb
source "https://rubygems.org"

group :production do
end

group :development do
  gem "rake"
end

group :ci, optional: true do
  gem "rake"
end
```

I think the message is misleading, because it shows even if all gems are specified **only once per group**.

I removed the _(per group)_ mention from the warning message in order to prevent confusion.

I chose this fix because it was the simplest way to bring your attention to the problem.

If Bundler's real intent is to allow the same gem to appear in multiple groups, as long as it does not appear more than once per group, then my suggestion is wrong, and instead, the logic around the warning message needs to be fixed.

### DIFF
--- a/lib/bundler/dsl.rb
+++ b/lib/bundler/dsl.rb
@@ -128,7 +128,7 @@ module Bundler
         else
           Bundler.ui.warn "Your Gemfile lists the gem #{current.name} (#{current.requirement}) more than once.\n" \
                           "You should probably keep only one of them.\n" \
-                          "Remove any duplicate entries and specify the gem only once (per group).\n" \
+                          "Remove any duplicate entries and specify the gem only once.\n" \
                           "While it's not a problem now, it could cause errors if you change the version of one of them later."
         end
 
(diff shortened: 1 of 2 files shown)


davidstosik:
Note that while Bundler raises this warning, I can _fix_ my Gemfile this way:
```rb
source "https://rubygems.org"

group :production do
end

group :development, :ci do
  gem "rake"
end

group :ci, optional: true do
end
```

davidstosik:
♻️

deivid-rodriguez:
Yeah, my guess is that this is to not give the false impression that bundler supports separate resolutions per group, and that you can specify separate (possibly incompatible) dependencies for the same gem in different groups.

So, to me, this PR is good :+1:.

hsbt:
@bundlerbot r+

PR #7303, opened 2019-08-15
Refactor ruby core integration

deivid-rodriguez:

The problem was that sometimes we break specs when integrating bundler changes into core.

My diagnosis was that sometimes we use paths dependent on the structure of this repo, but that break under ruby-core's structure.

My fix is only some refactoring so that usage of structure independent helpers is encouraged. After this set of changes, if you grep the repo for `ruby_core?`, the only result will be `spec/support/path.rb`. That means that all logic dealing with repo folder structure lives in a single place. 

I chose this fix because it makes the integration in core cleaner.

### DIFF
--- a/lib/bundler/shared_helpers.rb
+++ b/lib/bundler/shared_helpers.rb
@@ -304,7 +304,7 @@ module Bundler
       exe_file = File.expand_path("../../../exe/bundle", __FILE__)
 
       # for Ruby core repository testing
-      exe_file = File.expand_path("../../../bin/bundle", __FILE__) unless File.exist?(exe_file)
+      exe_file = File.expand_path("../../../libexec/bundle", __FILE__) unless File.exist?(exe_file)
 
       # bundler is a default gem, exe path is separate
       exe_file = Bundler.rubygems.bin_path("bundler", "bundle", VERSION) unless File.exist?(exe_file)
--- a/spec/bundler/shared_helpers_spec.rb
+++ b/spec/bundler/shared_helpers_spec.rb
@@ -394,9 +394,8 @@ RSpec.describe Bundler::SharedHelpers do
 
       it "sets BUNDLE_BIN_PATH to the bundle executable file" do
         subject.set_bundle_environment
-        bundle_exe = ruby_core? ? "../../../../libexec/bundle" : "../../../exe/bundle"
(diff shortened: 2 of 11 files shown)


deivid-rodriguez:
@hsbt I think you will like this :)

```
(refactor_ruby_core_integration) deivid@pantani ~/Code/bundler $ rg -l ruby_core?
spec/support/path.rb
```

lib/bundler/shared_helpers.rb
```diff
@@ -305,7 +305,7 @@ def set_bundle_variables
       exe_file = File.expand_path("../../../exe/bundle", __FILE__)
 
       # for Ruby core repository testing
-      exe_file = File.expand_path("../../../bin/bundle", __FILE__) unless File.exist?(exe_file)
+      exe_file = File.expand_path("../../../libexec/bundle", __FILE__) unless File.exist?(exe_file)
```
deivid-rodriguez: I think there was a bug here, although this code is only for tests and the unit test was also checking for the wrong file, so it doesn't really make a difference. This is supposed to use the actual bundle bin, not a bundler binstub.

hsbt:
I resolved to conflict with master branch at https://github.com/bundler/bundler/pull/7303/commits/01145c420724400b0c85ca130717801c0f53028b

Can you squash or rebase it? Thanks.

deivid-rodriguez:
I rebased it, and also added 71a29e286a34204e808eaf5910b6b23e2a467e2b to catch up with a recent fix in core :+1:.

deivid-rodriguez:
@hsbt I don't think it's caused by this PR, but the check you linked to shows a single test failure (https://github.com/ruby/ruby/pull/2369/checks?check_run_id=195021311). I guess the CI status is currently being ignored.

deivid-rodriguez:
Let's do it!

@bundlerbot r=hsbt

hsbt:
Ah, Sorry. I have a mistake to confirm `test-bundler` result on ruby repo checks. We add `continue-on-error` same as allow_failure option recently.

deivid-rodriguez:
Hehe, syncronization. Should I cancel the bot, or this is ok to merge? :)

hsbt:
This is okay!

If I got the same error when I sync the ruby repo, I can fix it 💎

PR #7305, opened 2019-08-15
Simplify setting `Bunder.ui`

deivid-rodriguez:

The problem was that https://github.com/bundler/bundler/pull/7294 didn't actually stop the deprecations from being printed during specs.

My diagnosis was that we still need to simplify the UI in order to fix that.

My fix tries #7284 again. The reason I closed that one is because I thought it was changing the behaviour because `-rbundler/setup` would start printing deprecations and before they wouldn't be printed. Apparently deprecations/warnings would only be printed when bundler is run directly (see my comment in https://github.com/bundler/bundler/pull/7284#issuecomment-520169497).

However, I actually tried it and the behavior in master is to indeed _print_ deprecations in any case, so what this PR is actually doing is fixing the spec to represent the real behavior, not actually changing the behavior.

Given that, I think this refactoring is fine, and it indeed works for cleaning up the warnings from the specs.

### DIFF
--- a/lib/bundler/deployment.rb
+++ b/lib/bundler/deployment.rb
@@ -1,6 +1,11 @@
 # frozen_string_literal: true
 
 require_relative "shared_helpers"
+
+require_relative "../bundler"
+
+Bundler.ui = Bundler::UI::Shell.new
+
 Bundler::SharedHelpers.major_deprecation 2, "Bundler no longer integrates with " \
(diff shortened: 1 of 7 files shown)


deivid-rodriguez:
I'm not really sure whether this PR is correct or not, but the original motivation for it was to avoid some warnings being printed to screen during specs, and I'm fixing that at #7362, so closing.

PR #7306, opened 2019-08-16
Fixup #7297

hsbt:

`build_metadata_spec.rb` was inconsistency structure.

I missed checking the file location of #7297 

Move file under `spec/bundler`.


hsbt:
@bundlerbot r+

PR #7307, opened 2019-08-16
Deprecate `--path` flag to `bundle check` (and related fixes)

### LINKED ISSUE #7300: Deprecate `--path` option to `bundle check` and `bundle cache`
Just like we're deprecating `bundle install --path` and always rely on configuration, we should deprecate these flags too.

deivid-rodriguez:

The problem was that this flag should also be deprecated for these commands in preparation for bundler no longer remembering CLI flags.

My diagnosis was that, on bundler 3, the standard use case for `bundle check --path` will no longer work, namely:

```
 $ bundle check --path .bundle
The following gems are missing
 * rake (12.3.3)
Install missing gems with `bundle install`
$ bundle install
Fetching gem metadata from https://rubygems.org/.
Installing rake 12.3.3
Using bundler 2.1.0.pre.1
Bundled gems are installed into `./.bundle`
```

In the case of `bundle cache --path`, it has never really work as I expect
So we should deprecate `bundle check --path` and instead use whatever path is configured.

The cache of `bundle cache --path` is not as clear. Currently it does remember the flag for subsequent `bundle install` runs, but it does not change the location where subsequent `bundle cache` or `bundle install` runs save their cache. So I'm not sure how the current behavior is useful.

My fix is, pending further discussion on what the expected behavior for `bundle cache` is, to only deprecate `bundle check --path`, and in the case of `bundle cache` I only fixed the option description to not say it remembers the option in bundler 3.

Finally, I added a minor change in the deprecation message to recommend `bundle config set path <path>` instead of `bundle config path <path>`, because the latter is deprecated.

I chose this fix because it's the subset of all this that seemed clearly like the way to go.

Fixes #7300.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -789,7 +789,7 @@ module Bundler
       Bundler::SharedHelpers.major_deprecation 2,\
         "The `#{flag_name}` flag is deprecated because it relies on being " \
         "remembered accross bundler invokations, which bundler will no longer " \
-        "do in future versions. Instead please use `bundle config #{name} " \
+        "do in future versions. Instead please use `bundle config set #{name} " \
         "'#{value}'`, and stop using this flag"
     end
   end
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
@bundlerbot r+

PR #7308, opened 2019-08-16
Metric collection and reporting functionality for Bundler (refactored)

Akrabut:
**This work has been done as part of Google Summer of Code 2019**
This is a metric reporting system for Bundler, the backend API that will collect and instrument some of the metrics in rubygems.org is [here](https://github.com/rubygems/rubygems.org/pull/2075)
This PR is for the the purpose of getting feedback (and the metrics are still being sent to localhost for testing purposes).
This project has been mentored by @indirect 

* This is the refactored version of [this PR](https://github.com/bundler/bundler/pull/7298), where `Bundler::Metrics` is now a module rather than a class.

# **Added functionality:**
### For all commands- **the following are appended into ~/.bundle/metrics.yml:**
Command used
Options if specified
Time taken to fully execute for commands that don't kill Bundler
Time taken to start executing for commands that do
Timestamp
### For install/outdated/package/update/pristine- **record the following system info and send it to the server, along with all the metrics collected so far in metrics.yml, then truncate the file:**
The prior, and-
A randomized hex ID
Remote git repository (hashed)
git version
rvm version
rbenv version
chruby version
Host system details
Ruby version
Bundler version
Rubygems version
Ruby engine
CI’s
Extra user agent strings
Gemfile gem count
Actually installed gem count
git gem count
Path gem count
Gem source count
List of gem sources (hashed)
### When gems are downloaded- **also appends:**
Gem download time
### When gems are installed- **also appends:**
Gem installation time
### When a gemfile is resolved- **also appends:**
Gemfile resolve time.
### When an installation fails- **appends the regular command metrics, and:**
Name and version of the gem that failed to install

### Opt out:
`bundle config set disable_metrics true` allows the user to opt out of metrics collection
Deletes the metrics.yml file and doesn't collect any farther metrics.
Also skips all added metric functionality when running Bundler.
### Opt in:
`bundle config set disable_metrics false` allows the user to opt back in to metric collection.
### Default: 
When no `disable_metrics` value is found in the global config file (or the file doesn't exist), Bundler behaves as if `disable_metrics` is false.

#### All added functionality has passed (including the added tests) the tests locally on my system.


hsbt:
see https://github.com/bundler/bundler/pull/7298#issuecomment-521976632

PR #7310, opened 2019-08-16
Freeze time to avoid failures at midnight

kazu9su:

The problem was it will potentially fail if tests are executed around midnight

My diagnosis was it should freeze time when executing spec

My fix is mocking `Time.now` when executing `Bundler::BuildMetadata` specs

I chose this fix because it doesn't have effects on other specs.

### DIFF
--- a/spec/bundler/build_metadata_spec.rb
+++ b/spec/bundler/build_metadata_spec.rb
@@ -4,9 +4,14 @@ require "bundler"
 require "bundler/build_metadata"
 
 RSpec.describe Bundler::BuildMetadata do
+  before do
+    allow(Time).to receive(:now).and_return(Time.at(0))
+    Bundler::BuildMetadata.instance_variable_set(:@built_at, nil)
+  end
+
   describe "#built_at" do
     it "returns %Y-%m-%d formatted time" do
-      expect(Bundler::BuildMetadata.built_at).to eq Time.now.strftime("%Y-%m-%d")
+      expect(Bundler::BuildMetadata.built_at).to eq "1970-01-01"
     end
   end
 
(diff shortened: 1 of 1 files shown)


spec/bundler/build_metadata_spec.rb
```diff
@@ -36,7 +40,7 @@
     subject { Bundler::BuildMetadata.to_h }
 
     it "returns a hash includes Built At, Git SHA and Released Version" do
-      expect(subject["Built At"]).to eq Time.now.strftime("%Y-%m-%d")
+      expect(subject["Built At"]).to eq Time.now.strftime("1970-01-01")
```
znz: `Time.now.strftime("1970-01-01")` → `"1970-01-01"` ?

deivid-rodriguez:
Hah! It's around midnight where I am, and I just run into these failures :)

Initial commit seemed alright for our CI, so maybe we can revert the second version?

kazu9su:
@deivid-rodriguez 
Thanks for commenting. Do you mean it's better to close this PR or revert to [954f4c5](https://github.com/bundler/bundler/commit/954f4c5da9b5012bee66ddf0adbd5961393e88b6)?

deivid-rodriguez:
Nono, I mean revert to whatever the initial version was (the one which passed CI). This PR is great, I hate those midnight only test failures... :sweat_smile:

deivid-rodriguez:
Sorry @kazu9su, I'm too bad at explaining myself. I didn't mean reverting the original PR adding the tests, I mean reverting to the initial version of this PR to fixed the tests (I think it is https://github.com/bundler/bundler/commit/954f4c5da9b5012bee66ddf0adbd5961393e88b6 as you mentioned).

kazu9su:
@deivid-rodriguez Don't apologize it! But actually, this change has a problem. Looks on some environments on Travis CI, Rspec::Mock method doesn't work. It means, `Time.now` returns actual `now` even specify `allow(Time).to receive(:now)...` on the spec file.
https://travis-ci.org/bundler/bundler/jobs/572757660

Looks it also happens on [954f4c5](https://github.com/bundler/bundler/commit/954f4c5da9b5012bee66ddf0adbd5961393e88b6) so I'm investigating but still couldn't figure out the cause...

deivid-rodriguez:
@kazu9su Maybe because a previous test uses this method, and thus get memoized and `Time.now` no longer gets called? You can try resetting the instance variable before the test to see if that's the problem.

kazu9su:
@deivid-rodriguez you're right... I forgot to initialize the instance variable. Thanks for specifying that!

deivid-rodriguez:
Thank YOU! Let's get this in.

@bundlerbot r+

PR #7311, opened 2019-08-17
Use the latest rubygems 3.0.6 in CI

deivid-rodriguez:

The problem was that [rubygems 3.0.6 was released](https://blog.rubygems.org/2019/08/16/3.0.6-released.html), and we haven't yet tested against it.

My fix is to use the new rubygems in CI.


deivid-rodriguez:
@bundlerbot r+

PR #7312, opened 2019-08-17
Remove old rubies stuff no longer needed

deivid-rodriguez:

The problem was that was still some code specific to old rubies we no longer support.

My fix is to remove that code.


deivid-rodriguez:
@bundlerbot r=hsbt

PR #7313, opened 2019-08-18
Fix a couple of typos

deivid-rodriguez:

The problem was typos.

My fix is to correct typos.


deivid-rodriguez:
@bundlerbot r+

PR #7317, opened 2019-08-19
Parallelize test suite (Take 2)

hsbt:

This branch reduced test time of bundler.

#7232 has a conflict with the current master branch.

I picked a commit for parallelized tests from #7232 

Fixed https://github.com/bundler/bundler/pull/7232

### DIFF
--- a/.rspec_parallel
+++ b/.rspec_parallel
@@ -1,0 +1,4 @@
+--format progress
+--format ParallelTests::RSpec::RuntimeLogger --out tmp/parallel_runtime_rspec.log
+--require spec_helper
+--require support/parallel.rb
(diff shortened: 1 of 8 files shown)


deivid-rodriguez:
Thanks for taking this! (I'll close the initial PR)

So, the reason I didn't follow up here and that I think this might cause unstability because our specs `Dir.chdir` back and forth and that's thread-global, so I saw some failures about that. And I don't know a good solution for that... Can you restart this two or three times to see if it reproduces?

hsbt:
> Can you restart this two or three times to see if it reproduces?

Fmm. I never faced these issues on my macOS.

hsbt:
Thanks!

>so I guess we can try this, and revert if it turns out to be problematic?

Agreed! we can revert this easily until the final release of the next version.

hsbt:
@bundlerbot r+

PR #7321, opened 2019-08-21
Don't (yet) enable parallelization in CI

deivid-rodriguez:

The problem was that unfortunately #7317 broken rubygems test suite. See https://github.com/rubygems/rubygems/pull/2888.

My diagnosis was that it has some issues.

My fix is to not (yet) use it in CI, so we can keep using it locally and investigating these issues without interfering with our CI. In a while we reevaluate how it is working a consider enabling in CI.


bronzdoc:
@bundlerbot r+

deivid-rodriguez:
@hsbt Sorry I had to partially revert this. I'm now working on reenabling it!

hsbt:
No problem!

PR #7329, opened 2019-08-26
Undo auto multiplatform

### LINKED ISSUE #7315: Regression related to platforms
Against current ransack's master (https://github.com/activerecord-hackery/ransack/commit/4e148cc0bfc54a95ed64c76e1b92ee43c8f1d1a4), on a fresh clone, using ruby 2.6 and bundler's master.

```
$ RAILS=v6.0.0 bundle _2.1.0.pre.1_ install
Fetching https://github.com/rails/rails.git
Fetching gem metadata from https://rubygems.org/............
Fetching gem metadata from https://rubygems.org/............
Resolving dependencies...
Bundler found conflicting requirements for the Ruby version:
  In Gemfile:
    Ruby

    mysql2 (~> 0.5.2) was resolved to 0.5.2, which depends on
      Ruby (< 2.6) x64-mingw32

    mysql2 (~> 0.5.2) was resolved to 0.5.2, which depends on
      Ruby (< 2.6) x86-mingw32

    mysql2 (~> 0.5.2) was resolved to 0.5.2, which depends on
      Ruby (< 2.6) x86-mswin32

Ruby (< 2.6), which is required by gem 'mysql2 (~> 0.5.2)', is not available in the local ruby installation
```

It's incorrectly taking in consideration only the platform specific gems (even if none of them is my platform), and ignoring the "ruby" version, which does not have an upper bound on the required ruby.

Using bundler 2.0.2 bundles just fine.

It might be related to https://github.com/bundler/bundler/pull/7215, I'll look into it tommorrow.

deivid-rodriguez:

The problem was that since #7215, more Gemfiles are going to fail resolution because of the current issues with multiplatform support, and that we'll be pushing current multiplatform problems into all users because bundler will resolve and lock all platforms included on a Gemfile by default. 

My diagnosis was that we probably need better multiplatform support before we start resolving all platforms by default.

My fix is to revert the relevant commits from #7215. I'll try to revisit in the future.

I chose this fix because it goes back to how things were before.

Closes #7315.

### DIFF
--- a/lib/bundler/cli/outdated.rb
+++ b/lib/bundler/cli/outdated.rb
@@ -169,8 +169,6 @@ module Bundler
     end
 
     def retrieve_active_spec(definition, current_spec)
-      return unless current_spec.match_platform(Bundler.local_platform)
-
       if strict
         active_spec = definition.find_resolved_spec(current_spec)
       else
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -518,6 +518,10 @@ module Bundler
       raise InvalidOption, "Unable to remove the platform `#{platform}` since the only platforms are #{@platforms.join ", "}"
     end
 
+    def add_current_platform
+      current_platforms.each {|platform| add_platform(platform) }
+    end
+
     def find_resolved_spec(current_spec)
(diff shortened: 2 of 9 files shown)


deivid-rodriguez:
@bundlerbot r+

Thanks!

PR #7332, opened 2019-08-27
Fix version number in release playbook

deivid-rodriguez:

The problem was that I'm about to release bundler, but I'm unsure which one is the right version number.

My diagnosis was that the release playbook is probably wrong, because looking at https://rubygems.org/gems/bundler/versions, we don't really ship `.pre.0`. versions.

My fix is to document `.pre.1` as the first prerelease on a minor series, since that's what we usually do.


deivid-rodriguez:
@bundlerbot r+

PR #7333, opened 2019-08-28
Release task improvements

deivid-rodriguez:

The problem was several issues I had when releasing bundler:

* The prerequisites can fail, but the gem is still pushed to rubygems.
* Some error messages could be better.
* The github release is a postrequisite not a prerequisite for the release task.

My diagnosis was that we need to make every prerequisite a prerequisite to the `rubygems:push` task to be really sure the gem is not pushed until every thing else is satisfied.

My fix implements that idea, plus other few improvements to the release task.


deivid-rodriguez:
I'll merge this, so I can use it for the next pre.

@bundlerbot r+

PR #7334, opened 2019-08-28
Test against latest rubies

deivid-rodriguez:

The problem was that patch level rubies were release, so we should test them.

My fix is to start using them in CI.


deivid-rodriguez:
I took the chance to regenerate VCR cassettes using the newest ruby. I think it's a reasonable frequency to regenerate the cassettes when new rubies are released.

deivid-rodriguez:
@bundlerbot r+

PR #7337, opened 2019-08-30
Fix a couple of bundler issues with keyword argument separation

jeremyevans:

`make test-bundler` in Ruby repository failed with recent changes to separate keyword arguments from positional arguments (see https://bugs.ruby-lang.org/issues/14183).

Bundler's specs check for lack of warnings, and the new changes cause warnings that cause Bundler's specs to fail.

In one case, I fix the code to work with keyword arguments and positional arguments in positional arguments.  In the other case, I fix the spec to ignore the keyword argument separation warnings.

I fixed this upstream in Ruby already (https://github.com/ruby/ruby/commit/b5b3afadfab4072f55320075ccac6afe333a140c). So this is a request to make bundler use the same patch.  You can certainly fix it differently, and there are more related issues to fix I think (they just don't cause test failures yet).  The issues definitely need to be fixed if you want bundler's specs to run correctly in Ruby 2.7.

Because it was the least amount of effort.

### DIFF
--- a/lib/bundler/vendor/thor/lib/thor.rb
+++ b/lib/bundler/vendor/thor/lib/thor.rb
@@ -90,9 +90,14 @@ class Bundler::Thor
     # ==== Parameters
     # Hash[String|Array => Symbol]:: Maps the string or the strings in the array to the given command.
     #
-    def map(mappings = nil)
+    def map(mappings = nil, **kw)
       @map ||= from_superclass(:map, {})
 
+      if mappings && !kw.empty?
+        mappings = kw.merge!(mappings)
+      else
+        mappings ||= kw
+      end
       if mappings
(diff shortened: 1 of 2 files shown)


hsbt:
@jeremy Can you update rubocop warning? https://travis-ci.org/bundler/bundler/jobs/578964090#L618

jeremyevans:
@hsbt Sure, I forced push a commit that should fix it.

hsbt:
@jeremy Thanks, We need to send the same patch to https://github.com/erikhuda/thor as the upstream.

hsbt:
@bundlerbot r+

deivid-rodriguez:
@bundlerbot retry

PR #7339, opened 2019-09-01
Bump man documentation's month

deivid-rodriguez:

The problem was that CI is green because `rake man:check` is failing.

My diagnosis was is that this is a gotcha of our current documentation building workflow.

I want to address this in the future but for now the easiest for me is to bump the documentation's date once a month.

I chose this fix because it's the simplest way to get CI green for me now.


deivid-rodriguez:
@bundlerbot r+

PR #7342, opened 2019-09-04
Start using automatiek 0.3.0

deivid-rodriguez:

The problem was that I added a feature to automatiek to vendor subdependencies, but we were not using it because it had not yet been released.

My fix is to start using the feature now that it has been released.

I chose this fix because it allows us to manage the `net-http-persistent` dependency in an easier way.


deivid-rodriguez:
@bundlerbot r+

PR #7350, opened 2019-09-09
Simplify `release:patch` task

deivid-rodriguez:

The problem was that I'm using a slightly different workflow for releasing, but the current rake task does not work my way.

My diagnosis was that previously the release task would push the changelog directly to the release branch and master. The way I'm doing it, I open a regular PR which can be reviewed and merged through bundlerbot just like any other PR.

My fix is to adapt the task to the workflow I'm using, and change it to only cherry-pick all the milestoned changes to the stable branch. I renamed it to `release:prepare_patch`.

I chose this fix because it allows the release process itself to be reviewed, and it would allow us to protect the master and stable branches if we wanted to.


deivid-rodriguez:
@bundlerbot r+

PR #7353, opened 2019-09-11
require `rubygems`, not `ubygems` in example

crazymykl:
The latter does not work on Ruby 2.6.3.

The problem was a broken example in the docs.

Ibid.

My fix was to write what was likely intended.

I chose this fix because the solution space is pretty small.


hsbt:
@bundlerbot r+

PR #7355, opened 2019-09-12
Restore version bumping to `release:prepare_patch`

deivid-rodriguez:

The problem was that I removed too much code in #7350.

My diagnosis was that since I removed the auto-bumping `version.rb`, the task no longer works before it tries to cherry-pick changes for the _current_, not the _to be released_ version, so I fails because it finds no open PR for that milestone (because it's already been released).

My fix is to restore the code that automatically bumps the version to the next patch level version, and commits that to the release branch.

I chose this fix because it works for me.

### DIFF
--- a/doc/playbooks/RELEASING.md
+++ b/doc/playbooks/RELEASING.md
@@ -48,7 +48,8 @@ using:
 $ git cherry-pick -m 1 dd6aef9
 ```
 
-The `rake release:patch` command will automatically handle cherry-picking, and is further detailed below.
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
The `release:prepare_patch` task can be run from master and it would work anyways, so there's no need to backport it. But I'll do it anyways because I see no harm on doing it.

I'll leave this opened for one more day anyways in case I get feedback for the new release workflow.

deivid-rodriguez:
@bundlerbot r+

PR #7358, opened 2019-09-15
More release fixes

deivid-rodriguez:

I just released [2.1.0.pre.2](https://rubygems.org/gems/bundler/versions/2.1.0.pre.2). It went smoother than 2.1.0.pre.1, but I still got an issue where I unintentionally pushed a release named "v" to github releases.

My diagnosis was that the release task by default should use the version in the gemspec for the `release:github` tasks.

My fix adds that fallback.

I chose this fix because it does what the user expects when the task is not passed any arguments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 2.1.0.pre.2
+## 2.1.0.pre.2 (September 15, 2019)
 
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
@bundlerbot r+

PR #7359, opened 2019-09-16
Spec simplifications

deivid-rodriguez:

The problem was that while working on #7296 and investigating why CI was failing there, I ended up adding several simplifications that are not necessarily related to that PR. In order to make that PR more digestable, I extracted those simplifications and improvements to this separate PR.

I chose this fix because it makes our dev and test environment better and simpler with some little tweaks.


Rakefile
```diff
@@ -21,7 +21,7 @@ end
 
 desc "Run specs"
 task :spec do
-  sh("bin/rspec")
+  sh("bin/rspec --format progress")
```
hsbt: 👍

deivid-rodriguez:
@bundlerbot r=indirect

PR #7360, opened 2019-09-23
Consistenly use the default RSpec formatter

deivid-rodriguez:

The problem was that our default RSpec formatter is `--documentation`, but we're overriding it almost everywhere to use `--progress`.

My diagnosis was that we should probably use the default formatter, since I think we all prefer it.

My fix is to do that.


deivid-rodriguez:
@bundlerbot r=hsbt

PR #7361, opened 2019-09-23
Add caller information to some deprecation messages

deivid-rodriguez:

The problem was that some deprecations mention usage of deprecated methods, but include no information about where the deprecated method is being called.

My diagnosis was that we should include this information with this kind of deprecation messages, to make migration easier.

My fix is to add an optional argument to the `major_deprecation` method, to print the location of the deprecated caller.

I chose this fix because it required little code changes.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -116,7 +116,7 @@ module Bundler
     end
 
     def environment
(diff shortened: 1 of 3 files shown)


indirect:
What a good idea 😅

indirect:
@bundlerbot r+

PR #7362, opened 2019-09-23
Fix spec warnings

deivid-rodriguez:

The problem was that our specs print some warnings to screen.

My diagnosis was that specs should have a clean output.

My fix is to remove the warnings.


deivid-rodriguez:
@bundlerbot r+

PR #7364, opened 2019-09-24
Follow up to `ENV["RGV"]` handling refactor

### LINKED ISSUE #7363: RubyGems test was broken with bundler master
I'm not sure why rubygems tests were failed with bundler master.

https://travis-ci.org/rubygems/rubygems/jobs/588913478

@deivid-rodriguez Do you have any idea related to this?

deivid-rodriguez:

The problem was that #7296 broke rubygems CI.

My diagnosis was that the handling of `ENV["RGV"]` when it contains the path of a rubygems checkout instead of a rubygems version (like it happens in the rubygems repo) was incorrect.

My fix is to correct the behaviour. I was able to reproduce the CI failures in https://github.com/rubygems/rubygems/pull/2924, and verify they are be fixed by this PR by patching the vendored bundler locally.

Fixes #7363.

### DIFF
--- a/spec/support/rubygems_version_manager.rb
+++ b/spec/support/rubygems_version_manager.rb
@@ -110,6 +110,8 @@ private
   end
 
   def resolve_target_gem_version
+    return local_copy_version if env_version_is_path?
+
     return @env_version[1..-1] if @env_version.match(/^v/)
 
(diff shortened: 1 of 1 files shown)


hsbt:
@bundlerbot r+

PR #4474, opened 2016-04-28
Improve readability of outdated

tlynam:
Haven't fixed tests since wanted buy-in.

Before:

```
Outdated gems included in the bundle:
  * active_model_serializers (newest 0.9.5, installed 0.8.3, requested ~> 0.8.0) in group "default"
  * activerecord-import (newest 0.13.0, installed 0.10.0) in group "default"
  * addressable (newest 2.4.0, installed 2.3.5, requested = 2.3.5) in group "default"
  * airbrake (newest 5.2.3, installed 5.1.0) in group "default"
  * bootstrap-sass (newest 3.3.6, installed 3.3.5.1) in group "default"
  * capistrano (newest 3.5.0, installed 3.4.0, requested ~> 3.4) in group "development"
  * capistrano-rails (newest 1.1.6, installed 1.1.3, requested ~> 1.1.1) in group "development"
```

After:

```
--pretty
Outdated gems included in the bundle:
Gem Name                New        Installed  Requested  Groups
airbrake                5.3.0      5.0.3                 default
better_errors           2.1.1      1.0.1      ~> 1.0.1   development
bootstrap-sass          3.3.6      3.3.0.1    ~> 3.3.0.1 default
factory_girl_rails      4.7.0      4.4.1                 test, development

--pretty --verbose
Outdated gems included in the bundle:
Gem Name                New        Installed  Requested  Groups            Load Path
airbrake                5.3.0      5.0.3                 default           /Users/todd/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/specifications/airbrake-5.3.0.gemspec
better_errors           2.1.1      1.0.1      ~> 1.0.1   development       /Users/todd/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/specifications/better_errors-2.1.1.gemspec
bootstrap-sass          3.3.6      3.3.0.1    ~> 3.3.0.1 default           /Users/todd/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/specifications/bootstrap-sass-3.3.6.gemspec
factory_girl            4.7.0      4.4.0                                   /Users/todd/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/specifications/factory_girl-4.7.0.gemspec
factory_girl_rails      4.7.0      4.4.1                 test, development /Users/todd/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/specifications/factory_girl_rails-4.7.0.gemspec
```

PR #7365, opened 2019-09-25
Refactor `bundle outdated`

deivid-rodriguez:

The problem was that while working on a better `bundle outdated` format (essentially getting #4474 ready to merge), I noticed a bunch of logic in `bundle outdated` that's overly complicated and not necessary.

My diagnosis was that I should simplify the implementation before changing the output.

My fix is to add my improvements as a separate PR. Afterwards I'll rebase and get #4474 ready.

I chose this fix because including this improvements with #4474 made that PR too big and hard to understand.

### DIFF
--- a/lib/bundler/cli/outdated.rb
+++ b/lib/bundler/cli/outdated.rb
@@ -109,8 +109,7 @@ module Bundler
         end
 
         if options_include_groups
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
@bundlerbot r+

PR #7368, opened 2019-09-27
Bump version to latest pre-release

deivid-rodriguez:

The problem was that I forgot to merge the version increase back to master after release.

My fix is to bump it.


deivid-rodriguez:
@bundlerbot r+

PR #7370, opened 2019-09-27
Fix `bunde outdated` regression with groupless dependencies

deivid-rodriguez:

The problem was that the refactoring in #7365 introduced a regression involving groupless dependencies.

My diagnosis was that `nil` groups were not being properly handled when sorting groups by name, and when this kind of outdated deps would be involved, `bundle outdated` would crash with a backtrace like the following:

```
NoMethodError: undefined method `split' for nil:NilClass
  /home/deivid/Code/bundler/lib/bundler/cli/outdated.rb:105:in `block in run'
  /home/deivid/Code/bundler/lib/bundler/cli/outdated.rb:104:in `each'
  /home/deivid/Code/bundler/lib/bundler/cli/outdated.rb:104:in `run'
  /home/deivid/Code/bundler/lib/bundler/cli.rb:412:in `outdated'
  /home/deivid/Code/bundler/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
  /home/deivid/Code/bundler/lib/bundler/vendor/thor/lib/thor/invocation.rb:126:in `invoke_command'
  /home/deivid/Code/bundler/lib/bundler/vendor/thor/lib/thor.rb:392:in `dispatch'
  /home/deivid/Code/bundler/lib/bundler/cli.rb:26:in `dispatch'
  /home/deivid/Code/bundler/lib/bundler/vendor/thor/lib/thor/base.rb:466:in `start'
  /home/deivid/Code/bundler/lib/bundler/cli.rb:17:in `start'
  /home/deivid/Code/bundler/exe/bundle:46:in `block in <main>'
  /home/deivid/Code/bundler/lib/bundler/friendly_errors.rb:123:in `with_friendly_errors'
  /home/deivid/Code/bundler/exe/bundle:34:in `<main>'
```

My fix is to consider `nil` groups as empty instead, so that sorting them works.

I chose this fix because it's reasonable simple.

### DIFF
--- a/lib/bundler/cli/outdated.rb
+++ b/lib/bundler/cli/outdated.rb
@@ -82,7 +82,7 @@ module Bundler
         next unless gem_outdated || (current_spec.git_version != active_spec.git_version)
 
         dependency = current_dependencies[current_spec.name]
-        groups = nil
+        groups = ""
         if dependency && !options[:parseable]
           groups = dependency.groups.join(", ")
         end
@@ -134,10 +134,10 @@ module Bundler
     end
 
     def header_group_message(groups)
-      if groups
-        "===== #{groups_text("Group", groups)} ====="
-      else
+      if groups.empty?
         "===== Without group ====="
+      else
+        "===== #{groups_text("Group", groups)} ====="
       end
     end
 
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
@bundlerbot r+

PR #7372, opened 2019-10-01
October manpages

deivid-rodriguez:

The problem was that outdated manpages.

My fix is to update them.


deivid-rodriguez:
That'd be super awesome.

deivid-rodriguez:
@bundlerbot r=hsbt

PR #7373, opened 2019-10-02
Refactor new gem specs

deivid-rodriguez:

The problem was that I rebased #6455 and I had trouble getting the specs to pass.

My diagnosis was that the new gemspecs could use a refactor.

My fix is to refactor these specs so that they are easier to read and modify.


deivid-rodriguez:
@bundlerbot r+

PR #7375, opened 2019-10-04
Bump vendored fileutils to 1.3.0

deivid-rodriguez:

The problem was that we were not using a released version of fileutils.

My diagnosis was that we can upgrade, because 1.3.0 has been released.

My fix is to upgrade.


deivid-rodriguez:
@bundlerbot r+

PR #7376, opened 2019-10-06
bundle info prints gem metadata

orien:

Rubygems has supported [gemspec metadata](https://guides.rubygems.org/specification-reference/#metadata) since version 2.0. I think it'd be great to display this information in response to the `bundle info <gem>` command.

Added a check for `metadata`, outputting this extra information if it has been included with the gem release. For example:

```
> bundle info rspec-mocks
  * rspec-mocks (3.8.1)
	Summary: rspec-mocks-3.8.1
	Homepage: https://github.com/rspec/rspec-mocks
	Documentation: https://rspec.info/documentation/
	Source Code: https://github.com/rspec/rspec-mocks
	Changelog: https://github.com/rspec/rspec-mocks/blob/v3.8.1/Changelog.md
	Bug Tracker: https://github.com/rspec/rspec-mocks/issues
	Mailing List: https://groups.google.com/forum/#!forum/rspec
	Path: /opt/asdf/installs/ruby/2.6.5/lib/ruby/gems/2.6.0/gems/rspec-mocks-3.8.1
```

For better readability, we could align the values. However, this would mean a change to the existing info property lines (`Summary`, `Homepage`, `Path` and `Default`). Which would potentially break existing scripts that are coupled to the precise output.

### DIFF
--- a/lib/bundler/cli/info.rb
+++ b/lib/bundler/cli/info.rb
@@ -50,7 +50,7 @@ module Bundler
     end
 
     def print_gem_info(spec)
-      metadata = spec.respond_to?(:metadata) ? spec.metadata : {}
+      metadata = spec.metadata
       gem_info = String.new
       gem_info << "  * #{spec.name} (#{spec.version}#{spec.git_version})\n"
       gem_info << "\tSummary: #{spec.summary}\n" if spec.summary
--- a/lib/bundler/cli/info.rb
+++ b/lib/bundler/cli/info.rb
@@ -50,10 +50,17 @@ module Bundler
     end
 
     def print_gem_info(spec)
+      metadata = spec.respond_to?(:metadata) ? spec.metadata : {}
       gem_info = String.new
       gem_info << "  * #{spec.name} (#{spec.version}#{spec.git_version})\n"
       gem_info << "\tSummary: #{spec.summary}\n" if spec.summary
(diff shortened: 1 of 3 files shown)


lib/bundler/cli/info.rb
```diff
@@ -50,10 +50,17 @@ def print_gem_path(spec)
     end
 
     def print_gem_info(spec)
+      metadata = spec.respond_to?(:metadata) ? spec.metadata : {}
```
deivid-rodriguez: Since our `required_rubygems_version` is `>= 2.5.2`, maybe we don't need this `respond_to?`?

orien: That makes sense to me. I was a little defensive when I wrote this, as I'm not too familiar with the codebase.

The `spec` here is an instance of `Bundler::StubSpecification`. So the `metadata` method comes in via method missing:

https://github.com/bundler/bundler/blob/fcb0fef62096bd8d64383f5bbfa584ce77f40fd8/lib/bundler/remote_specification.rb#L99-L107

`_remote_specification` is a `Gem::Specification` in all the experiments I've performed locally. I'm unaware of any scenario where it would be an instance of a different class.

I'll remove the check.

deivid-rodriguez:
Let's do it!

@bundlerbot r+

PR #7379, opened 2019-10-14
Switch to double quotes to match rest of file

jonathanhefner:

It is common to prefer double quotes, unless single quotes are needed.  It is also generally preferred to use a consistent quote style (whether double quotes or single quotes).

Switch the anomalous single quotes to double quotes.


deivid-rodriguez:
@bundlerbot r+

Thanks!

PR #7385, opened 2019-10-18
Remove taint usage on Ruby 2.7+

jeremyevans:

Ruby 2.7 is deprecating taint support. 

See https://bugs.ruby-lang.org/issues/16131 for details.

Bundler still uses `untaint` in a few places, which will cause deprecation warnings on Ruby 2.7.

`s/\.untaint/.tap{|x| x.untaint if RUBY_VERSION < "2.7" }/g`

It seemed simplest.


hsbt:
@bundlerbot r+

PR #7388, opened 2019-10-20
Fix help with command aliases

deivid-rodriguez:

The problem was that the output of `bundle help install` and `bundle install --help` was inconstent with `bundle i --help`, and the same issue with all of the other command aliases. 

My diagnosis was that the command reformatting so that `bundle <cmd> --help` "redirects" to `bundle help <cmd>` was ignoring command aliases.

My fix is to make this logic aware of command aliases.


deivid-rodriguez:
@bundlerbot r+

PR #7389, opened 2019-10-20
Wrap up `bundle cache` migration to current `bundle package`

deivid-rodriguez:

The problem was that after https://github.com/bundler/bundler/pull/7249, we added the current `bundle package` functionality to the current `bundle cache` command, so that these commands are now aliases of each other.

The initial plan was to do this in a more careful manner but we concluded that this was only _adding_ functionality to `bundle cache`, so not backwards incompatible.

We still need to wrap up the original plan, where `bundle cache` is the preferred command.

My fix is to migrate docs, tests and help text to use `bundle cache`, and to remove duplicated specs for `bundle cache` and `bundle package`, since they are testing the exact same code.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -130,7 +130,10 @@ namespace :man do
     begin
       Spec::Rubygems.gem_require("ronn")
     rescue Gem::LoadError => e
+      desc "Build the man pages"
       task(:build) { abort "We couln't activate ronn (#{e.requirement}). Try `gem install ronn:'#{e.requirement}'` to be able to build the help pages" }
+
+      desc "Verify man pages are in sync"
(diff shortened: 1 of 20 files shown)


deivid-rodriguez:
This essentially brings https://github.com/bundler/bundler/pull/4008 into master. That PR was merged to 2-0-dev, but was never incorporated.

As I explained in #7249, I believe we can do this without backwards compatibility concerns, because `bundle cache` functionality was as subset of `bundle package`, but brings no backwards incompatible changes otherwise.

indirect:
@bundlerbot r+

deivid-rodriguez:
Thanks!

PR #7390, opened 2019-10-20
Bump tested rubies and regenerate cassettes

deivid-rodriguez:

The problem was that we were not testing against the latest patch level rubies.

My fix is to upgrade our CI config, and regenerate the vcr cassettes, just to keep them up to date.


deivid-rodriguez:
@bundlerbot r+

Issue #7384, opened 2019-10-18
`bundle add` should update vendor/cache


benchristel:
If I have previously vendored gems using `bundle package`, and I then `bundle add` a new gem, the new gem is not cached in `vendor/cache`. I'd expect `bundle add` to cache the new gem, because if I manually add the gem to the Gemfile and `bundle install`, it *is* cached.

Using bundler 2.0.1, Ruby 2.4.2

Thanks for considering this issue!

PR #7393, opened 2019-10-25
Make 'bundle add' cache newly added gems when needed

andrehjr:

The problem was that when calling `bundle add` with cache_all as true, I have to call an additional `bundle install` to vendor gems.

My diagnosis was that a call to Bundler.load.cache was missing. For example Bundler::CLI::Update does the same after installing gems.

My fix was to call Bundler.load.cache when `Bundler.app_cache.exist? ` 

I chose this fix because it looks like this makes the behavior consistent with other commands.

I should also say that, as this is my first PR here, I'm not sure that this is the best solution, and it seems an easy fix for #7384.

### DIFF
--- a/lib/bundler/cli/add.rb
+++ b/lib/bundler/cli/add.rb
@@ -21,6 +21,7 @@ module Bundler
 
     def perform_bundle_install
       Installer.install(Bundler.root, Bundler.definition)
+      Bundler.load.cache if Bundler.app_cache.exist?
     end
 
     def inject_dependencies
--- a/spec/commands/add_spec.rb
+++ b/spec/commands/add_spec.rb
@@ -239,4 +239,13 @@ RSpec.describe "bundle add" do
       expect(err).not_to include("You may also need to change the version requirement specified in the Gemfile if it's too restrictive")
     end
   end
+
+  describe "when a gem is added and cache exists" do
+    it "caches all new dependencies added for the specified gem" do
+      bundle! :cache
+
+      bundle "add 'rack' --version=1.0.0"
+      expect(bundled_app("vendor/cache/rack-1.0.0.gem")).to exist
+    end
+  end
 end


lib/bundler/cli/add.rb
```diff
@@ -21,6 +21,7 @@ def run
 
     def perform_bundle_install
       Installer.install(Bundler.root, Bundler.definition)
+      Bundler.load.cache if Bundler.app_cache.exist? && !Bundler.frozen_bundle?
```
deivid-rodriguez: Why is the `frozen_bundle?` check needed?

spec/commands/add_spec.rb
```diff
@@ -239,4 +239,14 @@
       expect(err).not_to include("You may also need to change the version requirement specified in the Gemfile if it's too restrictive")
     end
   end
+
+  describe "when a gem is added and cache_all is true" do
```
deivid-rodriguez: I think this fix is independent from the `cache_all` setting?

That setting (when false) prevents`bundle cache` from caching `:git` and `:path` gems. It was added in https://github.com/bundler/bundler/commit/d351e68fa0a5df6de7aff587effce0801297848a, although I see no explanation on _why_.

Anyways, I don't think you need to set it here.

spec/commands/add_spec.rb
```diff
@@ -239,4 +239,14 @@
...
+    it "caches all new dependencies added for the specified gem" do
+      bundle "config set cache_all true"
+      bundle! :package
```
deivid-rodriguez: Use `:cache`, since this is our preferred alias for the command now.

lib/bundler/cli/add.rb
andrehjr: 🤔  I thought that it was needed because It makes the same call on https://github.com/bundler/bundler/blob/master/lib/bundler/cli/install.rb#L66.

But looking again, I think there's no need to call it from here. I'll remove the extra check.

spec/commands/add_spec.rb
andrehjr: You're right! Dropped that and updated the commit messages.

lib/bundler/cli/add.rb
deivid-rodriguez: I see, I'm not sure it's needed there either. I'll look into it and remove the check from `bundle install` too separately.

deivid-rodriguez:
@bundlerbot merge

PR #7394, opened 2019-10-26
Add inline RDoc documentation

zverok:
Since Bundler became part of the standard library, it renders in Ruby's docs. Unfortunately, what renders there...  is not really helpful: https://docs.ruby-lang.org/en/master/Bundler.html

I've added rudimentary documentation for `Bundler` module and two of its most user-facing methods to solve this problem at least partially.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -110,7 +110,7 @@ module Bundler
       end
     end
 
-    # Setups Bundler runtime. After +Bundler.setup+ call, all +load+ or
+    # Turns on the Bundler runtime. After +Bundler.setup+ call, all +load+ or
     # +require+ of the gems would be allowed only if they are part of
     # the Gemfile or Ruby's standard library. If the versions specified
     # in Gemfile, only those versions would be loaded.
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -14,6 +14,25 @@ require_relative "bundler/constants"
 require_relative "bundler/current_ruby"
 require_relative "bundler/build_metadata"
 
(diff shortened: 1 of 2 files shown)


lib/bundler/lockfile_parser.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+#--
```
zverok: This is necessary to make this (internal) comment invisible for RDoc.

hsbt:
👍 

Does anyone proofread this doc?

lib/bundler/lockfile_parser.rb
deivid-rodriguez: This comment is about really really old `bundler` versions, I don't think it's worth making it visible to RDoc.

lib/bundler.rb
```diff
@@ -107,6 +153,24 @@ def setup(*groups)
       end
     end
 
+    # Setups Bundler environment (see Bundler.setup) if it is not already set,
```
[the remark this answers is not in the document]
zverok: Changed it to less questionable verb :)

lib/bundler/lockfile_parser.rb
zverok: > I don't think it's worth making it visible to RDoc.

Exactly my point! The `#--` at the beginning of the comment tells RDoc to ignore it, otherwise, it would be rendered in the main class' docs :)

deivid-rodriguez: Oooops! I read "visible", not "unvisible"!

hsbt:
@bundlerbot merge

PR #7399, opened 2019-10-30
Fix some realworld specs testing nothing

deivid-rodriguez:

The problem was that this helper was printing nothing to the standard output, so the specs in this file were just checking if lockfiles contained the empty string, which is obviously true. For example,

https://github.com/bundler/bundler/blob/e70643c1be3a4417bd537d7e63470265465e693e/spec/realworld/edgecases_spec.rb#L31

My diagnosis was that the helper needs to actually print the extracted information about the resolved gem, so it can be checked that the lockfile is correct.

My fix adds the missing `puts`.


deivid-rodriguez:
@bundlerbot r+

Issue #7253, opened 2019-07-21
Core testing - spec/bundler/source_spec.rb - 'out of sync' file


MSP-Greg:
@hsbt,

I wrote some code to see if the bundler 'update from upstream' could be automated (you may have the same) for ruby/ruby and the bundler default gem files, and ran the CI in Pipelines in my ruby/ruby fork.

In the Ruby CI, some of the specs in this file are 'pending', but the same doesn't happen with the copy here pushed into Ruby.

I really haven't looked at the code, nor an I familiar with Thor, but https://github.com/ruby/ruby/commit/19408c785c6f258024efcaa69d0642ad995c50b3 may need to be added here...

With that added, all the commits added here since the last update can be pushed into ruby/ruby and yield the same results on the MacOS job.  The Ubuntu job is being corrupted by something with Rspec...

deivid-rodriguez:
These specs are no longer skipped everywhere since https://github.com/bundler/bundler/pull/7002. They are excluded from the Windows build though.

So I think this can be closed?

MSP-Greg:
Ok.  Bundler is a default gem in Ruby Core, but the files are not synced often.

As I stated, Ruby Core tests itself with the bundler specs with MacOS.  This is done in an Azure Pipeline build, tested on Ubuntu & MacOS (not sure about versions).

There is no testing here on MacOS.  When I moved the current Bundler files into Core and tested them in my fork, something was amiss with this file, as there were additional failures.

So, whenever the next 'bundler sync' happens, the test results from the MacOS should compared before and after, and whatever the cause, fixed.

When bundler runs in Core as it currently is, there are no failures in this file.  When I ran it with the current bundler master, the following errors occurred on MacOS:

```
rspec ./spec/bundler/bundler/source_spec.rb:62 # Bundler::Source#version_message when there are locked gems that contain the relevant gem spec with a different version with color should return a string with the spec name and version and locked spec version
rspec ./spec/bundler/bundler/source_spec.rb:83 # Bundler::Source#version_message when there are locked gems that contain the relevant gem spec with a more recent version with color should return a string with the locked spec version in yellow
rspec ./spec/bundler/bundler/source_spec.rb:104 # Bundler::Source#version_message when there are locked gems that contain the relevant gem spec with an older version with color should return a string with the locked spec version in green
```

Ok to close, but I thought a 'marker' about this might be helpful.

deivid-rodriguez:
[...]

PR #7401, opened 2019-10-30
Stop silencing output by default

deivid-rodriguez:

The problem was that bundler defaults to a silent UI:

https://github.com/bundler/bundler/blob/e70643c1be3a4417bd537d7e63470265465e693e/lib/bundler.rb#L66-L68

In my opinion, this isn't a good behavior for a CLI tool, and forces us to override it in many many different places. It has also caused several issues, for example, https://github.com/bundler/bundler/issues/3710 where `bundle list` was printing nothing.

The [solution to that issues](https://github.com/bundler/bundler/pull/3707) led us to add yet more places where we override the default UI, and @indirect [predicting that having to unset the UI everytime we want it to not be silent](https://github.com/bundler/bundler/pull/3707#issuecomment-108646127) would cause many headaches.

Well, yeah... I've lost a lot of time trying to figure out why UI was silent sometimes, and normal another times, why some specs printed warnings and some didn't. In particular, see my series of "big fail PRs" fighting against bundler's UI: https://github.com/bundler/bundler/pull/7284, https://github.com/bundler/bundler/pull/7294, https://github.com/bundler/bundler/pull/7305, https://github.com/bundler/bundler/pull/7362.

Another series of issues/PRs probably related to this is issue https://github.com/bundler/bundler/issues/6900, where the output would use a different UI on different environments. We had a lot of trouble to reliably fix it (https://github.com/bundler/bundler/pull/6994, https://github.com/bundler/bundler/pull/7002, https://github.com/bundler/bundler/issues/7253).

I also run into these issues again when trying out the `RUBYGEMS_GEMDEPS` environment variable that enables `bundler/setup` from rubygems.

My diagnosis was that we shouldn't silence UI by default.

My fix is to, instead of silencing and then overriding the default shell at many places, don't silence it by default and instead make it silent when needed. By doing this, I managed to get 100% of our specs green, so I'm pretty confident that the output is still the same (or if it's not, it's probably because some output/errors where being unintentionally swallowed). 

Now specs should pass, but they print a bunch of output to the screen. You can see error messages, hard crashes, success messages... Some of them might be showing actual issues with either the code or tests, so I plan to go through each of them and review them. I can do that in this PR or separately, no strong opinion.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -64,11 +64,11 @@ module Bundler
     end
 
     def ui
-      (defined?(@ui) && @ui) || (self.ui = UI::Silent.new)
+      (defined?(@ui) && @ui) || (self.ui = UI::Shell.new)
     end
 
     def ui=(ui)
-      Bundler.rubygems.ui = ui ? UI::RGProxy.new(ui) : nil
+      Bundler.rubygems.ui = UI::RGProxy.new(ui)
       @ui = ui
     end
 
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -15,9 +15,6 @@ module Bundler
 
     def self.start(*)
       super
-    rescue Exception => e # rubocop:disable Lint/RescueException
-      Bundler.ui = UI::Shell.new
-      raise e
     ensure
       Bundler::SharedHelpers.print_major_deprecations!
     end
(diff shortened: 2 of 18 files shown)


deivid-rodriguez:
Thanks @indirect!

I added a couple of extra tweaks so that the test output is clean again as it was PR this PR. It should be ready now I think.

indirect:
@bundlerbot r+

deivid-rodriguez:
@bundlerbot retry

PR #7402, opened 2019-10-30
Spec normalizations

deivid-rodriguez:

The problem was that I saw some room for improvement in our specs and I couldn't help myself :)

Essentially, the idea of this PR is to avoid using `__FILE__` or other folder-structure specific constructs to look for files, and instead use our helpers ìn `spec/support/path.rb`. The idea is that the helpers in this file are aware that `bundler` specs can be run from ruby-core or the upstream repo, and the folder structure is different for both.

Also, the PR normalizes some naming, some duplicated helpers, and the way we require files from specs.


deivid-rodriguez:
@bundlerbot merge

PR #7404, opened 2019-10-31
Support multiple groups for --without-group and --only-group options …

fatkodima:
This is a reworked version of #6939 
It accepts multiple groups for `--without-group` and `--only-group`.

Closes #6939

### DIFF
--- a/man/bundle-list.1.txt
+++ b/man/bundle-list.1.txt
@@ -6,8 +6,8 @@ NAME
        bundle-list - List all the gems in the bundle
 
 SYNOPSIS
-       bundle	 list	 [--name-only]	  [--paths]    [--without-group=GROUP]
-       [--only-group=GROUP]
+       bundle  list [--name-only] [--paths] [--without-group=GROUP[ GROUP...]]
+       [--only-group=GROUP[ GROUP...]]
 
 DESCRIPTION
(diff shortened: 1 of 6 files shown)


fatkodima:
@deivid-rodriguez Updated with documentation.

man/bundle-list.1.txt
```diff
@@ -31,11 +31,12 @@ OPTIONS
        --paths
 	      Print the path to each gem in the bundle.
 
-       --without-group
-	      Print all gems expect from a group.
+       --without-group=<list>
+	      A space-separated list of groups referencing gems to skip during
```
deivid-rodriguez: Super minor change but I think "groups of gems to skip during printing" reads better than "groups referencing gems to skip during printing".

man/bundle-list.1.txt
```diff
@@ -31,11 +31,12 @@ OPTIONS
...
+	      printing.
 
-       --only-group
-	      Print gems from a particular group.
+       --only-group=<list>
+	      A space-separated list of groups referencing gems to print.
```
deivid-rodriguez: Same thing about `s/referencing/of/`.

fatkodima:
@deivid-rodriguez Updated.

deivid-rodriguez:
@bundlerbot merge

PR #7406, opened 2019-10-31
Remove junit formatter in Azure for the moment

deivid-rodriguez:

The problem was it's not easy to inspect Windows failures at the moment. Github checks only show a few inline annotations, but we have ~330 failures at the moment.

My diagnosis was that for the moment, regular test output is better.

My fix is to remove the junit formatter for the moment.

### DIFF
--- a/.azure-pipelines/steps.yml
+++ b/.azure-pipelines/steps.yml
@@ -24,16 +24,5 @@ steps:
   displayName: 'ruby bin/rake spec:deps'
 
 - script: |
-    gem install --no-document --conservative rspec_junit_formatter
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Actually it's super simple to check the failing tests in Azure (there's a Tests tab with all the information ready to filter), I just didn't know how to use the UI. I'm closing this.

hsbt:
I'm going to migrate Azure Pipelines and Travis CI to GitHub Actions after releasing Bundler 2.1.0. Then We will drop to support JUnit formatter.

PR #7410, opened 2019-11-01
November man pages

deivid-rodriguez:

The problem was that man pages are outdated after the month change.

Well, the problem actually is that I need to fix in a better way (like a monthly scheduled rebuild or something), but I keep just updating the man pages every month... :sweat_smile:

My fix to run `bin/rake man:build` and commit the result.


deivid-rodriguez:
@bundlerbot merge

deivid-rodriguez:
@bundlerbot retry

PR #7411, opened 2019-11-01
Try out specs against MRI 2.6 + Rubygems 3.0

deivid-rodriguez:

The problem was that our Azure build was using outdated versions of ruby and rubygems.

My diagnosis was that we should update them.

My fix is to update them. With the latest versions, failures go down to 319.

### DIFF
--- a/.azure-pipelines/steps.yml
+++ b/.azure-pipelines/steps.yml
@@ -2,7 +2,7 @@ steps:
 
 - task: UseRubyVersion@0
   inputs:
-    versionSpec: '= 2.4'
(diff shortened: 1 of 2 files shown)


MSP-Greg:
@deivid-rodriguez 

Probably wait till the weekend. You ok with switching to Actions for Windows?

deivid-rodriguez:
Yes, that was in our plans anyways, so :+1:.

deivid-rodriguez:
I'll get this in for now.

@bundlerbot merge

Issue #7413, opened 2019-11-01
Please, configure github's merge button


fatkodima:
Please, configure github's merge button with rebasing and **squashing**, avoiding useless lists of commits for each PR and making possible to navigate through historical changes and debug problems.

hsbt:
We use the bors bot for git merge. So, We never use github merge button directly.

PR #7415, opened 2019-11-01
Squash before merging for bors

fatkodima:
@hsbt 

https://bors.tech/documentation/

#7413

### DIFF
--- a/.github/bors.toml
+++ b/.github/bors.toml
@@ -4,6 +4,7 @@ status = [
 
 timeout_sec = 14400
 delete_merged_branches = true
+use_squash_merge = true
 
 [committer]
 name = "Bundlerbot"


hsbt:
I don't think we need to change it. 

@deivid-rodriguez Do you have any opinion?

fatkodima:
Personally, I prefer to see all code changes for feature implementation in one place in one commit instead of spread across a wall of meaningless commits. It is then very hard to track and debug when and why something changes.

hsbt:
> I prefer to see all code changes for feature implementation in one place in one commit

IMO, I don't think so.

fatkodima:
How about this trash?

Screen Shot 2019-11-01 at 23 33 01

Screen Shot 2019-11-01 at 23 31 19

Screen Shot 2019-11-01 at 23 30 22

And 100500 similar commits without context.

hsbt:
I have no feelings about that. 

And your comment is not respectable for the developer team of bundler.

PR #7416, opened 2019-11-01
Don't use insecure temporary directory as home directory

### LINKED ISSUE #6501: tmp_home_path insecure
Hi,

If the home directory of the user is not writable the function tmp_home_path from lib/bundler.rb, creates an insecure temporary directory in tmp/:
- it is a world writable
- it has a with a predictable name.
Moreover, this temporary directory is not removed at the end.

See:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=796383
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=881749
for more details.

Thanks in advance.

fatkodima:
Closes #6501

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -167,29 +167,13 @@ module Bundler
         end
 
         if warning
-          Kernel.send(:require, "etc")
-          user_home = tmp_home_path(Etc.getlogin, warning)
+          user_home = tmp_home_path(warning)
           Bundler.ui.warn "#{warning}\nBundler will use `#{user_home}' as your home directory temporarily.\n"
           user_home
         else
           Pathname.new(home)
         end
       end
-    end
-
-    def tmp_home_path(login, warning)
-      login ||= "unknown"
-      Kernel.send(:require, "tmpdir")
-      path = Pathname.new(Dir.tmpdir).join("bundler", "home")
-      SharedHelpers.filesystem_access(path) do |tmp_home_path|
-        unless tmp_home_path.exist?
-          tmp_home_path.mkpath
-          tmp_home_path.chmod(0o777)
-        end
-        tmp_home_path.join(login).tap(&:mkpath)
-      end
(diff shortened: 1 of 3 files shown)


lib/bundler.rb
```diff
@@ -177,16 +176,12 @@ def user_home
       end
     end
 
-    def tmp_home_path(login, warning)
-      login ||= "unknown"
+    def tmp_home_path(_login, warning)
```
deivid-rodriguez: I assume you didn't change the signature to remove the `login` parameter for backwards compatibility?

I don't think anybody should be using this method, and it's returning an insecure path anyways, so I would vote for making it private and removing the now unused parameter.

Thoughts?

spec/bundler/bundler_spec.rb
```diff
@@ -282,31 +276,17 @@
...
-  describe "#tmp_home_path" do
-    it "should create temporary user home" do
-      allow(Dir).to receive(:tmpdir).and_return("/TMP")
-      allow(FileTest).to receive(:exist?).with("/TMP/bundler/home").and_return(false)
-      expect(FileUtils).to receive(:mkpath).once.ordered.with("/TMP/bundler/home")
-      expect(FileUtils).to receive(:mkpath).once.ordered.with("/TMP/bundler/home/USER")
-      expect(File).to receive(:chmod).with(0o777, "/TMP/bundler/home")
-      expect(Bundler.tmp_home_path("USER", "")).to eq(Pathname("/TMP/bundler/home/USER"))
-    end
-  end
```
deivid-rodriguez: I agree with removing this test if we decide to make this method private.

We could consider a graceful deprecation but I really doubt anybody is relying on the result of this method for anything?

fatkodima:
@deivid-rodriguez Updated with your suggestion of making method private. I am doubt too that anybody uses it.

deivid-rodriguez:
@bundlerbot merge

PR #7417, opened 2019-11-02
Fix bundler/inline warning: method redefined; discarding old root

### LINKED ISSUE #6167: bundler/inline warning: method redefined; discarding old root
`bundler/inline` works by redefining `Bundler.root` and then putting it back again... but `ruby -w` isn't a fan of that technique.

We get a warning [when it installs its new version](https://travis-ci.org/rails/rails/jobs/300535874#L797):
```
/home/travis/.rvm/rubies/ruby-2.4.1/lib/ruby/site_ruby/2.4.0/bundler/inline.rb:42: warning: method redefined; discarding old root
/home/travis/.rvm/rubies/ruby-2.4.1/lib/ruby/site_ruby/2.4.0/bundler.rb:216: warning: previous definition of root was here
```

and another when it [returns the original](https://travis-ci.org/rails/rails/jobs/300535874#L878):
```
/home/travis/.rvm/rubies/ruby-2.4.1/lib/ruby/site_ruby/2.4.0/bundler/inline.rb:73: warning: method redefined; discarding old root
/home/travis/.rvm/rubies/ruby-2.4.1/lib/ruby/site_ruby/2.4.0/bundler/inline.rb:42: warning: previous definition of root was here
```

This isn't a PR, mostly because I'm not sure which of the various not-great alternatives you'd prefer in order to make it happy. The ones I know of are:

* remove the method before redefining it -- which can be racy for other threads, though I guess inline isn't exactly thread-friendly to begin with
* *alias* the method before redefining it -- this is what we're moving to in Rails, but it feels pretty unclear for a future reader
* avoid redefining the method at all -- feels like a lot of work for no real gain, to fix a papercut of a warning

fatkodima:
Closes #6167

### DIFF
--- a/lib/bundler/inline.rb
+++ b/lib/bundler/inline.rb
@@ -38,6 +38,8 @@ def gemfile(install = false, options = {}, &gemfile)
   raise ArgumentError, "Unknown options: #{opts.keys.join(", ")}" unless opts.empty?
 
   old_root = Bundler.method(:root)
+  bundler_module = class << Bundler; self; end
+  bundler_module.send(:remove_method, :root)
   def Bundler.root
     Bundler::SharedHelpers.pwd.expand_path
   end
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
This was also the approach preferred by @segiddins to fix this, going in!

@bundlerbot merge

PR #7419, opened 2019-11-03
Add :glob to git source uniqueness

### LINKED ISSUE #7346: Bundler cannot add multiple gems from the same git source
I've created a POC project for this behavior (https://github.com/ajorgensen/bundler-poc) but I will also summarize the behavior here.

Essentially if you have a `Gemfile` with multiple gems coming from the same git source but with different subfolders then `bundle install` will error out because it cannot find the dependency. The most simple example I could come up with was:

```
source 'https://rubygems.org'

gem 'base_gem', git: 'git@github.com:ajorgensen/bundler-poc.git', glob: 'gems/base_gem/*.gemspec'
gem 'some_gem', git: 'git@github.com:ajorgensen/bundler-poc.git', glob: 'gems/some_gem/*.gemspec'

gemspec
```

Running `bundle install` with this gemfile will result in the following error:
```
Fetching git@github.com:ajorgensen/bundler-poc.git
Could not find gem 'base_gem' in git@github.com:ajorgensen/bundler-poc.git (at master@002e6fd).
The source does not contain any versions of 'base_gem'
```

This is surprising because I would expect that this *should* work correctly. The workaround I found was to do something like this:

```
source 'https://rubygems.org'

gem 'base_gem', git: 'git@github.com:ajorgensen/bundler-poc.git', glob: 'gems/**/*.gemspec'
gem 'some_gem', git: 'git@github.com:ajorgensen/bundler-poc.git', glob: 'gems/**/*.gemspec'

gemspec
```

By simply always including all of the gemspecs I can get `bundle install` to work correctly.

I did a bit of digging and my hypothesis is that the source is getting added for each of the different gems but the options are getting overwritten because the source itself is the same for both gems. So in the broken case, the source would get added first with the one `glob` option and then it is added a second time and overwrites the existing `glob` option. I believe this is happening somewhere around here: https://github.com/bundler/bundler/blob/master/lib/bundler/source_list.rb#L125-L128

I will admit I have very little knowledge of how bundler works but my naive thought would be that sources should be treated separately if they have different options or add in some special logic to merge the two sources so the options like `glob` aren't overwritten but merged.

I will continue to dig deeper into this and see if I can find an elegant way to fix this behavior but I thought I would put up an issue to see if someone who knows more has a more clever idea.

fatkodima:
Closes #7346

### DIFF
--- a/lib/bundler/source/git.rb
+++ b/lib/bundler/source/git.rb
@@ -8,7 +8,7 @@ module Bundler
     class Git < Path
       autoload :GitProxy, File.expand_path("git/git_proxy", __dir__)
 
-      attr_reader :uri, :ref, :branch, :options, :submodules
+      attr_reader :uri, :ref, :branch, :options, :glob, :submodules
 
       def initialize(options)
         @options = options
@@ -48,13 +48,14 @@ module Bundler
       end
 
       def hash
-        [self.class, uri, ref, branch, name, version, submodules].hash
+        [self.class, uri, ref, branch, name, version, glob, submodules].hash
       end
 
       def eql?(other)
         other.is_a?(Git) && uri == other.uri && ref == other.ref &&
           branch == other.branch && name == other.name &&
-          version == other.version && submodules == other.submodules
+          version == other.version && glob == other.glob &&
+          submodules == other.submodules
       end
 
       alias_method :==, :eql?
--- a/spec/install/git_spec.rb
+++ b/spec/install/git_spec.rb
@@ -61,5 +61,25 @@ RSpec.describe "bundle install" do
 
       expect(out).to include("Bundle complete!")
     end
+
+    it "allows multiple gems from the same git source" do
+      build_repo2 do
+        build_lib "foo", "1.0", :path => lib_path("gems/foo")
+        build_lib "zebra", "2.0", :path => lib_path("gems/zebra")
+        build_git "gems", :path => lib_path("gems"), :gemspec => false
+      end
+
+      install_gemfile <<-G
+        source "#{file_uri_for(gem_repo2)}"
+        gem "foo", :git => "#{lib_path("gems")}", :glob => "foo/*.gemspec"
+        gem "zebra", :git => "#{lib_path("gems")}", :glob => "zebra/*.gemspec"
+      G
+
+      bundle "info foo"
+      expect(out).to include("* foo (1.0 #{revision_for(lib_path("gems"))[0..6]})")
+
+      bundle "info zebra"
+      expect(out).to include("* zebra (2.0 #{revision_for(lib_path("gems"))[0..6]})")
+    end
   end
 end


deivid-rodriguez:
The test here is incorrect, but the fix looks good. The test should be modified like this to actually catch the issue:

```diff
diff --git a/spec/install/git_spec.rb b/spec/install/git_spec.rb
index 93b90a705..2ce5ddf15 100644
--- a/spec/install/git_spec.rb
+++ b/spec/install/git_spec.rb
@@ -64,18 +64,19 @@ RSpec.describe "bundle install" do
 
     it "allows multiple gems from the same git source" do
       build_repo2 do
-        build_git "foo", "1.0", :path => lib_path("foo")
-        build_git "zebra", "2.0", :path => lib_path("zebra")
+        build_lib "foo", "1.0", :path => lib_path("gems/foo")
+        build_lib "zebra", "2.0", :path => lib_path("gems/zebra")
+        build_git "gems", :path => lib_path("gems"), :gemspec => false
       end
 
       install_gemfile <<-G
         source "#{file_uri_for(gem_repo2)}"
-        gem "foo", :git => "#{lib_path("foo")}"
-        gem "zebra", :git => "#{lib_path("zebra")}"
+        gem "foo", :git => "#{lib_path("gems")}", :glob => "foo/*.gemspec"
+        gem "zebra", :git => "#{lib_path("gems")}", :glob => "zebra/*.gemspec"
       G
 
-      expect(out).to include("Using foo 1.0 from #{lib_path("foo")} (at master@#{revision_for(lib_path("foo"))[0..6]})")
-      expect(out).to include("Using zebra 2.0 from #{lib_path("zebra")} (at master@#{revision_for(lib_path("zebra"))[0..6]})")
+      expect(out).to include("Using foo 1.0 from #{lib_path("gems")} (at master@#{revision_for(lib_path("gems"))[0..6]})")
+      expect(out).to include("Using zebra 2.0 from #{lib_path("gems")} (at master@#{revision_for(lib_path("gems"))[0..6]})")
     end
   end
 end
```

fatkodima:
@deivid-rodriguez Updated with your suggestion.

deivid-rodriguez:
Ok! Now I see that CI is failing in Bundler 3 mode because `bundler 3` will no longer print "Using" messages to screen on `bundle install`, unless the dependency has changed, so the assertions are failing.

You can test the same thing in another way with something like:

```diff
diff --git a/spec/install/git_spec.rb b/spec/install/git_spec.rb
index 93b90a705..3bc8e1c06 100644
--- a/spec/install/git_spec.rb
+++ b/spec/install/git_spec.rb
@@ -74,8 +74,11 @@ RSpec.describe "bundle install" do
         gem "zebra", :git => "#{lib_path("zebra")}"
       G
 
-      expect(out).to include("Using foo 1.0 from #{lib_path("foo")} (at master@#{revision_for(lib_path("foo"))[0..6]})")
-      expect(out).to include("Using zebra 2.0 from #{lib_path("zebra")} (at master@#{revision_for(lib_path("zebra"))[0..6]})")
+      bundle "info foo"
+      expect(out).to include("* foo (1.0 #{revision_for(lib_path("foo"))[0..6]})")
+
+      bundle "info zebra"
+      expect(out).to include("* zebra (2.0 #{revision_for(lib_path("zebra"))[0..6]})")
     end
   end
 end
```

On a side note, I think we might want to still print the "Using" messages on the first `bundle install`, since technically, the dependency has changed in this case (from none to something). I'll open a separate issue to discuss this.

fatkodima:
@deivid-rodriguez Got it. Updated.

deivid-rodriguez:
@bundlerbot merge

PR #5521, opened 2017-03-17
bundler gem: Support test-unit as a testing framework

kou:

PR #7420, opened 2019-11-03
Add "--test=test-unit" option for gem command

mame:

The problem was that the `bundle gem --test` option supported only "minitest" and "rspec", not "test-unit".

My diagnosis was that the lack of "test-unit" was very unfortunate.  It is only one test framework gem that is bundled to the ruby distribution.

My fix is to add "--test=test-unit" option for `bundle gem` command.  I implemented it based on the implementation of "--test=minitest".  It creates "test/test_newgem.rb" (test-unit style) instead of "test/newgem_test.rb" (mnitest style).  It also tweaks Rakefile and gemspec files appropriately.

I chose this fix because I couldn't come up with other options.

### DIFF
--- a/lib/bundler/cli/gem.rb
+++ b/lib/bundler/cli/gem.rb
@@ -12,6 +12,7 @@ module Bundler
     TEST_FRAMEWORK_VERSIONS = {
       "rspec" => "3.0",
       "minitest" => "5.0",
+      "test-unit" => "3.0",
     }.freeze
 
     attr_reader :options, :gem_name, :thor, :name, :target
@@ -97,10 +98,14 @@ module Bundler
             "test/test_helper.rb.tt" => "test/test_helper.rb",
             "test/newgem_test.rb.tt" => "test/#{namespaced_path}_test.rb"
           )
+        when "test-unit"
+          templates.merge!(
+            "test/test_newgem.rb.tt" => "test/test_#{namespaced_path}.rb"
+          )
         end
       end
 
-      config[:test_task] = config[:test] == "minitest" ? "test" : "spec"
+      config[:test_task] = config[:test] == "minitest" || config[:test] == "test-unit" ? "test" : "spec"
 
       if ask_and_set(:mit, "Do you want to license your code permissively under the MIT license?",
         "This means that any other developer or company will be legally allowed to use your code " \
@@ -199,9 +204,9 @@ module Bundler
 
       if test_framework.nil?
         Bundler.ui.confirm "Do you want to generate tests with your gem?"
-        result = Bundler.ui.ask "Type 'rspec' or 'minitest' to generate those test files now and " \
-          "in the future. rspec/minitest/(none):"
-        if result =~ /rspec|minitest/
+        result = Bundler.ui.ask "Type 'rspec' or 'minitest' or 'test-unit' to generate those test files now and " \
+          "in the future. rspec/minitest/test-unit/(none):"
+        if result =~ /rspec|minitest|test-unit/
           test_framework = result
         else
           test_framework = false
--- a/lib/bundler/templates/newgem/Rakefile.tt
+++ b/lib/bundler/templates/newgem/Rakefile.tt
@@ -1,11 +1,15 @@
 require "bundler/gem_tasks"
-<% if config[:test] == "minitest" -%>
+<% if config[:test] == "minitest" || config[:test] == "test-unit" -%>
 require "rake/testtask"
 
 Rake::TestTask.new(:test) do |t|
   t.libs << "test"
   t.libs << "lib"
+<% if config[:test] == "minitest" -%>
   t.test_files = FileList["test/**/*_test.rb"]
+<% else -%>
+  t.test_files = FileList["test/**/test_*.rb"]
+<% end -%>
 end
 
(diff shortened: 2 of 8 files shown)


lib/bundler/cli/gem.rb
```diff
@@ -97,10 +98,14 @@ def run
             "test/test_helper.rb.tt" => "test/test_helper.rb",
             "test/newgem_test.rb.tt" => "test/#{namespaced_path}_test.rb"
           )
+        when "test-unit"
+          templates.merge!(
```
kou: Having `test/test_helper.rb` is better to share common helper codes.
See also: https://github.com/bundler/bundler/pull/5521/files#diff-7675ffc4b1b026242f4099e54c032b7b

lib/bundler/cli/gem.rb
```diff
@@ -97,10 +98,14 @@ def run
...
+            "test/test_newgem.rb.tt" => "test/test_#{namespaced_path}.rb"
```
kou: test-unit can use `XXX_test.rb` style.
So we can use `XXX_test.rb` here like other test framework.

lib/bundler/cli/gem.rb
```diff
@@ -97,10 +98,14 @@ def run
...
+          )
         end
       end
 
-      config[:test_task] = config[:test] == "minitest" ? "test" : "spec"
+      config[:test_task] = config[:test] == "minitest" || config[:test] == "test-unit" ? "test" : "spec"
```
kou: `config[:test] == "rspec" ? "spec" : "test"` will be simpler.
See also: https://github.com/bundler/bundler/pull/5521/files#diff-2ee2e162f4eceee05f93f6197078d10bR99

lib/bundler/templates/newgem/Rakefile.tt
```diff
@@ -1,11 +1,15 @@
...
 require "rake/testtask"
 
 Rake::TestTask.new(:test) do |t|
   t.libs << "test"
   t.libs << "lib"
+<% if config[:test] == "minitest" -%>
   t.test_files = FileList["test/**/*_test.rb"]
+<% else -%>
+  t.test_files = FileList["test/**/test_*.rb"]
+<% end -%>
```
kou: We can remove this clause when we use `XXX_test.rb` style.

lib/bundler/templates/newgem/test/test_newgem.rb.tt
```diff
@@ -0,0 +1,14 @@
+$LOAD_PATH.unshift File.expand_path("../lib", __dir__)
+require "<%= config[:namespaced_path] %>"
+
+require "test/unit"
+
+class Test<%= config[:constant_name] %> < Test::Unit::TestCase
+  def test_that_it_has_a_version_number
+    refute_nil ::<%= config[:constant_name] %>::VERSION
```
kou: The following code is test-unit style:

```ruby
  def test_version
    assert do
      ::<%= config[:constant_name] %>.const_defined?(:VERSION)
    end
  end
```

See also: https://github.com/bundler/bundler/pull/5521/files#diff-7675ffc4b1b026242f4099e54c032b7b

lib/bundler/templates/newgem/test/test_newgem.rb.tt
```diff
@@ -0,0 +1,14 @@
+$LOAD_PATH.unshift File.expand_path("../lib", __dir__)
+require "<%= config[:namespaced_path] %>"
+
+require "test/unit"
```
kou: `require "test-unit"` is better to ensure using test-unit gem.
(We couldn't require test-unit gem by `require "test/unit"` with old Ruby. Ruby 2.4 or later aren't old Ruby. So we can use `require "test/unit" here recently.)

lib/bundler/cli/gem.rb
```diff
@@ -199,9 +204,9 @@ def ask_and_set_test_framework
 
       if test_framework.nil?
         Bundler.ui.confirm "Do you want to generate tests with your gem?"
-        result = Bundler.ui.ask "Type 'rspec' or 'minitest' to generate those test files now and " \
-          "in the future. rspec/minitest/(none):"
-        if result =~ /rspec|minitest/
+        result = Bundler.ui.ask "Type 'rspec' or 'minitest' or 'test-unit' to generate those test files now and " \
```
kou: Is `'rspec', 'minitest' or 'test-unit'` better?

lib/bundler/cli/gem.rb
mame: Thank you for reviewing this, but I believe it is greater to merge #5521 than this PR because that was created by the maintainer of test-unit.  Could you rebase #5521 if the feature request is accepted?

kou: OK.

hsbt:
@mame Can you update with @kou 's feedbacks?

kou:
Is this request accepted?
Then I'll complete this pull request. ( https://github.com/bundler/bundler/pull/7420#discussion_r341933216 )

mame:
@hsbt As the comment, I thought I passed the ball to @kou.  This PR should be closed.  Closing.

@kou I think the feature has been already accepted.  Please go ahead!

PR #7422, opened 2019-11-05
Turn on warning mode for specs

deivid-rodriguez:

The problem was that we're missing verbose warnings like the one fixed by #7417.

My diagnosis was that our specs should let us know about this.

My fix is to enable warnings from our helpers that shell out, and fix the warnings they print. There's also some unrelated changes here that I'll try to split out once I get this green.

I chose this fix because it makes it so that our specs help us catching these warnings earlier.

### DIFF
--- a/spec/runtime/inline_spec.rb
+++ b/spec/runtime/inline_spec.rb
@@ -88,7 +88,6 @@ RSpec.describe "bundler/inline#gemfile" do
     RUBY
 
     expect(out).to include("Installing activesupport")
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
The only failing specs are now hitting the problem being fixed by https://github.com/bundler/bundler/pull/7417, so they should get green after that PR is merged and prevent future warnings from creeping in.

deivid-rodriguez:
@bundlerbot merge

PR #7424, opened 2019-11-06
Drop unnecessary `tempfile` require

deivid-rodriguez:

The `tempfile` dependency was dropped in [1], but the require was left in there. This could cause gem conflicts because `tempfile` requires `tmpdir` which requires `fileutils`, which loads the default gem instead of our namespaced version, and that could cause fileutils version mismatches and code overriding warnings.

[1]: https://github.com/bundler/bundler/commit/4a37d66f3f222739178d798b30fb135f2429fe45

My fix is to drop the unnecessary require.

### DIFF
--- a/lib/bundler/source/git/git_proxy.rb
+++ b/lib/bundler/source/git/git_proxy.rb
@@ -2,7 +2,7 @@
 
 require "open3"
 require "shellwords"
-require "tempfile"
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
@bundlerbot merge

PR #7427, opened 2019-11-06
Pass `env` as a parameter to `sys_exec`

deivid-rodriguez:

The problem was that Windows has trouble with commands of the form `BUNDLE_FROZEN=true bundle install`.

My diagnosis was that we should use `Open3.popen3` explicit `env` argument instead of passing a command line that has environment variables in the front.

My fix is to do just that. This is a follow up to #7011, but specifically for the `ruby` test helper. I expect this to fix some more Windows failures.

### DIFF
--- a/spec/support/helpers.rb
+++ b/spec/support/helpers.rb
@@ -172,10 +172,10 @@ module Spec
     end
 
     def ruby(ruby, options = {})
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
@bundlerbot merge

PR #7428, opened 2019-11-07
2.1.0.pre.3

deivid-rodriguez:
This is the upcoming 2.1.0.pre.3.

I only documented PRs that (I expect to) include user facing changes, but the release also includes:

* Some improvements in bundler integration as a default gem. I don't think these changes should change behavior but there's some default gem integration issues that I could never reproduce but the less we rely and mess with bundler's own `$LOAD_PATH` usually help here. That's #7398.
* Improvements in the experience debugging bundler by making the UI non silent by default. Again, I don't expect user facing changes, but if I'm missing something, I think it would be for the better. This is #7401.
* Bumping fileutils version to 1.3.0. We were using 1.2.0 with some patches on top. Again, no expected changes on the user side, but we should aim that our releases include vendorize versions of dependencies that exactly match something released on rubygems.org, and this change goes in that direction. This is #7375. 
* Refactorings and updates to the development environment.


deivid-rodriguez:
@bundlerbot merge

PR #7429, opened 2019-11-07
Remove rubygems integration dead code

deivid-rodriguez:

The problem was that I found some dead code.

My fix is to remove it because it's dead.


deivid-rodriguez:
@bundlerbot r=hsbt

PR #7430, opened 2019-11-07
Improve `release:prepare_patch` task

deivid-rodriguez:

The problem was some more little issues with the `bin/rake release:prepare_patch` task.

My fix is to improve the task with better fallbacks and better informational and error messages.


deivid-rodriguez:
@bundlerbot r=hsbt

PR #7433, opened 2019-11-11
Improve binstub specs

deivid-rodriguez:

The problem was that I run into some failures when testing rubygems against the latest bundler 2-1-stable branch of bundler.

In certain contexts, the previous approach of overwriting the version constant led into version mismatch issues.

Since in this case we are actually testing some rubygems stuff present in the generated binstub, but we don't really care what the bundler gem is actually doing, we can follow the same approach as other specs and install a "dummy" bundler gem to the system gems home used by our specs.

I chose this fix because it still tests the same thing but in a more resilient way, I believe.


deivid-rodriguez:
@bundlerbot merge

PR #7434, opened 2019-11-11
More lazy loading and relative requires

deivid-rodriguez:

The problem was that testing the latest version of bundler as a vendored copy inside rubygems surfaced some more issues about requiring the wrong code, or activating default gems too early.

My diagnosis was that we should follow similar techniques I've used in other PRs: delay as much as possible requiring default gems, and don't rely on the LOAD_PATH for internal requires.

My fix does just that, and also removes a bunch of `require`'s that were not necessary at all.


deivid-rodriguez:
@bundlerbot merge

PR #7435, opened 2019-11-11
Fix RGV env variable specified as a path

deivid-rodriguez:

The problem was that using `RGV` (the environment variable we use to select the proper rubygems version to test bundler against) with a path was not quite working.

Setting `RGV` to a path (`RGV=..`) is used in the rubygems repo to test the vendored copy of bundler (that lives at `bundler/` inside that repo) against the checked out copy of the repository.

My diagnosis was that the previous approach to switch the `rubygems` version, namely, reexec'ing with `--disable-gems` and explicitly requiring our "version switching code" was not quite working because nested subprocesses would end up "leaking" to the system copy of `rubygems` again.

My fix is to use `RUBYOPT` to configure the path to the rubygems copy that should be used, since I found that even the `require "rubygems"` present's in [ruby's `gem_prelude.rb`](https://github.com/ruby/ruby/blob/c5eb24349a4535948514fe765c3ddb0628d81004/gem_prelude.rb#L1) respects the `-I` option in the `RUBYOPT` env variable.

I chose this fix because it's simpler and works better.

### DIFF
--- a/spec/install/gemfile/git_spec.rb
+++ b/spec/install/gemfile/git_spec.rb
@@ -1056,7 +1056,6 @@ RSpec.describe "bundle install with git sources" do
 
       File.open(lib_path("install_hooks.rb"), "w") do |h|
         h.write <<-H
-          require '#{spec_dir}/support/rubygems'
           Gem.pre_install_hooks << lambda do |inst|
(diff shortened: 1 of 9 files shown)


deivid-rodriguez:
@bundlerbot merge

PR #7437, opened 2019-11-11
Extra patches for 2.1.0.pre.3

deivid-rodriguez:

The problem was that before releasing 2.1.0.pre.3 I found some issues when integrating the new version into rubygems.

My fix is to add the patches fixing the issues I found to the upcoming release.


deivid-rodriguez:
@bundlerbot merge

PR #7438, opened 2019-11-12
Add missing `travis_retry`'s to TravisCI config

deivid-rodriguez:

The problem was that CI failed in #7437 due to a network issue.

My diagnosis was that we should add `travis_retry` in the task that failed and other short tasks that may touch the network.

My fix is to do just that.


deivid-rodriguez:
@bundlerbot merge

PR #7443, opened 2019-11-18
Pick up changelog and version bump from 2-1-stable

deivid-rodriguez:

The problem was that the master branch needs to be updated with Changelog and version bump from the 2.1.0.pre.3 release.

My fix is to update it.


deivid-rodriguez:
@bundlerbot merge

PR #7450, opened 2019-11-22
Backport ruby-core changes

deivid-rodriguez:

The problem was that we haven't incorporated recent fixes for integration in ruby-core.

My fix is to cherry-pick the fixes.


deivid-rodriguez:
@bundlerbot merge

PR #7451, opened 2019-11-22
Windows spec fixes

deivid-rodriguez:

The problem was that we have a bunch of spec failures on Windows (~315).

My fix is to improve some stuff to be more OS-independent. This should bring the failures down to ~305.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,3 +1,12 @@
+trigger:
+  branches:
+    include:
+      - master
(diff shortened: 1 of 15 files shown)


deivid-rodriguez:
Added an extra change here to stop building in Azure Pipelines twice for every push to a PR. It should speed up our CI feedback from Azure.

deivid-rodriguez:
@bundlerbot merge

PR #7452, opened 2019-11-22
Bump rubocop and rubocop-performance

deivid-rodriguez:

The problem was not really a problem, but it's good to keep our dependencies up to date.

My fix is to bump rubocop and rubocop-performance.


deivid-rodriguez:
@bundlerbot merge

Issue #7431, opened 2019-11-08
Bundle remove does not remove multi line gems


luis1van:
Using bundle remove for a gem that is written in a multi line format wont work as the remove_gems_from_gemfile method will only match the name of the gem and remove that line the proper behavior would be to remove all the lines relevant to the gem on the Gemfile.

Gemfile:
```ruby
gem 'git'
gem 'some-gem',
    git: 'ssh://gitlab-server.com/project/repo.git',
    branch: 'master'
gem 'nokogiri'
```
Example: bundle remove some-gem
Expected Result:
```ruby
gem 'git'
gem 'nokogiri'
```
Actual result:
```ruby
gem 'git'
    git: 'ssh://gitlab-server.com/project/repo.git',
    branch: 'master'
gem 'nokogiri'
```

PR #7454, opened 2019-11-26
removing multiline gem specifications correctly

jethroo:

This PR fixes the issue #7431 where gems specified in multiple lines were not removed correctly.

The remove was implemented in a way that the match was only done for a line each. 

Now upon line removal it is checked if the gem specification was finished in this line. If there is a trailing `,` comma further lines needs to be removed.

### DIFF
--- a/spec/commands/remove_spec.rb
+++ b/spec/commands/remove_spec.rb
@@ -44,6 +44,30 @@ RSpec.describe "bundle remove" do
           source "#{file_uri_for(gem_repo1)}"
         G
       end
+
+      context "when gem is specified in multiple lines" do
+        it "shows success for removed gem" do
+          gemfile <<-G
+            source '#{file_uri_for(gem_repo1)}'
+
+            gem 'git'
+            gem 'rack',
+                git: 'https://github.com/rack/rack',
+                branch: 'master'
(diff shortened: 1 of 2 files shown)


spec/commands/remove_spec.rb
```diff
@@ -44,6 +44,25 @@
           source "#{file_uri_for(gem_repo1)}"
         G
       end
+
+      context "when gem is specified in multiple lines" do
+        it "shows success for removed gem" do
+          gemfile <<-G
+            source '#{file_uri_for(gem_repo1)}'
+
+            gem 'rack',
```
hsbt: Can you add the extra gems that are the original report?

Like this:

```
gem 'git'
gem 'some-gem',
    git: 'ssh://gitlab-server.com/project/repo.git',
    branch: 'master'
gem 'nokogiri'
```

And add expectation for git and nokogiri.

jethroo: thanks for the feedback, sure would give more test context, will do soonish ;)

deivid-rodriguez:
Haven't reviewed this, but I agree with fixing this :+1:.

However, this kind of maintenance burden reenforces my position that we should not accept features like #6623 or #6597.

PR #7455, opened 2019-11-26
Lazily load `open3`

deivid-rodriguez:

The problem was `open3` will be gemified in ruby 2.7, and since we use it inside `bundler`, we might activate a version causing a conflict with the user's choice.

My diagnosis was that only loading it when needed should be better.

My fix is to lazily load it.

I expect this PR to fix [some of the errors](https://travis-ci.org/bundler/bundler/jobs/615940817) currently happening in our CI against ruby-head.


deivid-rodriguez:
@bundlerbot merge

PR #7456, opened 2019-11-27
Lazily load CGI

deivid-rodriguez:

The problem was that our ruby-head build is failing in CI.

My diagnosis was that since `cgi` is now a default gem on ruby 2.7, we're getting someunintended activations of the new default gem inside our specs.

My fix is to lazily load CGI.


deivid-rodriguez:
@bundlerbot merge

PR #7457, opened 2019-11-27
Remove unneeded CI builds

deivid-rodriguez:

The problem was we're building more stuff than necessary in CI.

My diagnosis was that we don't need to build our protected branches because they are
guaranteed to always be green, and we don't push directly to them.

My fix is to remove the relevant branches from TravisCI whitelist.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,11 +9,8 @@ before_script:
 
 branches:
   only:
-    - master
     - staging
     - trying
-    - /.+-dev$/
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Oops, I misunderstood TravisCI docs. This whitelist not only controls which pushes are built, but also the potential target branches for PRs that will get built. So, removing `master` makes PRs to no longer be built :/

Closing.

PR #7459, opened 2019-11-28
Unskip quality specs

deivid-rodriguez:

The problem was after 5003a436765eb50d9208ffdce4bbe46cf0d6d63b backported from ruby-core, quality_specs are no longer run, not even in our repo.

My diagnosis was that the condition to check whether the root folder is a git folder was incorrect.

My fix is to correct the condition.


deivid-rodriguez:
@bundlerbot merge

PR #7460, opened 2019-11-29
Vendor `uri` library

deivid-rodriguez:

The problem was that after the gemification of the `uri` library (which will happen in ruby 2.7), `bundler` will activate the default version of the new library inside its own `bundler/setup` code. That means users won't be able to specify the version of the library they want/need to use in their own Gemfiles.

My diagnosis was that we should avoid using `uri` inside `bundler/setup` code.

My fix is to vendor the `uri` library, like we did with `fileutils`.

### DIFF
--- a/lib/bundler/cli.rb
+++ b/lib/bundler/cli.rb
@@ -784,7 +784,7 @@ module Bundler
       return unless SharedHelpers.md5_available?
 
       latest = Fetcher::CompactIndex.
-               new(nil, Source::Rubygems::Remote.new(URI("https://rubygems.org")), nil).
+               new(nil, Source::Rubygems::Remote.new(Bundler::URI("https://rubygems.org")), nil).
                send(:compact_index_client).
                instance_variable_get(:@cache).
                dependencies("bundler").
--- a/Rakefile
+++ b/Rakefile
@@ -268,6 +268,14 @@ else
       sublib.prefix = "Bundler"
       sublib.vendor_lib = "lib/bundler/vendor/connection_pool"
     end
+
+    lib.dependency("uri") do |sublib|
+      sublib.version = "master"
+      sublib.download = { :github => "https://github.com/ruby/uri" }
+      sublib.namespace = "URI"
+      sublib.prefix = "Bundler"
+      sublib.vendor_lib = "lib/bundler/vendor/uri"
+    end
   end
 end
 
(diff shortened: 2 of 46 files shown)


deivid-rodriguez:
This is now green against ruby-head, but this is still WIP since I want to clean things up, review some stuff, and extract some unrelated improvements to a separate PR.

deivid-rodriguez:
This should be ready now, and it gets build against ruby-head back to green.

Only concern is that I had to change some unit tests for methods that now return or receive `Bundler::URI`'s instead of `URI`'s. I think this should be fine since these are internal methods not meant to be used by end users (I believe).

lib/bundler/cli.rb
```diff
@@ -784,7 +784,7 @@ def warn_on_outdated_bundler
       return unless SharedHelpers.md5_available?
 
       latest = Fetcher::CompactIndex.
-               new(nil, Source::Rubygems::Remote.new(URI("https://rubygems.org")), nil).
+               new(nil, Source::Rubygems::Remote.new(Bundler::URI("https://rubygems.org")), nil).
                send(:compact_index_client).
                instance_variable_get(:@cache).
                dependencies("bundler").
```
deivid-rodriguez: I don't think all of these changes in `lib` are actually necessary because the `Bundler::URI` constant will beat `URI` anyways, but maybe they make it more explicit that we're using a vendored version of the `uri` gem?

lib/bundler/cli.rb
```diff
@@ -784,7 +784,7 @@ def warn_on_outdated_bundler
       return unless SharedHelpers.md5_available?
 
       latest = Fetcher::CompactIndex.
-               new(nil, Source::Rubygems::Remote.new(URI("https://rubygems.org")), nil).
+               new(nil, Source::Rubygems::Remote.new(Bundler::URI("https://rubygems.org")), nil).
```
deivid-rodriguez: I don't think all of these changes in lib are actually necessary because the Bundler::URI constant will beat URI anyways, but maybe they make it more explicit that we're using a vendored version of the uri gem?

hsbt: I think we should replace vendored `fileutils` and `uri` to the minimum code like [yaml_serializer](https://github.com/bundler/bundler/blob/master/lib/bundler/yaml_serializer.rb) in the next version. 

But there is no enough time to release Bundler 2.1. We can try it Bundler 2.2 or Bundler 3.

deivid-rodriguez: Thanks, I agree and also started some work to be able to vendor only parts of libraries while still automating the process with `automatiek` :+1:

deivid-rodriguez:
@bundlerbot r=hsbt

PR #7461, opened 2019-11-29
Set parallel_tests to output progress in the same format as regular RSpec

colby-swandale:
# What was the end-user problem that led to this PR?

When running tests using the `parallel_test` tool, there will be no output on the terminal to show the progress or result of the test suit.

```
$ ./bin/parallel_rspec spec         
8 processes for 162 specs, ~ 20 specs per process

Took 576 seconds (9:36)
```

You have to go diving into a file that's located in `tmp/parallel_runtime_rspec.log`.

It would be better to have the output that we get from the regular `./bin/rspec` and show the progress/result of the test suit.

Run the parallel test suit using `./bin/parallel_rspec spec`

Add the additional `progress` formatter to the parallel rspec configuration

### DIFF
--- a/.rspec_parallel
+++ b/.rspec_parallel
@@ -1,3 +1,4 @@
+--format progress
 --format ParallelTests::RSpec::RuntimeLogger --out tmp/parallel_runtime_rspec.log
 --require spec_helper
 --require support/parallel.rb


deivid-rodriguez:
The current formatter is used to balance test times across thread/processes.

I think `parallel_tests` support multiple formatters, so if we want another formmatter, I think we can add it as a separate configuration?

Also, are _failures_ not printed with the current format?

colby-swandale:
Oh, It's configured at the moment to record the test time and to balance out the different processors. There's a doc that uses both the `RuntimeLogger` and the `progress` formatter.

https://github.com/grosser/parallel_tests#rspec

deivid-rodriguez:
@bundlerbot merge

PR #7463, opened 2019-11-30
Vendor fileutils 1.4.1

deivid-rodriguez:

The problem was no big deal. Just a discrepancy between the fileutils version we're vendoring and the one that will ship with ruby 2.7. This doesn't cause any issues, but I prefer that they match.

My fix is to `bin/rake vendor:fileutils[v1.4.1]` and commit the result.


deivid-rodriguez:
This also fixes some `untaint` warnings in ruby 2.7. Merging!

@bundlerbot merge

PR #7464, opened 2019-12-01
Improve rubygems switcher

deivid-rodriguez:

The problem was we have some code in our tests to make sure that the `ENV["RGV"]` variable we use to set the rubygems version we want to test bundler against is always respected, including every subprocess our tests start. However:

* The code was overly complicated.
* It didn't support custom branches, for example, `RGV=lazily_load_uri`. This is a feature I find very useful when a PR you're working on also need some changes in the rubygems side, like it happened to me at #7460.

My diagnosis was that all the code needs to do is:

* Set up the location of the rubygems code we'll test bundler against, be it a path, a branch, or a released version.
* Modify `RUBYOPT` to include that location in the LOAD_PATH, so that `rubygems` is always loaded from there instead of the system's version.

My fix is to do just that, remove all the stuff that wasn't needed, and do a bit of renaming to improve readability.

### DIFF
--- a/spec/rubygems/rubygems.rb
+++ b/spec/rubygems/rubygems.rb
@@ -1,9 +1,0 @@
-# frozen_string_literal: true
-
-require_relative "../support/rubygems_version_manager"
-
-RubygemsVersionManager.new(ENV["RGV"]).switch
-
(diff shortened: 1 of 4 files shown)


deivid-rodriguez:
@bundlerbot merge

PR #7465, opened 2019-12-01
December manpages

deivid-rodriguez:

The problem was that man pages are out of date.

My fix is to update them.


deivid-rodriguez:
@bundlerbot merge

PR #7471, opened 2019-12-06
Vendor latest thor's master

deivid-rodriguez:

The problem was that I need to pull some changes from latest `thor` master to avoid loading the `uri` library for #7460.

My fix is to run `bin/rake vendor:thor[master]` and commit changes.

Not that I used an `automatiek` version including https://github.com/segiddins/automatiek/pull/7 that avoids some false positive when replacing namespaces.


deivid-rodriguez:
Just updating upstream changes, and no issues found, so merging.

@bundlerbot merge

PR #7473, opened 2019-12-06
Cleanup some unnecessary code

deivid-rodriguez:

The problem was #7460 is very big so I want to extract these changes to a separate PR, so that we're more aware of them.

My diagnosis was that this code can be removed. In particular, in the `GemRemoteFetcher` class there was the following comment

https://github.com/bundler/bundler/blob/25595896eb0f8dfd004d941093bf1d8f4a39aeeb/lib/bundler/gem_remote_fetcher.rb#L9-L10

After having a look, I think the comment would make sense if where it says "gemstash", it actually meant "bundler". That would make sense to me since this is about fetching remote specs, so I would assume it's the client side running it.

Assuming this is the correct interpretation, we can remove the code since our minimum supported rubygems version is 2.5.2, and this code was included in rubygems in 2.5.0.

My fix is to remove the `GemRemoteFetcher` class, plus simplify other related code.


deivid-rodriguez:
No feedback, so I'll do this. It should be ok, I think.

deivid-rodriguez:
@bundlerbot merge

indirect:
late feedback, but I like this :)

PR #7474, opened 2019-12-09
Fix `bundle exec rake install` failing

deivid-rodriguez:

The problem was that I noticed `bundle exec rake install` was failing on my gem using the latest prerelease of bundler.

My diagnosis was that `bundler` silences output of rubygems by default, and in a `bundle exec` context, `bundler/setup` is always required by subprocesses. So, the `rake install` task is silencing rubygems, but it requires it to not be silent:

https://github.com/bundler/bundler/blob/bada03dd6d4d15828fb5b2cf7f744948e88a69a3/lib/bundler/gem_helper.rb#L91-L92

My fix is to wrap the execution of the rubygems subprocess with `Bundler.with_original_env` so that the `gem` command runs without `bundler` involved.

### DIFF
--- a/lib/bundler/gem_helper.rb
+++ b/lib/bundler/gem_helper.rb
@@ -86,8 +86,8 @@ module Bundler
       built_gem_path ||= build_gem
       cmd = "#{gem_command} install #{built_gem_path}"
       cmd += " --local" if local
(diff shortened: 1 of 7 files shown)


deivid-rodriguez:
@bundlerbot merge

PR #7478, opened 2019-12-13
Revert "Write & read dummy gems binarily"

deivid-rodriguez:
This reverts commit f11c4757a17bea029e1a83ad67c425ccf16133ff.

The problem was that since #7451, azure logs are flooded with messages like

```
The file will have its original line endings in your working directory
warning: LF will be replaced by CRLF in foo.gemspec.
```

This makes it hard to scan for other potential errors.

My diagnosis was that some commit in that PR introduced this warnings.

My fix is to revert the offending commit.


deivid-rodriguez:
@bundlerbot merge

PR #7479, opened 2019-12-13
Get ready for 2.1.0 release

deivid-rodriguez:

The problem was we want to release bundler 2.1.0 before ruby 2.7.0 ships, so that bundler 2.1.0 can be included with ruby 2.7.0.

My fix is to get ready for the release.


deivid-rodriguez:
@bundlerbot merge

PR #7480, opened 2019-12-13
Vendor thor 1.0.0

deivid-rodriguez:

The problem was that `thor` 1.0.0 has been released, so we should vendor it.

My fix is to vendor it.


deivid-rodriguez:
@bundlerbot r=hsbt

deivid-rodriguez:
Missed one change.

@bundlerbot r-

deivid-rodriguez:
@bundlerbot merge

PR #7483, opened 2019-12-14
Delegate should be supported now

deivid-rodriguez:

The problem was that delegate is now supported since `did_you_mean` drop its dependency on it at https://github.com/ruby/ruby/commit/e2708068ad65f7f9986adf4fb3a4fa660f430a5a.

My diagnosis was that we can test that arbitrary `delegate` versions can be included in Gemfiles again.

My fix removes `delegate` from the exemptions.


deivid-rodriguez:
@bundlerbot merge

PR #7493, opened 2019-12-16
Fix another silent rubygems issue

### LINKED ISSUE #7490: `gem content NAME` outputs nothing after upgrading to 2.1.0
I've just updated to 2.1 a system that uses Bundler to install a bunch of gems. This system calls scripts/CMake code which itself tries to detect some gem parameters using `gem content`. Until 2.1 this worked fine.

I've locally upgraded to 2.1 today, and now `gem content NAME` just outputs nothing. The system I'm maintaining is using Bundler programatically, so I wonder if it could be an expected change of behavior ?

From the command line, if I run

~~~
RUBYOPT=-r/home/doudou/.local/share/autoproj/gems/ruby/2.5.0/gems/bundler-2.1.0/lib/bundler/setup \
  GEM_HOME=/home/doudou/.local/share/autoproj/gems/ruby/2.5.0 \
  BUNDLE_GEMFILE=/home/doudou/dev/build_area/bundler2.1/install/gems/Gemfile \
  gem2.5 content rice
~~~

I get no output. Removing the RUBYOPT solves it. Would anyone have an idea about why this started appearing ?

deivid-rodriguez:

The problem was that we still have some issues where using the `gem` script on a `bundler` context is silenced.

My diagnosis was that `bundle exec` adds `-rbundler/setup` to RUBYOPT which silences `rubygems` output, and that we need to reset `rubygems` UI after that so that if we end up shelling out to `gem`, it is not silent. The previous approach worked for loading the `gem` script in-process, but didn't work in other case.

My fix is to reset rubygems UI right after `bundler/setup`.

I chose this fix because it fixes the problem independently of the rubygems version being run, but we can probably also fix this more cleanly inside the `gem` script by adding something like `Gem::DefaultUserInteraction.ui = Gem::ConsoleUI.new` at the top of the script.

Fixes #7490.

### DIFF
--- a/lib/bundler/cli/exec.rb
+++ b/lib/bundler/cli/exec.rb
@@ -25,12 +25,12 @@ module Bundler
       SharedHelpers.set_bundle_environment
       if bin_path = Bundler.which(cmd)
         if !Bundler.settings[:disable_exec_load] && ruby_shebang?(bin_path)
-          return with_verbose_rubygems { kernel_load(bin_path, *args) }
+          return kernel_load(bin_path, *args)
         end
-        with_verbose_rubygems { kernel_exec(bin_path, *args) }
+        kernel_exec(bin_path, *args)
       else
         # exec using the given command
-        with_verbose_rubygems { kernel_exec(cmd, *args) }
+        kernel_exec(cmd, *args)
       end
     end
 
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
@doudou Does this work for you?

doudou:
> @doudou Does this work for you?

Seems to work, yes. Thanks !

deivid-rodriguez:
Thanks for confirming! I'll get this in and try to get a bug fix release as soon as possible.

@bundlerbot merge

doudou:
> Thanks for confirming! I'll get this in and try to get a bug fix release as soon as possible.
> 
> @bundlerbot merge

Hi ...

Would you maybe have an ETA on how fast the point release might get out ? It's pretty bad for us.

deivid-rodriguez:
Next day or two.

PR #7497, opened 2019-12-17
Merge changelog and version bump back to master

deivid-rodriguez:

The problem was that version and changelog need to be cherry-picked from the stable branch.

My fix is to do just that.


deivid-rodriguez:
@bundlerbot r=hsbt

PR #7502, opened 2019-12-17
Release 2.1.1

deivid-rodriguez:

The problem was I want to get regression fixed as soon as possible to reduce impact.

My fix is to release what we have already fixed as soon as possible.


deivid-rodriguez:
@bundlerbot merge

PR #7503, opened 2019-12-18
Bump version and changelog after release

deivid-rodriguez:

The problem was changelog and version bump not yet in master.

My fix is to merge back changelog and version bump after the release from the stable branch.


deivid-rodriguez:
@bundlerbot merge

PR #7505, opened 2019-12-18
Make sure to `require "rubygems"` explicitly

deivid-rodriguez:
### What was the end-user problem that led to this PR?

The problem was some setups have `--disable=gem` in `RUBYOPT`, and sometimes use `bundler` without modifying that env.

### What was your diagnosis of the problem?

My diagnosis was that we shouldn't break those setups just to save a `require`.

### What is your fix for the problem, implemented in this PR?

My fix is to add a `require "rubygems"` on top of the file that needs it.

Fixes #7487.

PR #7506, opened 2019-12-19
Remove unnecessary CI steps

deivid-rodriguez:

The problem was that merging #7505 failed.

My diagnosis was that some `apt` operation installing `graphviz` failed, but installing `graphviz` was unnecessary in the specific job that failed.

My fix is to only modifying the OS when strictly necessary.

Also fixed another test issue with the recently added test for integration with the `rubygems-bundler` gem by `rvm`.


deivid-rodriguez:
@bundlerbot merge

PR #7508, opened 2019-12-19
Release 2.1.2

deivid-rodriguez:

The problem was that I want to ship [a small regression fix](https://github.com/bundler/bundler/pull/7505), and also so tweaks to the development environment (#7506) that I want to use in the rubygems side to fix the CI errors in https://github.com/rubygems/rubygems/pull/3043.

My fix is to release 2.1.2


deivid-rodriguez:
@bundlerbot merge

PR #7509, opened 2019-12-19
Add `travis_retry` to apt-get operation

deivid-rodriguez:

The problem was another failed build due to [apt-get install failing](https://travis-ci.org/bundler/bundler/jobs/627348743).

My diagnosis was TravisCI networking is unstable at the moment.

Not a permanent fix, since that doesn't depend on us, but using `travis_retry` should increase reliability.


deivid-rodriguez:
@bundlerbot merge

PR #7513, opened 2019-12-20
Improve error message when not logged in

steveklabnik:
`gem push` makes you add other parameters, and is awkward to figure out the details. `gem signin` is a more direct way of doing this.


deivid-rodriguez:
@bundlerbot merge

PR #7514, opened 2019-12-20
Fix `rake build` when path has spaces on it

### LINKED ISSUE #7512: Bundler >= 2.1.1 ERROR:  While executing gem ... (Gem::CommandLineError) Too many gem names
We recently ran into a problem when we updated to bundler 2.1.1 and ran the following command:

```
jenkins@vpszap6s:~/jobs/is-babylon_scene Build and Release/workspace$ bundle _2.1.1_ exec rake build
```

output:
```
rake aborted!
ERROR:  While executing gem ... (Gem::CommandLineError)
    Too many gem names (/var/lib/jenkins/jobs/is-babylon_scene, Build, and, Release/workspace/is-babylon_scene.gemspec); please specify only one
/usr/local/rvm/gems/ruby-2.6.5/gems/bundler-2.1.1/lib/bundler/gem_helper.rb:190:in `sh'
/usr/local/rvm/gems/ruby-2.6.5/gems/bundler-2.1.1/lib/bundler/gem_helper.rb:76:in `build_gem'
/usr/local/rvm/gems/ruby-2.6.5/gems/bundler-2.1.1/lib/bundler/gem_helper.rb:40:in `block in install'
/usr/local/rvm/gems/ruby-2.6.5/gems/rake-13.0.1/exe/rake:27:in `<top (required)>'
/usr/local/rvm/gems/ruby-2.6.5/bin/ruby_executable_hooks:24:in `eval'
/usr/local/rvm/gems/ruby-2.6.5/bin/ruby_executable_hooks:24:in `<main>'
Tasks: TOP => build
```

Turns out it's because we're running it from a path which contains white spaces. The exact same error can be reproduced when running `rake build` with bundle _2.1.2_ but it seems to be working on older versions, such as 2.0.2. Can you please fix this? Thanks in advance!

deivid-rodriguez:

The problem was that `rake build` can no longer be run from a path that has spaces on it.

My diagnosis was that we need to escape the path for the shell somewhere.

My fix is to add `.shellescape` at the right place.

I chose this fix because it fixes the problem.

Fixes #7512.

### DIFF
--- a/lib/bundler/gem_helper.rb
+++ b/lib/bundler/gem_helper.rb
@@ -73,7 +73,7 @@ module Bundler
 
     def build_gem
       file_name = nil
-      sh("#{gem_command} build -V #{spec_path}".shellsplit) do
+      sh("#{gem_command} build -V #{spec_path.shellescape}".shellsplit) do
         file_name = File.basename(built_gem_path)
         SharedHelpers.filesystem_access(File.join(base, "pkg")) {|p| FileUtils.mkdir_p(p) }
         FileUtils.mv(built_gem_path, "pkg")
--- a/spec/runtime/gem_tasks_spec.rb
+++ b/spec/runtime/gem_tasks_spec.rb
@@ -57,6 +57,20 @@ RSpec.describe "require 'bundler/gem_tasks'" do
     expect(err).to be_empty
   end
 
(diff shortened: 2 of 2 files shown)


deivid-rodriguez:
@bundlerbot r=colby-swandale

PR #7515, opened 2019-12-22
Move `bin/rake man:check` to `linting` phase

deivid-rodriguez:
So that it's only run once.

The problem was that `bin/rake man:check` is run for every CI matrix entry and that's useless.

My diagnosis was that we should move the check to the linting phase and run it only once, like we did for `rubocop` at https://github.com/bundler/bundler/pull/6830.

My fix is to do that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,13 +54,7 @@ env:
 jobs:
   include:
     - rvm: 2.6.5
-      script: bin/rake rubocop
-      stage: linting
-    - rvm: 2.6.5
-      script: bin/rake check_rvm_integration
-      stage: linting
-    - rvm: 2.6.5
-      script: bin/rake man:check
+      script: bin/rake rubocop check_rvm_integration man:check
       stage: linting
     # Ruby 2.3 also tested in 2.x mode
     - rvm: 2.3.8
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ before_script:
   - travis_retry gem install rake:"~> 12.0"
   - travis_retry bin/rake override_version
   - travis_retry bin/rake spec:deps
-  - bin/rake man:check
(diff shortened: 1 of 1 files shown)


colby-swandale:
Is it worth making a script to run rubocop, `check_rvm_integration` and `man:check` within a single instance to help free up instances that can be used to run specs?

simi:
@colby-swandale it could get tricky to run all of them and collect all errors in some reasonable way. I think it is not worth it saving that 2* 1 minute needed for linting phase.

.travis.yml
```diff
@@ -55,10 +54,7 @@ env:
 jobs:
   include:
     - rvm: 2.6.5
-      script: bin/rake rubocop
-      stage: linting
-    - rvm: 2.6.5
-      script: bin/rake check_rvm_integration
+      script: bin/rake rubocop check_rvm_integration man:check
```
simi: Does this continue on first fail? Let's say `rubocop` will fail, is `man:check` triggered as well?

deivid-rodriguez: No, it aborts.

deivid-rodriguez:
I put them all in the same command as per @colby-swandale's suggestion, but I don't have a strong opinion either way.

.travis.yml
simi: Then the question is if you would like to trade 1 minute for a little less comfort. Having multiple jobs in linting phase makes it easier to see problems (since you'll see which job failed really quickly without need to open the log).

deivid-rodriguez: Yeah, as I said, I don't have a strong opinion either way.

deivid-rodriguez:
Let's do it like this for now. If it turns out to be inconvenient I'll split the jobs again.

@bundlerbot r=colby-swandale

colby-swandale:
@deivid-rodriguez you can always change the order and put rubocop last

PR #7516, opened 2019-12-22
Bump version and changelog

deivid-rodriguez:

The problem was master branch doesn't yet include the version bump and changelog from 2.1.2.

My fix is to include those.


deivid-rodriguez:
@bundlerbot merge

PR #7519, opened 2019-12-24
Make `bundle config deployment true` equivalent to `bundle install --deployment` in regards to configuration

deivid-rodriguez:

The problem was that while addressing comments for https://github.com/docker-library/official-images/pull/7188, I noticed that the alternative we're recommending for `bundle install --deployment`, i.e., `bundle config set deployment true`, is not equivalent to it.

My diagnosis was that whereas `bundle install --deployment` configures bundler to be frozen AND to install gems to `vendor/bundle`, `bundle config deployment true` only configures bundler to be frozen.

My fix is to make `bundle config deployment true`  behave just like `bundle install --deployment` in regards to configuration. 

I chose this fix because all the commands we're suggesting as alternatives for deprecations should be actual alternatives that work exactly in the same way.

Also, note that there's a change scheduled for bundler 3 where the `deployment` configuration will only mean `frozen`. But for now, we should focus on making sure that people moves away from CLI flags in favor of configuration, so we might want to delay that change to make things less confusing.

### DIFF
--- a/lib/bundler/cli/install.rb
+++ b/lib/bundler/cli/install.rb
@@ -38,7 +38,8 @@ module Bundler
         if Bundler.feature_flag.deployment_means_frozen?
           Bundler.settings.set_command_option :deployment, true
         else
-          Bundler.settings.set_command_option :frozen, true
+          Bundler.settings.set_command_option :deployment, true if options[:deployment]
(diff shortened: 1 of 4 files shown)


deivid-rodriguez:
@bundlerbot merge

PR #7522, opened 2019-12-29
Improve platform specific gem resolution

kou:

Platform specific gems aren't resolved when platform specific gems are conflicted.

For example, in the following situation, foo-1.0.0-x64-mingw32 that requires bar<1 is conflicted because there is no bar<1. Without this change, Bundler raises a conflict error. But users can use foo-1.0.0 (no x64-mingw32) in this situation. With this change, Bundler resolves to foo-1.0.0 (no x64-mingw32).

```ruby
@index = build_index do
  gem "bar", "1.0.0"
  gem "foo", "1.0.0"
  gem "foo", "1.0.0", "x64-mingw32" do
    dep "bar", "< 1"
  end
end
dep "foo"
platforms "x64-mingw32"
````

See also #6247. This change includes the specs that were added in #6247.

Not platform specific gem (foo-1.0.0 in the above case) isn't tried to be resolved when platform specific gem (foo-1.0.0-x64-mingw32 in the above case) is available.

In this PR, not platform specific gem (foo-1.0.0 in the above case) is also tried to be resolved even when platform specific gem (foo-1.0.0-x64-mingw32 in the above case) is available. Priority is "platform specific gem" -> "not platform specific gem". So platform specific gem is usable, platform specific gem is used. Not platform specific gem is used as fallback.

`search_for` represents this. Here is the `search_for` specification:
https://github.com/bundler/bundler/blob/master/lib/bundler/vendor/molinillo/lib/molinillo/modules/specification_provider.rb#L11-L13

> Search for the specifications that match the given dependency. The specifications in the returned array will be considered in reverse order, so the latest version ought to be last.

### CODE REVIEW

**Comment by kou** [comment]
Oh, I need more codes....

**Comment by kou** [comment]

**Comment by kou** [comment]
Umm. I need to care `specific_platform` feature flag.

File: lib/bundler/definition.rb [hunk 1]
```diff
@@ -228,12 +232,13 @@ def requested_specs
     end
 
     def current_dependencies
-      dependencies.select(&:should_include?)
+      dependencies.select do |d|
+        d.should_include? && !d.gem_platforms(@platforms).empty?
```
kou: `!d.gem_platforms(@platforms).empty?` is the same check done in `expand_dependencies` that removes dependencies not for the target platforms. For example, `nokogiri` gem is removed for `gem "nokogiri", :platforms => [:mingw, :mswin, :x64_mingw, :jruby]` with `forge_ruby_platform` configuration in `spec/runtime/platform_spec.rb:107`.

lib/bundler/definition.rb
```diff
@@ -706,9 +711,6 @@ def converge_dependencies
         elsif dep.source
           dep.source = sources.get(dep.source)
         end
-        if dep.source.is_a?(Source::Gemspec)
-          dep.platforms.concat(@platforms.map {|p| Dependency::REVERSE_PLATFORM_MAP[p] }.flatten(1)).uniq!
-        end
```
kou: This is a part of reverting 0a8ca4879e0b79aa4109a0dc424940b079ef38d0.
See also: #4150 and #4102.

lib/bundler/definition.rb
```diff
@@ -907,10 +909,16 @@ def expand_dependencies(dependencies, remote = false)
       deps
     end
 
+    def dependencies_for(groups)
+      current_dependencies.reject do |d|
+        (d.groups & groups).empty?
+      end
+    end
```
kou: It's a helper method to use the common code in this class to choose valid dependencies.

lib/bundler/dependency.rb
```diff
@@ -74,15 +74,6 @@ class Dependency < Gem::Dependency
...
     }.freeze
 
-    REVERSE_PLATFORM_MAP = {}.tap do |reverse_platform_map|
-      PLATFORM_MAP.each do |key, value|
-        reverse_platform_map[value] ||= []
-        reverse_platform_map[value] << key
-      end
-
-      reverse_platform_map.each {|_, platforms| platforms.freeze }
-    end.freeze
```
kou: This is a part of reverting 0a8ca4879e0b79aa4109a0dc424940b079ef38d0.
See also: #4150 and #4102.

lib/bundler/dsl.rb
```diff
@@ -75,8 +75,7 @@ def gemspec(opts = nil)
 
         @gemspecs << spec
 
-        gem_platforms = Bundler::Dependency::REVERSE_PLATFORM_MAP[Bundler::GemHelpers.generic_local_platform]
-        gem spec.name, :name => spec.name, :path => path, :glob => glob, :platforms => gem_platforms
+        gem spec.name, :name => spec.name, :path => path, :glob => glob
```
kou: This is a part of reverting 0a8ca4879e0b79aa4109a0dc424940b079ef38d0.
See also: #4150 and #4102.

spec/bundler/dsl_spec.rb
```diff
@@ -174,41 +174,6 @@
...
-      let(:platform) { "java" }
-
-      it "keeps track of the jruby platforms in the dependency" do
-        allow(Gem::Platform).to receive(:local).and_return(java)
-        subject.gemspec
-        expect(subject.dependencies.last.platforms).to eq(Bundler::Dependency::REVERSE_PLATFORM_MAP[Gem::Platform::JAVA])
-      end
-    end
-  end
-
```
kou: This is a part of reverting 0a8ca4879e0b79aa4109a0dc424940b079ef38d0.
See also: #4150 and #4102.

lib/bundler/resolver/spec_group.rb
```diff
@@ -5,10 +5,12 @@ class Resolver
     class SpecGroup
       include GemHelpers
 
+      attr_reader :all_specs
       attr_accessor :name, :version, :source
       attr_accessor :ignores_bundler_dependencies
 
       def initialize(all_specs)
+        @all_specs = all_specs
```
kou: This is used to create a copy of `SpecGroup` in `Resolver`.
We can add a copy method to `SpecGroup` as another approach. Which approach do you like?

Note that we can't use `#dup` because it copies all instance variables that include `@activated_platforms`. We want a copy that just have the same `all_specs` and `ignores_bundler_dependencies`.

lib/bundler/resolver/spec_group.rb
```diff
@@ -37,13 +39,17 @@ def activate_platform!(platform)
         @activated_platforms << platform
       end
 
+      def spec(platform)
```
kou: Another method name candidate is `for` because we use `for?(platform)`.
But `for` may confuse us because `for` is one of reserved word in Ruby.

lib/bundler/resolver/spec_group.rb
```diff
@@ -58,18 +64,26 @@ def ==(other)
...
           version == other.version &&
+          sorted_activated_platforms == other.sorted_activated_platforms &&
           source == other.source
       end
 
       def eql?(other)
         return unless other.is_a?(SpecGroup)
         name.eql?(other.name) &&
           version.eql?(other.version) &&
+          sorted_activated_platforms.eql?(other.sorted_activated_platforms) &&
```
kou: This is one of the most important changes in this pull request.
Without this change, `SpecGroup`s that have different activated platforms are the same `SpecGroup`.
With this change, `SpecGroup`s that have different activated platforms are the different `SpecGroup`s.

By this change, we can choose a gem that is conflicted with `x64-mingw32` version but is available with `ruby` version. See added examples into `spec/resolver/platform_spec.rb` for details.

spec/install/gems/resolving_spec.rb
```diff
@@ -150,6 +150,13 @@
 
       let(:ruby_requirement) { %("#{RUBY_VERSION}") }
       let(:error_message_requirement) { "~> #{RUBY_VERSION}.0" }
+      let(:error_message_platform) do
+        if Bundler.feature_flag.specific_platform?
+          " #{Bundler.local_platform}"
+        else
+          ""
+        end
+      end
```
kou: Only this is the incompatible behavior.
The current local platform information is shown in error message with `specific_platform` feature flag because we try platform specific version as the most priority candidate.
Is this acceptable?

spec/resolver/platform_spec.rb
```diff
@@ -28,6 +28,60 @@
...
+    @index = build_index do
+      gem "foo", "1.0.0"
+      gem "foo", "1.0.0", "x64-mingw32"
+      gem "foo", "1.1.0"
+    end
+    dep "foo"
+    platforms "x64-mingw32"
+
+    should_resolve_as %w[foo-1.1.0]
+  end
```
kou: Added in #6247.

spec/resolver/platform_spec.rb
```diff
@@ -28,6 +28,60 @@
...
+      gem "foo", "1.0.0"
+      gem "foo", "1.0.0", "x64-mingw32" do
+        dep "bar", "< 1"
+      end
+    end
+    dep "foo"
+    platforms "x64-mingw32"
+
+    should_resolve_as %w[foo-1.0.0]
+  end
```
kou: Added in #6247.

spec/resolver/platform_spec.rb
```diff
@@ -28,6 +28,60 @@
...
+        s.required_ruby_version = [">= 2.0", "< 2.4"]
+      end
+      gem "ruby\0", "2.5.1"
+    end
+    dep "foo"
+    dep "ruby\0", "2.5.1"
+    platforms "x64-mingw32"
+
+    should_resolve_as %w[foo-1.1.0]
+  end
```
kou: Added in #6247.

spec/resolver/platform_spec.rb
```diff
@@ -28,6 +28,60 @@
...
+  it "prefers the platform specific gem to the ruby version" do
+    @index = build_index do
+      gem "foo", "1.0.0"
+      gem "foo", "1.0.0", "x64-mingw32"
+    end
+    dep "foo"
+    platforms "x64-mingw32"
+
+    should_resolve_as %w[foo-1.0.0-x64-mingw32]
+  end
```
kou: Added for preventing regression.

spec/runtime/platform_spec.rb
```diff
@@ -113,6 +113,7 @@
       bundle! "install"
 
       expect(the_bundle).to include_gems "platform_specific 1.0 RUBY"
+      expect(the_bundle).to not_include_gems "nokogiri"
```
kou: Added for preventing regression.

lib/bundler/resolver.rb
```diff
@@ -146,7 +146,43 @@ def search_for(dependency)
...
+          if sg_ruby_platform.for?(Gem::Platform::RUBY)
+            sg_ruby_platform.activate_platform!(Gem::Platform::RUBY)
+            sg_ruby_platform.activate_platform!(platform)
+            selected_sgs << sg_ruby_platform
+          end
+        end
+        sg.activate_platform!(platform)
+        selected_sgs << sg
+      end
+      selected_sgs
```
kou: This change, `requirement_satisfied_by?` change and `self.platform_sort_key` change is the one of the most important changes for this improvement.

deivid-rodriguez:
Thanks so much @kou!

This is awesome!! I'll review this shortly. Even if this is a bug fix, I think it might be worth releasing it as 2.2.0 together with other platform improvements:

* Windows CI and test fixes.
* I'd like to considered reintroducing https://github.com/bundler/bundler/pull/7215, since I only reverted it due to these multiplatform resolution issues. Could you give me your opinion about it if you have time?

Thanks again for fixing this, I did try fix it myself in the past but never quite finished the work because it was hard :sweat_smile:.

kou:
> * I'd like to considered reintroducing #7215, since I only reverted it due to these multiplatform resolution issues. Could you give me your opinion about it if you have time?

Sure!

I confirmed #7215 and related issues: #7329, #7315, brianmario/mysql2#1064 and rails/rails#37224

We need one more improvement to reintroduce #7215. The following example should be passed:

```diff
```

larskanis:
I tried this PR with RubyInstaller-2.7.0-1-x64 on Windows Server 2016. It succeeds to run the new resolver specs, but `bundle install` still fails to resolve the following Gemfile in real world:

```
c:\Users\kanis\bundler>cat Gemfile
source "https://rubygems.org"
gem "ffi"

c:\Users\kanis\bundler>bundle install
Resolving dependencies...
Using bundler 2.2.0.dev
Fetching ffi 1.11.3 (x64-mingw32)
Installing ffi 1.11.3 (x64-mingw32)
Gem::RuntimeRequirementNotMetError: ffi requires Ruby version >= 2.2, < 2.7.dev.
The current ruby version is 2.7.0.0.
An error occurred while installing ffi (1.11.3), and Bundler cannot continue.
Make sure that `gem install ffi -v '1.11.3' --source 'https://rubygems.org/'`
succeeds before bundling.

In Gemfile:
  ffi
```

I verified that the new code paths are actually called.

deivid-rodriguez:
Thanks both so much :purple_heart:!

Just to make sure, does the remaining failing test case mentioned by @kou correspond to the realworld failing example found by @larskanis? It seems to me that the first one has an explicit maximum required ruby, but in @larskanis' case the upper bound is somehow being incorrectly considered.

deivid-rodriguez:
@kou I made the Windows CI check in master required,so if you rebase we can make sure that everything current passing is still passing on Windows, and I think with this patch we might be available to unskip some pending specs on Windows. For example, in `spec/install/gems/resolving_spec.rb` and `spec/install/gemfile/platform_spec.rb`.

Also, @larskanis, since the case you mention looks very simple, I guess that'd be a regression caused by this PR, or is it something already not working?

spec/install/gems/resolving_spec.rb
deivid-rodriguez: I think that's perfectly acceptable, yeah!

larskanis:
@deivid-rodriguez The error message [above](https://github.com/bundler/bundler/pull/7522#issuecomment-571090513) is exactly the same as on current bundler-2.1.2. It's no regression, but it doesn't do what the PR is supposed to do.

My expectation was, that with this PR the `platform=ruby` version of the gem is selected while `Resolving dependencies...`, but for some reason it isn't, so that the `platform=x64-mingw32` version is fetched. That unsurprisingly then fails at the install step.

deivid-rodriguez:
Right, and the current workaround people use to install that Gemfile under Windows is the `force_ruby_platform` config I guess?

larskanis:
Yes, it works with `force_ruby_platform`, but then doesn't consider `platform=x64-mingw32` any longer for any gems.

kou:
This is not a bug of resolver. This is a bug of `Bundler::CompactIndexClient#parse_gem`. 

We need the following patch:

```diff
diff --git a/lib/bundler/compact_index_client/cache.rb b/lib/bundler/compact_index_client/cache.rb
index f6105d3bb3..0dae227508 100644
--- a/lib/bundler/compact_index_client/cache.rb
+++ b/lib/bundler/compact_index_client/cache.rb
@@ -95,18 +95,52 @@ module Bundler
       def parse_gem(string)
         version_and_platform, rest = string.split(" ", 2)
         version, platform = version_and_platform.split("-", 2)
-        dependencies, requirements = rest.split("|", 2).map {|s| s.split(",") } if rest
-        dependencies = dependencies ? dependencies.map {|d| parse_dependency(d) } : []
-        requirements = requirements ? requirements.map {|r| parse_dependency(r) } : []
+        dependencies, requirements = rest.split("|", 2) if rest
+        dependencies = dependencies ? parse_dependencies(dependencies) : []
+        requirements = requirements ? parse_requirements(requirements) : []
         [version, platform, dependencies, requirements]
       end
 
+      def parse_dependencies(string)
+        string.split(",").map {|d| parse_dependency(d) }
+      end
+
       def parse_dependency(string)
         dependency = string.split(":")
         dependency[-1] = dependency[-1].split("&") if dependency.size > 1
         dependency
       end
 
+      # Parse the following format:
+      #
+      #   line = "checksum:#{checksum}"
+      #   line << ",ruby:#{ruby_version}" if ruby_version && ruby_version != ">= 0"
+      #   line << ",rubygems:#{rubygems_version}" if rubygems_version && rubygems_version != ">= 0"
+      #
+      # See compact_index/gem_version.rb for details.
+      def parse_requirements(string)
+        requirements = []
+        checksum = string.match(/\A(checksum):(\h+)/)
+        if checksum
+          requirements << [checksum[1], [checksum[2]]]
+          string = checksum.post_match
+          string = string[1..-1] if string.start_with?(",")
+        end
+        rubygems = string.match(/(rubygems):(.+)\z/)
+        if rubygems
+          string = rubygems.pre_match
+          string = string[1..-1] if string.start_with?(",")
+        end
+        ruby = string.match(/\A(ruby):(.+)\z/)
+        if ruby
+          requirements << [ruby[1], ruby[2].split(/\s*,\s*/)]
+        end
+        if rubygems
+          requirements << [rubygems[1], rubygems[2].split(/\s*,\s*/)]
+        end
+        requirements
+      end
+
       def info_roots
         [
           directory.join("info"),
```

kou:
I think that the fix should be done by a separated pull request.

kou:
https://github.com/bundler/bundler/pull/7554 for `Bundler::CompactIndexClient#parse_gem`.

kou:
> For example, in `spec/install/gems/resolving_spec.rb` and `spec/install/gemfile/platform_spec.rb`.

Trying.

kou:
We could unskip 2 examples in `spec/install/gemfile/platform_spec.rb`.

deivid-rodriguez:
Nice, that's something :sweat_smile:.

So, let me see if I got this:

* The realworld example about the ffi gem is not reproducible with our test setup, but it will be fixed by #7554.

* The missing functionality to be able to reintroduce #7215 would be to get the example you mentioned in https://github.com/bundler/bundler/pull/7522#issuecomment-571026880 passing. This is a different case from the ffi's realworld example and that work is yet to be done.

Is this correct?

kou:
>   * The realworld example about the ffi gem is not reproducible with our test setup, but it will be fixed by #7554.

Yes.

The following example with Ruby 2.6 on Windows may reproduce the ffi gem case:

```ruby
gemfile <<-G
  source "https://rubygems.org"

  gem 'ffi', '<= 1.11.3'
G

bundle! :install
```

But I didn't try it.

>   * The missing functionality to be able to reintroduce #7215 would be to get the example you mentioned in [#7522 (comment)](https://github.com/bundler/bundler/pull/7522#issuecomment-571026880) passing. This is a different case from the ffi's realworld example and that work is yet to be done.

Yes.

I think that the work should be done by a separated pull request. #7215's use case is different from this. #7215 always uses all platforms. This improves resolution for "ruby" platform and other specific platform.

lib/bundler/resolver/spec_group.rb
deivid-rodriguez: You mean extracting the logic from `search_for` to `SpecGroup` with something like:

```ruby
def copy_for(platform = nil)
 sg_ruby = dup
 sg_ruby.activated_platforms = []

  if sg_ruby.for?(Gem::Platform::RUBY)
    sg_ruby.activate_platform!(Gem::Platform::RUBY)
    sg_ruby.activate_platform!(platform) if platform
    sg_ruby
  end
end
```

I like it, seems more DRY.

lib/bundler/resolver/spec_group.rb
deivid-rodriguez: Maybe `spec_for`?

Note that we're already using a `for` method in `SpecSet`, so maybe `for` is ok here too. It probably won't be confusing if we make sure to always call it with a explicit receiver?

lib/bundler/resolver/spec_group.rb
kou: Thanks. I like the `copy_for` name. I'll use it.

kou:
Thanks for your comments. We got better names.

Once CI is green, could you merge this? Then could you also review #7554?

larskanis:
I did another try with this PR and #7554 merged together. It fails with a real world example:
```
$ cat Gemfile
source "https://rubygems.org"
gem "ffi", "~> 1.11"
$ ruby -Ilib exe/bundle install
Resolving dependencies...
Bundler found conflicting requirements for the Ruby  version:
  In Gemfile:
    Ruby  x64-mingw32

    ffi (~> 1.11) x64-mingw32 was resolved to 1.11.3, which depends on
      Ruby  (>= 2.0)

$ ruby -v
ruby 2.7.0p0 (2019-12-25 revision 647ee6f091) [x64-mingw32]
```
If I omit `"~> 1.11"` from the Gemfile, it installs ffi-1.9.25, which was the last version without ruby version constraint.

I think the same error can be simulated by changing a spec slightly:
```diff
diff --git a/spec/resolver/platform_spec.rb b/spec/resolver/platform_spec.rb
index 9a97a2c67c..134f18425a 100644
--- a/spec/resolver/platform_spec.rb
+++ b/spec/resolver/platform_spec.rb
@@ -69,7 +69,9 @@ RSpec.describe "Resolving platform craziness" do
     @index = build_index do
       gem "foo", "1.0.0"
       gem "foo", "1.0.0", "x64-mingw32"
-      gem "foo", "1.1.0"
+      gem "foo", "1.1.0" do |s|
+        s.required_ruby_version = [">= 2.0"]
+      end
       gem "foo", "1.1.0", "x64-mingw32" do |s|
         s.required_ruby_version = [">= 2.0", "< 2.4"]
       end
```
It fails with the following error, although the change should be irrelevant:
```
       expected: ["foo-1.1.0"]
            got: ["foo-1.0.0-x64-mingw32"]
```

deivid-rodriguez:
Woow, thanks again @larskanis for the thorough testing. I tried to have a look at the new failing spec, and this is what I got:

* First thing I noticed when debugging is that there were duplicated spec groups for ruby and rubygems version requirements, one with lower case name and one with uppercase. Not sure where the difference comes from (I would've expected all uppercased?), but downcasing things removed the duplication. This changed nothing regarding the failing spec. Patch is:

```diff
diff --git a/lib/bundler/definition.rb b/lib/bundler/definition.rb
index 3e5186039a..876a1d8db8 100644
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -864,8 +864,8 @@ module Bundler
           concat_ruby_version_requirements(locked_ruby_version_object) unless @unlock[:ruby]
         end
         [
-          Dependency.new("Ruby\0", ruby_versions),
-          Dependency.new("RubyGems\0", Gem::VERSION),
+          Dependency.new("ruby\0", ruby_versions),
+          Dependency.new("rubygems\0", Gem::VERSION),
         ]
       end
     end
diff --git a/lib/bundler/resolver/spec_group.rb b/lib/bundler/resolver/spec_group.rb
index 4ebb747956..ef089ae0ed 100644
--- a/lib/bundler/resolver/spec_group.rb
+++ b/lib/bundler/resolver/spec_group.rb
@@ -116,10 +116,10 @@ module Bundler
         return [] if !spec.is_a?(EndpointSpecification) && !spec.is_a?(Gem::Specification)
         dependencies = []
         if !spec.required_ruby_version.nil? && !spec.required_ruby_version.none?
-          dependencies << DepProxy.new(Gem::Dependency.new("Ruby\0", spec.required_ruby_version), platform)
+          dependencies << DepProxy.new(Gem::Dependency.new("ruby\0", spec.required_ruby_version), platform)
         end
         if !spec.required_rubygems_version.nil? && !spec.required_rubygems_version.none?
-          dependencies << DepProxy.new(Gem::Dependency.new("RubyGems\0", spec.required_rubygems_version), platform)
+          dependencies << DepProxy.new(Gem::Dependency.new("rubygems\0", spec.required_rubygems_version), platform)
         end
         dependencies
       end
diff --git a/lib/bundler/source/metadata.rb b/lib/bundler/source/metadata.rb
index 0867879861..09171f10dd 100644
--- a/lib/bundler/source/metadata.rb
+++ b/lib/bundler/source/metadata.rb
@@ -5,8 +5,8 @@ module Bundler
     class Metadata < Source
       def specs
         @specs ||= Index.build do |idx|
-          idx << Gem::Specification.new("Ruby\0", RubyVersion.system.to_gem_version_with_patchlevel)
-          idx << Gem::Specification.new("RubyGems\0", Gem::VERSION) do |s|
+          idx << Gem::Specification.new("ruby\0", RubyVersion.system.to_gem_version_with_patchlevel)
+          idx << Gem::Specification.new("rubygems\0", Gem::VERSION) do |s|
             s.required_rubygems_version = Gem::Requirement.default
           end
 
diff --git a/spec/install/gems/resolving_spec.rb b/spec/install/gems/resolving_spec.rb
index 0e933bfd8e..4d1fb98263 100644
--- a/spec/install/gems/resolving_spec.rb
+++ b/spec/install/gems/resolving_spec.rb
@@ -174,17 +174,17 @@ RSpec.describe "bundle install with install-time dependencies" do
             gem 'require_ruby'
           G
 
-          expect(out).to_not include("Gem::InstallError: require_ruby requires Ruby version > 9000")
+          expect(out).to_not include("Gem::InstallError: require_ruby requires ruby version > 9000")
 
           nice_error = strip_whitespace(<<-E).strip
-            Bundler found conflicting requirements for the Ruby\0 version:
+            Bundler found conflicting requirements for the ruby\0 version:
               In Gemfile:
-                Ruby\0 (#{error_message_requirement})#{error_message_platform}
+                ruby\0 (#{error_message_requirement})#{error_message_platform}
 
                 require_ruby was resolved to 1.0, which depends on
-                  Ruby\0 (> 9000)#{error_message_platform}
+                  ruby\0 (> 9000)#{error_message_platform}
 
-            Ruby\0 (> 9000), which is required by gem 'require_ruby', is not available in the local ruby installation
+            ruby\0 (> 9000), which is required by gem 'require_ruby', is not available in the local ruby installation
           E
           expect(err).to end_with(nice_error)
         end
@@ -222,7 +222,7 @@ RSpec.describe "bundle install with install-time dependencies" do
         gem 'require_rubygems'
       G
 
-      expect(err).to_not include("Gem::InstallError: require_rubygems requires RubyGems version > 9000")
+      expect(err).to_not include("Gem::InstallError: require_rubygems requires rubygems version > 9000")
       expect(err).to include("require_rubygems-1.0 requires rubygems version > 9000, which is incompatible with the current version, #{Gem::VERSION}")
     end
   end
diff --git a/spec/resolver/basic_spec.rb b/spec/resolver/basic_spec.rb
index 57897f89b4..63fefec560 100644
--- a/spec/resolver/basic_spec.rb
+++ b/spec/resolver/basic_spec.rb
@@ -91,7 +91,7 @@ RSpec.describe "Resolving" do
     @index = build_index do
       gem "mail", "2.7.0"
       gem "mail", "2.7.1.rc1"
-      gem "RubyGems\0", Gem::VERSION
+      gem "rubygems\0", Gem::VERSION
     end
     dep "mail"
     @locked = locked ["mail", "2.7.0"]
@@ -169,10 +169,10 @@ Bundler could not find compatible versions for gem "a":
         s.required_ruby_version = "~> 2.0.0"
       end
 
-      gem "Ruby\0", "1.8.7"
+      gem "ruby\0", "1.8.7"
     end
     dep "foo"
-    dep "Ruby\0", "1.8.7"
+    dep "ruby\0", "1.8.7"
 
     deps = []
     @deps.each do |d|
diff --git a/spec/support/indexes.rb b/spec/support/indexes.rb
index dc6e0bd1e9..83d8fc4180 100644
--- a/spec/support/indexes.rb
+++ b/spec/support/indexes.rb
@@ -77,7 +77,7 @@ module Spec
         gem "rack-mount", %w[0.4 0.5 0.5.1 0.5.2 0.6]
 
         # --- Pre-release support
-        gem "RubyGems\0", ["1.3.2"]
+        gem "rubygems\0", ["1.3.2"]
 
         # --- Rails
         versions "1.2.3 2.2.3 2.3.5 3.0.0.beta 3.0.0.beta1" do |version|
@@ -414,7 +414,7 @@ module Spec
         gem("b", %w[0.9.0 1.5.0 2.0.0.pre])
 
         # --- Pre-release support
-        gem "RubyGems\0", ["1.3.2"]
+        gem "rubygems\0", ["1.3.2"]
       end
     end
   end
```

* Then I noticed that when adding the ruby requirement for the `foo-1.1.0 (ruby)` gem, it was no longer being consider because the only spec group about ruby version requirements was for the x64-mingw32 platform. So I tried adding the Ruby spec group unconditionally with this patch:

```diff
diff --git a/lib/bundler/resolver.rb b/lib/bundler/resolver.rb
index 8a95273a0f..7de6920be1 100644
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -150,13 +150,10 @@ module Bundler
       search.each do |sg|
         spec = sg.spec_for(platform)
         next if spec.nil?
-        spec_platform = spec.platform
-        if spec_platform && spec_platform != Gem::Platform::RUBY
-          # Add a spec group for "non platform specific spec" as the fallback
-          # spec group.
-          sg_ruby = sg.copy_for(Gem::Platform::RUBY)
-          selected_sgs << sg_ruby if sg_ruby
-        end
+        # Add a spec group for "non platform specific spec" as the fallback
+        # spec group.
+        sg_ruby = sg.copy_for(Gem::Platform::RUBY)
+        selected_sgs << sg_ruby if sg_ruby
         if platform != Gem::Platform::RUBY
           # Add a spec group for ["non platform specific spec", "platform
           # specific spec"] to resolve a spec for multiple platforms.
```

That fixed the new failing spec, with the following issues:

* It needs the first patch about "casing" to not break other resolver specs, not sure why.
* It breaks the following spec:

```
 $ bin/rspec spec/resolver/platform_spec.rb 

Randomized with seed 58895
.............F

Retried examples: 0


Failures:

  1) Resolving platform craziness with conflicting cases reports on the conflict
     Failure/Error: raise "The resolve succeeded with: #{got.map(&:full_name).sort.inspect}"
     
     RuntimeError:
       The resolve succeeded with: ["bar-1.0.0", "baz-1.0.0", "foo-1.0.0", "foo-1.0.0"]
     # ./spec/support/indexes.rb:48:in `should_conflict_on'
     # ./spec/resolver/platform_spec.rb:170:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:112:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:112:in `block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:77:in `block (2 levels) in <top (required)>'
     # ./spec/support/rubygems_ext.rb:98:in `load'
     # ./spec/support/rubygems_ext.rb:98:in `gem_load_and_activate'
     # ./spec/support/rubygems_ext.rb:45:in `gem_load'

Finished in 0.25476 seconds (files took 0.13311 seconds to load)
14 examples, 1 failure

Failed examples:

rspec ./spec/resolver/platform_spec.rb:166 # Resolving platform craziness with conflicting cases reports on the conflict

Randomized with seed 58895
```

However, it seems to me that the new result is the correct one for this spec... :thinking:

kou:
Ah, sorry. I didn't confirm installed ffi gem version.

> Not sure where the difference comes from (I would've expected all uppercased?)

Are "Ruby" and "RubyGems" instead of "ruby" and "rubygems" expected? Then it will be fixed by improving `Bundler::CompactIndexClient::Cache`. https://rubygems.org/info/ffi returns `ruby:...,rubygems:...` and `Bundler::CompactIndexClient::Cache` use the "ruby" and "rubygems" as is.

Anyway, we could confirm that #7554 can solve Ruby version problem. Could you merge #7554 first? Then I can send a new pull request to change "ruby" and "rubygems" to "Ruby" and "RubyGems". Then I will rebase this branch and fix this case.

kou:
> However, it seems to me that the new result is the correct one for this spec... :thinking:

I think so too.

With this improvement, the spec should choose "bar-1.0.0" because "bar-1.0.0-java" isn't usable with both "ruby" and "java". If "bar-1.0.0-java" is used, "baz" is conflicted with "ruby" and "java".

deivid-rodriguez:
> Are "Ruby" and "RubyGems" instead of "ruby" and "rubygems" expected? Then it will be fixed by improving Bundler::CompactIndexClient::Cache. https://rubygems.org/info/ffi returns ruby:...,rubygems:... and Bundler::CompactIndexClient::Cache use the "ruby" and "rubygems" as is.

> Anyway, we could confirm that #7554 can solve Ruby version problem. Could you merge #7554 first? Then I can send a new pull request to change "ruby" and "rubygems" to "Ruby" and "RubyGems". Then I will rebase this branch and fix this case.

I don't know if "Ruby" and "RubyGems" or "ruby" and "rubygems" are expected. I think the only user visible difference is the case of the error messages that get printed. In that case, I think I actually prefer lowercase, but since that would be a user visible change (even if minimal) that requires changing expectations, I agree it's better to fix it for now `Bundler::CompactIndexClient::Cache` to use CamelCase consistently :+1: 

Your plan sounds good to me, I'll switch over to reviewing and merging #7554.

kou:
> I don't know if "Ruby" and "RubyGems" or "ruby" and "rubygems" are expected.

Ah, OK. I misunderstood.

> I think the only user visible difference is the case of the error messages that get printed. In that case, I think I actually prefer lowercase, but since that would be a user visible change (even if minimal) that requires changing expectations, I agree it's better to fix it for now `Bundler::CompactIndexClient::Cache` to use CamelCase consistently +1

I understand. It may be better that we change "Ruby" and "RubyGems" to "ruby" and "rubygemns" when we release Bundler 3.0.0.

deivid-rodriguez:
I can see how you misunderstood me, since first I said 

> I would've expected all uppercased?

and then

> I don't know if "Ruby" and "RubyGems" or "ruby" and "rubygems" are expected

:rofl:

First sentence I meant that from reading the code i didn't figure out the reason for the discrepancy and thought every spec group would be in CamelCase. In the second, I meant that I'm not sure what the preferred behavior is from an end user point of view.

Anyways, #7554 landed, so now I should be expecting a new PR to normalize spec groups about metadata to be named `RubyGems/Ruby` consistently (instead of `rubygems/ruby`), right?

> It may be better that we change "Ruby" and "RubyGems" to "ruby" and "rubygemns" when we release Bundler 3.0.0.

Maybe we don't need a major version bump for that tiny UI change, but we either way let's keep the same behaviour for now and we can change that little thing later :+1:.

Thanks again @kou!!

kou:
> Anyways, #7554 landed, so now I should be expecting a new PR to normalize spec groups about metadata to be named `RubyGems/Ruby` consistently (instead of `rubygems/ruby`), right?

Right.
I've created a pull request for it: https://github.com/rubygems/bundler/pull/7570

I'll also improve dependency resolution in this branch.

deivid-rodriguez:
Ok, so after #7570 being closed, what's the status here? Should we normalize things to always be "ruby/rubygems" instead of "Ruby/RubyGems" in a separate PR using the changes I mentioned in https://github.com/rubygems/bundler/pull/7522#issuecomment-572635007?

kou:
I'll explain more later but I just explain "ruby"/"rubygems" and "Ruby"/"RubyGems" for now.

We should use "Ruby\0" and "RubyGems\0" instead of "ruby\0" and "rubygems\0" because we use the former so far. We uses "ruby\0" (and "rubygems\0") only in the spec mentioned in https://github.com/rubygems/bundler/pull/7522#issuecomment-572441286 . So the spec is wrong. We should fix the spec ("ruby\0" -> "Ruby\0") instead of changing other locations: https://github.com/rubygems/bundler/pull/7522/commits/637b695640b147061a04b5c11ed3de4ec7a656c0

spec/resolver/platform_spec.rb
```diff
@@ -90,11 +165,11 @@
       end
     end
 
-    it "reports on the conflict" do
+    it "takes the ruby version as fallback" do
       platforms "ruby", "java"
       dep "foo"
 
-      should_conflict_on "baz"
+      should_resolve_as %w[bar-1.0.0 baz-1.0.0 foo-1.0.0]
```
kou: This should be resolved instead of conflicting.

See also:

* https://github.com/rubygems/bundler/pull/7522#issuecomment-572635007
* https://github.com/rubygems/bundler/pull/7522#issuecomment-572882729

spec/resolver/platform_spec.rb
```diff
@@ -28,6 +28,81 @@
...
+        s.required_ruby_version = [">= 2.0", "< 2.4"]
+      end
+      gem "Ruby\0", "2.5.1"
+    end
+    dep "foo"
+    dep "Ruby\0", "2.5.1"
+    platforms "x86_64-linux", "x64-mingw32"
+
+    should_resolve_as %w[foo-1.1.0]
+  end
```
kou: This is a spec mentioned at https://github.com/rubygems/bundler/pull/7522 .
Now, we can reintroduce #7215.

spec/resolver/platform_spec.rb
```diff
@@ -28,6 +28,81 @@
...
+      gem "foo", "1.1.0" do |s|
+        s.required_ruby_version = [">= 2.0"]
+      end
+      gem "foo", "1.1.0", "x64-mingw32" do |s|
+        s.required_ruby_version = [">= 2.0", "< 2.4"]
+      end
+      gem "Ruby\0", "2.5.1"
+    end
+    dep "foo"
+    dep "Ruby\0", "2.5.1"
```
kou: This must be `"Ruby\0"` instead of `"ruby\0"`.

lib/bundler/resolver/spec_group.rb
```diff
@@ -28,7 +29,7 @@ def to_specs
           lazy_spec = LazySpecification.new(name, version, s.platform, source)
           lazy_spec.dependencies.replace s.dependencies
           lazy_spec
-        end.compact
+        end.compact.uniq
```
kou: Don't return duplicated specs.
We can make `Bundler::Resolver` cleaner to remove duplicated specs here.

This needs a change in `LazySpecification`.

lib/bundler/lazy_specification.rb
```diff
@@ -46,6 +46,14 @@ def ==(other)
...
     end
 
+    def eql?(other)
+      identifier.eql?(other.identifier)
+    end
+
+    def hash
+      identifier.hash
+    end
+
```
kou: They are for supporting `uniq`. `uniq` uses `eql?`.

lib/bundler/resolver.rb
```diff
@@ -146,7 +146,25 @@ def search_for(dependency)
...
+        selected_sgs << sg_ruby if sg_ruby
+        sg_all_platforms = nil
+        self.class.sort_platforms(@platforms).reverse_each do |other_platform|
+          if sg_all_platforms.nil?
+            sg_all_platforms = sg.copy_for(other_platform)
+          else
+            sg_all_platforms.activate_platform!(other_platform)
+          end
+        end
+        selected_sgs << sg_all_platforms if sg_all_platforms
```
kou: `if sg_all_platforms` may not be needed.

spec/resolver/platform_spec.rb
deivid-rodriguez: This is the spec as modified by @larskanis in https://github.com/rubygems/bundler/pull/7522#issuecomment-572441286. Do you think it makes sense to keep both the old version (only `required_ruby_version` constraint in the platform specific `foo-1.1.0` gem) and the new one as two different specs?

spec/resolver/platform_spec.rb
deivid-rodriguez: Nice, and I remember that during my investigation I was getting a correct resolution, but with one spec being duplicated, and I see that you fixed that with your `uniq` addition :tada:.

spec/resolver/platform_spec.rb
deivid-rodriguez: That is really awesome :1st_place_medal:.

spec/resolver/platform_spec.rb
kou: Yes. I'll add the old version again.

kou: Done: https://github.com/rubygems/bundler/pull/7522/commits/7466d82cae0c302d919a81304088bb399edf2409

lib/bundler/resolver.rb
deivid-rodriguez: I don't think it's needed either. At least I think `@platforms` should never be empty here (which would cause this to crash), because in that case we would abort earlier (I think [here](https://github.com/rubygems/bundler/blob/4cc9dc992b2fa38c3cd157d5d1eb2396bac57a47/lib/bundler/definition.rb#L499-L509)?). Not source about the `copy_for` inside the loop always returning `nil` though.

I think we could maybe add a note that we could try to remove that clause in the future, and do it in the future once all these changes stabilize and we feel everything regarding multiplatform is solid.

larskanis:
Unfortunately it still doesn't work. FFI now installs fine, but only due to the fact, that I pushed ffi-1.12.1 yesterday with support for ruby-2.7. The situation remains with nokogiri:
```
c:\Users\kanis\bundler>ruby -Ilib exe/bundle install
Fetching gem metadata from https://rubygems.org/..................
Resolving dependencies...
nokogiri-1.10.7-x64-mingw32 requires ruby version >= 2.3, < 2.7.dev, which is
incompatible with the current version, ruby 2.7.0p0
c:\Users\kanis\bundler>cat Gemfile
source "https://rubygems.org"
gem "fxruby"
```
This is with commit 7466d82cae0c302d919a81304088bb399edf2409 and I checked that the code from this PR is actually called.

deivid-rodriguez:
Thanks for keeping on testing @larskanis, we'll get there! :muscle:

lib/bundler/resolver.rb
kou: OK. I've removed the `if` by adding the `platform` to target platforms. The `platform` must be valid because we check it by `sg.for?(platform)`. It means that at least one `sg.copy_for(other_platform)` succeeds. So we can remove the `if` safely.

kou:
@larskanis Thanks. I could reproduce it on Windows.

This is a problem of `LazySpecification#__materialize__`. It doesn't care about resolved platform. `Resolver` works well.

https://github.com/rubygems/bundler/pull/7522/commits/6db75e052e69a2aafcbb7bc98d99c8e8b4f65939 solves this case. Could you try again this branch? This branch has the commit.

@deivid-rodriguez Should we include https://github.com/rubygems/bundler/pull/7522/commits/6db75e052e69a2aafcbb7bc98d99c8e8b4f65939 in this pull request? It may be out of scope of this pull request. I intended that this pull request improves only resolver. But https://github.com/rubygems/bundler/pull/7522/commits/6db75e052e69a2aafcbb7bc98d99c8e8b4f65939 is a process after resolution.

If we split `LazySpeficition` improvement into other pull request, I'll remove https://github.com/rubygems/bundler/pull/7522/commits/6db75e052e69a2aafcbb7bc98d99c8e8b4f65939 from this pull request, @deivid-rodriguez merges this pull request and I'll create a pull request for https://github.com/rubygems/bundler/pull/7522/commits/6db75e052e69a2aafcbb7bc98d99c8e8b4f65939 .

lib/bundler/resolver.rb
deivid-rodriguez: Nice, that makes sense!

deivid-rodriguez:
I don't mind merging here or separately. It would be great though if we could add a spec to cover the last fix. It could be a realworld spec with response recorded through VCR if we can't find something more specific. I can add realworld specs to be run under windows so that this can be caught properly in CI.

kou:
OK.
I want to finish this branch as soon as possible because this pull request has a too long thread and many commits. So I want the last fix as a follow-up task.

deivid-rodriguez:
Ok, so you remove the commit and I merge, or I merge straight away and we add specs separately?

kou:
I'll remove the last fix.

larskanis:
@kou Now it works! It really works! I played with some scenarios on Windows with ruby-2.6 and 2.7 and with fxruby, nokogiri and pg and with `force_ruby_platform`. Everything worked as expected. I'm sure - bundler-2.2 will be the best version ever!

kou:
Done.

kou:
@larskanis Thanks for confirming this! :-)

kou:
I'll create a pull request with the last fix and its test after this pull request is merged.

deivid-rodriguez:
It's going in now!

@bundlerbot merge

deivid-rodriguez:
@kou PR is in! You can proceed with the followup task :heart:.

kou:
Done: #7578

PR #7523, opened 2019-12-30
Fix rspec stuck problem for large stderr external command

kou:

This PR just fixes a developer problem.

Some specs runs external commands by `sys_exec` in `spec/support/helpers.rb`. They may be stuck when they outputs many text to its stderr.

`sys_exec` uses `open3`. It uses pipe to communicate an external command. If `rspec` process doesn't read stderr of the external command, the external command is stuck when the external command writes many text to its stderr. Because the external command can't write to pipe infinitely.

The current `sys_exec` tries to fix this situation but it's incomplete:

```ruby
        command_execution.stdout = Thread.new { stdout.read }.value.strip
        command_execution.stderr = Thread.new { stderr.read }.value.strip
```

`Thread.new { stderr.read ` isn't started until `Thread.new { stdout.read }.value` is finished. Normally, it's finished when the external command is finished. It's late. We should read stderr of the external command while the external command is alive.

I don't have another option.

### DIFF
--- a/spec/support/helpers.rb
+++ b/spec/support/helpers.rb
@@ -210,8 +210,10 @@ module Spec
         yield stdin, stdout, wait_thr if block_given?
         stdin.close
 
-        command_execution.stdout = Thread.new { stdout.read }.value.strip
-        command_execution.stderr = Thread.new { stderr.read }.value.strip
+        stdout_read_thread = Thread.new { stdout.read }
+        stderr_read_thread = Thread.new { stderr.read }
+        command_execution.stdout = stdout_read_thread.value.strip
+        command_execution.stderr = stderr_read_thread.value.strip
         command_execution.exitstatus = wait_thr && wait_thr.value.exitstatus
       end
 


kou:
I think that CI failure is unrelated:

https://travis-ci.org/bundler/bundler/jobs/631000500#L536-L541

```text
       /home/travis/build/bundler/bundler/tmp/rubygems/lib/rubygems/core_ext/kernel_require.rb:92:in `require': cannot load such file -- ruby2_keywords (LoadError)
       	from /home/travis/build/bundler/bundler/tmp/rubygems/lib/rubygems/core_ext/kernel_require.rb:92:in `require'
       	from /home/travis/build/bundler/bundler/tmp/1/gems/base/gems/mustermann-1.1.0/lib/mustermann.rb:6:in `<top (required)>'
       	from /home/travis/build/bundler/bundler/tmp/rubygems/lib/rubygems/core_ext/kernel_require.rb:92:in `require'
       	from /home/travis/build/bundler/bundler/tmp/rubygems/lib/rubygems/core_ext/kernel_require.rb:92:in `require'
       	from /home/travis/build/bundler/bundler/tmp/1/gems/base/gems/sinatra-2.0.7/lib/sinatra/base.rb:8:in `<top (required)>'
```

kou:

deivid-rodriguez:
@bundlerbot merge

PR #7525, opened 2019-12-30
Use raise instead of flunk

kou:

This PR fixes a developer problem.

RSpec doesn't provide `flunk` method. It's a method name in test-unit and minitest.

3edfddbf7ebe4b942e1e2b6d693afd582e9a8147 may use `flunk` accidentally.

My fix uses `raise` instead of `flunk`.

I don't have another option.


kou:

deivid-rodriguez:
@bundlerbot merge

PR #7526, opened 2019-12-30
Fix CI failure with mustermann 1.1.0

kou:

This PR fixes a developer problem. Our CI is failed with mustermann 1.1.0.

This is caused because mustermann 1.1.0 starts depending on ruby2_keyword gem but our test script doesn't add a load path for ruby2_keyword gem yet.

See also:

  * The PR that adds ruby2_keyword gem dependency to mustermann: https://github.com/sinatra/mustermann/pull/97
  * Error message on CI: https://github.com/bundler/bundler/pull/7523#issuecomment-569736683

My fix adds a load path for ruby2_keywrod gem installed in `tmp/1/gems/`.

I don't have another option.

### DIFF
--- a/spec/support/artifice/endpoint.rb
+++ b/spec/support/artifice/endpoint.rb
@@ -4,7 +4,7 @@ require_relative "../path"
 require Spec::Path.lib_dir.join("bundler/deprecate")
 include Spec::Path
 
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
@bundlerbot merge

PR #7527, opened 2019-12-31
Report original exception when failing to load openssl

### LINKED ISSUE #7192: SSLError does not report the actual problem
I obtained the following output from bundler in our CI environment:

<pre>
[2019/06/07 12:49:12.626] + bundle install
[2019/06/07 12:49:12.626] [mongo] Warning: No private key present, creating unsigned gem.
[2019/06/07 12:49:12.647] Could not load OpenSSL.
[2019/06/07 12:49:12.647] You must recompile Ruby with OpenSSL support or change the sources in your
[2019/06/07 12:49:12.647] Gemfile from 'https' to 'http'. Instructions for compiling with OpenSSL using
[2019/06/07 12:49:12.647] RVM are available at http://rvm.io/packages/openssl.
[2019/06/07 12:49:12.648] Command failed: command encountered problem: exit status 1
</pre>

As far as I can tell, the Ruby installation I am using does have openssl extension compiled and loadable. Therefore I am stumped as to what the cause of this bundler failure is.

Looking at the code in fetcher.rb, when this exception is raised, bundler does not actually attempt to load openssl. It is not immediately obvious to me how to make bundler report the actual error.

p-mongo:

See #7192.

Bundler discards `LoadError` class and message when reporting it.

Report class and message of the original exception when reporting `LoadError`.

Fixes #7192

### DIFF
--- a/lib/bundler/friendly_errors.rb
+++ b/lib/bundler/friendly_errors.rb
@@ -23,7 +23,7 @@ module Bundler
         Bundler.ui.error error.message
       when LoadError
         raise error unless error.message =~ /cannot load such file -- openssl|openssl.so|libcrypto.so/
-        Bundler.ui.error "\nCould not load OpenSSL. #{error.class}: #{error}"
+        Bundler.ui.error "\nCould not load OpenSSL. #{error.class}: #{error}\n#{error.backtrace.join("\n  ")}"
         Bundler.ui.warn <<-WARN, :wrap => true
           You must recompile Ruby with OpenSSL support or change the sources in your \
           Gemfile from 'https' to 'http'. Instructions for compiling with OpenSSL \
--- a/lib/bundler/friendly_errors.rb
+++ b/lib/bundler/friendly_errors.rb
@@ -29,7 +29,6 @@ module Bundler
           Gemfile from 'https' to 'http'. Instructions for compiling with OpenSSL \
           using RVM are available at https://rvm.io/packages/openssl.
         WARN
-        Bundler.ui.trace error
       when Interrupt
         Bundler.ui.error "\nQuitting..."
         Bundler.ui.trace error
--- a/lib/bundler/friendly_errors.rb
+++ b/lib/bundler/friendly_errors.rb
@@ -23,7 +23,7 @@ module Bundler
         Bundler.ui.error error.message
       when LoadError
         raise error unless error.message =~ /cannot load such file -- openssl|openssl.so|libcrypto.so/
-        Bundler.ui.error "\nCould not load OpenSSL: #{error.class}: #{error}"
+        Bundler.ui.error "\nCould not load OpenSSL. #{error.class}: #{error}"
         Bundler.ui.warn <<-WARN, :wrap => true
           You must recompile Ruby with OpenSSL support or change the sources in your \
           Gemfile from 'https' to 'http'. Instructions for compiling with OpenSSL \
--- a/lib/bundler/friendly_errors.rb
+++ b/lib/bundler/friendly_errors.rb
@@ -23,7 +23,7 @@ module Bundler
         Bundler.ui.error error.message
       when LoadError
         raise error unless error.message =~ /cannot load such file -- openssl|openssl.so|libcrypto.so/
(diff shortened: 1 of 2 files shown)


lib/bundler/friendly_errors.rb
```diff
@@ -23,7 +23,7 @@ def log_error(error)
         Bundler.ui.error error.message
       when LoadError
         raise error unless error.message =~ /cannot load such file -- openssl|openssl.so|libcrypto.so/
-        Bundler.ui.error "\nCould not load OpenSSL."
+        Bundler.ui.error "\nCould not load OpenSSL. #{error.class}: #{error}"
```
deivid-rodriguez: We're showing the error class, message, and backtrace in `--verbose` mode as well, so I figure we can show everything by default and remove the `Bundler.ui.trace error` line below instead?

p-mongo: I do not recall ever seeing additional information in verbose mode for this exception therefore I have no objections to removing the verbose functionality.

deivid-rodriguez: Yeah, I read [your comment in the original issue](https://github.com/bundler/bundler/issues/7192#issuecomment-500029911), but I tried it myself in it does print all the information. Maybe it was a bug in an older version.

We should either recommend rerunning with the `--verbose` flag, or print everything. I tend to prefer the latter since as you point out, without additional information this particular error message is not useful.

deivid-rodriguez: @p-mongo Are you planning to keep working on this PR, or that I wrap it up myself? Thanks!

p-mongo: I have it in my queue but I do not have the bandwidth to tackle it right this moment. Whatever you think is best is fine with me.

deivid-rodriguez: Ok, kind of the same thing for me :). I guess whoever gets there first can wrap this up. I'll let you know before start working on this so we don't duplicate efforts.

egiurleo: @deivid-rodriguez I'm happy to finish this up. I noticed that the `trace` method also displays the error backtrace. Would you like the backtrace to always be printed as well?

deivid-rodriguez: Yeah, I think in this particular case I would go with "full verbose mode" by default.

egiurleo: Done.

lib/bundler/friendly_errors.rb
```diff
@@ -23,13 +23,12 @@ def log_error(error)
         Bundler.ui.error error.message
       when LoadError
         raise error unless error.message =~ /cannot load such file -- openssl|openssl.so|libcrypto.so/
-        Bundler.ui.error "\nCould not load OpenSSL."
+        Bundler.ui.error "\nCould not load OpenSSL. #{error.class}: #{error}\n#{error.backtrace.join("\n  ")}"
         Bundler.ui.warn <<-WARN, :wrap => true
           You must recompile Ruby with OpenSSL support or change the sources in your \
           Gemfile from 'https' to 'http'. Instructions for compiling with OpenSSL \
           using RVM are available at https://rvm.io/packages/openssl.
         WARN
```
deivid-rodriguez: Since we are here, do you think this message adds much value? I think I would remove this message entirely or maybe change the message to:

> Could not load OpenSSL, you ruby might not be compiled with OpenSSL support. The underlying error was #{error.class}: #{error}\n#{error.backtrace.join("\n  ")}"

The reason is:

* I don't think we want to explicitly recommend switching source the http.
* The RVM advice seems to specific to me, and it redirects to a page that's not very clear about how it helps to fix this issue.

What do you think? If you think this is too unrelated to this PR, I can tackle this myself separately.

egiurleo: I totally agree that this message isn't helpful. I'm happy to get you a new PR updating this message tomorrow morning if you want to merge this one now. :)

deivid-rodriguez: Sure!

deivid-rodriguez:
@bundlerbot merge

PR #7529, opened 2019-12-31
Remove optional require net-http-pipeline from vendored net-http-persistent.

### LINKED ISSUE #7521: Bundler does not vendor net-http-pipeline gem.
If we already have the net-http-pipeline gem installed and try to vendor it the gem is not vendored because Bundler loads it before vendoring (https://github.com/bundler/bundler/blob/master/lib/bundler/vendor/net-http-persistent/lib/net/http/persistent.rb#L7).

To reproduce this:

1- Gemfile
    ```
    source "https://rubygems.org"
    gem 'net-http-pipeline'
    ```
1- `bundle install`
1- `bundle install --path vendor/bundle``
1- `bundle check`

Bundle check gives this:
```
The following gems are missing
 * net-http-pipeline (1.0.1)
Install missing gems with `bundle install`
```

marciotoshio:

See issue https://github.com/bundler/bundler/issues/7521.

My diagnosis was If you have net-http-pipeline gem installed bundler loads it and you can't vendor it.

My fix is remove the optional require of net-http-pipeline.

I chose this fix because was a recommendation in the issue by @deivid-rodriguez 

Fixes #7521.

### DIFF
--- a/lib/bundler/vendor/net-http-persistent/lib/net/http/persistent.rb
+++ b/lib/bundler/vendor/net-http-persistent/lib/net/http/persistent.rb
@@ -3,11 +3,6 @@ require_relative '../../../../uri/lib/uri'
 require 'cgi' # for escaping
 require_relative '../../../../connection_pool/lib/connection_pool'
 
-begin
-  require 'net/http/pipeline'
-rescue LoadError
-end
-
 autoload :OpenSSL, 'openssl'
 
 ##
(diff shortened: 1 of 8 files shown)


deivid-rodriguez:
Thanks for fixing this @marciotoshio! I added one spec to verify the fix and I'll release this in a few days.

deivid-rodriguez:
@bundlerbot merge

PR #7532, opened 2020-01-01
Release 2.1.3

deivid-rodriguez:

The problem was we should release more bug fixes.

My fix is to prepare 2.1.3.


deivid-rodriguez:
@bundlerbot merge

PR #7533, opened 2020-01-01
Rebuild man pages for January

deivid-rodriguez:

The problem was man pages are out of date.

My fix is to bring them up to date.


deivid-rodriguez:
@bundlerbot merge

PR #7534, opened 2020-01-01
Set master's version to 2.2.0.dev

deivid-rodriguez:

The problem was that master version is set to the latest patch level release, but the code in master doesn't really match the latest patch level release.

My fix is to use 2.2.0.dev to name master's development version.

### DIFF
--- a/lib/bundler/version.rb
+++ b/lib/bundler/version.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: false
 
 module Bundler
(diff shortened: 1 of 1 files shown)


simi:
Any way to restart tests now?

deivid-rodriguez:
@bundlerbot retry

deivid-rodriguez:
@bundlerbot try

deivid-rodriguez:
@bundlerbot merge

PR #7539, opened 2020-01-03
Migrate Windows CI to Github Actions

deivid-rodriguez:

The problem was that current Windows CI is very slow.

My fix is to try migrate it to Github Actions to see if it gets better. It didn't really get much better, but I think Github Actions is better than Azure Pipelines.

Regarding the implementation, Github Actions environment doesn't seem to include an `rb-readline` version, and specs using `readline` just hang. My "solution" is to skip those specs for the time being.


simi:
I think it should be possible to install `rb-readline` in setup phase, anyway it will be probably painful to maintain which version to use.

deivid-rodriguez:
I tried but then we need to monkeypatch that and it gets messy. I think it's fine to skip readline specs for now.

Next step would be to skip every test that it's currently failing on Windows so that we have a real CI that fails when we break something on Windows.

simi:
@deivid-rodriguez sounds reasonable. Feel free to ping me if any help is needed. I have Windows machine around.

deivid-rodriguez:
I was working on implementing the second step, namely, actually make the build fail if something on Windows gets broken. But the PR had gotten too big, so I'll merge the migration to Github Actions first, and then I'll open a separate PR for the second task.

deivid-rodriguez:
@bundlerbot merge

benoittgt:
Curious about your result here.

deivid-rodriguez:
Regarding speed, no improvement, takes approximately the same time. But we planned to migrate to Github Actions anyways, so :man_shrugging:.

hsbt:
👍

Issue #6875, opened 2019-01-04
[Windows test failure] rspec './spec/bundler/bundler_spec.rb[1:2:2:1:1]' # Bundler#which when the executable in inside a quoted path behaves like it returns the correct executable returns the expected file


janpio:
As described in https://github.com/bundler/bundler/issues/6868#issuecomment-451418573 currently 665 out of 2509 tests of bundler are failing on my Windows environment. I am working on fixing this.

---

This is the first test that is failing:

```
rspec './spec/bundler/bundler_spec.rb[1:2:2:1:1]' # Bundler#which when the executable in inside a quoted path behaves like it returns the correct executable returns the expected file
```

Here is just that spec file:

```
C:\Projects\bundler (windows -> origin)
λ ruby bin/rspec ./spec/bundler/bundler_spec.rb:146
Run options:
  include {:locations=>{"./spec/bundler/bundler_spec.rb"=>[146]}}
  exclude {:ruby_repo=>false, :bundler=>"=< 2.0", :rubygems_master=>true, :git=>"=< 2.20.1", :rubygems=>"=< 2.6.14.1", :ruby=>"=< 2.4.4", :realworld=>true, :sudo=>true}

Bundler
  #which
    behaves like it returns the correct executable
      returns the expected file
    when the executable in inside a quoted path
      behaves like it returns the correct executable
        returns the expected file (FAILED - 1)
    when the executable is not found
      behaves like it returns the correct executable
        returns the expected file


Retried examples: 0

Failures:

  1) Bundler#which when the executable in inside a quoted path behaves like it returns the correct executable returns the expected file
     Failure/Error: expect(subject).to eq(expected)

       expected: "/e/executable"
            got: nil

       (compared using ==)
     Shared Example Group: "it returns the correct executable" called from ./spec/bundler/bundler_spec.rb:154
     # ./spec/bundler/bundler_spec.rb:146:in `block (4 levels) in <top (required)>'
     # ./spec/spec_helper.rb:122:in `block (2 levels) in <top (required)>'

Finished in 9.89 seconds (files took 1.6 seconds to load)
3 examples, 1 failure

Failed examples:

rspec './spec/bundler/bundler_spec.rb[1:2:2:1:1]' # Bundler#which when the executable in inside a quoted path behaves like it returns the correct executable returns the expected file
```

[...]

PR #7541, opened 2020-01-04
Adapt `Bundler#which` test to Windows

deivid-rodriguez:

The problem was that this spec was failing under Windows.

My diagnosis was that the test needs Windows specific paths.

My fix is to separate the test per OS.

I chose this fix because it was discussed and considered the better fix at https://github.com/bundler/bundler/issues/6875.

Closes https://github.com/bundler/bundler/issues/6875.


deivid-rodriguez:
@bundlerbot merge

Issue #6887, opened 2019-01-08
[Windows test failure] rspec ./spec/bundler/dsl_spec.rb:210 # Bundler::Dsl#gemspec with a ruby platform keeps track of the ruby platforms in the dependency


janpio:
As described in https://github.com/bundler/bundler/issues/6868#issuecomment-451418573 currently 665 out of 2509 tests of bundler are failing on my Windows environment. I am working on fixing this.

---

Next one:

```
C:\Projects\bundler (windows -> origin)                                                                                                                             
λ ruby bin/rspec ./spec/bundler/dsl_spec.rb:210                                                                                                                     
Run options:                                                                                                                                                        
  include {:locations=>{"./spec/bundler/dsl_spec.rb"=>[210]}}                                                                                                       
  exclude {:ruby_repo=>false, :bundler=>"=< 2.0", :rubygems_master=>true, :git=>"=< 2.20.1", :rubygems=>"=< 2.6.14.1", :ruby=>"=< 2.4.4", :realworld=>true, :sudo=>t
                                                                                                                                                                    
Bundler::Dsl                                                                                                                                                        
  #gemspec                                                                                                                                                          
    with a ruby platform                                                                                                                                            
      keeps track of the ruby platforms in the dependency (FAILED - 1)                                                                                              
                                                                                                                                                                    
                                                                                                                                                                    
Retried examples: 0                                                                                                                                                 
                                                                                                                                                                    
Failures:                                                                                                                                                           
                                                                                                                                                                    
  1) Bundler::Dsl#gemspec with a ruby platform keeps track of the ruby platforms in the dependency                                                                  
     Failure/Error: expect(subject.dependencies.last.platforms).to eq(Bundler::Dependency::REVERSE_PLATFORM_MAP[Gem::Platform::RUBY])                               
                                                                                                                                                                    
       expected: [:ruby, :ruby_18, :ruby_19, :ruby_20, :ruby_21, :ruby_22, :ruby_23, :ruby_24, :ruby_25, :mri, :mri_18, :mri_19, :mri_20, :mri_21, :mri_22, :mri_23,
, :rbx, :truffleruby]                                                                                                                                               
            got: [:x64_mingw, :x64_mingw_20, :x64_mingw_21, :x64_mingw_22, :x64_mingw_23, :x64_mingw_24, :x64_mingw_25]                                             
                                                                                                                                                                    
       (compared using ==)                                                                                                                                          
                                                                                                                                                                    
[...]

PR #7542, opened 2020-01-04
Make test platform independent

deivid-rodriguez:

The problem was that a test for the `gemspec` DSL is failing on Windows.

My diagnosis was that the test uses local platforms (that change between OS) but asserts for ruby platforms.

My fix is to add a proper stub.

I chose this fix because it was suggested in https://github.com/bundler/bundler/issues/6887 by @janpio and looks good.

Closes #6687.


deivid-rodriguez:
@bundlerbot merge

PR #7545, opened 2020-01-05
Enforce Windows CI

deivid-rodriguez:

The problem was that, although we have a Windows CI, errors are ignored, so whenever we break something on Windows, we don't notice, and having a CI become useless.

My diagnosis was that we should make the Windows CI required.

My fix is to actually go ahead and fix some issues, and skip the rest.


deivid-rodriguez:
Going in!

@bundlerbot merge

PR #7548, opened 2020-01-05
Fix escaping of some subprocesses during specs

deivid-rodriguez:

The problem was that some specs are failing on Windows due to commands like the following failing:

```
$ C:/hostedtoolcache/windows/Ruby/2.6.3/x64/bin/ruby.exe -Id:/a/bundler/bundler/lib -w -e  \
  <<EOS
            begin
               $:.unshift File.expand_path("bundle")
               require "bundler/setup"
     
               require "actionpack"
               puts ACTIONPACK
               require "spec"
          
            rescue LoadError => e
               $stderr.puts "ZOMG LOAD ERROR"# if e.message.include?("-- spec")
            end
   EOS
   -e:2: syntax error, unexpected backslash
                       \$:.unshift File.expand_path("...
                       ^
   -e:10: syntax error, unexpected backslash
             \$stderr.puts "ZOMG LOAD ERROR...
             ^
# $? => 1
```

My diagnosis was that if the code being run via `ruby -e` in a subprocess include "$", it was not being properly escaped under Windows.

My fix is to use `shellsplit` and `shellescape`.

I chose this fix because after a bit of testing it seemed to do the trick.

### DIFF
--- a/spec/runtime/require_spec.rb
+++ b/spec/runtime/require_spec.rb
@@ -155,7 +155,7 @@ RSpec.describe "Bundler.require" do
       begin
         Bundler.require
       rescue LoadError => e
-        $stderr.puts "ZOMG LOAD ERROR: \#{e.message}"
+        warn "ZOMG LOAD ERROR: \#{e.message}"
       end
     RUBY
     run(cmd)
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
Down to 279 failures! :muscle:

@bundlerbot merge

PR #7550, opened 2020-01-06
Enable parallel specs again

deivid-rodriguez:

The problem was that specs are slow.

My fix is to run them in parallel.

We enabled this a while ago, but run into some issues. I want to try again.


deivid-rodriguez:
It looks like this is working, let's try this!

@bundlerbot merge

PR #7554, opened 2020-01-08
Fix ruby version parsing in compact index

kou:

The end-user can't use multiple Ruby version requirements (e.g. ">= 2.2, < 2.7.dev") provided by each gem to resolve suitable version. 

See also: https://github.com/bundler/bundler/pull/7522#issuecomment-571090513

https://rubygems.org/info/ffi returns the following information:

```text
1.11.3-x86-mingw32 |checksum:99e4845796c8dec1c3fc80dc772860a01633b33291bd7534007f5c7724f0b876,ruby:>= 2.2, < 2.7.dev
```

The current parser doesn't parse the line correctly. It drops "< 2.7.dev" requirement for Ruby version.

It means that the parsed requirement is "ruby >= 2.2" instead of "ruby >=2.2 && ruby < 2.7.dev".

I implemented a new parsing logic for "requirements". It assumes that "requirements" has at most three fields: "checksum", "ruby" and "rubygems". Because https://github.com/rubygems/compact_index/blob/master/lib/compact_index/gem_version.rb#L26-L28 implements so.

I chose this fix because it's difficult to implement generic parsing logic without "," escape in version. For example, "," leading to "2.2" and "," leading to "2.7.dev" in "ruby:>= 2.2, < 2.7.dev,rubygems:> 1.3.1" are conflict. We need to know field name such as "ruby" and "rubygems" to detect "," as field delimiter.

"," in "dependencies" are escaped as "&": https://github.com/rubygems/compact_index/blob/master/lib/compact_index/gem_version.rb#L37

### DIFF
--- a/lib/bundler/compact_index_client/cache.rb
+++ b/lib/bundler/compact_index_client/cache.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_relative "gem_parser"
+
 module Bundler
   class CompactIndexClient
     class Cache
@@ -92,19 +94,9 @@ module Bundler
         header ? lines[header + 1..-1] : lines
       end
 
-      def parse_gem(string)
-        version_and_platform, rest = string.split(" ", 2)
-        version, platform = version_and_platform.split("-", 2)
-        dependencies, requirements = rest.split("|", 2).map {|s| s.split(",") } if rest
-        dependencies = dependencies ? dependencies.map {|d| parse_dependency(d) } : []
-        requirements = requirements ? requirements.map {|r| parse_dependency(r) } : []
-        [version, platform, dependencies, requirements]
-      end
-
-      def parse_dependency(string)
-        dependency = string.split(":")
-        dependency[-1] = dependency[-1].split("&") if dependency.size > 1
-        dependency
+      def parse_gem(line)
+        @dependency_parser ||= GemParser.new
+        @dependency_parser.parse(line)
       end
 
       def info_roots
--- a/lib/bundler/compact_index_client/cache.rb
+++ b/lib/bundler/compact_index_client/cache.rb
@@ -95,8 +95,8 @@ module Bundler
       end
 
       def parse_gem(line)
-        @dependency_parser ||= GemParser.new
-        @dependency_parser.parse(line)
+        @gem_parser ||= GemParser.new
+        @gem_parser.parse(line)
       end
 
       def info_roots
--- a/lib/bundler/compact_index_client/gem_parser.rb
+++ b/lib/bundler/compact_index_client/gem_parser.rb
@@ -1,0 +1,66 @@
+# frozen_string_literal: true
+
+module Bundler
+  class CompactIndexClient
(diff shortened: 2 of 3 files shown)


lib/bundler/compact_index_client/cache.rb
```diff
@@ -92,19 +94,9 @@ def lines(path)
...
-        requirements = requirements ? requirements.map {|r| parse_dependency(r) } : []
-        [version, platform, dependencies, requirements]
-      end
-
-      def parse_dependency(string)
-        dependency = string.split(":")
-        dependency[-1] = dependency[-1].split("&") if dependency.size > 1
-        dependency
+      def parse_gem(line)
+        @dependency_parser ||= GemParser.new
```
colby-swandale: Could this just be a class method to save from having instance var & instance object?

kou: Do you mean defining methods in `GemParser` as class methods instead of instance methods?

I don't think that it's a good design. Class method only class is one of global objects. Less global objects are better.

If we need to use class method only class, we should use module like `Bundler::SharedHelpers`.

BTW, do we need the instance saving optimization here? Normally, there is one `Bundler::CompactIndexClient::Cache` object in a process. So I don't think that we don't need it.

deivid-rodriguez: Is it intentional to call the instance variable `@dependency_parser` instead of `@gem_parser`?

deivid-rodriguez: I agree with @kou. @colby-swandale can you mark this as resolved if you're happy with @kou's reasoning?

kou: Oh, sorry. It's my fault.
I'll fix this.

At first I was implementing `GemParser` as `DependencyParser` then I renamed it to improve naming.
But I forgot to change this instance variable name.

kou:
Travis CI is failed.

https://travis-ci.org/rubygems/bundler/jobs/635459321#L532

```
       expected `["Rack::File is deprecated, please use Rack::Files instead."].empty?` to return true, got false
```

This is caused by rack-2.1.0 released at 2020-01-10. Sinatra uses `Rack::File`.

Rack removed this warning: https://github.com/rack/rack/commit/f80e65d5dde251ee446e4c0bd038f8bc4ec30314

This failure will be gone when rack-2.1.1 is released.

What should we do? Pin rack to 2.0.8 until rack-2.1.1 is released? Just ignore for now? rack-2.1.1 will be released soon.

deivid-rodriguez:
I'll pin the version :point_right: #7565.

kou:
Thanks!
Rebased.

deivid-rodriguez:
@colby-swandale I'm merging this PR to unblock others, but we can continue the code style discussion and do a followup PR if necessary.

@bundlerbot merge

deivid-rodriguez:
@bundlerbot retry

deivid-rodriguez:
@bundlerbot retry

PR #7558, opened 2020-01-09
Remove usused code and simplify concat method

larskanis:
The unused code is a leftover from commit 38b0e7ed64c3ca1c40f43c5aa9a1ead2f6cd7049 .
That commit disabled that RubyVersion.system is respected, but it seems that that feature wasn't missed.

Not known.

I read the bundler source code and noticed this.

My fix removes the code that seems to be of no use.

As a alternative I added #7559 that reactivates the code in question.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -858,11 +858,7 @@ module Bundler
 
     def metadata_dependencies
       @metadata_dependencies ||= begin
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Went ahead with the other PR, so closing. Thanks @larskanis!

PR #7559, opened 2020-01-09
Reactivate usused code for ruby version resolution

larskanis:
The code was changed in commit 38b0e7ed64c3ca1c40f43c5aa9a1ead2f6cd7049 so that RubyVersion.system was no longer respected. This is reactivated now and specs are adjusted accordingly.

Not known.

I read the bundler source code and noticed this.

My fix re-adds the ruby version as it was before commit 38b0e7ed64c3ca1c40f43c5aa9a1ead2f6cd7049 , since it seems to be a mistake.

As a alternative I added #7558 that removes the code in question.

### DIFF
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -860,8 +860,8 @@ module Bundler
       @metadata_dependencies ||= begin
         ruby_versions = concat_ruby_version_requirements(@ruby_version)
         if ruby_versions.empty? || !@ruby_version.exact?
-          concat_ruby_version_requirements(RubyVersion.system)
-          concat_ruby_version_requirements(locked_ruby_version_object) unless @unlock[:ruby]
+          concat_ruby_version_requirements(RubyVersion.system, ruby_versions)
+          concat_ruby_version_requirements(locked_ruby_version_object, ruby_versions) unless @unlock[:ruby]
         end
         [
           Dependency.new("Ruby\0", ruby_versions),
--- a/spec/install/gems/resolving_spec.rb
+++ b/spec/install/gems/resolving_spec.rb
@@ -194,7 +194,7 @@ RSpec.describe "bundle install with install-time dependencies" do
 
       describe "with a < requirement" do
         let(:ruby_requirement) { %("< 5000") }
-        let(:error_message_requirement) { "< 5000" }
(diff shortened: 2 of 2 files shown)


deivid-rodriguez:
Thanks @larskanis! Do you have any particular preference? Intuitively to me, less code is better than more code, so I would go with #7558. @kou any opinion?

kou:
I like #7558 too if we don't have regression by https://github.com/rubygems/bundler/commit/38b0e7ed64c3ca1c40f43c5aa9a1ead2f6cd7049. (I don't confirm it yet. Sorry.)

But I want to rename `concat_ruby_version_requirements` in #7558. Because it doesn't `concat` anything.

larskanis:
I honesty don't really understand the impact of this code and I didn't find a difference in resolution between #7558 and #7559. But there's at least a user visible difference. In case of ruby version conflicts with #7522 this PR prints:
```
$ ruby -Ilib exe/bundle install
Fetching gem metadata from https://rubygems.org/............
Resolving dependencies...
Bundler found conflicting requirements for the Ruby  version:
  In Gemfile:
    Ruby  (= 2.7.0.0) x64-mingw32

    ffi (~> 1.11) x64-mingw32 was resolved to 1.11.3, which depends on
      Ruby  (>= 2.0)
```
while #7558 doesn't print the current ruby version:
```
$ ruby -Ilib exe/bundle install
Resolving dependencies...
Bundler found conflicting requirements for the Ruby  version:
  In Gemfile:
    Ruby  x64-mingw32

    ffi (~> 1.11) x64-mingw32 was resolved to 1.11.3, which depends on
      Ruby  (>= 2.0)
```

kou:
OK. I'll check this with #7522.

deivid-rodriguez:
Mmm, I see. In that case, it sounds like this PR would be the way to go. At least I prefer the extra bit of clarification in the failure message.

deivid-rodriguez:
@larskanis Can you rebase this PR?

larskanis:
@deivid-rodriguez I rebased this PR and #7558 to master.

deivid-rodriguez:
I'm going to go with this PR out of the two, let me know if any of you have any concerns.

@bundlerbot merge

PR #7565, opened 2020-01-11
Lock rack to 2.0.8 for tests

deivid-rodriguez:

The problem was CI is failing due to a rack warning.

My diagnosis was that our CI shouldn't break because of third party changes.

My fix is to lock the rack version our tests use.

I chose this fix because it gives us a more deterministic test environment, which is good.

### DIFF
--- a/spec/support/rubygems_ext.rb
+++ b/spec/support/rubygems_ext.rb
@@ -15,7 +15,7 @@ module Spec
     }.freeze
 
     DEPS = {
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
@bundlerbot merge

deivid-rodriguez:
@bundlerbot retry

deivid-rodriguez:
@bundlerbot retry

PR #7568, opened 2020-01-12
Make parallel tests more stable

deivid-rodriguez:

The problem was that since we run parallel tests by default, our CI is much less stable even though specs are faster.

My diagnosis was that our specs change folders back and forth for every test, and that's not thread safe.

My fix is to stop changing folders in the main process during our specs.

I chose this fix because it should improve stability of our specs.


deivid-rodriguez:
@bundlerbot merge

PR #7569, opened 2020-01-13
Load `CompactIndexClient` upfront during specs

deivid-rodriguez:

The problem was failing CI:  https://github.com/rubygems/bundler/runs/386607433.

My diagnosis was that it could be caused by autoloading not being thread safe. The `CompactIndexClient` constant is autoloaded. If two parallel threads try to autoload the constant at the same time, it could happen that one of them sees the constant in an incomplete state (no `initialize` method defined yet).

My fix is to load the constant upfront.

I chose this fix because even if I could stop autoloading the constant, I don't want to touch code in `lib` at the moment just to fix a testing issue.

### DIFF
--- a/spec/bundler/fetcher/compact_index_spec.rb
+++ b/spec/bundler/fetcher/compact_index_spec.rb
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+# load CompactIndexClient upfront to prevent thread safety issues during parallel specs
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Seems to fix the issue #7554 is running into for me, so merging!

@bundlerbot merge

kou:
Wow! Thanks!

PR #7570, opened 2020-01-13
Use "Ruby" and "RubyGems" instead of "ruby" and "rubygems" with compact index

kou:

The problem was "Ruby" requirement isn't used correctly on solving gems.

See also:

  * https://github.com/rubygems/bundler/pull/7522#issuecomment-572635007
  * https://github.com/rubygems/bundler/pull/7522#issuecomment-572850885

My diagnosis was using "ruby" and "rubygems" is compact index isn't good. Because we use "Ruby" and "RubyGems" in other places.

See also #7554 for compact index content.

My fix uses "Ruby" and "RubyGems" explicitly for requirement keys instead of using keys in compact index directly.

I chose this fix because we already use "Ruby" and "RubyGems" in other places.

### DIFF
--- a/lib/bundler/compact_index_client/gem_parser.rb
+++ b/lib/bundler/compact_index_client/gem_parser.rb
@@ -54,10 +54,10 @@ module Bundler
         end
         ruby = raw_requirements.match(/\A(ruby):(.+)\z/)
         if ruby
-          requirements << [ruby[1], ruby[2].split(/\s*,\s*/)]
+          requirements << ["Ruby", ruby[2].split(/\s*,\s*/)]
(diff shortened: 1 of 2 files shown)


kou:
This may not be needed.
I'm looking into this...

kou:
Sorry. This is needles.
We need to use "ruby" and "rubygems" here. We use them in `Bundler::EndpointSpecification`: https://github.com/rubygems/bundler/blob/master/lib/bundler/endpoint_specification.rb#L120-L123

deivid-rodriguez:
I see, thanks.

PR #7574, opened 2020-01-15
Improve PR template

deivid-rodriguez:

The problem was that the current PR template sometimes asks too many questions, and the answers are a bit artificial.

My diagnosis was that experience tell me that some questions are too much, and a bit artificial to answer for many PRs.

My fix is to still let contributors know what kind of information we find valuable, but don't force all questions to be answered.


deivid-rodriguez:
@indirect You ok with these changes?

indirect:
@bundlerbot merge

deivid-rodriguez:
>  If you think these changes continue towards those goals, that sounds good to me. :)

Yes, I do, thanks @indirect!

PR #7578, opened 2020-01-17
Don't ignore resolved platform information

kou:
This is a follow-up change of #7522.

The problem was installing wrong platform gem even when resolver resolves the correct platform gem.

See also:

  * https://github.com/rubygems/bundler/pull/7522#issuecomment-574628275
  * https://github.com/rubygems/bundler/pull/7522#issuecomment-574882536

My diagnosis was `Bundler::LazySpecification#__materialize__` doesn't care about the resolved platform information.

My fix cares about the resolved platform information.

I chose this fix because we should respect the resolved platform information to install correct gem.

### DIFF
--- a/lib/bundler/match_platform.rb
+++ b/lib/bundler/match_platform.rb
@@ -15,7 +15,10 @@ module Bundler
       return true if Gem::Platform::RUBY == gemspec_platform
       return true if local_platform == gemspec_platform
       gemspec_platform = Gem::Platform.new(gemspec_platform)
-      return true if GemHelpers.generic(gemspec_platform) === local_platform
+      generic_gemspec_platform = GemHelpers.generic(gemspec_platform)
+      unless generic_gemspec_platform == Gem::Platform::RUBY
+        return true if generic_gemspec_platform === local_platform
+      end
       return true if gemspec_platform === local_platform
 
       false
--- a/lib/bundler/match_platform.rb
+++ b/lib/bundler/match_platform.rb
@@ -15,10 +15,7 @@ module Bundler
       return true if Gem::Platform::RUBY == gemspec_platform
       return true if local_platform == gemspec_platform
       gemspec_platform = Gem::Platform.new(gemspec_platform)
-      generic_gemspec_platform = GemHelpers.generic(gemspec_platform)
-      unless generic_gemspec_platform == Gem::Platform::RUBY
-        return true if generic_gemspec_platform === local_platform
-      end
+      return true if GemHelpers.generic(gemspec_platform) === local_platform
       return true if gemspec_platform === local_platform
 
       false
--- a/spec/install/gems/resolving_spec.rb
+++ b/spec/install/gems/resolving_spec.rb
@@ -143,6 +143,29 @@ RSpec.describe "bundle install with install-time dependencies" do
         expect(out).to_not include("rack-9001.0.0 requires ruby version > 9000")
         expect(the_bundle).to include_gems("rack 1.2")
       end
+
+      it "installs the older not platform specific version" do
+        build_repo4 do
+          build_gem "rack", "9001.0.0" do |s|
+            s.required_ruby_version = "> 9000"
+          end
+          build_gem "rack", "1.2" do |s|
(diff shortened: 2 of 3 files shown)


lib/bundler/match_platform.rb
```diff
@@ -15,7 +15,10 @@ def self.platforms_match?(gemspec_platform, local_platform)
       return true if Gem::Platform::RUBY == gemspec_platform
       return true if local_platform == gemspec_platform
       gemspec_platform = Gem::Platform.new(gemspec_platform)
-      return true if GemHelpers.generic(gemspec_platform) === local_platform
+      generic_gemspec_platform = GemHelpers.generic(gemspec_platform)
+      unless generic_gemspec_platform == Gem::Platform::RUBY
```
kou: This isn't needed for platforms that are listed in `Bundler::GemHelpers::GENERICS`. Because these platforms are resolved to a `Gem::Platform` object. Platforms that aren't listed such as `x86_64-linux` are resolved to `Gem::Platform::RUBY`. See `Bundler::GemHelpers.generic` for details: https://github.com/rubygems/bundler/blob/master/lib/bundler/gem_helpers.rb#L21

If `Gem::Platform::RUBY` is returned, it's matched for `local_platform` is `Gem::Platform::RUBY`. It means that this method returns `true` for `x86_64-linux` gem on `ruby` platform. But a platform specific gem such as a gem for `x86_64-linux` doesn't work all platform (= `ruby` platform).

So I added this condition.

kou: Umm... I think that this is a bug but some existing examples are failed with this change:

https://travis-ci.org/rubygems/bundler/jobs/638212688#L1303

```text
rspec ./spec/cache/gems_spec.rb:236 # bundle cache when previously cached doesn't remove gems that are for another platform
rspec ./spec/commands/install_spec.rb:220 # bundle install with gem sources the simple case with a gem that installs multiple platforms installs gems for the local platform as first choice
rspec ./spec/commands/update_spec.rb:39 # bundle update with --all updates the entire bundle
rspec ./spec/commands/update_spec.rb:49 # bundle update with --all doesn't delete the Gemfile.lock file if something goes wrong
rspec ./spec/commands/update_spec.rb:104 # bundle update with a top level dependency unlocks all child dependencies that are unrelated to other locked dependencies
rspec ./spec/commands/update_spec.rb:115 # bundle update with an unknown dependency should inform the user
rspec ./spec/commands/update_spec.rb:119 # bundle update with an unknown dependency should suggest alternatives
rspec ./spec/commands/update_spec.rb:126 # bundle update with a child dependency should update the child dependency
rspec ./spec/commands/update_spec.rb:299 # bundle update in a frozen bundle should suggest different command when frozen is set globally
rspec ./spec/runtime/platform_spec.rb:60 # Bundler.setup with multi platform stuff will add the resolve for the current platform
```

So I've reverted this change for now. And the new example uses `simulate_platform mingw`. It means that the new example is failed without `simulate_platform` on `x86_64-linux`.

spec/install/gems/resolving_spec.rb
```diff
@@ -143,6 +143,31 @@
...
+          end
+          build_gem "rack", "1.2" do |s|
+            s.platform = mingw
+            s.required_ruby_version = "> 9000"
+          end
+          build_gem "rack", "1.2"
+        end
+
+        simulate_platform mingw do
+          install_gemfile <<-G, :artifice => "compact_index_rate_limited", :env => { "BUNDLER_SPEC_GEM_REPO" => gem_repo4.to_s }
```
deivid-rodriguez: The API mock here is very specific to test API fallback behavior when the modern API responds with 429 code (too many requests). I'd use the "compact_index" mock instead which is more general and I think more appropriate for this test.

```suggestion
          install_gemfile <<-G, :artifice => "compact_index", :env => { "BUNDLER_SPEC_GEM_REPO" => gem_repo4.to_s }
```

kou: You're right.
Sorry. I didn't care about it.

deivid-rodriguez:
@bundlerbot merge

PR #7580, opened 2020-01-17
Automultiplatform again

deivid-rodriguez:
This PR is a reintroduction of #7215, now that I consider the multiplatform feature usable enough.

The problem was that currently if the gemfile include gems for specific platforms, like the default `gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]` Rails uses, bundler will always print a noisy warning, but I don't think users are doing anything wrong, just running `bundle install` on it.

Also, unless they run `bundle lock --add-platform x86-mingw32 x86-mswin32 x64-mingw32 java`, and commit the resulting lockfile, they will continue to get the same warning over and over again.

My diagnosis is that the fact that some gems will be unused under some platforms is inherent to the multiplatform feature itself, so we shouldn't warn about it because it's expected.

Take the following Gemfile for example (a simplification of the default Gemfile Rails generators create):

```ruby
source "https://rubygems.org"

gem "rails"

gem "tzinfo-data", platforms: "jruby"
```

If I type that Gemfile, it means that I'm explicitly opting into the multiplatform feature. So the behavior I would want as a user when I run `bundle install` on it is:

* Resolve and lock the Gemfile for all platforms (`jruby` and whatever platform I'm running).
* Install the resolution for the platform that I'm currently running.

That way, when the other developers of my team using `jruby` install the `Gemfile`, they install a predictable, deterministic resolution.

Currently, the only way to get that behavior is to run:

```
$ bundle install
$ bundle lock --add-platform java
```

But there's no way to do it without getting a warning on the first `bundle install`, which annoys people because it makes you think you're doing something wrong.

If you only plan to use MRI, you shouldn't specify any jruby-specific gems in your Gemfile and write instead:

```ruby
source "https://rubygems.org"

gem "rails"
```

If on the other hand you only plan to use jruby, then you should not specify any `platform` option at all and write

```ruby
source "https://rubygems.org"

gem "rails"

gem "tzinfo-data"
```

So, to sum up, I think the range of platforms users expect to support can be guessed implicit from the Gemfile and the running platform. So we should resolve by default for all those platforms and don't warn about behavior that's most likely expected.

My fix is to do the right thing by default, and not warn users at all. That is:

* Resolve the gemfile and lock it for all platforms present in the Gemfile.
* Install gems for the current platform if requested.

I chose this fix because I think it's better for our users. We currently have a specific setting to avoid this warning, which I guess we added given the complaints. We no longer need that setting, nor the warning, so I removed both.

### DIFF
--- a/lib/bundler/cli/outdated.rb
+++ b/lib/bundler/cli/outdated.rb
@@ -76,6 +76,8 @@ module Bundler
         next unless gems.empty? || gems.include?(current_spec.name)
 
         active_spec = retrieve_active_spec(definition, current_spec)
+        next unless active_spec
+
         next unless filter_options_patch.empty? || update_present_via_semver_portions(current_spec, active_spec, options)
 
         gem_outdated = Gem::Version.new(active_spec.version) > Gem::Version.new(current_spec.version)
@@ -144,6 +146,8 @@ module Bundler
     end
 
     def retrieve_active_spec(definition, current_spec)
+      return unless current_spec.match_platform(Bundler.local_platform)
+
       if strict
         active_spec = definition.find_resolved_spec(current_spec)
       else
@@ -231,8 +235,6 @@ module Bundler
     end
 
     def update_present_via_semver_portions(current_spec, active_spec, options)
-      return false if active_spec.nil?
-
       current_major = current_spec.version.segments.first
       active_major = active_spec.version.segments.first
 
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -523,10 +523,6 @@ module Bundler
       raise InvalidOption, "Unable to remove the platform `#{platform}` since the only platforms are #{@platforms.join ", "}"
     end
 
-    def add_current_platform
-      current_platforms.each {|platform| add_platform(platform) }
-    end
-
     def find_resolved_spec(current_spec)
       specs.find_by_name_and_platform(current_spec.name, current_spec.platform)
     end
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -117,7 +117,7 @@ module Bundler
       end
       @unlocking ||= @unlock[:ruby] ||= (!@locked_ruby_version ^ !@ruby_version)
 
-      add_current_platform unless Bundler.frozen_bundle?
+      add_platforms unless Bundler.frozen_bundle?
 
       converge_path_sources_to_gemspec_sources
       @path_changes = converge_paths
@@ -547,6 +547,12 @@ module Bundler
     end
 
   private
+
+    def add_platforms
+      (@dependencies.flat_map(&:expanded_platforms) + current_platforms).uniq.each do |platform|
+        add_platform(platform)
+      end
+    end
 
     def current_platforms
       current_platform = Bundler.local_platform
--- a/lib/bundler/definition.rb
+++ b/lib/bundler/definition.rb
@@ -898,17 +898,7 @@ module Bundler
       dependencies.each do |dep|
         dep = Dependency.new(dep, ">= 0") unless dep.respond_to?(:name)
         next if !remote && !dep.current_platform?
-        platforms = dep.gem_platforms(sorted_platforms)
-        if platforms.empty? && !Bundler.settings[:disable_platform_warnings]
-          mapped_platforms = dep.expanded_platforms
-          Bundler.ui.warn \
-            "The dependency #{dep} will be unused by any of the platforms Bundler is installing for. " \
-            "Bundler is installing for #{@platforms.join ", "} but the dependency " \
-            "is only for #{mapped_platforms.join ", "}. " \
-            "To add those platforms to the bundle, " \
-            "run `bundle lock --add-platform #{mapped_platforms.join " "}`."
-        end
-        platforms.each do |p|
+        dep.gem_platforms(sorted_platforms).each do |p|
           deps << DepProxy.new(dep, p) if remote || p == generic_local_platform
         end
       end
--- a/lib/bundler/inline.rb
+++ b/lib/bundler/inline.rb
@@ -58,7 +58,7 @@ def gemfile(install = false, options = {}, &gemfile)
 
       Bundler.ui = install ? ui : Bundler::UI::Silent.new
       if install || definition.missing_specs?
-        Bundler.settings.temporary(:inline => true, :disable_platform_warnings => true) do
+        Bundler.settings.temporary(:inline => true) do
           installer = Bundler::Installer.install(Bundler.root, definition, :system => true)
           installer.post_install_messages.each do |name, message|
             Bundler.ui.info "Post-install message from #{name}:\n#{message}"
--- a/lib/bundler/settings.rb
+++ b/lib/bundler/settings.rb
@@ -22,7 +22,6 @@ module Bundler
       disable_exec_load
       disable_local_branch_check
       disable_multisource
-      disable_platform_warnings
       disable_shared_gems
       disable_version_check
       force_ruby_platform
--- a/man/bundle-config.1
+++ b/man/bundle-config.1
@@ -187,9 +187,6 @@ The following is a list of all configuration keys and their purpose\. You can le
 \fBdisable_multisource\fR (\fBBUNDLE_DISABLE_MULTISOURCE\fR): When set, Gemfiles containing multiple sources will produce errors instead of warnings\. Use \fBbundle config unset disable_multisource\fR to unset\.
 .
 .IP "\(bu" 4
-\fBdisable_platform_warnings\fR (\fBBUNDLE_DISABLE_PLATFORM_WARNINGS\fR): Disable warnings during bundle install when a dependency is unused on the current platform\.
-.
-.IP "\(bu" 4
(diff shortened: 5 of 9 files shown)


deivid-rodriguez:
@kou @larskanis This is the last important PR I plan to include with bundler 2.2. What do you think about it? Do you think it's a good move?

kou:
I think that his is a good move too.

How about always including "ruby" platform?

Gemfile:

```ruby
source "https://rubygems.org/"
gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
```

`Gemfile.lock` on Debian GNU/Linux ("ruby" included):

```text
GEM
  remote: https://rubygems.org/
  specs:
    concurrent-ruby (1.1.5)
    tzinfo (2.0.1)
      concurrent-ruby (~> 1.0)
    tzinfo-data (1.2019.3)
      tzinfo (>= 1.0.0)

PLATFORMS
  java
  ruby
  x64-mingw32
  x86-mingw32
  x86-mswin32

DEPENDENCIES
  tzinfo-data

BUNDLED WITH
   2.2.0.dev
```

`Gemfile.lock` on Windows (no "ruby"):

```text
GEM
  remote: https://rubygems.org/
  specs:
    concurrent-ruby (1.1.5)
    tzinfo (2.0.1)
      concurrent-ruby (~> 1.0)
    tzinfo-data (1.2019.3)
      tzinfo (>= 1.0.0)

PLATFORMS
  java
  x64-mingw32
  x86-mingw32
  x86-mswin32

DEPENDENCIES
  tzinfo-data

BUNDLED WITH
   2.2.0.dev
```

deivid-rodriguez:
That makes sense to me yeah. I think it's good that `bundle install` on different platforms doesn't generate source control changes in the lock file. I'll work on it.

deivid-rodriguez:
@kou I added https://github.com/rubygems/bundler/pull/7580/commits/4edd84ffe39b4f62571f104d99eaccd650e89620 to implement your suggestion. It required a few spec changes, but maybe it's fine.

A case that this would break is if a gem provides no standard "ruby" alternative, and only platform specific gems. But I'm not sure whether somebody is actually doing that, or whether it's something that makes sense to do at all. This was actually caught by the specs and I needed to add a "ruby-platform" version of the dummy "rcov" gem to get them to pass.

Thoughts?

kou:
I've never seen a gem that provides only platform specific gems too.
So I confirmed real data:

```ruby
ARGF.gets
ARGF.gets

gem_versions = {}
ARGF.each_line do |line|
  name, versions, = line.split(" ")
  versions = versions.split(",").delete_if do |version|
    version.start_with?("-")
  end
  gem_versions[name] ||= []
  gem_versions[name].concat(versions)
end

gem_versions.each do |name, versions|
  grouped_versions = versions.group_by do |version|
    version.split("-")[0]
  end
  grouped_versions.each do |grouped_version, full_versions|
    unless full_versions.any? {|full_version| full_version.split("-", 2)[1].nil?}
      p [name, grouped_version, full_versions]
    end
  end
end
```

```shell
curl https://rubygems.org/versions | ruby a.rb
```

```text
["Auto", "4.0.0.alpha.1", ["4.0.0.alpha.1-x86-mingw32"]]
["Box2d", "1.2.0", ["1.2.0-mswin32"]]
["Box2d", "1.4.3", ["1.4.3-mswin32"]]
["Etch", "0.1.1", ["0.1.1-mswin32"]]
["Etch", "0.1.2", ["0.1.2-mswin32"]]
["Etch", "0.1.3", ["0.1.3-mswin32"]]
["LinkGrouper", "1.0.1", ["1.0.1-universal-linux"]]
["LinkGrouper", "1.0.2", ["1.0.2-universal-linux"]]
["POpen4-mingw", "0.1.6", ["0.1.6-x86-mingw32"]]
["Postcard", "0.0.0", ["0.0.0-x64-mingw32"]]
["Queen_Of_Diamonds", "0.3", ["0.3-mswin32"]]
["Queen_Of_Diamonds", "0.4", ["0.4-mswin32"]]
["RInotify", "0.9", ["0.9-i586-linux"]]
["Rlabel", "0.0.0", ["0.0.0-x86_64-darwin-12"]]
["SNMP4JR", "0.0.20", ["0.0.20-java"]]
["UDJrb", "0.0.1", ["0.0.1-universal-java-1.6"]]
["WaveSwissKnife", "0.0.1.20101110", ["0.0.1.20101110-x86-cygwin"]]
["WaveSwissKnife", "0.1.0.20110830", ["0.1.0.20110830-x86-cygwin"]]
["active_documentum", "0.3.2", ["0.3.2-java"]]
["active_documentum", "0.3.3", ["0.3.3-java"]]
["activerecord-jdbc-adapter", "0.9.3", ["0.9.3-java"]]
["activerecord-jdbc-adapter", "0.9.4", ["0.9.4-java"]]
["activerecord-jdbc-adapter", "0.9.5", ["0.9.5-java"]]
["activerecord-jdbc-adapter", "0.9.6", ["0.9.6-java"]]
["activerecord-jdbc-adapter", "0.9.7", ["0.9.7-java"]]
["activerecord-jdbc-adapter", "1.0.0", ["1.0.0-java"]]
...
```

There are many platform specific only gems...

It may be better that we don't always include "ruby" platform...

BTW, does automultiplatform work well with MRI and "activerecord-jdbc-adapter"?

deivid-rodriguez:
Wow, thanks so much for double checking! :heart_eyes:. I removed the changes to always include `ruby` platform. 

> BTW, does automultiplatform work well with MRI and "activerecord-jdbc-adapter"?

Yeah, it seems to be doing what I'd expect. Properly resolve and lock dependencies for jruby, but don't try to install `activerecord-jbdc-adapter` at all since it doesn't belong in the "MRI resolution".

Something else I noticed is that both the recent resolution fixes and this PR seem to affect bundler's resolution time. And that this slow down seems to affect all Gemfiles and all platforms... Thoughts about this?

kou:
I also confirmed with MRI and "activerecord-netezza-adapter" that has only java specific gems https://rubygems.org/gems/activerecord-netezza-adapter/versions ("activerecord-jbdc-adapter" has not platform specific gem https://rubygems.org/gems/activerecord-jdbc-adapter/versions/51.0 ).

Gemfile without `:platforms` is failed:

```ruby
source "https://rubygems.org/"
gem "activerecord-netezza-adapter"
```

Gemfile with `:platforms` works: 

```ruby
source "https://rubygems.org/"
gem "activerecord-netezza-adapter", platforms: [:jruby]
```

Is this expected behavior?

If this is expected, it seems that we can enable 4edd84f again. Gemfile with `:platforms` and 4edd84f works too.

> Something else I noticed is that both the recent resolution fixes and this PR seem to affect bundler's resolution time. And that this slow down seems to affect all Gemfiles and all platforms... Thoughts about this?

Do you already have any benchmark?

I think that we should focus on "works well" at first. Then we work on "works well and is fast".

How about releasing Bundler 2.2.0 without optimization? If there is a real world use case that isn't assumed by us, we need to support the use case before optimization.

If Bundler 2.2.0 works well in the real world, we can optimize our resolution logic and release Bundler 2.2.X or 3.0.0 as faster Bundler. Benchmark will be useful in optimization phase.

deivid-rodriguez:
Under MRI, what you showed is expected behaviour for those Gemfiles. The problem is when running that under jruby:

### With 4edd84f

```
$ bundle _2.2.0.dev_
Fetching gem metadata from https://rubygems.org/..................
Could not find gem 'activerecord-netezza-adapter' in any of the gem sources
listed in your Gemfile.
```

### Without 4edd84f

```
$ bundle _2.2.0.dev_
Fetching gem metadata from https://rubygems.org/..................
Resolving dependencies....
Using concurrent-ruby 1.1.5
Using i18n 1.8.2
Using minitest 5.14.0
Using thread_safe 0.3.6 (java)
Using tzinfo 1.2.6
Using zeitwerk 2.2.2
Using activesupport 6.0.2.1
Using activemodel 6.0.2.1
Using activerecord 6.0.2.1
Using activerecord-jdbc-adapter 60.1 (java)
Using activerecord-netezza-adapter 0.1.1 (java)
Using bundler 2.2.0.dev
Bundle complete! 1 Gemfile dependency, 12 gems now installed.
```

Regarding performance, I totally agree with you. First correctness, then speed. Also, full resolution is not run very often since you normally have a resolved lockfile, so hopefully it's not too bad.

Regarding benchmarking, I only did some tests using a couple of realworld Gemfiles. For example, with [activeadmin's main Gemfile](https://github.com/activeadmin/activeadmin/blob/9da88933d40ec29eaef3091054ef71ecb760c15f/Gemfile#L1) on my computer full resolution (`rm Gemfile.lock && bundle lock`) takes:

* ~3 seconds with 2.1.4.
* ~5 seconds with current master.
* ~10 seconds with this PR.

kou:
Thanks! I understand.
I agree with dropping 4edd84f.

It seems that activeadmin's Gemfile has enough many dependencies. I don't think that ~10s is too bad.

deivid-rodriguez:
Sorry I left this PR hang in there for a while, I'll rebase it now and get it in! Thanks for having a look @indirect and for having fixed all the resolver stuff @kou.

deivid-rodriguez:
@bundlerbot r=indirect

PR #7584, opened 2020-01-17
Fix flaky test failure

deivid-rodriguez:

The problem was that sometimes a spec is failing on Windows because the rename syscall sometimes is raising Errno::EXDEV.

I don't really know why this problem is happening.

My fix is to stop renaming the folder and copy it instead.

I chose this fix because copying will no longer make a rename syscall, so at least it should no longer fail in the same way.


deivid-rodriguez:
Thanks for the explanation!

@bundlerbot merge

PR #7588, opened 2020-01-18
Drop ruby 2.3 and rubygems 2.5 support

### LINKED ISSUE #7586: Drop support for ruby 2.3
I'd like to drop support for ruby 2.3, maybe in the upcoming 2.2 release, maybe right after.

deivid-rodriguez:
Closes #7586. 

The problem was that ruby 2.3 reached its end of life almost a year ago and we still support it.

My diagnosis was that we should drop support.

My fix is to drop support.

I chose this fix because it's part of our ruby support policy.

### DIFF
--- a/doc/POLICIES.md
+++ b/doc/POLICIES.md
@@ -17,7 +17,7 @@ Bundler tries for perfect backwards compatibility. That means that if something 
 
 Bundler will provide features and bugfixes to older versions on a schedule similar to Ruby itself. For example, when Bundler 4.x is the current version, Bundler 4 will be eligible for new features and bugfixes. Bundler 3 will be eligible for bugfixes only. Bundler 2 will be eligible for security bugfixes only. Bundler 1 will be unsupported.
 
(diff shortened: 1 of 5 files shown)


doc/POLICIES.md
```diff
@@ -17,7 +17,7 @@ Bundler tries for perfect backwards compatibility. That means that if something
...
 
-Bundler 2 and above will support Ruby and RubyGems versions for the same amount of time as the Ruby core team supports them. As of February 2018, that means no support for Bundler running on Ruby 2.2, security fixes only for Bundler running on Ruby 2.3, and full support (including new features and bugfixes) for Bundler running on Ruby 2.4 and 2.5. Unsupported Ruby versions will be dropped in the first Bundler minor release after support ends.
+Bundler 2 and supports Ruby and RubyGems versions for the same amount of time as the Ruby core team supports them. As of February 2020, that means no support for Bundler running on Ruby 2.3, security fixes only for Bundler running on Ruby 2.4, and full support (including new features and bugfixes) for Bundler running on Ruby 2.5, 2.6 and 2.7. Unsupported Ruby versions will be dropped in the first Bundler minor release after support ends.
```
kou: Can we remove "and" in "Bundler 2 and"?

kou (approved):
It may be needed tiny fix.

doc/POLICIES.md
deivid-rodriguez: Thanks, fixed!

PR #7589, opened 2020-01-18
Add support for BUNDLER_ prefixed debug environment variables

kou:
DEBUG_RESOLVER is used in RubyGems too. So we can't enable it only for
Bundler.

The problem was not the end-user problem. It's a developer problem.

My diagnosis was RubyGems also uses `DEBUG_RESOLVER` environment variable. So we can't use `DEBUG_RESOLVER` environment variable to enable debug mode only for Bundler. We can use `DEBUG_RESOLVER_TREE` instead of `DEBUG_RESOLVER` only for `Bundler::Resolver`. But we can't do it for `Bundler::GemVersionPromoter`.

My fix adds `BUNDLER_` prefix to `DEBUG_RESOLVER` and `DEBUG_RESOLVER_TREE` like other environment variables for Bundler such as `BUNDLER_EDITOR` do.

I chose this fix because adding prefix is a common way. Bundlerd Molinillo and Thor also use this way such as `MOLINILLO_DEBUG` and `THOR_SHELL`.

### DIFF
--- a/lib/bundler/gem_version_promoter.rb
+++ b/lib/bundler/gem_version_promoter.rb
@@ -7,7 +7,7 @@ module Bundler
   # available dependency versions as found in its index, before returning it to
   # to the resolution engine to select the best version.
   class GemVersionPromoter
-    DEBUG = ENV["DEBUG_RESOLVER"]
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
Thanks! :heart_eyes:

@bundlerbot r=colby-swandale

PR #7590, opened 2020-01-19
Prepend debug label to resolver's debug message

kou:

The problem was not a end-user problem. This was a developer problem.

If we have many debug messages (for example, I added many `pp [...]` to debug #7522), it's difficult to find resolver's debug messages.

My diagnosis was no common keywords in resolver's debug message is a problem. If we have a common keyword, we can find resolver's debug messages from many debug messages easily.

My fix prepends `BUNDLER: ` to all resolver's debug messages. If we have the same label in all resolver's debug messages, we can find resolver's debug message easily. For example, we can use "/BUNDLER:" in less.

I chose this fix because `BUNDLER: ` will not conflict. Other candidate is `DEBUG: ` but it may be used by other library. If the keyword conflicts, it's not easy to find resolver's debug messages. `Bundler::Resolver: ` will be more safer but it may be long.

### DIFF
--- a/lib/bundler/resolver.rb
+++ b/lib/bundler/resolver.rb
@@ -75,7 +75,7 @@ module Bundler
       return unless debug?
       debug_info = yield
       debug_info = debug_info.inspect unless debug_info.is_a?(String)
-      warn debug_info.split("\n").map {|s| "  " * depth + s }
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
@bundlerbot merge

deivid-rodriguez:
Hopefully the CI is a little more stable now.

@bundlerbot merge

PR #7591, opened 2020-01-20
Revise compatibility policies

indirect:

The POLICIES file was a bit contradictory, saying in one place that Ruby versions would be dropped in minor releases, and in another place that Ruby versions would be dropped in major releases.

The POLICIES file was not fully updated after several Bundler and RubyGems team discussions, agreements, and public statements during 2019.

This PR updates the POLICIES file to be consistent both with itself and with my understanding of the agreements and statements made by the Bundler and RubyGems teams in 2019.

/cc @deivid-rodriguez @colby-swandale

### DIFF
--- a/doc/POLICIES.md
+++ b/doc/POLICIES.md
@@ -25,7 +25,7 @@ These policies are not a guarantee that any particular fix will be backported. I
 
 Bugfix releases should generally be cut as soon as possible. A bugfix release for a single PR is totally okay.
 
-Minor/feature releases can be cut anytime at least one new feature is ready, but don't have to be. Minor version releases should include an update to the documentation website.
+Minor/feature releases can be cut anytime at least one new feature is ready, but don't have to be. Minor version releases must also include any needed updates to the documentation website, plus a "What's new" page summarizing what was added.
 
 Major version releases should be cut no more than once per year, and must include a new section of the website with documentation for that major version.
 
--- a/doc/POLICIES.md
+++ b/doc/POLICIES.md
@@ -29,7 +29,7 @@ Minor/feature releases can be cut anytime at least one new feature is ready, but
 
 Major version releases should be cut no more than once per year, and must include a new section of the website with documentation for that major version.
 
-Major releases ideally happen after a Ruby version loses support in February or March, which helps us stay in sync with Ruby versions uspported by the core team.  
+Major releases ideally happen after a Ruby version loses support in February or March, which helps us stay in sync with Ruby versions uspported by the core team.
 
 Breaking changes other than dropping support for old Ruby versions should be avoided whenever possible, but may be included in major releases. In general, breaking changes should include at least one major version (and one year elapsed) with a deprecation warning before the breaking change takes effect.
 
--- a/doc/POLICIES.md
+++ b/doc/POLICIES.md
@@ -23,13 +23,13 @@ These policies are not a guarantee that any particular fix will be backported. I
 
 ### Release guidelines
 
-Bugfix releases should generally be cut as soon as possible. A bugfix release for a single PR is totally okay.
+**tl;dr**: Majors about once per year, minors for any finished features with docs, patches for any committed bugfix.
 
-Minor/feature releases can be cut anytime at least one new feature is ready, but don't have to be. Minor version releases must also include any needed updates to the documentation website, plus a "What's new" page summarizing what was added.
+Patch (bugfix) releases should generally be cut as soon as possible. A patch release for a single bugfix PR is totally okay.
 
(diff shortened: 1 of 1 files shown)


doc/POLICIES.md
```diff
@@ -17,10 +17,22 @@ Bundler tries for perfect backwards compatibility. That means that if something
...
 
+### Release guidelines
+
+Bugfix releases should generally be cut as soon as possible. A bugfix release for a single PR is totally okay.
+
+Minor/feature releases can be cut anytime at least one new feature is ready, but don't have to be. Minor version releases should include an update to the documentation website.
+
+Major version releases should be cut no more than once per year, and must include a new section of the website with documentation for that major version.
+
+Major releases ideally happen after a Ruby version loses support in February or March, which helps us stay in sync with Ruby versions uspported by the core team.
```
deivid-rodriguez: There's trailing whitespace here, according to our CI :)

indirect: Oh no 😅 One second...

doc/POLICIES.md
```diff
@@ -17,10 +17,22 @@ Bundler tries for perfect backwards compatibility. That means that if something
...
+Bundler supports Ruby and RubyGems versions until the next major release after the Ruby core team drops support. For example, the Ruby core team will drop all Ruby 2.4 support on March 31, 2020. The next Bundler major release after that date will drop support for Ruby 2.4.
 
 These policies are not a guarantee that any particular fix will be backported. Instead, this is a way for us to set an upper limit on the versions of Ruby, RubyGems, and Bundler that we have to consider while making changes. Without the limit, the number of versions grows exponentially over time and quickly becomes overwhelming, which leads to maintainer burnout. We want to avoid that.
 
+### Release guidelines
+
+Bugfix releases should generally be cut as soon as possible. A bugfix release for a single PR is totally okay.
+
+Minor/feature releases can be cut anytime at least one new feature is ready, but don't have to be. Minor version releases should include an update to the documentation website.
```
deivid-rodriguez: I don't think I updated the documentation website for bundler 2.1. Not sure if someone did it, but we probably should if not at some point.

indirect: @deivid-rodriguez after thinking about this more, I think we should probably aim for one "version" site for each Bundler major version. Since minor versions are purely additive, we can put things in the docs like "Starting in Bundler 2.2, you can X". That way we don't have to make a completely new copy of the website except for major version changes. What do you think?

deivid-rodriguez: Hei @indirect. I haven't worked enough with our documentation website to have a strong opinion, but I don't mind putting up a new copy of docs explaining the relevant additions after every minor. I just don't know how that process works, and hadn't realized about this up until now.

The current way the documentation website is presented looks pretty neat to me, but if we decide to change it, we need to discuss how to do it:

In https://bundler.io/docs.html, how do users find out about new stuff in minors? I guess we should add new stuff to the "What's new in 2.0" section? It seems to me that if docs are "additive" and you can only browse one documentation copy per major version, that doesn't remove the need to update the documentation website on minor releases to "add the new stuff" to the major version documentation, right?

indirect: @deivid-rodriguez ah, my apologies—I should have explained the docs site situation context. In the 1.x versions, we created a completely new copy of the website for every minor version. The goal was to guarantee that you could always find accurate documentation even if you hadn't upgraded to the latest version available.

That approach also turns out to have some downsides:
- the documentation site has many duplicate pages, even when there are no changes between versions
- build times on the documentation site are getting quite slow because of the many copies of the documentation
- search results often point to outdated documentation pages—for example, searching for "bundle help" on Google shows results only for Bundler 1.10 and 1.12, and they are not clearly labelled :/

Now that we are aiming to release one new major version per year, I was imagining that it would make sense to switch to one copy of the docs site per major version, with minor versions simply adding new content to the existing pages for that major version.

I do think it makes sense to add a "What's new" page for 2.1 and 2.2, even if we don't create a new copy of the entire docs site for each of those minor versions.

deivid-rodriguez: I see, thanks!

Well, then the approach looks good to me, yeah. In that case, we should probably reword this to say something like

> Minor version releases should include an update to the documentation website, adding a new "What's new page", and updating the corresponding major version documentation to include the new features, with a specific mentions of the minor version where the are becoming available.

Or something like that?

indirect: @deivid-rodriguez done! what do you think?

deivid-rodriguez: Looks great!

indirect:
@bundlerbot merge

PR #7594, opened 2020-01-21
Fix flaky spec failure

deivid-rodriguez:

The problem was flaky tests: https://www.travis-ci.org/rubygems/bundler/jobs/639403049.

My diagnosis was that if our test code happens to run `Gem.configuration` before `ENV["HOME"]` is changed for our tests, this test starts failing because the test first writes to the global `~/.gemrc` file (not in `tmp/`), because `Gem.config_file` has that value memoized. Then, however, the `bundle install` subprocess infers `Gem.config_file` from the modified `ENV["HOME"]` and tries to read `tmp/home/.gemrc`. However, that file doesn't exist, so rubygems configuration is ignored and doesn't print an error like the test expects.
    
My fix is to make this test independent from the specific rubygems configuration in the main test process.

### DIFF
--- a/spec/bundler/friendly_errors_spec.rb
+++ b/spec/bundler/friendly_errors_spec.rb
@@ -7,13 +7,13 @@ require "cgi"
 RSpec.describe Bundler, "friendly errors" do
   context "with invalid YAML in .gemrc" do
     before do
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
@bundlerbot merge

PR #7597, opened 2020-01-23
Don't call Warn on LoadError

egiurleo:
Related to #7192 and #7527. When trying to run bundler without OpenSSL, the resulting warning recommends recompiling Ruby, changing the Gemfile sources to http, and then gives an RVM-specific link that won't be helpful for everyone.

I think the error message is now specific enough to stand on its own and help people debug, so we should remove the warning entirely. (Open to other opinions, though.)


deivid-rodriguez:
@bundlerbot merge

PR #7600, opened 2020-01-24
Improve cache specs

deivid-rodriguez:

I noticed that when running specs on Windows, sometime I got a `_gem` leftover folder in the root of the repo.

My fix is to run the particular spec creating this test folder in `tmp`, just like the rest of the specs.


deivid-rodriguez:
@bundlerbot merge

PR #7609, opened 2020-01-28
Simplify arguments to `sh`

nobu:
Get rid of escaping and splitting shell code repeatedly.

No changes for the end-users.

As `built_gem_path` in `install_gems` is not escaped, this may fix a trouble in the case the argument have shell special characters.


deivid-rodriguez:
@bundlerbot merge

PR #7614, opened 2020-01-29
Workaround jruby issue

### LINKED ISSUE #7598: Bundler 2.1.4 errors with different prefix for path source gems
When running my Ruby code via JRuby -- I've encountered an interesting failure specifically on **2.1.4** whereas **2.0.2** does not have the issue.

The failure at startup is:
```
eval_gemfile at uri:classloader:/gems/bundler-2.1.4/lib/bundler/dsl.rb:53
          evaluate at uri:classloader:/gems/bundler-2.1.4/lib/bundler/dsl.rb:12
             build at uri:classloader:/gems/bundler-2.1.4/lib/bundler/definition.rb:34
        definition at uri:classloader:/gems/bundler-2.1.4/lib/bundler.rb:195
             setup at uri:classloader:/gems/bundler-2.1.4/lib/bundler.rb:143
   block in (root) at uri:classloader:/gems/bundler-2.1.4/lib/bundler/setup.rb:20
        with_level at uri:classloader:/gems/bundler-2.1.4/lib/bundler/ui/shell.rb:136
           silence at uri:classloader:/gems/bundler-2.1.4/lib/bundler/ui/shell.rb:88
            <main> at uri:classloader:/gems/bundler-2.1.4/lib/bundler/setup.rb:20

[!] There was an error parsing `Gemfile`: different prefix: "uri:classloader://" and
"uri:classloader:/my-application". Bundler cannot continue.
```

The _Gemfile_ looks like:

```ruby
gem 'some-gem', :path => "ruby-gems/some-gem"
```

I have not _git-bisected_ yet but auditing the code; I have strong intuition to believe the issue is caused by https://github.com/rubygems/bundler/commit/c7532ced89bbc8ddc7296c56391c1c3cc981c54a
as it introduces `relative_path_from` for _path_ Gems.

## Environment

```
jruby 9.1.17.0 (2.3.3) 2018-04-20 d8b1ff9 OpenJDK 64-Bit Server VM 25.232-b09 on 1.8.0_232-b09 +jit [darwin-x86_64]

Darwin MacBook-Pro.localdomain 18.7.0 Darwin Kernel Version 18.7.0: Tue Aug 20 16:57:14 PDT 2019; root:xnu-4903.271.2~2/RELEASE_X86_64 x86_64
```

deivid-rodriguez:

The problem is that if a Gemfile contains `:path` gem with a relative path, the bundler environment no longer works when the app is packaged as a `jar` with [warbler](https://github.com/jruby/warbler).

Issue was introduced https://github.com/rubygems/bundler/commit/c7532ced89bbc8ddc7296c56391c1c3cc981c54a, since when `jruby` deals with urls of the form "uri:classloader", it has a weird inconsistency where `Pathname#expand_path` with an argument always returns `uri:classloader://` (double slash) as the canonical version of the path, while `Pathname#expand_path` without an argument always returns `uri:classloader/` (single slash) as the canonical version of it.

```
Pathname.new("uri:classloader://foo/bar").expand_path
#  => <Pathname:uri:classloader:/foo/bar>

Pathname.new("uri:classloader:/foo/bar").expand_path
# => <Pathname:uri:classloader:/foo/bar>

Pathname.new("foo/bar").expand_path("uri:classloader://")
# => <Pathname:uri:classloader://foo/bar>

Pathname.new("foo/bar").expand_path("uri:classloader:/")
# => <Pathname:uri:classloader://foo/bar>
```

That makes `Pathname#relative_path_from` (introduced with the offending commit) explode because we end up passing to different "uri:classloader" kind of paths to it.

I believe this should be fixed in `jruby` be doing either or both of the following things:

* Make `Pathname#expand_path` return consistent versions of "uri:classpath" paths.
* Make `Pathname#relative_path_from` support different versions of "uri:classpath" paths, since they are both absolute paths after all.

But I'm workarounding the issue in `bundler` by adding an extra `expand_path` call at an appropriate place to make sure that the URLs we pass to `Pathname#relative_path_from` have a consistent shape.

Fixes #7598.

NOTE: We currently don't have the ability to run a full test suite for every PR on jruby because it's too slow (#4796 attempted to improve that but it was never completed), and it doesn't even fully pass (#7603). As an alternative, I'm adding some realworld bare tests to be run under jruby, just to make sure common tasks like this one work.

### DIFF
--- a/.github/workflows/jruby.yml
+++ b/.github/workflows/jruby.yml
@@ -1,0 +1,31 @@
+name: jruby
+
+on:
+  pull_request:
+
+  push:
+    branches:
+      - staging
+      - trying
+
+jobs:
+  warbler:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v1
+
+      - name: Setup ruby
+        uses: eregon/use-ruby-action@v1
+        with:
+          ruby-version: jruby-9.2.9.0
+
+      - name: Install local bundler
+        run: bin/rake install:local
+
+      - name: Run a warbler project
+        run: |
+          cd spec/realworld/fixtures/warbler
+          bundle install
+          bundle exec warble
+          java -jar warbler.jar
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -408,6 +408,8 @@ Naming/ConstantName:
 
 Naming/FileName:
   Enabled: true
+  Exclude:
+    - 'spec/realworld/fixtures/warbler/bin/warbler-example.rb'
 
 Naming/HeredocDelimiterCase:
   Enabled: true
--- a/lib/bundler/source/path.rb
+++ b/lib/bundler/source/path.rb
@@ -132,7 +132,11 @@ module Bundler
       end
 
       def expand(somepath)
-        somepath.expand_path(root_path)
+        if Bundler.current_ruby.jruby? # TODO: Unify when https://github.com/rubygems/bundler/issues/7598 fixed upstream and all supported jrubies include the fix
+          somepath.expand_path(root_path).expand_path
+        else
+          somepath.expand_path(root_path)
+        end
       rescue ArgumentError => e
         Bundler.ui.debug(e)
         raise PathError, "There was an error while trying to use the path " \
--- a/spec/realworld/fixtures/warbler/.gitignore
+++ b/spec/realworld/fixtures/warbler/.gitignore
@@ -1,0 +1,2 @@
+*.jar
+Gemfile.lock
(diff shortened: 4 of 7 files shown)


.github/workflows/jruby.yml
```diff
@@ -0,0 +1,31 @@
...
+
+      - name: Install local bundler
+        run: bin/rake install:local
+
+      - name: Run a warbler project
+        run: |
+          cd spec/realworld/fixtures/warbler
+          bundle install
+          bundle exec warble
+          java -jar warbler.jar
```
fzakaria: great!

I think you can just add a second inner gem `demo-test` and have this same workflow and leverage this test setup for https://github.com/rubygems/bundler/issues/7611

fzakaria:
LGTM

.github/workflows/jruby.yml
deivid-rodriguez: Thanks, I'll keep that in mind :+1:

deivid-rodriguez:
@bundlerbot merge

Issue #7581, opened 2020-01-17
Update codeclimate to use correct repository


duckinator:
In the process of updating GitHub for #7577, I discovered that CodeClimate is still configured for `bundler/bundler`. If someone sets up `rubygems/bundler` on CodeClimate (I'm not sure if there's a way to tell them a repo has moved, or if you just need to set it up again), I can update the link. :+1:

hsbt:
@duckinator I added rubygems/bundler to codeclimate again.

see https://codeclimate.com/github/rubygems/bundler

Is it ok for your order?

duckinator:
@hsbt yup, that's what I needed! Thank you!

PR #7619, opened 2020-02-01
[docs] Update CodeClimate link.

duckinator:
This PR updates the CodeClimate link in `doc/contributing/HOW_YOU_CAN_HELP.md`, now that #7581 ("Update CodeClimate to use correct repository") is resolved.


duckinator:
I'm not sure why Travis CI is failing the way it is. It seems like it thinks the man pages are outdated, even though the _only_ change is the month mentioned in the generated files lmao

deivid-rodriguez:
I really need to fix this, but I'm lazy every month and just update it. The idea of the man pages check is to check that the different formats are always in sync with their source of truth (the `.ronn` files). However, the build tasks also updates the man pages with the current month, so they fall out of sync every month.

deivid-rodriguez:
@bundlerbot merge

PR #7620, opened 2020-02-01
February man pages

deivid-rodriguez:

No actual issue, just keeping the man pages current because the `build:check` task is failing.

Run `bin/rake man:build`.


deivid-rodriguez:
@bundlerbot merge

PR #7629, opened 2020-02-09
Adhere to gemfile name preference also for project skeleton.

### LINKED ISSUE #7626: init_gems_rb setting is ignored in bundle-gem
When **new** project skeleton is generated with `bundle gem`, it does not heed any of the following options

`init_gems_rb == true`
`prefer_gems_rb == true`

and creates `Gemfile` instead of expected `gems.rb` .

You have to change to projects directory, invoke `bundle init` which creates `gems.rb` afterwards, then throws following warning:

> Multiple gemfiles (gems.rb and Gemfile) detected. Make sure you remove Gemfile and Gemfile.lock since bundler is ignoring them in favor of gems.rb and gems.rb.locked.

and have to remove `Gemfile` manually.
It should not happen, while skeleton should pick `gems.rb` directly, in accordance to the preferences.

dunrix:
Fixes #7626.

Resolving gemfile name preference - original `Gemfile` or newer `gems.rb` moved to Bundler's singleton as being shared by multiple classes, ie. because of DRY principle.

### DIFF
--- a/lib/bundler.rb
+++ b/lib/bundler.rb
@@ -457,6 +457,10 @@ EOF
       Bundler.settings[:system_bindir] || Bundler.rubygems.gem_bindir
     end
 
+    def preferred_gemfile_name
+      Bundler.settings[:init_gems_rb] ? "gems.rb" : "Gemfile"
+    end
+
     def use_system_gems?
       configured_bundle_path.use_system_gems?
     end
--- a/lib/bundler/cli/init.rb
+++ b/lib/bundler/cli/init.rb
@@ -41,7 +41,7 @@ module Bundler
   private
 
     def gemfile
-      @gemfile ||= Bundler.settings[:init_gems_rb] ? "gems.rb" : "Gemfile"
+      @gemfile ||= Bundler.preferred_gemfile_name
     end
   end
 end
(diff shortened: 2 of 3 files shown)


lib/bundler.rb
```diff
@@ -451,6 +451,10 @@ def system_bindir
       Bundler.settings[:system_bindir] || Bundler.rubygems.gem_bindir
     end
 
+    def gemfile
```
deivid-rodriguez: There's a `Bundler.default_gemfile` method which returns the full path to the default `gems.rb` or `Gemfile` file, so I think introducing this method could create confusion.

Maybe a `Bundler.settings.preferred_gems_file_name` method instead?

```ruby
def preferred_gems_file_name
  self[:init_gems_rb] ? "gems.rb" : "Gemfile"
end
```

lib/bundler/cli/init.rb
```diff
@@ -38,10 +38,10 @@ def run
       puts "Writing new #{gemfile} to #{SharedHelpers.pwd}/#{gemfile}"
     end
 
-  private
+    private
```
deivid-rodriguez: Can you remove this change?

deivid-rodriguez:
Thanks @dunrix!

@bundlerbot merge

PR #7630, opened 2020-02-11
More paralelization improvements

deivid-rodriguez:

The developer problem is that CI is still failing sometimes. The reason is most likely due to side effects when parallelizing specs, since specs still change the current folder globally sometimes.

My changes stop changing folders globally in more places, to try to alleviate these issues.


deivid-rodriguez:
Merging this with the hope of making CI more stable. :crossed_fingers: 

@bundlerbot merge

PR #7635, opened 2020-02-16
Update code of conduct template to CC2.0

CoralineAda:
The Contributor Covenant was updated to version 2.0 last year, but bundler was still using the old version.

I updated the `.md.tt` file containing the template, careful to preserve the interpolation of the email address.

### DIFF
--- a/lib/bundler/templates/newgem/CODE_OF_CONDUCT.md.tt
+++ b/lib/bundler/templates/newgem/CODE_OF_CONDUCT.md.tt
@@ -2,73 +2,83 @@
 
 ## Our Pledge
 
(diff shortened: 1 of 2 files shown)


colby-swandale:
the CoC template needs to be added to the ignore list in the quality spec

https://github.com/rubygems/bundler/blob/13eefdc8fd166862ded7425c45270dab810003b2/spec/quality_spec.rb#L161

colby-swandale:
Thanks! @bundlerbot r+

colby-swandale:
@bundlerbot r+

CoralineAda:
Thanks for the support!

PR #7644, opened 2020-02-18
Unfreeze dependency hashes

deivid-rodriguez:

Running `bin/rake spec:deps` from jruby crashes with:

```
rake aborted!
FrozenError: can't modify frozen Hash
/path/to/bundler/spec/support/rubygems_ext.rb:36:in `dev_setup'
/path/to/bundler/Rakefile:29:in `block in <main>'
/path/to/bundler/Rakefile:14:in `block in invoke'
/path/to/bundler/Rakefile:13:in `invoke'
/path/to/bundler/spec/support/rubygems_ext.rb:98:in `gem_load_and_activate'
/path/to/bundler/spec/support/rubygems_ext.rb:45:in `gem_load'
Tasks: TOP => spec:deps
```

Unfreeze the hashes, since we are modifying them.

### DIFF
--- a/spec/support/rubygems_ext.rb
+++ b/spec/support/rubygems_ext.rb
@@ -4,7 +4,7 @@ require_relative "path"
 
 module Spec
   module Rubygems
-    DEV_DEPS = {
+    DEV_DEPS = { # rubocop:disable Style/MutableConstant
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
@bundlerbot merge

PR #7645, opened 2020-02-19
Use consistent white space in Gemfile template

mathisto:
`Gemfile` template has a trailing space before closing curly of block, but no leading space at the opening. Super nitpicky, but thought it was worthwhile considering it is going to be present in any use of `bundle init`. Not sure if you guys are religious about [this pattern](https://github.com/rubocop-hq/ruby-style-guide#blocks-procs--lambdas) from the style guide.

I added a single leading white space to open the block.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -363,7 +363,7 @@ Layout/SpaceInsideArrayPercentLiteral:
 
 Layout/SpaceInsideBlockBraces:
   Enabled: true
-  SpaceBeforeBlockParameters: false
+  SpaceBeforeBlockParameters: true
 
 Layout/SpaceInsideHashLiteralBraces:
   Enabled: true
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -361,6 +361,10 @@ Layout/SpaceInsideArrayLiteralBrackets:
 Layout/SpaceInsideArrayPercentLiteral:
   Enabled: true
 
+Layout/SpaceInsideBlockBraces:
+  Enabled: true
+  SpaceBeforeBlockParameters: true
+
 Layout/SpaceInsideHashLiteralBraces:
   Enabled: true
 
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -361,10 +361,6 @@ Layout/SpaceInsideArrayLiteralBrackets:
 Layout/SpaceInsideArrayPercentLiteral:
   Enabled: true
 
-Layout/SpaceInsideBlockBraces:
-  Enabled: true
-  SpaceBeforeBlockParameters: true
-
 Layout/SpaceInsideHashLiteralBraces:
   Enabled: true
 
--- a/Rakefile
+++ b/Rakefile
@@ -91,7 +91,7 @@ namespace :man do
       directory "man"
 
       index = []
-      sources = Dir["man/*.ronn"].map {|f| File.basename(f, ".ronn") }
+      sources = Dir["man/*.ronn"].map { |f| File.basename(f, ".ronn") }
       sources.map do |basename|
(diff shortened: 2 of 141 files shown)


colby-swandale:
I don't think we're religious about the space before the block argument, but you will need to remove the `Layout/SpaceInsideBlockBraces` rule from the Rubocop config

mathisto:
> I don't think we're religious about the space before the block argument, but you will need to remove the `Layout/SpaceInsideBlockBraces` rule from the Rubocop config

Turns out `Layout/SpaceInsideBlockBraces` should probably stay if it is desired to merge this. I did toggle the nested cop beneath it (`SpaceBeforeBlockParameters: true`) which was causing the Rubocop complaint during Travis' build attempt. 

Let me know if this is acceptable or if I really should just pull the whole cop.

deivid-rodriguez:
This is a duplicate of https://github.com/rubygems/bundler/pull/6860.

I'm happy to change our style and make this change, because I prefer the style, but also to make the generated gem layout consistent in style.

@indirect said he prefers the current style but that he's fine with switching in https://github.com/rubygems/bundler/pull/6860#issuecomment-450803326. So I would do this since it has come up a few times now.

mathisto:
> > I don't think we're religious about the space before the block argument, but you will need to remove the `Layout/SpaceInsideBlockBraces` rule from the Rubocop config
> 
> Turns out `Layout/SpaceInsideBlockBraces` should probably stay if it is desired to merge this. I did toggle the nested cop beneath it (`SpaceBeforeBlockParameters: true`) which was causing the Rubocop complaint during Travis' build attempt.
> 
> Let me know if this is acceptable or if I really should just pull the whole cop.

Looks like the entire codebase is riddled with this stylistic choice. Let me see if I can do some regex sub magic without breaking this in half. @colby-swandale ,I am assuming this is why you said just to gank the rule altogether. LOL. Mea culpa.

deivid-rodriguez:
`bin/rubocop --auto-correct` should do the trick.

mathisto:
> `bin/rubocop --auto-correct` should do the trick.

Ooooh, you tricksy!

mathisto:
I ganked the cop, as @colby-swandale recommended, which I think will work. Would you rather it stay, with the child enabled to match the RSG desires, and then update the codebase with `bin/rubocop --auto-correct` as you recommended? I can see a desire for either solution. If you ditch the cop, it won't be enforced, but you can keep using the style for the gem while promulgating the more common variation via the template.

mathisto:
Just ran the `bin/rubocop --auto-correct`. This PR just bloated in scope by a factor of 182 😆.
`364 files inspected, 457 offenses detected, 457 offenses corrected`
Changes pushed. Should be consistent across the board now.

deivid-rodriguez:
Wow, I guess I didn't stop to think how common it is to use single line hashes :sweat_smile:.

I'm personally :+1: on this, but I understand any reluctancy from other maintainers to switch away from a style we have sticked to for years. So I'm happy to close this too in case other maintainers disagree.

deivid-rodriguez:
Thanks for working on this by the way :purple_heart:.

mathisto:
> Thanks for working on this by the way 💜.

Dude, my pleasure. I owe the Ruby community a few beers as it is. You, specifically, even more so. My career in Ruby/Rails land started by spinning up an instance of `active_admin` for my old employer. Cheers to you, sir.

bpleslie (approved):
LGTM! Thanks for the contribution!

brand-it (approved):
❤️

hsbt:
@mathisto Bundler repo was merged into rubygems. If you still interested in this PR, Can you re-open this on https://github.com/rubygems/rubygems? Thanks.

PR #7646, opened 2020-02-19
Pin jruby-jars in warbler test

deivid-rodriguez:

The problem is that our CI is broken after the jruby-9.2.10.0 release.

My workaround is to pin the jruby-jars dependency of the warbler test to 9.2.9.0.


deivid-rodriguez:
@bundlerbot merge

deivid-rodriguez:
@bundlerbot retry

PR #7649, opened 2020-02-20
Fix resolving specs when run against a dev ruby snapshot

deivid-rodriguez:

The problem is that in some dev ruby builds, `RUBY_PATCHLEVEL` is `-1`, so the spec crashes because of an ill-formed requirement.

See for example https://github.com/rubygems/rubygems/pull/3077/checks?check_run_id=457988784.

My fix is to change the spec to respect a `RUBY_PATCHLEVEL` equal to `-1`, by reusing code that handles that case.


deivid-rodriguez:
@bundlerbot merge

PR #7650, opened 2020-02-21
Bump rake version to 13.0 everywhere

deivid-rodriguez:

The problem is rake 12 prints a bunch of warnings on ruby 2.7.

My fix is to upgrade rake everywhere.


deivid-rodriguez:
@bundlerbot merge

PR #7661, opened 2020-02-27
Revise CoC based on Contributor Covenant 2.0

indirect:
Bundler has had a code of conduct since before the Contributor Covenant was available as an option. In my (admittedly somewhat vague) memory, we continued to use our own custom CoC based on wanting clear enforcement examples directly in the CoC. The Contributor Covenant now contains enforcement examples as of 2.0, so I am proposing that we migrate Bundler's CoC to CC 2.0.

Additional benefits:
  - Merging with RubyGems gets easier, since RubyGems uses CC already (rubygems/rubygems#3157).
  - The `bundle gem` template and Bundler itself will now use the same CoC.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,136 +1,81 @@
 # Bundler Code of Conduct
 
-## When Something Happens
-
-If you see a Code of Conduct violation, follow these steps:
-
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Let's do it! Thanks @indirect!

@bundlerbot merge

PR #7664, opened 2020-03-04
Bump man pages month

deivid-rodriguez:

CI fails when man pages month gets out of date.

My fix is to update the month in the man pages to the current one.


deivid-rodriguez:
@bundlerbot merge

PR #7660, opened 2020-02-27
Try to migrate GitHub Actions from Travis CI

hsbt:
Fixes #7587

PR #7665, opened 2020-03-04
Fix sudo specs environment

deivid-rodriguez:

The developer problem is that sudo specs are not correctly setting the environment (except for ruby 2.3). The specs are accidentally passing but you can tell that they are running against the system ruby by the [warning messages they print](https://travis-ci.org/rubygems/bundler/jobs/658272719#L531-L558).

This causes problems when migrating to github actions. See https://github.com/rubygems/bundler/pull/7660#discussion_r385616844.

Another problem is that the environment is setup manually outside of `rake spec:sudo`, so there's no easy way of running sudo specs manually if your system does not have the right sudoers configuration.

My fix is to properly set sudo configuration to preserve the environment so that the correct ruby can be found by sudo specs, and also to move the logic inside `rake spec:sudo` so that the specs can easily be run manually without side effects.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: ruby
 dist: bionic
 script:
   - bin/parallel_rspec spec
+  - sudo sed -i '/secure_path/d' /etc/sudoers
   - test "$(ruby -v)" = "$(sudo -E ruby -v)"
   - sudo -E bin/rake spec:sudo
   - sudo chown -R $(whoami) tmp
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
@hsbt Is it ok if I merge this PR, so you can merge the sudo workflow with the main workflow in #7660?

deivid-rodriguez:
@bundlerbot r=hsbt

PR #7666, opened 2020-03-04
Fix `rake spec:parallel_deps` under ruby 2.3

deivid-rodriguez:

The problem is that `rake spec:parallel_deps` does not work under ruby 2.3. The only reason it's working on TravisCI is because the TravisCI configuration include `travis_retry` on `rake spec:parallel_deps` _does_ work on the second try.

I'm not sure why this is happening, but I would say it's a `Kernel.system` issue fixed on newer rubies. The reason for this is that this issue seems rubygems version independent. 

My fix is to install development dependencies programatically without shelling out. It should be faster, and it avoids the issue.

I'm also removing some unneeded `travis_retry` from the TravisCI configure. These steps shouldn't timeout and if they do, we should figure out why. Also, they could be hiding other issues not related to the network, like this one.

This fixes https://github.com/rubygems/bundler/pull/7660#discussion_r385603908.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,11 @@ script:
   - BUNDLER_SPEC_PRE_RECORDED=1 bin/rake spec:realworld
 
 before_script:
-  - travis_retry bin/rake override_version
-  - travis_retry bin/rake spec:parallel_deps
+  - bin/rake override_version
+  - bin/rake spec:parallel_deps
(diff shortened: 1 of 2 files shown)


hsbt:
@deivid-rodriguez Thanks. Your commit was working fine with my working branch. see https://github.com/rubygems/bundler/pull/7660#commits-pushed-b4dd50a

Can we close this pull-request for merging this?

deivid-rodriguez:
Sure!

